PR #23, opened 2024-01-22
Clang-tidy fixes

snnn:


src/models/model.cpp
```diff
@@ -50,7 +50,7 @@ Model::Model(std::unique_ptr<Config> config, OrtEnv& ort_env, const ProviderOpti
 
 Model::~Model() = default;
 
-void Model::InitDeviceAllocator(OrtSession& session) {
+void Model::InitDeviceAllocator(OrtSession& /*session*/) {
```
RyanUnderhill, line 53: Won't this break when USE_CUDA is true?

snnn: Changed to maybe unused.

RyanUnderhill (approved):

PR #26, opened 2024-01-23
Enable Onnxruntime-Extensions tokenizer support

RyanUnderhill:
With the build option USE_ORT_EXT you can then use model.CreateTokenizer() in Python or C++


CMakeLists.txt
```diff
@@ -20,6 +20,7 @@ if (CMAKE_CUDA_COMPILER)
     set (CUDA_FOUND TRUE)
 endif()
 set(USE_CUDA 1)
+set(USE_ORT_EXT 0) # Use the onnxruntime extensions library to support built in tokenizing
```
snnn, line 23: It's better to declare all such options at the beginning of the cmake file as cmake options.  Like this: https://cmake.org/cmake/help/latest/command/option.html . 
Otherwise it's hard to what build options this project has and how to use them.

snnn: Here is a detailed explanation: 

https://cliutils.gitlab.io/modern-cmake/chapters/basics/variables.html

snnn (approved):

PR #39, opened 2024-01-26
Merge Search & State into new "Generator" type

RyanUnderhill:
```
  // Original version
  auto search = params.CreateSearch();
  auto state = model->CreateState(search->GetSequenceLengths(), params);

  while (!search->IsDone()) {
    search->SetLogits(state->Run(search->GetSequenceLength(), search->GetNextTokens()));
    search->SelectTop();
  }
```

```
  // Generator version
  auto generator = model->CreateGenerator();

  while (!generator->IsDone()) {
    generator->ComputeLogits();
    generator->AppendNextToken_Top();
  }
```


examples/llama.py
```diff
@@ -10,48 +10,45 @@
...
 
 # Keep asking for input prompts in an loop
 while True:
     text = input("Input:")
-    # input_tokens = tokenizer.encode(text, return_tensors='np')
-    input_tokens = tokenizer.encode(text)
+    input_tokens = tokenizer.encode(text, return_tensors='np')
+    # input_tokens = tokenizer.encode(text)
 
     params=og.SearchParams(model)
```
yufenglee, line 25: could you please also change SearchParams to GeneratorParams?

examples/llama.py
```diff
@@ -10,48 +10,45 @@
...
-    while not search.IsDone():
-        search.SetLogits(state.Run(search.GetSequenceLength(), search.GetNextTokens()))
+    while not generator.IsDone():
+        generator.ComputeLogits()
 
         # search.Apply_MinLength(1)
         # search.Apply_RepetitionPenalty(1.0)
 
-        search.SampleTopP(0.7, 0.6)
+        generator.AppendNextToken_TopP(0.7, 0.6)
```
yufenglee, line 40: Append doesn't look like a good name to me. generator actually generate next token.

yufenglee: And can we make TopP a parameter instead of part of the name?

examples/llama.py
```diff
@@ -10,48 +10,45 @@
...
     print("Output:")
 
     print(text, end='', flush=True)
-    while not search.IsDone():
-        search.SetLogits(state.Run(search.GetSequenceLength(), search.GetNextTokens()))
+    while not generator.IsDone():
+        generator.ComputeLogits()
 
         # search.Apply_MinLength(1)
         # search.Apply_RepetitionPenalty(1.0)
```
yufenglee, line 38: could you please update this also

examples/llama.py
RyanUnderhill: I named it that since you first compute the logits, then append the next token based on the logits.
I initially had AddNextToken but 'Add' was less correct than 'Append' since it's appending tokens, not adding them.
Did you have a better name?

examples/llama.py
```diff
@@ -10,48 +10,45 @@
...
 
-        search.SampleTopP(0.7, 0.6)
+        generator.AppendNextToken_TopP(0.7, 0.6)
 
-        print(tokenizer.decode([search.GetNextTokens().GetArray()[0]]), ' ', end='', flush=True)
-        '''
+        # print(tokenizer.decode([generator.GetNextTokens().GetArray()[0]]), ' ', end='', flush=True)
         # Print each token as we compute it, we have to do some work to get newlines & spaces to appear properly:
-        word=tokenizer.convert_ids_to_tokens([search.GetNextTokens().GetArray()[0]])[0]
+        word=tokenizer.convert_ids_to_tokens([generator.GetNextTokens().GetArray()[0]])[0]
```
yufenglee, line 44: GetNexTokens looks fine. Is it better to use GetNewGeneratedTokens() considering that the tokens are already generated?

yufenglee: And we can make Generate(Append)NextTokens to return new generated tokens too

src/generators.h
```diff
@@ -124,6 +104,52 @@ struct SearchParams {
   std::variant<Whisper> inputs;
 };
 
+struct Search {
```
yufenglee, line 107: rename it to Sample?

examples/llama.py
yufenglee: It is a process of generate next token. Add or Append doesn't reflect the action, i think.

src/generators.h
RyanUnderhill: This code does the Greedy/Beam search, the sampling is just a tiny part of it. But I might be able to hide it from generators.h

examples/llama.py
RyanUnderhill: GetLastTokens()? Then it works in every case.

aciddelgado (approved):

PR #47, opened 2024-01-31
Rename AppendNextToken* to GenerateNextToken*

RyanUnderhill:
To be more intuitive, as we're not passing in tokens to append, we're generating the tokens based on parameters.

(there's also an internal AppendNextTokensToSequences that it could be confused with, which does append tokens passed in)

### DIFF
--- a/src/python/python.cpp
+++ b/src/python/python.cpp
@@ -178,24 +178,24 @@ struct PyGenerator {
     generator_->ComputeLogits();
   }
 
-  void AppendNextToken_TopK_TopP(int top_k, float top_p, float temperature) {
-    generator_->AppendNextToken_TopK_TopP(top_k, top_p, temperature);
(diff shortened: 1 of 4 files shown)


src/python/python.cpp
```diff
@@ -276,10 +276,10 @@ PYBIND11_MODULE(onnxruntime_genai, m) {
       .def(pybind11::init<Model&, PySearchParams&>())
       .def("IsDone", &PyGenerator::IsDone)
       .def("ComputeLogits", &PyGenerator::ComputeLogits)
```
yufenglee, line 278: Add a new function: GetLogits?

RyanUnderhill: Yep, adding that as part of the other API changes, just wanted to submit this separately.

yufenglee (approved):

PR #48, opened 2024-02-01
Using vcpkg

jchen351:


.github/workflows/win-cpu-arm64-build.yml
```diff
@@ -14,12 +14,28 @@ env:
...
 jobs:
   job:
     runs-on: ["self-hosted", "1ES.Pool=onnxruntime-genai-Win11-ARM-CPU"]
     steps:
-      - uses: actions/checkout@v2
+      - name: Checkout OnnxRuntime GenAI repo
+        uses: actions/checkout@v4
+      - name: Download latest Git
+        run: |
+          Invoke-WebRequest https://github.com/dennisameling/git/releases/download/v2.41.0.windows.1/PortableGit-2.41.0-arm64.7z.exe -OutFile ${{ github.workspace }}/git.exe
```
snnn, line 27: Are you saying the default git tool's version is too low?
Could you put the git tools in the VM image instead? And we'd better to get the tool from an official place, like https://git-scm.com/download/win. Otherwise we will get risk of leaking our Github token, which may result a massive data leak. 
Windows ARM64 should be able to run x86 binaries. If it doesn't, try to build git from source.

snnn:
Please create a json file to lockdown all the dependencies' versions. 
Doc: https://learn.microsoft.com/en-us/vcpkg/consume/lock-package-versions?tabs=inspect-powershell#5---force-a-specific-version

.github/workflows/linux-gpu-x64-build.yml
```diff
@@ -9,12 +9,21 @@ env:
...
   job:
     runs-on: [ "self-hosted", "1ES.Pool=onnxruntime-genai-Ubuntu2004-T4" ]
     steps:
       - name: Checkout OnnxRuntime GenAI repo
-        uses: actions/checkout@v2
+        uses: actions/checkout@v4
+        with:
+          submodules: true
+
+      - uses: lukka/get-cmake@latest
```
snnn, line 22: Please void using 3rd party Github actions when possible.

.github/workflows/win-cpu-arm64-build.yml
jchen351: default git isn't installed at all. I need to rebuild the image and see if it works.

snnn:
Before this PR is merged, please write a document to describe how to setup the build environment(install vcpkg and set cmake flags), and add Ryan's to the reviewers, because this change is going to change his daily dev workflow.

vcpkg.json
```diff
@@ -0,0 +1,9 @@
+{
+  "name": "onnxruntime-genai",
+  "version-string": "0.0.1",
+  "license": "MIT",
+  "dependencies": [
```
snnn, line 5: pin the versions.

tools/ci_build/github/linux/docker/inference/aarch64/default/cpu/scripts/install_deps.sh
```diff
@@ -67,3 +67,7 @@ tar --strip 1 -xf /tmp/src/node-v18.17.1-linux-${NODEJS_ARCH}.tar.gz -C /usr
 
 cd /
 rm -rf /tmp/src
+
+git clone https://github.com/Microsoft/vcpkg.git /
```
snnn, line 71: Pin the version. 
You can download the source code via curl and HTTPS as well, to avoid introducing a new dependency, git, in the docker.

tools/ci_build/github/linux/docker/inference/x64/default/gpu/Dockerfile
```diff
@@ -1,17 +1,21 @@
...
 
-ENV PATH="/usr/local/cuda/binet:/opt/rh/gcc-toolset-12/root/usr/bin:${PATH}"
+ENV PATH="/opt/rh/gcc-toolset-12/root/usr/bin:${PATH}"
 ENV LANG=en_US.UTF-8
 ENV LC_ALL=en_US.UTF-8
+
+# Set up CUDA and CUDNN environment variables
 ENV CUDA_HOME=/usr/local/cuda
 ENV CUDNN_HOME=/usr/lib/x86_64-linux-gnu/
+ENV PATH="/usr/local/cuda/binet:${PATH}"
```
snnn, line 14: Misspelling. 
How does it work today???

jchen351:
To use vcpkg with CMake
First clone vckpg as submodule
``` bash
git submodule update --init --recursive
```
Second run bootstrap file 
For Windows
```cmd
.\extern\vcpkg\bootstrap-vcpkg.bat
```
On Unix
```bash
./extern/vcpkg/bootstrap-vcpkg.sh
```

.github/workflows/linux-cpu-arm64-build.yml
```diff
@@ -7,42 +7,48 @@ env:
...
           rm ${{ env.ort_zip }}
-
-      - name: Rename OnnxRuntime to ort
-        run: |
+          echo "Renaming ${{ env.ort_dir }} to ort"
           mv ${{ env.ort_dir }} ort
 
-      - name: Git Submodule Update
+      - name: Replace vcpkg linux.cmake aarch64-linux-gnu-gcc with gcc
         run: |
```
snnn, line 32: Set CC and CXX environment variables instead. 
Like:
export CC=gcc
export CXX=g++

jchen351: Tried, but does not work.

jchen351:
```
elseif(VCPKG_TARGET_ARCHITECTURE STREQUAL "arm64")
    set(CMAKE_SYSTEM_PROCESSOR aarch64 CACHE STRING "")
    if(CMAKE_HOST_SYSTEM_NAME STREQUAL "Linux")
        if(NOT DEFINED CMAKE_CXX_COMPILER)
            set(CMAKE_CXX_COMPILER "aarch64-linux-gnu-g++")
        endif()
        if(NOT DEFINED CMAKE_C_COMPILER)
            set(CMAKE_C_COMPILER "aarch64-linux-gnu-gcc")
        endif()
        if(NOT DEFINED CMAKE_ASM_COMPILER)
            set(CMAKE_ASM_COMPILER "aarch64-linux-gnu-gcc")
        endif()
        if(NOT DEFINED CMAKE_ASM-ATT_COMPILER)
            set(CMAKE_ASM-ATT_COMPILER "aarch64-linux-gnu-as")
        endif()
```

snnn: These settings are for cross-compiling. It means vcpkg probably doesn't support build natively on arm64.

jchen351: So, should we use it?

snnn: I don't know. Let me search around.

CMakeLists.txt
```diff
@@ -3,8 +3,10 @@
 
 cmake_minimum_required(VERSION 3.26)
 include(FetchContent)
-include(cmake/external/onnxruntime_external_deps.cmake)
-
+#include(cmake/external/onnxruntime_external_deps.cmake)
+set(CMAKE_TOOLCHAIN_FILE "${CMAKE_CURRENT_SOURCE_DIR}/extern/vcpkg/scripts/buildsystems/vcpkg.cmake"
```
snnn, line 7: What if a user really needs to supply a toolchain file? For example, if you open the main ORT repo's build.py and search "toolchain", you will see in many cases for cross-compiling we must provide such a file. For example, emsdk has its own toolchain file. Then how to combine them?

snnn: Given there are a lot of questions that need to be figured out, I suggest making vcpkg optional. It means users can use vcpkg to build and install packages, but it is not mandatory, so that users still can manually fetch and build the dependencies in case vcpkg doesn't support the usage case.  I suggest avoiding mentioning vcpkg in our CMakefiles. Instead, we asks users to:
1. Install vcpkg
2. The install the packages by using "vcpkg install ..."
3. Then our CMake file locate the packages by using the "find_package" function.

What do you think?

.github/workflows/linux-cpu-arm64-build.yml
```diff
@@ -7,42 +7,49 @@ env:
...
-
-
+      - name: Docker -- Build with CMake and GCC
+        run: |
+          echo "Running docker image ${{ env.docker_image }}"
+          docker run \
+          --rm \
+          --volume $GITHUB_WORKSPACE:/onnxruntime_src \
+          --volume $RUNNER_TEMP:/build \
+          -w /build ${{ env.docker_image }} bash -c "/onnxruntime_src/extern/vcpkg/bootstrap-vcpkg.sh && /usr/bin/cmake -B build -S /onnxruntime_src -G Ninja -DCMAKE_BUILD_TYPE=Release && cmake --build build --parallel 8 --config Release"
```
snnn, line 53: You need to use a preset, or disable USE_CUDA

natke:
Closing for now. Can re-open later

PR #50, opened 2024-02-02
Enable shared libraries.

jchen351:


RyanUnderhill:
How are the python bindings built with this, they should be statically linked as they don't use the C API.

RyanUnderhill (approved):

PR #51, opened 2024-02-02
Add build.py and add control for building the Python wheel

baijumeswani:
To build:

`python build.py`

To build without the wheel

`python build.py --skip_wheel`


RyanUnderhill:
Do we want this to be built by default? Most people developing it won't be building the wheel every time they build. I thought we'd have it as a parameter to build.sh like in onnxruntime --build_wheel

baijumeswani:
> Do we want this to be built by default? Most people developing it won't be building the wheel every time they build. I thought we'd have it as a parameter to build.sh like in onnxruntime --build_wheel

Makes sense. I think we should add a way to control the build params. This PR only avoids the need for 2 steps:

`make` and `make PyPackageBuild`

We should still add the parameter to control what gets built. i.e. Adding a build param to indicate whether or not we want to build the python wheel. Similarly for `USE_CUDA` and tokenizer and so on.

baijumeswani:
> Do we want this to be built by default? Most people developing it won't be building the wheel every time they build. I thought we'd have it as a parameter to build.sh like in onnxruntime --build_wheel

I think it is fair to assume that most people wanting to build python would want the wheel to be built as well. In fact, I think we could adopt the idea where if we want to build with python enabled, we should probably also install the wheel for the user (similar to what pytorch does).

RyanUnderhill:
> > Do we want this to be built by default? Most people developing it won't be building the wheel every time they build. I thought we'd have it as a parameter to build.sh like in onnxruntime --build_wheel
> 
> I think it is fair to assume that most people wanting to build python would want the wheel to be built as well. In fact, I think we could adopt the idea where if we want to build with python enabled, we should probably also install the wheel for the user (similar to what pytorch does).

Then we'll need a way to not build the wheel so we don't slow down development. --dont_build_wheel?

baijumeswani:
oh...

I was thinking more on the lines of:

```
if (enable_python):
    build_pybind_lib()
    ....
    build_wheel()
```

^ translated to cmake.

where enable_python can be something the user controls through the build arguments. Because for python, it almost always makes sense to use the wheel (even for local testing) as opposed to using the file structure directly.

natke (approved):

build.py
```diff
@@ -0,0 +1,122 @@
...
+        command += ["-G", cmake_generator, "-S", ".", "-B", "build", f"-DBUILD_WHEEL={build_wheel}"]
+        run_subprocess(command).check_returncode()
+        make_command = ["cmake", "--build", ".", "--config", "Release"]
+        run_subprocess(make_command, cwd="build").check_returncode()
+    elif is_linux():
+        cuda_version = "cuda"
+        cuda_arch = 80
+        cuda_compiler = f"/usr/local/{cuda_version}/bin/nvcc"
+
+        env = {"CUDA_HOME": "/usr/local/cuda", "CUDNN_HOME": "/usr/lib/x86_64-linux-gnu/"}
```
yufenglee, line 93: make those part of the parameters

build.py
```diff
@@ -0,0 +1,122 @@
...
+def build(skip_wheel: bool = False):
+    """Generates the CMake build tree and builds the project.
+
+    Args:
+        skip_wheel: Whether to skip building the Python wheel. Defaults to False.
+    """
+    command = [resolve_executable_path("cmake")]
+    build_wheel = "OFF" if skip_wheel else "ON"
+    if is_windows():
+        cmake_generator = "Visual Studio 17 2022"
```
yufenglee, line 83: and for the windows version too

build.py
baijumeswani: Was trying to keep the build script parameters simple to begin with. But have added the parameters now.

PR #54, opened 2024-02-03
Create fast and efficient model builder tool

kunal-vaishnavi:
### Description

This PR creates an initial version of a fast and efficient model builder tool. It allows users to build an optimized and quantized ONNX model in only a few minutes. The tool currently supports model architectures that follow LLaMA-2, Mistral, and Phi-2.

### Motivation and Context

Currently, users have to first export large language models (LLMs) on large machines with massive amounts of memory. Then, users have to run the ORT transformer optimizer script to optimize the exported ONNX model. Finally, users have to choose and run an ORT quantization script to quantize the optimized ONNX model. For the state-of-the-art LLMs, these steps are time-consuming and require machines with large amounts of memory.

This tool combines all of these steps and greatly reduces both the time taken and memory used. For example, generating a LLaMA-2 7B FP16 optimized ONNX model can take more than an hour. With this tool, it takes only 3 minutes to generate the same model.

### DIFF
--- a/models/export.py
+++ b/models/export.py
@@ -1,0 +1,1310 @@
+"""
+Run this script to create the desired ONNX model.
+Example usage:
+python export.py -m model_name -o model.onnx -p precision -e execution_provider
+"""
+
+from onnx import helper, numpy_helper, TensorProto, external_data_helper, load_model, save_model
+from onnxruntime.quantization.matmul_4bits_quantizer import MatMul4BitsQuantizer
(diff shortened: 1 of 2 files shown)


natke (commented):
Can we the script into the python package?

yufenglee:
name it "convert_hf_to_onnx.py"?

natke:
I validated loading a model from a folder on disk

natke (approved):

PR #60, opened 2024-02-06
Add googletest framework

natke:
Add googletest ahead of our testing hackathon tomorrow


CMakeLists.txt
```diff
@@ -208,3 +208,17 @@ source_group("Sources" FILES ${test_srcs})
...
 
 set_property(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR} PROPERTY VS_STARTUP_PROJECT Tests)
+
+enable_testing()
+
+add_executable(
+  tests
+  test/test.cpp
+)
+target_link_libraries(
```
baijumeswani, line 218: Do you also want to link to onnxruntime-genai?

natke: Probably!

natke: Done

snnn (approved):

CMakeLists.txt
snnn: It's not enough. But we can do it later, increamentally.

CMakeLists.txt
```diff
@@ -210,21 +210,22 @@ source_group(TREE ${GENERATORS_ROOT} FILES ${generator_srcs})
...
+  PRIVATE
+  ${CMAKE_SOURCE_DIR}/src
 )
 
+
 target_link_libraries(
-  onnxruntime-genai
   unit_tests
+  PRIVATE
   GTest::gtest_main
```
snnn, line 227: Here we need to link to GenAI static libraries, and all the dependencies that GenAI uses, such as SPM.

natke: Is there a variable for that, or do we need to list them out?

snnn: No such a thing yet. We can do it later.

PR #64, opened 2024-02-07
Add  CMakePresets.json

snnn:
Here is how to use it in Visual Studio 2022: https://learn.microsoft.com/en-us/cpp/build/cmake-presets-vs?view=msvc-170

This feature is totally optional. If you want to use it, you use it. Otherwise, you can stick to your existing dev workflow.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,13 +3,15 @@
 
 cmake_minimum_required(VERSION 3.26)
 include(FetchContent)
-include(cmake/external/onnxruntime_external_deps.cmake)
-
+include(CMakeDependentOption)
 project(Generators LANGUAGES C CXX)
 option(USE_CUDA "Build with CUDA support" ON)
 option(USE_TOKENIZER "Build with Tokenizer support" ON)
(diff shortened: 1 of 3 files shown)


jchen351 (approved):

CMakeLists.txt
```diff
@@ -3,13 +3,15 @@
...
 include(FetchContent)
-include(cmake/external/onnxruntime_external_deps.cmake)
-
+include(CMakeDependentOption)
 project(Generators LANGUAGES C CXX)
 option(USE_CUDA "Build with CUDA support" ON)
 option(USE_TOKENIZER "Build with Tokenizer support" ON)
 option(ENABLE_PYTHON "Enable python buildings" ON)
-cmake_dependent_option(BUILD_WHEEL "Build the python wheel" ON "ENABLE_PYTHON" ON)
+cmake_dependent_option(BUILD_WHEEL "Build the python wheel" ON "ENABLE_PYTHON" OFF)
```
baijumeswani, line 11: Thank you for correcting this bug.

baijumeswani (approved):

PR #71, opened 2024-02-12
fix tests.cpp and linux parallel build

aciddelgado:
Integrate gtest into tests.cpp, and make sure tests run correctly. Linux now builds in parallel. Also, the Phi-2 Tests.cpp now works as example for how to use library.


test/tests.cpp
```diff
@@ -246,38 +242,38 @@ Print all primes between 1 and n
...
+    generator->ComputeLogits();
+    generator->GenerateNextToken_Top();
   }
 
   auto result = generator->GetSequence(0);
 
+  // TODO: Not sure if we want this either
   // High level version
-  auto result = model->Generate(params);
+  // auto result = Generators::Generate(model, params);
```
natke, line 274: We do want this - but it should be in a separate test to show the high level API end to end

test/c_api.cpp
```diff
@@ -35,8 +36,10 @@ void CheckResult(OgaResult* result) {
   throw std::runtime_error(OgaResultGetError(result));
 }
 
-void Test_GreedySearch_Gpt_Fp32_C_API() {
-  std::cout << "Test_GreedySearch_Gpt fp32 C API" << std::flush;
+TEST(CAPITests, GreedySearchGptFp32CAPI) {
+  std::unique_ptr<OrtEnv> g_ort_env;
+  Ort::InitApi();
```
snnn, line 41: This function should not be called multiple times in a single process. Therefore it's better to put it in main.

test/main.cpp
```diff
@@ -1,42 +0,0 @@
...
-
-#if USE_CUDA
-    Test_GreedySearch_Gpt_Cuda();
-    Test_BeamSearch_Gpt_Cuda();
-    Test_Phi2_Cuda();
-#endif
-  } catch (const std::exception& e) {
-    std::cout << "Fatal Exception: " << e.what() << std::endl;
-  }
-  return 0;
```
snnn, line 41: The main function should release the OrtEnv before the function returns.

test/main.cpp
```diff
@@ -1,42 +1,25 @@
...
-#endif
-
-int main() {
+int main(int argc, char **argv) {
   std::cout << "Generators Utility Library" << std::endl;
-
-  std::cout << "Initializing OnnxRuntime...";
+  std::cout << "Initializing OnnxRuntime... ";
   std::cout.flush();
   Ort::InitApi();
```
snnn, line 16: This comment is not related to your change, but we should think about how to handle the failures that may come from this function. A common failure is the test program failed to load onnxruntime.dll(or loaded the wrong one). We see such issues quite often. So it's better to have a good error message here. If we do nothing, I think by default it will just crash.

natke: Suggestion is to move this into the try block

natke (commented):
Can we move the test_models folder into the test folder please?

test/c_api.cpp
```diff
@@ -102,8 +100,7 @@ void Test_GreedySearch_Gpt_Fp32_C_API() {
     std::span<const int32_t> sequence{OgaSequencesGetSequenceData(sequences, i), OgaSequencesGetSequenceCount(sequences, i)};
    
     auto* expected_output_start = &expected_output[i * max_length];
-    if (!std::equal(expected_output_start, expected_output_start + max_length, sequence.begin(), sequence.end()))
-      throw std::runtime_error("Test Results Mismatch");
+    EXPECT_TRUE(0 == std::memcmp(expected_output_start, sequence.data(), max_length * sizeof(int32_t)));
   }
 
   std::cout << " - complete\r\n";
```
RyanUnderhill, line 106: These should go too...

test/main.cpp
RyanUnderhill: Ort::InitApi() will throw an exception with a message, so it should be within a try block.

test/main.cpp
```diff
@@ -1,41 +1,23 @@
...
-}
+  std::cout << "Initializing OnnxRuntime... ";
+  std::cout.flush();
+  Ort::InitApi();
+  g_ort_env = OrtEnv::Create();
+  std::cout << "done" << std::endl;
+  ::testing::InitGoogleTest(&argc, argv);
+  int result = RUN_ALL_TESTS();
+  std::cout << "Shutting down OnnxRuntime... ";
+  g_ort_env.release();
```
RyanUnderhill, line 20: No, we don't want to leak it, we want to destroy it. release should be reset()

baijumeswani (approved):

PR #72, opened 2024-02-12
ONNX Runtime GenAI C# API

baijumeswani:
This pull-request introduces the C# API for `ONNX Runtime GenAI`. The C# API follows the [C API](https://github.com/microsoft/onnxruntime-genai/blob/main/src/ort_genai_c.h) closely and leverages the C API as the interface with the ONNX Runtime GenAI core.

#### The API 📚

The C# API primarily revolves around three classes:

- `Model` - Entry point for running GenAI with the library. Used for loading the model and the config.
- `GeneratorParams` - Used to set the configurable parameters for search.
- `Generator` - Main computation engine. Used for computing the logits and retrieving the generated tokens.

#### Sample usage 👨‍💻

```csharp
using Model model = new Model(modelPath, DeviceType.CPU)
using GeneratorParams generatorParams = new GeneratorParams(model)

generatorParams.SetMaxLength(maxLength);
generatorParams.SetInputIDs(inputIDs, sequenceLength, batchSize);

using Generator generator = new Generator(model, generatorParams)

while (!generator.IsDone())
{
    generator.ComputeLogits();
    generator.GenerateNextTokenTop();
}

for (ulong i = 0; i < batchSize; i++)
{
    var sequence = generator.GetSequence((int)i);
    ... // use the output tokens
}
```

#### What does this Pull-Request Include? 🤔

- The C# API for Windows. The build generates a managed dll called `Microsoft.ML.OnnxRuntimeGenAI.dll` for `x64` using `net8.0`.
- Test that leverages the managed dll built.
- This pull-request does not include packaging.
- Module definition for exporting the native functions in the C library.

#### What's next? 🔄

- Run build and tests in CI pipeline.
- C# packaging.
- Linux C# support
- Build the C# API as part of the cmake build.
- Tokenizer functions to be added to C# API to complete the API


src/onnxruntime-genai.def
```diff
@@ -0,0 +1,16 @@
+LIBRARY
```
snnn, line 1: On Linux you will need a similar thing. That's why the main ORT repo has a https://github.com/microsoft/onnxruntime/blob/main/tools/ci_build/gen_def.py file that reads a config file, symbols.txt, then generates different link config files on different platforms.

baijumeswani: Yes, I will add support for linux in a subsequent PR. This PR is mostly for the API and has been validated on windows so far.
Thanks for the information. I had to search quite a bit in ort to understand how it exported the functions. :)

yuslepukhin:
The summary example features nested using blocks. It would be better to use `using` statements that are scope based, so blocks are not required, although they may be useful when you want to dispose of something before the current scope ends.

src/csharp/Microsoft.ML.OnnxRuntimeGenAI.csproj
```diff
@@ -0,0 +1,55 @@
+<Project Sdk="MSBuild.Sdk.Extras/3.0.22">
+  <PropertyGroup>
+    <TargetFrameworks>net8.0</TargetFrameworks>
+    <Platforms>x64</Platforms>
+    <LangVersion>latest</LangVersion>
```
yuslepukhin, line 5: LangVersion must be default so it is correctly set by the IDE according to the TargetFramework. This way it will not allow you to use language features that are not supported.

src/csharp/NativeMethods.cs
```diff
@@ -0,0 +1,90 @@
...
+        }
+
+        [DllImport(NativeLib.DllName, CallingConvention = CallingConvention.Winapi)]
+        public static extern IntPtr /* const char* */ OgaResultGetError(IntPtr /* OgaResult* */ result);
+
+        [DllImport(NativeLib.DllName, CallingConvention = CallingConvention.Winapi)]
+        public static extern void OgaDestroyResult(IntPtr /* OgaResult* */ result);
+
+        [DllImport(NativeLib.DllName, CallingConvention = CallingConvention.Winapi)]
+        public static extern IntPtr /* OgaResult* */ OgaCreateModel(string /* const char* */ configPath,
```
yuslepukhin, line 23: You can not use string, because it will not be converted to ANSI by default, not UTF-8. In this case, the client code must covert the string to UTF-8 `byte[]`. Then you can pass it as such.

src/csharp/NativeMethods.cs
```diff
@@ -0,0 +1,90 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
yuslepukhin, line 1: General comment for the file. Please, add detailed docs for each entry, explaining what parameters are really const, in or out.
Mention C types where appropriate.

It is especially important to understand who owns what and the lifespan of the entities.
For example, it is not clear who owns and destroys OgaGenerate() generated object.

For that we need a clear C API documentation with in/out params and whether the objects and buffers must be destroyed by C# API.

src/csharp/NativeMethods.cs
```diff
@@ -0,0 +1,90 @@
...
+        [DllImport(NativeLib.DllName, CallingConvention = CallingConvention.Winapi)]
+        public static extern IntPtr /* OgaResult* */ OgaCreateGeneratorParams(IntPtr /* OgaModel* */ model,
+                                                                              out IntPtr /* OgaGeneratorParams** */ generatorParams);
+
+        [DllImport(NativeLib.DllName, CallingConvention = CallingConvention.Winapi)]
+        public static extern void OgaDestroyGeneratorParams(IntPtr /* OgaGeneratorParams* */ generatorParams);
+
+        [DllImport(NativeLib.DllName, CallingConvention = CallingConvention.Winapi)]
+        public static extern IntPtr /* OgaResult* */ OgaGeneratorParamsSetMaxLength(IntPtr /* OgaGeneratorParams* */ generatorParams,
+                                                                                    int maxLength);
```
yuslepukhin, line 39: Because int size can vary on different platforms, you should use `IntPtr`

src/csharp/NativeMethods.cs
```diff
@@ -0,0 +1,90 @@
...
+
+        [DllImport(NativeLib.DllName, CallingConvention = CallingConvention.Winapi)]
+        public static extern IntPtr /* OgaResult* */ OgaGeneratorParamsSetInputIDs(IntPtr /* OgaGeneratorParams* */ generatorParams,
+                                                                                   int[] /* int32_t* */ inputIDs,
+                                                                                   UIntPtr inputIDsCount,
```
yuslepukhin, line 44: Mention what C type is expected on the native end.

src/csharp/NativeMethods.cs
```diff
@@ -0,0 +1,90 @@
...
+        public static extern bool OgaGenerator_IsDone(IntPtr /* OgaGenerator* */ generator);
+
+        [DllImport(NativeLib.DllName, CallingConvention = CallingConvention.Winapi)]
+        public static extern IntPtr /* OgaResult* */ OgaGenerator_ComputeLogits(IntPtr /* OgaGenerator* */ generator);
+
+        [DllImport(NativeLib.DllName, CallingConvention = CallingConvention.Winapi)]
+        public static extern IntPtr /* OgaResult* */ OgaGenerator_GenerateNextToken_Top(IntPtr /* OgaGenerator* */ generator);
+
+        [DllImport(NativeLib.DllName, CallingConvention = CallingConvention.Winapi)]
+        public static extern IntPtr /* OgaResult* */ OgaGenerator_GetSequence(IntPtr /* OgaGenerator* */ generator,
```
yuslepukhin, line 66: It would be better to have a separate call to get the size of the pre-allocated buffer. Then we could have int[] in the signature. It would take care of the pinning, now the pinning must be done by C# API.

src/csharp/NativeMethods.cs
```diff
@@ -0,0 +1,90 @@
...
+                                                                              int index,
+                                                                              IntPtr /* int32_t* */ tokens,
+                                                                              out UIntPtr /* size_t* */ tokensCount);
+
+        [DllImport(NativeLib.DllName, CallingConvention = CallingConvention.Winapi)]
+        public static extern void OgaDestroySequences(IntPtr /* OgaSequences* */ sequences);
+
+        [DllImport(NativeLib.DllName, CallingConvention = CallingConvention.Winapi)]
```
yuslepukhin, line 74: Does this return the number of tokens in the Sequence? Can it be used for pre-allocating buffer for GetSequence?

src/csharp/NativeMethods.cs
```diff
@@ -0,0 +1,90 @@
...
+        public static extern ulong OgaSequencesCount(IntPtr /* OgaSequences* */ sequences);
```
yuslepukhin, line 75: C API returns size_t and size_t size changes depending on 32/64 bit. For that reason, it must return UIntPtr.

src/csharp/NativeMethods.cs
```diff
@@ -0,0 +1,90 @@
...
+
+        [DllImport(NativeLib.DllName, CallingConvention = CallingConvention.Winapi)]
+        public static extern ulong OgaSequencesGetSequenceCount(IntPtr /* OgaSequences* */ sequences,
```
yuslepukhin, line 78: This API must return UIntPtr and accept UIntPtr.

baijumeswani:
> The summary example features nested using blocks. It would be better to use `using` statements that are scope based, so blocks are not required, although they may be useful when you want to dispose of something before the current scope ends.

Updated the sample example to avoid using the nested using statements.

src/csharp/NativeMethods.cs
```diff
@@ -0,0 +1,90 @@
...
+                                                                ulong sequenceIndex);
+
+        [DllImport(NativeLib.DllName, CallingConvention = CallingConvention.Winapi)]
+        public static extern IntPtr /* const int32_t* */ OgaSequencesGetSequenceData(IntPtr /* OgaSequences* */ sequences,
```
yuslepukhin, line 82: This must take UIntPtr. How do we know the length of the buffer to which the ptr is returned? The original C API does not have good documentation.
The starting point should be to get a clear documentation for the C API and build on top of that.

yuslepukhin (changes requested):

src/onnxruntime-genai.def
```diff
@@ -0,0 +1,27 @@
...
+EXPORTS
```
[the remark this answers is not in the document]
snnn: onnxruntime-genai doesn't need to maintain backwards compatibility, therefore either way should work. Otherwise the order of the symbols matters.

src/csharp/NativeMethods.cs
baijumeswani: I added documentation to both the C API as well as what is in this file.

src/csharp/NativeMethods.cs
baijumeswani: This returns the number of sequences in the object `OgaSequences`.
OgaSequences can be thought of as a 2D array where this function returns the size in the first dimension.

src/csharp/NativeMethods.cs
baijumeswani: Addressed.

src/onnxruntime-genai.def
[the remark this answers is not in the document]
baijumeswani: I removed the onnxruntime-genai.def file and am now using `__declspec(__dllexport)` at build time.

src/csharp/Microsoft.ML.OnnxRuntimeGenAI.csproj
```diff
@@ -0,0 +1,55 @@
+<Project Sdk="MSBuild.Sdk.Extras/3.0.22">
+  <PropertyGroup>
+    <TargetFrameworks>net8.0</TargetFrameworks>
```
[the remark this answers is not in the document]
baijumeswani: Made this project netstandard2.1 and the test is now using net6.0.

src/csharp/Microsoft.ML.OnnxRuntimeGenAI.csproj
baijumeswani: Addressed this.

src/csharp/NativeMethods.cs
baijumeswani: Addressed this.

src/csharp/NativeMethods.cs
baijumeswani: Addressed this.

src/csharp/NativeMethods.cs
baijumeswani: I added the expected C type for all the native methods.

src/csharp/NativeMethods.cs
baijumeswani: Changed the API to now have two functions. One for getting the count of tokens in the sequence which can be used for allocating the sequence on the c# side and the other for getting the data that can be copied into this buffer.

src/csharp/NativeMethods.cs
baijumeswani: Addressed this.

src/csharp/NativeMethods.cs
baijumeswani: Addressed this.

src/ort_genai_c.h
```diff
@@ -47,34 +60,154 @@ OGA_EXPORT size_t OGA_API_CALL OgaBufferGetDimCount(const OgaBuffer*);
...
+ * \return The number of tokens in the sequence at the given index
+ */
+OGA_EXPORT size_t OGA_API_CALL OgaSequencesGetSequenceCount(const OgaSequences* sequences, size_t sequence_index);
+
+/*
+ * \brief Returns a pointer to the sequence data at the given index. The number of tokens in the sequence
+ *        is given by OgaSequencesGetSequenceCount
+ * \param[in] sequences
+ * \return The pointer to the sequence data at the given index. The sequence data is owned by the OgaSequences
+ *         and will be freed when the OgaSequences is destroyed. The caller must copy the data if it needs to
```
RyanUnderhill, line 87: Would it be better to explain based on lifetime, like:
"The pointer to the sequence data at the given index. The pointer is valid until the OgaSequences is destroyed. ..."

src/ort_genai_c.h
```diff
@@ -47,34 +60,154 @@ OGA_EXPORT size_t OGA_API_CALL OgaBufferGetDimCount(const OgaBuffer*);
...
+ *         be used after the OgaSequences is destroyed.
+ */
+OGA_EXPORT const int32_t* OGA_API_CALL OgaSequencesGetSequenceData(const OgaSequences* sequences, size_t sequence_index);
+
+/*
+ * \brief Creates a model from the given configuration directory and device type.
+ * \param[in] config_path The path to the model configuration directory.
```
RyanUnderhill, line 94: Should we say the config_path is utf8?

src/ort_genai_c.h
```diff
@@ -47,34 +60,154 @@ OGA_EXPORT size_t OGA_API_CALL OgaBufferGetDimCount(const OgaBuffer*);
...
-OGA_EXPORT OgaResult* OGA_API_CALL OgaGenerate(const OgaModel*, OgaGeneratorParams* generator_params, OgaSequences** out);
+/*
+ * \brief Destroys the given model.
+ * \param[in] model The model to be destroyed.
+ */
+OGA_EXPORT void OGA_API_CALL OgaDestroyModel(OgaModel* model);
 
+/*
+ * \brief Generates sequences of tokens from the model execution based on the given generator params.
+ *        The sequences can be thought of as a 2D array of tokens where the first dimension is the batch size
```
RyanUnderhill, line 109: We shouldn't think of it as a 2D array. It's really an 'array of token arrays', as we should keep the ability for each one to have a different length (vs padding them all in every case). Also not sure if that's best put in the explanation for OgaSequences.

RyanUnderhill (approved):

src/csharp/GeneratorParams.cs
```diff
@@ -0,0 +1,53 @@
...
+// Licensed under the MIT License.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Runtime.InteropServices;
+
+namespace Microsoft.ML.OnnxRuntimeGenAI
+{
+    public class GeneratorParams : IDisposable
```
yuslepukhin, line 11: We need to follow IDisposable pattern here according to the MS Docs.
Introduce bool _disposed = false; member.
Check in the beggining of the Dispose() and return.
Do not check for the handle to be Zero.

This is a simple class, but generally, Dispose() takes care of 2 classes of resources.
1) disposable members, those are taken care of when disposing argument is true. In this case, we do not have such yet.
2) Native resources. They are disposed unconditionally, but once because _diposed guards the entrance.

Set _disposed to true.

The reason for the above is a complex dynamic and interactions between GC, finalizer thread and the runtime.
See here https://learn.microsoft.com/en-us/dotnet/api/system.idisposable?view=net-8.0
and the link at the bottom how to implement IDisposable.

src/csharp/Model.cs
```diff
@@ -0,0 +1,72 @@
...
+namespace Microsoft.ML.OnnxRuntimeGenAI
+{
+    public enum DeviceType : long
+    {
+        Auto = 0,
+        CPU = 1,
+        CUDA = 2
+    }
+
+    public class Model : IDisposable
```
yuslepukhin, line 16: Same comments for IDisposable.

src/csharp/Model.cs
```diff
@@ -0,0 +1,72 @@
...
+        private IntPtr _modelHandle;
+
+        public Model(string modelPath, DeviceType deviceType)
+        {
+            Result.VerifySuccess(NativeMethods.OgaCreateModel(Utils.ToUtf8(modelPath), deviceType, out _modelHandle));
+        }
+
+        internal IntPtr Handle { get { return _modelHandle; } }
+
+        public int[][] Generate(GeneratorParams generatorParams)
```
yuslepukhin, line 27: This creates a number of garbage objects. I think it would be much better to return ReadOnlySpan to the individual native buffers and let the user to decide what to do with them. 
To achive that introduce a Sequence class (and other classes as appriate) with the appropriate methods that return spans directly to native data.
This reduces the amount of garbage and eliminates copying.

Dealing with the multi-dim arrays is very inconvinient. Let the user build them if they want to.

src/csharp/Generator.cs
```diff
@@ -0,0 +1,64 @@
...
+        {
+            Result.VerifySuccess(NativeMethods.OgaGenerator_ComputeLogits(_generatorHandle));
+        }
+
+        public void GenerateNextTokenTop()
+        {
+            Result.VerifySuccess(NativeMethods.OgaGenerator_GenerateNextToken_Top(_generatorHandle));
+        }
+
+        public int[] GetSequence(int index)
```
yuslepukhin, line 33: Same comment, return span directly to native data, this reduces the amount of garbage and eliminates data copying.
Also implement this[index] that returns a span instead of the method. C# developers expect appropriate patterns.
Add an out of bounds check for the index argument.

yuslepukhin (changes requested):

src/csharp/NativeMethods.cs
```diff
@@ -0,0 +1,112 @@
...
+        // This function returns the number of tokens in the sequence at the given index of the OgaSequences object.
+        [DllImport(NativeLib.DllName, CallingConvention = CallingConvention.Winapi)]
+        public static extern UIntPtr OgaSequencesGetSequenceCount(IntPtr /* const OgaSequences* */ sequences,
+                                                                  UIntPtr /* size_t */ sequenceIndex);
+
+        // This function returns the sequence data at the given index of the OgaSequences object. The returned
+        // pointer is owned by the OgaSequences object and will be freed when the OgaSequences object is destroyed.
+        // It is expected that the caller copies the data returned by this function after calling this function.
+        [DllImport(NativeLib.DllName, CallingConvention = CallingConvention.Winapi)]
+        public static extern IntPtr /* const int32_t* */ OgaSequencesGetSequenceData(IntPtr /* const OgaSequences* */ sequences,
```
yuslepukhin, line 102: >OgaSequencesGetSequenceData [](http://example.com/codeflow?start=57&length=27)

Please, mention here which API returns the length of the returned data.

src/csharp/Sequences.cs
```diff
@@ -0,0 +1,70 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
```
yuslepukhin, line 4: All classes and methods should be documented

src/csharp/Sequences.cs
```diff
@@ -0,0 +1,70 @@
...
+                ulong sequenceLength = NativeMethods.OgaSequencesGetSequenceCount(_sequencesHandle, (UIntPtr)sequenceIndex).ToUInt64();
+                IntPtr sequencePtr = NativeMethods.OgaSequencesGetSequenceData(_sequencesHandle, (UIntPtr)sequenceIndex);
+                unsafe
+                {
+                    return new ReadOnlySpan<int>(sequencePtr.ToPointer(), (int)sequenceLength);
+                }
+            }
+        }
+
+        public int[][] ToArray()
```
yuslepukhin, line 38: I would remove this method. Clients can call them if they wanted, but that would be their code generating garbage

src/csharp/GeneratorParams.cs
```diff
@@ -0,0 +1,56 @@
...
+        internal IntPtr Handle { get { return _generatorParamsHandle; } }
+
+        public void SetMaxLength(ulong maxLength)
+        {
+            Result.VerifySuccess(NativeMethods.OgaGeneratorParamsSetMaxLength(_generatorParamsHandle, (UIntPtr)maxLength));
+        }
+
+        public void SetInputIDs(IReadOnlyCollection<int> inputIDs, ulong sequenceLength, ulong batchSize)
+        {
+            int[] inputIDsArray = inputIDs.ToArray();
```
yuslepukhin, line 30: This produces a copy + garbage collected array on every call.

Here is my suggestion, the inputIDs must be in a contagious memory for this API call. Therefore, let's change the input parameter to ReadOnlySpan<int>.

Then, you can obtain a pointer to its buffer, pin it and feed directly to the native call.

here is an example:
[https://github.com/microsoft/onnxruntime/blob/main/csharp/src/Microsoft.ML.OnnxRuntime/OrtValue.shared.cs#L783](https://github.com/microsoft/onnxruntime/blob/main/csharp/src/Microsoft.ML.OnnxRuntime/OrtValue.shared.cs#L783)

src/csharp/Sequences.cs
baijumeswani: Would it be ok to get to the API documentation in another PR to avoid delaying other tasks dependent on this PR?

yuslepukhin (approved):
:shipit:

baijumeswani:
Thank you for the review @yuslepukhin @RyanUnderhill :)

PR #83, opened 2024-02-14
Disable USE_CUDA if CMAKE_CUDA_COMPILER is not found instead of ERROR

jchen351:
This pull request primarily focuses on changes in the `CMakeLists.txt` file to improve the handling of CUDA-related files during the build process. The changes ensure that CUDA files are only included when CUDA is available and properly configured. Additionally, the error message when CUDA is not found has been modified to a warning, and CUDA support is disabled instead of terminating the build process.

Changes to CUDA file handling:

* [`CMakeLists.txt`](diffhunk://#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR47-R49): Added a new glob pattern to exclude CUDA files from `generator_srcs` unless CUDA is being used. This is done by creating a new list `generator_cuda_exclude_srcs` that matches all CUDA files, and then removing these files from `generator_srcs`.

Changes to CUDA availability checks:

* [`CMakeLists.txt`](diffhunk://#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aL61-R70): Modified the `if(USE_CUDA AND CMAKE_CUDA_COMPILER)` condition to append the excluded CUDA files to `generator_srcs` only when CUDA is available and properly configured. If CUDA is requested but the compiler is not found, a warning message is now displayed and CUDA support is disabled, instead of terminating the build process with a fatal error.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -44,6 +44,9 @@ file(GLOB generator_srcs CONFIGURE_DEPENDS
   "${MODELS_ROOT}/*.h"
   "${MODELS_ROOT}/*.cpp"
 )
+file(GLOB generator_cuda_exclude_srcs "${GENERATORS_ROOT}/*_cuda*.*")
+# Remove the cuda files from the generator_srcs unless we are using cuda
+list(REMOVE_ITEM generator_srcs ${generator_cuda_exclude_srcs})
 
 if(USE_CUDA AND CMAKE_CUDA_COMPILER)
   # Don't let cmake set a default value for CMAKE_CUDA_ARCHITECTURES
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -58,15 +61,13 @@ if(USE_CUDA AND CMAKE_CUDA_COMPILER)
...
+  list(APPEND generator_srcs ${generator_cuda_srcs} ${generator_cuda_exclude_srcs})
   add_compile_definitions(USE_CUDA=1)
   include_directories("${CMAKE_CUDA_TOOLKIT_INCLUDE_DIRECTORIES}")
 elseif(USE_CUDA)
   # USE_CUDA is true but cmake could not find the cuda compiler
-  message(FATAL_ERROR "USE_CUDA is ON but no cuda compiler was found.")
-else()
-  file(GLOB generator_cuda_srcs "${GENERATORS_ROOT}/*_cuda*.*")
-  list(REMOVE_ITEM generator_srcs ${generator_cuda_srcs})
+  message(WARNING "USE_CUDA is ON but no cuda compiler was found. Disabling CUDA support.")
```
baijumeswani, line 69: What is the context behind giving a warning instead of terminating the build?

CMakeLists.txt
```diff
@@ -44,6 +44,9 @@ file(GLOB generator_srcs CONFIGURE_DEPENDS
   "${MODELS_ROOT}/*.h"
   "${MODELS_ROOT}/*.cpp"
 )
+file(GLOB generator_cuda_exclude_srcs "${GENERATORS_ROOT}/*_cuda*.*")
```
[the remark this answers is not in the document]
jchen351: `generator_cuda_srcs` is used by later line 58

CMakeLists.txt
jchen351: Because USE_CUDA is ON by default. So if someone work on a machine doesn't have CUDA, they have to change USE_CUDA to OFF every time sync with main.

baijumeswani: They can turn it off using -DUSE_CUDA=OFF, right?
Silently falling back to not using cuda when the user requests it (for example when user says -DUSE_CUDA=ON) might give users the wrong impression of what is getting built.

jchen351: But I think by default, most of user doesn't have CUDA. And we don't tell them how to build without CUDA from source (-DUSE_CUDA=OFF).

PR #84, opened 2024-02-14
Switch USE_TOKENIZER to opposite NO_TOKENIZER

RyanUnderhill:
Since USE_TOKENIZER is the default, it makes sense to opt out vs opt in when including headers.

This also removes most of the #if statements and instead puts them in one single place and makes it a runtime error that reports "Tokenizer not enabled" if anyone tries to use it.

If we remove this option later, this change makes it even simpler to remove this small piece.


src/models/model.cpp
```diff
@@ -31,7 +31,18 @@ void State::ClearIO() {
   outputs_.clear();
 }
 
-#if USE_TOKENIZER
+#ifdef NO_TOKENIZER
```
yufenglee, line 34: #ifndef

test/tests.cpp
```diff
@@ -226,62 +226,4 @@ void Test_BeamSearch_Gpt_Cuda() {
   for (auto model_path : c_tiny_gpt2_model_paths)
     Test_BeamSearch_Gpt_Cuda(model_path.first, model_path.second);
 }
-
-void Test_Phi2_Cuda() {
```
yufenglee, line 230: why do you remove this?

yufenglee (approved):
:shipit:

test/tests.cpp
RyanUnderhill: It was disabled code that was out of date and never used. I was just cleaning it up when I saw it.

PR #86, opened 2024-02-14
Update span.h to fix breaks

RyanUnderhill:
Hasn't been built for a while, so some errors crept in. This should allow compiling on certain older compilers/platforms again.


jchen351:
Still fail on MacOS with  
 add_compile_definitions(USE_CXX17=1)
  message("USE C++17")
  set(CMAKE_CXX_STANDARD 17)

snnn (approved):

snnn:
Would you mind using absl::span? Then we will not need to spend time on maintaining this file.

snnn:
Anyway gtest, protobuf and a lot of others already depends on absl.

PR #90, opened 2024-02-15
Add tokenizer C APIs

RyanUnderhill:
Note that the encode is 'batch encode' as it returns an OgaSequences to be symmetrical with the OgaGenerate function inputs/outputs.
The decode is a single token array decode, since that avoids needing to create a separate "array of strings" type that can be created by the C API and managed by the users.

The streaming API is a guess, as that one is still hypothetical based on the streaming tokenizer API.


src/ort_genai_c.h
```diff
@@ -76,6 +80,22 @@ OGA_EXPORT OgaResult* OGA_API_CALL OgaGenerator_GenerateNextToken_TopP(OgaGenera
...
+ */
+OGA_EXPORT OgaResult* OGA_API_CALL OgaTokenizerDecode(const OgaTokenizer*, const int32_t* tokens, size_t token_count, const char** out_string);
+
+OGA_EXPORT OgaResult* OGA_API_CALL OgaCreateTokenizerStream(const OgaTokenizer*, OgaTokenizerStream** out);
+OGA_EXPORT void OGA_API_CALL OgaDestroyTokenizerStream(OgaTokenizerStream*);
+
+/* Streaming decoder for tokens. Allows displaying a decoded token stream as the tokens are generated.
+ * As not every token will result in bytes written to the string, bytes_written may sometimes be zero.
+ */
+OGA_EXPORT OgaResult* OGA_API_CALL OgaTokenizerDecodeStream(const OgaTokenizer*, OgaTokenizerStream*, int32_t token, char* string, size_t string_size, size_t* bytes_written);
```
yufenglee, line 97: >OgaTokenizerDecodeStream [](http://example.com/codeflow?start=35&length=24)

Where is this function implemented?

yufenglee, line 97: >int32_t token [](http://example.com/codeflow?start=102&length=13)

we need to support multiple batch

RyanUnderhill: If the user needs to decode multiple streams, the user should iterate over multiple OgaTokenizerStream objects.

Otherwise we'd need to create more C API types, OgaTokenizerStreams, and have it return an array of strings each iteration. Is that worth it? There's no perf gain.

RyanUnderhill: It wasn't :) The new update has them all implemented and tests that demonstrate them. The streaming versions match the output for all of the test cases.

natke (changes requested):
Can we remove the streaming API interfaces from this PR. This opens the possibility of releasing without them. They can remain in a separate PR until we have the implementation

src/ort_genai_c.h
natke: @RyanUnderhill please add sample usage to the PR

RyanUnderhill: Yep, there's a sample of all the APIs added to c_api.cpp: TokenizerCAPI

baijumeswani (approved):

PR #91, opened 2024-02-15
add extra options for export.py

yufenglee:
1. Add an extra_options option for export. It allows user to easily specify some parameters, like for quantization.
2. Re-order the ReadMe a littlt bit.


src/python/models/export.py
```diff
@@ -102,8 +103,8 @@ def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir):
         # Quantization-specific variables (INT4, INT8, etc.)
         self.quant_attrs = {
             "int4": {
-                "block_size": 32,
-                "accuracy_level": None,
+                "block_size": int(extra_options["int4_block_size"]) if "int4_block_size" in extra_options else 32,
+                "accuracy_level": int(extra_options["int4_accuracy_level"]) if "int_accuracy_level" in extra_options else None,
```
kunal-vaishnavi, line 107:
```suggestion
                "accuracy_level": int(extra_options["int4_accuracy_level"]) if "int4_accuracy_level" in extra_options else None,
```

PR #96, opened 2024-02-20
C# Nuget Package Spec

baijumeswani:
This pull request introduces the nuspec for the native and managed package so the nuget package can be locally built.

To create a package, we need to run this command:

```bash
nuget.exe pack Microsoft.ML.OnnxRuntimeGenAI.nuspec -Prop version=0.1.0 -Prop id="Microsoft.ML.OnnxRuntimeGenAI" # for the native package
nuget.exe pack Microsoft.ML.OnnxRuntimeGenAI.nuspec -Prop version=0.1.0 -Prop id="Microsoft.ML.OnnxRuntimeGenAI.Gpu" # for the native gpu package
nuget.exe pack Microsoft.ML.OnnxRuntimeGenAI.Managed.nuspec -Prop version=0.1.0 # for the managed package
```

Note that the build paths inside the nuspec will need to be updated to adapt to ones build director structure.


yuslepukhin (commented):
I do not have much comments. Using NuGet one can test it locally. However, I think the pipelines do not actually run NuGet directly, but through a project target. So the PR comment is not entirely correct.

natke (approved):

PR #98, opened 2024-02-20
Update phi2 example

natke:


examples/phi2/README.md
```diff
@@ -2,8 +2,19 @@
 
 ## Get the model
 
+Install the model builder script dependencies
```
edgchen1, line 5: maybe put these into a requirements.txt file and refer to that instead so they don't need to be updated in multiple places.
e.g., src/python/models/export.requirements.txt

examples/phi2/README.md
```diff
@@ -2,8 +2,19 @@
...
+pip install numpy
+pip install transformers
+pip install torch
+pip install onnx
+pip install onnxruntime
+```
+
 ```bash
-python -m onnxruntime_genai.scripts.export.py -m microsoft/phi-2 -e cpu -p int4 -o model/model.onnx
+wget https://github.com/microsoft/onnxruntime-genai/blob/main/src/python/models/export.py
```
edgchen1, line 16: this download step seems unnecessary if we already assume the user has a copy of the repo (the next step directly runs `phi2.py`). perhaps just refer to the script path (e.g., `<repo root>/src/python/models/export.py`)

examples/phi2/README.md
```diff
@@ -2,8 +2,19 @@
...
+python export.py -m microsoft/phi-2 -e cpu -p int4 -o model/model.onnx
```
edgchen1, line 17: consider being explicit about which directory the user should be in. this directory is assumed with the relative path `model/model.onnx`. similar comment for running phi2.py. maybe just have initial instructions for the user to change to this directory.

examples/phi2/README.md
```diff
@@ -2,8 +2,19 @@
...
 ```
 
 ## Run the phi-2 model
```
edgchen1, line 20: consider adding instructions for installing `onnxruntime_genai`

natke:
Superseded by #111

PR #106, opened 2024-02-22
Fix Batching, add C API, and C example

RyanUnderhill:
position_ids.cpp/h almost supported batching but now it does it properly. In the future we can allow the user to specify an attention_mask as an external input.

Added C API `OGA_EXPORT OgaResult* OGA_API_CALL OgaGeneratorParamsSetInputSequences(OgaGeneratorParams* generator_params, const OgaSequences* sequences);`

Added C API example of batching using above API and existing tokenizer code.

Cleaned up some of the cuda code affected by the changes while I was at it.

Note: phi-2 config.json must have the proper token_id settings:
```
  "pad_token_id": 50256,
  "bos_token_id": 50256,
  "eos_token_id": 50256,
```


test/python/test_onnxruntime_genai_api.py
```diff
@@ -49,3 +49,28 @@ def test_greedy_search(device, test_data_path, relative_model_path):
...
+@pytest.mark.parametrize("relative_model_path", [Path("phi-2")])
+def test_batching(device, test_data_path, relative_model_path):
+    model_path = os.fspath(
+        Path(test_data_path) / relative_model_path
+    )
+
+    model = og.Model(model_path, device)
+    tokenizer = model.create_tokenizer()
+
+    '''
```
baijumeswani, line 63: I think you forgot to uncomment your test. :)

test/python/readme.txt
baijumeswani, line 1: Change to README.md so it shows up on github?

baijumeswani (approved):

baijumeswani (approved):

PR #108, opened 2024-02-22
efficient kernels for update position ids

aciddelgado:
parallelize cuda kernels that were previously inefficient


src/span.h
```diff
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 #ifndef USE_CXX17
+#pragma once
```
RyanUnderhill, line 4: The Pragma once should be at the top

RyanUnderhill (approved):

PR #112, opened 2024-02-23
Introduce the C# Tokenizer API

baijumeswani:
This pull-request introduces the C# `Tokenizer` related classes.

Sample usage:

```csharp
using Model model = new Model(modelPath, DeviceType.CPU);
using Tokenizer tokenizer = new Tokenizer(model);

var strings = new string[] { 
    "This is a test.",
    "Rats are awesome pets!",
    "The quick brown fox jumps over the lazy dog."
};

var sequences = tokenizer.EncodeBatch(strings);

using GeneratorParams generatorParams = new GeneratorParams(model);
generatorParams.SetMaxLength(20);
generatorParams.SetInputSequences(sequences);

var outputSequences = model.Generate(generatorParams);

var outputStrings = tokenizer.DecodeBatch(outputSequences);

```


src/ort_genai_c.h
```diff
@@ -218,9 +225,8 @@ OGA_EXPORT const int32_t* OGA_API_CALL OgaGenerator_GetSequence(const OgaGenerat
 
 OGA_EXPORT OgaResult* OGA_API_CALL OgaCreateTokenizer(const OgaModel* model, OgaTokenizer** out);
 OGA_EXPORT void OGA_API_CALL OgaDestroyTokenizer(OgaTokenizer*);
-OGA_EXPORT OgaResult* OGA_API_CALL OgaTokenizerEncodeBatch(const OgaTokenizer*, const char** strings, size_t count, OgaSequences** out);
```
RyanUnderhill, line 221: We should have a simple/faster common case function vs requiring the complex batching functions. Since this change requires copying every string before tokenizing, and using another object, we should also have a low level function like:

`OgaTokenizerEncode(const OgaTokenizer*, const char* string, OgaSequences** out);`

It's a common case to just pass in one string, so we should make this simple and efficient. It also would pair with OgaTokenizerDecode.

Or perhaps use the OgaBuffer. The batching APIs shouldn't be less efficient than the non batching ones.

baijumeswani: I added a new function for single encode.

baijumeswani: I can also go back to using the previous signature if you feel that would be better. It will just make the c# code quite unpleasant with all the Marshalling to make it work with const char** and const char*** types.

src/csharp/NativeMethods.cs
```diff
@@ -51,6 +51,11 @@ internal class NativeLib
                                                                                           UIntPtr /* size_t */ sequenceLength,
                                                                                           UIntPtr /* size_t */ batchSize);
 
+
+        [DllImport(NativeLib.DllName, CallingConvention = CallingConvention.Winapi)]
+        public static extern unsafe IntPtr /* OgaResult* */ OgaGeneratorParamsSetInputSequences(IntPtr /* OgaGeneratorParams* */ generatorParams,
```
yuslepukhin, line 56: Why in this case OgaResult is `unsafe` and in other cases it is not?

src/csharp/Tokenizer.cs
```diff
@@ -0,0 +1,119 @@
...
+        public Sequences EncodeBatch(string[] strings)
+        {
+            IntPtr nativeSequences = IntPtr.Zero;
+            IntPtr stringArray = IntPtr.Zero;
+            Result.VerifySuccess(NativeMethods.OgaCreateStringArray(out stringArray));
+            try
+            {
+                foreach (string s in strings)
+                {
+                    Result.VerifySuccess(NativeMethods.OgaStringArrayAddString(stringArray, Utils.ToUtf8(s)));
```
yuslepukhin, line 28: This produces a temporary `byte[]` which is garbage. 
Here is a suggestion that works best:
1) Calculate the length of the buffer required for the UTF-8 string, see example in the OrtValue API
2) Make an API that allocates a buffer for that string (May be the same API, but change the signature): essentially calls std::string::resize() and return a ptr to that buffer
3) Copy convert the string straight into the native buffer.

src/csharp/Tokenizer.cs
```diff
@@ -0,0 +1,119 @@
...
+        {
+            Result.VerifySuccess(NativeMethods.OgaCreateTokenizer(model.Handle, out _tokenizerHandle));
+        }
+
+        public Sequences EncodeBatch(string[] strings)
+        {
+            IntPtr nativeSequences = IntPtr.Zero;
+            IntPtr stringArray = IntPtr.Zero;
+            Result.VerifySuccess(NativeMethods.OgaCreateStringArray(out stringArray));
+            try
```
yuslepukhin, line 24: Suggest inlining the declaration with the call: out IntPtr stringArray

yuslepukhin (changes requested):

src/ort_genai_c.h
RyanUnderhill: Since we control both sides, what would be your ideal C API to interface with C# for this part?

The C API should ideally allow all language bindings to be as efficient as possible, so I want to avoid forced copies. Do you know if Java/Rust would need to do similar here?

baijumeswani: I don't know the answer to this question. If we don't claim backwards compatibility, we can change this going forward. But i think we should try to make a reasonable decision now to avoid problems for end users.

RyanUnderhill: Let's have both for now, and call the new ones OgaTokenizer*BatchStrings.

Should we call OgaStringArray just OgaStrings? Since we don't say OgaSequenceArray

src/ort_genai_c.cpp
```diff
@@ -246,6 +276,63 @@ OgaResult* OGA_API_CALL OgaTokenizerStreamDecode(OgaTokenizerStream* p, int32_t
...
 }
 
+OgaResult* OGA_API_CALL OgaCreateStrings(OgaStrings** out) {
+  OGA_TRY
+  *out = reinterpret_cast<OgaStrings*>(std::make_unique<std::vector<std::string>>().release());
+  return nullptr;
+  OGA_CATCH
+}
+
+OgaResult* OGA_API_CALL OgaCreateAllocatedStrings(size_t num_strings, size_t* str_lengths, OgaStrings** out) {
```
yuslepukhin, line 286: The problem with this API is that there is a chance of discrepancy betwen num_strings and the actual number of lenghts. With a one by one approach, there is not such a change.

src/ort_genai_c.cpp
```diff
@@ -246,6 +276,63 @@ OgaResult* OGA_API_CALL OgaTokenizerStreamDecode(OgaTokenizerStream* p, int32_t
...
+  OGA_CATCH
+}
+
+size_t OGA_API_CALL OgaStringsGetCount(const OgaStrings* string_array) {
+  return reinterpret_cast<const std::vector<std::string>*>(string_array)->size();
+}
+
+OgaResult* OGA_API_CALL OgaStringsGetString(const OgaStrings* string_array, size_t index, const char** out) {
+  OGA_TRY
+  *out = (*reinterpret_cast<const std::vector<std::string>*>(string_array))[index].c_str();
```
yuslepukhin, line 327: Should check the bounds.

src/ort_genai_c.cpp
```diff
@@ -214,6 +214,36 @@ void OGA_API_CALL OgaTokenizerDestroyStrings(const char** strings, size_t count)
   delete strings;
 }
 
+OgaResult* OGA_API_CALL OgaTokenizerEncodeBatchStrings(const OgaTokenizer* p, const OgaStrings* string_array, OgaSequences** out) {
+  OGA_TRY
+  auto& tokenizer = *reinterpret_cast<const Generators::Tokenizer*>(p);
+  auto& strings = *reinterpret_cast<const std::vector<std::string>*>(string_array);
+  *out = reinterpret_cast<OgaSequences*>(std::make_unique<Generators::TokenSequences>(std::move(tokenizer.EncodeBatch(strings))).release());
```
yuslepukhin, line 221: What does EncodeBatch() returns and why move is necessary?

src/ort_genai_c.cpp
```diff
@@ -214,6 +214,36 @@ void OGA_API_CALL OgaTokenizerDestroyStrings(const char** strings, size_t count)
...
+      std::make_unique<std::vector<std::string>>(tokenizer.DecodeBatch(sequences)).release());
+  return nullptr;
+  OGA_CATCH
+}
+
+OgaResult* OGA_API_CALL OgaTokenizerEncode(const OgaTokenizer* p, const char* str, OgaSequences** out) {
+  OGA_TRY
+  auto& tokenizer = *reinterpret_cast<const Generators::Tokenizer*>(p);
+  auto out_tokens = std::make_unique<Generators::TokenSequences>();
+  out_tokens->emplace_back(std::move(tokenizer.Encode(str)));
```
yuslepukhin, line 241: Move does not seem to be necessary

src/ort_genai_c.cpp
```diff
@@ -246,6 +276,63 @@ OgaResult* OGA_API_CALL OgaTokenizerStreamDecode(OgaTokenizerStream* p, int32_t
...
+  *out = reinterpret_cast<OgaStrings*>(std::make_unique<std::vector<std::string>>().release());
+  return nullptr;
+  OGA_CATCH
+}
+
+OgaResult* OGA_API_CALL OgaCreateAllocatedStrings(size_t num_strings, size_t* str_lengths, OgaStrings** out) {
+  OGA_TRY
+  auto strings = std::make_unique<std::vector<std::string>>(num_strings);
+  for (size_t i = 0; i < num_strings; i++) {
+    strings->at(i).resize(str_lengths[i]);
```
yuslepukhin, line 290: Because it is a bounded loop, at() is not necessary

src/ort_genai_c.cpp
```diff
@@ -246,6 +276,63 @@ OgaResult* OGA_API_CALL OgaTokenizerStreamDecode(OgaTokenizerStream* p, int32_t
...
+  }
+  *out = reinterpret_cast<OgaStrings*>(strings.release());
+  return nullptr;
+  OGA_CATCH
+}
+
+OgaResult* OGA_API_CALL OgaStringsGetBuffer(OgaStrings* string_array, size_t index, char** out) {
+  OGA_TRY
+  *out = &((*reinterpret_cast<std::vector<std::string>*>(string_array))[index])[0];
```
yuslepukhin, line 299: No bounds check

src/ort_genai_c.cpp
```diff
@@ -246,6 +276,63 @@ OgaResult* OGA_API_CALL OgaTokenizerStreamDecode(OgaTokenizerStream* p, int32_t
...
+}
+
+OgaResult* OGA_API_CALL OgaStringsAddString(OgaStrings* string_array, const char* str) {
+  OGA_TRY
+  reinterpret_cast<std::vector<std::string>*>(string_array)->push_back(str);
+  return nullptr;
+  OGA_CATCH
+}
+
+OgaResult* OGA_API_CALL OgaStringsAddStrings(OgaStrings* string_array, const char* const* strings, size_t count) {
```
yuslepukhin, line 311: Same thing here, there is not a way to verify that strings is an array of count.

src/csharp/Tokenizer.cs
```diff
@@ -0,0 +1,133 @@
...
+            IntPtr stringArray = IntPtr.Zero;
+            UIntPtr[] stringLengths = new UIntPtr[strings.Length];
+            for (int i = 0; i < strings.Length; i++)
+            {
+                stringLengths[i] = (UIntPtr)UTF8Encoding.UTF8.GetByteCount(strings[i]);
+            }
+
+            unsafe
+            {
+                fixed (UIntPtr* stringLengthsPtr = stringLengths)
```
yuslepukhin, line 31: fixed is not necessary here. UIntPtr is a blittable type. Change the function signature to accept that and it will do the pinning for you.

src/csharp/Tokenizer.cs
```diff
@@ -0,0 +1,133 @@
...
+                {
+                    Result.VerifySuccess(NativeMethods.OgaCreateAllocatedStrings((UIntPtr)strings.Length, stringLengthsPtr, out stringArray));
+                }
```
yuslepukhin, line 34: Wondering why we need this array that would get destroyed at the end of the function? Can we do without it?
Can we pass strings one by one and encode it directly to the result?

src/csharp/Tokenizer.cs
```diff
@@ -0,0 +1,133 @@
...
+            finally
+            {
+                NativeMethods.OgaDestroyStrings(stringArray);
+            }
+        }
+
+        public string[] DecodeBatch(Sequences sequences)
+        {
+            IntPtr stringArray = IntPtr.Zero;
+            Result.VerifySuccess(NativeMethods.OgaTokenizerDecodeBatchStrings(_tokenizerHandle, sequences.Handle, out stringArray));
```
yuslepukhin, line 57: Same here, stringArray seems to be a pure overhead.

yuslepukhin (changes requested):

src/csharp/Tokenizer.cs
baijumeswani: Changed the code so I now do the EncodeBatch and DecodeBatch by calling Encode and Decode in a loop. Was able to remove the OgaStrings type as a result of this.

src/ort_genai_c.cpp
baijumeswani: Removed OgaStrings type.

src/ort_genai_c.cpp
baijumeswani: Removed OgaStrings type.

src/ort_genai_c.cpp
baijumeswani: Move was not necessary here. Removed it.

src/ort_genai_c.cpp
baijumeswani: Removed OgaStrings type.

src/ort_genai_c.cpp
baijumeswani: Removed OgaStrings type.

src/ort_genai_c.cpp
baijumeswani: Removed OgaStrings type.

src/csharp/Tokenizer.cs
baijumeswani: Removed OgaStrings type. And will call Encode and Decode in a loop for the batch calls.

src/ort_genai_c.cpp
baijumeswani: Move was not necessary here. Removed it.

src/csharp/Tokenizer.cs
```diff
@@ -0,0 +1,116 @@
...
+                throw;
+            }
+        }
+
+        public string[] DecodeBatch(Sequences sequences)
+        {
+            string[] result = new string[sequences.NumSequences];
+            for (ulong i = 0; i < sequences.NumSequences; i++)
+            {
+                IntPtr outStr = IntPtr.Zero;
```
yuslepukhin, line 44: outStr not used?

yuslepukhin (changes requested):

baijumeswani:
Thanks for the review. :)

PR #117, opened 2024-02-24
Update config parsing to handle new builder.py script format

RyanUnderhill:
Tested with Phi2 and the built in tests (includes updated genai_config.json files for tests)

Note: Config parsing will now give errors for unknown values in the config. So if a config loads without errors, all values were recognized.


src/config.cpp
```diff
@@ -16,120 +16,231 @@ ONNXTensorElementDataType TranslateTensorType(std::string_view value) {
...
+    } else if (name == "diversity_penalty") {
+      v_.diversity_penalty = static_cast<float>(value);
+    } else if (name == "length_penalty") {
+      v_.length_penalty = static_cast<float>(value);
+    } else
+      throw JSON::unknown_value_error{};
   }
 
   void OnBool(std::string_view name, bool value) override {
     if (name == "early_stopping") {
```
kunal-vaishnavi, line 212: Is this needed if `early_stopping = true` by default?

src/python/CMakeLists.txt
```diff
@@ -10,7 +10,7 @@ set_target_properties(python PROPERTIES OUTPUT_NAME "onnxruntime_genai")
 
 if(USE_CUDA AND CMAKE_CUDA_COMPILER)
   set_target_properties(python PROPERTIES LINKER_LANGUAGE CUDA)
-  target_link_libraries(python PRIVATE cublasLt cublas cudnn curand cufft cudart)
+  # target_link_libraries(python PRIVATE cublasLt cublas cudnn curand cufft cudart)
```
yufenglee, line 13: why commenting out this line?

src/config.cpp
```diff
@@ -16,120 +16,231 @@ ONNXTensorElementDataType TranslateTensorType(std::string_view value) {
...
  private:
-  Config::Model& model_;
+  Config::Model& v_;
+  EncoderDecoderInit_Element encoder_decoder_init_{v_.encoder_decoder_init};
+  Decoder_Element decoder_{v_.decoder};
 };
 
-struct Root_Element : JSON::Element {
-  explicit Root_Element(Config& config) : config_{config} {}
+struct Search_Element : JSON::Element {
```
yufenglee, line 175: nit: name it as Generato_Element for consistency

test/test_models/hf-internal-testing/tiny-random-gpt2-fp16/genai_config.json
```diff
@@ -0,0 +1,23 @@
+{
+  "model": {
+    "type": "gpt2",
+    "pad_token_id": 98,
+    "bos_token_id": 98,
+    "eos_token_id": 98,
+    "vocab_size": 1000,
+    "logits_type": "float16",
```
yufenglee, line 8: Could we add the capability to infer the type from model?

src/python/CMakeLists.txt
RyanUnderhill: Oops, I didn't mean to check this in.

src/config.cpp
RyanUnderhill: If you want to set it to false, perhaps?

test/test_models/hf-internal-testing/tiny-random-gpt2-fp16/genai_config.json
RyanUnderhill: Yep, that'll be in the next PR. This unblocks people using the updated builder.py

src/config.cpp
RyanUnderhill: Hmm, the values here are for how to do the search (beamsearch/greedysearch).
But I can see how they're also how the generation loop works.
But the GeneratorParams also has the model inputs.
Up to you, I can call this Generator_Element, but it's not directly related to GeneratorParams.

yufenglee (approved):

PR #119, opened 2024-02-26
Update options.cmake to utilize CUDAToolkit_FOUND

skyline75489:
Use `CUDAToolkit_FOUND` forged by [FindCUDAToolkit](https://cmake.org/cmake/help/latest/module/FindCUDAToolkit.html#) (available in CMake 3.17+)

The use case is to automatically fallback to CPU build if CUDA toolkit is not found.


natke:
Thank you @skyline75489. We will consider this functionality. Right now, we will leave it as is, to simplify each of the scenarios.

PR #129, opened 2024-02-28
Search options now settable through a dictionary in python

RyanUnderhill:
Added C APIs to do the equivalent by name/values
Enabled min_length & repetition_penalty search options
Renamed 'search_params' to 'params' as there is now a separate 'search' member. To avoid confusion

Example difference (assuming params is already a GeneratorParams)

```
params.set_search_options({"max_length":256, "min_length":20})
```
old:
```
params.max_length=256
(no equivalent for min_length)
```


src/ort_genai_c.h
```diff
@@ -137,13 +137,8 @@ OGA_EXPORT OgaResult* OGA_API_CALL OgaCreateGeneratorParams(const OgaModel* mode
...
 
-/*
- * \brief Sets the maximum length that the generated sequence can have.
- * \param[in] params The generator params to set the maximum length on.
- * \param[in] max_length The maximum length of the generated sequences.
- * \return OgaResult containing the error message if the setting of the maximum length failed.
- */
-OGA_EXPORT OgaResult* OGA_API_CALL OgaGeneratorParamsSetMaxLength(OgaGeneratorParams* generator_params, size_t max_length);
+OGA_EXPORT OgaResult* OGA_API_CALL OgaGeneratorParamsSetSearchNumber(OgaGeneratorParams* generator_params, const char* name, double value);
+OGA_EXPORT OgaResult* OGA_API_CALL OgaGeneratorParamsSetSearchBool(OgaGeneratorParams* generator_params, const char* name, bool value);
```
yufenglee, line 141: how about 2 overload functions with name OgaGeneratorParamsSetSearchOptions?

yufenglee: Just like what C# APIs

yufenglee: Never mind. C doesn't support overload

RyanUnderhill: I would, but C can't do overloaded functions. No parameter name mangling in C.

src/ort_genai_c.cpp
```diff
@@ -105,10 +105,24 @@ OgaResult* OGA_API_CALL OgaCreateGeneratorParams(const OgaModel* model, OgaGener
 }
 
 OgaResult* OGA_API_CALL OgaGeneratorParamsSetMaxLength(OgaGeneratorParams* params, size_t max_length) {
```
yufenglee, line 107: remove this function

RyanUnderhill: It's already removed. You might be looking at before Baiju's change

RyanUnderhill: Done

yufenglee (approved):

PR #130, opened 2024-02-28
--use_cuda build argument

baijumeswani:


natke (approved):

build.py
```diff
@@ -110,8 +108,11 @@ def build(
     """
     if not is_windows() and not is_linux():
         raise OSError(f"Unsupported platform {platform()}.")
+    
+    if cuda_home and not use_cuda:
+        raise ValueError("cuda_home is specified but use_cuda is not specified.")
```
yufenglee, line 113: I think we may set use_cuda to true if cuda_home is specified explicitly.

PR #131, opened 2024-02-29
Add kv_shared_past_present option

RyanUnderhill:
In the config, there is a new bool option: `{ "model": { "decoder": { "kv_shared_past_present": true } } }` that will cause the KV cache buffers to be allocated once to the max size, and to have the past & present buffers be the same buffers (so no memory allocations after the first iteration, and only half of the max memory).

Does not apply to KV_Cache_Combined.

Tested with phi2 model on cuda

### DIFF
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -107,6 +107,13 @@ struct Decoder_Element : JSON::Element {
       throw JSON::unknown_value_error{};
   }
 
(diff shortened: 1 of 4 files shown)


yufenglee:
could you please set it true in the genai_config.json for cuda case while building the model with builder.py?

yufenglee (approved):

PR #133, opened 2024-02-29
Add from_chars replacement for USE_CXX17

RyanUnderhill:
Uses std::strtod which is available in older C++ versions


RyanUnderhill:
(merged into Jian's change)

PR #135, opened 2024-03-01
Pull latest tokenizer update

RyanUnderhill:


src/tokenizer/tokenizer.cc
```diff
@@ -15,22 +26,26 @@ TfmStatus CreateBPETokenizer(const std::string& tokenizer_path,
...
-               token_cfg->tokenizer_class_ == "CLIPTokenizer" ||
-               token_cfg->tokenizer_class_ == "CodeGenTokenizer") {
-      token_ptr = std::make_unique<BPETokenizer>();
+  std::string type = tokenizer_type;
+  if (type.empty()) {
+    if (BPETokenizer::IsSupportedModel(GetModelName(token_cfg->tokenizer_class_))) {
+      type = "BPE";
+    } else if (std::filesystem::exists(tokenizer_path + "/tokenizer.model")) {
+      // if 'tokenizer.model exists in the tokenizer_path, then it is a sentencepiece model
+      type = "SPM";
```
yufenglee, line 35: This doesn't look right. SPM is not supported anymore.

wenbingl: It's just a type check and only output an error message if the model is unsupported.

yufenglee (approved):

PR #136, opened 2024-03-01
Enable Python Tests on Linux CI and Add a Nightly Pipeline

baijumeswani:
This pull-request:

- enables the python tests on linux x64 cpu and gpu pipelines.
- adds a new pipeline that runs phi-2 end to end by first downloading the model using the model builder and then running the model using onnxruntime-genai. This pipeline runs on linux x64 cpu.

Missing pieces in this pull-request:
- Cuda tests are not enabled yet. This is because of errors with cuda 11.8 and onnxruntime_genai.
- linux arm64 python tests are not enabled. There were problems trying to get that running with uninstalled dependencies such as cython and numpy.


jchen351 (approved):
LGTM

baijumeswani:
Thank you for the review. :)

PR #140, opened 2024-03-02
update outdate parameter

yufenglee:


examples/gemma/gemma-loop.py
```diff
@@ -15,7 +15,7 @@
 input_tokens = tokenizer.encode_batch(prompts)
 
 params=og.GeneratorParams(model)
-params.max_length = 256
+params.set_search_options({"max_length":64})
```
RyanUnderhill, line 18: Was it intentional to change the max from 256 to 64?

examples/gemma/gemma.py
```diff
@@ -14,7 +14,7 @@
     input_tokens = tokenizer.encode(text)
 
     params=og.GeneratorParams(model)
-    params.max_length = 64
+    params.set_search_options({"max_length":256})
```
RyanUnderhill, line 17: Was it intentional change the max from 64 to 256?

RyanUnderhill (approved):

examples/gemma/gemma.py
yufenglee: yes, want to generate longer to interactive case and short for the demon cases

PR #141, opened 2024-03-02
Remove 'set_input_ids' by making tokenizer output 2D arrays

RyanUnderhill:
Simplifies the python, and makes the API more uniform. Also removes the need for get_array()
Also removes unused C APIs as C# doesn't use the batching tokenizer functions.
Updated the C examples to do batching with the tokenizer APIs


examples/phi2/python/phi2-streaming.py
```diff
@@ -30,4 +30,4 @@
 while not generator.is_done():
     generator.compute_logits()
     generator.generate_next_token_top_p(0.7, 0.6)
-    print(tokenizer_stream.decode(generator.get_sequence(0).get_array()[-1]), end='', flush=True)
+    print(tokenizer_stream.decode(generator.get_next_tokens()[0]), end='', flush=True)
```
yufenglee, line 33: could you please also update other examples?

yufenglee (approved):

PR #143, opened 2024-03-04
Fix linux gpu pipeline

aciddelgado:
Two issues resolved here... 
Firstly the main CMakeLists.txt would default to C++17 and the other CMakeLists would sometimes use C++20. Second, on certain compilers on Linux with C++17, filesystem library is not included by default, causing errors in various areas of the code.


jchen351 (approved):

yufenglee:
Could you please add a description on what causes the issue?

PR #150, opened 2024-03-05
Add do_sample config attribute

RyanUnderhill:
Consolidate multiple models into a single 'decoder_only' type
Add additional parameter checking to the search parameters


kunal-vaishnavi (approved):

src/generators.cpp
```diff
@@ -114,35 +114,40 @@ bool Generator::IsDone() const {
...
+  }
+
+  // The user explicitly called TopK_TopP on a beam search
+  if (search_->params_.search.num_beams != 1)
+    throw std::runtime_error("TopK and TopP cannot be used with a beam search");
+
+  // Sanity checks
+  if (top_p < 0.0f || top_p > 1.0f)
+    throw std::runtime_error("top_p must be between 0.0 and 1.0");
+  if (top_k < 1)
```
yufenglee, line 133: The GenerateNextToken_TopP function calls this function with top_k equal to 0

PR #156, opened 2024-03-05
Update Ort to 1.17.1

jchen351:


yufenglee (approved):

baijumeswani:
Do you think we could query the ort version from a file in the repo so instead of updating all these locations, we can just update a single one?
Does not have to be now. We can do this in the future.

baijumeswani (approved):

PR #158, opened 2024-03-06
Add Mistral example

kunal-vaishnavi:
### Description

This PR adds Mistral as an example in the `examples` folder.

### Motivation and Context

Mistral is one of the supported models in ONNX Runtime GenAI.


yufenglee (approved):

README.md
```diff
@@ -30,7 +30,7 @@ Users can call a high level `generate()` method, or run each iteration of the mo
 
 ## Coming very soon
 
-* Support for the Mistral and Whisper model architectures
+* Support for the Whisper model architectures
```
RyanUnderhill, line 33: We do have Whisper support in there now, but it just doesn't handle multiple audio chunks

yufenglee:
Thanks!

README.md
```diff
@@ -49,7 +49,7 @@ Install onnxruntime-genai.
 ```python
 
-model=og.Model(f'models/microsoft/phi-2', device_type)
+model = og.Model(f'models/microsoft/phi-2', device_type)
 
 tokenizer = model.create_tokenizer()
```
RyanUnderhill, line 54: This should be:
`tokenizer = og.Tokenizer(model)`

RyanUnderhill: (while you're modifying this file)

README.md
```diff
@@ -60,11 +60,11 @@ prompt = '''def print_prime(n):
 
 tokens = tokenizer.encode(prompt)
 
-params=og.SearchParams(model)
+params = og.SearchParams(model)
 params.max_length = 200
```
RyanUnderhill, line 64: This should be:

`params.set_search_options({"max_length":200})`

PR #175, opened 2024-03-07
Make Examples Model-Agnostic

aciddelgado:
PR aims to generalize example code to fit any generator model rather than having separate but highly similar examples for different models.


examples/python/model-generate.py
```diff
@@ -0,0 +1,43 @@
...
+def main(args):
+    print("Loading model...")
+    model=og.Model(f'{args.model}', og.DeviceType.CPU if args.execution_provider == 'cpu' else og.DeviceType.CUDA)
+    print("Model loaded")
+    tokenizer=og.Tokenizer(model)
+    tokenizer_stream=tokenizer.create_stream()
+    print("Tokenizer created")
+
+    if args.input is not None:
+        prompt = args.input
```
yufenglee, line 13: could we make prompt an array?

examples/python/model-chat.py
```diff
@@ -0,0 +1,43 @@
...
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description="End-to-end chat-bot example for gen-ai")
+    parser.add_argument('-m', '--model', type=str, required=True, help='Onnx model folder path (must contain config.json and model.onnx)')
+    parser.add_argument('-ep', '--execution_provider', type=str, choices=['cpu', 'cuda'], required=True, help='Execution provider (device) to use, default is CPU, use CUDA for GPU')
+    parser.add_argument('-l', '--max_length', type=int, default=512, help='Max number of tokens to generate after prompt')
+    parser.add_argument('-p', '--top_p', type=float, default=0.9, help='Top p probability to sample with')
+    parser.add_argument('-k', '--top_k', type=int, default=50, help='Top k tokens to sample from')
+    parser.add_argument('-v', '--verbose', action='store_true', help='Print verbose output')
+    args = parser.parse_args()
```
yufenglee, line 42: could we also add a temperature?

examples/python/README.md
```diff
@@ -23,16 +21,30 @@ pip install onnx
 pip install onnxruntime
 ```
 
+Choose a model. Examples of supported ones are:
+- Phi-2
+- Mistral
+- Gemma 2B IT
+- LLama 7B
+
```
yufenglee, line 29: please update the example folder below:

'''bash
cd examples/phi2/python
python -m onnxruntime_genai.models.builder -m microsoft/phi-2 -e cpu -p int4 -o ./example-models/phi2-int4-cpu
'''

examples/python/model-generate.py
```diff
@@ -0,0 +1,45 @@
+import onnxruntime_genai as og
+import argparse
+
+def main(args):
+    if args.verbose: print("Loading model...")
+    model=og.Model(f'{args.model}', og.DeviceType.CPU if args.execution_provider == 'cpu' else og.DeviceType.CUDA)
```
kunal-vaishnavi, line 6:
```suggestion
    model = og.Model(f'{args.model}', og.DeviceType.CPU if args.execution_provider == 'cpu' else og.DeviceType.CUDA)
```

examples/python/model-generate.py
```diff
@@ -0,0 +1,45 @@
...
+    if args.verbose: print("Model loaded")
+    tokenizer=og.Tokenizer(model)
```
kunal-vaishnavi, line 8:
```suggestion
    tokenizer = og.Tokenizer(model)
```

examples/python/model-generate.py
```diff
@@ -0,0 +1,45 @@
...
+    tokenizer_stream=tokenizer.create_stream()
```
kunal-vaishnavi, line 9:
```suggestion
    tokenizer_stream = tokenizer.create_stream()
```

examples/python/model-generate.py
```diff
@@ -0,0 +1,45 @@
...
+    if args.verbose: print("Tokenizer created")
+
+    if args.input is not None:
+        prompt = args.input
+    else:
+        prompt = "I like walking my cute dog"
```
kunal-vaishnavi, line 15: Can we show batched inference with prompts of multiple lengths?

kunal-vaishnavi, line 15: Can we show batched inference with prompts of multiple lengths?

examples/python/model-generate.py
```diff
@@ -0,0 +1,45 @@
...
+
+    generator=og.Generator(model, params)
+    if args.verbose: print("Generator created")
+
+    if args.verbose: print("Running generation loop ...")
+    print(f'\n{prompt}', end='')
+    while not generator.is_done():
+        generator.compute_logits()
+        generator.generate_next_token_top_k_top_p(args.top_k, args.top_p, args.temperature)
+        print(tokenizer_stream.decode(generator.get_next_tokens()[0]), end='', flush=True)
```
kunal-vaishnavi, line 32: Does `flush=True` introduce a noticeable performance hit?

examples/python/model-generate.py
```diff
@@ -0,0 +1,45 @@
...
+
+    if args.input is not None:
+        prompt = args.input
+    else:
+        prompt = "I like walking my cute dog"
+    input_tokens = tokenizer.encode(prompt)
+    if args.verbose: print("Prompt(s) encoded")
+
+    params = og.GeneratorParams(model)
+    params.set_search_options({"max_length": args.max_length, "top_p": args.top_p, "top_k": args.top_k})
```
kunal-vaishnavi, line 20: Should we add `repetition_penalty` as a search option since the GenAI config will use the default value of 1.0?

examples/python/README.md
```diff
@@ -23,16 +21,30 @@ pip install onnx
 pip install onnxruntime
```
kunal-vaishnavi, line 21: Since installing `onnxruntime` vs `onnxruntime-gpu` does not impact the model builder, should we tell the user to install `onnxruntime-gpu` instead to make sure they can run their models on GPU as well?

examples/python/model-generate.py
aciddelgado: should the user be able to input multiple prompts somehow?

examples/python/model-generate.py
```diff
@@ -0,0 +1,53 @@
...
+
+def main(args):
+    if args.verbose: print("Loading model...")
+    model = og.Model(f'{args.model}', og.DeviceType.CPU if args.execution_provider == 'cpu' else og.DeviceType.CUDA)
+    if args.verbose: print("Model loaded")
+    tokenizer = og.Tokenizer(model)
+    if args.verbose: print("Tokenizer created")
+
+    if args.input is not None:
+        prompts = [args.input]
```
yufenglee, line 13: >[args.input] [](http://example.com/codeflow?start=18&length=12)

i meant can we take in array of prompts from users?

examples/python/README.md
```diff
@@ -20,19 +18,33 @@ pip install numpy
...
+```bash
+python model-chat.py -m {path to model folder} -ep {cpu or cuda}
+```
+
+To run the token generation loop example...
 
 ```bash
-python phi2.py
+python model-generate.py -m {path to model folder} -ep {cpu or cuda} -i {string prompt}
 ```
```
yufenglee, line 50: could you please add 2 explicit examples? one with phi2 for chat and one with phi2 for code generation.

examples/python/model-generate.py
```diff
@@ -0,0 +1,53 @@
...
+    
+    print()
+    print(f"Tokens: {len(output_tokens[0])} Time: {run_time:.2f} Tokens per second: {len(output_tokens[0])/run_time:.2f}")
+    print()
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description="End-to-end token generation loop example for gen-ai")
+    parser.add_argument('-m', '--model', type=str, required=True, help='Onnx model folder path (must contain config.json and model.onnx)')
+    parser.add_argument('-ep', '--execution_provider', type=str, choices=['cpu', 'cuda'], required=True, help='Execution provider (device) to use, default is CPU, use CUDA for GPU')
+    parser.add_argument('-i', '--input', type=str, required=False, help='Input prompt to generate tokens from')
```
yufenglee, line 45: >input [](http://example.com/codeflow?start=33&length=5)

nit: rename it to prompts and allow it to take in multiple inputs

examples/python/README.md
aciddelgado: both examples are there, the other is underneath! i can join them into the same bash entry though

yufenglee: how about add 2 concrete examples, like
python -m onnxruntime_genai.model.builder -m microsoft/phi2 -e cpu -p int4 models/phi2-cpu-int4
cd genai_models
python model-chat.py -m "models/phi2-cpu-int4" ...

Maybe create a .sh script to showcase the running of the model?

PR #176, opened 2024-03-08
Remove DeviceType and add "session_options" to genai_config.json

RyanUnderhill:
**Edit: Examples need to be updated, but will wait for the other example simplification fix to go in first**

Example of the new section of the genai_config (and showing some of the new values that can be set):

```
"session_options": {
  "intra_op_num_threads" : 1,
  "inter_op_num_threads": 1,
  "enable_cpu_mem_arena": true,
  "enable_mem_pattern": true,
  "log_id": "genai",
  "log_severity_level": 4,
  "provider_options": [
    {
      "cuda": { "arena_extend_strategy":"kSameAsRequested", "gpu_mem_limit":1000000 }
    }
  ]
},
```

All of the cuda options listed here can be set: https://onnxruntime.ai/docs/execution-providers/CUDA-ExecutionProvider.html#configuration-options
The config just passes the key/value pairs listed to Onnxruntime, genai does not hardcode the options.

Also updated model builder.py to add the "cuda" section when generating models for cuda.


src/python/python.cpp
```diff
@@ -250,6 +243,9 @@ PYBIND11_MODULE(onnxruntime_genai, m) {
         return false;
 #endif
   });
+
+  m.def("set_current_gpu_device_id", [](int device_id) { Ort::SetCurrentGpuDeviceId(device_id); });
+  m.def("get_current_gpu_device_id", []() { return Ort::GetCurrentGpuDeviceId(); });
```
baijumeswani, line 248: Do we not need this in other language bindings?

test/test_models/hf-internal-testing/tiny-random-gpt2-fp32-cuda/past.onnx
baijumeswani, line 1: Can generate these files on the fly using the model builder?

src/python/py/models/builder.py
```diff
@@ -174,6 +178,10 @@ def make_genai_config(self, model_name_or_path, extra_kwargs, out_dir):
             },
         }
 
+        if self.ep == "cuda":
+            cuda_options = { "cuda" : { } }
```
kunal-vaishnavi, line 182: Could we generalize this and remove the `if` statement?

```suggestion
            ep_options = { self.ep : { } }
```

RyanUnderhill: Up to you, but won't any options inside here be provider specific?

test/test_models/hf-internal-testing/tiny-random-gpt2-fp32-cuda/past.onnx
RyanUnderhill: Not yet. Kunal and I tried but the output didn't match yet. It was close, just not the same. There's also bugs in that model, so I'm fine redoing the outputs with the correct token ids (eos token/pad token are not valid in the HF config)

src/python/python.cpp
RyanUnderhill: We should probably have it. I'll add it to the C# bindings.

RyanUnderhill: Done

baijumeswani (approved):

baijumeswani:
Let's cherry pick this and add it to the rc3 package (along with the other fixes).

kunal-vaishnavi (approved):

PR #179, opened 2024-03-08
adding cuda graph 

wangyems:


src/python/py/models/builder.py
```diff
@@ -1164,13 +1180,23 @@ def make_common_mask_reformat_subgraph(self, basename, root_input, unsqueeze_for
...
+            self.model_inputs = ["input_ids", "attention_mask", "position_ids"]
 
     def make_attention_mask_reformatting(self):
         if not self.attention_attrs["use_gqa"]:
             super().make_attention_mask_reformatting()
             return
 
+        if self.enable_cuda_graph:
+            # ORT does not allow nodes to be placed on mulitple execution providers 
+            # with cuda graph enabled. Thus the attention mask is deprecated and the
```
kunal-vaishnavi, line 1195: Should we consider having `seqlens_k` and `total_seq_len` as raw inputs by default instead of `attention_mask` when GQA is used?

wangyems: Yes using seqlens_k and total_seq_len by default for GQA is the goal. But at the moment I'm concerning if the change will affect some customers that already have the inference code that supports masks so just keep both options

src/config.h
```diff
@@ -8,6 +8,11 @@ struct Config {
 
   std::filesystem::path config_path;  // Path of the config directory
 
+  bool use_cuda_graphs{false};
+  // TODO: pass it from config file/generator ctor when serving stack is supported
+  // Hardcoded for now
+  size_t max_batch_size{16};
```
wangyems, line 14: I'm not quite sure how do we pass the max_batch into the config. One option is through config file from model builder, another option is to have an api in the generator. let me know if you have any preference

src/models/position_metadata.h
```diff
@@ -0,0 +1,59 @@
...
+  std::unique_ptr<OrtValue> seqlens_k_;
+  std::array<int64_t, 0> total_sequence_length_shape_{};  // scalar
+  std::unique_ptr<OrtValue> total_sequence_length_;
+
+  std::unique_ptr<OrtValue> position_ids_next_;  // Replaces position_ids_ after the first Run() call
+  std::vector<int32_t> initial_sequence_lengths_;
+
+  // Used for decoding runs with cuda graphs.
+  std::unique_ptr<StaticBuffer> sb_position_ids_{nullptr};
+  std::unique_ptr<StaticBuffer> sb_seqlens_k_{nullptr};
```
RyanUnderhill, line 53: Oops, {nullptr} not needed here for these two.

src/models/position_metadata.h
```diff
@@ -0,0 +1,59 @@
...
+
+  ONNXTensorElementDataType type_;  // Common type for position_ids and attention_mask
+
+  std::array<int64_t, 2> position_ids_shape_{};  // {params.batch_size*params.beam_size, params.sequence_length}
+  std::unique_ptr<OrtValue> position_ids_;
+  std::array<int64_t, 2> attention_mask_shape_{};  // {params.batch_size*params.beam_size, params.sequence_length}
+  std::unique_ptr<OrtValue> attention_mask_;
+  std::array<int64_t, 1> senlens_k_shape_{};  // {params.batch_size*params.beam_size}
+  std::unique_ptr<OrtValue> seqlens_k_;
+  std::array<int64_t, 0> total_sequence_length_shape_{};  // scalar
```
RyanUnderhill, line 45: Interesting, the 0 length array is intentional to create an OrtValue of 1 single value by passing in a 0 sized shape?

src/config.h
wangyems: it's from the user

yufenglee:
@wangyems, Kunal's PR is in. Could you please update this PR?

src/models/static_buffer.cpp
```diff
@@ -0,0 +1,55 @@
+#include "../generators.h"
+#include "../span.h"
+#include "static_buffer.h"
+
+namespace Generators {
+
+StaticBuffer::StaticBuffer(Ort::Allocator* allocator, size_t max_beam_batch_size) : allocator_{allocator}, info_{allocator_->GetInfo()}, max_beam_batch_size_{max_beam_batch_size} {
+}
+
+std::unique_ptr<OrtValue> StaticBuffer::GetOrCreateTensor(std::span<const int64_t> shape,
```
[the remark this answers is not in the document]
wangyems: rename it to CreateTensorOnStaticBuffer()

src/config.h
```diff
@@ -8,6 +8,11 @@ struct Config {
 
   std::filesystem::path config_path;  // Path of the config directory
 
+  bool use_cuda_graphs{false};
```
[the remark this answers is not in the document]
wangyems: I've added the cuda graph config in the json file. the use_cuda_graph is model dependent as of now (we specify which model can use cuda graph and which can not) so config seems a good place to store this parameter.

src/config.h
wangyems: I feel it's better to be put into Generator

src/python/py/models/builder.py
```diff
@@ -1166,8 +1181,14 @@ def has_final_norm(self, module, model):
         return hf_norm or hf_final_layernorm or gguf_final_norm
 
     def make_attention_mask_reformatting(self):
+        if self.enable_cuda_graph:
+            # ORT does not allow nodes to be placed on mulitple execution providers
+            # with cuda graph enabled. Thus the attention mask is deprecated and the
+            # subgraph is replaced with seqlens_k and total_seq_len as the raw
+            # inputs to GroupQueryAttention.
+            return
```
yufenglee, line 1189: so all nodes need to be run on CUDA to enable cuda graph?

wangyems: Yes

src/python/py/models/builder.py
```diff
@@ -40,6 +40,7 @@ def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
 
         # EP-specific variables
         self.ep = ep
+        self.enable_cuda_graph = "enable_cuda_graph" in extra_options and extra_options["enable_cuda_graph"] == "1"
```
kunal-vaishnavi, line 43: Can this be moved within the `cuda` dict in `ep_attrs`?

src/python/py/models/builder.py
```diff
@@ -49,19 +50,21 @@ def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
...
+
         self.inputs = []
         self.outputs = []
         self.initializers = []
         self.value_infos = []
         self.nodes = []
 
         # Map input names to their types and shapes
-        self.input_names = ["input_ids", "attention_mask", "position_ids"]
+        self.input_names = ["input_ids", "attention_mask", "position_ids"] if not self.enable_cuda_graph else ["input_ids", "position_ids", "seqlens_k", "total_seq_len"]
```
kunal-vaishnavi, line 61: Can this conditional logic be moved after the below as another branch case instead?
https://github.com/microsoft/onnxruntime-genai/blob/2fb0c6dacc46169dcbbdeb560492389efa639e6d/src/python/py/models/builder.py#L82-L84

src/python/py/models/builder.py
```diff
@@ -239,6 +244,8 @@ def make_genai_config(self, model_name_or_path, extra_kwargs, out_dir):
 
         if self.ep != "cpu":
             ep_options = { self.ep : self.ep_attrs[self.ep] }
+            if self.enable_cuda_graph and self.ep == "cuda":
```
kunal-vaishnavi, line 247: If `enable_cuda_graph` is defined in `self.ep_attrs["cuda"]`, this logic should get automatically added to `genai_config` and won't be needed.

src/python/py/models/builder.py
```diff
@@ -1789,6 +1804,7 @@ def get_args():
                 exclude_lm_head = Remove language modeling head from your ONNX model.
                     Use this option when you want to remove the language modeling head from within your ONNX model.
                     Instead of `logits`, you will have `hidden_states` as the output to your ONNX model.
+                enable_cuda_graph = 1 : The model can use CUDA graph capture for CUDA execution provider. Limitations may apply.
```
kunal-vaishnavi, line 1807: Can you add how the graph inputs change with CUDA graph enabled?

src/config.h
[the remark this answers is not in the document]
RyanUnderhill: Oh, I just mean this extra value that is not actually in the json, the one inside the json cuda options is fine. It's just duplicated inside the same config. It should either be in the provider or in the root.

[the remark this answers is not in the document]
wangyems: removed

RyanUnderhill (approved):

kunal-vaishnavi (approved):

PR #198, opened 2024-03-14
Cjian/rc3

jchen351:


.pipelines/stages/jobs/steps/nuget-win-step.yml
```diff
@@ -16,7 +16,7 @@ steps:
     DisplayName: 'ESRP - Sign C# dlls'
 
 - powershell: |
-    $VERSION = '0.1.0-rc1'
+    $VERSION = '0.1.0-rc3'
```
baijumeswani, line 19: Can we read from VERSION_INFO file?

jchen351: The rc format for python and nuget is different. Unless we create a different version info file.

baijumeswani (approved):

.pipelines/stages/jobs/steps/nuget-win-step.yml
baijumeswani: Yeah, is it possible to find a common ground between the two?

jchen351: 0.1.0-rc3 vs 0.1.0rc3 we would need to have different suffix, but python packaging directly read from a file.

PR #210, opened 2024-03-16
Add repeat KV to model builder

kunal-vaishnavi:
### Description

This PR adds `repeat_kv` to the model builder for models where `num_attention_heads != num_key_value_heads`.

### Motivation and Context

By supporting `repeat_kv`, models where `num_attention_heads != num_key_value_heads` can now run on both CPU and GPU.

### DIFF
--- a/src/python/py/models/builder.py
+++ b/src/python/py/models/builder.py
@@ -590,39 +590,7 @@ class Model:
         self.make_node("RotaryEmbedding", inputs=inputs, outputs=[output], name=name, domain="com.microsoft", interleaved=0, **kwargs)
         self.make_value_info(output, self.io_dtype, shape=['batch_size', 'sequence_length', self.head_size * (self.num_kv_heads if "k_rotary" in name else self.num_attn_heads)])
 
-    # def make_repeat_kv(self, layer_id, root_input, past_kv, present_kv, **kwargs):
-    #     basename = f"/model/layers.{layer_id}/attn/{'k_proj' if past_kv.endswith('key') else 'v_proj'}/repeat_kv"
-        
-    #     reshape_1_name = f"{basename}/Reshape_1"
-    #     reshape_1_inputs = [root_input, f"/model/constants/TensorProto.INT64/1D/0, 0, {self.num_kv_heads}, -1"]
-    #     self.make_reshape(reshape_1_name, reshape_1_inputs, dtype=self.io_dtype, shape=['batch_size', 'sequence_length', self.num_kv_heads, self.head_size])
-    #     transpose_1_name = f"{basename}/Transpose_1"
-    #     transpose_1_input = f"{reshape_1_name}/output_0"
-    #     self.make_transpose(transpose_1_name, transpose_1_input, dtype=self.io_dtype, shape=['batch_size', self.num_kv_heads, 'sequence_length', self.head_size], perm=[0,2,1,3])
-    #     concat_name = f"{basename}/Concat"
-    #     concat_inputs = [past_kv, f"{transpose_1_name}/output_0"]
-    #     self.make_node("Concat", inputs=concat_inputs, outputs=[present_kv], name=concat_name, axis=2)
-
-    #     unsqueeze_name = f"{basename}/Unsqueeze"
-    #     unsqueeze_inputs = [present_kv, "/model/constants/TensorProto.INT64/1D/2"]
-    #     self.make_unsqueeze(unsqueeze_name, unsqueeze_inputs, dtype=self.io_dtype, shape=['batch_size', self.num_kv_heads, 1, 'sequence_length', self.head_size])
-    #     expand_name = f"{basename}/Expand"
-    #     expand_inputs = [f"{unsqueeze_name}/output_0", ???]
-    #     self.make_expand(expand_name, expand_inputs, dtype=self.io_dtype, shape=['batch_size', self.num_kv_heads, self.num_attn_heads // self.num_kv_heads, 'sequence_length', self.head_size])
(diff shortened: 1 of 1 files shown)


src/python/py/models/builder.py
```diff
@@ -166,15 +166,15 @@
                 "do_sample": config.do_sample if hasattr(config, "do_sample") else False,
                 "early_stopping": True,
                 "length_penalty": config.length_penalty if hasattr(config, "length_penalty") else 1.0,
-                "max_length": config.max_length if hasattr(config, "max_length") else 20,
+                "max_length": self.context_length,
```
yufenglee, line 169: >self.context_length, [](http://example.com/codeflow?start=30&length=20)

is it better to use:
"max_length": config.max_length if hasattr(config, "max_length") else self.context_length,

src/python/py/models/builder.py
```diff
@@ -166,15 +166,15 @@
...
                 "min_length": 0,
                 "no_repeat_ngram_size": config.no_repeat_ngram_size if hasattr(config, "no_repeat_ngram_size") else 0,
                 "num_beams": config.num_beams if hasattr(config, "num_beams") else 1,
                 "num_return_sequences": config.num_return_sequences if hasattr(config, "num_return_sequences") else 1,
-                "past_present_share_buffer": True if self.attention_attrs["op_type"] == "GroupQueryAttention" else False,
+                "past_present_share_buffer": self.attention_attrs["op_type"] == "GroupQueryAttention",
                 "repetition_penalty": config.repetition_penalty if hasattr(config, "repetition_penalty") else 1.0,
                 "temperature": config.temperature if hasattr(config, "temperature") else 1.0,
-                "top_k": config.top_k if hasattr(config, "top_k") else 50,
+                "top_k": 1,
```
yufenglee, line 177: >1, [](http://example.com/codeflow?start=25&length=2)

similar here. is it better to use:
"top_k": config.top_k if hasattr(config, "top_k") else 1,

src/python/py/models/builder.py
```diff
@@ -576,11 +580,195 @@
...
+    def make_repeat_kv(self, layer_id, root_input, past_kv, present_kv, **kwargs):
+        # Make subgraph that repeats tensor of shape (batch_size, sequence_length, num_kv_heads, head_size)
+        # to shape (batch_size, sequence_length, num_attn_heads, head_size) in an interleaved pattern
+        # and updates the KV caches
+        #
+        #           root_input
+        #                |
+        #             Reshape
+        #                |
+        #            Transpose
```
yufenglee, line 603: >Transpose [](http://example.com/codeflow?start=21&length=9)

Why it needs Transpose? Transpose is quite expensive

src/python/py/models/builder.py
kunal-vaishnavi: The `config` object appears to always have the `max_length` attribute so the `else` case is never used. If `max_length` is in the original JSON config, then `config.max_length = json_value`. If `max_length` is not in the original JSON config, then `config.max_length = 20` by default. Because the latter case is more frequent, `max_length` is often set to 20 in the GenAI config.

If `max_length` is in the original JSON config, it is typically set to the context length (e.g. in LLaMA-2 7B [here](https://huggingface.co/meta-llama/Llama-2-7b-hf/blob/8cca527612d856d7d32bd94f8103728d614eb852/generation_config.json#L7)). Thus, it is easier to set `max_length = context_length` as the default in the GenAI config. Users can then manually change the `max_length` as needed.

src/python/py/models/builder.py
kunal-vaishnavi: The `config` object appears to always have the `top_k` attribute so the `else` case is never used. If `top_k` is in the original JSON config, then `config.top_k = json_value`. If `top_k` is not in the original JSON config, then `config.top_k = 50` by default. Because the latter case is more frequent, `top_k` is often set to 50 in the GenAI config.

The `top_k` attribute has not been seen in any of the original JSON configs for the supported model architectures. Setting `top_k = 1` as the default in the GenAI config avoids potential issues with incomprehensible outputs. Users can manually change `top_k` as needed.

src/python/py/models/builder.py
kunal-vaishnavi:
```
                                        past_kv (BxNxSxH)
                                           |
root_input --> Reshape --> Transpose --> Concat (BxNxSxH)
  (BxSxD)     (BxSxNxH)    (BxNxSxH)       |
                                       present_kv (BxNxSxH)
```
The root input for the K path is the output from `RotaryEmbedding` and the output is of shape `BxSxD`. The root input for the V path is the output from `MatMul` and the output is of shape `BxSxD`. Thus, the root input to `repeat_kv` is of shape `BxSxD` in both cases.

The past key/past value and present key/present value are each of shape `BxNxSxH`. In order to concatenate the past key/past value with the current key/current value, the current key/current value has to be of shape `BxNxSxH`. The `Reshape` converts the root input to shape `BxSxNxH` and the `Transpose` converts it to shape `BxNxSxH`. Hence, the `Transpose` is needed.

src/python/py/models/builder.py
yufenglee: if config always have the top_k attribute, we should use it instead of setting it to constant 1, right?

kunal-vaishnavi: In the `config` object, `config.top_k = 50` usually so the GenAI config will have `"top_k": 50` and then model outputs won't make sense.

yufenglee:
Could you please run a benchmark on the Gemm 2b model on CPU?

yufenglee (approved):

PR #212, opened 2024-03-16
Make Model and GeneratorParams be a shared_ptr vs unique_ptr

RyanUnderhill:
For safety. This will ensure the Model object's lifetime matches that of any Generator using it.

Might also be useful to do the same for TokenizerStream and Tokenizer.

Starting with this so reviewers can judge if the approach makes sense. The C API is unaffected.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,6 +10,8 @@ Users can call a high level `generate()` method, or run each iteration of the mo
 * Built in logits processing like repetition penalties
 * Easy custom scoring
 
+See full documentation at [https://onnxruntime.ai/docs/genai].
+
 ## Features
 
 * Supported model architectures:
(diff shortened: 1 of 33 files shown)


baijumeswani:
In what scenarios would the Model get destructed before the Generator? I am ok with the change, but it feels unnecessary?

From the C api perpective, the Model would be the first thing that should get created, and then all other onnxruntime-genai components (like the Tokenizer, GeneratorParams, Generator). So, it would make sense that the Model should get destructed last. Is the concern that the user may accidentally destroy the Model before the other components and this is a safeguard against that?

RyanUnderhill:
> In what scenarios would the Model get destructed before the Generator? I am ok with the change, but it feels unnecessary?
> 
> From the C api perpective, the Model would be the first thing that should get created, and then all other onnxruntime-genai components (like the Tokenizer, GeneratorParams, Generator). So, it would make sense that the Model should get destructed last. Is the concern that the user may accidentally destroy the Model before the other components and this is a safeguard against that?

Python could garbage collect in any order, and C API users could also destroy in any order. Similar with any other languages. It's simple to just make it always work no matter what the user does vs allowing accessing deleted memory bugs.

baijumeswani (approved):

PR #221, opened 2024-03-21
Remove GenerateNextToken* special case functions

RyanUnderhill:
set_search_options already supports the functionality, so the extra functions are confusing users since there are multiple ways to do the same thing.
set_search_options is also more flexible as it supports all future options without the need for extra APIs.


yufenglee:
Could you please check all the examples and benchmark tool?

RyanUnderhill:
> Could you please check all the examples and benchmark tool?

Yep, this previous change removed all references to the functions: https://github.com/microsoft/onnxruntime-genai/pull/217

So this change doesn't affect any of the existing python scripts.

yufenglee (approved):

PR #225, opened 2024-03-22
Easy to use C++ API Wrapper

RyanUnderhill:
Wraps the C API in a 0-overhead C++ style API that automatically manages resources and gives C++ style interfaces. It removes all of the C "glue" and should be much simpler and safer to use.

I switched the C API example to use it, plus our C API unit tests.

Before:
```
  OgaModel* model;
  CheckResult(OgaCreateModel("phi-2", OgaDeviceTypeCPU, &model));
  OgaModelPtr model_ptr{model};

  OgaTokenizer* tokenizer;
  CheckResult(OgaCreateTokenizer(model, &tokenizer));
  OgaTokenizerPtr tokenizer_ptr{tokenizer};

  const char* prompt = "def is_prime(num):";
  std::cout << "Prompt: " << std::endl << prompt << std::endl;

  OgaSequences* sequences;
  CheckResult(OgaCreateSequences(&sequences));
  OgaSequencesPtr sequences_ptr{sequences};
  CheckResult(OgaTokenizerEncode(tokenizer, prompt, sequences));

  OgaGeneratorParams* params;
  CheckResult(OgaCreateGeneratorParams(model, &params));
  OgaGeneratorParamsPtr params_ptr{params};
  CheckResult(OgaGeneratorParamsSetSearchNumber(params, "max_length", 200));
  CheckResult(OgaGeneratorParamsSetInputSequences(params, sequences));

  OgaSequences* output_sequences;
  CheckResult(OgaGenerate(model, params, &output_sequences));
  OgaSequencesPtr output_sequences_ptr{output_sequences};

  size_t sequence_length = OgaSequencesGetSequenceCount(output_sequences, 0);
  const int32_t* sequence = OgaSequencesGetSequenceData(output_sequences, 0);

  const char* out_string;
  CheckResult(OgaTokenizerDecode(tokenizer, sequence, sequence_length, &out_string));

  std::cout << "Output: " << std::endl << out_string << std::endl;
```

After:
```
  auto model = OgaModel::Create("phi-2");

  const char* prompt = "def is_prime(num):";
  std::cout << "Prompt: " << std::endl << prompt << std::endl;

  auto sequences = OgaSequences::Create();
  tokenizer->Encode(prompt, *sequences);

  params->SetSearchOption("max_length", 200);
  params->SetInputSequences(*sequences);

  auto output_sequences = model->Generate(*params);
  auto out_string = tokenizer->Decode(output_sequences->Get(0));

  std::cout << "Output: " << std::endl << out_string << std::endl;
```


yufenglee:
for the example, could we keep both C and C?

RyanUnderhill:
> for the example, could we keep both C and C++?

Ah, it's not easy to see in the diff, but there are both examples in the main.cpp file:

void CXX_API()
void C_API()

https://github.com/microsoft/onnxruntime-genai/blob/9d37e073f5cb1b545876af46c132bef3ef67f758/examples/c/src/main.cpp

src/ort_genai.h
baijumeswani, line 1: Shall we rename to `onnxruntime_genai.h` and `onnxruntime_genai_c.h`?

yufenglee (approved):

src/ort_genai.h
RyanUnderhill: We can, but I'll do that as a separate change as it's going to break a lot of links.

PR #227, opened 2024-03-23
Add Android build support

skottmckay:
Update build.py
  - setup like ORT/ort-ext with update/build/test/clean phases
  - support cross compiling for android
  - refactored to try and simplify
    - split out arg creation and validation cleanly
    - use helper from util folder

To use: Download ORT AAR from https://mvnrepository.com/artifact/com.microsoft.onnxruntime/onnxruntime-android/1.17.1
Extract AAR and use `--ort_path` arg to point to the directory with that. Will look for the lib and headers in the correct place for the ABI.


build.py
```diff
@@ -5,245 +5,400 @@
...
 
-    build_wheel = "OFF" if skip_wheel else "ON"
-    build_dir = os.path.abspath(build_dir) if build_dir else os.path.join(os.getcwd(), "build")
+    command += [f"-DCMAKE_BUILD_TYPE={args.config}"]
+
+    build_wheel = "OFF" if args.skip_wheel else "ON"
 
     command += [
         "-S",
         ".",
```
edgchen1, line 312:
```suggestion
        str(REPO_ROOT),
```

build.py
```diff
@@ -5,245 +5,400 @@
...
+    )
 
-    Args:
-        command_or_path (str): The command or path to resolve.
+    parser.add_argument(
+        "--config",
+        default="RelWithDebInfo",
+        type=str,
+        choices=["Debug", "MinSizeRel", "Release", "RelWithDebInfo"],
+        help="Configuration(s) to build.")
```
edgchen1, line 50:
```suggestion
        help="Configuration to build.")
```

build.py
```diff
@@ -5,245 +5,400 @@
...
+            raise ValueError(f"{args.ort_home} does not exist or is not a directory.")
+
+        args.ort_home = args.ort_home.resolve(strict=True)
+
+
+def _create_env(args: argparse.Namespace):
+    env = os.environ.copy()
+
+    if args.use_cuda:
+        env["CUDA_HOME"] = args.cuda_home
```
edgchen1, line 255:
```suggestion
        env["CUDA_HOME"] = str(args.cuda_home)
```

build.py
```diff
@@ -5,245 +5,400 @@
...
+    build_wheel = "OFF" if args.skip_wheel else "ON"
 
     command += [
         "-S",
         ".",
         "-B",
-        build_dir,
+        str(args.build_dir),
         "-DCMAKE_POSITION_INDEPENDENT_CODE=ON",
         "-DUSE_CXX17=ON",
```
edgchen1, line 316: looks like the variable `USE_CXX17` is not used now in the CMake files

build.py
```diff
@@ -5,245 +5,400 @@
...
-        parallel=args.parallel,
-        config=args.config,
-    )
+    if not (util.is_windows() or util.is_linux() or util.is_mac()):
+        raise OSError(f"Unsupported platform {sys.platform}.")
+
+    arguments = _parse_args()
+
+    _validate_args(arguments)
+    environment = _create_env(args)
```
edgchen1, line 395:
```suggestion
    environment = _create_env(arguments)
```

build.py
```diff
@@ -5,245 +5,394 @@
...
 import sys
 import warnings
 
+from pathlib import Path
 
-def is_windows():
-    """Check if the current platform is Windows."""
-    return sys.platform.startswith("win")
+REPO_ROOT = Path(__file__).parent
+sys.path.append(str(REPO_ROOT / "tools" / "python"))
```
baijumeswani, line 16:
```suggestion
sys.path.append(os.fspath(REPO_ROOT / "tools" / "python"))
```

build.py
```diff
@@ -5,245 +5,394 @@
...
+
+        # override if we're cross-compiling
+        # TODO: Add ios and arm64 support
+        if args.android:
+            target_sys = "Android"
+        elif platform.system() == "Darwin":
+            # also tweak build directory name for mac builds
+            target_sys = "macOS"
+
+        args.build_dir = Path("build/" + target_sys)
```
baijumeswani, line 162:
```suggestion
        args.build_dir = Path("build") / target_sys
```

build.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+
+python3 $SCRIPT_DIR/build.py "$@"
```
edgchen1, line 7:
```suggestion
python3 "$SCRIPT_DIR/build.py" "$@"
```

build.py
```diff
@@ -5,245 +5,394 @@
...
-        native_lib_path = f"/p:NativeBuildOutputDir={os.path.join(build_dir, config)}"
+    if util.is_windows() and not args.skip_csharp:
+        dotnet = str(_resolve_executable_path("dotnet"))
 
         # Build the library
-        properties = [configuration, platorm, native_lib_path]
         csharp_build_command = [dotnet, "build", ".",]
-        run_subprocess(csharp_build_command + properties, cwd=os.path.join("src", "csharp")).check_returncode()
+        csharp_build_command += _get_csharp_properties(args)
+        util.run(csharp_build_command, cwd=str(REPO_ROOT / "src" / "csharp"))
```
edgchen1, line 355: nit: the `util.run` type annotation says `cwd` is `Path | None`

tools/python/util/run.py
```diff
@@ -37,18 +39,22 @@ def run(
...
+
+    _log.info(f"Running subprocess in '{cwd}'\n  {' '.join([shlex.quote(arg) for arg in cmd])}")
 
     def output(is_stream_captured):
         return subprocess.PIPE if is_stream_captured else (subprocess.DEVNULL if quiet else None)
 
     completed_process = subprocess.run(
         cmd,
-        cwd=cwd,
+        cwd=str(cwd),
```
edgchen1, line 57: subprocess.run's cwd also accepts Path

edgchen1 (approved):

build.py
skottmckay: Does the pathlib.Path str() implementation not handle this?

baijumeswani: pathlib.Path str() will handle this. This can become a problem when the type of input is not known to be of type pathlib.Path. For example, if REPO_ROOT was not a pathlib.Path type, it would lead to unexpected errors.

In this scenario, I think it doesn't matter. So, I think it should be ok to leave this as is.

build.py
skottmckay: Updated type annotation to match what subprocess run supports (it calls os.fsdecode on the input) and removed unnecessary conversion to string.

baijumeswani (approved):

edgchen1 (approved):

PR #229, opened 2024-03-25
Android app starter.

skottmckay:
Very basic app to get the JNI side of thing setup. Sets the output in a text box and hangs as the model runs on startup with a hardcoded prompt and there are no UI elements for the user to do anything.  

Need to manually copy model to device as it's too large to input in the app asserts. See the readme. 

Tested on emulator. Currently set to log each token as generated by calling back to the Java level (calls `GenAIWrapper.gotNextToken`), but that needs to be wired up to display the token somewhere the user can see (e.g. something like pass a delegate to `GenAIWrapper.run` that update a text box in the UI so `gotNextToken` can call that). 

When token generation is complete it returns the full string for convenience.

Throws a Java exception (GenAIException) is there's an issue but exceptions are currently unhandled in the Java code.

I _think_ all the low-level stuff is implemented and it just needs the UI built out.


natke:
This has been moved to https://github.com/microsoft/onnxruntime-inference-examples/pull/420

PR #231, opened 2024-03-25
Mergin rel-0.1.0 back to main

jchen351:


examples/python/README.md
```diff
@@ -4,6 +4,10 @@
 
 Install the python package according to the [installation instructions](https://onnxruntime.ai/docs/genai/howto/install).
 
+  ```bash
+  cd build/wheel
+  pip install onnxruntime-genai-*.whl
```
baijumeswani, line 9: I think this should be `pip install onnxruntime_genai-*.whl`

baijumeswani (approved):

PR #234, opened 2024-03-25
/CETCOMPAT

jchen351:


snnn (approved):
You may consider putting the settings in cmake preset json files

PR #236, opened 2024-03-26
add benchmark for memory usage

rui-ren:
## Add memory footprint option

Command line:
```
python benchmark_e2e.py -i models/ -b 1 -l 128 -g 256 -r 100 -w 10 -k 5 -o myTest -pm True
```

Memory usage will be printed in the terminal and save in csv file for the benchmark.


benchmark/python/output
```diff
@@ -0,0 +1,13 @@
+Batch Size,1.0
```
zhijxu-MS, line 1: do we need to commit the output file into repo?

rui-ren: Yes, I am about to do that. Thanks!

yufenglee: any update on this? we can merge after removing the output

yufenglee:
Does the memory monitor have impact on the performance? if so, could we add an option to control it and turn it off by default

natke:
@rui-ren Can you please address the comment above and then we can complete this review

rui-ren:
> @rui-ren Can you please address the comment above and then we can complete this review

Done, thank you.

benchmark/python/benchmark_e2e.py
```diff
@@ -36,36 +79,78 @@ def generate_prompt(model, tokenizer, prompt_length, use_graph_capture) -> str:
...
+    return metrics
+
 def run_benchmark(args, model, tokenizer, batch_size, prompt_length, generation_length, max_length):
+
     # Get user arguments
     num_repetitions = args.repetitions
     temperature = 1.0
 
+    # Get tokenizer, and model
+    if args.verbose: print(f"Loading model... ")
```
pengwa, line 148:
```suggestion
    if args.verbose: print("Loading model... ")
```

pengwa (approved):

yufenglee (approved):

PR #239, opened 2024-03-27
Introduce a simple local server built into the library

baijumeswani:


src/ort_genai.h
```diff
@@ -84,12 +91,21 @@ struct OgaSequences : OgaAbstract {
     return OgaSequencesCount(this);
   }
 
+#if __cplusplus >= 202002L
   std::span<const int32_t> Get(size_t index) const {
```
RyanUnderhill, line 95: Hmm, maybe just have separate 'GetData' and 'GetCount' functions that always exist, and just make the span one not exist with less than C++20?

test/c_api_tests.cpp
```diff
@@ -123,7 +123,11 @@ TEST(CAPITests, GreedySearchGptFp32CAPI) {
     auto sequence = generator->GetSequence(i);
 
     auto* expected_output_start = &expected_output[i * max_length];
+#if __cplusplus >= 202002L
     EXPECT_TRUE(0 == std::memcmp(expected_output_start, sequence.data(), max_length * sizeof(int32_t)));
```
RyanUnderhill, line 127: Where was this failing to build without this change?

PR #241, opened 2024-03-29
DML - Preload DirectML.dll to not use OS version

RyanUnderhill:
If we fail to preload, we abort using DirectML as using the OS copy will fail when we try to use it.
Also Copy DirectML.dll to the install folder as part of the other onnxruntime files

### DIFF
--- a/src/models/model.cpp
+++ b/src/models/model.cpp
@@ -13,6 +13,23 @@
 //  Because dml_provider_factory includes windows headers that #define min and max, this next line will prevent this from happening
 #define NOMINMAX
 #include "dml_provider_factory.h"
+
+EXTERN_C IMAGE_DOS_HEADER __ImageBase;
+
+static std::wstring CurrentModulePath() {
+  wchar_t path[MAX_PATH];
+  GetModuleFileNameW((HINSTANCE)&__ImageBase, path, _countof(path));
+
+  wchar_t absolute_path[MAX_PATH];
+  wchar_t* name;
+  GetFullPathNameW(path, _countof(path), absolute_path, &name);
+
+  auto idx = std::distance(absolute_path, name);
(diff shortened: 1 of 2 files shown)


yufenglee (approved):

src/models/model.cpp
```diff
@@ -302,6 +319,9 @@ void Model::CreateSessionOptions() {
       Ort::ThrowOnError(Ort::api->GetExecutionProviderApi("DML", ORT_API_VERSION, reinterpret_cast<const void**>(&p_dml_api)));
       if (!p_dml_api)
         throw std::runtime_error("Unexpected nullptr getting OrtDmlApi");
+      auto directml_dll = CurrentModulePath() + L"DirectML.dll";
+      if (LoadLibraryExW(directml_dll.c_str(), nullptr, 0) == NULL)
+        throw std::runtime_error("DirectML.dll not found");
```
snnn, line 324: If the DLL is not in current module path, it still could somewhere else?
And here you may need to distinguish between "failed to load" vs "DLL not found".

src/models/model.cpp
```diff
@@ -13,6 +13,23 @@
...
 #define NOMINMAX
 #include "dml_provider_factory.h"
+
+EXTERN_C IMAGE_DOS_HEADER __ImageBase;
+
+static std::wstring CurrentModulePath() {
+  wchar_t path[MAX_PATH];
+  GetModuleFileNameW((HINSTANCE)&__ImageBase, path, _countof(path));
+
+  wchar_t absolute_path[MAX_PATH];
```
snnn, line 23: Try to avoid using MAX_PATH, because the value of the macro is very low. If I remember correctly, it is just around 256. We should support longer paths.

PR #242, opened 2024-03-29
Add fp32 test to nightly run

natke:


RyanUnderhill (approved):

test/python/test_onnxruntime_genai_e2e.py
```diff
@@ -53,5 +53,7 @@ def run_model(model_path: str | bytes | os.PathLike):
     for model_name in ["microsoft/phi-2"]:
         with tempfile.TemporaryDirectory() as temp_dir:
             device = "cuda" if og.is_cuda_available() else "cpu"
-            download_model(temp_dir, device, model_name)
+            download_model(temp_dir, device, model_name, int4)
```
baijumeswani, line 56: I am a little confused. How does python know that `int4` is a `str`? Should it not be `"int4"`?

test/python/test_onnxruntime_genai_e2e.py
```diff
@@ -53,5 +53,7 @@ def run_model(model_path: str | bytes | os.PathLike):
...
             run_model(temp_dir)
+            download_model(temp_dir, "cpu", model_name, fp32)
```
baijumeswani, line 58: Do you not want to execute the model in this test?

test/python/test_onnxruntime_genai_e2e.py
natke: Fixed

test/python/test_onnxruntime_genai_e2e.py
natke: Fixed

baijumeswani: I think you can do this:

```python
    for precision in ["int4", "fp32"]:
        with tempfile.TemporaryDirectory() as temp_dir:
            device = "cuda" if og.is_cuda_available() else "cpu"
            download_model(temp_dir, device, model_name, precision)
            run_model(temp_dir)
```

natke: Updated

baijumeswani (approved):

PR #243, opened 2024-03-30
Add C++ benchmark program

edgchen1:
Add a C++ model benchmark program. It is modeled after the existing Python benchmark script (benchmark/python/benchmark_e2e.py).

The motivation for a C++ version is to be able to run without Python. This is useful for Android.

### DIFF
--- a/build.py
+++ b/build.py
@@ -95,6 +95,7 @@ def validate_cuda_home(cuda_home: str | bytes | os.PathLike | None):
 
 def build(
     skip_wheel: bool = False,
+    skip_python: bool = False,
     use_cuda: bool | None = None,
     use_dml: bool | None = None,
     cuda_home: str | bytes | os.PathLike | None = None,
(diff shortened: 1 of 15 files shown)


build.py
```diff
@@ -95,6 +95,7 @@ def validate_cuda_home(cuda_home: str | bytes | os.PathLike | None):
 
 def build(
     skip_wheel: bool = False,
+    skip_python: bool = False,
```
RyanUnderhill, line 98: What is this solving? Should probably be a separate change if needed.

edgchen1: reverted these changes for this PR. I found it useful to make optional though.

src/ort_genai.h
```diff
@@ -4,7 +4,7 @@
 #pragma once
 
 #include <memory>
-#include <span>
+#include "span.h"  // TODO Use the std span header if we can assume C++20.
```
RyanUnderhill, line 7: Hmm, this will break our release package as we don't ship span.h with this file. Baiju has another PR where we put the span functions inside an `#if __cplusplus >= 202002L`

https://github.com/microsoft/onnxruntime-genai/pull/239

What do you think?

test/c_api_tests.cpp
```diff
@@ -120,21 +120,27 @@ TEST(CAPITests, GreedySearchGptFp32CAPI) {
 
   // Verify outputs match expected outputs
   for (int i = 0; i < batch_size; i++) {
-    auto sequence = generator->GetSequence(i);
```
RyanUnderhill, line 123: Why did this need to change?

edgchen1: this file builds with C++17 too. now, GetSequence() is only defined for C++20.

RyanUnderhill (approved):

PR #253, opened 2024-04-08
Refactor model-qa

natke:


examples/python/assistant.ipynb
```diff
@@ -0,0 +1,150 @@
...
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "id": "569d2618-ff32-466a-8bec-eeb967ee364b",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import onnxruntime_genai as og\n",
+    "import argparse\n",
```
baijumeswani, line 11: argparse is not needed in this notebook.

examples/python/assistant.ipynb
```diff
@@ -0,0 +1,150 @@
...
+   "execution_count": 7,
+   "id": "add45ace-14be-4ab3-a68c-303aebeea18c",
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "Loading model...\n",
+      "Model loaded in 41.10 seconds\n"
```
baijumeswani, line 26: Loading phi2 on my machine is much faster than this.
Since this is really machine dependent, can we avoid having the output cached in the notebook?

examples/python/model-qa.py
```diff
@@ -0,0 +1,74 @@
...
+    model = og.Model(f'{args.model}')
+    model_loaded_timestamp  = time.time()
+    if args.verbose:
+        print("Model loaded in {:.2f} seconds".format(model_loaded_timestamp - app_started_timestamp))
+    tokenizer = og.Tokenizer(model)
+    tokenizer_stream = tokenizer.create_stream()
+    if args.verbose: print("Tokenizer created")
+    if args.verbose: print()
+
+    # Keep asking for input prompts in an loop
```
baijumeswani, line 22:
```suggestion
    # Keep asking for input prompts in a loop
```

examples/python/assistant.ipynb
natke: We can. How do we do that?

examples/python/assistant.ipynb
```diff
@@ -0,0 +1,299 @@
...
+     ]
```
baijumeswani, line 174:
```suggestion
```

examples/python/assistant.ipynb
baijumeswani: I added a suggestion. Just remove the text block from the ipynb

natke: Do you mean manually, before checking it in?

baijumeswani: Yes, or you could also click on the clear cell outputs before checking it in.

yufenglee:
why an olive example here?

natke:
> why an olive example here?

This is for the demo

natke:
> why an olive example here?

I moved it into a separate repo

yufenglee (approved):

PR #254, opened 2024-04-09
Switch set_search_options from taking a dict to kwargs

RyanUnderhill:
Now 20% more pythonic!


baijumeswani (approved):

src/python/python.cpp
```diff
@@ -94,7 +94,7 @@ struct PyGeneratorParams {
     }
   }
 
-  void SetSearchOptions(const pybind11::dict& dict) {
+  void SetSearchOptions(const pybind11::kwargs& dict) {
```
yufenglee, line 97: could we list out all the valid search options here:
https://github.com/microsoft/onnxruntime-genai/blob/18adb67dce145579779dd0b6a3e2b25826a7b459/src/python/python.cpp#L182

RyanUnderhill: Do it as python help on that function? We should provide a url since the options match the config documentation.

yufenglee: i mean add a comment here

baijumeswani (approved):

src/python/python.cpp
RyanUnderhill: Done

PR #260, opened 2024-04-11
Add cpu logits conversion from fp16 to fp32

RyanUnderhill:
Convenience addition for models that run on devices besides CPU but don't output fp32.

Prior to this change, there would be an error if the model outputs fp16 and our scoring device is the CPU.


baijumeswani (approved):
Changes look good. Any way to add a test?

natke:
> Changes look good. Any way to add a test?

You're singing my song!

yufenglee:
how is the perf? the Float16ToFloat32 implementation is not efficient.

yufenglee:
> how is the perf? the Float16ToFloat32 implementation is not efficient.

https://github.com/microsoft/onnxruntime/blob/ee603ee3265dbf6eac112baf273b6b69bf696085/onnxruntime/core/mlas/inc/mlas.h#L1034

PR #262, opened 2024-04-16
Updates to Android genai demo app

edgchen1:
- Main change is to avoid creating the GenAIWrapper repeatedly. The instance also needs to be explicitly closed. This should fix the crash observed after repeatedly sending a prompt.
- Some other refactoring.


YUNQIUGUO (approved):
ah awesome! thanks for the fix - yeah it should be avoided from creating the GenAIWrapper instance repeatedly.

PR #266, opened 2024-04-16
Add basic DML support

PatriceVignola:
This adds basic DML support (but not performant). It currently requires a private local build to run, but for the sake of timelines and keeping PRs small, I'm making the PR right now.


PatriceVignola:
Closing. Just noticed that the CUDA graph has been merged and will need to make it work with it (and fix merge conflicts)

PR #268, opened 2024-04-16
Add basic DML support

PatriceVignola:
This adds basic DML support (although not performant). The next step will be to implement DML graph functionality.

This currently requires private local ORT and DML builds to run, but official builds are coming soon.


RyanUnderhill (approved):
Looks good, the only part I had to think about for a sec were these:
`auto& allocator = model_.device_type_ == DeviceType::DML ? model_.allocator_cpu_ : *model_.allocator_device_;`

I initially thought it should be checking for cuda also, but the logic is really "DML uses CPU for this codepath", so maybe put a comment?

wangyems:
LGTM. Please fix the conflicts

wangyems (approved):

PR #271, opened 2024-04-18
Attention mask with cuda graph

wangyems:
To keep attention_mask as model's input when using cuda graph, we assume shared_kv_cache is used and attention_mask is allocated with [max_batch, max_seq_length]

1. revert seqlens_k and total_seq_len in model builder and GenAI tool.
2. allocate attention_mask with fixed size [max_batch, max_seq_length] if using cuda graph
3. add readme

### DIFF
--- a/src/models/kernels.cu
+++ b/src/models/kernels.cu
@@ -1,71 +1,87 @@
-#include <cuda_runtime.h>
 #include <cuda_fp16.h>
+#include <cuda_runtime.h>
 #include <stdint.h>
 
 namespace Generators {
 namespace cuda {
 
-template <typename T>
-__global__ void UpdatePositionIds(T* positions, int batch_beam_size) {
-  int i = blockIdx.x * blockDim.x + threadIdx.x;
-  if (i < batch_beam_size)
-    positions[i]++;
+template <typename T> __global__ void UpdatePositionIds(T *positions, int batch_beam_size) {
+    int i = blockIdx.x * blockDim.x + threadIdx.x;
+    if (i < batch_beam_size)
+        positions[i]++;
 }
 
-template <typename T>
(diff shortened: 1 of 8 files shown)


src/python/py/models/README.md
```diff
@@ -3,6 +3,7 @@
 This folder contains the model builder for quickly creating optimized and quantized ONNX models within a few minutes that run with ONNX Runtime GenAI.
 
 # Contents
+
```
kunal-vaishnavi, line 6: Can you add `Enable CUDA Graph` as a section in the table of contents?

src/models/kernels.cu
```diff
@@ -1,71 +1,87 @@
...
-    } else {
-      mask_data[i * current_length + j] = 1;
+__global__ void CopyAndUpdateAttentionMask(T *mask_data, const T *old_mask_data, int batch_beam_size,
+                                           int current_length, int max_length) {
+    int global_index = blockIdx.x * blockDim.x + threadIdx.x;
+    int i = global_index / current_length;
+    int j = global_index % current_length;
+    if (i < batch_beam_size) {
+        if (j < current_length - 1) {
+            mask_data[i * max_length + j] = old_mask_data[i * (current_length - 1) + j];
```
yufenglee, line 29: >max_length [](http://example.com/codeflow?start=26&length=10)

is it max_length or current_length? Do you need a max_length for copy and update scenario?

yufenglee (approved):
:shipit:

src/models/kernels.cu
wangyems: The new mask uses max length and the old mask uses current length because we allocate max length to mask and enable cuda graph in the second iteration

kunal-vaishnavi (approved):

PR #284, opened 2024-04-23
Support eos_token_id being an array

RyanUnderhill:
Some models have the eos_token_id being an array of tokens vs a single value. This adds support for it by processing the logits in such a way that the first eos_token_id in the array gets the max score of all of the ones in the array.

This way, if any of the eos_token_id scores is high scoring, the primary eos_token_id score will get that score and the scoring code will check for it as before.

### DIFF
--- a/src/models/logits.cpp
+++ b/src/models/logits.cpp
@@ -186,8 +186,11 @@ void Logits::HandleEOSArray(cpu_span<float> batched_logits) {
   for (int index = 0; index < shape_[0]; index++) {
     auto logits = batched_logits.subspan(vocab_index, vocab_size);
     float max = std::numeric_limits<float>::lowest();
-    for (auto id : model_.config_->model.eos_token_ids)
+    for (auto id : model_.config_->model.eos_token_ids) {
       max = std::max(max, logits[id]);
+      logits[id] = std::numeric_limits<float>::lowest();  // Set all EOS token options to never happen (the first will get the max of all)
+    }
+
     logits[model_.config_->model.eos_token_id] = max;  // Set the score of the primary EOS token to the highest of any of the EOS tokens
     vocab_index += vocab_size;
   }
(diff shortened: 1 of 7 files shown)


src/models/model.cpp
```diff
@@ -537,6 +537,7 @@ std::unique_ptr<OrtValue> Model::ExpandInputs(std::unique_ptr<OrtValue>& input,
 }
 
 void Model::GetMaxBatchSizeFromGeneratorParams(const GeneratorParams& params) {
```
yufenglee, line 539: >GetMaxBatchSizeFromGeneratorParams [](http://example.com/codeflow?start=12&length=34)

why is it commented out?

src/models/logits.cpp
```diff
@@ -126,33 +139,58 @@ RoamingArray<float> Logits::Get() {
...
+  size_t vocab_index = 0;  // Simpler math to have this index go up by vocab_size for every logit chunk we process
+
+  for (int index = 0; index < shape_[0]; index++) {
+    auto logits = batched_logits.subspan(vocab_index, vocab_size);
+    float max = std::numeric_limits<float>::lowest();
+    for (auto id : model_.config_->model.eos_token_ids)
+      max = std::max(max, logits[id]);
+    logits[model_.config_->model.eos_token_id] = max;  // Set the score of the primary EOS token to the highest of any of the EOS tokens
+    vocab_index += vocab_size;
+  }
```
yufenglee, line 193: does it work for sampling?

src/models/model.cpp
RyanUnderhill: Whoops, that wasn't supposed to be merged.

src/models/logits.cpp
RyanUnderhill: It does now

yufenglee (approved):
:shipit:

kunal-vaishnavi (approved):

PR #322, opened 2024-04-24
Update badges to show latest pre release

natke:
You can't currently show a pre-release version of a PyPI package, so thinking it's more informative to show the latest NuGet, which will display pre-release


EwoutH:
Related issue: https://github.com/badges/shields/issues/4692

baijumeswani (approved):

PR #330, opened 2024-04-25
Cherry-pick #327 for merging to main

baijumeswani:
Cherry-pick #327 for merging to main

This PR is to sync rel-0.2.0 branch with main as far as the c api goes.

We should probably try to make OgaGenerate function take in a `const OgaModel*` and the OgaCreateGenerator should also take in a const OgaModel* instead of a OgaModel*.

This will require moving max_batch_size_ and use_cuda_graph_ out of Model, and probably use the generator params to control these variables. I don't think these two should be part of Model. @wangyems

### DIFF
--- a/benchmark/c/main.cpp
+++ b/benchmark/c/main.cpp
@@ -112,7 +112,7 @@ void WriteE2EStats(std::string_view label,
             << "\n";
 }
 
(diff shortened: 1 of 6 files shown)


wangyems:
Yes, we can refactor the input_ids/position_ids.. 's class to be constructed with generator params to restore the constness of OgaModel*.

baijumeswani:
Closing as this has already been merged to main.

PR #356, opened 2024-04-30
Make OgaModel* const again

baijumeswani:
The changes in https://github.com/microsoft/onnxruntime-genai/pull/179 was setting a `Model` member variable at `Generator` creation time. At `Generator` creation time, the `Model` is const and as a result cannot set the member variable.

The intermediate changes that needed to be made to support 0.2.0 release was to make `OgaModel*` non const in the c api.

This pull-request tries to set the variable at the time of `GeneratorParams` creation and avoids needing to make `const OgaModel*` non-const.

### DIFF
--- a/src/ort_genai_c.cpp
+++ b/src/ort_genai_c.cpp
@@ -108,7 +108,7 @@ OgaResult* OGA_API_CALL OgaGeneratorParamsSetSearchBool(OgaGeneratorParams* gene
 OgaResult* OGA_API_CALL OgaGeneratorParamsTryGraphCaptureWithMaxBatchSize(OgaGeneratorParams* generator_params, int32_t max_batch_size) {
   OGA_TRY
   auto* params = reinterpret_cast<Generators::GeneratorParams*>(generator_params);
-  params->max_batch_size = max_batch_size;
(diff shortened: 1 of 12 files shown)


src/ort_genai_c.cpp
```diff
@@ -108,7 +108,7 @@ OgaResult* OGA_API_CALL OgaGeneratorParamsSetSearchBool(OgaGeneratorParams* gene
 OgaResult* OGA_API_CALL OgaGeneratorParamsTryGraphCaptureWithMaxBatchSize(OgaGeneratorParams* generator_params, int32_t max_batch_size) {
   OGA_TRY
   auto* params = reinterpret_cast<Generators::GeneratorParams*>(generator_params);
-  params->max_batch_size = max_batch_size;
+  params->TryGraphCapture(max_batch_size);
```
yufenglee, line 111: >TryGraphCapture [](http://example.com/codeflow?start=10&length=15)

OgaGeneratorParamsTryGraphCaptureWithMaxBatchSize can be called for CPU EP. in this case, it should does nothing

yufenglee (approved):
:shipit:

wangyems (approved):

PR #357, opened 2024-04-30
Shutdown in C++ and C#

baijumeswani:


examples/c/src/main.cpp
```diff
@@ -92,6 +92,9 @@ int main(int argc, char** argv) {
     return -1;
   }
 
+  // Responsible for cleaning up the library during shutdown
```
baijumeswani, line 95: This will impact users trying 0.2.0-rc4 with this sample.
Maybe we should wait to merge this PR? Or cherry-pick it to 0.2.0-rc5?

examples/csharp/HelloPhi2/Program.cs
```diff
@@ -1,6 +1,8 @@
 ﻿// See https://aka.ms/new-console-template for more information
 using Microsoft.ML.OnnxRuntimeGenAI;
 
+OgaHandle ogaHandle = new OgaHandle();
```
baijumeswani, line 4: Same for this sample.

src/ort_genai.h
```diff
@@ -235,3 +235,8 @@ struct OgaGenerator : OgaAbstract {
 
   static void operator delete(void* p) { OgaDestroyGenerator(reinterpret_cast<OgaGenerator*>(p)); }
 };
+
+struct OgaHandle {
+  OgaHandle() = default;
+  ~OgaHandle() { OgaCheckResult(OgaShutdown()); }
```
edgchen1, line 241: probably shouldn't throw an exception from the dtor.
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c36-a-destructor-must-not-fail

baijumeswani: Thanks for the reminder. :)

yufenglee (approved):

PR #372, opened 2024-05-01
Remove debug prompt print

natke:


examples/python/phi3-qa.py
```diff
@@ -29,8 +29,6 @@ def main(args):
         # If there is a chat template, use it
         prompt = f'{chat_template.format(input=text)}'
 
-        print(f"Prompt: {prompt}")
```
kunal-vaishnavi, line 32: Is the prompt already being printed with the generated text?

natke: No it's not. The text that the user enters is printed as they enter it. Then only the output is printed

baijumeswani (approved):

Issue #336, opened 2024-04-27
Error while running Phi-3 with DML


tomas-pet:
Here was my input command: 
python model-qa.py -m Phi-3-mini-128k-instruct-onnx/directml/directml-int4-awq-block-128 -l 2048

Here is the error I am getting:
Input: hi

Output: Traceback (most recent call last):
  File "model-qa.py", line 82, in <module>
    main(args)
  File "model-qa.py", line 47, in main
    generator.compute_logits()
onnxruntime_genai.onnxruntime_genai.OrtException: Failed to parse the cuda graph annotation id: -1

PatriceVignola:
Hi @tomas-pet,
Which version of onnxruntime-genai-directml do you have?

tomas-pet:
I am using latest

tomas-pet:
Any update on this?

natke:
@tomas-pet We need some more information to try and repro this.

Can you share the output of pip list?

Which model are you using? Did you download it from HuggingFace?

Can you share the genai_config.json file from the model folder please.

tomas-pet:
Here is output of pip list:
Package                    Version
-------------------------- ---------------------
accelerate                 0.29.2
aiohttp                    3.9.3
aiosignal                  1.3.1
async-timeout              4.0.3
attrs                      23.2.0
auto-gptq                  0.7.1
certifi                    2024.2.2
charset-normalizer         3.3.2
cmake                      3.29.1
colorama                   0.4.6
coloredlogs                15.0.1
datasets                   2.18.0
diffusers                  0.27.2
dill                       0.3.8
filelock                   3.13.3
flatbuffers                24.3.25
frozenlist                 1.4.1
fsspec                     2024.2.0
gekko                      1.1.0
huggingface-hub            0.22.2
humanfriendly              10.0
idna                       3.6
importlib-metadata         7.1.0
inquirerpy                 0.3.4
Jinja2                     3.1.3
MarkupSafe                 2.1.5
mpmath                     1.3.0
multidict                  6.0.5
multiprocess               0.70.16
networkx                   3.1
numpy                      1.24.4
onnx                       1.16.0
onnxruntime-directml       1.17.3
onnxruntime-genai          0.1.0
onnxruntime-genai-directml 0.2.0rc4
optimum                    1.18.0
ort-nightly-qnn            1.18.0.dev20240428001
packaging                  24.0
pandas                     2.0.3
peft                       0.10.0
pfzy                       0.3.4
[...]

PR #377, opened 2024-05-01
Ignore software adapters from DML device enumeration and use DXGI enumeration instead of dxcore

PatriceVignola:
There are 2 changes in this PR:

1. Ignore software adapters like WARP and remote desktop from the enumeration. There's a weird behavior with these adapters where they will show up first no matter what, even if the `HighPerformance` enumeration is requested. This can cause very slow execution when using remote desktop, or simply if a device decides to enumerate the WARP software emulation device first.
2. Use DXGI instead of DXCORE. DXGI works everywhere that dxcore works (except for Linux), but DXGI is supported on a wider range of Windows machines going back to the early days of Windows 10. This should allow the enumeration to work on far more devices.


baijumeswani (approved):

yufenglee:
Does it fix this issue: https://github.com/microsoft/onnxruntime-genai/issues/336

yufenglee (approved):

PR #385, opened 2024-05-02
Initial PR for enabling iOS build

YUNQIUGUO:
Initial PR for enabling iOS build for ORT GenAI

- Support build .dylib for the initial change and one architecture/platform at a time for now.
- Usage:
1. Have libonnxruntime.dylib available or build one manually from ORT using "--build_dynamic_framework" option.
2. Under ${ORT_GENAI_PROJECT_SOURCE_REPO}, create a folder as the following structure:
```
     ort/
       include/
          onnxruntime_c_api.h
       lib/
          libonnxruntime.dylib
```
This will be used as ${ORT_HOME} root directory in Oga build.

3. Use the build command as below: (or similarly adjust for iphoneos/arm64, etc.)
```
python3 build.py --parallel --build_dir ./build_ios_simulator --ios --ios_sysroot iphonesimulator --ios_arch x86_64 --ios_deployment_target 13.0 --cmake_generator Xcode
```


build.py
```diff
@@ -39,33 +50,48 @@ def _parse_args():
...
-    parser.add_argument("--skip_csharp", action="store_true", help="Skip building the C# API.")
+    parser.add_argument(
+        "--skip_tests", action="store_true", help="Skip all tests. Overrides --test."
+    )
+    parser.add_argument(
+        "--skip_wheel", action="store_true", help="Skip building the Python wheel."
+    )
+    parser.add_argument(
+        "--skip_csharp", action="store_true", help="Skip building the C# API."
+    )
```
skottmckay, line 82: Is there a requirement for this sort of formatting? personally I think it reduces readability in cases like this so I'm not a fan. IIRC it's the `black` formatter. I know we use it in ORT but don't in extensions. 

I'm fine with using formatters for consistency though - I'd just prefer to use things that don't reduce readability when the developer has made specific choices about where to put line breaks. FWIW this has added ~100 lines of whitespace to the existing script.

build.py
```diff
@@ -214,13 +295,41 @@ def _validate_android_args(args: argparse.Namespace):
...
+            )
 
         # no C# on Android so automatically skip
         if not args.skip_csharp:
             args.skip_csharp = True
 
 
+def _validate_ios_args(args: argparse.Namespace):
+    if args.ios:
+        # TODO: cross compile for iOS on Linux?
```
skottmckay, line 309: Is cross compiling on Linux possible?

build.py
```diff
@@ -214,13 +295,41 @@ def _validate_android_args(args: argparse.Namespace):
...
+        if util.is_mac():
```
skottmckay, line 310: Should it be an error if this is not a mac?

build.py
```diff
@@ -214,13 +295,41 @@ def _validate_android_args(args: argparse.Namespace):
...
+            ]
+            arg_names = [
+                "--ios_sysroot          "
+                + "<the location or name of the macOS platform SDK>",
+                "--osx_arch             "
+                + "<the Target specific architectures for iOS>",
+                "--apple_deploy_target  "
+                + "<the minimum version of the target platform>",
+            ]
+            if not all(_ is not None for _ in needed_args):
```
skottmckay, line 324: nit: would avoiding a double negative be a little easier to read?

```suggestion
            have_required_args = all(_ is not None for _ in needed_args)
            if not have_required_args:
```

build.py
```diff
@@ -319,16 +429,41 @@ def update(args: argparse.Namespace, env: dict[str, str]):
...
+                    args.android_ndk_path
+                    / "build"
+                    / "cmake"
+                    / "android.toolchain.cmake"
+                ).resolve(strict=True)
+            ),
             f"-DANDROID_PLATFORM=android-{args.android_api}",
             f"-DANDROID_ABI={args.android_abi}",
-            f"-DENABLE_PYTHON=OFF"
+            f"-DENABLE_PYTHON=OFF",
```
skottmckay, line 450:
```suggestion
            "-DENABLE_PYTHON=OFF",
```

Existing code, but this line doesn't actually need the `f`

build.py
```diff
@@ -214,13 +295,41 @@ def _validate_android_args(args: argparse.Namespace):
...
 
 
+def _validate_ios_args(args: argparse.Namespace):
+    if args.ios:
+        # TODO: cross compile for iOS on Linux?
+        if util.is_mac():
+            needed_args = [
+                args.ios_sysroot,
+                args.osx_arch,
+                args.apple_deployment_target,
```
edgchen1, line 314: maybe think about the naming of these options - using "ios", "osx", and "apple" here seems inconsistent. if we want to make it general, maybe use "apple". if it's iOS-specific, use "ios".

build.py
edgchen1: I think it's probably simpler to just require a Mac host.

build.py
```diff
@@ -12,11 +12,22 @@
 
 from pathlib import Path
 
+
+class BaseError(Exception):
+    """Base class for errors originating from build.py."""
+
+
+class BuildError(BaseError):
```
edgchen1, line 20: maybe just use `ValueError` to be consistent with the existing code.

build.py
```diff
@@ -319,16 +429,41 @@ def update(args: argparse.Namespace, env: dict[str, str]):
...
             f"-DANDROID_PLATFORM=android-{args.android_api}",
             f"-DANDROID_ABI={args.android_abi}",
-            f"-DENABLE_PYTHON=OFF"
+            f"-DENABLE_PYTHON=OFF",
+        ]
+
+    if args.ios:
+        command += [
+            "-DCMAKE_SYSTEM_NAME=iOS",
+            "-DENABLE_TESTS=OFF",
```
edgchen1, line 456: do we need to not build the tests, or can we just skip running them for now when cross compiling?

build.py
```diff
@@ -112,6 +153,32 @@ def _parse_args():
         help="Path to the Android NDK. Typically `<Android SDK>/ndk/<ndk_version>`.",
     )
 
+    # iOS build options
+    parser.add_argument("--ios", action="store_true", help="build for ios")
```
edgchen1, line 157: could we put all the cross compiling options (`ios`, `android`, and whatever other ones may be added) into a mutually exclusive group? I assume we won't support cross-compiling for more than one platform in one build.py call.

build.py
YUNQIUGUO: right - I think it's possible but there are probably a couple issues as cross build ios on Linux, such as requires a complicated tool chain setup/no easier build set up for ios simulator for CI testing purposes, etc..

removed this comment. just put it there as a question/potential todo.

build.py
YUNQIUGUO: make sense - reverted some of the formatting changes.

build.py
YUNQIUGUO: make sense - made them all "ios"

build.py
```diff
@@ -221,6 +250,34 @@ def _validate_android_args(args: argparse.Namespace):
...
+            args.ios_arch,
+            args.ios_deployment_target,
+        ]
+        arg_names = [
+            "--ios_sysroot          "
+            + "<the location or name of the macOS platform SDK>",
+            "--ios_arch             "
+            + "<the Target specific architectures for iOS>",
+            "--ios_deployment_target  "
+            + "<the minimum version of the target platform>",
```
skottmckay, line 269: nit: I don't think the line breaks are required and it makes it harder to parse that this is 1:1 with the needed_args list.

```suggestion
            "--ios_sysroot           <the location or name of the macOS platform SDK>",
            "--ios_arch              <the Target specific architectures for iOS>",
            "--ios_deployment_target <the minimum version of the target platform>",
```

skottmckay (approved):

PR #389, opened 2024-05-02
Model generate template

natke:
A few updates to the example model scripts:
- Add template functionality to the model-generate script
- Set a default length for the phi-3 script so that users don't have to
- Calculate token timings for the generate script

### DIFF
--- a/examples/python/phi3-qa.py
+++ b/examples/python/phi3-qa.py
@@ -15,6 +15,12 @@ def main(args):
     if args.verbose: print("Tokenizer created")
     if args.verbose: print()
     search_options = {name:getattr(args, name) for name in ['do_sample', 'max_length', 'min_length', 'top_p', 'top_k', 'temperature', 'repetition_penalty'] if name in args}
+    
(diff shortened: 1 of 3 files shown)


examples/python/phi3-qa.py
```diff
@@ -15,6 +15,12 @@ def main(args):
     if args.verbose: print("Tokenizer created")
     if args.verbose: print()
     search_options = {name:getattr(args, name) for name in ['do_sample', 'max_length', 'min_length', 'top_p', 'top_k', 'temperature', 'repetition_penalty'] if name in args}
+    
+    # Set the max length to something sensible by default, unless it is specified by the user,
+    # since otherwise it will be set to the entire context length
+    if 'max_length' not in search_options:
+        search_options['max_length'] = 2048
```
RyanUnderhill, line 22: Is this only for phi-3 vs the other examples?

natke: Just for Phi-3. This is the entry point that folks are running, linked from the HuggingFace model card and blog

RyanUnderhill (approved):

PR #391, opened 2024-05-03
Exclude libnvinfer_plugin.so.10

jchen351:


baijumeswani (approved):
Do we know what triggered the sudden failure?

yufenglee:
Can we have a include list instead of an exclude list?

jchen351:
There are so many of them from Cuda.

jchen351:
> Do we know what triggered the sudden failure?

Might be some changes from ORT that triggers it. We are using the latest nightly build now.

Issue #384, opened 2024-05-02
Created onnx version of ChatQA-1.5-8B model with Unknown value: pad_token_id error on inferencing


AshD:
Model: https://huggingface.co/nvidia/ChatQA-1.5-8B
Conversion Command: python -m onnxruntime_genai.models.builder -m ./ChatQA-1.5-8B -o chatqacpu -p int4 -e cpu -c cache

Runs without errors and creates the onnx model. 

Ran the onnx model. C# exception: 
`Microsoft.ML.OnnxRuntimeGenAI.OnnxRuntimeGenAIException: 'Error encountered while parsing 'C:\AI\onnx\chatqacpu\genai_config.json' JSON Error: Unknown value: pad_token_id at line 33 index 26'`

Thanks,
Ash

natke:
Hi @AshD We are looking into this

baijumeswani:
@AshD what version of onnxruntime-genai are you using?

AshD:
Microsoft.ML.OnnxRuntimeGenAI.DirectML 0.2.0-rc4

kunal-vaishnavi:
The `genai_config.json` file shows a list of pad token ids rather than a single value. This will require a change in ONNX Runtime GenAI to support a list of pad token ids. A similar change was made [here](https://github.com/microsoft/onnxruntime-genai/pull/284) to support a list of EOS token ids.

AshD:
Thanks for the fix. I did a github pull of the latest code and created a new model.
python builder.py -m ./ChatQA-1.5-8B -o chatqa -p int4 -e dml -c cache

When running the newly created model, I get a new error. 

`Microsoft.ML.OnnxRuntimeGenAI.OnnxRuntimeGenAIException: 'NO_SUCH_FIELD: The JSON field referenced does not exist in this object.'` 

Using Microsoft.ML.OnnxRuntimeGenAI.DirectML 0.2.0-rc4 in C# when running.

AshD:
Exception Stack Trace: 
```
 	Microsoft.ML.OnnxRuntimeGenAI.dll!Microsoft.ML.OnnxRuntimeGenAI.Result.VerifySuccess(nint nativeResult) Line 20	C#
>	Microsoft.ML.OnnxRuntimeGenAI.dll!Microsoft.ML.OnnxRuntimeGenAI.Tokenizer.Tokenizer(Microsoft.ML.OnnxRuntimeGenAI.Model model) Line 14	C#
[...]

PR #394, opened 2024-05-03
Fix pad token id in config

kunal-vaishnavi:
### Description

This PR sets `pad_token_id` in `genai_config.json` to a single value when a model does not specify a pad token id but it specifies a list of EOS token ids.

### Motivation and Context

When the pad token id is not specified, `pad_token_id` in `genai_config.json` stores the same value that `eos_token_id` in `genai_config.json` contains. When `eos_token_id` has a list of EOS token ids, then `pad_token_id` also has a list of pad token ids. This causes a parsing issue in ONNX Runtime GenAI because it expects only one pad token id.

This PR also fixes [this issue](https://github.com/microsoft/onnxruntime-genai/issues/384).

### DIFF
--- a/src/python/py/models/builder.py
+++ b/src/python/py/models/builder.py
@@ -228,7 +228,7 @@ class Model:
                     "num_key_value_heads": self.num_kv_heads,
                 },
                 "eos_token_id": config.eos_token_id,
-                "pad_token_id": config.pad_token_id if hasattr(config, "pad_token_id") and config.pad_token_id is not None else config.eos_token_id,
+                "pad_token_id": config.pad_token_id if hasattr(config, "pad_token_id") and config.pad_token_id is not None else config.eos_token_id[0] if type(config.eos_token_id) == list else config.eos_token_id,
                 "type": self.model_type[ : self.model_type.find("For")].lower(),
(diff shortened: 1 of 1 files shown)


src/python/py/models/builder.py
```diff
@@ -228,7 +228,7 @@ def make_genai_config(self, model_name_or_path, extra_kwargs, out_dir):
                     "num_key_value_heads": self.num_kv_heads,
                 },
                 "eos_token_id": config.eos_token_id,
-                "pad_token_id": config.pad_token_id if hasattr(config, "pad_token_id") and config.pad_token_id is not None else config.eos_token_id,
+                "pad_token_id": config.pad_token_id if hasattr(config, "pad_token_id") and config.pad_token_id is not None else config.eos_token_id[0] if type(config.eos_token_id) == list else config.eos_token_id,
```
baijumeswani, line 231:
```suggestion
                "pad_token_id": config.pad_token_id if hasattr(config, "pad_token_id") and config.pad_token_id is not None else config.eos_token_id[0] if isinstance(config.eos_token_id, list) else config.eos_token_id,
```

baijumeswani (approved):

PR #401, opened 2024-05-04
Fix DML crash and update release version

PatriceVignola:


VERSION_INFO
```diff
@@ -1 +1 @@
-0.2.0rc5
\ No newline at end of file
+0.2.0rc6
```
baijumeswani, line 1: no new line.

PatriceVignola, line 1: Is this newline needed? I don't remember.

yufenglee: the new line needs to be removed

yufenglee (approved):

PR #402, opened 2024-05-04
re-enable DML GQA fast path

zhangxiang1993:


PatriceVignola:
We can't merge this until we fix the regression in DML and release the next version.

PR #411, opened 2024-05-07
[Experiment] Use ONNX IR in model builder

justinchuby:


baijumeswani:
@justinchuby is this PR still relevant?

justinchuby:
Depending on what people think. This PR is trying to prove a point and show how one can use the IR in model builder to construct models robustly. You can decide if you would like it to be taken further.

justinchuby:
I do recommend that any proto construction and manipulation logic be migrated to use the IR

PR #413, opened 2024-05-07
Fetch tokenizer from onnxruntime-extensions

baijumeswani:
With this PR, we depend on onnxruntime-extensions for the tokenizer instead of maintaining a duplicate copy inside ort-genai.


RyanUnderhill (approved):
Nice, looks like the code changes were what I expected.

baijumeswani:
Thank you for the review. :)

PR #418, opened 2024-05-08
Adding dependency libraries to zip package

jchen351:
This PR request primarily focuses on changes to adding ONNXRUNTIME libraries to the zip packages. Depends on the Target EP, it could also contain cuda_provider library or DML libraries.

Changes to `cmake/package.cmake`:

* [`cmake/package.cmake`](diffhunk://#diff-5f5cbcf4c6721a0a3c7b347a598906a68658622f06352b6f8c07d45d1e11e0c4L7-R67): Modified the `set_target_properties` to provide more specific destination paths for the library, runtime, and public header. Also, updated the `CPACK_PACKAGE_FILE_NAME` to be more specific based on the system and processor type. This helps in creating more specific and identifiable package names.
* 

Changes to `libonnxruntime.so` handling: To reduce linux package size. 
 The major changes include replacing copy commands with symbolic links for `libonnxruntime.so` and modifying the `set_target_properties` in `cmake/package.cmake` to provide more specific destination paths and package file names based on the system and processor type.

* [`.github/workflows/linux-cpu-x64-build.yml`](diffhunk://#diff-e758ed36b14737d7c89353329cb735a6d081f5fa288b809d49e47bfc0ade96c3L67-R68): Replaced the copy command with a symbolic link for `libonnxruntime.so`. This creates a symbolic link to the original file instead of creating a copy, saving disk space and ensuring any updates to the original file are reflected in the linked file.
* [`.github/workflows/linux-gpu-x64-build.yml`](diffhunk://#diff-e941545683ca21b146fc6f414d9fe9e311f5f5efc44fce47287e8b0d835434aeL81-R82): Similar to the above change, replaced the copy command with a symbolic link for `libonnxruntime.so`.
* [`.pipelines/stages/jobs/steps/utils/download-ort.yml`](diffhunk://#diff-b33e2529c1e7627822171d08975dbb8c52cdde1bcce3a5931a90ddf24c5019b1L50-R50): Again, replaced the copy command with a symbolic link for `libonnxruntime.so`.


cmake/package.cmake
```diff
@@ -4,26 +4,65 @@ set_target_properties(
 )
 install(TARGETS
   onnxruntime-genai
-  LIBRARY
-  PUBLIC_HEADER
+  LIBRARY DESTINATION lib
+  RUNTIME DESTINATION lib
```
jchen351, line 8: without doing this onnxruntime-genai.dll will be in bin folder

snnn: But, why it should not be in the bin folder?

.github/workflows/linux-cpu-x64-build.yml
```diff
@@ -64,7 +64,8 @@ jobs:
           mv ${{ env.ORT_PACKAGE_NAME }}/build/native/include ort/
```
snnn, line 64: Is it a pull request pipeline  or a packaging pipeline?

snnn:
What dependency libraries are you adding? Would you mind explaining that?

.github/workflows/linux-cpu-x64-build.yml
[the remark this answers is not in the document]
snnn: Supposedly you should do it in ADO?

cmake/package.cmake
jchen351: I thought ORT has the dll in lib, but if you think we can put it in bin folder, that is fine too.

.github/workflows/linux-cpu-x64-build.yml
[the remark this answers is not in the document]
jchen351: Removed

cmake/package.cmake
```diff
@@ -13,17 +13,52 @@ set(CPACK_RESOURCE_FILE_LICENSE "${PROJECT_SOURCE_DIR}/LICENSE")
...
+
 set(CPACK_INCLUDE_TOPLEVEL_DIRECTORY TRUE)
 install(FILES
   "${PROJECT_SOURCE_DIR}/README.md"
   "${PROJECT_SOURCE_DIR}/ThirdPartyNotices.txt"
   "${PROJECT_SOURCE_DIR}/SECURITY.md"
   "${PROJECT_SOURCE_DIR}/LICENSE"
   DESTINATION .)
+
+install(FILES
```
jchen351, line 60: @snnn here is all libraries we are adding

cmake/package.cmake
```diff
@@ -13,17 +13,52 @@ set(CPACK_RESOURCE_FILE_LICENSE "${PROJECT_SOURCE_DIR}/LICENSE")
...
+  ${onnxruntime_libs}
```
jchen351, line 61: @snnn here is all libraries we are adding

.pipelines/stages/jobs/steps/utils/download-ort.yml
```diff
@@ -47,7 +47,7 @@ steps:
       set -e -x
       ort_stable_version=$(echo $(ort_version) | cut -d- -f1)
       echo $ort_stable_version
-      cp libonnxruntime.so libonnxruntime.so.$ort_stable_version
+      ln -s libonnxruntime.so libonnxruntime.so.$ort_stable_version
```
snnn, line 50: The order is wrong. 
```
ln -s <source> <target>
```
libonnxruntime.so should be a symlink.

jchen351: libonnxruntime.so here is source here, unless we want to rename the it to libonnxruntime.so.$ort_stable_version first.

snnn: When you build a C/C++ program that uses onnxruntime, the program's binary will require loading libonnxruntime.so.***.***  (with the version number) instead of libonnxruntime.so.  However, the file you found from our nuget package is without the version number, because zip format doesn't support symlink. That's the reason why this extra step is needed.

snnn: Read this： https://tldp.org/HOWTO/Program-Library-HOWTO/shared-libraries.html

natke: @snnn does this need updating?

jchen351: It is already updated?

cmake/package.cmake
snnn: On Windows DLLs should be in BIN folder instead of LIB. You may need to manually check where the files go.

snnn (approved):

PR #420, opened 2024-05-08
Add GroupQueryAttention on CPU in model builder

kunal-vaishnavi:
### Description

This PR adds `GroupQueryAttention` to ONNX models generated for CPU.

### Motivation and Context

This PR is a follow up to [this PR](https://github.com/microsoft/onnxruntime-genai/pull/270).


src/python/py/models/builder.py
```diff
@@ -167,17 +167,21 @@ def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
...
+        valid_gqa_configurations = [
+            ("cpu", TensorProto.FLOAT),
+            ("cuda", TensorProto.FLOAT16),
+            ("dml", TensorProto.FLOAT16),
+        ]
+        if (self.ep, self.io_dtype) in valid_gqa_configurations:
             # Change model settings for GroupQueryAttention
             self.attention_attrs["op_type"] = "GroupQueryAttention"
-            print("GroupQueryAttention (GQA) is used in this model.")
+            print("GroupQueryAttention (GQA) is used in this model. GQA is not currently supported for FP32 CUDA.")
```
yufenglee, line 178: what is this message for "GQA is not currently supported for FP32 CUDA."? It is already checked with float16.

src/python/py/models/builder.py
```diff
@@ -563,11 +567,13 @@ def make_matmul_fp16_or_fp32(self, matmul, name, root_input, **kwargs):
...
-        # Note: Packed MatMul is of shape 3NxH instead of Hx3N because `make_matmul` will apply a transpose before saving
-        N, H = q_matmul.shape
-        matmul = np.stack((q_matmul.transpose(), k_matmul.transpose(), v_matmul.transpose()), axis=1).reshape(H, 3*N).transpose()
+        # N_q = num_attention_heads * head_size, N_kv = num_key_value_heads * head_size, H = hidden_size
+        # Combine 3 MatMuls of shape N_q x H, N_kv x H, N_kv x H into 1 packed MatMul of shape (N_q+N_kv+N_kv)xH
+        # Note: Packed MatMul is of shape (N_q+N_kv+N_kv)xH instead of Hx(N_q+N_kv+N_kv) because `make_matmul` will
+        # apply a transpose before saving
+        N_q, H = q_matmul.shape
+        N_kv, _ = k_matmul.shape
+        matmul = np.concatenate([q_matmul.transpose(), k_matmul.transpose(), v_matmul.transpose()], axis=1).reshape(H, N_q + N_kv + N_kv).transpose()
```
yufenglee, line 576: does it work to concatenate on axis = 0?

src/python/py/models/builder.py
kunal-vaishnavi: Models created with `--precision fp32 --execution_provider cuda` via the model builder will not have `GroupQueryAttention`. This message is to let users know that.

yufenglee: the message is already guarded by the condition: (self.ep, self.io_dtype) in valid_gqa_configurations

src/python/py/models/builder.py
kunal-vaishnavi: Fixed

yufenglee (approved):

PR #422, opened 2024-05-08
Refacor and decluttering top level Cmakelist.txt

jchen351:


snnn (approved):

cmake/global_variables.cmake
```diff
@@ -0,0 +1,50 @@
...
+  message(FATAL_ERROR "Expected the ONNX Runtime library to be found at ${ORT_LIB_DIR}/${ONNXRUNTIME_LIB}. Actual: Not found.")
+endif()
+if(NOT EXISTS "${ORT_HEADER_DIR}/onnxruntime_c_api.h")
+  message(FATAL_ERROR "Expected the ONNX Runtime C API header to be found at \"${ORT_HEADER_DIR}/onnxruntime_c_api.h\". Actual: Not found.")
+endif()
+
+
+# Define the dependency libraries
+
+if(WIN32)
```
snnn, line 26: Why the var "ONNXRUNTIME_LIB " is used before set?

cmake/global_variables.cmake
```diff
@@ -0,0 +1,50 @@
...
+
+if(NOT EXISTS "${ORT_LIB_DIR}/${ONNXRUNTIME_LIB}")
+  message(FATAL_ERROR "Expected the ONNX Runtime library to be found at ${ORT_LIB_DIR}/${ONNXRUNTIME_LIB}. Actual: Not found.")
+endif()
+if(NOT EXISTS "${ORT_HEADER_DIR}/onnxruntime_c_api.h")
+  message(FATAL_ERROR "Expected the ONNX Runtime C API header to be found at \"${ORT_HEADER_DIR}/onnxruntime_c_api.h\". Actual: Not found.")
+endif()
+
+
+# Define the dependency libraries
```
snnn, line 24: Are these vars only used in cpack?

jchen351: No, they are also used in building

cmake/global_variables.cmake
jchen351: Nice catch. reoreded.

PR #426, opened 2024-05-09
support for ort-web

guschmue:


src/python/py/models/builder.py
```diff
@@ -1338,11 +1349,10 @@ def make_attention_mask_reformatting_for_mha(self):
         end_add_shape = ["batch_size", 1, "source_sequence_length", "target_sequence_length"]
         self.make_add(end_add_name, end_add_inputs, dtype=self.io_dtype, shape=end_add_shape) # Shape of mask is now (B, 1, S, T)
 
-        # TODO: replace Concat with Expand for performance gains
         concat_name = f"{basename}/Concat"
-        concat_inputs = [f"{end_add_name}/output_0" for _ in range(self.num_attn_heads)]
+        concat_inputs = [f"{end_add_name}/output_0", f"/model/constants/TensorProto.INT64/1D/1, {self.num_kv_heads}, 1, 1"]
```
guschmue, line 1353: ort-web/webgpu has a limit on the number of buffers and a concat with 32 inputs isn't going to work. 
Replace it with Tile.
It is enabled for all flavors since term should be better. But if we have concerns about it I can special case it for ort-web.

src/python/py/models/builder.py
```diff
@@ -1939,7 +1949,7 @@ def get_args():
         "-e",
         "--execution_provider",
         required=True,
-        choices=["cpu", "cuda", "dml"],
+        choices=["cpu", "cuda", "dml", "web"],
         help="Execution provider to target with precision of model (e.g. FP16 CUDA, INT4 CPU)",
```
yufenglee, line 1953: could you please also update the information for web here:
https://github.com/microsoft/onnxruntime-genai/pull/426/files#diff-11d14636083c4529597f09601b8f66e5df0094978debab425bfe6dd850cfcb50R1998

src/python/py/models/builder.py
yufenglee: Let me check how tile works on cpu and cuda. and not sure if dml supports Tile operator

kunal-vaishnavi:
Can you merge `main` into your branch? It looks like your version is slightly out of sync with `main` (e.g. `enable_GQA_on_CPU` is in your branch but it is not in `main` anymore).

src/python/py/models/builder.py
```diff
@@ -1338,11 +1349,10 @@ def make_attention_mask_reformatting_for_mha(self):
         end_add_shape = ["batch_size", 1, "source_sequence_length", "target_sequence_length"]
         self.make_add(end_add_name, end_add_inputs, dtype=self.io_dtype, shape=end_add_shape) # Shape of mask is now (B, 1, S, T)
 
-        # TODO: replace Concat with Expand for performance gains
         concat_name = f"{basename}/Concat"
```
kunal-vaishnavi, line 1352: If `Tile` is [good to use](https://github.com/microsoft/onnxruntime-genai/pull/426/files#r1596059754), then can this be updated to the following?

```suggestion
        tile_name = f"{basename}/Tile"
```

If you can make similar changes to replace `concat` with `tile` in the below lines, that would be helpful.

src/python/py/models/builder.py
yufenglee: ah, after reading it carefully, Tile is much better.

src/python/py/models/builder.py
guschmue: done

src/python/py/models/builder.py
guschmue: done

src/python/py/models/builder.py
```diff
@@ -1344,13 +1355,12 @@ def make_attention_mask_reformatting_for_mha(self):
         end_add_shape = ["batch_size", 1, "source_sequence_length", "target_sequence_length"]
         self.make_add(end_add_name, end_add_inputs, dtype=self.io_dtype, shape=end_add_shape) # Shape of mask is now (B, 1, S, T)
 
-        # TODO: replace Concat with Expand for performance gains
-        concat_name = f"{basename}/Concat"
-        concat_inputs = [f"{end_add_name}/output_0" for _ in range(self.num_attn_heads)]
-        concat_shape = ["batch_size", self.num_attn_heads, "source_sequence_length", "target_sequence_length"]
-        self.make_concat(concat_name, concat_inputs, dtype=self.io_dtype, shape=concat_shape, axis=1) # Shape of mask is now (B, N, S, T)
+        tile_name = f"{basename}/Tile"
+        tile_inputs = [f"{end_add_name}/output_0", f"/model/constants/TensorProto.INT64/1D/1, {self.num_kv_heads}, 1, 1"]
```
kunal-vaishnavi, line 1359:
```suggestion
        tile_inputs = [f"{end_add_name}/output_0", f"/model/constants/TensorProto.INT64/1D/1, {self.num_attn_heads}, 1, 1"]
```

guschmue:
thanks for catching!

kunal-vaishnavi (approved):

PR #428, opened 2024-05-09
AddExtraInput API for C/C++/C#

RyanUnderhill:
We added add_extra_input for python, but now we have it for the other languages

NOTE: Not final, but open to review to figure out if it's right.

Example usage:

Python:
```
        data = np.load('C:/models/qlora/adapter_weights.npz')
        for key in data.keys():
            params.add_extra_input(key, data[key])
```

C++:
```
   // Assume we have a buffer with the data to pass in
   auto tensor=OgaTensor::Create(data, std::array<int64_t>{shape1, shape2}, OgaElementType_float32);
   params.AddExtraInput("input_name", *tensor);
```
C#:
```
   // Assume data is a pointer to a block of memory to float16 values, and shape is an Int64[] array
   using var tensor = new Tensor(data, shape, ElementType.float16);
   params.AddExtraInput("input_name", tensor);
```


src/python/python.cpp
```diff
@@ -135,7 +135,7 @@ std::string ToFormatDescriptor(ONNXTensorElementDataType type) {
   }
 }
 
-std::unique_ptr<OrtValue> ToTensor(pybind11::array& v) {
+std::unique_ptr<OrtValue> ToOrtValue(pybind11::array& v) {
```
yuslepukhin, line 138: One thing we recently discovered about numpy arrays, and we are going to address it, they donot support BFloat16 and any 8-bit types. So users can not use it in Python.

src/csharp/GeneratorParams.cs
```diff
@@ -51,6 +50,11 @@ public void SetInputSequences(Sequences sequences)
             Result.VerifySuccess(NativeMethods.OgaGeneratorParamsSetInputSequences(_generatorParamsHandle, sequences.Handle));
         }
 
+        public void AddExtraInput(string name, Tensor value)
```
yuslepukhin, line 53: Same comment for naming as for C interface.

src/csharp/NativeMethods.cs
```diff
@@ -171,6 +177,25 @@ internal class NativeLib
                                                                               int /* int32_t */ token,
                                                                               out IntPtr /* const char** */ outStr);
 
+        [DllImport(NativeLib.DllName, CallingConvention = CallingConvention.Winapi)]
+        public static extern unsafe IntPtr /* OgaResult* */ OgaCreateTensorFromBuffer(IntPtr /* data* */ data, long* shape_dims, UIntPtr shape_dims_count, ElementType element_Type,
```
yuslepukhin, line 181: Replace two shape arguments to long[] and pass it directly, no pinning required.

src/csharp/NativeMethods.cs
```diff
@@ -171,6 +177,25 @@ internal class NativeLib
...
+        public static extern void OgaDestroyTensor(IntPtr /* OgaTensor * */ tensor);
+
+        [DllImport(NativeLib.DllName, CallingConvention = CallingConvention.Winapi)]
+        public static extern IntPtr /* OgaResult* */ OgaTensorGetType(IntPtr /* OgaTensor * */ tensor, out ElementType element_type);
+
+        [DllImport(NativeLib.DllName, CallingConvention = CallingConvention.Winapi)]
+        public static extern IntPtr /* OgaResult* */ OgaTensorGetShapeSize(IntPtr /* OgaTensor * */ tensor, out UIntPtr size);
+
+        [DllImport(NativeLib.DllName, CallingConvention = CallingConvention.Winapi)]
+        public static extern unsafe IntPtr /* OgaResult* */ OgaTensorGetShape(IntPtr /* OgaTensor * */ tensor, long* shape_dims, UIntPtr /* size_t */ shape_dims_count);
```
yuslepukhin, line 194: Same thing, pass a pre-allocated long[]

src/csharp/Tensor.cs
```diff
@@ -0,0 +1,80 @@
...
+    public class Tensor : IDisposable
+    {
+        private IntPtr _tensorHandle;
+        private bool _disposed = false;
+
+        public Tensor(IntPtr data, Int64[] shape, ElementType type)
+        {
+            unsafe
+            {
+                fixed (long* p = shape)
```
yuslepukhin, line 36: Not required if native delegate is changed as suggested.

src/csharp/Tensor.cs
```diff
@@ -0,0 +1,80 @@
...
+        int64,
+        string_t,
+        bool_t,
+        float16,
+        float64,
+        uint32,
+        uint64,
+    }
+
+    public class Tensor : IDisposable
```
yuslepukhin, line 27: Documentation.

src/csharp/Tensor.cs
```diff
@@ -0,0 +1,80 @@
...
+            Dispose(false);
+        }
+
+        Int64[] GetShape()
+        {
+            Result.VerifySuccess(NativeMethods.OgaTensorGetShapeSize(_tensorHandle, out UIntPtr size));
+            Int64[] shape = new Int64[size.ToUInt64()];
+
+            unsafe
+            {
```
yuslepukhin, line 55: Not required if delegate is adjusted.

yuslepukhin (changes requested):

src/ort_genai.h
```diff
@@ -236,9 +240,61 @@ struct OgaGenerator : OgaAbstract {
   static void operator delete(void* p) { OgaDestroyGenerator(reinterpret_cast<OgaGenerator*>(p)); }
 };
 
+struct OgaTensor : OgaAbstract {
+#if __cplusplus >= 202002L
+  static std::unique_ptr<OgaTensor> Create(void* data, std::span<const int64_t> shape, OgaElementType element_type) {
+    OgaTensor* p;
+    OgaCheckResult(OgaCreateTensorFromBuffer(data, shape.data(), shape.size(), element_type, &p));
+    return std::unique_ptr<OgaTensor>(p);
```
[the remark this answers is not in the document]
RyanUnderhill: This is a wrapper around the C API, so it can't call make_unique as there's no C++ object to be constructed. We just get back the pointer from the C API.

src/csharp/NativeMethods.cs
RyanUnderhill: How do I change it? C# knows how to split the long[] into the separate parts for a C API?

yuslepukhin: Keep both arguments but replace the first one with long[]. Remove unsafe keyword.

src/tensor.h
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+namespace Generators {
+
+struct Tensor : std::enable_shared_from_this<Tensor> {
+
+  Tensor() = default;
+  Tensor(std::unique_ptr<OrtValue> ort_tensor) : ort_tensor_ {std::move(ort_tensor)} { }
+
+  std::unique_ptr<OrtValue> ort_tensor_;
```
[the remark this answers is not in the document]
RyanUnderhill: We get an OrtValue back from the ort C API, it's an opaque type to our code. We store it in a std::unique_ptr to ensure proper C++ RAII (the unique_ptr destrustor calls OrtDestroyValue automatically)

src/tensor.h
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+namespace Generators {
+
+struct Tensor : std::enable_shared_from_this<Tensor> {
```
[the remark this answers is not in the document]
RyanUnderhill: Yes, we only need help in the C# files, the rest are to get a demo working for build!

src/ort_genai_c.h
```diff
@@ -155,8 +173,9 @@ OGA_EXPORT OgaResult* OGA_API_CALL OgaGeneratorParamsSetInputIDs(OgaGeneratorPar
  */
 OGA_EXPORT OgaResult* OGA_API_CALL OgaGeneratorParamsSetInputSequences(OgaGeneratorParams* generator_params, const OgaSequences* sequences);
 
-OGA_EXPORT OgaResult* OGA_API_CALL OgaGeneratorParamsSetWhisperInputFeatures(OgaGeneratorParams*, const int32_t* inputs, size_t count);
-OGA_EXPORT OgaResult* OGA_API_CALL OgaGeneratorParamsSetWhisperDecoderInputIDs(OgaGeneratorParams*, const int32_t* input_ids, size_t input_ids_count);
+OGA_EXPORT OgaResult* OGA_API_CALL OgaGeneratorParamsAddExtraInput(OgaGeneratorParams* generator_params, const char* name, OgaTensor* tensor);
```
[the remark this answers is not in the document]
RyanUnderhill: Amusingly it was called 'SetInput' initially but that confused people, because it sounds like there's only one and it's special. So it was renamed to 'AddExtraInput' to let you pass in extra data for LoRA models.

Documentation will solve it going forward.

src/ort_genai_c.h
```diff
@@ -229,6 +248,13 @@ OGA_EXPORT void OGA_API_CALL OgaDestroyTokenizerStream(OgaTokenizerStream*);
  */
 OGA_EXPORT OgaResult* OGA_API_CALL OgaTokenizerStreamDecode(OgaTokenizerStream*, int32_t token, const char** out);
 
+OGA_EXPORT OgaResult* OGA_API_CALL OgaCreateTensorFromBuffer(void* data, const int64_t* shape_dims, size_t shape_dims_count, OgaElementType element_type, OgaTensor** out);
+OGA_EXPORT void OGA_API_CALL OgaDestroyTensor(OgaTensor* tensor);
+OGA_EXPORT OgaResult* OGA_API_CALL OgaTensorGetType(OgaTensor*, OgaElementType* out);
```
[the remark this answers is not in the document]
RyanUnderhill: :) I specifically avoided having to pass in the data size for that reason. I internally calculate it based on the shape before calling into Ort.

src/ort_genai_c.h
```diff
@@ -155,8 +173,9 @@ OGA_EXPORT OgaResult* OGA_API_CALL OgaGeneratorParamsSetInputIDs(OgaGeneratorPar
  */
 OGA_EXPORT OgaResult* OGA_API_CALL OgaGeneratorParamsSetInputSequences(OgaGeneratorParams* generator_params, const OgaSequences* sequences);
 
-OGA_EXPORT OgaResult* OGA_API_CALL OgaGeneratorParamsSetWhisperInputFeatures(OgaGeneratorParams*, const int32_t* inputs, size_t count);
-OGA_EXPORT OgaResult* OGA_API_CALL OgaGeneratorParamsSetWhisperDecoderInputIDs(OgaGeneratorParams*, const int32_t* input_ids, size_t input_ids_count);
+OGA_EXPORT OgaResult* OGA_API_CALL OgaGeneratorParamsAddExtraInput(OgaGeneratorParams* generator_params, const char* name, OgaTensor* tensor);
+
+OGA_EXPORT OgaResult* OGA_API_CALL OgaGeneratorParamsSetWhisperInputFeatures(OgaGeneratorParams*, OgaTensor* tensor);
```
yufenglee, line 178: not to this PR. Why whisper has special public API?

src/ort_genai_c.h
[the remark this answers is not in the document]
yufenglee: as Dmitri suggested, could you please add some comments here to explain the ShapeSize

src/tensor.h
```diff
@@ -0,0 +1,13 @@
...
+  Tensor() = default;
+  Tensor(std::unique_ptr<OrtValue> ort_tensor) : ort_tensor_{std::move(ort_tensor)} {}
```
yufenglee, line 7: >std::unique_ptr<OrtValue> [](http://example.com/codeflow?start=9&length=25)

use std::unique_ptr<OrtValue> && to specify the ort_tensor object will be moved

yufenglee:
could you please add unit test to cover C/C++, C# API?

yufenglee:
and also an example to use C/C++

---
In reply to: [2103689004](https://github.com/microsoft/onnxruntime-genai/pull/428#issuecomment-2103689004) [](http://example.com/codeflow?ancestors=2103689004)

src/ort_genai_c.h
```diff
@@ -229,6 +248,13 @@ OGA_EXPORT void OGA_API_CALL OgaDestroyTokenizerStream(OgaTokenizerStream*);
  */
 OGA_EXPORT OgaResult* OGA_API_CALL OgaTokenizerStreamDecode(OgaTokenizerStream*, int32_t token, const char** out);
 
+OGA_EXPORT OgaResult* OGA_API_CALL OgaCreateTensorFromBuffer(void* data, const int64_t* shape_dims, size_t shape_dims_count, OgaElementType element_type, OgaTensor** out);
```
yufenglee, line 251: Could you please add another one to create without data managed by Oga

RyanUnderhill: This is the API for that. Oga doesn't own the memory. I'll add documentation.

src/tensor.h
RyanUnderhill: Amusingly it's recommended to pass by value to enforce that it's taking ownership, as it can only be passed by moving it into the parameter you pass. Vs rvalue reference "might" take ownership.

https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#r32-take-a-unique_ptrwidget-parameter-to-express-that-a-function-assumes-ownership-of-a-widget

R.32: Take a unique_ptr<widget> parameter to express that a function assumes ownership of a widget

Reason Using unique_ptr in this way both documents and enforces the function call’s ownership transfer.

src/ort_genai_c.h
RyanUnderhill: Whisper took extra inputs that we manage, this PR just hooked up the function that was already there in a better way.

src/ort_genai_c.h
yufenglee: sorry, i typed it wrongly. I meant to add one to create another with data managed by oga

src/csharp/GeneratorParams.cs
```diff
@@ -37,20 +36,19 @@ public void TryGraphCaptureWithMaxBatchSize(int maxBatchSize)
...
         public void SetInputIDs(ReadOnlySpan<int> inputIDs, ulong sequenceLength, ulong batchSize)
         {
-            unsafe
-            {
-                fixed (int* inputIDsPtr = inputIDs)
-                {
-                    Result.VerifySuccess(NativeMethods.OgaGeneratorParamsSetInputIDs(_generatorParamsHandle, inputIDsPtr, (UIntPtr)inputIDs.Length, (UIntPtr)sequenceLength, (UIntPtr)batchSize));
-                }
-            }
+            Result.VerifySuccess(NativeMethods.OgaGeneratorParamsSetInputIDs(_generatorParamsHandle, inputIDs.ToArray(), (UIntPtr)inputIDs.Length, (UIntPtr)sequenceLength, (UIntPtr)batchSize));
```
baijumeswani, line 39: I think we need not do this. `inputIDs.ToArray()` will make a copy of the underlying array if I remember correctly. Instead, we could just send in a pointer to the underlying memory buffer using the previous implementation.
For this function, maybe we can leave it unchanged.

src/csharp/NativeMethods.cs
```diff
@@ -59,17 +59,23 @@ internal class NativeLib
 
         // This function is used to set the input IDs for the generator.
         [DllImport(NativeLib.DllName, CallingConvention = CallingConvention.Winapi)]
-        public static extern unsafe IntPtr /* OgaResult* */ OgaGeneratorParamsSetInputIDs(IntPtr /* OgaGeneratorParams* */ generatorParams,
-                                                                                          int* /* const int32_t* */ inputIDs,
-                                                                                          UIntPtr /* size_t */ inputIDsCount,
-                                                                                          UIntPtr /* size_t */ sequenceLength,
-                                                                                          UIntPtr /* size_t */ batchSize);
+        public static extern IntPtr /* OgaResult* */ OgaGeneratorParamsSetInputIDs(IntPtr /* OgaGeneratorParams* */ generatorParams,
```
baijumeswani, line 62: Same comment as above. We can probably avoid making the change for this function.

src/csharp/Tokenizer.cs
```diff
@@ -65,13 +65,7 @@ public Sequences Encode(string str)
...
         {
             IntPtr outStr = IntPtr.Zero;
-            unsafe
-            {
-                fixed (int* sequencePtr = sequence)
-                {
-                    Result.VerifySuccess(NativeMethods.OgaTokenizerDecode(_tokenizerHandle, sequencePtr, (UIntPtr)sequence.Length, out outStr));
-                }
-            }
+            Result.VerifySuccess(NativeMethods.OgaTokenizerDecode(_tokenizerHandle, sequence.ToArray(), (UIntPtr)sequence.Length, out outStr));
```
baijumeswani, line 68: Same comment as setinputids. For spans, let's not change the implementation.

src/ort_genai.h
```diff
@@ -236,9 +240,61 @@ struct OgaGenerator : OgaAbstract {
...
+  }
+#endif
+  static std::unique_ptr<OgaTensor> Create(void* data, const int64_t* shape_dims, size_t shape_dims_count, OgaElementType element_type) {
+    OgaTensor* p;
+    OgaCheckResult(OgaCreateTensorFromBuffer(data, shape_dims, shape_dims_count, element_type, &p));
+    return std::unique_ptr<OgaTensor>(p);
+  }
+
+  OgaElementType GetType() {
```
baijumeswani, line 257: How about naming it `Type()` instead of `GetType()`? Same for `Shape()` and `Data()`.

src/ort_genai.h
```diff
@@ -236,9 +240,61 @@ struct OgaGenerator : OgaAbstract {
...
+
 struct OgaHandle {
   OgaHandle() = default;
   ~OgaHandle() noexcept {
     OgaShutdown();
   }
 };
+
+// Global Oga functions
+namespace Oga {
```
baijumeswani, line 288: 👍🏽

src/ort_genai_c.cpp
```diff
@@ -122,7 +122,7 @@ OgaResult* OGA_API_CALL OgaGeneratorParamsSetInputIDs(OgaGeneratorParams* oga_pa
   OGA_CATCH
 }
 
-OGA_EXPORT OgaResult* OGA_API_CALL OgaGeneratorParamsSetInputSequences(OgaGeneratorParams* oga_params, const OgaSequences* p_sequences) {
+OgaResult* OGA_API_CALL OgaGeneratorParamsSetInputSequences(OgaGeneratorParams* oga_params, const OgaSequences* p_sequences) {
```
baijumeswani, line 125: 👍🏽

src/csharp/GeneratorParams.cs
RyanUnderhill: 👍Reverted

src/csharp/NativeMethods.cs
RyanUnderhill: 👍Reverted

src/csharp/Tokenizer.cs
RyanUnderhill: 👍Reverted

src/tensor.h
yufenglee: I prefer to use && to keep consistent.

yufenglee (approved):
:shipit:

src/csharp/GeneratorParams.cs
yuslepukhin: Does C API expect a zero terminated string and does ToUtf8 append zero terminator?

src/csharp/NativeMethods.cs
```diff
@@ -171,6 +177,25 @@ internal class NativeLib
...
+        public static extern IntPtr /* OgaResult* */ OgaTensorGetType(IntPtr /* OgaTensor * */ tensor, out ElementType element_type);
+
+        [DllImport(NativeLib.DllName, CallingConvention = CallingConvention.Winapi)]
+        public static extern IntPtr /* OgaResult* */ OgaTensorGetShapeSize(IntPtr /* OgaTensor * */ tensor, out UIntPtr size);
+
+        [DllImport(NativeLib.DllName, CallingConvention = CallingConvention.Winapi)]
+        public static extern IntPtr /* OgaResult* */ OgaTensorGetShape(IntPtr /* OgaTensor * */ tensor, long[] shape_dims, UIntPtr /* size_t */ shape_dims_count);
+
+        [DllImport(NativeLib.DllName, CallingConvention = CallingConvention.Winapi)]
+        public static extern IntPtr /* OgaResult* */ OgaTensorGetData(IntPtr /* OgaTensor * */ tensor, out IntPtr /* void* */ data);
```
yuslepukhin, line 197: How does the caller know the length of the data without multiplying the shape?
Do you plan to expose it via Tensor class?

src/csharp/NativeMethods.cs
```diff
@@ -171,6 +177,25 @@ internal class NativeLib
...
+                                                                               out IntPtr /* OgaTensor** */ tensor);
+
+        [DllImport(NativeLib.DllName, CallingConvention = CallingConvention.Winapi)]
+        public static extern void OgaDestroyTensor(IntPtr /* OgaTensor * */ tensor);
+
+        [DllImport(NativeLib.DllName, CallingConvention = CallingConvention.Winapi)]
+        public static extern IntPtr /* OgaResult* */ OgaTensorGetType(IntPtr /* OgaTensor * */ tensor, out ElementType element_type);
+
+        [DllImport(NativeLib.DllName, CallingConvention = CallingConvention.Winapi)]
+        public static extern IntPtr /* OgaResult* */ OgaTensorGetShapeSize(IntPtr /* OgaTensor * */ tensor, out UIntPtr size);
```
yuslepukhin, line 191: Suggest GetShapeRank() instead

src/csharp/Tensor.cs
```diff
@@ -0,0 +1,67 @@
...
+            Result.VerifySuccess(NativeMethods.OgaCreateTensorFromBuffer(data, shape, (UIntPtr) shape.Length, type, out _tensorHandle));
+        }
+        internal IntPtr Handle { get { return _tensorHandle; } }
+
+        ~Tensor()
+        {
+            Dispose(false);
+        }
+
+        Int64[] GetShape()
```
yuslepukhin, line 43: Shape, but no data?

yuslepukhin (changes requested):

src/csharp/Tensor.cs
RyanUnderhill: Good point, easy enough to add. We just didn't have a need for them yet so I implemented one as a test.

src/csharp/NativeMethods.cs
RyanUnderhill: We could certainly add it eventually. The current APIs were to allow people to pass data in, this could be added in the future as needed.

src/ort_genai_c.h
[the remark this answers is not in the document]
RyanUnderhill: Done, I documented all of the C API functions I added.

src/csharp/Tensor.cs
RyanUnderhill: Oops, wasn't even public. The unit test caught that.

baijumeswani: While you're at it, could we also call this function Shape() and Data() and so on?

src/csharp/Tensor.cs
RyanUnderhill: We plan on having it here once it stabilizes: https://onnxruntime.ai/docs/genai/api/csharp.html

baijumeswani (approved):

src/csharp/Tensor.cs
RyanUnderhill: Yep! Was locally changed, now it's pushed.

src/python/python.cpp
RyanUnderhill: Interesting, I messaged you offline about how another team sent us a PR that added 8-bit types. Not BFloat though.

src/tensor.h
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+namespace Generators {
```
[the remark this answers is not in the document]
RyanUnderhill: What header is needed? This is included by our main generators.h header.

src/python/python.cpp
RyanUnderhill: Ah, 8-bit floats. It all makes sense now. Yes, we haven't needed those yet. As the data is coming from numpy it should be ok.

yuslepukhin (approved):

PR #445, opened 2024-05-13
Test fix - ensure random indices don't overlap

RyanUnderhill:
CreateRandomLogits might hit the same index twice in the random sampling, this can result in num_large not being hit and resulting in fewer large values in the logits. This will make our unit test fail randomly.

For example, we have a TopK(10) test, and we randomly set 10 score values to be 25 (the rest between 0-1). We test it by setting 10 random values to be 25, and verify the TopK results are all '25'.

But, if a duplicate random index occurs, we'll only have 9 (or fewer) score values as 25, and our Top 10 will include some lower values.

### DIFF
--- a/test/sampling_tests.cpp
+++ b/test/sampling_tests.cpp
@@ -26,8 +26,8 @@ TEST(SamplingTests, BatchedSamplingTopPCpu) {
   int batch_size = 4;
   auto params = Generators::CreateGeneratorParams();
   params->search.max_length = 10;
-  params->search.do_sample=true;
-  params->search.top_p=0.25f;
+  params->search.do_sample = true;
+  params->search.top_p = 0.25f;
   params->batch_size = batch_size;
   params->sequence_length = 1;
   params->vocab_size = vocab_size;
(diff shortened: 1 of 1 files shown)


test/sampling_tests.cpp
```diff
@@ -116,12 +116,16 @@ void CreateRandomLogits(float* logits, int num_large, int vocab_size, int batch_
...
+
   // Randomly set num_large elements to be large
   std::uniform_int_distribution<> dist_large(0, vocab_size - 1);
   for (int b = 0; b < batch_size; b++) {
     for (int i = 0; i < num_large; i++) {
-      int index = dist_large(engine);
-      logits[index + b * vocab_size] = 25.0f;
+      float& value = logits[dist_large(engine) + b * vocab_size];
+      if (value == 25.0f)
+        i--;  // We hit the same number twice, so do it again to ensure num_large values are set to 25.0f
```
baijumeswani, line 126: Can we have a sanity check to make sure that `num_large` < logits.size(). Else, we will run into an infinite loop.

aciddelgado: could also increment index until reaching an unused one

baijumeswani (approved):

PR #447, opened 2024-05-13
Moving ORT-Genai DLL into lib folder

jchen351:
Moving ORT-Genai Windows dll into lib folder

### DIFF
--- a/cmake/package.cmake
+++ b/cmake/package.cmake
@@ -5,6 +5,7 @@ set_target_properties(
 install(TARGETS
   onnxruntime-genai
   LIBRARY DESTINATION lib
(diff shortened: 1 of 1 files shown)


snnn (approved):

cmake/package.cmake
```diff
@@ -5,6 +5,7 @@ set_target_properties(
 install(TARGETS
   onnxruntime-genai
   LIBRARY DESTINATION lib
+  RUNTIME DESTINATION lib
```
snnn, line 8: Does this change have any impact on Linux?

jchen351: No, we don't currently have any executable.

PR #461, opened 2024-05-15
Initial Java bindings. 

skottmckay:
Java bindings for GenAI C API, as well as a SimpleGenAI class for simplified basic usage (text in, text out, single request with optional callback for token-by-token output).

Builds and the unit tests pass on Windows.

The Android setup has been copied from ORT but is currently untested so can be ignored for now.


src/java/src/main/native/generator_params.cpp
```diff
@@ -0,0 +1,63 @@
...
+  ThrowIfError(env, OgaGeneratorParamsSetInputSequences(generator_params, sequences));
+}
+
+extern "C" JNIEXPORT void JNICALL
+Java_ai_onnxruntime_1genai_GeneratorParams_setInputIDs(JNIEnv* env, jobject thiz, jlong native_handle,
+                                                       jintArray tokenIds, jint sequenceLength, jint batchSize) {
+  OgaGeneratorParams* generator_params = reinterpret_cast<OgaGeneratorParams*>(native_handle);
+
+  auto num_tokens = env->GetArrayLength(tokenIds);
+  jint* jtokens = env->GetIntArrayElements(tokenIds, nullptr);
```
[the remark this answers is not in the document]
skottmckay: I added the release but it turns out the GenAI C code doesn't copy the array, so the release made generation fail due to invalid inputs.

What's the best approach to this? I assume I need to add an `int[]` member to the GeneratorParam java class to keep the array alive at that level, but not clear how to handle at the JNI level. IIUC GetIntArrayElements doesn't control whether a copy of the array is made or not.

src/java/Debugging.md
```diff
@@ -0,0 +1,96 @@
...
+
+Build the native library and ensure it's in the native-jni path mentioned above when you run the tests.
+That should result in an error window with Debug/Retry/Ignore buttons.
+When that error window appears, you first need to attach to the Java process that is running the tests. 
+Process Explorer (from sysinternals) is a good way to find the process as you can use it's 'target' feature to select the error window.
+Attach to that process in Visual Studio, ideally using the solution for the native library you built with the heap violation.
+
+Once you have attached, hit the 'Retry' button in the error window. That _should_ drop you into VS with full debug info and call stack in the genai code. Note that you may need to change to the current thread to see the call stack as it may be in the Java thread that is displaying the dialog box.
+
+If someone knows the magic incantation to run java.exe directly with a unit test (with class paths etc. set up correctly) that would be a much better way to debug the native code.
```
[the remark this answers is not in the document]
skottmckay: I got it to work a bit more gracefully by updating the build to put all the native libs in one place (I couldn't see anywhere that required native-lib and native-jib), and passed that path through to build.gradle in `-DnativeLibPath=...`. That at least overcomes the need to manually set java.library.path on Windows and hopefully works on other platforms.

With that, this command line will find the tests (needed a newer junit version to handle EnabledIf) and run them:

> D:\Java\jdk-11.0.17\bin\java.exe "-Djava.library.path=D:\src\github\ort.genai\build\Windows\Debug\src\java\native-lib\ai\onnxruntime-genai\native\win-x64" -jar junit-platform-console-standalone-1.10.2.jar -cp D:\src\github\ort.genai\src\java\build\classes\java\test -cp D:\src\github\ort.genai\src\java\build\resources\test  -cp D:\src\github\ort.genai\src\java\build\classes\java\main --scan-classpath

src/java/src/main/java/ai/onnxruntime/genai/SimpleGenAI.java
```diff
@@ -0,0 +1,132 @@
...
+   * @param listener Optional callback for tokens to be provided as they are generated. NOTE: Token
+   *     generation will be blocked until the listener's `accept` method returns.
+   * @return The generated text.
+   * @throws GenAIException on failure
+   */
+  public String generate(GeneratorParams generatorParams, Consumer<String> listener)
+      throws GenAIException {
+    String result = null;
+    try (Tokenizer tokenizer = new Tokenizer(model)) {
+      int[] output_ids = null;
```
Craigacp, line 99: Both result and output_ids don't need to be set to null, the Java compiler knows they are assigned if the method terminates normally.

src/java/src/main/native/generator_params.cpp
```diff
@@ -0,0 +1,77 @@
...
+extern "C" JNIEXPORT void JNICALL
+Java_ai_onnxruntime_genai_GeneratorParams_setInputIDs(JNIEnv* env, jobject thiz, jlong native_handle,
+                                                       jintArray token_ids, jint sequence_length, jint batch_size) {
+  OgaGeneratorParams* generator_params = reinterpret_cast<OgaGeneratorParams*>(native_handle);
+
+  auto num_tokens = env->GetArrayLength(token_ids);
+  jboolean is_copy = false;
+  jint* jtokens = env->GetIntArrayElements(token_ids, &is_copy);
+  if (is_copy) {
+    // we're dead as GenAI doesn't copy the inputs so the input data address will be invalid
```
Craigacp, line 64: If it doesn't copy the memory then the Java side method should create a direct byte buffer, map it into an int buffer, write the array to it, store it in a field in the GeneratorParams and then pass the buffer down into JNI. The JNI code can then pull out the pointer from the buffer and hand it to the native code and Java will guarantee that the buffer lives as long as the Java GeneratorParams object does. This byte buffer dance is how ORT's tensors work, that's how we get zero copy through from Java into ORT. It'll look a bit like this - https://github.com/microsoft/onnxruntime/blob/main/java/src/main/java/ai/onnxruntime/OrtUtil.java#L507, and then on the JNI side - https://github.com/microsoft/onnxruntime/blob/main/java/src/main/native/ai_onnxruntime_OnnxTensor.c#L98.

What happens if this method is called multiple times is then something we need to decide the semantics for.

src/java/src/main/native/generator_params.cpp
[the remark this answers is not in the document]
Craigacp: I added a comment here - https://github.com/microsoft/onnxruntime-genai/pull/461#discussion_r1603419301

cmake/global_variables.cmake
```diff
@@ -53,4 +55,53 @@ if(NOT EXISTS "${ORT_LIB_DIR}/${ONNXRUNTIME_LIB}")
...
+  elseif (genai_target_platform STREQUAL "x64" OR 
+          genai_target_platform STREQUAL "x86_64" OR 
+          genai_target_platform STREQUAL "AMD64" OR 
+          CMAKE_GENERATOR MATCHES "Win64")
+    set(genai_target_platform "x64")
+  else()
+    message(FATAL_ERROR "Unsupported architecture. CMAKE_SYSTEM_PROCESSOR: ${CMAKE_SYSTEM_PROCESSOR}")
+  endif()
+elseif(APPLE)
+  if (CMAKE_OSX_ARCHITECTURES STREQUAL "arm64")
```
Craigacp, line 81: This seems to be unset when I try to build on a Mac. Is there a way of passing it through?

Craigacp (approved):
Looks good. I still need to make build.sh executable to get the build to run on a clean checkout, and the test generation is still locked off to Windows, but otherwise everything passes cleanly on Linux x64 when I checked it. MacOS will be fine when the environment detection is fixed.

src/java/build.gradle
```diff
@@ -0,0 +1,249 @@
...
+	}
+}
+
+if (cmakeNativeLibDir != null) {
+	// generate tasks to be called from cmake
+
+	// Overwrite jar location
+	task allJar(type: Jar) {
+		manifest {
+			attributes('Automatic-Module-Name': project.group,
```
Craigacp, line 122: I'd missed this initially & checked again after noticing a similar issue in ORT-extensions (https://github.com/microsoft/onnxruntime-extensions/pull/730), but `com.microsoft.onnxruntime-genai` isn't a valid Java module name as it has a hypen in. Could this be changed to `com.microsoft.onnxruntime.genai`?

src/java/src/main/java/ai/onnxruntime/genai/Model.java
```diff
@@ -0,0 +1,92 @@
...
+  // model.createTokenizer is consistent with the python setup at least and agreed upon.
+
+  /**
+   * Creates a GeneratorParams instance for executing the model. NOTE: GeneratorParams internally
+   * uses the Model, so the Model instance must remain valid
+   *
+   * @return The GeneratorParams instance.
+   * @throws GenAIException If the call to the GenAI native API fails.
+   */
+  public GeneratorParams createGeneratorParams() throws GenAIException {
```
skottmckay, line 38: I have both the tokenizer and the generator params creation being done via the Model class currently. The rationale is that creating those classes involves Model internals (the config), hence we pass the model to the ctor of each. It's also more consistent this way vs creating the tokenizer via the model but the generator params directly. 

Happy to discuss further and adjust the other language bindings to be consistent.

baijumeswani: I think this is ok for now. We need to standardize the API. This can happen as a separate PR.

src/java/src/test/android/.gitignore
baijumeswani, line 1: Is the plan here to run the android app and tests here in CI?

baijumeswani (approved):

src/java/src/test/android/.gitignore
skottmckay: Yep. Have that working in next branch. ORT does the same. there's a gradle target to copy the test app to the emulator and run it.

src/java/Debugging.md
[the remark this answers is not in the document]
skottmckay: Took me long enough to figure out and nobody else should have to waste that much time in the future. :)

src/java/src/main/java/ai/onnxruntime_genai/Generator.java
```diff
@@ -0,0 +1,141 @@
...
+ * <p>The expected usage is to loop until isDone returns false. Within the loop, call computeLogits
+ * followed by generateNextToken.
+ *
+ * <p>The newly generated token can be retrieved with getLastTokenInSequence and decoded with
+ * TokenizerStream.Decode.
+ *
+ * <p>After the generation process is done, GetSequence can be used to retrieve the complete
+ * generated sequence if needed.
+ */
+public class Generator implements AutoCloseable {
```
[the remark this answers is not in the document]
skottmckay: Will look at doing this in the next PR.

PR #464, opened 2024-05-16
Support extra inputs in graph captures

PatriceVignola:
Since extra inputs can change between generators, we need a way to make sure that a captured graph can be reused. For a graph to be reused, the following conditions need to be met:

1. All extra inputs need to have the same names as the extra inputs that were provided when first recording the graph
2. All shapes need to match
3. All data types need to match
4. The number of extra inputs needs to be the same

As long as those conditions are satisfied, we are allowed to reuse the graph. Then, at model creation time, we need to copy the extra inputs to the gpu-backed static buffers that were created when recording the graph.


src/models/captured_graph_pool.h
```diff
@@ -5,20 +5,92 @@
...
 // From boost http://www.boost.org/doc/libs/1_35_0/doc/html/hash/combine.html
 template <class T>
 inline void hash_combine(size_t& seed, T const& v) {
   seed ^= std::hash<T>()(v) + 0x9e3779b9 + (seed << 6) + (seed >> 2);
 }
 
-struct tuple_hash {
-  size_t operator()(const std::tuple<int, int, int>& p) const {
+struct InputKey {
+  InputKey(std::string name, ONNXTensorElementDataType tensor_type, std::vector<int64_t> tensor_shape)
```
yufenglee, line 17: >name [](http://example.com/codeflow?start=23&length=4)

InputKey

yufenglee (approved):
:shipit:

src/models/model.cpp
```diff
@@ -36,13 +36,7 @@ static std::wstring CurrentModulePath() {
 
 namespace Generators {
 
-State::State(const GeneratorParams& params) : params_{params.shared_from_this()} {
-  // Add extra user inputs
-  for (auto& input : params.extra_inputs) {
-    input_names_.push_back(input.name.c_str());
-    inputs_.push_back(input.tensor->ort_tensor_.get());
-  }
-}
```
baijumeswani, line 45: Does this change impact other models (whisper) which might depend on extra inputs?
It seems like now every type of Model must have this extra input defined in their class? Can this input be constructed on the parent class instead of constructing them for all the children classes?

PatriceVignola: I now added it for GPT and Whisper, but you're right it would be better if it was constructed in the `State` class. It would require some refactoring though since the `State` class doesn't own a captured graph or doesn't even have a reference on the model, because GPT and Whisper don't support CUDA graph. I'd rather not mess with that flow at the moment, especially since I don't have a good way to teste GPT and Whisper (they don't work on DML).

PR #467, opened 2024-05-17
Add custom DML allocators and non-global env to work around memory issues

PatriceVignola:
This change achieves 2 things:

1. Use non-global envs for DML only. It's also possible to enable it for CUDA, but I wanted to avoid breaking things. I saw comments in the code that indicated that CUDA somehow needs a global environment that outlives the model, but it doesn't seem to be necessary for the DML EP. Global envs tend to hold references on allocators, which makes it hard to clean all memory when a model gets released. By having the lifetime of the env be tied to the lifetime of the model, we avoid such problems.
2. Use a custom allocator for DML. `onnxrutime-genai-directml` is a very specific scenario where we know in advance where all memory will be coming from, so we don't need a fancy arena allocator to manage the memory. In fact, having the memory as tight as possible allows models to run on more hardware and reduces fragmentation.


pranavsharma:
> Global envs tend to hold references on allocators

By default, the environment doesn't hold references to any allocators. Can you give an example?
In general, we want the OrtEnv instance to be created only once per process. It's actually a singleton so calling Create again will return the same instance.

PatriceVignola:
You're right, it looks like the implementation itself is a singleton. I was looking at the documentation [over here](https://skottmckay.github.io/onnxruntime/docs/reference/api/c-api.html) which says `Create and register a shared allocator with the env using the CreateAndRegisterAllocator API. This allocator is then reused by all sessions that use the same env instance unless a session`. The `that use the same env instance` seems to imply that there can be more than one env instance.

It turns out that the fix was instead in another change I had made, which I didn't realize was holding a hard reference on the allocator.

yufenglee (approved):
:shipit:

pranavsharma:
The PR's title and desc don't match what has been implemented. This is quite misleading when someone goes through the commit logs.

PR #468, opened 2024-05-17
Drop std::filesystem for custom file path handling

skyline75489:


src/file_path.h
```diff
@@ -0,0 +1,115 @@
...
+
+// Windows Header Files:
+#include <windows.h>
+#define ENABLE_INTSAFE_SIGNED_FUNCTIONS // Only unsigned intsafe math/casts available without this def
+#include <intsafe.h>
+#include <tchar.h>
+
+using path_type = std::wstring;
+
+[[nodiscard]] inline std::wstring ConvertToW(const UINT codePage, const std::string_view source) {
```
skyline75489, line 23: Copied from https://github.com/microsoft/terminal/blob/3486111722296f287158e0340789c607642c1067/src/types/convert.cpp#L20  . Dropped all error handling because we don't have WIL included.

src/models/model.cpp
```diff
@@ -122,7 +122,7 @@ const std::string& TokenizerStream::Decode(int32_t token) {
 }
 
 Tokenizer::Tokenizer(Config& config) : pad_token_id_{config.model.pad_token_id} {
-  CheckResult(OrtxCreateTokenizer(tokenizer_.Address(), reinterpret_cast<const char*>(config.config_path.u8string().c_str())));
+  CheckResult(OrtxCreateTokenizer(tokenizer_.Address(), reinterpret_cast<const char*>(config.config_path.c_str())));
```
skyline75489, line 125: Needs to sync with Wenbing about the API boundary.

skyline75489 (commented):
Currently this PR expects everything users provide to be UTF-8 encoded. However, in older version of Windows, the system codepage can vary based on the locale. For example, the Simplified Chinese version of Windows is likely to use GB18030. Handling this scenario requires more work.

RyanUnderhill:
What problem is this solving? We wanted to avoid having an API for Windows and an API for all other OS's, and std::filesystem let us use utf8 on all OS's. We shouldn't be using the system code page, only utf8.

src/config.cpp
```diff
@@ -63,7 +63,12 @@ struct SessionOptions_Element : JSON::Element {
     if (name == "log_id")
       v_.log_id = value;
     else if (name == "enable_profiling")
+#ifdef _WIN32
```
RyanUnderhill, line 66: #ifdef _WIN32 everywhere is what we wanted to avoid with std::filesystem. What problem is this change solving?

RyanUnderhill (dismissed):

src/config.cpp
RyanUnderhill: Saw the other comment, hopefully you can put the #ifdef _WIN32 inside the "oga_std_filesystem" replacement so that we can switch to it once the compilers sort themselves out.

skyline75489:
Does [this one](https://github.com/microsoft/onnxruntime-extensions/commit/c3c5f1cbb135fc9ccc673ae99b31e4ce6a9e2669) from @wenbingl looks fine to you? If so, I can follow the same fashion

skyline75489:
Closed in favor of #480

PR #480, opened 2024-05-21
Remove std::filesystem for the compatibility of old systems

skyline75489:
From https://github.com/microsoft/onnxruntime-extensions/pull/721

### DIFF
--- a/src/models/model.cpp
+++ b/src/models/model.cpp
@@ -121,7 +121,7 @@ const std::string& TokenizerStream::Decode(int32_t token) {
 }
 
 Tokenizer::Tokenizer(Config& config) : pad_token_id_{config.model.pad_token_id} {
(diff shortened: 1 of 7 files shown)


src/models/model.cpp
```diff
@@ -121,7 +121,7 @@ const std::string& TokenizerStream::Decode(int32_t token) {
 }
 
 Tokenizer::Tokenizer(Config& config) : pad_token_id_{config.model.pad_token_id} {
-  CheckResult(OrtxCreateTokenizer(tokenizer_.Address(), reinterpret_cast<const char*>(config.config_path.u8string().c_str())));
+  CheckResult(OrtxCreateTokenizer(tokenizer_.Address(), config.config_path.string().c_str()));
```
baijumeswani, line 124: can we just do `.c_str()`?

baijumeswani (approved):

src/models/model.cpp
natke: Cannot use .c_str() because it returns wide string

PR #491, opened 2024-05-21
Single quotes around .../* command

simonw:
Without single quotes this throws an error on Zsh.


baijumeswani (approved):

natke:
Does not work with bash. Closing PR

PR #498, opened 2024-05-22
Add Azure pipeline script to release onnxruntime-genai pkg for Windows ARM64 CPU

gunandrose4u:
Add Azure pipeline script to release onnxruntime-genai pkg for Windows ARM64 CPU

- Release onnxruntime-genai python wheel for python 3.11
- Release onnxruntime-genai python wheel for python 3.12
- Release onnxruntime-genai nuget pkg


.pipelines/stages/jobs/nuget-packaging-job.yml
```diff
@@ -79,14 +81,40 @@ jobs:
...
   workspace:
     clean: all
   steps:
-  - task: UsePythonVersion@0
-    inputs:
-      versionSpec: 3.10
-      addToPath: true
-      architecture: $(arch)
+  - script: |
+      MKDIR $(Agent.ToolsDirectory)\Python\3.11.0\arm64
```
snnn, line 96: Could you please put them in the VM's image?

.pipelines/stages/jobs/steps/nuget-win-step.yml
```diff
@@ -15,6 +19,18 @@ steps:
     FolderPath: '$(Build.Repository.LocalPath)\src\csharp\bin\Release\'
     DisplayName: 'ESRP - Sign C# dlls'
     Pattern: '*OnnxRuntimeGenAI*.dll'
+- powershell: |
```
snnn, line 22: Put it in the image.
Or use "dotnet" command instead.

.pipelines/stages/jobs/nuget-packaging-job.yml
gunandrose4u: Yeah, will move this part in image later

.pipelines/stages/jobs/steps/nuget-win-step.yml
gunandrose4u: Will install nuget in image in order to keep consistence with other nuget pkg release script.

.pipelines/stages/jobs/nuget-packaging-job.yml
gunandrose4u: Base  on  Adrian's suggestion, it would be better to create a new image and new agent pool for new project. I would like to know whether ort-genai project can reuse the pool and image which are used by ORT build and release. If yes, then I think just modify the image config and rebuild image is enough. If not, then it will take more effort and time  to create new image and pool.

mszhanyi: personally, I wish ORT eco-system team can use their own pool and image as possible.
Because there're higher code standards, performance and stability requirements.
And we'll give a high priority on ORT pipelines. @snnn

snnn: It's fine if you want to create a new image and new agent pool . But since you need to use QNN, you need to get help from the QNN EP team. Otherwise you would not be able to get QNN SDK at all.

.pipelines/stages/jobs/steps/nuget-win-step.yml
gunandrose4u: Put the nuget install in image build

.pipelines/stages/jobs/nuget-packaging-job.yml
gunandrose4u: in order to avoid breaking build and release process of ORT since they are build so often and release daily. I have set up new image and agent pool for ort-genai to publish windows arm64 pkgs

snnn (approved):
In general please try to get the ORT  binaries from nuget.org instead of Github. 
Nuget has code sign, so that you will not get a corrupted file
Github has a rate limit that may throttle your download.

PR #513, opened 2024-05-24
Add ModelOpt support

sam-india-007:
Change in builder utility to enable an option to use NVIDIA's TensorRT Model Optimizer https://github.com/NVIDIA/TensorRT-Model-Optimizer


sam-india-007:

src/python/py/models/builder.py
```diff
@@ -42,6 +43,8 @@ def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
         self.filename = extra_options["filename"] if "filename" in extra_options else "model.onnx"
         self.extra_options = extra_options
 
+        self.quant_provider = extra_options["quant_provider"] if "filename" in extra_options else "inc"
```
kunal-vaishnavi, line 46:
```suggestion
        self.quant_provider = extra_options["quant_provider"] if "quant_provider" in extra_options else "inc"
```

src/python/py/models/builder.py
```diff
@@ -331,15 +334,19 @@ def save_model(self, out_dir):
...
+                model=model,
+                block_size=self.quant_attrs["int4"]["block_size"],
+                is_symmetric=True,
+                accuracy_level=self.quant_attrs["int4"]["accuracy_level"],
+                nodes_to_exclude=[],
+            )
+            quant.process()
+            return quant.model.model
+        else:
+            quant = modelopt.onnx.quantization.int4.quantize_int4("int4_awq_clip", model)
```
kunal-vaishnavi, line 348:
```suggestion
            import modelopt
            quant = modelopt.onnx.quantization.int4.quantize_int4("int4_awq_clip", model)
```

src/python/py/models/builder.py
```diff
@@ -12,6 +12,7 @@
 from transformers import AutoConfig, AutoModelForCausalLM, AutoTokenizer, GenerationConfig
 import numpy as np
 import torch
+import modelopt
```
kunal-vaishnavi, line 15: Added a suggestion to move the import statement down

```suggestion
```

src/python/py/models/builder.py
```diff
@@ -1996,6 +2003,7 @@ def get_args():
                 enable_cuda_graph = 1 : The model can use CUDA graph capture for CUDA execution provider. If enabled, all nodes being placed on the CUDA EP
                     is the prerequisite for the CUDA graph to be used correctly. It is not guaranteed that cuda graph be enabled as it depends on the model
                     and the graph structure.
+                quant_provider = {inc, modelopt} : Quantization provider
```
kunal-vaishnavi, line 2006: Can you add some more descriptive information about `modelopt` here? You can look at the above options for description examples.

kunal-vaishnavi, line 2006: Can you add some more descriptive information about the `quant_provider` options here? You can look at the above options for description examples.

Can you also add this same information in the [README](https://github.com/microsoft/onnxruntime-genai/blob/main/src/python/py/models/README.md) as a new section under "Extra Options" after "Enable Cuda Graph"?

kunal-vaishnavi, line 2006: Can you add some more descriptive information about the `quant_provider` options here? You can look at the above options for description examples.

Can you also add examples in the [README](https://github.com/microsoft/onnxruntime-genai/blob/main/src/python/py/models/README.md) as a new section under "Extra Options" after "Enable Cuda Graph"?

kunal-vaishnavi, line 2006: Can you add some more descriptive information about the `quant_provider` options here? You can look at the above options for description examples.

Can you also add usage examples in the [README](https://github.com/microsoft/onnxruntime-genai/blob/main/src/python/py/models/README.md) as a new section under "Extra Options" after "Enable Cuda Graph"?

kunal-vaishnavi:
Thank you for your contribution!

How long does the `int4_awq_clip` quantization take with `modelopt`? The goal of the model builder is to quickly produce and save the final ONNX model to disk within a few minutes max. If `int4_awq_clip` takes a while, a note will need to be added somewhere so users are aware in advance that the model builder will take time to run with `int4_awq_clip` quantization enabled.

sam-india-007:
ModelOpt quantization with `int4_awq_clip` can take significant amount of time, but results in smaller and better quantization. Added a disclaimer in the README to this end

src/python/py/models/README.md
```diff
@@ -158,6 +158,20 @@ python3 -m onnxruntime_genai.models.builder -i path_to_local_folder_on_disk -o p
 python3 builder.py -i path_to_local_folder_on_disk -o path_to_output_folder -p precision -e execution_provider -c cache_dir_to_store_temp_files --extra_options enable_cuda_graph=1
 ```
 
+#### Quantization Provider
```
kunal-vaishnavi, line 161: Can you add a link to this section in the table of contents after "Enable Cuda Graph"?

https://github.com/microsoft/onnxruntime-genai/blob/cfb37d495db9067a0ea1402aa7ff2f76657830de/src/python/py/models/README.md?plain=1#L5-L22

src/python/py/models/builder.py
```diff
@@ -331,15 +333,20 @@ def save_model(self, out_dir):
...
+                model=model,
+                block_size=self.quant_attrs["int4"]["block_size"],
+                is_symmetric=True,
+                accuracy_level=self.quant_attrs["int4"]["accuracy_level"],
+                nodes_to_exclude=[],
+            )
+            quant.process()
+            return quant.model.model
+        else:
+            import modelopt
```
jambayk, line 347: Hi, I don't think the modelopt int4 quantization is compatible with the model builder. At least, not with the current state of things. 

The default int4 quantizer `MatMul4BitsQuantizer` uses an onnxruntime contrib op called `MatMulNBits` for int4 quantization. But the modelopt quantizes uses int4 DequantizeLinear and regular MatMul. 
However, onnxruntime currently doesn't support onnx int4 datatype so it cannot even load a model with int4 dq nodes.  Initial support for int4 to ort main was only merged recently https://github.com/microsoft/onnxruntime/pull/20362 and it only supports cpu ep. 

I think a more general approach might be to just use model builder to create a float32/float16 model, and then quantize using modelopt subsequently? Nvidia also contributed the nvmo quantization pass to Olive https://github.com/microsoft/Olive/blob/3d3277b646c2c5a0fa96bcf629a117aa78423176/olive/passes/onnx/nvmo_quantization.py#L47 so it is possible to create a workflow in Olive to first do model builder and then nvmo quantization if desired.

jambayk: Also, I think the modelopt quantizer requires calibration data but is not provided here?

src/python/py/models/builder.py
```diff
@@ -1996,6 +2003,8 @@ def get_args():
                 enable_cuda_graph = 1 : The model can use CUDA graph capture for CUDA execution provider. If enabled, all nodes being placed on the CUDA EP
                     is the prerequisite for the CUDA graph to be used correctly. It is not guaranteed that cuda graph be enabled as it depends on the model
                     and the graph structure.
+                quant_provider = {inc, modelopt} : Quantization provider - You can choose between Intel's Neural Compressor (default) and NVIDIA's TensorRT
+                    Model Optimizer (ModelOpt). Expect slower processing but better compression when using ModelOpt.
```
riyadshairi979, line 2007: `Expect slower processing but better compression when using ModelOpt.`: remove, as are not publishing any experimental data.

sam-india-007: Noted and removed

PR #544, opened 2024-05-30
Update to support running on Android.

skottmckay:
Android test app runs on the emulator, executes a model and validates the output. Complete build including running the emulator is supported from build.py.

Unzip the [onnxruntime-android AAR from maven](https://mvnrepository.com/artifact/com.microsoft.onnxruntime/onnxruntime-android) and provide the path in `--ort_home`. 

e.g. on Windows the command line will look something like this.
`python .\build.py --android --parallel --cmake_generator=Ninja --build_java --android_api=27 --config=Release --android_abi=x86_64 --ort_home=onnxruntime-android-1.18.0 --android_run_emulator`

Script added to build AAR for Android. It's setup to build x64 and arm64. Not sure 32-bit platforms are applicable given the typical model size so have excluded for now. 

e.g.
`python .\tools\ci_build\github\android\build_aar_package.py --ort_home onnxruntime-android-1.18.0 .\tools\ci_build\github\android\default_aar_build_settings.json`

Couple of misc fixes to C++ warnings that show up in the Android build.


yufenglee:
Could we add a ReadMe on how to build on android and IOS?

snnn:
In https://github.com/microsoft/onnxruntime-genai/blob/main/cmake/cxx_standard.cmake, it set CXX standard to 17 if the target OS is Android, in other cases it uses C++20.  I wonder if the special treatment for Android is still needed.

tools/python/util/android.py
```diff
@@ -11,44 +11,40 @@
...
+__all__ = ["get_sdk_tool_paths", "create_virtual_device", "start_emulator", "stop_emulator"]
+
 _log = get_logger("util.android")
 
 
 SdkToolPaths = collections.namedtuple("SdkToolPaths", ["emulator", "adb", "sdkmanager", "avdmanager"])
 
 
-def get_sdk_tool_paths(sdk_root: str):
+def get_sdk_tool_paths(sdk_root: Path):
```
skottmckay, line 31: Match update from ORT repo to make sure we're using the correct emulator.

tools/ci_build/github/android/build_aar_package.py
```diff
@@ -0,0 +1,208 @@
...
+import util  # REPO_ROOT/tools/python/util noqa: E402
+
+# Default to only build 64-bit ABIs. TBD if we need to support any 32-bit ABIs.
+# DEFAULT_BUILD_ABIS = ["armeabi-v7a", "arm64-v8a", "x86", "x86_64"]
+DEFAULT_BUILD_ABIS = ["arm64-v8a", "x86_64"]
+
+# Match onnxruntime and use NDK API 21 by default
+# It is possible to build from source for Android API levels below 21, but it is not guaranteed.
+# As level 21 is Android 5.0 from 2014, it's highly unlikely that we will need to support anything lower.
+DEFAULT_ANDROID_MIN_SDK_VER = 21
```
skottmckay, line 26: This feels really old, but as we're not dependent on any specific Android functionality it shouldn't matter.

src/java/src/test/android/app/src/androidTest/java/ai/onnxruntime_genai/example/javavalidator/SimpleTest.kt
```diff
@@ -1,42 +1,121 @@
...
 
-    @After
-    fun TearDown() {
-        reportHelper.label("Stopping App")
+    @Throws(IOException::class)
+    private fun copyModelFromAssets(): String {
+        // NOTE: We have to read from the app's assets (app/src/main/assets) and write the the app's filesDir.
+        // The unit test's context cannot be used. you'll get mysterious errors like File.mkdirs() will return false and
+        // assertManager.open() throws even if the filename is valid if you try and use it.
+        // Test context is InstrumentationRegistry.getInstrumentation().targetContext.
```
skottmckay, line 34: Important to note this restriction as it will apply to any app using GenAI. 

Any model included in the app will need to be copied to the app's `filesDir` so that it's not part of a compressed archive and the individual files (e.g. the config) are accessible from the native code.

src/models/captured_graph_pool.cpp
```diff
@@ -1,4 +1,3 @@
-#include "onnxruntime_c_api.h"
```
skottmckay, line 1: Cleaned up so the only place we include the ORT API header is in GenAI's onnxruntime_api.h. Any GenAI source that needs access to ORT types should include that file.

snnn (approved):

src/java/AndroidBuild.md
```diff
@@ -0,0 +1,56 @@
...
+
+### Example build commands
+
+#### Windows
+
+.\build.bat --parallel  --config=Release --cmake_generator=Ninja --build_java --android --android_home=D:\Android --android_ndk_path=D:\Android\ndk\26.2.11394342 --android_api=27 --android_abi=x86_64 --ort_home=<path to unzipped onnxruntime-android.aar> --android_run_emulator
+
+#### Linux
+
+./build.sh .\build.bat --parallel  --config=Release --build_java --android --android_home=/home/me/Android --android_ndk_path=/home/me/Android/ndk/26.2.11394342 --android_api=27 --android_abi=x86_64 --ort_home=<path to unzipped onnxruntime-android.aar> --android_run_emulator
```
edgchen1, line 47:
```suggestion
./build.sh --parallel  --config=Release --build_java --android --android_home=/home/me/Android --android_ndk_path=/home/me/Android/ndk/26.2.11394342 --android_api=27 --android_abi=x86_64 --ort_home=<path to unzipped onnxruntime-android.aar> --android_run_emulator
```

src/models/onnxruntime_api.h
```diff
@@ -78,9 +85,72 @@ namespace Ort {
...
+    }
+  }
+
+  if (!api) {
+    __android_log_assert("api != nullptr", "GenAI",
+                         "%s did not have an ORT API version between %d and %d.",
+                         path.c_str(), ORT_API_VERSION, genai_min_ort_api_version);
+  }
+#else
   api = OrtGetApiBase()->GetApi(ORT_API_VERSION);
```
edgchen1, line 150: should we make the behavior regarding valid ORT API versions consistent between Android and non-Android code?

src/models/onnxruntime_api.h
```diff
@@ -78,9 +85,72 @@ namespace Ort {
...
   if (!api)
     throw std::runtime_error("Onnxruntime is installed but is too old, please install a newer version");
+#endif
```
edgchen1, line 153: nit: comments identifying the condition on `#else` and `#endif` are helpful when the original `#if` is farther away

build.py
```diff
@@ -343,6 +349,69 @@ def _get_csharp_properties(args: argparse.Namespace):
...
+        except subprocess.CalledProcessError as e:
+            exception = e
+            print(e)
+            print(f"Output:\n{e.output.decode('utf-8')}")
+            print(f"stderr:\n{e.stderr.decode('utf-8')}")
+
+        # Print test log output so we can easily check that the test ran as expected
+        util.run([adb, "logcat", "-s", "-d", "GenAI:*"])
+        util.run([adb, "logcat", "-s", "-d", "ORTGenAIAndroidTest:*"])
+        util.run([adb, "logcat", "-s", "-d", "TestRunner:*"])
```
edgchen1, line 407: how about using a single logcat call so the relative ordering of the logs from different tags is shown? could use a filter like `"GenAI:V ORTGenAIAndroidTest:V TestRunner:V *:S"`

src/models/onnxruntime_api.h
```diff
@@ -78,9 +85,72 @@ namespace Ort {
...
+
+  const std::string path = "libonnxruntime.so";  // "libonnxruntime4j_jni.so" is also an option if we have issues
+  __android_log_print(ANDROID_LOG_INFO, "GenAI", "Attempting to dlopen %s native library", path.c_str());
+
+  using OrtApiBaseFn = const OrtApiBase* (*)(void);
+  OrtApiBaseFn ort_api_base_fn = nullptr;
+
+  void* ort_lib_handle = dlopen(path.c_str(), RTLD_LOCAL);
+  if (ort_lib_handle == nullptr) {
+    __android_log_assert("ort_lib_handle != nullptr", "GenAI", "Failed to load %s", path.c_str());
```
edgchen1, line 118: should we instead throw an exception to be consistent with the non-Android code path?

skottmckay: This code executes in the context on a Java System.loadLibrary call and it's not clear to me how safe throwing a C++ exception in the genai library and hoping for it and its message to get back to that level with the message intact is.

The use of android log guarantees the message will be output, and the assert will kill the app, which I think is desirable for such a fundamental issue.

src/models/onnxruntime_api.h
skottmckay: For non-Android scenarios we're currently linking against the shared library and packaging them together so there's a fixed version involved. I'm told that will change to separate the packages in the next release, so any looping could be added as part of that.

src/models/onnxruntime_api.h
edgchen1: logging and throwing is also a possibility.

regarding the terminate on error behavior, would that also be more appropriate for other platforms?

build.py
edgchen1: was this one addressed?

skottmckay:
> Could we add a ReadMe on how to build on android and IOS?

@yufenglee I added the Android readme. Not familiar with the iOS setup currently so that will have to be added in a separate PR.

src/models/onnxruntime_api.h
skottmckay: In this specific case I'd prefer to completely die vs throwing and hoping the app that is using the library handles the exception well. It's a fundamental setup error for the ORT library to not be available so should only happen during development. That expected setup/behavior is however Android specific. 

As for other platforms I think we can consider the desired behavior when we separate out ORT from genai in those packages. Not sure what is available/useful. Android's __android_log_assert has the nice property of being debugger friendly (as below, and as I think this is a development-time issue that's very helpful). But doing something like an `abort` in c++ is generally horrible (immediate exit, hard to debug). 

```c
/*
 * Log an assertion failure and SIGTRAP the process to have a chance
 * to inspect it, if a debugger is attached. This uses the FATAL priority.
 */
void __android_log_assert(const char *cond, const char *tag, const char *fmt, ...)
```

build.py
skottmckay: Sorry - I tested it and validated it worked but it ended up in a different branch (juggling too many changes at once). I'm going to checkin this PR so it's done, and include the change in the PR to add the Android packaging pipeline.

Issue #257, opened 2024-04-10
terminated by signal SIGSEGV (Address boundary error) if import onnxruntime before onnxruntime-genai


trajepl:
Cannot create og.Model if user import onnxruntime before the onnxruntime-gneai.
The job is terminated by signal SIGSEGV (Address boundary error)

```python

import onnxruntime as ort

def genai_run(prompt, model_path, max_length=200):

    import time

    import onnxruntime_genai as og

    print("Loading model...")
    app_started_timestamp = time.time()
    model = og.Model(model_path)
    model_loaded_timestamp = time.time()
    print("Model loaded in {:.2f} seconds".format(model_loaded_timestamp - app_started_timestamp))
    tokenizer = og.Tokenizer(model)
    tokenizer_stream = tokenizer.create_stream()
    input_tokens = tokenizer.encode(prompt)
    started_timestamp = time.time()

    print("Creating generator ...")
    params = og.GeneratorParams(model)
    params.set_search_options(
        {
            "do_sample": False,
            "max_length": max_length,
            "min_length": 0,
            "top_p": 0.9,
            "top_k": 40,
            "temperature": 1.0,
            "repetition_penalty": 1.0,
        }
    )
    params.input_ids = input_tokens
    generator = og.Generator(model, params)
    print("Generator created")

    first = True
    new_tokens = []

    while not generator.is_done():
        generator.compute_logits()
        generator.generate_next_token()
        if first:
            first_token_timestamp = time.time()
            first = False

        new_token = generator.get_next_tokens()[0]
        print(tokenizer_stream.decode(new_token), end="")
        new_tokens.append(new_token)

    run_time = time.time() - started_timestamp
    print(
        f"Prompt tokens: {len(input_tokens)}, New tokens: {len(new_tokens)},"
        f" Time to first: {(first_token_timestamp - started_timestamp):.2f}s,"
        f" New tokens per second: {len(new_tokens)/run_time:.2f} tps"
    )


model_path = "xxxx"
genai_run("helo world", model_path)

```
How to reproduce:

1. python -m onnxruntime_genai.models.builder -m microsoft/phi-2 -e cpu -p int4 -o ./models/phi2
2. run above scripts

trajepl:
Not only onnxruntime, it failed when import `torch, transformer`.

baijumeswani:
Could you share more details on this error? And if possible, a stacktrace. That would help me reproduce this.

[...]

PR #548, opened 2024-05-31
Update base image and upgrade GCC version

snnn:
1. Use an internal prebuilt base image instead of a public image, so that we do not need to rebuild the manylinux part again and again.
2. The new CUDA 11 base image is based on almalinux 8 instead of ubi8, so that we can get GCC 11.  ubi8 only has GCC 8 and GCC 12, but GCC 12 is compatible with CUDA 11. So before this change we use GCC 8 in CUDA 11 build.  After this change we will use GCC 11 instead. 
3. Drop the support for GCC 10 and below.

This PR provides another solution for #257 .


tools/ci_build/github/linux/docker/manylinux/Dockerfile.manylinux2_28_cpu
```diff
@@ -1,156 +1,4 @@
...
-     /build_scripts/
-COPY build_scripts/requirements-tools/* /build_scripts/requirements-tools/
-RUN manylinux-entrypoint /build_scripts/finalize.sh && rm -rf /build_scripts
-
-ENV SSL_CERT_FILE=/opt/_internal/certs.pem
-
-CMD ["/bin/bash"]
-
-#Build manylinux2014 docker image end
+FROM onnxruntimebuildcache.azurecr.io/internal/azureml/onnxruntime/build/cpu_x64_ubi8_gcc12:20240530.3
```
skyline75489, line 1: Does this mean the tag (2024530.3) need to be regular updated?

baijumeswani (approved):

PR #557, opened 2024-06-03
Enable ort-genai release pkg for cuda 12.2

gunandrose4u:
Enable ort-genai release python pkg for cuda 12.2


.pipelines/stages/jobs/py-packaging-job.yml
```diff
@@ -81,6 +81,12 @@ jobs:
   - name: os
     value: ${{ parameters.os }}
 
+  - name: feed_name
+    ${{ if eq(parameters.cuda_version, '12.2') }}:
```
skyline75489, line 85: Better also check `eq(parameters.ep, 'cuda`)

gunandrose4u: why

skyline75489: All other packages (DirectML, ROCm) needs the original feed.

gunandrose4u: Make sense. In the future, when cuda default value switch to 12.2. We do not need change here.

baijumeswani (approved):

snnn (approved):

snnn (commented):
I don't see a new yaml file is added. 
Please setup two new pipelines in ADO: one for nuget CUDA 12, one for python CUDA 12. And they should be triggered every day, so that we will have nightly packages published.

baijumeswani:
@snnn until now, we only have a manual trigger to run ort-genai packaging.
But I think we should add a new yml to support nightly packaging for all the packages: ort-genai cpu, dml, cuda11.8, cuda 12.2.

But the matrix can get really big (similar to ort). We will probably need to prioritize which packages we want to run nightly.

snnn:
If you don't have a nightly pipeline if there was any thing wrong with the setting, the problem will only be found when you are doing a release, which could make your release process very tough.

baijumeswani:
You're right. The goal should be to have all the packages build nightly. I'll sync with the team to see if we can prioritize this work.

PR #568, opened 2024-06-05
(WIP) Add DirectML and Phi-3 support to Genny example

### LINKED ISSUE #569: [.NET] Genny chat-bot sample doesn't support DirectML and Phi-3
I would like to update ..examples\csharp\Genny sample to support DirectML and Phi-3 model.

I managed to do it for Stateless mode (#568), however I faces with an issue for Stateful mode:

Inference fails with OnnxRuntimeGenAIException: 'Non-zero status code returned while running DmlFusedNode_0_0 node. Name:'DmlFusedNode_0_0' Status Message: invalid unordered_map<K, T> key'

This happens due to current implementation of `private void AddPastTokens(Sequences sequences)` method in Genny sample:

```csharp
// Only keep (context_length - max_length) worth of history
while (_pastTokens.Count > ModelOptions.ContextLength - SearchOptions.MaxLength)
{
    _pastTokens.RemoveAt(0);
}
```

for Phi3 both ModelOptions.ContextLength and SearchOptions.MaxLength are equal to 4096, so this method removes all tokens from session history and passes empty collection of tokens to `generatorParams.SetInputIDs`. 

Should it be just `while (_pastTokens.Count > ModelOptions.ContextLength)` instead so number of passed tokens doesn't exceed model ContextLength?

Documention on GenAI C# API (https://onnxruntime.ai/docs/genai/api/csharp.html) doesn't provide any description on what each SearchOptions fields mean and how SetInputIDs works, so I don't have a clear view what this check is aimed to achive. 

Could you please provide more info? If my understanding is correct, fix can be applied. If you approve it, I'll finish #568 PR

asmirnov82:
Fixes #569

### DIFF
--- a/examples/csharp/Genny/Genny.sln
+++ b/examples/csharp/Genny/Genny.sln
@@ -8,17 +8,23 @@ EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
 		Debug_Cuda|x64 = Debug_Cuda|x64
+		Debug_DirectML|x64 = Debug_DirectML|x64
 		Debug|x64 = Debug|x64
 		Release_Cuda|x64 = Release_Cuda|x64
+		Release_DirectML|x64 = Release_DirectML|x64
 		Release|x64 = Release|x64
 	EndGlobalSection
 	GlobalSection(ProjectConfigurationPlatforms) = postSolution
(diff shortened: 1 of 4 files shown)


asmirnov82:

asmirnov82:

skyline75489:
Hi! Thanks for the contribution. The CI pipeline was broken yesterday and fixed now. Feel free to re-run the failed jobs.

natke (approved):

PR #576, opened 2024-06-06
Initial support for ROCm build & packaging

skyline75489:
Internal workitem: https://task.ms/aii/29719


cmake/presets/CMakeLinuxBuildPresets.json
```diff
@@ -100,6 +100,38 @@
     {
       "name": "linux_gcc_cuda_minsizerel",
       "configurePreset": "linux_gcc_cuda_minsizerel"
+    },
```
skyline75489, line 103: Do we really need all of these presets? I just copied the CUDA ones

tools/ci_build/github/linux/docker/manylinux/Dockerfile.manylinux2_28_rocm
```diff
@@ -0,0 +1,11 @@
+FROM onnxruntimebuildcache.azurecr.io/internal/azureml/onnxruntime/build/cpu_x64_ubi8_gcc12:20240530.3
```
skyline75489, line 1: Currently no ROCm library is needed during build. So the CPU image is used.

.pipelines/pypl-publishing.yml
```diff
@@ -44,6 +49,11 @@ parameters:
   type: string
   default: '1.18.0-dev-20240426-0116-b842effa29'
 
+- name: ort_rocm_version
+  displayName: 'OnnxRuntime ROCm version'
+  type: string
+  default: '1.19.0-dev-20240602-1103-217b66f'
```
skyline75489, line 55: I found no 1.18 build with ROCm and only 1.19.0 build suits here. I could be wrong.

PR #577, opened 2024-06-06
Initialize Python package validation pipeline

skyline75489:


.pipelines/stages/jobs/py-validation-job.yml
```diff
@@ -0,0 +1,217 @@
...
+        StepName: 'Download Python Wheel Artifacts'
+        ArtifactName: $(ArtifactName)
+        TargetPath: '$(Build.BinariesDirectory)/wheel'
+        SpecificArtifact: ${{ parameters.specificArtifact }}
+        BuildId: ${{ parameters.BuildId }}
+  - ${{ if eq(parameters.os, 'win') }}:
+    - template: steps/utils/flex-download-pipeline-artifact.yml
+      parameters:
+        StepName: 'Download Python Wheel Artifacts'
+        ArtifactName: $(ArtifactName)-wheel
```
skyline75489, line 148: The artifact name should be same between windows & linux build. But currently it's not. See:

https://github.com/microsoft/onnxruntime-genai/blob/a6bec83bf02667f6a9c6aa6340842f46e001578c/.pipelines/stages/jobs/steps/capi-win-step.yml#L161

And:

https://github.com/microsoft/onnxruntime-genai/blob/a6bec83bf02667f6a9c6aa6340842f46e001578c/.pipelines/stages/jobs/steps/capi-linux-step.yml#L159

.pipelines/pypl-publishing.yml
```diff
@@ -1,4 +1,14 @@
 parameters:
+- name: SpecificArtifact
```
skyline75489, line 2: The name is from ORT repo. Should we make it snake_case?

.pipelines/pypl-publishing.yml
```diff
@@ -62,16 +72,31 @@ resources:
...
-    enable_win_cpu: ${{ parameters.enable_win_cpu }}
-    enable_win_cuda: ${{ parameters.enable_win_cuda }}
-    enable_win_dml: ${{ parameters.enable_win_dml }}
-    enable_win_arm64_cpu: ${{ parameters.enable_win_arm64_cpu }}
-    ort_version: ${{ parameters.ort_version }}
-    ort_cuda_version: ${{ parameters.ort_cuda_version }}
-    ort_dml_version: ${{ parameters.ort_dml_version }}
-    cuda_version: ${{ parameters.cuda_version }}
-
+- ${{ if eq(parameters.SpecificArtifact, false) }}:
```
skyline75489, line 75: I'm not sure whether having a separate YAML would be better than using a unified one like this.

.pipelines/stages/jobs/py-validation-job.yml
snnn: It depends. 
Az pipeline has two kinds of artifacts: 
1. https://learn.microsoft.com/en-us/azure/devops/pipelines/artifacts/pipeline-artifacts?view=azure-devops&tabs=yaml
2. https://learn.microsoft.com/en-us/azure/devops/pipelines/tasks/reference/publish-build-artifacts-v1?view=azure-pipelines

If you use the second one, different jobs can use the same job. The first one doesn't. However, the second one should be deprecated.

skyline75489: So the currently used approach is considered deprecated? I guess this is out of the scope of this PR, though. All I want is a consistent way to retrieve the artifacts.

.pipelines/pypl-publishing.yml
baijumeswani: Can we do the validation post packaging in the same pipeline?
Or maybe, we can use a separate pipeline for validation compared with packaging?

skyline75489: That's a great idea! I've update the PR. You can now select whether or not enable post packaging validation:

屏幕截图 2024-07-01 115547

This is my experimental run: https://aiinfra.visualstudio.com/ONNX%20Runtime/_build/results?buildId=499908&view=results

baijumeswani (approved):

PR #582, opened 2024-06-07
Adding Net Standard 2.0 compatibility

### LINKED ISSUE #382: Multi-target with .NET Standard 2.0 and .NET runtimes
Since you're targeting .NET standard, it'd be great to see multi-target support for .NET Standard 2.0 in order to support UWP and .NET Framework to increase reach of availability of these model tools.

You can see this is the general recommendation in the cross-platform guidance support doc here: https://learn.microsoft.com/en-us/dotnet/standard/library-guidance/cross-platform-targeting

There are NuGet packages for allowing for Span for instance for .NET Standard 2.0 which can be included for that target, for instance, see this example from the .NET Community Toolkit: https://github.com/CommunityToolkit/dotnet/blob/7b53ae23dfc6a7fb12d0fc058b89b6e948f48448/src/CommunityToolkit.HighPerformance/CommunityToolkit.HighPerformance.csproj#L28-L37

You could also multi-target the newer runtimes as well this way to have them better optimize performance when applications are targeting those runtimes. E.g. `<TargetFrameworks>netstandard2.0;netstandard2.1;net6.0;net7.0</TargetFrameworks>` ([from here also](https://github.com/CommunityToolkit/dotnet/blob/7b53ae23dfc6a7fb12d0fc058b89b6e948f48448/src/CommunityToolkit.HighPerformance/CommunityToolkit.HighPerformance.csproj#L4)).

RogerBarreto:
To increase the adoption and support to SDKs like `Semantic Kernel` is very desired to also have support for `netstandard 2.0`.

This is an initial suggestion, the implementations might be improved for best performance.

Resolves #382

### DIFF
--- a/src/csharp/Microsoft.ML.OnnxRuntimeGenAI.csproj
+++ b/src/csharp/Microsoft.ML.OnnxRuntimeGenAI.csproj
@@ -1,6 +1,6 @@
 <Project Sdk="MSBuild.Sdk.Extras/3.0.22">
   <PropertyGroup>
-    <TargetFrameworks>netstandard2.0</TargetFrameworks>
+    <TargetFrameworks>net8.0;netstandard2.0</TargetFrameworks>
     <Platforms>AnyCPU</Platforms>
     <LangVersion>default</LangVersion>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
--- a/src/csharp/Microsoft.ML.OnnxRuntimeGenAI.csproj
+++ b/src/csharp/Microsoft.ML.OnnxRuntimeGenAI.csproj
@@ -1,6 +1,6 @@
 <Project Sdk="MSBuild.Sdk.Extras/3.0.22">
   <PropertyGroup>
-    <TargetFrameworks>netstandard2.1</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0</TargetFrameworks>
     <Platforms>AnyCPU</Platforms>
     <LangVersion>default</LangVersion>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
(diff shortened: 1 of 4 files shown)


stephentoub:
I was about to submit a PR to similarly add a netstandard2.0 target. It's important to have.

stephentoub:
cc: @luisquintanilla

test/csharp/TestOnnxRuntimeGenAIAPI.cs
```diff
@@ -5,10 +5,7 @@
 using System.IO;
 using Xunit;
 using Xunit.Abstractions;
-using Microsoft.ML.OnnxRuntimeGenAI;
-using System.Collections.Generic;
 using System.Linq;
-using System.Reflection.Emit;
```
stephentoub, line 11: Most of the changes in this PR are now just style. I'd suggest undoing all of those. In the end, I think this is basically just about updating the TFMs and adding the project reference :smile:

stephentoub (approved):
Nice. Thanks.

Separately, we should consider multitargeting here to both netstandard2.0 and net8.0 at a minimum.

baijumeswani (approved):
Thanks for the contribution. Looks good. :)

src/csharp/Microsoft.ML.OnnxRuntimeGenAI.csproj
```diff
@@ -41,4 +41,8 @@
     <Compile Include="**\*.cs" Link="%(Filename)%(Extension)" />
   </ItemGroup>
 
+  <ItemGroup>
```
WeihanLi, line 44:
```suggestion
  <ItemGroup Condition="'$(TargetFramework)'=='netstandard2.0'">
```

RogerBarreto:
@baijumeswani let me know what I might be missing from the pipeline side, getting 401 errors.

jchen351:
CI should pass now after manally adding upstream dependencies to 
https://aiinfra.visualstudio.com/PublicPackages/_artifacts/feed/ORT-Nightly
If not, keep adding the missing dependencies with the correct version indicated in the [build log](https://github.com/microsoft/onnxruntime-genai/actions/runs/9420849771/job/25954367397#step:9:69).

baijumeswani:
Thanks for the contribution. :)

PR #583, opened 2024-06-07
support netstandard2.0 for csharp

WeihanLi:
contributes to #382


baijumeswani:
Thanks for your contribution [WeihanLi](https://github.com/WeihanLi).

Seems the same as https://github.com/microsoft/onnxruntime-genai/pull/582. Shall we use that PR instead?

WeihanLi:
> Seems the same as #582. Shall we use that PR instead?

Oh sorry, I missed that. Support `netstandard2.0` both, while this just supports `netstandard2.0`, the https://github.com/microsoft/onnxruntime-genai/pull/582 seems mainly to support multi-targeting, maybe we could support `netstandard2.0` firstly, then focus on multi-targeting?

WeihanLi:
closing in favor of https://github.com/microsoft/onnxruntime-genai/pull/582

Issue #592, opened 2024-06-10
Cannot get device_type from Python API


natke:
After loading a model, I tried:

```
     print(model.device_type)
```

but got the following error:

```
 E       TypeError: Unable to convert function return value to a Python type! The signature was
E               (arg0: onnxruntime_genai.onnxruntime_genai.Model) -> Generators::DeviceType
```

PR #593, opened 2024-06-10
Query device type as a string

baijumeswani:
Addresses #592

### DIFF
--- a/test/python/test_onnxruntime_genai_api.py
+++ b/test/python/test_onnxruntime_genai_api.py
@@ -149,3 +149,23 @@ def test_logging():
     og.set_log_options(enabled=True, model_input_values=True, model_output_shapes=True)
     og.set_log_options(model_input_values=False, model_output_shapes=False)
     og.set_log_options(enabled=False)
+
+
+@pytest.mark.parametrize(
+    "relative_model_path",
+    (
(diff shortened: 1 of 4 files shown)


test/python/test_onnxruntime_genai_api.py
```diff
@@ -149,3 +149,23 @@ def test_logging():
...
+        [
+            (Path("hf-internal-testing") / "tiny-random-gpt2-fp32", "CPU"),
+            (Path("hf-internal-testing") / "tiny-random-gpt2-fp32-cuda", "CUDA"),
+            (Path("hf-internal-testing") / "tiny-random-gpt2-fp16-cuda", "CUDA"),
+        ]
+        if og.is_cuda_available()
+        else [(Path("hf-internal-testing") / "tiny-random-gpt2-fp32", "CPU")]
+    ),
+)
+def test_model_device_type(test_data_path, relative_model_path):
```
natke, line 166: Would we want to test it with DML too?

baijumeswani: Unfortunately, we do not have any DML specific pipeline right now. We should add one.

yufenglee (approved):

PR #596, opened 2024-06-11
Make build.py compatible with ARM64 Windows build.

edgchen1:


baijumeswani (approved):

cmake/global_variables.cmake
```diff
@@ -67,7 +67,8 @@ if (MSVC)
     set(genai_target_platform ${CMAKE_SYSTEM_PROCESSOR})
   endif()
 
-  if (genai_target_platform STREQUAL "arm64")
+  if (genai_target_platform STREQUAL "arm64" OR
+      genai_target_platform STREQUAL "ARM64")
```
skottmckay, line 71: Should this set it from 'ARM64' -> 'arm64' so all the downstream logic only have to check for the one casing?

baijumeswani (approved):

PR #598, opened 2024-06-11
Switch "check_language(CUDA)" to "enable_language(CUDA)"

snnn:
Without this change even if we have added "-allow-unsupported-compiler" to CMAKE_CUDA_FLAGS_INIT, the check will still fail. Sorry I don't know why.

### DIFF
--- a/cmake/check_cuda.cmake
+++ b/cmake/check_cuda.cmake
@@ -2,7 +2,8 @@
 include(CheckLanguage)
 
 if(USE_CUDA)
-  check_language(CUDA)
+  enable_language(CUDA)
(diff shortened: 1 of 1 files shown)


snnn:
It looks like this change is still necessary

yufenglee (approved):

yufenglee:
it still doesn't work. Looks like a bug in CMake: 
https://gitlab.kitware.com/cmake/cmake/-/issues/26003
https://gitlab.kitware.com/cmake/cmake/-/merge_requests/9546

snnn:
Which cmake version do you use? The latest one should work fine

PR #603, opened 2024-06-12
Add CMAKE_CUDA_FLAGS_INIT

snnn:
Add CMAKE_CUDA_FLAGS_INIT to the cmake preset files, and append "-allow-unsupported-compiler" there.

We need this change because the latest VC++ compiler isn't compatible with CUDA SDKs we have. 

This PR is similar to https://github.com/microsoft/onnxruntime/pull/21004


snnn:
You may need to use the latest cmake to make it work.

baijumeswani (approved):

yuslepukhin:
Can you add --presets flag to build.py?

snnn:
> Can you add --presets flag to build.py?

I am not familiar with the code. I think the need is better to be addressed separately.

PR #606, opened 2024-06-13
Cleanup code and follow best practices

baijumeswani:
onnxruntime-genai was using `operator=` that has a different meaning than what it was being used as.

This pull request cleans up the code, deletes copy constructor and assignment operator for some of the classes.


yuslepukhin (approved):
:shipit:

PR #612, opened 2024-06-17
Add D3D12 agility SDK dependency to fix crashes on long prompts

PatriceVignola:


baijumeswani (approved):
Will need to merge https://github.com/microsoft/onnxruntime-genai/pull/613 to get the CIs green.

src/models/model.cpp
```diff
@@ -345,13 +345,14 @@ void Model::CreateSessionOptions() {
...
+      auto current_module_path = CurrentModulePath();
+      dml_objects_ = DmlHelpers::CreateDmlObjects(current_module_path);
 
-      auto directml_dll = CurrentModulePath() + L"DirectML.dll";
-      wil::unique_hmodule smart_directml_dll(LoadLibraryExW(directml_dll.c_str(), nullptr, 0));
+      auto directml_dll = current_module_path + "DirectML.dll";
+      wil::unique_hmodule smart_directml_dll(LoadLibraryExA(directml_dll.c_str(), nullptr, 0));
       THROW_LAST_ERROR_IF(!smart_directml_dll);
 
-      if (LoadLibraryExW(directml_dll.c_str(), nullptr, 0) == NULL) {
```
skyline75489, line 354: Why the change from `W` to `A`? I thought we *wanted* to use wide string on Windows for better compatibility.

PatriceVignola: I changed it because the D3D12 factory function takes a non-wide path, but I think it doesn't matter since we don't compile with `UNICODE` defined anyway. Otherwise, changing to `LoadLibraryEx` like I just did wouldn't compile. Either way using `LoadLibraryEx` instead of `LoadLibraryExA` or `LoadLibraryExW` is probably the right thing to do here.

PR #624, opened 2024-06-20
Make GenAi CUDA build run on Windows.

yuslepukhin:
With these changes I am able to use build.py to run the compiler locally.

### DIFF
--- a/cmake/check_cuda.cmake
+++ b/cmake/check_cuda.cmake
@@ -2,6 +2,12 @@
 include(CheckLanguage)
 
 if(USE_CUDA)
+  # Temporary add -allow-unsupported-compiler
+  # Do this before enable_cuda
+  if(WIN32 AND NOT CMAKE_CUDA_FLAGS_INIT)
+    set(CMAKE_CUDA_FLAGS "${CMAKE_CUDA_FLAGS} -allow-unsupported-compiler")
+  endif()
+
   enable_language(CUDA)
   message( STATUS "CMAKE_CUDA_COMPILER_VERSION: ${CMAKE_CUDA_COMPILER_VERSION}")
   if(CMAKE_CUDA_COMPILER)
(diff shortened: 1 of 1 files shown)


cmake/check_cuda.cmake
```diff
@@ -30,7 +36,33 @@ if(USE_CUDA AND CMAKE_CUDA_COMPILER)
   # enable_language(CUDA)
   # message(STATUS "CMAKE_CUDA_COMPILER_VERSION: ${CMAKE_CUDA_COMPILER_VERSION}")
   # set(CUDA_PROPAGATE_HOST_FLAGS ON)
+
+  if(WIN32 AND NOT CMAKE_CUDA_FLAGS_INIT)
+      set(CMAKE_CUDA_FLAGS "${CMAKE_CUDA_FLAGS} /DWIN32 /D_WINDOWS /DWINAPI_FAMILY=100 /DWINVER=0x0A00 /D_WIN32_WINNT=0x0A00 /DNTDDI_VERSION=0x0A000000
```
snnn, line 41: For a local build, you do not need the flags.

snnn: If you set it('/Qspectre'), people who didn't install the VC Qspectre runtime will fail to build our code. 
Therefore I think it's better to not having it.

skyline75489: I thought BinSkim requires `/QSpectre` and is widely used in https://github.com/microsoft/onnxruntime-genai/blob/main/cmake/presets/CMakeWinConfigPresets.json

TLDR: We already need VC QSpectre runtime to build our code.

snnn: You can build this repo without the preset files.

snnn (approved):

baijumeswani (approved):

PR #635, opened 2024-06-22
Add Android AAR packaging pipeline.

skottmckay:
Pipeline will create and publish an AAR with x86_64 and arm64-v8a builds.


skottmckay:
Successful run with changes: https://dev.azure.com/aiinfra/ONNX%20Runtime/_build/results?buildId=492962&view=results

.pipelines/stages/jobs/android-java-api-aar.yml
```diff
@@ -0,0 +1,59 @@
...
+      script: |
+        set -e -x
+        export BUILD_DIR=$(Build.BinariesDirectory)
+        export BUILD_CONFIG=${{parameters.build_config}}
+        export ORT_VERSION=${{parameters.ort_version}}
+        export GENAI_VERSION=$(genai_version)
+        export PACKAGE_NAME=${{parameters.package_name}}
+        export ARTIFACTS_DIR=$(Build.ArtifactStagingDirectory)
+
+        /bin/bash tools/ci_build/github/android/build_aar_and_copy_artifacts.sh
```
edgchen1, line 53: nit: setting the variables here and using them in another script makes it a little harder to see how they are used.

if the logic in build_aar_and_copy_artifacts.sh is not expected to become much more complex, maybe just inline that script here as it is fairly simple now. otherwise, consider a (Python?) script that takes named arguments.

also, is `ORT_VERSION` used here?

edgchen1 (approved):

.pipelines/stages/jobs/android-java-api-aar.yml
skottmckay: CI with changes https://dev.azure.com/aiinfra/ONNX%20Runtime/_build/results?buildId=495491&view=results

edgchen1 (approved):

PR #639, opened 2024-06-24
Support cuda 12 and cudnn 9 in packaging pipeline

baijumeswani:
A few fixes in this PR:

- Packaging pipeline to support building cuda 12 python packages for windows.
- Update cuda 12 docker image to use latest image with cudnn 9.
- Update auditwheel to ignore cudnn9
- Use CXX 17 for all cuda versions in light of https://forums.developer.nvidia.com/t/cudacompile-nvcc-error-cudafe-died-with-status-0xc0000409/260651


snnn (approved):

cmake/cxx_standard.cmake
```diff
@@ -2,9 +2,8 @@ if ("${CMAKE_C_COMPILER_ID}" STREQUAL "GNU" AND CMAKE_C_COMPILER_VERSION VERSION
     add_compile_definitions(USE_CXX17=1)
     message("Test is using C++17 because GCC Version is less than 10")
     set(CMAKE_CXX_STANDARD 17)
-elseif (USE_CUDA AND CMAKE_CUDA_COMPILER AND CMAKE_CUDA_COMPILER_VERSION VERSION_LESS 12)
+elseif (USE_CUDA AND CMAKE_CUDA_COMPILER)
     add_compile_definitions(USE_CXX17=1)
```
snnn, line 6: Do we still need to set it to C++17?

snnn: You hit a problem?

baijumeswani: https://aiinfra.visualstudio.com/ONNX%20Runtime/_build/results?buildId=494247&view=logs&j=61857388-bb1d-59f7-b698-f28b530ffdbd&t=af18331a-2c96-5b9f-1a79-979428754583&l=333

Compiling onnxruntime-genai with cuda 12.2 will yield 

```
CUDACOMPILE : nvcc error : 'cudafe++' died with status 0xC0000409 [D:\a\_work\1\onnxruntime-genai\build\cuda\onnxruntime-genai-static.vcxproj]
```

According to https://forums.developer.nvidia.com/t/cudacompile-nvcc-error-cudafe-died-with-status-0xc0000409/260651, this is a problem only with C++ 20 and VS 2022 v17.10. So, updating to using C++ 17.

jywu-msft: we don't hit this in ORT build?  i think the base image is also using cuda 12.2 ?

baijumeswani: ORT seems to be using CXX 17 for all builds except for mac os builds: https://github.com/microsoft/onnxruntime/blob/4743803944d6f6333bf93b89a62d8083d4466710/cmake/CMakeLists.txt#L43-L47

baijumeswani: With this change, the new build succeeds: https://aiinfra.visualstudio.com/ONNX%20Runtime/_build/results?buildId=494928&view=results

PR #640, opened 2024-06-25
[WIP] Trying /WX

skyline75489:


cmake/presets/CMakeWinConfigPresets.json
```diff
@@ -6,8 +6,8 @@
       "generator": "Visual Studio 17 2022",
       "architecture": "x64",
       "cacheVariables": {
-        "CMAKE_CXX_FLAGS_INIT": "/w15038 /wd4100 /W4 /WX",
```
skyline75489, line 9: `CMAKE_CXX_FLAGS_INIT` gets overwritten and doesn't really work.

src/models/logits.cpp
```diff
@@ -56,7 +56,6 @@ RoamingArray<float> Logits::Get() {
     logits_of_last_token = output_last_tokens_.get();
 
     size_t element_size = type_ == Ort::TypeToTensorType<float>::type ? 4 : 2;
-    size_t size_in_bytes = vocab_size * element_size;
```
skyline75489, line 59: Redefinition in the next following line.

test/c_api_tests.cpp
```diff
@@ -137,7 +137,7 @@ TEST(CAPITests, GreedySearchGptFp32CAPI) {
 
   auto params = OgaGeneratorParams::Create(*model);
   params->SetSearchOption("max_length", max_length);
-  params->SetInputIDs(input_ids.data(), input_ids.size(), sequence_length, batch_size);
```
skyline75489, line 140: Redefinition in the next following line with the same name

skyline75489:
Note that this only affects cmake presets. Using `setup.py` won't be affected.

baijumeswani:
> Note that this only affects cmake presets. Using `setup.py` won't be affected.

Can we enable the warnings as error in build.py also?

skyline75489:
@baijumeswani That's actually what I would prefer. As you can see there's a lot of duplicate flags in CMake presets. If we move them into `CMakeLists.txt`, it can presumably be used in both cmake presets and `setup.py` scenarios. I wonder why we choose to duplicate them flags in presets files instead of directly putting them into `CMakeLists.txt` in the first place.

CMakeLists.txt
```diff
@@ -35,7 +35,10 @@ include(cmake/cxx_standard.cmake)
 add_compile_definitions(BUILDING_ORT_GENAI_C)
 if(MSVC)
   # set updated value for __cplusplus macro instead of 199711L
-  add_compile_options($<$<COMPILE_LANGUAGE:CXX>:/Zc:__cplusplus>)
+  SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /Zc:__cplusplus")
```
skyline75489, line 38: I just found that `add_compile_options` won't work, due to the existence o f global `CMAKE_CXX_FLAGS`

baijumeswani:
>  @baijumeswani That's actually what I would prefer. As you can see there's a lot of duplicate flags in CMake presets. If we move them into CMakeLists.txt, it can presumably be used in both cmake presets and setup.py scenarios. I wonder why we choose to duplicate them flags in presets files instead of directly putting them into CMakeLists.txt in the first place.

In that case, let's make the two consistent. Developers can benefit from catching compilation errors during development time as opposed to during CI stage.

skyline75489:
@baijumeswani in order to reduce the size of the changeset, I moved *some* flags into the main `CMakeLists.txt`, to make `/WX` work. 

I don't know if you know this, but there's this very important difference between global `CMAKE_CXX_FLAG` and `add_compile_options` I discovered while crafting this PR. Using `CMAKE_CXX_FLAG` affects **all** projects including third party deps, whereas `add_compile_options` only affect current projects (to be specific, [the current directory and below](https://cmake.org/cmake/help/latest/command/add_compile_options.html)). 

I think the current `CMAKE_CXX_FLAG` method somehow make it easy to break because we can't control all third-party deps. I also notice that onnxruntime itself actually uses `add_compile_options` instead. With all these considered, I'm hesitated to perform a major overhaul to all the presets and `CMakeLists.txt`, just to add some warning flags.

CMakeLists.txt
```diff
@@ -35,7 +35,10 @@ include(cmake/cxx_standard.cmake)
...
+
+  SET(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} /w15038 /wd4100 /W4 /WX")
```
yufenglee, line 40: Why disable 4100?

skyline75489: I don't really know. This is from the existing CMake presets.

skyline75489:
After talking with @yufenglee I found that I was wrong about `add_compile_options` not working. It does work by combining the global `CMAKE_CXX_FLAGS` with the added option, So I guess using `add_compile_options` is the better way. It limits the `/WX` to ort-genai itself.

cmake/deps.txt
```diff
@@ -14,4 +14,4 @@ pybind11;https://github.com/pybind/pybind11/archive/refs/tags/v2.10.1.zip;769b6a
 googletest;https://github.com/google/googletest/archive/530d5c8c84abd2a46f38583ee817743c9b3a42b4.zip;5e3a61db2aa975cfd0f97ba92c818744e7fa7034
 microsoft_wil;https://github.com/microsoft/wil/archive/refs/tags/v1.0.230629.1.zip;e4a542a323c070376f7c2d1973d0f7ddbc1d2fa5
 directx_headers;https://github.com/microsoft/DirectX-Headers/archive/refs/tags/v1.613.1.zip;47653509a3371eabb156360f42faf582f314bf2e
-onnxruntime_extensions;https://github.com/microsoft/onnxruntime-extensions.git;2f9595af74464822f6b8d5c23153931c9d7a2871
+onnxruntime_extensions;https://github.com/microsoft/onnxruntime-extensions.git;0f1f454867b57511e4ebdf5449483d0ce207b368
```
yufenglee, line 17: why do you update the onnxruntime_extension in this PR?

skyline75489: I've fixed some warnings on their side and need to update the repo to catch up those fixes.

yufenglee (approved):

PR #650, opened 2024-06-27
Address shortcomings of C++17 compatible span substitute

yuslepukhin:
Address problems that prevent span automatic construction on API calls.

### DIFF
--- a/src/span.h
+++ b/src/span.h
@@ -8,7 +8,6 @@
 #include <array>
 #include <stdexcept>
 #include <vector>
-#include <initializer_list>
 
 namespace std {
 namespace generators_span {
--- a/src/span.h
+++ b/src/span.h
@@ -4,40 +4,92 @@
 #ifndef USE_CXX17
 #include <span>
 #else
-#include <array>
-#include <vector>
-namespace std {
 
+#include <array>
+#include <stdexcept>
+#include <vector>
+#include <initializer_list>
+
+namespace std {
+namespace generators_span {
 template <typename T>
 struct span {
   span() = default;
-  span(T* p, size_t length) : p_{p}, length_{length} {}
+  constexpr span(T* p, size_t length) noexcept : p_{p}, length_{length} {}
+  constexpr span(const span<T>& s) noexcept : p_{s.p_}, length_{s.length_} {}
 
-  span(const span<std::remove_const_t<T> >& s) : p_{const_cast<T*>(s.data())}, length_{s.size()} {}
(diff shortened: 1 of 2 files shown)


src/span.h
```diff
@@ -4,40 +4,92 @@
...
+  span() = default;
+  constexpr span(const T* p, size_t length) noexcept : p_{p}, length_{length} {}
+
+  constexpr span(const span<const T>& s) noexcept : p_{s.p_}, length_{s.length_} {}
+  constexpr span(const span<T>& s) noexcept : p_{s.data()}, length_{s.size()} {}
+
+  span(const std::vector<T>& s) noexcept : p_{s.data()}, length_{s.size()} {}
+  template <size_t N>
+  constexpr span(const std::array<T, N>& s) noexcept : p_{s.data()}, length_{s.size()} {}
+  constexpr span(std::initializer_list<T> list) noexcept : p_(&*list.begin()), length_(list.size()) {}
```
edgchen1, line 62: the `std::span(std::initializer_list)` ctor is available in C++26. if we want to switch to `std::span` from C++20, it won't be available yet.

yuslepukhin: I am removing it.

src/span.h
```diff
@@ -4,40 +4,90 @@
...
-  T& back() const { return p_[length_ - 1]; }
+  constexpr T& back() const noexcept { return p_[length_ - 1]; }
 
-  T* begin() const { return p_; }
-  T* end() const { return p_ + length_; }
+  constexpr T* begin() const noexcept { return p_; }
+  constexpr T* end() const noexcept { return p_ + length_; }
 
-  span subspan(size_t index, size_t length) const { return span(p_ + index, length); }
+  span subspan(size_t index, size_t length) const {
```
edgchen1, line 38: https://en.cppreference.com/w/cpp/container/span/subspan
some differences:
- `subspan(size(), 0)` should also be valid.
- there's a default argument value for `length` of `std::dynamic_extent` which has special handling.
- undefined behavior vs. exception.

not sure if the second two are worth implementing now, but maybe the index/length validation can be updated.

"The behavior is undefined if either Offset or Count is out of range. This happens if
- Offset is greater than size(), or
- Count is not std::dynamic_extent and Count is greater than size() - Offset."

yuslepukhin: I intentionally veered off the undefined behavior. I will update the index validation.

yuslepukhin: Made subspan(size(), 0) valid. Kept exceptions. It would not harm compatibility.

edgchen1 (approved):

PR #662, opened 2024-07-01
Load onnxruntime shared lib from onnxruntime python package

baijumeswani:
Now that the onnxruntime python package contains the onnxruntime shared lib, onnxruntime-genai can depend on this shared lib and does not need to embed the onnxruntime shared lib.

Open items: Avoid packaging DirectML.dll inside onnxruntime-genai-directml python package.


yufenglee:
Looks like Windows CPU CI is stuck

yufenglee (approved):

baijumeswani:
The arm64 pipeline provisioning is failing. I am working on that in parallel. Will merge this change in and in try to fix that so as to avoid blocking merges.

PR #699, opened 2024-07-15
Support the Whisper model in onnxruntime-genai

baijumeswani:
Many changes in this PR are from @RyanUnderhill's whisper branch.

This pull-request introduces support for running the openai/whisper model. In particular, this pull-request introduces the following:

- Whisper preprocessing: Given an audio file, the preprocessing stage creates the mel spectrogram and the decoder input ids that are needed for the model execution.
- Whisper model execution: Given the log mel spectrogram and the decoder input ids, the whisper onnx models can be executed. The model execution is split into three phases
  - `EncoderDecoderInit`: The corresponding ONNX model for this phase contains the Encoder and the Decoder with the `Attention` operator. This model is run on the first token generation only.
  - `DecoderInit`: This phase manages the transition from first token generation to the second token generation. i.e. From managing the inputs/outputs of the EncoderDecoderInit ONNX models to managing the inputs/outputs of the Decoder ONNX model.
  - `Decoder`: This phase manages all the remaining token generation steps. The corresponding ONNX model contains only the decoder logic with the `DecoderMaskedMultiHeadAttention` operator.
  
  The model execution also manages the outputs of the model: `logits` and optionally the `cross_qk` buffers required for computing the word level timestamps.
- The Python, C, C++ APIs for loading audios, preprocessing the audio files, and executing the whisper model.

With this pull-request, onnxruntime-genai can execute the openai/whisper model on CPU (fp32) and CUDA (fp16 and fp32) EPs with batch_size >= 1 and beam_size >= 1

Changes required that are not part of this pull-request:
- Making the `cross_qk` buffers available as outputs for the CPU EP.
- Refining the user API.
- Add C# API to load multiple audio files for batch size > 1.
- Add C# example.
- Splitting the phases into two Encoder and Decoder to avoid the decoder weight duplication in EncoderDecoderInit.
- An end-to-end example with word-level timestamps

### DIFF
--- a/src/models/audio_processor.cpp
+++ b/src/models/audio_processor.cpp
@@ -4,7 +4,6 @@
 #include "../generators.h"
 #include "model.h"
 
-
 namespace Generators {
 
 namespace {


examples/python/whisper.py
```diff
@@ -0,0 +1,83 @@
...
+
+class Format:
+    end = "\033[0m"
+    underline = "\033[4m"
+
+
+def run(args: argparse.Namespace):
+    print("Loading model...")
+    model = og.Model(args.model_path)
+    processor = model.create_multimodal_processor()
```
kunal-vaishnavi, line 24: Should we consider creating a processor as `og.Processor(model)` in the future similar to how `og.Tokenizer(model)` exists?

src/generators.h
```diff
@@ -99,7 +99,8 @@ struct GeneratorParams : std::enable_shared_from_this<GeneratorParams>, LeakChec
   std::span<const int32_t> input_ids;  // Array of [batchsize][sequence_length]
 
   struct Whisper {
-    std::shared_ptr<Tensor> input_features;  // float32 [batch_size, number_of_mels, something that is 3000]
+    std::shared_ptr<Tensor> input_features;   // float32 [batch_size, number_of_mels, something that is 3000]
```
kunal-vaishnavi, line 102:
```suggestion
    std::shared_ptr<Tensor> input_features;   // float32 [batch_size, number_of_mels, number_of_frames]
```

src/models/kernels.cu
```diff
@@ -107,5 +121,257 @@ void LaunchInt32ToInt64(const int32_t* src, int64_t* dst, int count, cudaStream_
...
+                                                           beam_ids,
+                                                           batch_size,
+                                                           beam_width,
+                                                           input_seq_length,
+                                                           max_seq_length,
+                                                           current_length);
+}
+
+template <typename T>
+__global__ void CopyCrossQKSingleDecodeStepKernel(T* target, // shape [batch_beam_size, num_alignment_heads, max_length, frame]
```
kunal-vaishnavi, line 245:
```suggestion
__global__ void CopyCrossQKSingleDecodeStepKernel(T* target, // shape [batch_beam_size, num_alignment_heads, max_length, frames]
```

src/models/kernels.cu
```diff
@@ -107,5 +121,257 @@ void LaunchInt32ToInt64(const int32_t* src, int64_t* dst, int count, cudaStream_
   ConvertInt32ToInt64<<<num_blocks, block_size, 0, stream>>>(src, dst, count);
 }
 
+// Support head_size up to 128
+constexpr unsigned int kTileSize = 32;
+constexpr unsigned int kSeqTileSize = 16;
```
kunal-vaishnavi, line 126: Can we move these constants inside `ReorderPastStatesKernelLauncher` and pass them to `ReorderPastStatesKernel`?

src/models/whisper.cpp
```diff
@@ -66,21 +256,174 @@ RoamingArray<float> Whisper_State::Run(int current_length, RoamingArray<int32_t>
...
+      {
+        bool search_buffers = false;
+        UpdateInputsOutputs(next_tokens, next_indices, current_length, search_buffers);
+      }
       break;
+
+    case RunState::Decoder: {
+      bool search_buffers = true;
+      UpdateInputsOutputs(next_tokens, next_indices, current_length, search_buffers);
+    } break;
```
kunal-vaishnavi, line 312: Should the `break;` be before the `}`?

src/models/whisper.cpp
```diff
@@ -66,21 +256,174 @@ RoamingArray<float> Whisper_State::Run(int current_length, RoamingArray<int32_t>
...
+  }
+
+  if (cache_indirection_) {
+#if USE_CUDA
+    gpu_span<int32_t> beam_indices_gpu = beam_indices.GetGPU();
+    cuda_unique_ptr<int32_t> beam_indices_ptr;
+    if (beam_indices_gpu.empty()) {
+      beam_indices_ptr = CudaMallocArray<int32_t>(params_->batch_size, &beam_indices_gpu);
+      std::vector<int32_t> beam_indices_cpu(params_->batch_size, 0);
+      std::iota(beam_indices_cpu.begin(), beam_indices_cpu.end(), 0);
```
kunal-vaishnavi, line 344: Should we do this elsewhere instead? This initialization isn't Whisper specific.

src/python/python.cpp
```diff
@@ -275,7 +278,8 @@ struct PyGeneratorParams {
   }
 
   pybind11::array_t<int32_t> py_input_ids_;
-  pybind11::array_t<float> py_whisper_input_features_;
+  pybind11::array py_whisper_input_features_;
```
kunal-vaishnavi, line 281: Can we add a TODO comment here that this needs to be renamed since we will do that in a subsequent PR?

src/models/kernels.cu
baijumeswani: Seems like it must be `constexpr` within the context of this function since there is a memory allocation in the kernel:

```cpp
__shared__ float4 tile[kSeqTileSize][kTileSize + 1];
```

I'll put these in an unnamed namespace for now. Or if we want to, we can also create the \_\_shared\_\_ allocation outside of the kernel and define it as extern. Let me know if you have a preference.

src/models/whisper.cpp
baijumeswani: I think this should be local to file since the beam indices (for non-beam-search search algorithms) is only needed by this whisper model due to DMMHA. Other models will probably never need to care about beam_indices in case beam-search is not used. 
If you think we can move the logic to some other place, I can try to move it.

src/models/kernels.cu
baijumeswani: I ended up putting the constexprs in a struct of its own. This way, we continue benefitting from compile time memory allocation as opposed to dynamic memory allocation and also make the consts <somewhat> specific to the `ReorderPastStates` kernel

src/models/audio_processor.cpp
```diff
@@ -0,0 +1,103 @@
...
+  ort_extensions::OrtxObjectPtr<OrtxTensorResult> result;
+  CheckResult(OrtxSpeechLogMel(processor_.get(), audios->audios_.get(), ort_extensions::ptr(result)));
+
+  ort_extensions::OrtxObjectPtr<OrtxTensor> mel;
+  CheckResult(OrtxTensorResultGetAt(result.get(), 0, ort_extensions::ptr(mel)));
+
+  named_tensors->emplace(std::string(Config::Defaults::InputFeaturesName),
+                         std::make_shared<Tensor>(ProcessMel(mel, input_features_type_, allocator)));
+
+  constexpr auto start_of_transcript = "<|startoftranscript|>";
```
yufenglee, line 83: those are whisper specific. i am thinking how we can make it more flexible to support models that need audio processing.

src/models/audio_processor.cpp
```diff
@@ -0,0 +1,103 @@
...
+  const int32_t start_of_transcript_token_id = tokenizer.TokenToTokenId(start_of_transcript);
+  const auto prompt_token_ids = tokenizer.GetDecoderPromptIds(audios->num_audios_, language, task, no_timestamps);
+
```
yufenglee, line 86: can we treat those as an prompt of the whisper model and allow user to compose this by themselves?

yufenglee: and Whispler allows user to provide 'prev' tokens to help get better transcription. It can also be part of the prompt of the whipser model

baijumeswani: The problem with doing this approach is that then we need to detect `<|startoftranscript|>` in the user prompt. The `<|startoftranscript|>` may not always map to a single token when encoding with the tokenizer.

We can handle this by detecting model type here or by asking users to provide the token ids instead of a string prompt.

But I think we need to still add the ability for users to supply previously generated tokens which is not available right now. I'll add this functionality.

examples/python/whisper.py
yufenglee: Whisper is a typical encoder-decoder model. It is different with the Multi-Modal model. the process i am thinking:

prompt = "<prev>prev_text<SOT><en><transcrib>"
prompt_tokens = Processor.encode(prompt)
encoder_input = Processor.ProcessAudio(audio);

generator.append(prompt_tokens)
generaotr.set_encoder_input(encoder_input)

RyanUnderhill (approved):

src/ort_genai_c.h
```diff
@@ -273,8 +280,21 @@ OGA_EXPORT void OGA_API_CALL OgaDestroyMultiModalProcessor(OgaMultiModalProcesso
...
+ * \brief Converts the given string to a single token id and adds it to the end of the sequence at the given index.
+          If the sequence does not exist, a new sequence is created at the given index if sequence_idx is equal to the current sequences count.
+ * \param[in] tokenizer The tokenizer to use to convert the string to a token id.
+ * \param[in] str The string to convert to a token id.
+ * \param[in] sequences The OgaSequences to add the token id to.
+ * \param[in] sequence_idx The index of the sequence to add the token id to.
+ * \return OgaResult containing the error message if the conversion of the string to a token id failed.
+ */
+OGA_EXPORT OgaResult* OGA_API_CALL OgaTokenizerToTokenId(const OgaTokenizer* tokenizer, const char* str, OgaSequences* sequences, size_t sequence_idx);
+
```
yufenglee, line 293: I'm thinking if it is better to return a token id directly instead of add it to OgaSequences. 

We may add an API to allow OgaSequences to append an token id

baijumeswani: Sure, can do that.

src/models/audio_processor.cpp
baijumeswani: These are now removed. And the user must provide the input ids directly.

src/ort_genai_c.h
baijumeswani: Updated the API and example.

examples/python/whisper.py
baijumeswani: Many models have very blurred lines between the encoder and the decoder (such as the phi3 vision model).
I don't think the user wants to know whether a model is encoder-decoder model or if it is only a decoder model (or if it something else).

Unless we want to expose ways for the user to manipulate the output of the encoder before it is fed back into the decoder, we can potentially hide this complexity from the user.

For now, this pull-request separates out the encoder and decoder inputs and the user must explicitly set both the encoder as well as the decoder inputs by themselves.

We should discuss and review this design choice in a meeting/scrum when possible.

src/models/model.h
```diff
@@ -74,6 +76,11 @@ struct Tokenizer : std::enable_shared_from_this<Tokenizer>, LeakChecked<Tokenize
   std::vector<int32_t> EncodeBatch(std::span<const std::string> strings) const;
   std::vector<std::string> DecodeBatch(std::span<const int32_t> sequences, size_t count) const;
 
+  std::vector<int32_t> GetDecoderPromptIds(size_t batch_size, const std::string& language,
+                                           const std::string& task, int32_t no_timestamps) const;
```
yufenglee, line 80: do we still need this function?

yufenglee:
we need to add some tests in following PR.

src/models/whisper.cpp
```diff
@@ -66,21 +252,174 @@ RoamingArray<float> Whisper_State::Run(int current_length, RoamingArray<int32_t>
...
+                                cross_qk_search_buffer_->GetTensorData<float>(),
+                                cross_qk_final_->GetTensorMutableData<float>(),
+                                params_->search.num_return_sequences,
+                                cache_indirection_->GetTensorData<int32_t>());
+#endif
+  }
 }
 
+OrtValue* Whisper_State::GetOutput(const char* name) {
+  if (std::strcmp("cross_qk", name) == 0) {
```
yufenglee, line 419: >"cross_qk" [](http://example.com/codeflow?start=18&length=10)

could we remove this dirty hard code?

yufenglee:
and add a description on how to create the model

---
In reply to: [2347714980](https://github.com/microsoft/onnxruntime-genai/pull/699#issuecomment-2347714980) [](http://example.com/codeflow?ancestors=2347714980)

src/models/model.h
baijumeswani: We don't. I'll remove it.

src/models/whisper.cpp
baijumeswani: Added some code comments highlighting that the cross_qk_final_ is a buffer maintained by ort-genai.

We need to document it in an example and probably in our API docs if possible. We can do this in a follow up pull-request.

yufenglee (approved):

src/models/kernels.cu
```diff
@@ -107,5 +121,262 @@ void LaunchInt32ToInt64(const int32_t* src, int64_t* dst, int count, cudaStream_
...
+  } else if (time_step == (current_length - 1)) {
+    // For the final (newly generated) time step,
+    // the beam that it comes from is always the beam that we
+    // are currently processing (i.e.) from this point on, these time-steps
+    // form the new beams.
+    tgt_indir_cache[tgt_offset] = static_cast<int32_t>(beam_id);
+  } else {
+    // For all other time-steps, we look up the source indirection, to
+    // see which beam it came from based on the `src_beam`.
+    // printf("else branch\n");
```
kunal-vaishnavi, line 223:
```suggestion
```

src/models/audio_processor.cpp
```diff
@@ -0,0 +1,84 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#include "../generators.h"
+#include "model.h"
+
+#include <regex>
```
kunal-vaishnavi, line 7:
```suggestion
```

kunal-vaishnavi (approved):

PR #702, opened 2024-07-16
Oga global functions should be inline

skyline75489:
The `inline` is needed to prevent duplicated symbols when this header is imported more than once.


skyline75489:
To reviewers: should this be `inline`, `static` or `static inline`? I'm not sure...

baijumeswani (approved):

baijumeswani:
> To reviewers: should this be inline, static or static inline? I'm not sure...

inline is good. Thanks for making the change.

PR #706, opened 2024-07-17
fix, potential fix, thoughts

aciddelgado:


benchmark/python/benchmark_e2e.py
```diff
@@ -259,7 +258,7 @@ def run_benchmark(args, model, tokenizer, batch_size, prompt_length, generation_
         if args.use_graph_capture:
             params.try_graph_capture_with_max_batch_size(batch_size)
 
-        generator = og.Generator(model, params)
+        generator = og.Generator(model, params) # TODO: Generator allocates memory. Should we measure this? Should we add functionality so we can reuse this memory?
```
baijumeswani, line 261: I think we should detect when the generator goes out of scope and release the allocator memory associated with the released generator so the memory can be reused by others.

But I think we can do it as a separate feature (and not through gc.collect which might/might not work).

PR #709, opened 2024-07-18
Initialize Objective-C bindings

skyline75489:
This PR contains Objective-C wrapper code, podspec generation infras and the accompanying test code.

Internal workitem: https://task.ms/aii/35199


src/objectivec/error_utils.h
```diff
@@ -0,0 +1,27 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
skyline75489, line 1: Copied from onnxruntime repo: https://github.com/microsoft/onnxruntime/blob/main/objectivec/error_utils.h

src/objectivec/oga_internal.mm
```diff
@@ -0,0 +1,54 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#import "oga_internal.h"
+
+@implementation OGAInt32Span {
```
skyline75489, line 6: This is to mimic `std::span` which is not available until C++ 20, also not suitable in an ObjC environment.

src/objectivec/oga_internal.h
```diff
@@ -0,0 +1,81 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
skyline75489, line 1: Keep things we don't want to expose internal, mostly C++ things.

edgchen1 (commented):
please document public APIs and add unit tests

src/objectivec/oga_internal.mm
```diff
@@ -0,0 +1,54 @@
...
+  return _ptr;
+}
+
+- (size_t)size {
+  return _size;
+}
+
+- (int32_t)last {
+  return *(_ptr + (_size - 1));
+}
```
skottmckay, line 27: What's the use case for this? std::span has things like `begin` and `end` but that's slightly different to returning the value of the last element (and this would also do the wrong thing if size was 0 wouldn't it)?

src/objectivec/oga_internal.h
skottmckay: I'm not an objective-c expert. can you explain how this keeps things internal? maybe we could add a comment in the file if that would help the next developer understand how it works as we rarely work on objective-c.

src/objectivec/oga_tokenizer.mm
```diff
@@ -0,0 +1,85 @@
...
+    return nil;
+  }
+  try {
+    _tokenizer->Encode([str UTF8String], [sequences CXXAPIOgaSequences]);
+    return sequences;
+  }
+  OGA_OBJC_API_IMPL_CATCH_RETURNING_NULLABLE(error)
+}
+
+- (nullable NSString*)decode:(OGAInt32Span*)data error:(NSError**)error {
```
skottmckay, line 36: The C# API has encode/decode of batches. Is there a reason for this to be inconsistent and only support a single input?

src/objectivec/oga_tokenizer.mm
```diff
@@ -0,0 +1,85 @@
...
+  OGA_OBJC_API_IMPL_CATCH_RETURNING_NULLABLE(error)
+}
+
+- (const OgaTokenizer&)CXXAPIOgaTokenizer {
+  return *(_tokenizer.get());
+}
+
+@end
+
+@implementation OGATokenizerStream {
```
skottmckay, line 50: nit: could this be in oga_tokenizer_stream.mm? we usually keep it 1:1 between non-trivial types and the file they're in

src/objectivec/oga_internal.h
skyline75489: Oh I didn't include the podspec here because I think it's not ready. The way to keep it internal is by tweaking the podspec.

src/objectivec/oga_tokenizer.mm
skyline75489: No. This PR is just the first step towards a usable example. More methods including batches support will be added.

src/objectivec/oga_tokenizer.mm
skyline75489: sure. will do

src/objectivec/oga_generator_params.mm
```diff
@@ -0,0 +1,68 @@
...
+
+- (BOOL)setInputs:(OGANamedTensors*)namedTensors error:(NSError**)error {
+  try {
+    _generatorParams->SetInputs([namedTensors CXXAPIOgaNamedTensors]);
+    return YES;
+  }
+  OGA_OBJC_API_IMPL_CATCH_RETURNING_BOOL(error)
+}
+
+- (BOOL)setInputSequences:(OGASequences*)sequences error:(NSError**)error {
```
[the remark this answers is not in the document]
skyline75489: Added `SetInputIDs`

tools/ci_build/github/apple/get_simulator_device_info.py
edgchen1, line 1: FYI, this was updated recently in onnxruntime. you may want to copy the latest as it allows you to specify the simulator version, which could be important depending on which Xcode version is used to build and test.

src/objectivec/include/ort_genai_objc.h
```diff
@@ -0,0 +1,245 @@
...
+
+- (nullable OGAInt32Span*)sequenceAtIndex:(size_t)index;
+
+@end
+
+@interface OGATensor : NSObject
+
+- (instancetype)init NS_UNAVAILABLE;
+- (nullable instancetype)initWithDataPointer:(void*)data
+                                       shape:(OGAInt64Span*)shape
```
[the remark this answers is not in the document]
skyline75489: Good idea! I'll see if I can keep the span-like structure inside GenAI

src/objectivec/oga_generator.mm
```diff
@@ -0,0 +1,53 @@
...
+  }
+  OGA_OBJC_API_IMPL_CATCH_RETURNING_NULLABLE(error)
+}
+
+- (BOOL)isDone {
+  return _generator->IsDone();
+}
+
+- (void)computeLogits {
+  _generator->ComputeLogits();
```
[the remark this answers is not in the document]
skyline75489: sure

src/objectivec/oga_images.mm
```diff
@@ -0,0 +1,35 @@
...
+@implementation OGAImages {
+  std::unique_ptr<OgaImages> _images;
+}
+
+- (nullable instancetype)initWithPath:(NSArray<NSString *> *)paths error:(NSError**)error {
+  if ((self = [super init]) == nil) {
+    return nil;
+  }
+
+  std::vector<const char *> cpp_paths;
```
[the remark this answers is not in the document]
skyline75489: nice catch

src/objectivec/oga_internal.h
```diff
@@ -0,0 +1,65 @@
...
+@interface OGATokenizer ()
+
+- (const OgaTokenizer&)CXXAPIOgaTokenizer;
+
+@end
+
+@interface OGASequences ()
+
+- (nullable instancetype)initWithError:(NSError**)error;
+- (instancetype)initWithNativePointer:(std::unique_ptr<OgaSequences>)ptr;
```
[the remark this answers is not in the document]
skyline75489: yeah...makes sense

src/objectivec/oga_internal.mm
```diff
@@ -0,0 +1,66 @@
...
+- (size_t)size {
+  return _size;
+}
+
+- (int32_t)last {
+  if (_size - 1 <= 0) {
+    NSException* exception = [NSException exceptionWithName:@"onnxruntime-genai"
+                                                     reason:@"The size of this span is invalid"
+                                                   userInfo:nil];
+    @throw exception;
```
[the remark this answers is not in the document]
skyline75489: sure

src/objectivec/oga_sequences.mm
```diff
@@ -0,0 +1,46 @@
...
+  OGA_OBJC_API_IMPL_CATCH_RETURNING_NULLABLE(error)
+}
+
+- (size_t)count {
+  return _sequences->Count();
+}
+
+- (nullable OGAInt32Span*)sequenceAtIndex:(size_t)index {
+  if (index >= [self count]) {
+    return nil;
```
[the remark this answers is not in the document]
skyline75489: done

src/objectivec/oga_internal.mm
```diff
@@ -0,0 +1,66 @@
...
+- (const int32_t*)pointer {
+  return _ptr;
+}
+
+- (size_t)size {
+  return _size;
+}
+
+- (int32_t)last {
+  if (_size - 1 <= 0) {
```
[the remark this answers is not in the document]
skyline75489: done

src/objectivec/test/model_tests.mm
```diff
@@ -0,0 +1,114 @@
...
+
+    const auto max_length = 10;
+    const auto batch_size = input_ids_shape[0];
+    const auto input_sequence_length = input_ids_shape[1];
+
+    NSBundle* bundle = [NSBundle mainBundle];
+    NSString* path = [[bundle resourcePath] stringByAppendingString:@"hf-internal-testing/tiny-random-gpt2-fp32"];
+
+    NSError *error = nil;
+    OGAModel* model = [[OGAModel alloc] initWithPath:path error:&error];
```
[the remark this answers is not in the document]
skyline75489: done

src/objectivec/include/ort_genai_objc.h
```diff
@@ -0,0 +1,245 @@
...
+ *
+ * @param str The text to be encoded.
+ * @param error Optional error information set if an error occurs.
+ * @return The encoding result, or nil if an error occurs.
+ */
+- (nullable OGASequences*)encode:(NSString*)str
+                           error:(NSError**)error;
+
+/**
+ * Decode sequences to text
```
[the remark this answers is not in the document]
skyline75489: Added, except the APIs that will be removed in the following refactor effort.

tools/ci_build/github/apple/get_simulator_device_info.py
skyline75489: done

src/objectivec/include/ort_genai_objc.h
```diff
@@ -0,0 +1,392 @@
...
+ */
+- (nullable instancetype)initWithModel:(OGAModel*)model
+                                params:(OGAGeneratorParams*)params
+                                 error:(NSError**)error NS_DESIGNATED_INITIALIZER;
+/**
+ * Whether generation is done
+ * @param error Optional error information set if an error occurs.
+ * @return The result, or nil if an error occurs.
+ */
+- (nullable NSNumber*)isDoneWithError:(NSError**)error NS_SWIFT_NAME(isDone());
```
skyline75489, line 339: This one is tricky. If I use `BOOL`, it will fall into the "BOOL as if the method success" category. So I'm using nullable NSNumber and check it's `boolValue` for success or failure.

src/objectivec/include/ort_genai_objc.h
```diff
@@ -0,0 +1,406 @@
...
+ * @return The last element, or nil if an error occurs.
+ */
+- (int32_t)lastElementWithError:(NSError**)error NS_SWIFT_NAME(lastElement());
+
+@end
+
+/**
+ * Wraps a raw int64_t pointer and its size. Represents a data sequence.
+ */
+@interface OGAInt64Span : NSObject
```
edgchen1, line 189: do we still need `OGAInt64Span`?

src/objectivec/test/model_tests.mm
```diff
@@ -0,0 +1,125 @@
...
+#import <XCTest/XCTest.h>
+
+#import "ort_genai_objc.h"
+#import "assertion_utils.h"
+#import <vector>
+#import <array>
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface ORTGenAIModelTest : XCTestCase
```
edgchen1, line 14: may want to set [`continueAfterFailure`](https://developer.apple.com/documentation/xctest/xctestcase/1496260-continueafterfailure?language=objc) so that the test logic doesn't keep going after an assertion failure.

src/objectivec/test/model_tests.mm
```diff
@@ -0,0 +1,125 @@
...
+
+@end
+
+@implementation ORTGenAIModelTest
+
+
+- (void)GreedySearchGptFp32 {
```
edgchen1, line 21: do test methods need to start with "test"?

https://developer.apple.com/documentation/xctest/defining_test_cases_and_test_methods?language=objc

src/objectivec/include/ort_genai_objc.h
```diff
@@ -0,0 +1,406 @@
...
+ */
+- (const int32_t*)pointer;
+/**
+ * The underlying data size
+ */
+- (size_t)size;
+/**
+ * The last element in this data sequence
+ * @param error Optional error information set if an error occurs.
+ * @return The last element, or nil if an error occurs.
```
edgchen1, line 180: the return type can't be nil

src/objectivec/oga_internal.h
```diff
@@ -0,0 +1,65 @@
...
+- (instancetype)initWithCXXPointer:(std::unique_ptr<OgaTensor>)ptr;
+
+- (OgaTensor&)CXXAPIOgaTensor;
+
+@end
+
+@interface OGANamedTensors ()
+
+- (instancetype)initWithCXXPointer:(std::unique_ptr<OgaNamedTensors>)ptr
+    NS_DESIGNATED_INITIALIZER;
```
edgchen1, line 54: should some of the other initializers in this file also be marked `NS_DESIGNATED_INITIALIZER`?

src/objectivec/oga_sequences.mm
```diff
@@ -0,0 +1,52 @@
...
+
+- (size_t)count {
+  return _sequences->Count();
+}
+
+- (nullable OGAInt32Span*)sequenceAtIndex:(size_t)index
+                                    error:(NSError**)error {
+  if (index >= [self count]) {
+    NSDictionary *errorDictionary = @{NSLocalizedDescriptionKey : @"The index is out of bounds"};
+    *error = [[NSError alloc] initWithDomain:kOgaErrorDomain code:-1 userInfo:errorDictionary];
```
edgchen1, line 37: check `error != nil` first?

to simplify, could just put this in the try block and throw a normal C++ exception.

src/objectivec/oga_sequences.mm
```diff
@@ -0,0 +1,52 @@
...
+
+  try {
+    _sequences = OgaSequences::Create();
+    return self;
+  }
+  OGA_OBJC_API_IMPL_CATCH_RETURNING_NULLABLE(error)
+}
+
+- (size_t)count {
+  return _sequences->Count();
```
edgchen1, line 30: nit: also catch exceptions here. even if it doesn't throw now, it's safer to treat all C++ code as potentially throwing.

edgchen1: general: maybe use methods instead of properties so NSError may be used

src/objectivec/oga_sequences.mm
skyline75489: yeah this seems unnecessary, I'll just delete it and make use of the C++ exceptions.

src/objectivec/error_utils.mm
```diff
@@ -0,0 +1,33 @@
...
+void OGASaveCodeAndDescriptionToError(int code, NSString* description, NSError** error) {
+  if (!error) return;
+
+  *error = [NSError errorWithDomain:kOgaErrorDomain
+                               code:code
+                           userInfo:@{NSLocalizedDescriptionKey : description}];
+}
+
+void OGASaveExceptionToError(const std::exception& e, NSError** error) {
+  OGASaveCodeAndDescriptionToError(0x0A, e.what(), error);
```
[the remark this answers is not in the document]
skyline75489: GenAI doesn't have a set of error code like ORT, so I just use 0x0A as the temp error. Will make it a named constant.

src/objectivec/include/ort_genai_objc.h
skyline75489: yeah we don't. will move it into `oge_internal.h` and out of public interfaces.

src/objectivec/test/model_tests.mm
skyline75489: sure

src/objectivec/test/model_tests.mm
skyline75489: yeah. thanks for catching this. i forgot when i modified it, and since then the test are actually not run (just compiles). And because I haven't set up the testing pipeline yet, i didn't notice it until now.

src/objectivec/include/ort_genai_objc.h
skyline75489: fixed

src/objectivec/oga_internal.h
skyline75489: yeah. I reorganized some of them initializers.

src/objectivec/oga_sequences.mm
skyline75489: done

src/objectivec/include/ort_genai_objc.h
edgchen1: yeah, that is tricky. does the suggestion in this answer about `swift_error(nonnull_error)` work?
https://stackoverflow.com/a/45566261

it might be nice to have a boolean return value in swift if possible.

tools/ci_build/github/apple/objectivec/objc.podspec.template
```diff
@@ -0,0 +1,54 @@
...
+      ]
+    end
+
+    core.pod_target_xcconfig = {
+      "HEADER_SEARCH_PATHS" => include_dirs.join(" "),
+      "OTHER_CPLUSPLUSFLAGS" => "-fvisibility=hidden -fvisibility-inlines-hidden",
+    }
+
+  end
+  s.dependency 'onnxruntime-objc', '~> @ORT_VERSION@'
```
edgchen1, line 52: why does it need to depend on onnxruntime-objc?

.pipelines/stages/jobs/macos-ios-cocoapods-packaging-job.yml
```diff
@@ -0,0 +1,40 @@
...
+      parameters:
+        build_config: ${{parameters.build_config}}
+
+
+    - task: CmdLine@2
+      displayName: Build MacOS & iOS CocoaPods Packages
+      inputs:
+        script: |
+          set -e -x
+          python tools/ci_build/github/apple/build_and_assemble_apple_pods.py \
```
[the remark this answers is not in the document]
skyline75489: Yeah sure. I didn't know this exists. Thanks

src/objectivec/oga_tokenizer.mm
```diff
@@ -0,0 +1,49 @@
...
+  try {
+    _tokenizer = OgaTokenizer::Create([model CXXAPIOgaModel]);
+    return self;
+  }
+  OGA_OBJC_API_IMPL_CATCH_RETURNING_NULLABLE(error)
+}
+
+- (nullable OGASequences*)encode:(NSString*)str error:(NSError**)error {
+  OGASequences* sequences = [[OGASequences alloc] initWithError:error];
+  if (*error) {
```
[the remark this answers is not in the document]
skyline75489: This is a pure mistake. Will correct it.

src/objectivec/oga_tensor.mm
```diff
@@ -0,0 +1,49 @@
...
+}
+
+- (nullable instancetype)initWithDataPointer:(void*)data
+                          shape:(NSArray<NSNumber*>*)shape
+                           type:(OGAElementType)elementType
+                          error:(NSError**)error {
+  try {
+    std::vector<int64_t> cxxShape;
+    for (NSNumber* object in shape) {
+      cxxShape.push_back([object intValue]);
```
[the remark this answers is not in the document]
skyline75489: yeah sure.

tools/ci_build/github/apple/objectivec/objc.podspec.template
skyline75489: i'm not sure about this. technically it doesn't *need* `onnxruntime-objc` at this moment. but I'm not sure if depending on `onnxruntime-c` will always suffice in the future. There isn't really harm including `onnxruntime-objc`, is there?

src/objectivec/oga_internal.mm
```diff
@@ -0,0 +1,76 @@
...
+      *error = [[NSError alloc] initWithDomain:kOgaErrorDomain code:-1 userInfo:errorDictionary];
+    }
+    return -1;
+  }
+  return *(_ptr + (_size - 1));
+}
+
+@end
+
+@implementation OGAInt64Span {
```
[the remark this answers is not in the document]
skyline75489: I think in the future after the API refactor, this will be used. but you're right. I'll remove it and add back if needed.

src/objectivec/include/ort_genai_objc.h
```diff
@@ -0,0 +1,379 @@
...
+ * A series of generated sequences
+ */
+@interface OGASequences : NSObject
+
+- (instancetype)init NS_UNAVAILABLE;
+
+/**
+ * The count of generated sequences
+ * @param error Optional error information set if an error occurs.
+ * @return The count of sequences, or -1 if an error occurs.
```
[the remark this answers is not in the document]
skyline75489: Should I make it `NSNumber` and use `nil` to indicate error?

src/objectivec/oga_tensor.mm
```diff
@@ -0,0 +1,49 @@
...
+  OGA_OBJC_API_IMPL_CATCH_RETURNING_NULLABLE(error)
+}
+
+- (OGAElementType)type {
+  return OGAElementType(_tensor->Type());
+}
+
+- (void*)data {
+  return _tensor->Data();
+}
```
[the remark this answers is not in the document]
skyline75489: I was under the impression that this is just a pointer access. but yeah it actually can throw. Thanks.

src/objectivec/include/ort_genai_objc.h
skyline75489: The attribute `swift_error(nonnull_error)` will handle the naming. But in objective-c, for a signature like `(BOOL)isDoneWithError`,  should I just return `false` to indicate error?

tools/ci_build/github/apple/objectivec/objc.podspec.template
edgchen1: > There isn't really harm including onnxruntime-objc, is there?

well, it's an additional thing to download. if we need the dependency later, we can add it then.

tools/ci_build/github/apple/test_apple_packages.py
```diff
@@ -0,0 +1,299 @@
...
+
+        # Download phi3 model
+        model_dir = target_proj_path / "models" / "Phi-3-mini-4k-instruct-onnx"
+
+        print(f"Downloading models:\n{model_dir}")
+        snapshot_download(
+            repo_id="microsoft/Phi-3-mini-4k-instruct-onnx",
+            allow_patterns="cpu_and_mobile/cpu-int4-rtn-block-32-acc-level-4/*",
+            local_dir=model_dir
+        )
```
edgchen1, line 132: where is this model used?

tools/ci_build/github/apple/package_assembly_utils.py
```diff
@@ -0,0 +1,173 @@
...
+            if not ios_deployment_target:
+                ios_deployment_target = target
+            else:
+                # should be consistent
+                assert ios_deployment_target == target
+
+    return (ios_deployment_target, macos_deployment_target, weak_framework)
+
+
+def get_ort_version():
```
edgchen1, line 166: nit: maybe name it `get_ort_genai_version`?

tools/ci_build/github/apple/c/c.podspec.template
```diff
@@ -0,0 +1,29 @@
...
+  s.license	            = { :type => "MIT", :file => "@LICENSE_FILE@" }
+  s.author              = {"ONNX Runtime" => "onnxruntime@microsoft.com" }
+  s.source              = { :http => 'file:///http_source_placeholder' }
+
+  s.ios.deployment_target = "@IOS_DEPLOYMENT_TARGET@"
+  s.osx.deployment_target = "@MACOSX_DEPLOYMENT_TARGET@"
+
+  s.vendored_frameworks = "@ORTGENAI_C_FRAMEWORK@"
+  s.static_framework    = true
+  s.framework           = ['ImageIO', 'CoreGraphics']
```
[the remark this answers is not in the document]
skyline75489: yes ort-genai requires ImageIO and CoreGraphics for image decoding.

src/objectivec/include/ort_genai_objc.h
skyline75489: alright i'll document this behaviour then. thanks.

src/objectivec/include/ort_genai_objc.h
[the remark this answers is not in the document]
skyline75489: Changed it to `size_t(-1)` and documented this behaviour.

tools/ci_build/github/apple/objectivec/objc.podspec.template
skyline75489: removed

tools/ci_build/github/apple/package_assembly_utils.py
skyline75489: done

src/objectivec/oga_sequences.mm
```diff
@@ -0,0 +1,50 @@
...
+  try {
+    _sequences = OgaSequences::Create();
+    return self;
+  }
+  OGA_OBJC_API_IMPL_CATCH_RETURNING_NULLABLE(error)
+}
+
+- (size_t)getCountWithError:(NSError**)error {
+  try {
+      return _sequences->Count();
```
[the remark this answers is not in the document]
skyline75489: fixed

src/objectivec/oga_named_tensors.mm
```diff
@@ -0,0 +1,22 @@
...
+#import "oga_internal.h"
+#import "ort_genai_objc.h"
+
+@implementation OGANamedTensors {
+  std::unique_ptr<OgaNamedTensors> _tensor;
+}
+
+- (instancetype)initWithCXXPointer:(std::unique_ptr<OgaNamedTensors>)ptr
+{
+  _tensor = std::move(ptr);
```
edgchen1, line 14: general: add `if ((self = [super init]) == nil) { return nil; }` to initializers

also in `OGASequences`

src/objectivec/oga_internal.h
```diff
@@ -0,0 +1,69 @@
...
+@interface OGATokenizer ()
+
+- (const OgaTokenizer&)CXXAPIOgaTokenizer;
+
+@end
+
+@interface OGASequences ()
+
+- (nullable instancetype)initWithError:(NSError**)error;
+- (instancetype)initWithCXXPointer:(std::unique_ptr<OgaSequences>)ptr;
```
edgchen1, line 25: should these also have `NS_DESIGNATED_INITIALIZER`?

src/objectivec/oga_multi_modal_processor.mm
```diff
@@ -0,0 +1,48 @@
...
+  OGA_OBJC_API_IMPL_CATCH_RETURNING_NULLABLE(error)
+}
+
+- (nullable OGANamedTensors*)processImages:(NSString*)prompt
+                                    images:(OGAImages*)images
+                                     error:(NSError**)error {
+  try {
+    OGANamedTensors* result = [[OGANamedTensors alloc]
+        initWithCXXPointer:_processor->ProcessImages([prompt UTF8String],
+                                                        [images CXXAPIOgaImages])];
```
edgchen1, line 30: nit: please run formatter on all objc code

src/objectivec/include/ort_genai_objc.h
```diff
@@ -0,0 +1,360 @@
...
+ * @param length The number of tokens to rewind.
+ * @param error Optional error information set if an error occurs.
+ */
+- (BOOL)rewindTo:(NSUInteger)length error:(NSError**)error;
+
+/**
+ * Generate next token
+ * @param error Optional error information set if an error occurs.
+ */
+- (BOOL)generateNextTokenWithError:(NSError**)error NS_SWIFT_NAME(generateNextToken());
```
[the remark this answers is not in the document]
skyline75489: thanks for the catch!

src/objectivec/oga_internal.mm
```diff
@@ -0,0 +1,40 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#import "oga_internal.h"
+
+@implementation OGAInt32Span {
+  const int32_t* _ptr;
```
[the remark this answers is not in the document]
skyline75489: After consideration, I decided to remove `OGAInt32Span` and use raw `const int32_t*` together a length `size_t` instead. The functionality remains the same.

src/objectivec/test/ort_genai_api_test.mm
```diff
@@ -0,0 +1,139 @@
...
+
+    NSError *error = nil;
+    BOOL ret = NO;
+    OGAModel* model = [[OGAModel alloc] initWithPath:[ORTGenAIAPITest getModelPath] error:&error];
+    ORTAssertNullableResultSuccessful(model, error);
+
+    OGAGeneratorParams *params = [[OGAGeneratorParams alloc] initWithModel:model error:&error];
+    ORTAssertNullableResultSuccessful(params, error);
+
+    [params setSearchOption:@"max_length" doubleValue:max_length error:&error];
```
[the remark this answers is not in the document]
skyline75489: yes for sure.

.pipelines/stages/jobs/macos-ios-cocoapods-packaging-job.yml
```diff
@@ -0,0 +1,41 @@
+parameters:
+- name: build_config
+  type: string
+  default: 'release'
```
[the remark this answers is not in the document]
skyline75489: removed.

src/objectivec/oga_generator.mm
```diff
@@ -0,0 +1,92 @@
...
+    return YES;
+  }
+  OGA_OBJC_API_IMPL_CATCH_RETURNING_BOOL(error)
+}
+
+- (BOOL)appendTokens:(NSArray<NSNumber*>*)tokens error:(NSError**)error {
+  std::vector<int32_t> cxxTokens;
+  for (NSNumber* object in tokens) {
+    cxxTokens.push_back([object intValue]);
+  }
```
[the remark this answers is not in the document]
skyline75489: done

src/objectivec/include/ort_genai_objc.h
```diff
@@ -0,0 +1,360 @@
...
+/**
+ * Appends token sequences to the generator.
+ * @param tokens The tokens to append.
+ * @param error Optional error information set if an error occurs.
+ */
+- (BOOL)appendTokens:(NSArray<NSNumber*>*)tokens error:(NSError**)error;
+
+/**
+ * Rewinds the generator by the specified number of tokens.
+ * @param length The number of tokens to rewind.
```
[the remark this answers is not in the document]
skyline75489: I copied the comments from the Java binding:

https://github.com/microsoft/onnxruntime-genai/blob/c5745fd6d91a2ba2d94c6a52f6255652dde02508/src/java/src/main/java/ai/onnxruntime/genai/Generator.java#L102

Now I wonder which one of those docs are correct. @aciddelgado Could you help take a look?

src/objectivec/include/ort_genai_objc.h
```diff
@@ -0,0 +1,360 @@
...
+ * @param error Optional error information set if an error occurs.
+ */
+- (BOOL)rewindTo:(NSUInteger)length error:(NSError**)error;
```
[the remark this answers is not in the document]
skyline75489: Use `size_t` instead.

src/objectivec/oga_images.mm
```diff
@@ -0,0 +1,35 @@
...
+      cpp_paths.push_back([path UTF8String]);
+    }
+
+    _images = OgaImages::Load(cpp_paths);
+    return self;
+  }
+  OGA_OBJC_API_IMPL_CATCH_RETURNING_NULLABLE(error)
+}
+
+- (OgaImages*)CXXAPIOgaImages {
```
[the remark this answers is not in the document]
skyline75489: Fixed.

src/objectivec/oga_multi_modal_processor.mm
skyline75489: Clang-formated, but the result is still pretty much the same with this file. I think it's because of line length limitation.

src/objectivec/oga_internal.h
skyline75489: sure why not.

src/objectivec/error_utils.mm
```diff
@@ -0,0 +1,34 @@
...
+NS_ASSUME_NONNULL_BEGIN
+
+NSString* const kOgaErrorDomain = @"onnxruntime-genai";
+const int kOgaErrorCode = 0x0A;
+
+void OGASaveCodeAndDescriptionToError(int code, const char* descriptionCstr, NSError** error) {
+  if (!error) return;
+
+  NSString* description = [NSString stringWithCString:descriptionCstr
+                                             encoding:NSASCIIStringEncoding];
```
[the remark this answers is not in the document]
skyline75489: I just copied from ORT code. And yeah maybe this should be UTF8. do you want me to change it?

src/objectivec/oga_named_tensors.mm
skyline75489: fixed.

src/objectivec/error_utils.mm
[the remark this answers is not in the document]
edgchen1: yeah, let's change it. I created a corresponding PR for ORT:
https://github.com/microsoft/onnxruntime/pull/22982

src/objectivec/include/ort_genai_objc.h
```diff
@@ -309,13 +320,22 @@ typedef NS_ENUM(NSInteger, OGAElementType) {
...
+ */
+- (nullable const int32_t*)sequenceDataAtIndex:(size_t)index
+                                         error:(NSError**)error;
+
+/**
+ * Retrieve the sequence count at the given index.
  * @param index The index needed.
  * @param error Optional error information set if an error occurs.
- * @return The sequence at the given index, or nil if an error occurs.
+ * @return The sequence length at the given index, or nil if an error occurs.
```
edgchen1, line 335:
```suggestion
 * @return The sequence count at the given index, or size_t(-1) if an error occurs.
```

"count" for consistency with `OGASequences sequenceCountAtIndex`

src/objectivec/include/ort_genai_objc.h
```diff
@@ -187,13 +189,22 @@ typedef NS_ENUM(NSInteger, OGAElementType) {
...
-- (nullable OGAInt32Span*)sequenceAtIndex:(size_t)index
-                                    error:(NSError**)error;
+- (nullable const int32_t*)sequenceDataAtIndex:(size_t)index
+                                         error:(NSError**)error;
+
+/**
+ * Retrieve the sequence count at the given index.
+ * @param index The index needed.
+ * @param error Optional error information set if an error occurs.
+ * @return The sequence count at the given index, or nil if an error occurs.
```
edgchen1, line 204:
```suggestion
 * @return The sequence count at the given index, or size_t(-1) if an error occurs.
```

src/objectivec/test/ort_genai_api_test.mm
```diff
@@ -0,0 +1,138 @@
...
+
+    // check for the 1st token generation
+    // full logits has shape [2, 1, 1000]. Sample 1 for every 200 tokens and the expected sampled logits has shape [2, 1, 5]
+    std::vector<float> expected_sampled_logits_token_gen{0.03742531f, -0.05752287f, 0.14159015f, 0.04210977f, -0.1484456f,
+                                                         0.3041716f, -0.08701379f, -0.03778192f, 0.07471392f, -0.02049096f};
+
+    OGATensor* token_gen_logits_ptr = [generator getOutput:@"logits" error:&error];
+    ORTAssertNullableResultSuccessful(token_gen_logits_ptr, error);
+
+    auto token_gen_logits = static_cast<float*>([token_gen_logits_ptr getDataPointerWithError:&error]);
```
edgchen1, line 126: also check `error` here

src/objectivec/test/ort_genai_api_test.mm
```diff
@@ -0,0 +1,138 @@
...
+                                                      0.30323744f, 0.0545997f, 0.03894716f, 0.11702324f, 0.0410665f,
+                                                      -0.12675379f, -0.04443946f, 0.14492269f, 0.03021223f, -0.03212897f,
+                                                      0.29694548f, 0.00955007f, 0.0430819f, 0.10063869f, 0.0437237f,
+                                                      0.27329233f, 0.00841076f, -0.1060291f, 0.11328877f, 0.13369876f,
+                                                      -0.04699047f, 0.17915794f, 0.20838135f, 0.10888482f, -0.00277808f,
+                                                      0.2938929f, -0.10538938f, -0.00226692f, 0.12050669f, -0.10622668f};
+
+    OGATensor* prompt_logits_ptr = [generator getOutput:@"logits" error:&error];
+    ORTAssertNullableResultSuccessful(prompt_logits_ptr, error);
+    auto prompt_logits = static_cast<float*>([prompt_logits_ptr getDataPointerWithError:&error]);
```
edgchen1, line 104: also check `error` here

src/objectivec/test/ort_genai_api_test.mm
```diff
@@ -0,0 +1,138 @@
...
+#import <array>
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface OGATensor (Testing)
+
+- (nullable instancetype)initWithDataPointer:(void*)data
+                                       shape:(NSArray<NSNumber*>*)shape
+                                        type:(OGAElementType)elementType
+                                       error:(NSError**)error;
```
edgchen1, line 18: instead of declaring `initWithDataPointer:shape:type:error:` again here, should we just make it part of OGATensor's public interface?

skyline75489: In [C#](https://github.com/microsoft/onnxruntime-genai/blob/d9cb8548a576d0d5f43cb2214188268209cafd54/src/csharp/Tensor.cs#L38) and [Java](https://github.com/microsoft/onnxruntime-genai/blob/d9cb8548a576d0d5f43cb2214188268209cafd54/src/java/src/main/java/ai/onnxruntime/genai/Tensor.java#L75) bindings, the ctor of `OGATensor` is intentionally made `internal`. I think the reason is that it's not possible for users ton construct a native handle of `OGATensor`. I think in Objective-C I should follow the same fashion.

src/objectivec/test/ort_genai_api_test.mm
skyline75489: These two assertions are tricky because `ORTAssertNullableResultSuccessful` expects an `id`, instead of native C++ pointer. Using it will cause error:

```text
Cannot initialize a variable of type '__strong id' with an lvalue of type 'float *'
```

Do you want me to add another one for native C++ pointer? Or maybe just use `ORTAssertNullableResultSuccessful` and checks for `prompt_logits != nullptr`?

src/objectivec/include/ort_genai_objc.h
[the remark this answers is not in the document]
skyline75489: This is fixed in #1118 and I'll modify the Objective-C code accordingly. Thanks for catching this!

src/objectivec/test/ort_genai_api_test.mm
edgchen1: oh, `XCTAssertNotNil` expects `id`.

how about `XCTAssertNil(error)` and `XCTAssertNotEqual(prompt_logits, nullptr)`? up to you if you want to add another helper macro.

src/objectivec/test/ort_genai_api_test.mm
edgchen1: ok, if there's a good reason to not make it public then we shouldn't. but as a follow up question, why is there a C API to create a tensor from a buffer?

https://github.com/microsoft/onnxruntime-genai/blob/4b12d57dfbc0097bf3f1b9c7660e2e79f02ff003/src/ort_genai_c.h#L440-L449

I wonder if the reason that Java and C# don't expose it is language-specific, e.g., because it is difficult to access native memory from a managed language?

skyline75489: You're right. I was mistaken. There're **two** ctors in `OGATensor`. One is using raw buffer (`void*`) and another is using a `OgaTensor*`. This one `initWithDataPointer:shape:type:error:` is the raw buffer one and should made public, just like in C, C# and Java. The intentionally made internal one is the second one using a `OgaTensor*` as parameter, which is not possible to construct from outside this library.

src/objectivec/test/ort_genai_api_test.mm
skyline75489: Using `XCTAssertNil(error)` and `XCTAssertNotEqual(prompt_logits, nullptr)` seems to be working fine.

src/objectivec/test/ort_genai_api_test.mm
skyline75489: Fixed

src/objectivec/test/ort_genai_api_test.mm
skyline75489: Fixed

src/objectivec/include/ort_genai_objc.h
```diff
@@ -0,0 +1,384 @@
...
+ * @return The decoding result, or nil if an error occurs.
+ */
+- (nullable NSString*)decode:(int32_t)token
+                       error:(NSError**)error;
+@end
+
+/**
+ * Wraps a raw int32_t pointer and its size. Represents a data sequence.
+ */
+@interface OGAInt32Span : NSObject
```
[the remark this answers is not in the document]
skyline75489: Yes for sure. I *thought* I removed it but guess I forgot.

src/objectivec/oga_multi_modal_processor.mm
```diff
@@ -0,0 +1,48 @@
...
+  try {
+    OGANamedTensors* result = [[OGANamedTensors alloc]
+        initWithCXXPointer:_processor->ProcessImages([prompt UTF8String],
+                                                     &[images CXXAPIOgaImages])];
+    return result;
+  }
+  OGA_OBJC_API_IMPL_CATCH_RETURNING_NULLABLE(error)
+}
+
+- (nullable NSString*)decode:(OGAInt32Span*)data error:(NSError**)error {
```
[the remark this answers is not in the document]
skyline75489: It is now declared in `ort_genai_obc.h`. Thanks for catching it.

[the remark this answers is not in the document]
skyline75489: I plan to have a documentation overhaul after #1107 is done. So don't worry about the missing documentation. Just need  to figure out a better way to maintain them.

edgchen1 (approved):
looks good. thanks for adding the objective-c API!

PR #711, opened 2024-07-18
Add clean option to build script

natke:


build.py
```diff
@@ -299,7 +299,7 @@ def _validate_cmake_args(args: argparse.Namespace):
 
 def _validate_args(args: argparse.Namespace):
     # default to all 3 stages
-    if not args.update and not args.build and not args.test:
+    if not args.update and not args.build and not args.test and not args.clean:
```
baijumeswani, line 302: I think args.clean should override all other args. For example, if user provided --clean --build, we should probably only clean (or raise an error or warning).
Or alternatively, clean first and then build again? We probably should document the expected behavior.

natke: Agreed. Will fix

natke: Is clean supposed remove the build directory for the specified or default config? doesn't appear to be doing so

natke: Apparently cmake clean only removes the cmake generated files. I wanted to remove all of the previous build artifacts

natke: Updated to remove the build folder as well

build.py
```diff
@@ -550,10 +556,13 @@ def clean(args: argparse.Namespace, env: dict[str, str]):
     """
     Clean the build output.
     """
-    log.info("Cleaning targets")
-    cmd_args = [str(args.cmake), "--build", str(args.build_dir), "--config", args.config, "--target", "clean"]
+    log.info("Cleaning cmake targets")
+    cmd_args = [str(args.cmake_path), "--build", str(args.build_dir), "--config", args.config, "--target", "clean"]
```
baijumeswani, line 560: Do we need to clean if we are deleting the build directory? If we are deleting the build directory, it offers no advantage to the user since doing

```
rm -rf build/
```

or 

```
rmdir /s /q "build"
```

is probably easier for the user. Maybe we can just let clean be the same as cleaning the cmake output dir.

baijumeswani:
Could we merge main and update this PR?

natke:
Closing as we don't think it adds much value

Issue #715, opened 2024-07-22
Warn users if image processor is not available


skyline75489:
Introduced by #478, the following code:

https://github.com/microsoft/onnxruntime-genai/blob/60269a0f7bf8e26bab971d64713a7a2eab1d13d4/src/models/model.cpp#L558

Skips the creation of `image_processor` when no vision field is found in config. This leads to `image_processor` being nullptr. If users then try to use `ProcessImage`, the code will immediately crash.

We should probably come up with a better warning when this happens (instead of crashing).

PR #717, opened 2024-07-22
Avoid crash when non MultiModal model tries to process images

baijumeswani:
Addresses https://github.com/microsoft/onnxruntime-genai/issues/715 and other user complaints.

### DIFF
--- a/src/models/model.cpp
+++ b/src/models/model.cpp
@@ -556,11 +556,7 @@ std::unique_ptr<OrtValue> Model::ExpandInputs(std::unique_ptr<OrtValue>& input, 
 MultiModalProcessor::MultiModalProcessor(Config& config, const SessionInfo& session_info)
     : tokenizer_{std::make_shared<Tokenizer>(config)} {
   if (config.model.type == "phi3v") {
-    if (!config.model.vision.filename.empty()) {
-      image_processor_ = std::make_shared<ImageProcessor>(config, session_info);
-    } else {
-      throw std::runtime_error("ImageProcessor cannot be created. Vision model not found.");
-    }
+    image_processor_ = std::make_shared<ImageProcessor>(config, session_info);
   } else {
     throw std::runtime_error("MultiModalProcessor cannot be created. Expected a multimodal model. Actual: " + config.model.type);
   }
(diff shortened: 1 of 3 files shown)


src/models/model.cpp
```diff
@@ -555,8 +555,14 @@ std::unique_ptr<OrtValue> Model::ExpandInputs(std::unique_ptr<OrtValue>& input,
 
 MultiModalProcessor::MultiModalProcessor(Config& config, const SessionInfo& session_info)
     : tokenizer_{std::make_shared<Tokenizer>(config)} {
-  if (!config.model.vision.filename.empty()) {
-    image_processor_ = std::make_shared<ImageProcessor>(config, session_info);
+  if (config.model.type == "phi3v") {
+    if (!config.model.vision.filename.empty()) {
+      image_processor_ = std::make_shared<ImageProcessor>(config, session_info);
+    } else {
+      throw std::runtime_error("ImageProcessor cannot be created. Vision model not found.");
```
skyline75489, line 562: Is `MultiModalProcessor` designed to work without vision capability? If so, this should probably be just a warning.

src/ort_genai_c.cpp
```diff
@@ -370,6 +370,9 @@ OgaResult* OGA_API_CALL OgaProcessorProcessImages(const OgaMultiModalProcessor*
   OGA_TRY
   auto& processor = *reinterpret_cast<const Generators::MultiModalProcessor*>(p);
   auto* images = images_p ? reinterpret_cast<const Generators::Images*>(images_p) : nullptr;
```
skyline75489, line 372: Maybe also throw runtime exception when `images` is `nullptr`? I mean...what we gonna do with a nullptr `images`, anyway.

src/models/model.cpp
baijumeswani: MultiModalProcessor is designed to support vision capabilities for now. Audio capabilities will be added in the near future.
What capabilities the MultiModalProcessor supports is model type dependent. For example, phi3-vision should support creating the image processor. The whisper model should create the audio processor. And maybe other model types will require other types of processors in the future.

Maybe I can remove the check for the filename being empty since it is not used here. But certainly, if the model is phi3v, the image processor should be created.

src/ort_genai_c.cpp
baijumeswani: The phi3-vision model works with no images. It is one of the features. Essentially, if no image is provided, it works like a regular decoder only model. But it still requires the processing using ProcessImages.

skyline75489 (approved):

PR #723, opened 2024-07-25
Initial support for QNN EP

skyline75489:
Added support for QNN EP and related session options


src/config.cpp
```diff
@@ -84,6 +84,12 @@ struct SessionOptions_Element : JSON::Element {
       v_.enable_cpu_mem_arena = value;
     else if (name == "enable_mem_pattern")
       v_.enable_mem_pattern = value;
+    else if (name == "disable_cpu_ep_fallback")
```
skyline75489, line 87: These options are used in the reference script.

src/java/src/main/native/utils.h
```diff
@@ -30,19 +31,24 @@ bool ThrowIfError(JNIEnv* env, OgaResult* result);
...
   CString(JNIEnv* env, jstring str)
-      : env_{env}, str_{str}, cstr{env->GetStringUTFChars(str, /* isCopy */ nullptr)} {
+      : env_{env}, str_{str}, cstr{env->GetStringUTFChars(str, /* isCopy */ nullptr)}, len_{env->GetStringUTFLength(str)} {
   }
 
   const char* cstr;
 
   operator const char*() const { return cstr; }
 
+  std::string utf8String() {
```
skyline75489, line 41: Strictly speaking, this is Java modified version of UTF8. It'll work for ASCII. But I'm not sure how good it is comparing to "normal" UTF8.

src/java/src/main/java/ai/onnxruntime/genai/Model.java
```diff
@@ -6,6 +6,14 @@
 public final class Model implements AutoCloseable {
   private long nativeHandle;
 
+  public Model(String modelPath, String qnnLibSearchPath) throws GenAIException {
```
skyline75489, line 9: I don't feel like adding public interfaces like this. But I don't have better ideas. @skottmckay  any thought?

baijumeswani (approved):

src/models/model.cpp
```diff
@@ -388,6 +422,13 @@ void Model::CreateSessionOptions() {
 
       device_type_ = DeviceType::DML;  // We use a DML allocator for input/output caches, but other tensors will use CPU tensors
 #endif
+    } else if (provider_options.name == "qnn") {
+      std::unordered_map<std::string, std::string> opts;
+      for (auto& option : provider_options.options) {
+        opts.emplace(option.first, option.second);
+      }
+
+      ort_options.AppendExecutionProvider("QNN", opts);
```
skyline75489, line 431: From ORT doc, currently supported providers includes:

   *   QNN
   *   SNPE
   *   XNNPACK

Issue #693, opened 2024-07-12
Consider removing direct link-time dependency on ORT on Linux/macOS


skyline75489:
# Introduction

During the work of #687, I notice some issues that blocks GenAI NuGet from working on Linux. After investigation and discussion with @baijumeswani , I'll like to propose removing direct link-time dependency on ORT, to improve the overall dynamic library loading scenarios. @skottmckay has already implemented it on Android platform. This proposal is to further extend it on Linux/macOS.

# The problem with link-time dependency

Again, @skottmckay has a detailed writeup at https://github.com/microsoft/onnxruntime-genai/blob/e41fb2c1430693157c3cea70d3bdcc207ae84ef3/src/models/onnxruntime_api.h#L89 

I suggest checking this out first to have a basic understanding of the issue.

TLDR: ort-genai C++ library depends on a specific version of ORT. Simply having `libonnxruntime.so` isn't enough, It needs specifically `libonnxruntime.so.1.18.0`. This is the `ldd` output:

```text
ldd -v libonnxruntime-genai.so
        linux-vdso.so.1 (0x00007ffecc25e000)
        libonnxruntime.so.1.18.0 => not found
        libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007f59e6ab0000)
        libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f59e6aab000)
......
```

Traditionally, this should be handled by symlinks, if those libraries are installed system-wide. But we need ort-genai to be distributed in wheels and NuGets.

## Wheel scenario

To make GenAI work with direct linking against ORT, a exact copy of `libonnxruntime.so` currently exists and renamed to `libonnxruntime.so.1.18.0`:

```text
(genai) skyline@workstation ~/miniconda3/envs/genai/lib/python3.10/site-packages/onnxruntime_genai
>ls -al
total 33804
drwxr-xr-x  4 skyline skyline     4096 Jul 12 11:29 .
drwxr-xr-x 43 skyline skyline     4096 Jul 12 12:08 ..
-rw-r--r--  1 skyline skyline      711 Jul 12 11:29 __init__.py
drwxr-xr-x  2 skyline skyline     4096 Jul 12 11:29 __pycache__
-rw-r--r--  1 skyline skyline      617 Jul 12 11:29 _dll_directory.py
-rw-r--r--  1 skyline skyline 15285336 Jul 12 11:29 libonnxruntime.so
[...]

PR #724, opened 2024-07-25
Use dlopen in favor of direct dependency of ORT on Linux

skyline75489:
Implement #693 on Linux. 

Fixed CI breakage caused by https://github.com/microsoft/onnxruntime/pull/21339, which modifies the SONAME to `libonnxruntime.so.1`. But the ORT lib bundled in ORT 1.19(nightly) is named `libonnxruntime.so.1.19.0`.

Related: https://github.com/microsoft/onnxruntime/issues/21281

### DIFF
--- a/.github/workflows/linux-cpu-arm64-build.yml
+++ b/.github/workflows/linux-cpu-arm64-build.yml
@@ -95,4 +95,4 @@ jobs:
         run: |
           docker run --rm \
           --volume $GITHUB_WORKSPACE:/onnxruntime_src \
-          -w /onnxruntime_src ort_genai_linux_arm64_gha bash -c "/onnxruntime_src/build/cpu/test/unit_tests"   
+          -w /onnxruntime_src ort_genai_linux_arm64_gha bash -c "LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/onnxruntime_src/build/cpu/ /onnxruntime_src/build/cpu/test/unit_tests"
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -84,7 +84,7 @@ target_link_directories(onnxruntime-genai PRIVATE ${ORT_LIB_DIR})
 
 # we keep the shared libraries disconnected on Android as they will come from separate AARs and we don't want to force
 # the ORT version to match in both.
-if(NOT ANDROID)
+if(NOT (ANDROID OR LINUX))
 target_link_libraries(onnxruntime-genai PRIVATE ${ONNXRUNTIME_LIB})
 endif()
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -84,7 +84,7 @@ target_link_directories(onnxruntime-genai PRIVATE ${ORT_LIB_DIR})
 
 # we keep the shared libraries disconnected on Android as they will come from separate AARs and we don't want to force
 # the ORT version to match in both.
-if(NOT (ANDROID OR LINUX))
+if(NOT (CMAKE_SYSTEM_NAME STREQUAL "Android" OR CMAKE_SYSTEM_NAME STREQUAL "Linux"))
 target_link_libraries(onnxruntime-genai PRIVATE ${ONNXRUNTIME_LIB})
 endif()
 
--- a/src/models/onnxruntime_api.h
+++ b/src/models/onnxruntime_api.h
@@ -113,7 +113,7 @@ inline const OrtApi* api{};
 
 #if defined(__linux__)
 inline void* LoadDynamicLibraryIfExists(const std::string& path) {
-  LOG_INFO("Attempting to dlopen %s native library", path.c_str());
+  LOG_INFO("Attempting to dlopen %s", path.c_str());
   void* ort_lib_handle = dlopen(path.c_str(), RTLD_NOW | RTLD_LOCAL);
   if (ort_lib_handle == nullptr) {
     return nullptr;
@@ -201,7 +201,6 @@ inline void InitApi() {
   void* ort_lib_handle = LoadDynamicLibraryIfExists(path);
 
 #if !defined(__ANDROID__)
-  // Search for pip installation
   if (ort_lib_handle == nullptr) {
     const std::array<std::string, 4> target_libraries = {
         std::string("libonnxruntime.so"),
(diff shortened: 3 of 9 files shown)


CMakeLists.txt
```diff
@@ -84,7 +84,7 @@ target_link_directories(onnxruntime-genai PRIVATE ${ORT_LIB_DIR})
 
 # we keep the shared libraries disconnected on Android as they will come from separate AARs and we don't want to force
 # the ORT version to match in both.
-if(NOT ANDROID)
+if(NOT (CMAKE_SYSTEM_NAME STREQUAL "Android" OR CMAKE_SYSTEM_NAME STREQUAL "Linux"))
```
baijumeswani, line 87: Why not do this for windows also? Then we have only one solution instead of 2.

skyline75489: Mostly because `dlopen` isn't available on Windowsl. We'll need to use `LoadLibraryA` or something like that. You're right, one solution does sounds better.

baijumeswani: Let's do it in another PR. I think this PR is good to go. But we can wait for @snnn to share his thoughts as well.

baijumeswani (approved):

CMakeLists.txt
skyline75489: Technically this PR lacks more fine-grained searching for libraries. In #662 you added the search path like this for wheel scenario:

```cpp
 set_property(TARGET python APPEND_STRING PROPERTY LINK_FLAGS " -Xlinker -rpath=\\$ORIGIN/../onnxruntime/capi/")
```

With this PR, we'll need to search for `../onnxruntime/capi` in C++. While waiting for Changming, I'll add the necessary searching accroding to #662 today.

src/models/onnxruntime_api.h
```diff
@@ -108,49 +198,43 @@ inline void InitApi() {
...
-  }
-
-  ort_api_base_fn = (OrtApiBaseFn)dlsym(ort_lib_handle, "OrtGetApiBase");
-  if (ort_api_base_fn == nullptr) {
-    __android_log_assert("ort_api_base_fn != nullptr", "GenAI", "OrtGetApiBase not found");
+    const std::array<std::string, 4> target_libraries = {
+        std::string("libonnxruntime.so"),
+        std::string("libonnxruntime.so.1.18.0"),
+        std::string("libonnxruntime.so.1.19.0"),
+        std::string("libonnxruntime.so.1.20.0")};
```
skyline75489, line 210: To reviewer: the ORT lib bundled in ort-nightly package has version number. This seems to be the only way to have forward compatibility. I don't like it, but I don't have better idea.

src/models/onnxruntime_api.h
```diff
@@ -68,24 +68,114 @@ p_session_->Run(nullptr, input_names, inputs, std::size(inputs), output_names, o
...
+
+#if !defined(__ANDROID__)  // RTLD_DI_ORIGIN not available on Android
+  char pathname[PATH_MAX];
+  dlinfo((void*)ort_lib_handle, RTLD_DI_ORIGIN, &pathname);
+  LOG_INFO("Loaded native library at %s", pathname);
+#endif
+  return ort_lib_handle;
+}
+
+inline std::string GetCurrentModuleDir() {
```
skyline75489, line 130: This is to find the location of the current ort-genai library, in order to locate the `site-pacakges` and `onnxruntime` folder in it.

.github/workflows/linux-cpu-arm64-build.yml
```diff
@@ -95,4 +95,4 @@ jobs:
         run: |
           docker run --rm \
           --volume $GITHUB_WORKSPACE:/onnxruntime_src \
-          -w /onnxruntime_src ort_genai_linux_arm64_gha bash -c "/onnxruntime_src/build/cpu/test/unit_tests"   
+          -w /onnxruntime_src ort_genai_linux_arm64_gha bash -c "LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/onnxruntime_src/build/cpu/ /onnxruntime_src/build/cpu/test/unit_tests"
```
skyline75489, line 98: Unlike Python wheel, we can't reliably determine where to find ORT for native C++ binaries like `unit_tests`. Fortunately , `dlopen` also respect `LD_LIBRARY_PATH`.

src/models/onnxruntime_api.h
```diff
@@ -218,12 +218,14 @@ inline void InitApi() {
...
-    // Search for pip installation
-    for (const std::string& lib_name : target_libraries) {
-      std::string pip_path{current_module_dir + "/../onnxruntime/capi/" + lib_name};
-      ort_lib_handle = LoadDynamicLibraryIfExists(pip_path);
-      if (ort_lib_handle != nullptr) {
-        break;
+    if (ort_lib_handle == nullptr) {
+      // Search for pip installation
+      for (const std::string& lib_name : target_libraries) {
+        std::string pip_path{current_module_dir + "/../onnxruntime/capi/" + lib_name};
```
snnn, line 224: ../onnxruntime probably doesn't exist. 
For example, some python modules may get installed in a system location like /usr/lib, then some get installed in a user location that is in the user's home folder ($HOME).

skyline75489: Yeah that's a thing. I was just copying what #662 is doing. There's certainly more than one place that onnxruntime can reside.

snnn: There might be a better solution.

Search RTLD_GLOBAL in 

https://github.com/microsoft/onnxruntime/blob/bbbaef3fa60f1ef1920af5104459251b42390135/cmake/onnxruntime_python.cmake#L320

You may turn it always ON, and let the onnxruntime_pybind11_state.so   expose all the symbols that libonnxrutime.so has. Then, in theory in another package(ONNXRuntime-GENAI) you should be able to directly use the symbols, without doing any dlopen. Because, RTLD_GLOBAL  means globally visible. Here globally means process-wide.

snnn (approved):
You may merge it first then continue improving it.

src/models/onnxruntime_api.h
skyline75489: The SONAME is now `libonnxruntime.so.1`, but the file name is still `libonnxruntime.so.1.19.0`.

src/models/onnxruntime_api.h
skyline75489: The thing is, `libonnxruntime-genai.so` is loaded *before* any ORT libs. `RTLD_GLOBAL` will only expose symbol *after* ORT libs are loaded.

A better way might be directly interacting with CPython API and asked it to find `onnxruntime` for us.

baijumeswani: > A better way might be directly interacting with CPython API and asked it to find onnxruntime for us.

Not sure if we have the same idea, but here is one solution. We can use on linux to find the path of ort shared so. From python before importing onnxruntime_genai pybind layer, we can do something like this:

```py
import importlib.util
import ctypes
import glob

ort_package = importlib.util.find_spec("onnxruntime")
ort_package_path = ort_package.submodule_search_locations[0]
ort_lib_path = glob.glob(os.path.join(ort_package_path, "capi", "libonnxruntime.so*"))[0]
_ = ctypes.CDLL(ort_lib_path)

import onnxruntime_genai
```

We do something similar in windows by doing `os.add_dll_directory`. Since there is no linux equivalent, best is to load the shared lib and then the system will not load it again.

Changes in this PR: https://github.com/microsoft/onnxruntime-genai/pull/731

baijumeswani:
Let's merge this to unblock other PRs. And then revisit the search strategy.

PR #729, opened 2024-07-29
Support for pipeline decoder model

baijumeswani:
This pull request adds support for scenarios where the decoder only model is split into multiple smaller onnx models.

onnxruntime-genai will execute these models in a pipeline fashion. The pipeline is expected to be defined in the genai config.

### Update to the GenAI config

The outputs of the previous model in the pipeline can be fed into inputs of the next model in the pipeline. Configurations exposed to the user:
- `filename`: file name of the onnx model in the pipeline. Required field.
- `session_options`: Each `PipelineModel` can define its own `SessionOptions`. In essence, that pipeline model session will use this session options to execute. By not provided, the default session options are used.
- `run_on_first_token_generation`: Whether that pipeline model should be run on the first token generation. Default: true 
- `run_on_nth_token_generation`: Whether that pipeline model should be run on nth token generation (`n != 1`). Default: true
- `output_names_forwarder`: In case output names from the previous model do not align with input names of the following model, this mapping can be defined in the config.
- `inputs`: input names of the pipeline model. Required field.
- `outputs`: output names of the pipeline model. Required field.

### Example pipeline

Here we have split the decoder model in 3 parts:
1. Embeddings
2. Transformer
3. Model head

```json
"pipeline": [
    {
        "embedding": {
            "filename": "phi-3-embedding.onnx",
            "inputs": [
                "input_ids"
            ],
            "outputs": [
                "inputs_embeds"
            ]
        },
        "transformer_model": {
            "filename": "phi-3-transformer.onnx",
            "inputs": [
                "inputs_embeds",
                "past_keys_0",
                "past_values_0",
                "..."
            ],
            "outputs": [
                "transformer_output",
                "present_keys_0",
                "present_values_0",
                "..."
            ],
        },
        "transformer_head": {
            "filename": "phi-3-transformer-head.onnx",
            "inputs": [
                "transformer_output"
            ],
            "outputs": [
                "logits"
            ]
        }
    }
]
```

In the above example, the outputs of the embedding pipeline model are fed into the inputs of the transformer model. Similarly, the outputs of the transformer model are fed into the inputs of the transformer model head pipeline model.

### Assumptions and limitations

- Final model inputs and outputs are expected to be the same as is currently supported in decoder only models. No other inputs/outputs are managed by the pipeline model. Inputs/outputs managed by the pipeline model:
  - input_ids (input)
  - kv cache (input)
  - attention_mask (input)
  - logits (output)
  - kv cache (output)
- The managed inputs and outputs (listed above) must be allocated on the device where the search is expected to take place. i.e. the pipeline does not move/copy data from one device to another after a session runs.
- The intermediate (unmanaged inputs and outputs must reside on CPU) and the ort session is responsible to make copies to and from the device (-host) the session options is registered for.

### Where can this feature be used

- In scenarios where the user would like to run different parts of the model using different session options.
- In scenarios where it is hard to combine multiple smaller models into one big model (due to limitations of the model or device being used to run the model).

Co-authors: @edgchen1 @ajindal1


wangyems:
qq: is this for pipeline parallelism?

baijumeswani:
> qq: is this for pipeline parallelism?

No, the work here is not intended for pipeline parallelism. However, it could potentially be useful in pipeline parallelism. Sorry for the late response.

yufenglee:
could you please add an unit test to cover the pipeline?

src/models/model.cpp
```diff
@@ -271,167 +281,193 @@ void Model::InitDeviceAllocator([[maybe_unused]] OrtSession& session) {
...
-
-      if (LoadLibraryEx(directml_dll, nullptr, 0) == NULL) {
-        throw std::runtime_error("DirectML.dll not found");
-      }
-
-      auto dml_create_device1_fn = reinterpret_cast<decltype(&DMLCreateDevice1)>(GetProcAddress(smart_directml_dll.get(), "DMLCreateDevice1"));
-      THROW_LAST_ERROR_IF(!dml_create_device1_fn);
-      THROW_IF_FAILED(dml_create_device1_fn(dml_objects_.d3d12_device.Get(), DML_CREATE_DEVICE_FLAG_NONE, DML_FEATURE_LEVEL_5_0, IID_PPV_ARGS(&dml_device_)));
+      } else if (provider_options.name == "dml") {
+        if (!p_dml_api_) {
```
yufenglee, line 395: >if (!p_dml_api_) { [](http://example.com/codeflow?start=8&length=18)

why add this check? the p_dmp_api_ is set below.

baijumeswani: All the pipeline model session options are created by calling this lambda function multiple times (once for each pipeline session options).
However, this lambda function not only updates the session options passed in, it also sets members on `this` `Model`. Particularly, it sets the `device_type_` and the `p_dml_api_` on this. These members once created should not be updated again. So, I check if the members are already created. If not, we create them. If they are, we skip this block of code.

src/config.cpp
```diff
@@ -183,6 +183,101 @@ struct Outputs_Element : JSON::Element {
   Config::Model::Decoder::Outputs& v_;
 };
 
+struct StringArray_Element : JSON::Element {
+  explicit StringArray_Element(std::vector<std::string>& v) : v_{v} {}
+
+  void OnString(std::string_view name, std::string_view value) override {
+    v_.push_back(std::string(value));
```
RyanUnderhill, line 190: Is the std::string needed?
```suggestion
    v_.push_back(value);
```

src/config.cpp
```diff
@@ -183,6 +183,101 @@ struct Outputs_Element : JSON::Element {
...
+
+ private:
+  std::vector<std::string>& v_;
+};
+
+struct StringStringMap_Element : JSON::Element {
+  explicit StringStringMap_Element(std::unordered_map<std::string, std::string>& v) : v_{v} {}
+
+  void OnString(std::string_view name, std::string_view value) override {
+    v_[std::string(name)] = std::string(value);
```
RyanUnderhill, line 201: Same here:
```suggestion
    v_[std::string(name)] = value;
```

src/config.cpp
```diff
@@ -183,6 +183,101 @@ struct Outputs_Element : JSON::Element {
...
+  StringStringMap_Element output_names_forwarder_{v_.output_names_forwarder};
+};
+
+struct PipelineModelObject_Element : JSON::Element {
+  explicit PipelineModelObject_Element(std::vector<Config::Model::Decoder::PipelineModel>& v) : v_{v} {}
+
+  Element& OnObject(std::string_view name) override {
+    auto& model = v_.emplace_back();
+    model.model_id = name;
+    pipeline_model_elements_.push_back(PipelineModel_Element(model));
```
RyanUnderhill, line 260: Does this work?
```suggestion
    pipeline_model_elements_.emplace_back(model);
```

src/models/decoder_only_pipeline.cpp
```diff
@@ -0,0 +1,229 @@
...
+                      [](const auto& elem) { return !elem.name.empty(); })) {
+        InitDeviceAllocator(*sessions_.back());
+        device_allocator_created = true;
+      }
+    }
+  }
+
+  if (!device_allocator_created) {
+    // If the device allocator has not been created, it implies all
+    // sessions are configured torun on CPU.
```
RyanUnderhill, line 28:
```suggestion
    // sessions are configured to run on CPU.
```

src/models/model.h
```diff
@@ -31,7 +31,11 @@ struct State {
   virtual RoamingArray<float> Run(int current_length, RoamingArray<int32_t> next_tokens, RoamingArray<int32_t> next_indices = {}) = 0;
   virtual const CapturedGraphInfo* GetCapturedGraphInfo() const { return nullptr; }
 
-  OrtValue* GetOutput(const char* name);
+  OrtValue* GetInput(const char* name);
+
+  virtual OrtValue* GetOutput(const char* name);
```
RyanUnderhill, line 36: @kunal-vaishnavi Just FYI since in your whisper changes you added 'GetSpecialOutput' that is called when 'GetOutput' has no match, this should work for your case also.

src/config.cpp
baijumeswani:
```
error: no matching function for call to 'std::vector<std::__cxx11::basic_string<char> >::push_back(std::string_view&)
```

Conversion from std::string_view to std::string must be explicit.

src/config.cpp
baijumeswani: Same as above.

test/test_models/pipeline-model/genai_config.json
```diff
@@ -0,0 +1,96 @@
...
+                            "attention_mask",
+                            "past_key_values.0.key",
+                            "past_key_values.0.value"
+                        ],
+                        "outputs": [
+                            "/model/layers.1/final_norm_layernorm/output_0",
+                            "present.0.key",
+                            "present.0.value"
+                        ],
+                        "session_options": {
```
kunal-vaishnavi, line 52: This doesn't have to be cleaned up now but having a `session_options` section in the `decoder` and then having another `session_options` section within the `transformer` section seems confusing. It may make more sense to have a `session_options` section for each `pipeline` entry (e.g. `embeddings`, `transformer`, `language_model_head`) and remove the global `session_options` section within `decoder`.

baijumeswani: The way it is currently structured is that the session options on the higher level is a fallback session options. In case users do not provide any session options on the pipeline model level, the fallback session options are used for that pipeline model session. The fallback session options also enables users to choose the device to use for search and sampling when the device used for pipelines is different compared to the one intended to be used for search and sampling.

Since the config structure is shared between the decoder-only models and the decoder-pipeline models, I cannot remove the session options on the decoder level. We could however change the structure to make it more user friendly, but that will break other published models.

src/config.cpp
```diff
@@ -147,6 +147,10 @@ struct Inputs_Element : JSON::Element {
       v_.cross_past_key_names = value;
     } else if (name == "cross_past_value_names") {
       v_.cross_past_value_names = value;
+    } else if (name == "curr_seq_length") {
```
kunal-vaishnavi, line 150: We should try to use the full-length names as the labels in the config. The abbreviated names can be the values of the labels.

src/models/decoder_only_pipeline.cpp
```diff
@@ -0,0 +1,229 @@
...
+  bool device_allocator_created = false;
+  for (const auto& model : config_->model.decoder.pipeline) {
+    sessions_.emplace_back(OrtSession::Create(ort_env, (config_->config_path / fs::path(model.filename)).c_str(),
+                                              GetSessionOptions(model.model_id)));
+
+    if (!device_allocator_created && model.session_options.has_value()) {
+      const auto& provider_options = (*model.session_options).provider_options;
+      if (std::any_of(provider_options.begin(), provider_options.end(),
+                      [](const auto& elem) { return !elem.name.empty(); })) {
+        InitDeviceAllocator(*sessions_.back());
```
RyanUnderhill, line 20: Why not check allocator_device_ for being set vs empty intead of device_allocator_created?

src/models/model.cpp
```diff
@@ -277,167 +287,194 @@ void Model::InitDeviceAllocator([[maybe_unused]] OrtSession& session) {
...
-      ort_options.DisableCpuMemArena();
-  }
-
-  if (options.enable_mem_pattern.has_value()) {
-    if (options.enable_cpu_mem_arena.value())
-      ort_options.EnableMemPattern();
-    else
-      ort_options.DisableMemPattern();
-  }
+  const auto create_session_options_from_config = [&](const Config::SessionOptions& config_session_options,
```
RyanUnderhill, line 290: Can this just be a separate function vs a lambda?

baijumeswani: I try to avoid changing the header file, if possible, to save on compile time. But I think it just makes reading better if this were a function of its own. Will make the change.

src/models/decoder_only_pipeline.cpp
```diff
@@ -0,0 +1,229 @@
...
+}
+
+IntermediatePipelineState::IntermediatePipelineState(const DecoderOnlyPipelineModel& model, const GeneratorParams& params,
+                                                     size_t pipeline_state_index)
+    : State{params, model},
+      id_{pipeline_state_index},
+      model_{model} {}
+
+bool IntermediatePipelineState::HasInput(std::string_view name) const {
+  for (const auto& input : model_.config_->model.decoder.pipeline[id_].inputs) {
```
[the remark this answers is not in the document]
baijumeswani: I do not remember why I did it this way. I'll use std::any_of. :)

src/config.h
```diff
@@ -116,6 +120,20 @@ struct Config {
...
 
+      struct PipelineModel {
+        std::string model_id;
+        std::string filename;
+        std::optional<SessionOptions> session_options;
+
+        std::vector<std::string> inputs;
+        std::vector<std::string> outputs;
+        std::unordered_map<std::string, std::string> output_names_forwarder;
+        bool run_on_first_token_generation{true};
```
yufenglee, line 131: >run_on_first_token_generation [](http://example.com/codeflow?start=13&length=29)

nit: change it to run_on_prompt?

src/config.h
```diff
@@ -116,6 +120,20 @@ struct Config {
...
+        bool run_on_nth_token_generation{true};
```
yufenglee, line 132: >run_on_nth_token_generation [](http://example.com/codeflow?start=13&length=27)

nit: change name to run_on_token_gen

yufenglee (approved):
:shipit:

src/models/kv_cache.cpp
```diff
@@ -113,6 +113,12 @@ void KV_Cache_Combined::PickPastState(std::span<const int32_t> beam_indices, int
   }
 }
 
+bool KV_Cache::IsCacheNeeded(const Model& model) {
+  char past_key_name[64];
+  snprintf(past_key_name, std::size(past_key_name), model.config_->model.decoder.inputs.past_key_names.c_str(), 0);
```
edgchen1, line 118: should we check the return value to see if the buffer was actually big enough? also noticed a few other similar calls to `snprintf` in this file.

kunal-vaishnavi (approved):

src/models/kv_cache.cpp
baijumeswani: Made changes to address this.

RyanUnderhill (approved):

baijumeswani:
Thank you all for the review. :)

PR #731, opened 2024-07-29
Load the onnxruntime shared lib in python

baijumeswani:


snnn (approved):

skyline75489 (approved):
This is basically what I was thinking in #724. I was planning to do it in C++. Your choice is obviously better.

src/models/onnxruntime_api.h
```diff
@@ -213,32 +213,7 @@ inline void InitApi() {
...
-      // Search for pip installation
-      for (const std::string& lib_name : target_libraries) {
-        std::string pip_path{current_module_dir + "/../onnxruntime/capi/" + lib_name};
-        ort_lib_handle = LoadDynamicLibraryIfExists(pip_path);
-        if (ort_lib_handle != nullptr) {
-          break;
-        }
-      }
-    }
+    ort_lib_handle = LoadDynamicLibraryIfExists("libonnxruntime.so.1");
```
skyline75489, line 216: Is this going to work with older ORT installation, or do we want to force ORT 1.19 with ort-genai 0.0.4?

baijumeswani: I don't think we need to make it work for older ort installation considering that the library depends on certain features only available in 1.19.0.
Moreover, --in the case of python-- the shared library is not packaged inside ort package before 1.19.0.

PR #732, opened 2024-07-29
Added null check for proper disposal of model class when could not laad the onnx runtime.

azchohfi:


yufenglee (approved):

src/csharp/Model.cs
```diff
@@ -42,8 +42,11 @@ protected virtual void Dispose(bool disposing)
             {
                 return;
             }
-            NativeMethods.OgaDestroyModel(_modelHandle);
-            _modelHandle = IntPtr.Zero;
+            if (_modelHandle != IntPtr.Zero)
```
baijumeswani, line 45: Just out of curiosity, what problem was the existing code giving?

baijumeswani: ok, I see that it would hit if the _modelHandle was not created correctly. Thanks for the fix.

azchohfi: Very rare case, as the initialization on the ctor would have to had failed, but I did hit this once and it crashed the app. Its a simple fix, so I rather not crash the whole app and fail gracefully here.

PR #745, opened 2024-08-01
Update README for gradio app

natke:
This PR attempts to address component governance issues on vulnerable components, by firstly removing unncessary dependencies from the example application, and then moving what is necessary from the requirements.txt into the README.


rui-ren (approved):
Thank you!

PR #748, opened 2024-08-01
Add nuget package README

natke:
Not sure if the example output is too wordy?


nuget/PACKAGE.md
baijumeswani, line 1: I don't think this is too wordy. But I was thinking we could have the same README.md for all packages and only change the language code snippet.

Anytime we update our README, we can update all language READMEs. 

We could do this by doing something like that. Showing for examples here, but can be replaced for readmes instead.

[<mark>Python Examples</mark>](https://github.com/microsoft/onnxruntime-genai/tree/main/examples/python) | [C# Examples](https://github.com/microsoft/onnxruntime-genai/tree/main/examples/csharp) | [C++ Examples](https://github.com/microsoft/onnxruntime-genai/tree/main/examples/c)

And the corresponding code snippet below. But this can be done separately later.

nuget/Microsoft.ML.OnnxRuntimeGenAI.Managed.nuspec
```diff
@@ -19,7 +19,7 @@
   </metadata>
   <files>
     <file src="..\LICENSE" target="LICENSE" />
-    <file src="..\README.md" target="README.md" />
+    <file src="PACKAGE.md" target="PACKAGE.md" />
```
baijumeswani, line 22: Considering the location of the [python package description](https://github.com/microsoft/onnxruntime-genai/blob/main/src/python/package_description.md), shall we add this package md file to https://github.com/microsoft/onnxruntime-genai/tree/main/src/csharp?

baijumeswani (approved):

nuget/Microsoft.ML.OnnxRuntimeGenAI.Managed.nuspec
natke: Good idea. I can move it

nuget/PACKAGE.md
natke: From the feedback from the .NET team, they like their READMEs to be .NET/C# specific. I know it is more for us but it is a better user experience

nuget/Microsoft.ML.OnnxRuntimeGenAI.Managed.nuspec
natke: Moved

baijumeswani (approved):

Issue #721, opened 2024-07-24
[Feature request] Support for LLama3.1


nguyenhoangthuan99:
Llama 3.1 has been released, and it includes a small change in the model's forward part that makes the current version of ONNX unable to run it.

I'm a member of [homebrew/jan](https://github.com/janhq). Currently, we support the Llama 3 model, and now I'm planning to update the model execution part of ONNX to support Llama 3.1 for integration into our project at [cortex.onnx](https://github.com/janhq/cortex.onnx).

kunal-vaishnavi:
The PR to support LLaMA 3.1 can be found [here](https://github.com/microsoft/onnxruntime-genai/pull/750). Once merged, you can use the ["from source"](https://github.com/microsoft/onnxruntime-genai/blob/main/src/python/py/models/README.md) commands for the model builder to build your desired ONNX model.

PR #750, opened 2024-08-01
Add LLaMA 3.1 to model builder

kunal-vaishnavi:
### Description

This PR adds [Meta's LLaMA 3.1](https://ai.meta.com/blog/meta-llama-3-1/) to the model builder. Here is an example command to get the ONNX model.

```bash
$ git clone https://github.com/microsoft/onnxruntime-genai
$ cd onnxruntime-genai/src/python/py/models
$ python3 builder.py -m meta-llama/Meta-Llama-3.1-8B-Instruct -o ./llama3.1_8b_instruct_fp16/ -p fp16 -e cuda -c ./cache/
```

### Motivation and Context

LLaMA 3.1 is the next generation of the LLaMA models. This PR also helps [this issue](https://github.com/microsoft/onnxruntime-genai/issues/721).

### DIFF
--- a/src/python/py/models/builder.py
+++ b/src/python/py/models/builder.py
@@ -201,7 +201,6 @@ class Model:
                     "low_freq_factor": low_freq_factor,
                     "high_freq_factor": high_freq_factor,
                 }
-                print(self.rotemb_attrs["rescale_inv_freq"])
 
         # Attention-specific variables (MHA, GQA, GQA + Rot.Emb., etc.)
         # Block-sparse attention-specific variables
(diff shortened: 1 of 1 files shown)


hanbitmyths (approved):

tianleiwu (approved):
Can we add a test with some tiny model like https://huggingface.co/yujiepan/llama-3.1-tiny-random/tree/main?

kunal-vaishnavi:
> Can we add a test with some tiny model like https://huggingface.co/yujiepan/llama-3.1-tiny-random/tree/main?

We are still working on adding a full suite of tests for the currently supported models. Once it is ready, we can add LLaMA 3.1 to the test suite.

PR #753, opened 2024-08-02
Only log when enabled to avoid unnecessary printing on stdout

baijumeswani:
#724 logs when loading the ort lib. The log can create problems if the logging is not enabled (in particular [this line](https://github.com/microsoft/onnxruntime-genai/blob/32378d96cbb7e921e0d986e5e1a56f086c32fe96/src/logging.cpp#L82)) in debug builds. And it also prints to console unnecessary information.

Changing the behavior in this PR.

### DIFF
--- a/src/models/env_utils.cpp
+++ b/src/models/env_utils.cpp
@@ -3,6 +3,8 @@
 
 #include "env_utils.h"
 
+#include <stdexcept>
+
 #if _MSC_VER
 #include <Windows.h>
 #endif
@@ -31,11 +33,25 @@ std::string GetEnvironmentVariable(const char* var_name) {
     return buffer;
   }
 
-  return std::string();
+  return {};
 #else
   const char* val = getenv(var_name);
   return val == nullptr ? "" : std::string(val);
 #endif  // _MSC_VER
+}
+
+void GetEnvironmentVariable(const char* var_name, bool& value) {
+  std::string str_value = GetEnvironmentVariable(var_name);
+  if (str_value == "1" || str_value == "true") {
+    value = true;
+  } else if (str_value == "0" || str_value == "false") {
+    value = false;
+  } else if (!str_value.empty()) {
+    throw std::invalid_argument("Invalid value for environment variable " + std::string(var_name) + ": " + str_value +
+                                ". Expected '1' or 'true' for true, '0' or 'false' for false.");
+  }
+
+  // Otherwise, value will not be modified.
 }
 
 }  // namespace Generators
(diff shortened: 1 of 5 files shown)


src/models/env_utils.cpp
```diff
@@ -0,0 +1,41 @@
...
+  // If the function succeeds, the return value is the number of characters stored in the buffer pointed to by lpBuffer, not including the terminating null character.
+  // Therefore, If the function succeeds, kBufferSize should be larger than char_count.
+  auto char_count = ::GetEnvironmentVariableA(var_name, buffer.data(), kBufferSize);
+
+  if (kBufferSize > char_count) {
+    buffer.resize(char_count);
+    return buffer;
+  }
+
+  return std::string();
```
skyline75489, line 34: This can just be `return {}` if you like

src/models/onnxruntime_api.h
```diff
@@ -93,11 +94,14 @@ p_session_->Run(nullptr, input_names, inputs, std::size(inputs), output_names, o
 #define PATH_MAX (4096)
 #endif
 
-#define LOG_DEBUG(...) Generators::Log("debug", __VA_ARGS__)
-#define LOG_INFO(...) Generators::Log("info", __VA_ARGS__)
-#define LOG_WARN(...) Generators::Log("warning", __VA_ARGS__)
-#define LOG_ERROR(...) Generators::Log("error", __VA_ARGS__)
-#define LOG_FATAL(...) Generators::Log("fatal", __VA_ARGS__)
+#define LOG_WHEN_ENABLED(LOG_FUNC) \
```
skyline75489, line 97: How about move all these macros to `logging.h`, including the androids and other ones the I added in my PRs. That would make this file less bloated.

src/models/onnxruntime_api.h
```diff
@@ -175,6 +179,12 @@ inline void InitApi() {
     return;
   }
 
+  const std::string ort_lib = Generators::GetEnvironmentVariable("ORTGENAI_LOG_ORT_LIB");
+  if (ort_lib == "1" || ort_lib == "true") {
```
skyline75489, line 183: Do we have similar checks elsewhere in the project? Maybe we need a consolidated way to check for `true` and `false`.

baijumeswani: Will make a utility for this.

src/models/onnxruntime_api.h
baijumeswani: Let's avoid using macros as a general-purpose logging utility for now. If it comes to that, we can make this happen later.

skyline75489: sure, no problem

skyline75489 (approved):
Thanks for amending it :)

PR #755, opened 2024-08-02
Add GetOutput for C API

ajindal1:
Add GetOutput for C API


src/ort_genai_c.cpp
```diff
@@ -208,6 +208,15 @@ OgaResult* OGA_API_CALL OgaGenerator_GenerateNextToken(OgaGenerator* generator)
   OGA_CATCH
 }
 
+OgaResult* OGA_API_CALL OgaGenerator_GetOutput(const OgaGenerator* oga_generator, const char* name, OgaTensor** out) {
+  OGA_TRY
+  auto& generator = *reinterpret_cast<const Generators::Generator*>(oga_generator);
+  auto* ortValueOutput = generator.state_->GetOutput(name);
+  *out = reinterpret_cast<OgaTensor*>(ortValueOutput);
```
yufenglee, line 215: The buffer of 'OgaTensor** out' is not handled properly here. Its buffer is shared with the buffer of output tensors of the model. The lifetime of 'OgaTensor** out' can be different with the output tensor. 

We need to change OgaTensor to make it able to own the buffer.

src/ort_genai_c.cpp
```diff
@@ -208,6 +208,34 @@ OgaResult* OGA_API_CALL OgaGenerator_GenerateNextToken(OgaGenerator* generator)
   OGA_CATCH
 }
 
+OgaResult* OGA_API_CALL OgaGenerator_GetOutput(const OgaGenerator* oga_generator, const char* name, OgaTensor** out) {
```
yufenglee, line 211: OgaTensor doesn't have a destroy function. We need a function for it.

baijumeswani: I think we already have a destroy function for `OgaTensor`:

https://github.com/microsoft/onnxruntime-genai/blob/f3a77dadf94ea35e1e7b61f0533c1470137671f5/src/ort_genai_c.h#L286

yufenglee: yes, it does. thanks

src/ort_genai_c.cpp
```diff
@@ -208,6 +208,34 @@ OgaResult* OGA_API_CALL OgaGenerator_GenerateNextToken(OgaGenerator* generator)
...
+  OGA_TRY
+  auto& generator = *reinterpret_cast<const Generators::Generator*>(oga_generator);
+  auto* ortvalue_output = generator.state_->GetOutput(name);
+  auto type_info = ortvalue_output->GetTensorTypeAndShapeInfo();
+  std::unique_ptr<OrtValue> ortvalue_clone = OrtValue::CreateTensor(*generator.model_->allocator_device_,
```
yufenglee, line 216: the allocator_device_ can be allocator on GPU. allocator_cpu_ is the one we should use.

src/ort_genai_c.cpp
```diff
@@ -208,6 +208,34 @@ OgaResult* OGA_API_CALL OgaGenerator_GenerateNextToken(OgaGenerator* generator)
...
+                                                                    type_info->GetShape(),
+                                                                    type_info->GetElementType());
+  // Copy data to ortvalue_clone
+  auto element_size = Generators::SizeOf(type_info->GetElementType());
+  auto data_size = type_info->GetElementCount() * element_size;
+#if USE_CUDA
+  if (ortvalue_output->GetTensorMemoryInfo().GetDeviceType() == OrtMemoryInfoDeviceType_GPU && generator.model_->device_type_ == Generators::DeviceType::CUDA) {
+    cudaMemcpy(ortvalue_clone->GetTensorMutableRawData(), ortvalue_output->GetTensorMutableRawData(), data_size, cudaMemcpyDeviceToHost);
+  }
+#else
```
yufenglee, line 226: could you please add support for DML?

src/ort_genai_c.h
```diff
@@ -224,6 +224,8 @@ OGA_EXPORT bool OGA_API_CALL OgaGenerator_IsDone(const OgaGenerator* generator);
 OGA_EXPORT OgaResult* OGA_API_CALL OgaGenerator_ComputeLogits(OgaGenerator* generator);
 OGA_EXPORT OgaResult* OGA_API_CALL OgaGenerator_GenerateNextToken(OgaGenerator* generator);
 
+OGA_EXPORT OgaResult* OGA_API_CALL OgaGenerator_GetOutput(const OgaGenerator* oga_generator, const char* name, OgaTensor** out);
```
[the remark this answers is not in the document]
yufenglee: and please also update the OgaTensor document to specify that the OgaTensor can also own the buffer.

src/ort_genai_c.cpp
ajindal1: Okay, I thought we need to return the logits on the same device. So, I will return the values on the CPU only.

src/ort_genai_c.cpp
```diff
@@ -208,6 +208,48 @@ OgaResult* OGA_API_CALL OgaGenerator_GenerateNextToken(OgaGenerator* generator)
...
+                                                                    type_info->GetElementType());
+  // Copy data to ortvalue_clone
+  auto element_size = Generators::SizeOf(type_info->GetElementType());
+  auto data_size = type_info->GetElementCount() * element_size;
+#if USE_CUDA
+  if (ortvalue_output->GetTensorMemoryInfo().GetDeviceType() == OrtMemoryInfoDeviceType_GPU && generator.model_->device_type_ == Generators::DeviceType::CUDA) {
+    cudaMemcpy(ortvalue_clone->GetTensorMutableRawData(), ortvalue_output->GetTensorMutableRawData(), data_size, cudaMemcpyDeviceToHost);
+  }
+#elif USE_DML
+  if (*generator.model_->device_type_ == DeviceType::DML) {
```
baijumeswani, line 227: I think here we need to check where the ortvalue is allocated. Because even though the device_type_ is DML, the ortvalue could still reside on cpu (for example in the case of logits).

ajindal1: Okay, makes sense. Also, for every such kind of scenario (CUDA, DML), there should be an else statement as well to ensure that the data is copied using std::copy, right?

src/ort_genai_c.cpp
yufenglee: yes, we return the value on CPU only.

src/ort_genai_c.cpp
ajindal1: Thanks, fixed.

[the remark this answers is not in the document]
ajindal1: Returning a copy

src/ort_genai_c.cpp
```diff
@@ -208,6 +208,48 @@ OgaResult* OGA_API_CALL OgaGenerator_GenerateNextToken(OgaGenerator* generator)
...
+    generator.model_->GetDmlReadbackHeap()->ReadbackFromGpu(
+        std::span(reinterpret_cast<uint8_t*>(cpu_tensor), data_size),
+        gpu_resource.Get(),
+        0,
+        D3D12_RESOURCE_STATE_UNORDERED_ACCESS);
+  }
+#else
+  std::copy(static_cast<uint8_t*>(ortvalue_output->GetTensorMutableRawData()),
+            static_cast<uint8_t*>(ortvalue_output->GetTensorMutableRawData()) + data_size,
+            static_cast<uint8_t*>(ortvalue_clone->GetTensorMutableRawData()));
```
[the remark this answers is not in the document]
ajindal1: Done, thanks.

yufenglee (approved):
:shipit:

src/ort_genai_c.h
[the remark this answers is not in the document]
ajindal1: Updated.

baijumeswani (approved):

PR #756, opened 2024-08-05
Adds OgaAppendTokenSequence function

MorkTheOrk:
Adding a function to the C API that allows the developer to add a custom token sequence into a OgaSeqence. 

There is no option currently to add tokens to the OgaSequences because the `std::vector` is hidden from the developer.

### DIFF
--- a/src/ort_genai_c.cpp
+++ b/src/ort_genai_c.cpp
@@ -58,12 +58,12 @@ OgaResult* OGA_API_CALL OgaCreateSequences(OgaSequences** out) {
   OGA_CATCH
 }
 
-OgaResult* OGA_API_CALL OgaAppendTokenSequence(const int32_t * token_ptr, int32_t token_cnt, OgaSequences* sequence) {
+OgaResult* OGA_API_CALL OgaAppendTokenSequence(const int32_t * token_ptr, uint32_t token_cnt, OgaSequences* sequence) {
   OGA_TRY
   Generators::TokenSequences* toks = reinterpret_cast<Generators::TokenSequences*>(sequence);
   std::vector<int32_t> tmp(token_cnt);
   for (size_t i = 0; i < token_cnt; i++) {
-    tmp[i] = (token_ptr[i]);
+    tmp[i] = token_ptr[i];
   }
   toks->emplace_back(std::move(tmp));
   return nullptr;
--- a/src/ort_genai_c.cpp
+++ b/src/ort_genai_c.cpp
@@ -58,6 +58,18 @@ OgaResult* OGA_API_CALL OgaCreateSequences(OgaSequences** out) {
   OGA_CATCH
 }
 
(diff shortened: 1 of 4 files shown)


yufenglee:
Could you please add an unit test and example to use the API?

baijumeswani:
> There is no option currently to add tokens to the OgaSequences because the std::vector is hidden from the developer.

What is the motivation behind exposing this functionality to end users? Is it to support scenarios where user would use an external tokenizer?

MorkTheOrk:
Hi @baijumeswani, Yes this is exactly for an external tokenizer, otherwise the developer would need to transcode the tokens using the OGA Tokenizer. Usually this results into the same tokens it just adds some constant overhead which is undesired.

src/ort_genai_c.cpp
```diff
@@ -58,6 +58,18 @@ OgaResult* OGA_API_CALL OgaCreateSequences(OgaSequences** out) {
   OGA_CATCH
 }
 
+OgaResult* OGA_API_CALL OgaAppendTokenSequence(const int32_t * token_ptr, int32_t token_cnt, OgaSequences* sequence) {
+  OGA_TRY
+  Generators::TokenSequences* toks = reinterpret_cast<Generators::TokenSequences*>(sequence);
+  std::vector<int32_t> tmp(token_cnt);
+  for (size_t i = 0; i < token_cnt; i++) {
+    tmp[i] = (token_ptr[i]);
```
baijumeswani, line 66:
```suggestion
    tmp[i] = token_ptr[i];
```

src/ort_genai_c.h
baijumeswani, line 1: Could you also please add the C++ implementation [here](https://github.com/microsoft/onnxruntime-genai/blob/62daf5e8e7519c7eef25ee2ccf371e280fb07d14/src/ort_genai.h#L87)?

baijumeswani:
[MorkTheOrk](https://github.com/MorkTheOrk) thanks for the PR. Do you think you could address the open comments and the failing pipeline (clang-format) soon? We can then include it in the release branch for onnxruntime-genai 0.4.0.

src/ort_genai_c.h
MorkTheOrk: Added, see [here](https://github.com/microsoft/onnxruntime-genai/blob/0a48ca6b67b132b4a9394640801513b184b25e95/src/ort_genai.h#L106)

MorkTheOrk:
> Could you please add an unit test and example to use the API?

Added a C API test [here](https://github.com/MorkTheOrk/onnxruntime-genai/blob/0a48ca6b67b132b4a9394640801513b184b25e95/test/c_api_tests.cpp#L63)

MorkTheOrk:
@microsoft-github-policy-service agree company="Nvidia"

baijumeswani:
@MorkTheOrk could you please fix the clang-format error?

MorkTheOrk:
@baijumeswani could we run the workflows to see if anything else is wrong? If everything checks out, feel free to merge.

yufenglee (approved):

PR #762, opened 2024-08-06
Prepare validation pipeline for 0.4.0 release

skyline75489:
Improve both wheel & NuGet packaging & validation pipelines.

### DIFF
--- a/.pipelines/nuget-publishing.yml
+++ b/.pipelines/nuget-publishing.yml
@@ -36,7 +36,7 @@ parameters:
 - name: ort_version
   displayName: 'OnnxRuntime version'
   type: string
-  default: '1.18.0-dev-20240426-1256-b842effa29'
+  default: '1.19.0-dev-20240805-1630-ee2fe87e2d'
 
 - name: ort_cuda_version
   displayName: 'OnnxRuntime GPU version'
@@ -46,7 +46,7 @@ parameters:
 - name: ort_dml_version
   displayName: 'OnnxRuntime DML version'
   type: string
-  default: '1.18.0-dev-20240426-0116-b842effa29'
+  default: '1.19.0-dev-20240805-1630-ee2fe87e2d'
 
 - name: cuda_version
   displayName: 'CUDA version'
--- a/.pipelines/nuget-publishing.yml
+++ b/.pipelines/nuget-publishing.yml
@@ -41,7 +41,7 @@ parameters:
 - name: ort_cuda_version
   displayName: 'OnnxRuntime GPU version'
   type: string
-  default: '1.18.0-dev-20240426-0614-b842effa29'
+  default: '1.19.0-dev-20240805-0337-88c811b638'
 
 - name: ort_dml_version
   displayName: 'OnnxRuntime DML version'
@@ -98,6 +98,7 @@ stages:
     enable_win_dml: ${{ parameters.enable_win_dml }}
     enable_win_arm64: ${{ parameters.enable_win_arm64 }}
     ort_version: ${{ parameters.ort_version }}
+    ort_cuda_version: ${{ parameters.ort_cuda_version }}
     ort_dml_version: ${{ parameters.ort_dml_version }}
     build_config: ${{ parameters.build_config }}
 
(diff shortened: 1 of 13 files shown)


baijumeswani (approved):

.pipelines/stages/jobs/capi-packaging-job.yml
```diff
@@ -55,10 +55,7 @@ jobs:
   - name: os
     value: ${{ parameters.os }}
   - name: feed_name
-    ${{ if and(eq(parameters.cuda_version, '12.2'), eq(parameters.ep, 'cuda')) }}:
```
skyline75489, line 58: The main `ORT-Nightly` feed now have CUDA12 packages produced. I'm not sure if we should keep using `ort-cuda-12-nightly`

src/models/onnxruntime_api.h
```diff
@@ -144,6 +129,31 @@ inline std::string GetCurrentModuleDir() {
   return module_directory;
 }
 
+inline void* LoadDynamicLibraryIfExists(const std::string& path) {
+  LOG_INFO("Attempting to dlopen %s", path.c_str());
+  void* ort_lib_handle = dlopen(path.c_str(), RTLD_NOW | RTLD_LOCAL);
+  if (ort_lib_handle == nullptr) {
+    char* err = dlerror();
+    LOG_WARN("Error while dlopen: %s", (err != nullptr ? err : "Unknown"));
+    // Trying current dir
```
skyline75489, line 138: This is needed for NuGet on Linux, because `dlopen` won't search the current directory.

.pipelines/nuget-publishing.yml
```diff
@@ -36,17 +36,17 @@ parameters:
 - name: ort_version
   displayName: 'OnnxRuntime version'
   type: string
-  default: '1.18.0-dev-20240426-1256-b842effa29'
+  default: '1.19.0-dev-20240805-1630-ee2fe87e2d'
```
skyline75489, line 39: Note that these versions will be written to the final nuspec as a dependency. We should probably only use official ORT release as the dependency.

baijumeswani: I think when the 1.19.0 stable is published, we can update the version here.

src/models/onnxruntime_api.h
baijumeswani: My understanding is that this should typically be resolved with LD_LIBRARY_PATH. Is that not what is recommended?

.pipelines/nuget-publishing.yml
skyline75489: Come to think about it, the current dependency has strict version requirement like this:

```xml
<dependency id="Microsoft.ML.OnnxRuntime.Gpu" version="1.19.0-dev-20240805-1630-ee2fe87e2d" />
```

Maybe we can relax it a bit?

Reference: https://learn.microsoft.com/en-us/nuget/concepts/package-versioning?tabs=semver20sort

.pipelines/stages/jobs/capi-packaging-job.yml
baijumeswani: We can change to ort-nightly feed and use ort-nightly-gpu.
Should we keep the option for cuda 11? We will need to change to ort-cuda-11-nightly feed for that.

skyline75489: If we want to always test the latest nightly, we should primarily use `ORT-Nightly` which has the latest packages. From the names, `ort-cuda-11-nightly` and `ort-cuda-12-nightly` should also have latest package. But right now the packages in `ort-cuda-12-nightly` are 3-months old. I'm quite confused by which feed is the "correct" one to use.

src/models/onnxruntime_api.h
skyline75489: Do you mean we add instructions for user to tweak `LD_LIBRARY_PATH`?

In #731 you managed to load the ORT library using `ctypes` in Python. Python has this global location for site-packages. But for NuGet & C# projects, dependencies are copied and stored locally in the artifact folder. Therefore, I think it would be better for us to just load the nearby libraries, because this *should* be the desired library.

baijumeswani: ok, if it is guaranteed that the ort lib will end up in the same folder as ortgenai lib, this should be a good strategy.

.pipelines/nuget-publishing.yml
skyline75489: @natke This is how the dependency is added, as I mentioned in your doc PR.

PR #774, opened 2024-08-09
Fix quant model loading from multiple safetensors

BowenBao:
It is possible to create and append incomplete/empty layers, when weights of a layer is split across multiple safetensors files.

This is one of such example models https://huggingface.co/yujiepan/Meta-Llama-3-8B-Instruct-gptq-w8asym/tree/main

### DIFF
--- a/src/python/py/models/quantized_model.py
+++ b/src/python/py/models/quantized_model.py
@@ -90,10 +90,11 @@ class QuantizedModel:
         self.lm_head = TensorModule()
         self.layers = []
 
+        layers = {}
         layer_id = 0
         for weight_file in os.listdir(input_path):
             if weight_file.endswith(".safetensors"):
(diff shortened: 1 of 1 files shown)


src/python/py/models/quantized_model.py
```diff
@@ -90,10 +90,11 @@ def __init__(self, quant_type, input_path, bits, group_size, q_size, kv_size, in
         self.lm_head = TensorModule()
         self.layers = []
```
kunal-vaishnavi, line 92: Should we have `self.layers = {}` instead and then cast it to a list at the end instead of holding two copies (`self.layers` and `layers`)?

kunal-vaishnavi:
Closing this PR since its changes are merged into [this PR](https://github.com/microsoft/onnxruntime-genai/pull/780) now.

PR #785, opened 2024-08-12
Simplify the provider detection using the preprocessor

RyanUnderhill:
I saw this opportunity while reviewing the APIs.

It looks more consistent to have the preprocessor variable be 1 when present and 0 when not (vs 1 when present and undefined when not). This also simplifies the python APIs a lot.


yufenglee (approved):
:shipit:

cmake/check_cuda.cmake
```diff
@@ -60,6 +60,7 @@ elseif(USE_CUDA)
 else()
   file(GLOB generator_cuda_srcs "${GENERATORS_ROOT}/*_cuda*.*")
   list(REMOVE_ITEM generator_srcs ${generator_cuda_srcs})
+  add_compile_definitions(USE_CUDA=0)
```
baijumeswani, line 63: Any code usage of ```#ifndef USE_CUDA``` would no longer work with this change, right?

baijumeswani (approved):

cmake/check_cuda.cmake
RyanUnderhill: Correct, we don't have any usage like that as it'd be inconsistent. And this will prevent anyone from trying!

PR #796, opened 2024-08-14
Support for loading multiple images for phi3 vision

baijumeswani:
So far, onnxruntime-genai only offered a way to load one image at a time for running vision-based models. This pull-request adds the necessary API to load multiple images for running the phi 3 vision model.

It also adds support in Python to load from a memory buffer. This support is restricted to Python. If needed, it can also be added to C/C++/C#.


examples/c/src/phi3v.cpp
```diff
@@ -92,36 +103,49 @@ void C_API(const char* model_path) {
...
-    std::cout << "Image Path (leave empty if no image):" << std::endl;
-    std::getline(std::cin, image_path);
+    std::string image_paths_str;
+    std::cout << "Image Path (comma separated; leave empty if no image):" << std::endl;
+    std::getline(std::cin, image_paths_str);
     OgaImages* images = nullptr;
-    if (image_path.empty()) {
+    std::vector<std::string> image_paths;
+    for (size_t start = 0, end = 0; end < image_paths_str.size(); start = end + 1) {
+      end = image_paths_str.find(',', start);
```
kunal-vaishnavi, line 112: Will the processor be affected by whitespace in the paths when using `.find()` and `.substr()` or is leading and trailing whitespace already stripped?

For example, whether the returned paths are `path_1`, ` path_2`, and ` path_3` (if a user passes the paths as `path_1, path_2, path_3` with a space after each comma) or whether the returned paths are `path_1`, `path_2`, and `path_3` (if a user passes the paths as `path_1,path_2,path_3` without a space after each comma), both cases should work.

src/ort_genai.h
```diff
@@ -279,12 +279,26 @@ struct OgaTensor : OgaAbstract {
...
     OgaImages* p;
-    OgaCheckResult(OgaLoadImage(image_path, &p));
+    OgaStringArray* strs;
+    OgaCheckResult(OgaCreateStringArrayFromStrings(image_paths.data(), image_paths.size(), &strs));
+    OgaCheckResult(OgaLoadImages(strs, &p));
+    OgaDestroyStringArray(strs);
     return std::unique_ptr<OgaImages>(p);
   }
 
+#if __cplusplus >= 202002L
```
kunal-vaishnavi, line 291: Why is this `if` check needed?

examples/c/src/phi3v.cpp
baijumeswani: Whitespaces will cause a problem as it stands right now. I can modify to trim the strings after extracting the paths using the delimiter.

src/ort_genai.h
baijumeswani: std::span is not available in c++ 17. This #if checks if the compiler language is c++20 or above.

james-valbuena:
@baijumeswani Many thanks for integrating multi-image processing into the codebase, it is really useful. Do you know this merged PR will be integrated into the latest pip package?

baijumeswani:
I expect the new ort-genai package (0.5.0) to be published at the beginning of the next month.

For the 0.5.0 release, we are targeting getting ort-genai ready for phi3.5 family of models and the whisper model. We expect all changes ready and validated by the end of this month.

james-valbuena:
@baijumeswani  Thank you for the update, should you require any UAT support for this ticket, I would be more than happy to assist.

PR #799, opened 2024-08-15
Check for resource leaks on Shutdown() to help users

RyanUnderhill:
Adds an instance count to the main types users allocate and destroy through our external API.

On calling Shutdown(), if any count is non zero it will say the count and type of what hasn't been cleaned up.
Will also say if Shutdown() hasn't been called.

Also adds a check on process exit that Shutdown() is called. If not, will give an error message and abort the process.

This should help users more easily write correct code. Trivial to extend to more types, just add to the type list and inherit from LeakChecked.

Example output:
```
[----------] Global test environment tear-down
[==========] 31 tests from 4 test suites ran. (45750 ms total)
[  PASSED  ] 31 tests.
Shutting down OnnxRuntime... done
OGA Error: 1 instances of struct Generators::Model were leaked.
    Please see the documentation for the API being used to ensure proper cleanup.
```

### DIFF
--- a/src/generators.cpp
+++ b/src/generators.cpp
@@ -20,11 +20,11 @@ std::atomic<int> TrackedResource::count_{};
 // Validate process exit conditions, as this is done atexit, we print errors to stderr and throw an exception to stop the process
 void ValidateShutdown() {
   if (GetOrtGlobals()) {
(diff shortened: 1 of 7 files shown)


baijumeswani (approved):

ramkrishna2910:
@RyanUnderhill can you to the documentation for proper shutdown. Are there python APIs exposed for this?

PR #800, opened 2024-08-15
Add NuGet package README for native packages and link from managed package README

natke:


nuget/MANAGED_PACKAGE.md
```diff
@@ -0,0 +1,3 @@
+## About
+
+This package is a dependency of [Microsoft.ML.OnnxRuntimeGenAI](https://www.nuget.org/packages/Microsoft.ML.OnnxRuntimeGenAI) and does not need to be installed directly.
```
baijumeswani, line 3: Since we have multiple packages, do we need to highlight that?

tools/nuget/generate_nuspec_for_native_nuget.py
```diff
@@ -74,7 +74,7 @@ def generate_license(line_list):
     line_list.append('<license type="file">LICENSE</license>')
 
 def generate_readme(line_list):
-    line_list.append('<readme>README.md</readme>')
+    line_list.append('<readme>nuget/PACKAGE.md</readme>')
```
baijumeswani, line 77: README is inside src\csharp:

https://github.com/microsoft/onnxruntime-genai/blob/main/src/csharp/README.md

nuget/MANAGED_PACKAGE.md
natke: Maybe?! I thought about it. What do you think?

tools/nuget/generate_nuspec_for_native_nuget.py
natke: I moved it

baijumeswani (approved):

nuget/MANAGED_PACKAGE.md
natke: I think I'll leave as hopefully the instruction is enough to say the user does not have to depend on this package

ajindal1:
Seems like this [error](https://aiinfra.visualstudio.com/ONNX%20Runtime/_build/results?buildId=536126&view=logs&j=cfddc5e9-9413-5333-1aa2-ff2db0d6a1ac&t=ba149374-d9ca-5d3a-6629-e4055ab8f597&l=29) is caused by this PR, might be wrong. Need to investigate more on this.

Issue #552, opened 2024-05-31
Phi-3-mini-4k-instruct-onnx model generates nonsensical results when prompt is longer than half of the context window length


jackylu0124:
I am running the `Phi-3-mini-4k-instruct-onnx` model on desktop CPU, and one behavior I have noticed is that, after the back and forth conversation is longer than half of the context window length (in other words, longer than 2048 tokens), the model starts generating nonsensical and unreadable results. This is weird because I would expect the model to continue generating readable results at least to the point that's closer to the end of the context window length. I would really appreciate any insights on this problematic behavior as well as any ways that I can fix this issue. Thanks a lot in advance!

To reproduce this issue, you can run the example `phi3-qa.py` script ([https://github.com/microsoft/onnxruntime-genai/blob/main/examples/python/phi3-qa.py](https://github.com/microsoft/onnxruntime-genai/blob/main/examples/python/phi3-qa.py)) (with command like `python phi3-qa.py -m .\Phi-3-mini-4k-instruct-onnx\cpu_and_mobile\cpu-int4-rtn-block-32-acc-level-4`) with the if-block setting `search_options['max_length'] = 2048` commented out to allow longer input and also replace the `prompt` in the line `input_tokens = tokenizer.encode(prompt)`([https://github.com/microsoft/onnxruntime-genai/blob/6be88357c924de71da67f96a70b3c0f45803249f/examples/python/phi3-qa.py#L38](https://github.com/microsoft/onnxruntime-genai/blob/6be88357c924de71da67f96a70b3c0f45803249f/examples/python/phi3-qa.py#L38)) with the following conversation string (which is a conversation between the user and the assistant, and the string contains 2271 tokens):

```
[...]

PR #802, opened 2024-08-16
Update the attention_mask reformat for MHA

apsonawane:
For phi3 models the attention_mask reformat was incorrect. Updated the pattern in this below PR.
This PR also helps with the issue open here: https://github.com/microsoft/onnxruntime-genai/issues/552

### DIFF
--- a/src/python/py/models/builder.py
+++ b/src/python/py/models/builder.py
@@ -550,6 +550,11 @@ class Model:
         self.make_node("Tile", inputs=inputs, outputs=[output], name=name)
         self.make_value_info(output, dtype, shape=shape)
 
+    def make_trilu(self, name, inputs, upper, dtype, shape):
+        output = f"{name}/output_0"
+        self.make_node("Trilu", inputs=inputs, outputs=[output], name=name, upper=upper)
+        self.make_value_info(output, dtype, shape=shape)
+
     def make_equal(self, name, inputs, shape):
         output = f"{name}/output_0"
         self.make_node("Equal", inputs=inputs, outputs=[output], name=name)
@@ -660,7 +665,7 @@ class Model:
             # print(f"Quantizing to {self.onnx_dtype} on-the-fly is not currently supported.")
             # print(f"Saving as {self.io_dtype} on-the-fly and quantizing to {self.onnx_dtype} at the end.")
             return self.make_matmul_fp16_or_fp32(matmul, basename, root_input, **kwargs)
-
+        
         name = f"{basename}NBits"
 
         # Input weights are quantized, save quantized MatMul numpy weights for onnx model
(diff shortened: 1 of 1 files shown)


src/python/py/models/builder.py
```diff
@@ -1689,61 +1694,73 @@ def make_attention_mask_reformatting_for_mha(self):
         # Make nodes for the attention mask subgraphs that reformat the
         # 2D attention mask (B, S) to 4D causal attention mask (B, N, S, T)
```
tianleiwu, line 1695: I do not under stand this. I think the 2D mask is (B, T), and MHA supports 2D mask, shall we use that directly instead of converting to 4D in onnx graph?

natke (commented):
Does this change get validated by tests?

baijumeswani:
[apsonawane](https://github.com/apsonawane) is this pull-request still relevant? Could you please address the comments and update the PR when possible?

apsonawane:
Yes, I will be updating the PR

myadav2:
I've been experiencing a similar issue when fine-tuning phi3 and phi3.5 models as well. Produces a lot of tokens at the end that are gibberish even after fine-tuning. 

Looks like this has been solved with the latest ONNX release, but fine-tuning these ONNX models by converting them to torch is really tricky. Has the fix been made to the non-ONNX models as well? Any workaround for that?

yufenglee:
> Looks like this has been solved with the latest ONNX release, but fine-tuning these ONNX models by converting them to torch is really tricky. Has the fix been made to the non-ONNX models as well? Any workaround for that?

@myadav2, the overall process for fine-tuning is that: fine-tune with Pytorch -> use [ModelBuilder](https://github.com/microsoft/onnxruntime-genai/tree/main/src/python/py/models) or PyTorch ONNX export to convert the model to ONNX -> serve with ORT GenAI API

myadav2:
> > Looks like this has been solved with the latest ONNX release, but fine-tuning these ONNX models by converting them to torch is really tricky. Has the fix been made to the non-ONNX models as well? Any workaround for that?
> 
> @myadav2, the overall process for fine-tuning is that: fine-tune with Pytorch -> use [ModelBuilder](https://github.com/microsoft/onnxruntime-genai/tree/main/src/python/py/models) or PyTorch ONNX export to convert the model to ONNX -> serve with ORT GenAI API

Yes, but the issue of gibberish output after fine-tuning long context text is present for the base Pytorch models. I don't think we have a fix for that as far as I know

yufenglee:
> > > Looks like this has been solved with the latest ONNX release, but fine-tuning these ONNX models by converting them to torch is really tricky. Has the fix been made to the non-ONNX models as well? Any workaround for that?
> > 
> > 
> > @myadav2, the overall process for fine-tuning is that: fine-tune with Pytorch -> use [ModelBuilder](https://github.com/microsoft/onnxruntime-genai/tree/main/src/python/py/models) or PyTorch ONNX export to convert the model to ONNX -> serve with ORT GenAI API
> 
> Yes, but the issue of gibberish output after fine-tuning long context text is present for the base Pytorch models. I don't think we have a fix for that as far as I know

@myadav2, that's annoying. Could you please describe your issue in details and share some examples? I can bring the issues to the Phi3 model training team and see if they can help.

kunal-vaishnavi:
> Has the fix been made to the non-ONNX models as well? Any workaround for that?

The [Phi-3.5 PyTorch models](https://huggingface.co/collections/microsoft/phi-3-6626e15e9585a200d2d761e3) should also have this fix. If you still observe this behavior, you can open an issue in this repo so this can be tracked.

PR #808, opened 2024-08-16
Update VERSION_INFO

natke:


VERSION_INFO
```diff
@@ -1 +1 @@
-0.5.0-dev
+0.5.0-dev
```
baijumeswani, line 1: Adding a new line might break the build.

baijumeswani (approved):

PR #813, opened 2024-08-19
Time the generation process in C# example

skyline75489:
Make it more like the Python `model-generate.py` example.

Validation run: https://aiinfra.visualstudio.com/ONNX%20Runtime/_build/results?buildId=536632&view=results

### DIFF
--- a/examples/csharp/HelloPhi/Program.cs
+++ b/examples/csharp/HelloPhi/Program.cs
@@ -107,6 +107,6 @@ do
         var runTimeInSeconds = watch.Elapsed.TotalSeconds;
         var outputSequence = generator.GetSequence(0);
         var totalTokens = outputSequence.Length;
-        Console.WriteLine($"Tokens: {totalTokens} Time: {runTimeInSeconds:0.00} Tokens per second: {totalTokens / runTimeInSeconds:0.00}");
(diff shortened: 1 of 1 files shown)


examples/csharp/HelloPhi/Program.cs
```diff
@@ -80,23 +80,33 @@ void PrintUsage()
...
             generator.ComputeLogits();
             generator.GenerateNextToken();
             Console.Write(tokenizerStream.Decode(generator.GetSequence(0)[^1]));
         }
         Console.WriteLine();
+        watch.Stop();
+        var runTimeInSeconds = watch.Elapsed.TotalSeconds;
+        var outputSequence = generator.GetSequence(0);
+        var totalTokens = outputSequence.Length;
+        Console.WriteLine($"Tokens: {totalTokens} Time: {runTimeInSeconds:0.00} Tokens per second: {totalTokens / runTimeInSeconds:0.00}");
```
RyanUnderhill, line 110: Should the comment say what option it is, like 'Streaming Tokens:' so we know which one was timed?

skyline75489: that's better. will make it "Streaming Tokens:" for clarity

RyanUnderhill (approved):

PR #834, opened 2024-08-22
Simplify tensor type related code

RyanUnderhill:
Use variadic templates to make the code more generic and simpler.

Switch to using a template variable for TypeToTensorType, which avoids needing ::type all over the code.


ajindal1 (approved):
LGTM

PR #850, opened 2024-08-29
support phi3-moe in model builder

wangyems:


src/python/py/models/builder.py
```diff
@@ -249,6 +249,17 @@ def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
             "output_0": "",             # Output 0 for MLP layer
         }
 
+        # MoE-specific variables
+        self.moe_attrs = {
+            "top_k": 1,                             # Number of experts to select in MoE layer
```
kunal-vaishnavi, line 254: Can we set the default values similar to how the default values are set for attention, rotary embeddings, etc? Something like this:

https://github.com/microsoft/onnxruntime-genai/blob/ae18a2cf276ad7800c7e368d313e667b1865c298/src/python/py/models/builder.py#L159-L166

That way, the config's values will be used by default when available.

src/python/py/models/builder.py
```diff
@@ -1495,6 +1506,125 @@ def make_mlp_fc(self, layer_id, mlp, root_input):
...
+        #                 output
+        num_experts = self.moe_attrs["num_experts"]
+        top_k = self.moe_attrs["top_k"]
+        activation_type = self.moe_attrs["activation_type"]
+        normalize_routing_weights = self.moe_attrs["normalize_routing_weights"]
+        use_sparse_mixer = self.moe_attrs["use_sparse_mixer"]
+        use_int4 = self.moe_attrs["use_int4"]
+
+        gate_ops_base = f"/model/layers.{layer_id}/moe/gate/"
+        moe_name = f"/model/layers.{layer_id}/moe"
```
kunal-vaishnavi, line 1537:
```suggestion
        moe_name = f"/model/layers.{layer_id}/moe"
        gate_ops_base = f"{moe_name}/gate/"
```

src/python/py/models/builder.py
```diff
@@ -1495,6 +1506,125 @@ def make_mlp_fc(self, layer_id, mlp, root_input):
...
+
+        # Make MoE nodes
+        gate_name = gate_ops_base + "MatMul"
+        self.make_matmul(bsm.gate, gate_name, root_input)
+
+        shape_name = gate_ops_base + "Shape"
```
kunal-vaishnavi, line 1543: Can we create all of the node names with f-string formatting (e.g. `f"{gate_ops_base}/Shape"`) to stay consistent with how other names are created?

src/python/py/models/builder.py
```diff
@@ -1495,6 +1506,125 @@ def make_mlp_fc(self, layer_id, mlp, root_input):
...
+        w1_list = []
+        w2_list = []
+        w3_list = []
+        w1_scale_list = []
+        w2_scale_list = []
+        w3_scale_list = []
+
+        for i in range(num_experts):
+            # Quantize the weights with uint8
+            w1_scale, pre_qweight1= quant_dequant(bsm.experts[i].w1.weight, use_int4)
```
kunal-vaishnavi, line 1578: Is `QMoE` the only op available to create? Can a regular `MoE` op also be created if the user wants the non-quantized version?

src/python/py/models/builder.py
```diff
@@ -1495,6 +1506,125 @@ def make_mlp_fc(self, layer_id, mlp, root_input):
...
+        concat_name = gate_ops_base + "Concat"
+        self.make_concat(concat_name, ["/model/constants/TensorProto.INT64/1D/-1", f"{unsqueeze_name}/output_0"], dtype=TensorProto.INT64, shape=[2], axis=0)
+
+        gate_reshape_name = gate_ops_base + "Reshape"
+        self.make_reshape(gate_reshape_name, [f"{gate_name}/output_0", f"{concat_name}/output_0"], dtype=self.io_dtype, shape=['num_rows', num_experts])
+
+        def quant_dequant(weights, quant_mode: bool = True):
+            type = torch.quint4x2 if quant_mode else torch.int8
+            # The following line is needed use torch.ops.trtllm._symmetric_quantize_last_axis_of_batched_matrix()
+            import tensorrt_llm
```
kunal-vaishnavi, line 1561: We should make it obvious why `tensorrt_llm` is needed for `QMoE` with the model builder. Should we wrap the import in a `try-except` block in case the user does not have it installed? We can print an error saying why it has to be installed for the error message.

src/python/py/models/builder.py
```diff
@@ -1495,6 +1506,125 @@ def make_mlp_fc(self, layer_id, mlp, root_input):
...
+        unsqueeze_name = gate_ops_base + "Unsqueeze"
+        self.make_unsqueeze(unsqueeze_name, [f"{gather_name}/output_0", "/model/constants/TensorProto.INT64/1D/0"], dtype=TensorProto.INT64, shape=[1])
+
+        concat_name = gate_ops_base + "Concat"
+        self.make_concat(concat_name, ["/model/constants/TensorProto.INT64/1D/-1", f"{unsqueeze_name}/output_0"], dtype=TensorProto.INT64, shape=[2], axis=0)
+
+        gate_reshape_name = gate_ops_base + "Reshape"
+        self.make_reshape(gate_reshape_name, [f"{gate_name}/output_0", f"{concat_name}/output_0"], dtype=self.io_dtype, shape=['num_rows', num_experts])
+
+        def quant_dequant(weights, quant_mode: bool = True):
```
kunal-vaishnavi, line 1558: Should we move this function outside of `make_block_sparse_moe` to allow other methods in the future to utilize it?

src/python/py/models/builder.py
```diff
@@ -1495,6 +1506,125 @@ def make_mlp_fc(self, layer_id, mlp, root_input):
...
+
+        moe_expert_weight_1_name = f"model.layers.{layer_id}.moe.weight_1"
+        moe_expert_weight_2_name = f"model.layers.{layer_id}.moe.weight_2"
+        moe_expert_weight_3_name = f"model.layers.{layer_id}.moe.weight_3"
+
+        moe_expert_scales_1_name = f"model.layers.{layer_id}.moe.scales_1"
+        moe_expert_scales_2_name = f"model.layers.{layer_id}.moe.scales_2"
+        moe_expert_scales_3_name = f"model.layers.{layer_id}.moe.scales_3"
+
+        def make_moe_external_tensor(w_list, moe_expert_name, numpy_type):
```
kunal-vaishnavi, line 1598: Should we move this function outside of `make_block_sparse_moe` to allow other methods in the future to utilize it?

src/python/py/models/builder.py
```diff
@@ -1495,6 +1506,125 @@ def make_mlp_fc(self, layer_id, mlp, root_input):
...
+
+        bias_ph = "" # Placeholder for bias
+        inputs = [root_input, f"{gate_reshape_name}/output_0", \
+                  moe_expert_weight_1_name, moe_expert_scales_1_name, bias_ph, \
+                  moe_expert_weight_2_name, moe_expert_scales_2_name, bias_ph, \
+                  moe_expert_weight_3_name, moe_expert_scales_3_name]
+        output = f"{moe_name}/output_0"
+
+        op_type = "QMoE"
+        self.make_node(op_type, inputs=inputs, outputs=[output], name=moe_name, domain="com.microsoft",
```
kunal-vaishnavi, line 1619: Can we create two generic methods called `make_moe` and `make_qmoe` and call those instead? We can use the logic similar to how `make_mlp` works.

src/python/py/models/builder.py
```diff
@@ -2602,6 +2769,7 @@ def get_args():
                 enable_cuda_graph = 1 : The model can use CUDA graph capture for CUDA execution provider. If enabled, all nodes being placed on the CUDA EP
                     is the prerequisite for the CUDA graph to be used correctly. It is not guaranteed that cuda graph be enabled as it depends on the model
                     and the graph structure.
+                use_8bits_moe = 1 : Use 8-bit quantization for MoE layers. Default is using 4-bit quantization.
```
kunal-vaishnavi, line 2772: Can you add an example for `use_8bits_moe` in `README.md` [after the CUDA graph example](https://github.com/microsoft/onnxruntime-genai/tree/main/src/python/py/models#enable-cuda-graph)?.

src/python/py/models/builder.py
```diff
@@ -2491,6 +2653,11 @@ def create_model(model_name, input_path, output_dir, precision, execution_provid
             onnx_model = Phi3Mini4KModel(config, io_dtype, precision, execution_provider, cache_dir, extra_options)
         elif config.architectures[0] == "Phi3ForCausalLM" and config.max_position_embeddings == 131072:
             onnx_model = Phi3Mini128KModel(config, io_dtype, precision, execution_provider, cache_dir, extra_options)
+        elif config.architectures[0] == "PhiMoEForCausalLM" and config.max_position_embeddings == 131072:
```
kunal-vaishnavi, line 2656: The `config.max_position_embeddings == 131072` check exists for Phi-3 mini to distinguish between the 4K and 128K variants. Since Phi-3.5 MoE has only one variant, I think we can remove that part of the check.

src/python/py/models/builder.py
```diff
@@ -2447,6 +2577,38 @@ def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
...
+        self.rotemb_attrs["rotemb_name"] = "rotary_emb"
+
+        self.layernorm_attrs["simple"] = False
+
+        self.moe_attrs["num_experts"] = 16
+        self.moe_attrs["top_k"] = 2
+        self.moe_attrs["activation_type"] = "silu"
+        self.moe_attrs["normalize_routing_weights"] = 0
+        self.moe_attrs["use_sparse_mixer"] = 1
+        self.moe_attrs["use_int4"] = 0 if "use_8bits_moe" in extra_options and extra_options["use_8bits_moe"] == "1" else 1
```
kunal-vaishnavi, line 2594:
```suggestion
        self.moe_attrs["use_int4"] = 0 if "use_8bits_moe" in extra_options else 1
```

Even though the documentation shows the official way to select extra options, users seem to use a wide set of versions to denote that they want to use an extra option (e.g. `use_8bits_moe=1`, `use_8bits_moe=true`, `use_8bits_moe=True`, etc). I think we can just check if the user passed the option or not to understand that they want to utilize it.

src/python/py/models/builder.py
```diff
@@ -2447,6 +2577,38 @@ def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
...
+        super().__init__(config, io_dtype, onnx_dtype, ep, cache_dir, extra_options)
+        assert io_dtype == TensorProto.FLOAT16, "This model only supports float16 io type."
+
+        self.rotemb_attrs["rotemb_name"] = "rotary_emb"
+
+        self.layernorm_attrs["simple"] = False
+
+        self.moe_attrs["num_experts"] = 16
+        self.moe_attrs["top_k"] = 2
+        self.moe_attrs["activation_type"] = "silu"
```
kunal-vaishnavi, line 2591: Once these are auto-added when `moe_attrs` is created, I think these entries can be removed.

kunal-vaishnavi, line 2591: Once these are [auto-added](https://github.com/microsoft/onnxruntime-genai/pull/850/files#r1742802952) when `moe_attrs` is created, I think these entries can be removed.

src/python/py/models/builder.py
```diff
@@ -2447,6 +2577,38 @@ def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
         super().__init__(config, io_dtype, onnx_dtype, ep, cache_dir, extra_options)
 
 
+class Phi3MoE128KModel(MistralModel):
+    def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
+        super().__init__(config, io_dtype, onnx_dtype, ep, cache_dir, extra_options)
+        assert io_dtype == TensorProto.FLOAT16, "This model only supports float16 io type."
+
+        self.rotemb_attrs["rotemb_name"] = "rotary_emb"
```
kunal-vaishnavi, line 2585: I think this can be removed since it isn't used anywhere.

src/python/py/models/builder.py
```diff
@@ -2491,6 +2653,11 @@ def create_model(model_name, input_path, output_dir, precision, execution_provid
             onnx_model = Phi3Mini4KModel(config, io_dtype, precision, execution_provider, cache_dir, extra_options)
         elif config.architectures[0] == "Phi3ForCausalLM" and config.max_position_embeddings == 131072:
             onnx_model = Phi3Mini128KModel(config, io_dtype, precision, execution_provider, cache_dir, extra_options)
+        elif config.architectures[0] == "PhiMoEForCausalLM" and config.max_position_embeddings == 131072:
+            print("WARNING: This model only works for CUDA currently because `MoE` is only supported for CUDA in ONNX Runtime. Setting `--execution_provider cuda` by default.")
+            print("WARNING: This model currently only supports quantized version.")
+            execution_provider = "cuda"
+            onnx_model = Phi3MoE128KModel(config, io_dtype, precision, execution_provider, cache_dir, extra_options)
```
kunal-vaishnavi, line 2660:
```suggestion
            onnx_model = Phi3MoEModel(config, io_dtype, precision, execution_provider, cache_dir, extra_options)
```

src/python/py/models/builder.py
wangyems: Will add MoE when GenAI tool supports multi-gpu

src/python/py/models/builder.py
wangyems: this is currently unique to moe

src/python/py/models/builder.py
wangyems: This is unique to moe as of now. we can refactor it later if needed.

src/python/py/models/builder.py
wangyems: will refactor it when adding the sharded moe part

src/python/py/models/builder.py
wangyems: the team shared the 4K version to us before. just want to make it clear that it's the 128k version.

src/python/py/models/builder.py
wangyems: let's use the 128K for now

src/python/py/models/builder.py
wangyems: sure

src/python/py/models/builder.py
wangyems: sure

src/python/py/models/builder.py
wangyems: A corner case is that user may add use_8bits_moe=0 which causes logic flaw. we can do some checks for extra_options

src/python/py/models/builder.py
wangyems: I have concerns about the naming issue with autofill. e.g. activation_type is named hidden_act in phi3 moe, top_k is named num_experts_per_tok. these are not standard names as others. I prefer manually setting the config for each model class instead of doing naming check in the autofill process.

src/python/py/models/README.md
```diff
@@ -172,6 +172,18 @@ python3 -m onnxruntime_genai.models.builder -i path_to_local_folder_on_disk -o p
 python3 builder.py -i path_to_local_folder_on_disk -o path_to_output_folder -p precision -e execution_provider -c cache_dir_to_store_temp_files --extra_options enable_cuda_graph=1
 ```
 
+#### Use 8 Bits Quantization in QMoE
```
kunal-vaishnavi, line 175: Can you add a link to this section after `Enable Cuda Graph`?

https://github.com/microsoft/onnxruntime-genai/blob/a477802a411cc1ab77f5c23834ca807081e20ed9/src/python/py/models/README.md?plain=1#L15-L19

src/python/py/models/builder.py
```diff
@@ -1495,6 +1506,128 @@ def make_mlp_fc(self, layer_id, mlp, root_input):
...
+        #           |        |     |
+        #         Gather     |     |
+        #           |        |     |
+        #       Unsqueeze    |     |
+        #           |        |    /
+        #        Concat     /    /
+        #             \    /    /
+        #             Reshape  /
+        #                 \   /
+        #                  MoE
```
kunal-vaishnavi, line 1526:
```suggestion
        #                  QMoE
```

src/python/py/models/builder.py
```diff
@@ -1495,6 +1506,128 @@ def make_mlp_fc(self, layer_id, mlp, root_input):
...
+            type = torch.quint4x2 if quant_mode else torch.int8
+
+            try:
+                import tensorrt_llm
+
+                _, processed_q_weight, torch_weight_scales = (
+                    torch.ops.trtllm._symmetric_quantize_last_axis_of_batched_matrix(weights.T.cpu().contiguous(), type)
+                )
+            except:
+                print("import tensorrt_llm is needed to use torch.ops.trtllm._symmetric_quantize_last_axis_of_batched_matrix()")
```
kunal-vaishnavi, line 1568: If `tensorrt_llm` is not installed and it is required to continue, should we raise an `Exception` (or an `ImportError`) instead and print this message?

kunal-vaishnavi, line 1568: If `tensorrt_llm` is not installed and it is required to continue, should we raise an `Exception` (or an `ImportError`) instead and add this message there?

src/python/py/models/builder.py
```diff
@@ -1495,6 +1506,128 @@ def make_mlp_fc(self, layer_id, mlp, root_input):
         # Assign output 0 of MLP layer as output of last layer
         self.mlp_attrs["output_0"] = f"{fc2_add_name}/output_0"
 
+    def make_block_sparse_moe(self, layer_id, bsm, root_input):
+        # Make nodes for the MoE subgraph
```
kunal-vaishnavi, line 1510:
```suggestion
        # Make nodes for the QMoE subgraph
```

src/python/py/models/builder.py
wangyems: without trtllm package,  torch.ops.trtllm._symmetric_quantize_last_axis_of_batched_matrix() will throw error that saying .trtllm is not existed in torch. I think we just need an extra message to tell the user how to proceed instead of throwing again.

kunal-vaishnavi: But with the try-except block, that error won't get thrown. Instead it will print this message and continue to this line.

https://github.com/microsoft/onnxruntime-genai/blob/234db941273c66bb9256f802c1648436b87cf364/src/python/py/models/builder.py#L1571

Since `torch_weight_scales = None`, a `NoneType` error will be thrown instead.

kunal-vaishnavi: I just ran it to verify that would be the case.

```
Reading embedding layer
Reading decoder layer 0
import tensorrt_llm is needed to use torch.ops.trtllm._symmetric_quantize_last_axis_of_batched_matrix()
  File "/path/to/onnxruntime-genai/src/python/py/models/builder.py", line 2794, in <module>
    create_model(args.model_name, args.input, args.output, args.precision, args.execution_provider, args.cache_dir, **extra_options)
  File "/path/to/onnxruntime-genai/src/python/py/models/builder.py", line 2686, in create_model
    onnx_model.make_model(input_path)
  File "/path/to/onnxruntime-genai/src/python/py/models/builder.py", line 1768, in make_model
    self.make_layer(self.layer_id, module)
  File "/path/to/onnxruntime-genai/src/python/py/models/builder.py", line 2606, in make_layer
    self.make_block_sparse_moe(layer_id, layer.block_sparse_moe, root_input=self.layernorm_attrs["output_0"])
  File "/path/to/onnxruntime-genai/src/python/py/models/builder.py", line 1582, in make_block_sparse_moe
    w1_scale, pre_qweight1= quant_dequant(bsm.experts[i].w1.weight, use_int4)
  File "/path/to/onnxruntime-genai/src/python/py/models/builder.py", line 1571, in quant_dequant
    return torch_weight_scales.to(torch.float16), processed_q_weight
AttributeError: 'NoneType' object has no attribute 'to'
```

The `AttributeError: 'NoneType' object has no attribute 'to'` message is misleading so I think it would be better to raise an exception and print the `tensorrt_llm` message as part of the exception.

wangyems: you are right, let me change it

src/python/py/models/builder.py
```diff
@@ -2491,6 +2664,11 @@ def create_model(model_name, input_path, output_dir, precision, execution_provid
             onnx_model = Phi3Mini4KModel(config, io_dtype, precision, execution_provider, cache_dir, extra_options)
         elif config.architectures[0] == "Phi3ForCausalLM" and config.max_position_embeddings == 131072:
             onnx_model = Phi3Mini128KModel(config, io_dtype, precision, execution_provider, cache_dir, extra_options)
+        elif config.architectures[0] == "PhiMoEForCausalLM" and config.max_position_embeddings == 131072:
+            print("WARNING: This model only works for CUDA currently because `MoE` is only supported for CUDA in ONNX Runtime. Setting `--execution_provider cuda` by default.")
+            print("WARNING: This model currently only supports quantized version.")
+            execution_provider = "cuda"
```
kunal-vaishnavi, line 2670: Should we set `precision = "int4"` as well since that's the only option?

wangyems: fp16 should also work but yeah let me hardcode it to int4 for less ambiguity

kunal-vaishnavi (approved):

PR #853, opened 2024-08-29
refine the benchmark script

yufenglee:
1. add check for max length
2. speed up the warming up
3. fix the prompt generation with greedy search

### DIFF
--- a/benchmark/python/benchmark_e2e.py
+++ b/benchmark/python/benchmark_e2e.py
@@ -66,11 +66,10 @@ def monitor_cpu_memory():
 
 # Use input model to generate prompt
 def generate_prompt(model, tokenizer, prompt_length, use_graph_capture) -> str:
-    temperature = 1.0
     prompt = "a"
     tokens = tokenizer.encode(prompt)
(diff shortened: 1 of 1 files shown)


benchmark/python/benchmark_e2e.py
```diff
@@ -226,9 +226,8 @@ def run_benchmark(args, batch_size, prompt_length, generation_length, max_length
     if args.verbose: print("Running warmup runs...")
     for _ in tqdm(range(args.warmup)):
         generator = og.Generator(model, params)
-        while not generator.is_done():
-            generator.compute_logits()
-            generator.generate_next_token()
+        generator.compute_logits()
```
RyanUnderhill, line 229: Why is there no more generation loop? It seems to just run once? Is this just to "process the prompt tokens" and not to generate any tokens beyond it?

yufenglee: it is for warming up. ORT can be warmed up by just running the prompt

RyanUnderhill (approved):

PR #864, opened 2024-09-03
Clarify build availability of Arm64 now Windows Arm64 builds are available

sirredbeard:
With 0.4.0 builds are available for Arm on Windows, but building for Arm64 on other platforms still requires building from source


sirredbeard:

skyline75489:
Thanks for the contribution! I think @natke would like this :)

yufenglee (approved):

PR #867, opened 2024-09-03
Continuous Decoding and System Prompt Preprocessing

aciddelgado:
#### Overview
This PR introduces a flexible solution to enable continuous decoding and system prompt preprocessing. Continuous decoding supports multi-turn conversations, while system prompt preprocessing processes the application's system prompt before the user inputs their prompt. The following changes are proposed:

#### Changes
1. **New Method: `generator.append_tokens(List: token_ids)`**
   - Appends system/user tokens to internal `input_ids`.
   - Opportunistically runs computation on the given `token_ids`.
   - Replaces `params.input_ids`.

2. **New Method: `generator.rewind_to(int: n_tokens)`**
   - Rewinds the KV-cache state to `n_tokens`.
   - Allows for reuse of KV-cache memory.
   - Enables rewinding to the system prompt between generation cycles.

3. **Deprecation: `generator.compute_logits()`**
   - This method is deprecated as it is unnecessary and confusing for users to call manually.
   - The model will compute/run:
     - At the end of an `append_tokens()` call.
     - At the beginning of a `generate_next_token()` call.

4. **Deprecation: `generator_params.input_ids`**
   - Input IDs are deprecated, and their functionality is replaced by `append_tokens`. This way, tokens are added to the generator rather than the parameters.

#### Limitations
- **append_tokens:** Continuous decoding and system prompt preprocessing only work for `batch size == 1`. When `batch_size > 1`, `append_tokens` can only be invoked once without error. For `batch size == 1`, `append_tokens` can be called arbitrarily to enable the application designer to create their own continuous decoding/system prompt preprocessing solution.
- **rewind_to:** For `batch size > 1`, `rewind_to` can only rewind to index 0, resetting the generation state. For `batch_size == 1`, however, `rewind_to` can rewind to any point in the conversation.
- **Model types:** Currently, `append_tokens` and `rewind_to` are not supported for multi-modal models like phi-3v and whisper. Please use `generator_params.set_inputs()` to provide inputs for these model types. In the future, this workflow will likely change.

#### Examples
- Python example with continuous decoding and system prompt preprocessing
  - https://github.com/microsoft/onnxruntime-genai/blob/aciddelgado/continuous/examples/python/model-qa.py
- Python example with batch_size > 1
  - https://github.com/microsoft/onnxruntime-genai/blob/aciddelgado/continuous/examples/python/model-generate.py


src/cuda/search_cuda.cpp
```diff
@@ -216,6 +205,44 @@ std::span<float> Search_Cuda::GetScores() {
...
+// Set user input tokens (batch_beam_size, sequence_length)
+void GreedySearch_Cuda::SetUserTokens(DeviceSpan<int32_t>& next_tokens) {
+  cudaMemsetAsync(eos_meet_.data(), 0, eos_meet_.size_bytes(), params_->cuda_stream);
+  *done_cpu_ = false;
+
+  auto next_tokens_gpu = const_cast<DeviceSpan<int32_t>&>(next_tokens).Span();
+  cuda::Launch_AppendNextTokensToSequences(next_tokens_gpu, sequences_.GetSequences().Span(), params_->BatchBeamSize(), sequences_.GetSequenceLength(), sequences_.max_length_, GetStream());
+  sequences_.AfterAppendNextTokens(next_tokens, params_->BatchBeamSize());
+
+  if (sequences_.GetSequenceLength() == params_->search.max_length) {
```
RyanUnderhill, line 217: If next_tokens goes past the max_length, what happens?

src/cuda/search_cuda.h
```diff
@@ -46,6 +46,8 @@ struct GreedySearch_Cuda : Search_Cuda {
   void SampleTopK(int k, float t) override { SampleTopKTopP(k, 0.0, t); }
   void SampleTopP(float p, float t) override { SampleTopKTopP(-1, p, t); }
   void SampleTopKTopP(int k, float p, float t) override;
+  void SetUserTokens(DeviceSpan<int32_t>& next_tokens) override;  // shape (batch_size, sequence_length)
```
RyanUnderhill, line 49: Should it be AppendUserTokens vs SetUserTokens? Also what does the 'User' part mean?

src/generators.cpp
```diff
@@ -354,28 +395,32 @@ void Generator::GenerateNextToken() {
...
 
-DeviceSpan<int32_t> Generator::GetSequence(size_t index) const {
-  return search_->GetSequence(index);
+void Generator::RewindToLength(size_t new_length) {
+  if (model_->config_->model.type == "whisper" || model_->config_->model.type == "phi3v" || model_->config_->model.type == "decoder-pipeline")
+    throw std::runtime_error("RewindTo is currently not supported for " + model_->config_->model.type + ".");
+  if (new_length > search_->GetSequenceLength())
+    throw std::runtime_error("Cannot rewind to a length greater than the current sequence length");
+  if (new_length == search_->GetSequenceLength())
+    return;
```
RyanUnderhill, line 404:
```suggestion
    return;  // Nothing to do, length doesn't change
```

What happens if you don't early exit on this check? If it works fine and it's not costly, we should just remove the check.

src/search.cpp
```diff
@@ -277,6 +267,56 @@ void GreedySearch_Cpu::AppendNextTokensToSequences() {
   }
 }
 
+void GreedySearch_Cpu::SetUserTokens(DeviceSpan<int32_t>& next_tokens) {
+  // Set user-defined next tokens
+  auto next_tokens_cpu = const_cast<DeviceSpan<int32_t>&>(next_tokens).Span();
```
RyanUnderhill, line 272: Why this const_cast?

src/search.h
```diff
@@ -53,7 +58,7 @@ struct Search_Cpu : Search {
 
   DeviceSpan<float> next_token_scores_;  // shape (beam_size*batch_size, vocab_size)
 
-  bool done_{};
+  bool done_{false};
```
RyanUnderhill, line 61:
```suggestion
  bool done_{};
```

Don't fight the zero init, embrace it.

src/generators.cpp
aciddelgado: It will set done to fasle. I think this is the best way to ensure it does nothing when we rewind to the current sequence length

kunal-vaishnavi:
I have some questions:

1. For models such as Whisper where we want to run custom logic on the logits before generating the next token, how will this be done with the deprecation of `generator.compute_logits()`? Previously, users could do the following.

```python
    generator.compute_logits()
    logits = generator.get_logits()
    ...
    # modify logits with custom logic
    ...
    generator.set_logits(logits)
```

2. Can `model.generate()` still be supported rather than deprecated? It is the highest-level API that is currently offered and it matches Hugging Face's `model.generate()` API. There are also several [video](https://www.youtube.com/watch?v=tNq9MP9wlBg) [tutorials](https://www.youtube.com/watch?v=Wp5PaRpudlk) that already use it. I think it is fine to not support this PR's features through `model.generate()` and we can print a message to encourage users to use the lower-level APIs instead for these features. For new users who want to quickly try out ONNX Runtime GenAI, `model.generate()` is the closest API to use when transitioning from Hugging Face's APIs that they may be more familiar with.

3. Can you fix the Whisper example so that the original sequence length is still passed into the CUDA kernels? I think we can support this PR's features for Whisper after my rewrite.

4. It appears that `params_->batch_size` is renamed to `params_->search.batch_size`. Does the `genai_config.json` need to have a `batch_size` entry now?

5. For multi-modal models, `set_inputs()` only works with `NamedTensors` produced by our pre-processors. For Phi-3 vision/Phi-3.5 vision, this is not an issue because the pre-processor [prepares all of the inputs](https://github.com/microsoft/onnxruntime-genai/blob/ec89e49542b168072836a2091fc66ed65d580a86/examples/python/phi3v.py#L55-L59). However, this is [not the case](https://github.com/microsoft/onnxruntime-genai/blob/ec89e49542b168072836a2091fc66ed65d580a86/examples/python/whisper.py#L42-L55) for Whisper. Since Whisper's pre-processor doesn't produce input IDs, the removal of `params.input_ids` would break it. For external users, can we add input IDs to Whisper's pre-processor so that `set_inputs` can be used? For internal users, `params.input_ids` will be calculated outside of ONNX Runtime GenAI. How should we set those inputs with the removal of `params.input_ids` since `set_inputs` only works for `NamedTensors`?

6. Is `generator.generate_next_token()` supposed to be called twice here? The comment says to check for the first token generated but two tokens will have been generated.

https://github.com/microsoft/onnxruntime-genai/blob/d1038f8ffc47dc20344790b13714c44b18e4309d/test/python/test_onnxruntime_genai_api.py#L345-L348

examples/python/model-qa.py
```diff
@@ -93,5 +100,7 @@ def main(args):
     parser.add_argument('-v', '--verbose', action='store_true', default=False, help='Print verbose output and timing information. Defaults to false')
     parser.add_argument('-g', '--timings', action='store_true', default=False, help='Print timing information for each generation step. Defaults to false')
     parser.add_argument('-c', '--chat_template', type=str, default='', help='Chat template to use for the prompt. User input will be injected into {input}')
+    parser.add_argument('-s', '--system_prompt', type=str, default='You are a helpful assistant. You are friendly, courteous, and professional. All your responses must end with an exclamation point!', help='System prompt to use for the prompt.')
+    parser.add_argument('-re', '--rewind', action='store_true', default=False, help='Rewind to the system prompt after each generation. Defaults to false')
```
kunal-vaishnavi, line 104:
```suggestion
    parser.add_argument('-r', '--rewind', action='store_true', default=False, help='Rewind to the system prompt after each generation. Defaults to false')
```

aciddelgado:
@kunal-vaishnavi 
To answer your questions:

1. In order to implement the workflow described, this is how you could modify the code snippet given:
`generator.append_tokens(prompt_tokens)
    logits = generator.get_logits()
    ...
    # modify logits with custom logic
    ...
    generator.set_logits(logits)
    generator.generate_next_token()`
This works because get_logits will automatically compute logits if they haven't been computed already.

2. As a team we had decided to remove `model.generate()`. The reason is that input_ids has been removed from generator_params. However, if the team decides it's a good idea to reimplement `model.generate` I can do so. We can pass the prompt tokens like so `model.generate(prompt_tokens)` or we can reimplement input_ids but I think the latter would be confusing.
3. Will discuss offline.
4. Yes, I will specify that in the PR description now that you mention it.
5. I'm not familiar with how Whisper works so we can discuss offline. However, to clarify, This PR intends to include two workflows. The first is for Decoder Only models (Phi-3, LLama, etc.) and uses the `generator.append_tokens(prompt)` API to add input ids. The second is for multi-modal models (Phi-3v, etc.) and uses `parameters.set_inputs()` API to add inputs. We can figure out how to support Whisper cooperatively.
6. Yes, this is because we are measuring how long it takes to compute the logits and generate the token. The first generate_next_token call simply runs a search on logits already generated by the previous append_tokens call. The second one actually runs the model and also runs a search.

kunal-vaishnavi:
All of this sounds great. Regarding `model.generate()`, I'm fine with deprecating it as well since the while loop approach is only an extra line of code and it gives much greater flexibility to the user. But I think we will need to update our documentation in a lot of places (this repo, ORT docs site, any blogs, etc.) to make this change very clear.

For example, the name of this repo is "ONNX Runtime generate() API" and we have the following sentence in the main README: "You can call a high level generate() method to generate all of the output at once, or stream the output one token at a time." Neither of these will be accurate since `generate()` won't exist anymore.

src/models/kv_cache.cpp
```diff
@@ -213,25 +267,83 @@ void KV_Cache::Add() {
...
+    OrtValue& present = *presents_[i];
+    std::unique_ptr<OrtValue> past = OrtValue::CreateTensor(*model_.allocator_kvcache_, shape_, type_);
+    for (int j = 0; j < batch_x_num_heads; j++) {
+      auto present_data = present.GetTensorData<T>() + j * old_length_x_head_size;
+      auto past_data = past->GetTensorMutableData<T>() + j * new_length_x_head_size;
+#if USE_CUDA
+      if (model_.device_type_ == DeviceType::CUDA) {
+        cudaMemcpyAsync(past_data, present_data, new_length_x_head_size * sizeof(T), cudaMemcpyDeviceToDevice, model_.cuda_stream_);
+      } else
+#elif USE_DML
```
baijumeswani, line 335: It is possible that we may enable use cuda and use dml together. Do you want to handle this differently here?

baijumeswani (approved):

tarekziade:
This breaks the provided examples throughout the documentation, because you've removed `params.input_ids` and `generator.compute_logits`

PR #882, opened 2024-09-09
Add re-designed support for Phi-3 vision and Phi-3.5 vision

kunal-vaishnavi:
### Description
This PR supports the re-designed export of Phi-3 vision and Phi-3.5 vision. The new design natively supports multi-image and the `select` logic inside the ONNX models.

### Motivation and Context
With the re-designed export, some of the logic inside ONNX Runtime GenAI is no longer needed as it is now inside the ONNX model. This allows other models to more easily re-use the vision and embedding components within ONNX Runtime GenAI.


examples/python/phi3v.py
```diff
@@ -29,9 +28,10 @@ def run(args: argparse.Namespace):
                 "Image Path (comma separated; leave empty if no image): "
             ).split(",")
         ]
+        image_paths = [image_path for image_path in image_paths if len(image_path)]
```
baijumeswani, line 31: Why is this needed?

src/models/image_features.cpp
```diff
@@ -0,0 +1,66 @@
+#include "../generators.h"
```
baijumeswani, line 1: license.

src/models/image_features.cpp
```diff
@@ -0,0 +1,66 @@
...
+  // 4) Created as an input for embedding model (num_image_tokens = 0)
+  //    The tensor does not need to be pre-allocated because it will be created during (2).
+  if (mode == ImageFeatures::Mode::Output) {
+    image_features_ = OrtValue::CreateTensor(*model_.allocator_device_, shape_, type_);
+  }
+}
+
+void ImageFeatures::Add() {
+  if (mode_ == ImageFeatures::Mode::Input) {
+    // In case the input_features are an input to a model, they are added
```
baijumeswani, line 34: Shall we call is image_features wherever possible?

src/models/multi_modal_vision_model.cpp
```diff
@@ -104,41 +30,19 @@ int64_t GetNumImageTokens(const std::vector<GeneratorParams::Input>& extra_input
     return 0;
   }
 
-  if (image_sizes->ort_tensor_->GetTensorTypeAndShapeInfo()->GetShape() != std::vector<int64_t>{1, 2}) {
-    throw std::runtime_error("image_sizes tensor must have 2 elements");
+  auto num_images = pixel_values->ort_tensor_->GetTensorTypeAndShapeInfo()->GetShape()[0];
+  if (image_sizes->ort_tensor_->GetTensorTypeAndShapeInfo()->GetShape() != std::vector<int64_t>{num_images, 2}) {
+    throw std::runtime_error("image_sizes tensor must be of shape (num_images, 2)");
```
baijumeswani, line 35: Can we also provide details on what num_images is here in the error?

src/models/multi_modal_vision_model.cpp
```diff
@@ -104,41 +30,19 @@ int64_t GetNumImageTokens(const std::vector<GeneratorParams::Input>& extra_input
...
-                                            const std::string& visual_features_name, int32_t hidden_size,
-                                            int64_t num_image_tokens) {
-  constexpr int32_t batch_size = 1;
-  if (!session_info.HasOutput(visual_features_name)) {
-    throw std::runtime_error("Visual features output not found in the model");
+  int64_t num_image_tokens = 0;
+  for (int i = 0; i < num_images; i++) {
+    int64_t h = image_sizes_data[i * num_images] / 336;
+    int64_t w = image_sizes_data[i * num_images + 1] / 336;
+    num_image_tokens += static_cast<int64_t>((h * w + 1) * 144) + 1 + ((h + 1) * 12);
```
baijumeswani, line 43: I don't think we need the static_cast here since h and w and num_image_tokens are all int64_t.

src/models/multi_modal_vision_model.cpp
```diff
@@ -104,41 +30,19 @@ int64_t GetNumImageTokens(const std::vector<GeneratorParams::Input>& extra_input
...
-  return ((h * w + 1) * 144) + 1 + ((h + 1) * 12);
-}
-
-std::unique_ptr<OrtValue> GetVisualFeatures(OrtAllocator& device_allocator, const SessionInfo& session_info,
-                                            const std::string& visual_features_name, int32_t hidden_size,
-                                            int64_t num_image_tokens) {
-  constexpr int32_t batch_size = 1;
-  if (!session_info.HasOutput(visual_features_name)) {
-    throw std::runtime_error("Visual features output not found in the model");
+  int64_t num_image_tokens = 0;
```
baijumeswani, line 39: We can do it later. But we should think about how to make this Model State execution not specific to phi3 vision.
Maybe we can move this num_image_tokens computation to the vision preprocessing or some other place. 

But we can do this later on when we add support for other vision models.

src/models/multi_modal_vision_model.cpp
```diff
@@ -185,22 +92,17 @@ RoamingArray<float> EmbeddingState::Run(int current_length, RoamingArray<int32_t
...
-                                         params_->hidden_size, num_image_tokens_);
-    output_names_.push_back(model_.config_->model.vision.outputs.visual_features.c_str());
-    outputs_.push_back(visual_features_.get());
-  }
+  image_features_.Add();
 }
 
 RoamingArray<float> VisionState::Run(int current_length, RoamingArray<int32_t> next_tokens, RoamingArray<int32_t> next_indices) {
-  State::Run(*model_.vision_session_, *model_.run_options_, 1);
+  int num_images = static_cast<int>(inputs_[0]->GetTensorTypeAndShapeInfo()->GetShape()[0]);
```
baijumeswani, line 104: Is the batch size dimension = num_images?
Is it possible to run this batched? Multiple images in a batch?

baijumeswani, line 104:
```suggestion
  const int num_images = static_cast<int>(inputs_[0]->GetTensorTypeAndShapeInfo()->GetShape()[0]);
```

src/models/image_features.cpp
```diff
@@ -0,0 +1,66 @@
...
+    state_.input_names_.push_back(name_.c_str());
+  } else {
+    index_ = state_.outputs_.size();
+    state_.outputs_.push_back(image_features_.get());
+    state_.output_names_.push_back(name_.c_str());
+  }
+}
+
+void ImageFeatures::Update() {
+  // Initialize empty image_features tensor for no-image or after-prompt input scenarios
```
baijumeswani, line 48: This function is only called after prompt processing. The comment here can reflect only the second part:

```
Initialize empty image_features tensor after-prompt input scenarios
```

You can highlight that the num_image_tokens will be 0 when no image is provided in the constructor of ImageFeatures where we do the buffer allocation.

test/test_models/create_dummy_model.py
```diff
@@ -0,0 +1,128 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation.  All rights reserved.
+# Licensed under the MIT License.  See License.txt in the project root for
+# license information.
+# --------------------------------------------------------------------------
+"""
+Create dummy ONNX models that contain only inputs and outputs
```
baijumeswani, line 7: Could you also add a comment here detailing why a dummy model with input output names might be needed?

test/test_models/images/10809054.jpg
baijumeswani, line 1: Seems like this image is not being used anywhere. Do we need it?

examples/python/phi3v.py
kunal-vaishnavi: The image paths are otherwise passed as empty strings to the processor and the following error occurs.

```
Loading model...
Image Path (comma separated; leave empty if no image): 
['']
Loading images...
  File "/path/to/onnxruntime-genai/examples/python/phi3v.py", line 79, in <module>
    run(args)
  File "/path/to/onnxruntime-genai/examples/python/phi3v.py", line 42, in run
    raise FileNotFoundError(f"Image file not found: {image_path}")
FileNotFoundError: Image file not found:
```

src/models/image_features.cpp
kunal-vaishnavi: Added

src/models/image_features.cpp
kunal-vaishnavi: This is using the same logic as how the embeddings are added.

https://github.com/microsoft/onnxruntime-genai/blob/dac90d557dbd90c816cb4a44f39367346bca1c1c/src/models/embeddings.cpp#L33-L46

src/models/multi_modal_vision_model.cpp
kunal-vaishnavi: Added the user's input shape to the error message

src/models/multi_modal_vision_model.cpp
kunal-vaishnavi: Removed

src/models/multi_modal_vision_model.cpp
kunal-vaishnavi: I agree. Besides the `num_image_tokens` calculation, I think everything else is generic enough to work for other models.

src/models/multi_modal_vision_model.cpp
kunal-vaishnavi: The two dimensions are different. For example, you can pass 2 images and 1 prompt. Then `num_images = 2` in the vision component and `batch_size = 1` in the text component.

kunal-vaishnavi: Fixed

src/models/image_features.cpp
kunal-vaishnavi: Changed

test/test_models/create_dummy_model.py
kunal-vaishnavi: Added

test/test_models/images/10809054.jpg
kunal-vaishnavi: I used it for testing the multi-image scenario. I think it is okay to keep since there are only two images in the test folder, and it gives us another image to test vision models with.

examples/python/phi3v.py
baijumeswani: I see. This would happen if we provide empty spaces in the file paths without listing the files. Cool, thanks.

src/models/multi_modal_vision_model.cpp
baijumeswani: So the vision component cannot be run with a batch_size > 1 for now, right?

baijumeswani (approved):

src/models/multi_modal_vision_model.cpp
kunal-vaishnavi: The `num_images` is the batch size for the vision component and the `batch_size` is the batch size for the text component. With these changes, you can run this model batched (i.e. multiple images in a batch) as well. The processor just has to prepare the inputs correctly.

src/config.cpp
```diff
@@ -248,8 +244,8 @@ struct VisionOutputs_Element : JSON::Element {
   explicit VisionOutputs_Element(Config::Model::Vision::Outputs& v) : v_{v} {}
 
   void OnString(std::string_view name, std::string_view value) override {
-    if (name == "visual_features") {
-      v_.visual_features = value;
+    if (name == "image_features") {
```
yufenglee, line 247: Will this change break the old models?

kunal-vaishnavi: It will and it is intentional. The old design is quite restrictive in terms of both obtaining and running the vision models. The new design makes it easier to obtain and run the vision models. It also allows users to run Phi-3 vision and Phi-3.5 vision with multiple images, which was not previously supported in the uploaded ONNX models as well as ONNX Runtime GenAI. These changes also make the multi-modal pipeline more generic to onboard new vision models.

Since backwards compatibility is not fully guaranteed, it should be fine to make this change. We will publish updated ONNX models for Phi-3 vision along with the Phi-3.5 vision ONNX models.

Issue #772, opened 2024-08-09
Nightly builds


jarroddavis68:
Can you make the nightly builds available?

baijumeswani:
This is something we would like to add soon.
It requires creating scheduled packaging and publishing jobs and a feed where the packages can be hosted.

While I do not have a timeline that I can share for this, I think we will certainly get to this in the near future.

jarroddavis68:
Thanks! It would allow me to have the latest fixes. I use Delphi. I can usually get a lib built, but for whatever reason I just cannot seem to get this one to build properly. So, for me, the nightly builds are most important. Thanks again.

natke:
While we work on nightly builds, can you share the issue you are having with building? Perhaps there is something we can help you with there

jarroddavis68:
> While we work on nightly builds, can you share the issue you are having with building? Perhaps there is something we can help you with there

I primarily work with Delphi, so I require the project to be in DLL form for integration. I avoid using C/C++ unless absolutely necessary. Typically, if a CMake project can handle all dependencies during the generation of the Visual Studio solution, I can compile it successfully. However, this particular project doesn't seem to manage dependencies as seamlessly, or at least I haven't been able to get it working properly.

For instance, with the llama.cpp project, I have a script that pulls the latest changes from the repository. I can then use CMake to generate a Visual Studio 2022 solution, which automatically handles all dependencies. From there, I can compile the project into static libraries and incorporate these libraries into my project's dependency DLL, which exports the routines from the various C libraries used in my project.

With this current project, it appears that some dependencies need to be placed manually in specific folders, and the dependency checks performed by CMake don't seem to complete successfully. I’m wondering why CMake can't be configured to automatically download and correctly place all necessary dependencies to streamline the process. My goal is to compile the ONNX DLLs as static libraries so that I can include them in my project's dependency DLL.

elephantpanda:
I would would be happy with weekly or every 2 weeks. 
I was looking at the gap between version 0.30 and 0.40 and that looks like 2 months. In this fast moving world of AI that seems like a lifetime 🙂

It would also be quite nice to see somewhere a list of the bugs found, if they could be reproduced or not, and the priority. There's 54 issues open at the moment so it's not clear which are being worked on and which aren't. And what the timelines might be. May be some "high" or "low" priority tags might be useful.

That's me 2 cents 😊

jarroddavis68:
[...]

PR #890, opened 2024-09-12
Auto download ort at compile time

baijumeswani:
Up until now, the `onnxruntime-genai` build required users to manually provide the path to the ort home folder. The ort home folder was expected to contain the headers inside the `include` directory and the libraries inside the `lib` directory. This was cumbersome to the developer since it required an extra step for building ort. Moreover, it made libraries depending on `onnxruntime-genai` unable to build the project with ease since it required this extra step as can be seen in [this issue](https://github.com/microsoft/onnxruntime-genai/issues/772#issuecomment-2313811101).

With this pull-request, the onnxruntime-genai dependencies are auto downloaded and used for compilation without the user needing to know where to fetch those from. The user can alternatively provide a local ort folder if they wish to override the default version used during compilation using the `--ort_home` build arg.


snnn (approved):
Thank you!

PR #898, opened 2024-09-18
Added extra Option to To control HF Token

### LINKED ISSUE #830: huggingface-cli login error when building quantized model
**Describe the bug**
I'm trying to quantize a fine-tuned Phi-3 model using `builder.py` and I'm running into the error below

Screenshot 2024-08-21 at 12 02 49 PM

Looking at the error message it's because the `configuration_phi3.py` file is missing from my fine-tuned model folder, so HuggingFace is trying to download it. However on line 1615 and line 2473 of the `builder.py` the `use_auth_token` is set to `True` which requires local HuggingFace login token.

```
config = AutoConfig.from_pretrained(hf_name, use_auth_token=True, trust_remote_code=True, **extra_kwargs)
model = AutoModelForCausalLM.from_pretrained(self.model_name_or_path, cache_dir=self.cache_dir, use_auth_token=True, trust_remote_code=True, **extra_kwargs)
```

After setting `use_auth_token=False` I was able to avoid this error.

**To Reproduce**
1. Download builder.py `curl https://raw.githubusercontent.com/microsoft/onnxruntime-genai/main/src/python/py/models/builder.py -o builder.py`
2. Run the quantization `python builder.py -i <fine-tuned-model-folder> -o quantized_model -p int4 -e cpu`

**Expected behavior**
No huggingface_hub token error is thrown if I'm using a public model like Phi-3. Can we change to `use_auth_token=False` by default?

**Environment**
 - OS: MacOS
 - onnx==1.16.2
 - onnxruntime==1.18.1

nmoeller:
fixes #830.

I've added a new option `hf_token` in the extra options, the user can pass a Token or True or False.
Also i upgraded `use_auth_token` to `token`.


src/python/py/models/builder.py
```diff
@@ -42,6 +42,7 @@ def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
 
         self.cache_dir = cache_dir
         self.filename = extra_options["filename"] if "filename" in extra_options else "model.onnx"
+        self.hf_token=parse_hf_token(extra_options["hf_token"])  if "hf_token" in extra_options else True
```
kunal-vaishnavi, line 45:
```suggestion
        self.hf_token = parse_hf_token(extra_options["hf_token"]) if "hf_token" in extra_options else True
```

src/python/py/models/builder.py
```diff
@@ -2720,7 +2729,8 @@ def create_model(model_name, input_path, output_dir, precision, execution_provid
     # Load model config
     extra_kwargs = {} if os.path.isdir(input_path) else {"cache_dir": cache_dir}
     hf_name = input_path if os.path.isdir(input_path) else model_name
-    config = AutoConfig.from_pretrained(hf_name, use_auth_token=True, trust_remote_code=True, **extra_kwargs)
+    hf_token=parse_hf_token(extra_options["hf_token"]) if "hf_token" in extra_options else True
```
kunal-vaishnavi, line 2732:
```suggestion
    hf_token = parse_hf_token(extra_options["hf_token"]) if "hf_token" in extra_options else True
```

src/python/py/models/builder.py
```diff
@@ -2859,6 +2869,7 @@ def get_args():
                     is the prerequisite for the CUDA graph to be used correctly. It is not guaranteed that cuda graph be enabled as it depends on the model
                     and the graph structure.
                 use_8bits_moe = 1 : Use 8-bit quantization for MoE layers. Default is using 4-bit quantization.
+                hf_token = 1 : Use this to enable or disable the HF Authentication, can be True,False or the token itself.
```
kunal-vaishnavi, line 2872:
```suggestion
                hf_token = true/false/token: Use this to enable or disable authentication with Hugging Face. Default is true.
                    If you have already authenticated, you do not need to use this.
```

kunal-vaishnavi, line 2872:
```suggestion
                hf_token = false/token: Use this to disable authentication with Hugging Face or provide a custom authentication token that differs from the one stored in your environment. Default behavior is to use the authentication token stored in the `HF_TOKEN` environment variable.
                    If you have already authenticated via `huggingface-cli login`, you do not need to use this flag because Hugging Face has already stored your authentication token in `HF_TOKEN` for you.
```

src/python/py/models/builder.py
kunal-vaishnavi, line 45:
```suggestion
        self.hf_token = parse_hf_token(extra_options.get("hf_token", "true"))
```

src/python/py/models/builder.py
kunal-vaishnavi, line 2732:
```suggestion
    hf_token = parse_hf_token(extra_options.get("hf_token", "true"))
```

kunal-vaishnavi:
Thank you for your contribution!

There are many ways that users can authenticate with Hugging Face.
1. [Always logged in via Hugging Face CLI](https://huggingface.co/docs/huggingface_hub/quick-start#login-command) (comes installed with `transformers`): `huggingface-cli login`
2. [Global environment variable](https://huggingface.co/docs/huggingface_hub/package_reference/environment_variables#hftoken): `export HF_TOKEN=<token>`
3. Per-run environment variable: `HF_TOKEN=<token> python3 -m onnxruntime.models.builder <...>`
4. Token in code: `from_pretrained(token=<token>)` 
5. Boolean in code: `from_pretrained(token=True)`
6. Default in Hugging Face's code: `from_pretrained(token=None)`

There are also several scenarios for when the token is needed or not needed.

| | Trust Remote Code = True | Trust Remote Code = False |
| ---- | ----------------------------- | ----------------------------- |
| **Use Auth Token = True** | Gated model, modeling file not implemented in installed `transformers` package | Gated model |
| **Use Auth Token = False** | Modeling file not implemented in installed `transformers` package | Local model |

Given this information, I think we should change the behavior of how the Hugging Face token is obtained to the following.

```python
# For the `Model` class
self.hf_token = parse_hf_token(extra_options.get("hf_token", "true"))
```

```python
# For `create_model`
```

```python
def parse_hf_token(hf_token):
   """
   Returns the authentication token needed for Hugging Face.
   Token is obtained either from the user or the environment.
   """
    if hf_token.lower() in {"false", "0"}:
        # Default is `None` for disabling authentication
        return None

    if hf_token.lower() in {"true", "1"}:
        # Return token in environment
        return os.environ["HF_TOKEN"]

    # Return user-provided token as string
    return hf_token
```

Then, the two options for `hf_token` would be `false` or the user-provided token. While not officially supported, the "0" and "1" flags are also checked because some users don't follow the documentation.

The `hf_token` flag would allow us to disable authentication with Hugging Face or provide a custom authentication token that differs from the one stored in the user's environment. When not using the flag, the default behavior would be to use the authentication token stored by `huggingface-cli login` in the `HF_TOKEN` environment variable.

What do you think?

nmoeller:
You are absolutely right and your feedback is very appreciated ! I implemented the changes you proposed. 
Sorry for my missing intelligence to not have the idea by my own 😄

kunal-vaishnavi:
> You are absolutely right and your feedback is very appreciated ! I implemented the changes you proposed.
Sorry for my missing intelligence to not have the idea by my own 😄

No worries and happy to review!

> @kunal-vaishnavi are we sure that the huggingface-cli login set the HF_TOKEN variable ? I think the token is stored locally in a file or ?
> I think we should return true from the method instead of the env variable ? Because when using from_pretrained(token=True) it will check for a local token set by huggingface-cli login and if no token is set it will use the HF_TOKEN automatically or ?

Upon further testing, it appears that the environment variables are not always set. I agree that returning `True` instead makes sense since the `from_pretrained(token=True)` command should pick up the token saved locally by `huggingface-cli login`.

According to [Hugging Face's documentation](https://huggingface.co/docs/huggingface_hub/package_reference/environment_variables#hftoken), the `HF_TOKEN` environment variable can override the token saved at `$HF_HOME/token`, which is where `huggingface-cli login` [saves the token](https://huggingface.co/docs/huggingface_hub/package_reference/environment_variables#hftokenpath). If a user wants to use a different token and doesn't want to pass it to `--hf_token`, then the user can either login with a different token via `huggingface-cli login` or use a per-run environment variable (e.g. `HF_TOKEN=<token> python3 -m onnxruntime.models.builder <...>`).

src/python/py/models/builder.py
```diff
@@ -2859,6 +2878,8 @@ def get_args():
                     is the prerequisite for the CUDA graph to be used correctly. It is not guaranteed that cuda graph be enabled as it depends on the model
                     and the graph structure.
                 use_8bits_moe = 1 : Use 8-bit quantization for MoE layers. Default is using 4-bit quantization.
+                hf_token = false/token: Use this to disable authentication with Hugging Face or provide a custom authentication token that differs from the one stored in your environment. Default behavior is to use the authentication token stored in the `HF_TOKEN` environment variable.
```
kunal-vaishnavi, line 2881:
```suggestion
                hf_token = false/token: Use this to disable authentication with Hugging Face or provide a custom authentication token that differs from the one stored in your environment. Default behavior is to use the authentication token stored by `huggingface-cli login`.
```

src/python/py/models/builder.py
```diff
@@ -2859,6 +2878,8 @@ def get_args():
...
+                    If you have already authenticated via `huggingface-cli login`, you do not need to use this flag because Hugging Face has already stored your authentication token in `HF_TOKEN` for you.
```
kunal-vaishnavi, line 2882:
```suggestion
                    If you have already authenticated via `huggingface-cli login`, you do not need to use this flag because Hugging Face has already stored your authentication token for you.
```

kunal-vaishnavi:
In the README, can you add a new section in "Extra Options" after "Use 8 Bits Quantization in QMoE" to show how to use this new extra option?

https://github.com/microsoft/onnxruntime-genai/blob/17648698c37c1178d3b5f58f78884682913e2d30/src/python/py/models/README.md?plain=1#L176-L186

Once added, can you also add a reference to the new section in the table of contents after "Use 8 Bits Quantization in QMoE"?

https://github.com/microsoft/onnxruntime-genai/blob/17648698c37c1178d3b5f58f78884682913e2d30/src/python/py/models/README.md?plain=1#L15-L20

src/python/py/models/README.md
```diff
@@ -185,6 +188,24 @@ python3 -m onnxruntime_genai.models.builder -i path_to_local_folder_on_disk -o p
 python3 builder.py -i path_to_local_folder_on_disk -o path_to_output_folder -p precision -e execution_provider -c cache_dir_to_store_temp_files --extra_options use_8bits_moe=1
 ```
 
+#### Hugging Face Authentication
+
+This scenario is for when you need control over the Hugging Face Authentication, to enable, disable or us a token.
+
+Possible values :
+
+- hf_token=True
```
kunal-vaishnavi, line 197: Since `hf_token = False` or `hf_token = token`, we should remove this line.

src/python/py/models/README.md
```diff
@@ -185,6 +188,24 @@ python3 -m onnxruntime_genai.models.builder -i path_to_local_folder_on_disk -o p
 python3 builder.py -i path_to_local_folder_on_disk -o path_to_output_folder -p precision -e execution_provider -c cache_dir_to_store_temp_files --extra_options use_8bits_moe=1
 ```
 
+#### Hugging Face Authentication
+
+This scenario is for when you need control over the Hugging Face Authentication, to enable, disable or us a token.
```
kunal-vaishnavi, line 193: Can we change this to the following and include the hyperlink?

This scenario is for when you need to disable the Hugging Face authentication or use a different authentication token than the one stored in [`huggingface-cli login`](https://huggingface.co/docs/huggingface_hub/main/en/guides/cli#huggingface-cli-login).

src/python/py/models/README.md
```diff
@@ -4,24 +4,27 @@ This folder contains the model builder for quickly creating optimized and quanti
...
-  - [Original PyTorch Model from Disk](#original-pytorch-model-from-disk)
-  - [Customized or Finetuned PyTorch Model](#customized-or-finetuned-pytorch-model)
-  - [Quantized PyTorch Model](#quantized-pytorch-model)
-  - [GGUF Model](#gguf-model)
-  - [Extra Options](#extra-options)
-    - [Config Only](#config-only)
-    - [Exclude Embedding Layer](#exclude-embedding-layer)
-    - [Exclude Language Modeling Head](#exclude-language-modeling-head)
-    - [Enable Cuda Graph](#enable-cuda-graph)
-    - [Use 8 Bits Quantization in QMoE](#use-8-bits-quantization-in-qmoe)
```
kunal-vaishnavi, line 20: Can we just insert the new "Hugging Face Authentication" section here and then keep the table of contents formatting as is? I think the first two links are self-explanatory since a user can see the README title and the table of contents heading as soon as the README doc is opened.

My concern is that users may click on those links, not see the page move since the user is already at those positions in the README, and then wonder why the link is not working (even though it is).

kunal-vaishnavi, line 20: Can we just keep the insertion of the new "Hugging Face Authentication" section below "Use 8 Bits Quantization in QMoE" and then keep the original table of contents formatting as is? I think the first two links, "ONNX Runtime GenAI Model Builder" and "Contents", are self-explanatory since a user can see the README title and the table of contents heading as soon as the README doc is opened.

My concern is that users may click on the new "ONNX Runtime GenAI Model Builder" and "Contents" links, not see the page move since the user is already at those positions in the README, and then wonder why the links are not working (even though it is).

nmoeller: Very sorry for this, my VSCode Markdown Extension automatically adjusted the ToC.... this was not intended. I think this is now how it should be or ?

kunal-vaishnavi: Looks good to me now. Thank you!

kunal-vaishnavi:
This PR looks good to me now! It appears you need to resolve two conflicting files as those files were recently updated by [another PR](https://github.com/microsoft/onnxruntime-genai/pull/900). Once those conflicts are resolved, I can approve and merge your changes!

kunal-vaishnavi (approved):

PR #900, opened 2024-09-18
Add QDQ MatMul option to model builder

PatriceVignola:


yufenglee:
The operators GroupQueryAttention, MultiHeadAttention, RotaryEmbedding, etc. are contrib ops. The model builder can not build a model without contrib ops now.

PatriceVignola:
> The operators GroupQueryAttention, MultiHeadAttention, RotaryEmbedding, etc. are contrib ops. The model builder can not build a model without contrib ops now.

@yufenglee Yes, this would be more of a work in progress kind of thing, starting with this QDQ change, and possibly adding other contrib ops to the list later on. We could fork this in a feature branch until all the work is ready (or replace the `--no_contrib_ops` switch with something like `--use_qdq`).

The reason for this PR is that the pytorch exporter route currently is slow, not great, not flexible enough and generates models with patterns that are sometimes too verbose. It's also harder to automatically generate some patterns (e.g. GQA or stacked GQA) with the pytorch exporter. Even though it should be the most obvious path for no-contrib-ops model, I feel like with just a few changes to the model builder that decompose the contrib ops, we can use the same ort-genai pipeline for both contrib ops and non-contrib-ops models.

I'd be happy to be proven wrong, but already with the QDQ path I hit limitations with the pytorch exporter where it wouldn't generate the correct QDQ pattern that we need (e.g. Dequantize + Transpose + MatMul is more natural than Dequantize + MatMul due to how blockwise quantization works).

I'd be happy to explore another path though, this PR is more of a proof of concept of the idea I had in mind.

yufenglee:
> > The operators GroupQueryAttention, MultiHeadAttention, RotaryEmbedding, etc. are contrib ops. The model builder can not build a model without contrib ops now.
> 
> @yufenglee Yes, this would be more of a work in progress kind of thing, starting with this QDQ change, and possibly adding other contrib ops to the list later on. We could fork this in a feature branch until all the work is ready (or replace the `--no_contrib_ops` switch with something like `--use_qdq`).
> 
> The reason for this PR is that the pytorch exporter route currently is slow, not great, not flexible enough and generates models with patterns that are sometimes too verbose. It's also harder to automatically generate some patterns (e.g. GQA or stacked GQA) with the pytorch exporter. Even though it should be the most obvious path for no-contrib-ops model, I feel like with just a few changes to the model builder that decompose the contrib ops, we can use the same ort-genai pipeline for both contrib ops and non-contrib-ops models.
> 
> I'd be happy to be proven wrong, but already with the QDQ path I hit limitations with the pytorch exporter where it wouldn't generate the correct QDQ pattern that we need (e.g. Dequantize + Transpose + MatMul is more natural than Dequantize + MatMul due to how blockwise quantization works).
> 
> I'd be happy to explore another path though, this PR is more of a proof of concept of the idea I had in mind.

@PatriceVignola, yes, we can do this incrementally. If you want to merge into main branch, i think use_qdq is a better option.

src/python/py/models/builder.py
```diff
@@ -429,7 +431,7 @@ def order_repeated_field(self, repeated_proto, key_name, order):
         order = list(order)
         repeated_proto.sort(key=lambda x: order.index(getattr(x, key_name)))
 
-    def make_external_tensor(self, np_data, name, **kwargs):
+    def make_external_tensor(self, np_data, name, unpack_int4=False):
```
kunal-vaishnavi, line 434:
```suggestion
    def make_external_tensor(self, np_data, name, unpack_int4=False, **kwargs):
```

src/python/py/models/builder.py
```diff
@@ -2813,6 +2874,12 @@ def get_args():
         help="Precision of model",
     )
 
+    parser.add_argument(
+        "--no_contrib_ops",
```
kunal-vaishnavi, line 2878: I agree that this should be changed to `use_qdq` instead. We should also add this via the `extra_options` instead.

src/python/py/models/builder.py
```diff
@@ -2859,6 +2920,7 @@ def get_args():
                     is the prerequisite for the CUDA graph to be used correctly. It is not guaranteed that cuda graph be enabled as it depends on the model
                     and the graph structure.
                 use_8bits_moe = 1 : Use 8-bit quantization for MoE layers. Default is using 4-bit quantization.
+                use_qdq = 1 : Use the QDQ decomposition for quantized MatMul instead of the MatMulNBits operator.
```
kunal-vaishnavi, line 2923: Can you add a section that shows example usage in the README after the MoE section [here](https://github.com/microsoft/onnxruntime-genai/blob/main/src/python/py/models/README.md#use-8-bits-quantization-in-qmoe)?

src/python/py/models/builder.py
```diff
@@ -712,6 +721,58 @@ def make_matmul_int4(self, matmul, basename, root_input, **kwargs):
...
+
+        if hasattr(matmul, "qzeros") and matmul.qzeros is not None:
+            zeros = dequantize_name[1:].replace("/", ".") + ".qzeros"
+            zeros_npy = matmul.qzeros.detach().numpy()
+            zeros_npy = zeros_npy.reshape(*qweight_npy.shape[:-1], qweight_npy.shape[-1] // matmul.group_size)
+            self.make_external_tensor(zeros_npy, zeros, True)
+            dequantize_inputs.append(zeros)
+
+        dequantize_output = f"{dequantize_name}/output_0"
+        self.make_node("DequantizeLinear", inputs=dequantize_inputs, outputs=[dequantize_output], name=dequantize_name, block_size=matmul.group_size)
```
kunal-vaishnavi, line 755: Can we put this in a `make_dequantize_linear` method?

src/python/py/models/builder.py
```diff
@@ -712,6 +721,58 @@ def make_matmul_int4(self, matmul, basename, root_input, **kwargs):
...
+        # compute quantized matmul when the weights are transposed. In most implementations, the transpose should usually be converted to a "transposeB"
+        # attribute on the MatMul itself. A more natural way to represent this would have been to use Gemm since it already supports a transB attribute,
+        # but unfortunately Gemm doesn't support batches.
+        perms = list(range(0, len(scales_npy.shape)))
+        perms[-2], perms[-1] = perms[-1], perms[-2]
+        transposed_shape = [*scales_npy.shape[:-2], scales_npy.shape[-1] * matmul.group_size, scales_npy.shape[-2]]
+
+        transpose_name = f"{matmul_name}/Transpose"
+        transpose_output = f"{transpose_name}/output_0"
+        self.make_node("Transpose", inputs=[dequantize_output], outputs=[transpose_output], name=transpose_name, perm=perms)
```
kunal-vaishnavi, line 768: Can we use `make_transpose` instead?

https://github.com/microsoft/onnxruntime-genai/blob/0f193ac0561726d8432cffcda1be55652a390814/src/python/py/models/builder.py#L645-L648

src/python/py/models/README.md
```diff
@@ -185,6 +185,18 @@ python3 -m onnxruntime_genai.models.builder -i path_to_local_folder_on_disk -o p
 python3 builder.py -i path_to_local_folder_on_disk -o path_to_output_folder -p precision -e execution_provider -c cache_dir_to_store_temp_files --extra_options use_8bits_moe=1
 ```
 
+#### Use QDQ Pattern for Quantization
```
kunal-vaishnavi, line 188: Can you add a reference to this section in the table of contents after the QMoE reference?

https://github.com/microsoft/onnxruntime-genai/blob/0f193ac0561726d8432cffcda1be55652a390814/src/python/py/models/README.md?plain=1#L15-L20

src/python/py/models/builder.py
```diff
@@ -713,6 +722,60 @@ def make_matmul_int4(self, matmul, basename, root_input, **kwargs):
...
+
+        if hasattr(quantized_op, "qzeros") and quantized_op.qzeros is not None:
+            zeros = dequantize_name[1:].replace("/", ".") + ".qzeros"
+            zeros_npy = quantized_op.qzeros.detach().numpy()
+            zeros_npy = zeros_npy.reshape(*qweight_npy.shape[:-1], qweight_npy.shape[-1] // quantized_op.group_size)
+            self.make_external_tensor(zeros_npy, zeros, True)
+            dequantize_inputs.append(zeros)
+
+        dequantize_output = f"{dequantize_name}/output_0"
+        self.make_node("DequantizeLinear", inputs=dequantize_inputs, outputs=[dequantize_output], name=dequantize_name, block_size=quantized_op.group_size)
```
yufenglee, line 747: The default axis is 1. it is 0 in this case.

src/python/py/models/builder.py
```diff
@@ -713,6 +722,60 @@ def make_matmul_int4(self, matmul, basename, root_input, **kwargs):
...
+
+        # Add a transpose instead of transposing the weights offline. The reason for this is that it is more natural and usually more performant to
+        # compute quantized matmul when the weights are transposed. In most implementations, the transpose should usually be converted to a "transposeB"
+        # attribute on the MatMul itself. A more natural way to represent this would have been to use Gemm since it already supports a transB attribute,
+        # but unfortunately Gemm doesn't support batches.
+        qweight_shape = matmul.scales.detach().numpy().shape
+        qweight_shape = [*qweight_shape[:-2], qweight_shape[-2] * qweight_shape[-1] * 2]
+        perms = list(range(0, len(qweight_shape)))
+        perms[-2], perms[-1] = perms[-1], perms[-2]
+        transposed_shape = [*qweight_shape[:-2], qweight_shape[-1] * matmul.group_size, qweight_shape[-2]]
```
yufenglee, line 769: >transposed_shape [](http://example.com/codeflow?start=8&length=16)

the output of transpose should be a 2D tensor

src/python/py/models/builder.py
PatriceVignola: The correct axis is actually `1` here, I can make it explicit by hardcoding the axis to the last dimension, but the second dimension is the one that's blocked, not the first one.

src/python/py/models/builder.py
PatriceVignola: It is in this case, but this code also handles cases where the shape has leading dimensions or is reshaped differently for some reason. `[*qweight_shape[:-2], qweight_shape[-1] * matmul.group_size, qweight_shape[-2]]` will still result in a 2D tensor if `qweight_shape[:-2]` is empty.

yufenglee (approved):
:shipit:

kunal-vaishnavi (approved):

PR #903, opened 2024-09-19
Update python benchmark script

thevishalagarwal:
* fixed warmup generation loop to complete the generation of tokens instead of generating just one token
* removed redundant generator object initialization in the benchmark loop. This improves (reduces) memory utilization
* fixed things related to printing memory usage

Issue: If multiple prompt lengths are used in benchmark arg, the script would print the same memory usage for all cases. 
Solution: Need to reset monitoring and peak_gpu_memory variables in each benchmark loop and print at the end of the loop along with other metrics


thevishalagarwal:
@baijumeswani to review

baijumeswani (approved):

baijumeswani:
Thanks for your contribution. :)

PR #907, opened 2024-09-20
Optimize Decoder Pipeline Model Execution

baijumeswani:
#729 introduced support to execute a pipeline of ort sessions sequentially as defined by the config file.

This pull-request builds on top by:

- Fixing a memory leak bug where the ortvalues stored in the `ortvalue_pool_` were raw pointers and were not released.
- Clears outputs from pipeline state that was only run on prompt. Before this pull-request, those ortvalues lived in the `ortvalue_store_` forever using unnecessary memory.
- Enable registering the managed inputs as pipeline model outputs.
- Sharing an allocator across ort sessions.

Co-contributor: @edgchen1 who helped identify the bug(s).


src/models/decoder_only_pipeline.cpp
```diff
@@ -179,16 +204,31 @@ RoamingArray<float> DecoderOnlyPipelineState::Run(int current_length, RoamingArr
     // Run the intermediate pipeline state
     pipeline_state->Run(current_length, next_tokens, next_indices);
 
-    // Store the non-managed outputs from the current pipeline state in the ortvalue pool.
+    // Transfer ownership of all the non-managed outputs from the current pipeline state to the ortvalue store.
     // All non managed outputs are assumed to be on CPU
     for (size_t i = 0; i < pipeline_state->output_names_.size(); ++i) {
       if (std::none_of(output_names_.begin(), output_names_.end(),
+                       [&](const std::string& elem) { return elem == pipeline_state->output_names_[i]; }) &&
+          std::none_of(input_names_.begin(), input_names_.end(),
```
edgchen1, line 212: just for my understanding, why was it changed from `output_names` to `input_names` here? also, what is the difference between a managed and non-managed output?

baijumeswani: Yes. managed inputs/outputs are the model inputs/outputs that ort-genai is aware of. For example, the kv cache, the input_ids, the position_ids and so on. The input/outputs are updated by ort-genai after each run (as required).

Non managed inputs and outputs are those pipeline-model inputs and outputs that ort-genai does not care/know of. For example intermediate outputs of one pipeline model that feed directly into the next pipeline model is a non managed input/output. These ortvalues are stored in the `ortvalue_store_` map.

src/generators.cpp
```diff
@@ -13,7 +13,12 @@ namespace Generators {
 
 static bool _ = (Ort::InitApi(), false);
 
-OrtGlobals::OrtGlobals() : env_{OrtEnv::Create(OrtLoggingLevel::ORT_LOGGING_LEVEL_ERROR)} {}
+OrtGlobals::OrtGlobals()
+    : env_{OrtEnv::Create(OrtLoggingLevel::ORT_LOGGING_LEVEL_ERROR)},
+      arena_config_{OrtArenaCfg::Create(0, -1, -1, -1)} {
```
yufenglee, line 18: >arena_config_ [](http://example.com/codeflow?start=6&length=13)

Do you need to hold the arena_config_ in OrtGlobals? Isn't a temporary variable enough?

src/models/decoder_only_pipeline.cpp
```diff
@@ -179,16 +204,31 @@ RoamingArray<float> DecoderOnlyPipelineState::Run(int current_length, RoamingArr
...
         } else {
-          ortvalue_pool_[pipeline_state->output_names_[i]] = pipeline_state->outputs_[i];
+          ortvalue_store_[pipeline_state->output_names_[i]] = std::unique_ptr<OrtValue>(pipeline_state->outputs_[i]);
+        }
+      }
+    }
+  }
+
+  // Clear the outputs of the pipeline models that are only run on prompt since this cannot happen earlier.
+  if (!first_run_) {
```
yufenglee, line 225: is it better to clean on the first_run_? only the managed inputs/outputs used cross DecoderOnlyPipelineState::run()

baijumeswani: We cannot clean on the first run.

The pipeline model can be set up in a way such that the outputs of the prompt processor models are used as inputs in the token generation models.

src/generators.cpp
```diff
@@ -13,7 +13,12 @@ namespace Generators {
...
+  Ort::Allocator& allocator_cpu_{Ort::Allocator::GetWithDefaultOptions()};
+  env_->CreateAndRegisterAllocator(allocator_cpu_.GetInfo(), *arena_config_);
```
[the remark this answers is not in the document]
skottmckay: One risk would be how the arena grows. 

The default is to grow from 1MB, in 2x increments (until the current allocation can be satisfied) up to 1GB, at which point it grows in 1GB chunks. On some platforms 1GB growth could be excessive, but all these things are configurable. 

I believe you can load initializers without using the arena so the arena growth is from inferencing not model loading, which can help avoid large jumps. 

https://github.com/microsoft/onnxruntime/blob/ae66d0e7cf6774dc1b6435e122d3589251e6fbc8/include/onnxruntime/core/session/onnxruntime_session_options_config_keys.h#L110-L112

src/generators.cpp
baijumeswani: Thanks. Changed the code to use a local variable.

src/generators.cpp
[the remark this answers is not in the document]
baijumeswani: I added the option to configure the usage of the env allocator. This option is set to true by default but will only be applied for sessions that use just the CPU provider.

yufenglee (approved):

src/models/model.cpp
```diff
@@ -364,6 +364,11 @@ void Model::CreateSessionOptionsFromConfig(const Config::SessionOptions& config_
     session_options.SetEpContextFilePath(config_session_options.ep_context_file_path.value().c_str());
   }
 
+  if (config_session_options.provider_options.empty() && config_session_options.use_env_allocators) {
```
edgchen1, line 367: is it also possible to specify CPU EP provider options? would `provider_options` not be empty in that case?

edgchen1 (approved):

src/models/model.cpp
baijumeswani: The only way to set CPU provider is by leaving the provider options empty. We do not have a way to explicitly set the CPU provider right now.

baijumeswani:
Thank you for the review. :)

PR #910, opened 2024-09-23
Rm download-ort pipeline step

skyline75489:
Thanks to #890 , these steps are no longer needed.


.pipelines/stages/jobs/android-java-api-aar.yml
```diff
@@ -27,17 +27,6 @@ jobs:
   - template: steps/utils/use-android-ndk.yml
   - template: steps/utils/set-genai-version.yml
 
-  - task: CmdLine@2
```
skyline75489, line 30: Note that the Android pipeline is currently still broken. I'll fix it in the following several PRs.

snnn (approved):

PR #911, opened 2024-09-23
C, C++, and Python API for Adapters (Multi-LoRA and others)

baijumeswani:
This pull-request introduces a way to apply adapters to a model with onnxruntime-genai. It leverages onnxruntime API changes from https://github.com/microsoft/onnxruntime/pull/22046 and allows users to load, apply, and unload adapters as needed.

This pull-request adds the C, C++ and the Python API. Other language bindings will follow soon.

Example usage:

__C++__
```cpp
auto model = OgaModel::Create("<adapter/model/path>");
auto adapters = OgaAdapters::Create(*model);
adapters->LoadAdapter("</absolute/adapter/file/path>", "<unique_user_supplied_adapter_identifier>");

const char* input_strings[] = {
    "This is a test.",
    "Rats are awesome pets!",
    "The quick brown fox jumps over the lazy dog.",
};

auto input_sequences = OgaSequences::Create();
for (auto& string : input_strings)
  tokenizer->Encode(string, *input_sequences);

{
  params->SetSearchOption("max_length", 20);
  params->SetInputSequences(*input_sequences);

  generator->SetActiveAdapter(*adapters, "<unique_user_supplied_adapter_identifier>");

  while (!generator->IsDone()) {
    generator->ComputeLogits();
  }
}

// Optionally, unload the adapter. Will error out if the adapter is still active.
// So, the generator must go out of scope before the adapter can be unloaded.
adapters->UnloadAdapter("<unique_user_supplied_adapter_identifier>");
```

__Python__
```python
model = og.Model("<model/path>")
adapters = og.Adapters(model)
adapters.load("</absolute/path/to/adapters>", "<unique_user_supplied_adapter_identifier>")

prompts = [
    "This is a test.",
    "Rats are awesome pets!",
    "The quick brown fox jumps over the lazy dog.",
]

params.set_search_options(max_length=20)
params.input_ids = tokenizer.encode_batch(prompts)

generator.set_active_adapter(adapters, "<unique_user_suppled_adapter_identifier>")

    generator.compute_logits()
```


src/models/adapters.cpp
```diff
@@ -0,0 +1,73 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#include "../generators.h"
+#include "model.h"
+
+namespace Generators {
+
+Adapter::Adapter(const char* adapter_file_path, Ort::Allocator* allocator)
+    : adapter_{OrtLoraAdapter::Create(fs::path(adapter_file_path).c_str(), *allocator)} {}
```
RyanUnderhill, line 10: What does fs::path(adapter_file_path).c_str() do vs just passing adapter_file_path?

src/ort_genai_c.cpp
```diff
@@ -531,6 +531,39 @@ size_t OGA_API_CALL OgaStringArrayGetCount(const OgaStringArray* string_array) {
   return reinterpret_cast<const std::vector<std::string>*>(string_array)->size();
 }
 
+OgaResult* OgaCreateAdapters(const OgaModel* model, OgaAdapters** adapters_p) {
```
RyanUnderhill, line 534: Should `adapters_p` be `out` for consistency with other methods?

src/models/adapters.cpp
```diff
@@ -0,0 +1,73 @@
...
+
+const OrtLoraAdapter* Adapter::AcquireRef() {
```
RyanUnderhill, line 12: Are the AcuireRef() and ReleaseRef() management entirely inside our code? A user can't mess these up?

src/models/adapters.cpp
baijumeswani: It manages the os specific differences with regards to the path file (wide string vs regular string).

On Windows, the ort api expects const w_char* and on linux it expects const char*. The fs::path utility inside ort-genai manages that.

src/models/adapters.cpp
baijumeswani: Users don't have access to individual adapters. The adapters that are active have ref count > 0 where as the adapters that are inactive have ref count = 0.

When the user wants to unload an adapter, they must have no generator object that has that same adapter active. Otherwise, the unload will fail.

However, if we inside our code try to AcquireRef and forget to ReleaseRef (i.e. there is mismatch between AcquireRef and ReleaseRef within our code), it can lead to failures when unload is invoked since the adapter considers itself still active.

The generator holds a shared pointer to the adapters, so even if the user attempts to destroy the adapters container (that holds all adapters), the container won't be destroyed unless all generators that reference it are destroyed.

baijumeswani: But maybe we can automate this further by the use of shared pointers and use the shared_ptr::use_count variable to determine whether unload can be executed? I can explore this further.

test/c_api_tests.cpp
```diff
@@ -365,6 +368,46 @@ TEST(CAPITests, TopKTopPCAPI) {
 
 #endif  // TEST_PHI2
 
+TEST(CAPITests, AdaptersTest) {
```
natke, line 371: Can we have a test with more than one adapter?

pranavsharma (approved):
In the interest of time, let's resolve the ref counting issues later as they don't affect the public APIs.

src/models/adapters.cpp
baijumeswani: I will defer this to a future commit, once we have more clarity on whether we should be ref counting to begin with.

test/c_api_tests.cpp
baijumeswani: Will add the test.

baijumeswani: Will defer the test to a future PR to unblock the C# API. Will add them in soon.

src/ort_genai_c.cpp
baijumeswani: Will address this also in a separate PR so I can unblock the C# API development. Will get to it soon.

PR #913, opened 2024-09-23
Share downloaded ort with ort-extensions

baijumeswani:
So far, ort-extensions and ort both download ort to build.

With this pull-request, ort-genai can share the downloaded ort header dir with ort-extensions.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -20,8 +20,8 @@ if(MSVC)
   add_compile_definitions(_DISABLE_CONSTEXPR_MUTEX_CONSTRUCTOR)
 endif()
 
-include(cmake/external/onnxruntime_external_deps.cmake)
 include(cmake/ortlib.cmake)
+include(cmake/external/onnxruntime_external_deps.cmake)
 # All Global variables, including GLOB, for the top level CMakeLists.txt should be defined here
 include(cmake/global_variables.cmake)
 # Checking if CUDA is supported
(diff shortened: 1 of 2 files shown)


skyline75489:
From what I can tell, the exact version of ORT downloaded by ort-extension is NOT the same as the one download by GenAI, is this going to be a problem?

baijumeswani:
> From what I can tell, the exact version of ORT downloaded by ort-extension is NOT the same as the one download by GenAI, is this going to be a problem?

It should not matter. ort lib api is backwards compatible and ort-extensions only uses header file of ort for building. So, if any problems, it can be caught at compile time.

skyline75489:
Is it possible to prevent onnxruntime-extension from downloading its own ORT at all?  Seems that `ONNXRUNTIME_PKG_DIR` would help. We probably needs more tweak for Android and iOS, but it's a start.

See: https://github.com/microsoft/onnxruntime-extensions/blob/6b94f4d7a597f0199f3e44d020d601d408605736/cmake/ext_ortlib.cmake#L4

skyline75489:
Oh I was mistaken. Setting `ONNXRUNTIME_INCLUDE_DIR` would also prevent ortx from downloading ORT. Nice!

See: https://github.com/microsoft/onnxruntime-extensions/blob/6b94f4d7a597f0199f3e44d020d601d408605736/CMakeLists.txt#L297

skyline75489 (approved):

baijumeswani:
Thanks for the review :)

PR #914, opened 2024-09-24
Remove unnecessary HF token query

baijumeswani:
The pipelines do not really need any HF token to run. The pipelines started recently failing due to key vault access issues.

This pull-request removes the need to query the HF token by setting a dummy token.


skyline75489 (approved):
this is long overdue.

PR #920, opened 2024-09-24
Remove redundant config copy from GeneratorParams 

RyanUnderhill:
Some model config options were being copied into the GeneratorParams and then relayed to the Generator functions, this is unnecessary and confusing. Instead the GeneratorParams just holds a pointer to the config and these options are now taken directly from the config.

This simplifies the GeneratorParams


yufenglee (approved):

src/generators.h
```diff
@@ -57,44 +57,21 @@ enum struct DeviceType {
 std::string to_string(DeviceType device_type);
 
 struct GeneratorParams : std::enable_shared_from_this<GeneratorParams>, LeakChecked<GeneratorParams> {
-  GeneratorParams() = default;  // This constructor is only used if doing a custom model handler vs built-in
+  GeneratorParams(const Config& config);  // This constructor is only used for internal generator benchmarks
```
baijumeswani, line 60: Can we remove this constructor completely? It requires us to maintain two constructors and can lead to unnecessary overhead if we update one and not the other.
Since this is only use for the benchmark code, maybe we can update the benchmark code to overwrite the config file of the model instead of exposing this constructor?

We can do this later though.

baijumeswani (approved):

src/generators.h
RyanUnderhill: I agree, I wanted to remove it but having all of the benchmarks require separate configs felt like a lot of work. I think after the continuous decoding change and the GeneratorParams no longer have input IDs I can simplify it further.

PR #921, opened 2024-09-24
Added Support for Vanilla and Quantized ChatGLM3 Models to Model Builder

amd-sudo-sh:
We have integrated support for both vanilla and quantized versions of the [ChatGLM3](https://huggingface.co/THUDM/chatglm3-6b) models into the Model Builder. Additionally, we have successfully performed parity checks for both model types to ensure consistency and reliability across different configurations.

PR #922, opened 2024-09-24
Extend model type with ChatGLM3

BowenBao:
In addition to #921. Afterwards, what remains is tokenizer support for ChatGLM3.

edit: also added a small fix to populate `bos_token_id` as it is not present in ChatGLM3 source config.

### DIFF
--- a/src/python/py/models/builder.py
+++ b/src/python/py/models/builder.py
@@ -298,9 +298,15 @@ class Model:
             "past_key_names": "past_key_values.%d.key",
             "past_value_names": "past_key_values.%d.value",
         })
+        if self.model_type.find("ChatGLM") != -1:
+            # config.bos_token_id not present in ChatGLM model configs.
+            bos_token_id = 1
+        else:
+            bos_token_id = config.bos_token_id
+
         genai_config = {
             "model": {
-                "bos_token_id": config.bos_token_id,
+                "bos_token_id": bos_token_id,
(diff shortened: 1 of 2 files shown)


baijumeswani:
@BowenBao could you please pose the link to the HF model card so @wenbingl can take a look at what needs to be done to support the model from the tokenizer end?

baijumeswani:
I think it is this one from another PR: https://huggingface.co/THUDM/chatglm-6b

kunal-vaishnavi (approved):

src/python/py/models/builder.py
```diff
@@ -298,9 +298,15 @@ def make_genai_config(self, model_name_or_path, extra_kwargs, out_dir):
...
+        if self.model_type.find("ChatGLM") != -1:
+            # config.bos_token_id not present in ChatGLM model configs.
+            bos_token_id = 1
+        else:
+            bos_token_id = config.bos_token_id
+
         genai_config = {
             "model": {
-                "bos_token_id": config.bos_token_id,
+                "bos_token_id": bos_token_id,
```
kunal-vaishnavi, line 309: Can we simplify this change into one line and say `"bos_token_id": config.bos_token_id if hasattr(config, "bos_token_id") else 1`?

BowenBao: sure

PR #926, opened 2024-09-25
Add llama C++ example

natke:


examples/c/src/llama.cpp
```diff
@@ -0,0 +1,86 @@
...
+    auto generator = OgaGenerator::Create(*model, *params);
+
+    while (!generator->IsDone()) {
+      generator->ComputeLogits();
+      generator->GenerateNextToken();
+
+      // Show usage of GetOutput
+      std::unique_ptr<OgaTensor> output_logits = generator->GetOutput("logits");
+
+      // Assuming output_logits.Type() is float as it's logits
```
RyanUnderhill, line 42: Just fyi, the raw model output can be float16 if the model runs on cuda. Our internal "processed logits" are always float32

examples/c/src/llama.cpp
```diff
@@ -0,0 +1,86 @@
...
+    params->SetInputSequences(*sequences);
+
+    auto generator = OgaGenerator::Create(*model, *params);
+
+    while (!generator->IsDone()) {
+      generator->ComputeLogits();
+      generator->GenerateNextToken();
+
+      // Show usage of GetOutput
+      std::unique_ptr<OgaTensor> output_logits = generator->GetOutput("logits");
```
RyanUnderhill, line 40: Is the `std::unique_ptr<OgaTensor>` for clarity vs auto?

examples/c/src/llama.cpp
```diff
@@ -0,0 +1,86 @@
...
+  }
+
+  // Responsible for cleaning up the library during shutdown
+  OgaHandle handle;
+
+  std::cout << "-------------" << std::endl;
+  std::cout << "Run Llama" << std::endl;
+  std::cout << "-------------" << std::endl;
+
+#ifdef USE_CXX
```
RyanUnderhill, line 80: What is the USE_CXX here for, given the whole file is C++?

baijumeswani:
What changes from one decoder-only model to another? Shall we instead create a standardized example that can work with other decoder-only models?

kunal-vaishnavi:
> What changes from one decoder-only model to another? Shall we instead create a standardized example that can work with other decoder-only models?

I agree that one standardized example should be sufficient. From a user's perspective, the main change between decoder-only models is the chat template.

examples/c/CMakeLists.txt
```diff
@@ -6,6 +6,7 @@ set(CMAKE_CXX_STANDARD 20)
 option(USE_CUDA "Build with CUDA support" OFF)
 option(USE_CXX "Invoke the C++ example" ON)
 option(PHI3 "Build the Phi example" OFF)
+option(LLAMA "Build the Llama example" OFF)
```
kunal-vaishnavi, line 9: Could we have something like the following?

```suggestion
option(LLM "Build the large-language model example" OFF)
option(VLM "Build the vision-language model example" OFF)
option(ALM "Build the audio-language model example" OFF)
```

The abbreviations and names can be different, but the idea would be to create examples grouped by input and output modality.
- The `LLM` could be for decoder-only models (e.g. LLaMA, Phi)
  - Inputs: text
  - Outputs: text
- The `VLM` could be for vision-text models (e.g. Phi-3/Phi-3.5 vision)
  - Inputs: images, text
  - Outputs: text
- The `ALM` could be for audio-text models (e.g. Whisper)
  - Inputs: audios, text
  - Outputs: text

natke:
> What changes from one decoder-only model to another? Shall we instead create a standardized example that can work with other decoder-only models?

I agree that we should have a base decoder class, maybe? And these scripts can inherit from that class. But I'd like to get this sample in soonish so that folks can run Llama

examples/c/CMakeLists.txt
natke: I think this is a great proposal for a new PR!

examples/c/src/llama.cpp
natke: So is this correct, or not?

examples/c/src/llama.cpp
natke: So `auto` would be simpler? I'll update it

PR #930, opened 2024-09-26
Disable graph capture for embedding model

PatriceVignola:
The new embedding model for the vision pipeline contains `if` nodes, so graph capture is not possible.

### DIFF
--- a/src/models/model.h
+++ b/src/models/model.h
@@ -132,6 +132,7 @@ struct Model : std::enable_shared_from_this<Model>, LeakChecked<Model> {
   std::unique_ptr<Config> config_;
   std::unique_ptr<OrtSessionOptions> session_options_;
   std::unique_ptr<OrtSessionOptions> vision_session_options_;
+  std::unique_ptr<OrtSessionOptions> embedding_session_options_;
   std::unique_ptr<OrtRunOptions> run_options_;
 
   cuda_stream_holder cuda_stream_;
--- a/src/models/model.h
+++ b/src/models/model.h
@@ -131,8 +131,6 @@ struct Model : std::enable_shared_from_this<Model>, LeakChecked<Model> {
 
   std::unique_ptr<Config> config_;
   std::unique_ptr<OrtSessionOptions> session_options_;
-  std::unique_ptr<OrtSessionOptions> vision_session_options_;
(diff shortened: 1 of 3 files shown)


kunal-vaishnavi:
Can graph capture still be enabled for the vision component? It has a `Loop` node with a sub-graph now.

PatriceVignola:
> Can graph capture still be enabled for the vision component? It has a `Loop` node with a sub-graph now.

Hmm I'm pretty sure loops are also a problem here, the problem is with control flow nodes in general, not only if nodes. Also, graph capture is not really useful for the vision model because it's executed only once per prompt, so it's not like the text model that we need to keep executing dozens of times with the same static dimensions.

kunal-vaishnavi:
> Hmm I'm pretty sure loops are also a problem here, the problem is with control flow nodes in general, not only if nodes. Also, graph capture is not really useful for the vision model because it's executed only once per prompt, so it's not like the text model that we need to keep executing dozens of times with the same static dimensions.

Can we disable graph capture for the vision model in this PR as well then?

PatriceVignola:
> > Hmm I'm pretty sure loops are also a problem here, the problem is with control flow nodes in general, not only if nodes. Also, graph capture is not really useful for the vision model because it's executed only once per prompt, so it's not like the text model that we need to keep executing dozens of times with the same static dimensions.
> 
> Can we disable graph capture for the vision model in this PR as well then?

It was never enabled because it never used to work with the vision model (the previous first model also had issues not suitable for graph capture, like nodes being executed on the CPU).

src/models/model.h
```diff
@@ -132,6 +132,7 @@ struct Model : std::enable_shared_from_this<Model>, LeakChecked<Model> {
   std::unique_ptr<Config> config_;
   std::unique_ptr<OrtSessionOptions> session_options_;
   std::unique_ptr<OrtSessionOptions> vision_session_options_;
+  std::unique_ptr<OrtSessionOptions> embedding_session_options_;
```
baijumeswani, line 135: Can these session options be owned by the Model in question (like MultiModalPipelineModel) instead of being on the base class?

baijumeswani (approved):

kunal-vaishnavi (approved):

PR #931, opened 2024-09-26
Support tensor dumping for DML

PatriceVignola:


baijumeswani:
I think we have an intermittent crash (probably a result of https://github.com/microsoft/onnxruntime-genai/pull/907/commits/213f74f3b84eaa5085af504a879f22ba73e91fbe).

Will debug this issue.

src/models/debugging.cpp
```diff
@@ -71,7 +76,7 @@ void DumpValues(std::ostream& stream, ONNXTensorElementDataType type, const void
   stream << SGR::Fg_Green << "]" << SGR::Reset << std::endl;
 }
 
-void DumpTensor(std::ostream& stream, OrtValue* value, bool dump_value) {
+void DumpTensor(const Model& model, std::ostream& stream, OrtValue* value, bool dump_value) {
```
baijumeswani, line 79: Do we need to pass in the Model to dump the tensors? Is there any way around it?

baijumeswani (approved):

src/models/debugging.cpp
PatriceVignola: Unlike CUDA, D3D12 doesn't have a way to do synced copies without accessing an instantiated object first (called a command queue). The best way to access this object is by retrieving it from the model. Otherwise, we'd have to pass the D3D12 objects directly to the function, which would be uglier because it would require different function signatures between CUDA and DML.

Issue #933, opened 2024-09-26
Genny CUDA support does not work


lilhoser:
**Describe the bug**

Note: I'm using this model `https://huggingface.co/microsoft/Phi-3-mini-128k-instruct-onnx/tree/main/cuda/cuda-int4-rtn-block-32`

The readme for the Genny sample C# app states:

```
Debug_Cuda or Release_Cuda to launch the application with CUDA GPU support
```

However, when building with `Debug_Cuda`, I get the runtime error `CUDA execution provider is not enabled in this build.`

Genny app does specify a different nuget package in the `csproj`:

```
      <PackageReference Include="Microsoft.ML.OnnxRuntimeGenAI" Version="0.4.0" />
      <PackageReference Include="Microsoft.ML.OnnxRuntimeGenAI.Cuda" Version="0.4.0" Condition=" '$(Configuration)' == 'Debug_Cuda' OR '$(Configuration)' == 'Release_Cuda' " />
      <PackageReference Include="Microsoft.ML.OnnxRuntimeGenAI.DirectML" Version="0.4.0" />
```

But in the console output, it's loading the wrong package:

```
'Genny.exe' (CoreCLR: clrhost): Loaded 'C:\tools\onnxruntime-genai\examples\csharp\Genny\Genny\bin\x64\Debug_Cuda\net8.0-windows7.0\Microsoft.ML.OnnxRuntimeGenAI.dll'. 
```

How is this supposed to work when all three `Microsoft.ML.OnnxRuntimeGenAI.dll` are referenced in the project?

**To Reproduce**
1. Clone genny repo
2. Build `Debug_Cuda` configuration
3. Load model `dir https://huggingface.co/microsoft/Phi-3-mini-128k-instruct-onnx/tree/main/cuda/cuda-int4-rtn-block-32`

**Expected behavior**

Expect the app to run with CUDA support and the right nuget package.

**Screenshots**
<img width="584" alt="image" src="https://github.com/user-attachments/assets/b8f56c15-b866-4704-a4d3-c95725b2060c">

**Desktop (please complete the following information):**
 - OS: Windows 11

lilhoser:
If I uninstall all but the CUDA nuget package and manually install CUDNN. it appears to work.

Is it not possible to have all packages included in the solution to build these different configurations?

baijumeswani:
I think this should resolve the issue you were seeing: https://github.com/microsoft/onnxruntime-genai/pull/937.

[...]

PR #937, opened 2024-09-27
[Genny] Add missing build flags on DirectML package

saddam213:
The build flags are missing from the `DirectML:` Nuget packages, this is causing it to supersede the `CUDA` package if selected

Adding back the `Debug_DirectML` and `Release_DirectML` conditional flags

Mentioned in Issue https://github.com/microsoft/onnxruntime-genai/issues/933

### DIFF
--- a/examples/csharp/Genny/Genny/Genny.csproj
+++ b/examples/csharp/Genny/Genny/Genny.csproj
@@ -13,7 +13,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.ML.OnnxRuntimeGenAI" Version="0.4.0" />
+    <PackageReference Include="Microsoft.ML.OnnxRuntimeGenAI" Version="0.4.0" Condition=" '$(Configuration)' == 'Debug' OR '$(Configuration)' == 'Release' "/>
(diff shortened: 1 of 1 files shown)


baijumeswani (approved):

baijumeswani:
We probably need the same for the cpu variant as well. Would you be able to add that in this PR?

baijumeswani:
Microsoft.ML.OnnxRuntimeGenAI is the CPU package
Microsoft.ML.OnnxRuntimeGenAI.Cuda is the Cuda package
Microsoft.ML.OnnxRuntimeGenAI.DirectML is the DirectML package

The problem is that when Debug_DirectML is enabled, the CPU package will also be installed since it has no condition associated with it.

saddam213:
> Microsoft.ML.OnnxRuntimeGenAI is the CPU package Microsoft.ML.OnnxRuntimeGenAI.Cuda is the Cuda package Microsoft.ML.OnnxRuntimeGenAI.DirectML is the DirectML package
> 
> The problem is that when Debug_DirectML is enabled, the CPU package will also be installed since it has no condition associated with it.

Ahh I though we needed `Microsoft.ML.OnnxRuntimeGenAI` + `Microsoft.ML.OnnxRuntimeGenAI.XXXX`

added CPU flags

baijumeswani:
Thanks for the contribution. :)

PR #940, opened 2024-09-27
Add new INT4 quantization features to model builder

kunal-vaishnavi:
### Description

This PR adds new INT4 quantization features to the model builder.

1. The model builder can now quantize the embedding layer and the language modeling head to INT4 precision by default.
2. For already-quantized PyTorch models that are passed to the model builder, any ops that are still created with `MatMul` can now be quantized to `MatMulNBits` via RTN.
3. A new optional flag in the extra options called `int4_op_types_to_quantize` has been added to allow more flexibility with INT4 quantization.

### Motivation and Context

With these PR changes, the size of the ONNX models can be reduced by quantizing the embedding layer and/or the language modeling head.

For the ONNX models built from already-quantized PyTorch models, one example is with using AutoAWQ. AutoAWQ does not quantize the language modeling head. The resulting ONNX model typically contains a `MatMul` op for the language modeling head. Now, that `MatMul` op will be quantized via RTN to `MatMulNBits` to reduce memory.


src/python/py/models/builder.py
```diff
@@ -275,8 +275,9 @@ def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
         # Quantization-specific variables (INT4, INT8, etc.)
         self.quant_attrs = {
             "int4": {
-                "block_size": int(extra_options["int4_block_size"]) if "int4_block_size" in extra_options else 32,
                 "accuracy_level": int(extra_options["int4_accuracy_level"]) if "int4_accuracy_level" in extra_options else 0,   # Default is 0 for non-QDQ formats, default is 4 for QDQ formats
+                "block_size": int(extra_options["int4_block_size"]) if "int4_block_size" in extra_options else 32,
+                "op_types_to_quantize": (extra_options["op_type_to_quantize"], ) if "int4_op_type_to_quantize" in extra_options else ("MatMul", "Gather", ),
```
yufenglee, line 280: Could we make the default "MatMul"?

kunal-vaishnavi: Fixed

fajin-corp (approved):
:shipit:

src/python/py/models/builder.py
```diff
@@ -401,7 +402,7 @@ def save_model(self, out_dir):
 
         # Quantize ONNX model to desired precision
         # TODO: Replace by quantizing the MatMuls as they are created
-        if self.onnx_dtype == "int4" and self.quant_type is None:
+        if self.onnx_dtype == "int4":
             model = self.to_int4(model)
```
hanbitmyths, line 406: Is there a check to prevent quantizing twice somewhere? When an AWQ model is an input, is this code not running a quantization on already quantized nodes?

kunal-vaishnavi: In ORT's 4-bits quantizer, the [target op types](https://github.com/microsoft/onnxruntime/blob/7b9db658c3930aec0ea9f8bd550145f87df1f59f/onnxruntime/python/tools/quantization/matmul_4bits_quantizer.py#L1070-L1074) are `MatMul` and `Gather`.

The below check is for which MatMul op(s) to create.

https://github.com/microsoft/onnxruntime-genai/blob/16b11ee9488430da8f0aa78e451d0df33af0e771/src/python/py/models/builder.py#L692-L701

For already-quantized models, `make_matmul_int4` will create a `MatMulNBits` node but `make_matmul_int4_qdq` will create `DequantizeLinear --> Transpose --> MatMul`.

I'll add a check to prevent quantizing the latter `MatMul` again for the already-quantized, use-QDQ case.

hanbitmyths (approved):

Issue #929, opened 2024-09-26
The option `use_env_allocators` crashes the inference with CPU EP


skyline75489:
**Describe the bug**

Introduced in #907, when there is no EP specified, `use_env_allocators` will be `true`. This crashes the inference on iOS with the following stacktrace:

```plaintext
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x11217c000)
    frame #0: 0x000000010543b7a8 TestGenAI.debug.dylib`onnxruntime::common::Status onnxruntime::contrib::RunRotaryEmbedding<float>(onnxruntime::concurrency::ThreadPool*, onnxruntime::contrib::rotary_embedding_helper::RotaryParameters, float const*, long long const*, float const*, float const*, float*, bool)::'lambda'(long, long)::operator()(long, long) const + 488
    frame #1: 0x0000000104d3e950 TestGenAI.debug.dylib`onnxruntime::concurrency::ThreadPool::ParallelFor(long, onnxruntime::TensorOpCost const&, std::__1::function<void (long, long)> const&) + 484
    frame #2: 0x000000010543a19c TestGenAI.debug.dylib`onnxruntime::common::Status onnxruntime::contrib::RunRotaryEmbedding<float>(onnxruntime::concurrency::ThreadPool*, onnxruntime::contrib::rotary_embedding_helper::RotaryParameters, float const*, long long const*, float const*, float const*, float*, bool) + 276
    frame #3: 0x000000010525a25c TestGenAI.debug.dylib`onnxruntime::contrib::GroupQueryAttention<float>::Compute(onnxruntime::OpKernelContext*) const + 2980
    frame #4: 0x0000000104dfc51c TestGenAI.debug.dylib`onnxruntime::ExecuteKernel(onnxruntime::StreamExecutionContext&, unsigned long, unsigned long, bool const&, onnxruntime::SessionScope&) + 312
    frame #5: 0x0000000104dbd3b4 TestGenAI.debug.dylib`onnxruntime::LaunchKernelStep::Execute(onnxruntime::StreamExecutionContext&, unsigned long, onnxruntime::SessionScope&, bool const&, bool&) + 48
    frame #6: 0x0000000104e2f3f8 TestGenAI.debug.dylib`onnxruntime::RunSince(unsigned long, onnxruntime::StreamExecutionContext&, onnxruntime::SessionScope&, bool const&, unsigned long) + 216
[...]

PR #941, opened 2024-09-28
Keep arena config alive

baijumeswani:
Fixes https://github.com/microsoft/onnxruntime-genai/issues/929


baijumeswani:
Had my test wrong. This does not fix the issue.

PR #955, opened 2024-10-05
LoRA support in model builder

apsonawane:
This PR adds LoRA MatMul changes in model builder. It includes changes made by Kunal and few changes done to make it work with Olive.
It covers the scenario where base_layer of float and adapters are float.

This PR will be followed with support for quant models scenario

### DIFF
--- a/src/python/py/models/README.md
+++ b/src/python/py/models/README.md
@@ -100,6 +100,20 @@ python3 -m onnxruntime_genai.models.builder -i path_to_local_folder_on_disk -o p
 python3 builder.py -i path_to_local_folder_on_disk -o path_to_output_folder -p int4 -e execution_provider -c cache_dir_to_store_temp_files
 ```
 
+### LoRA Models
+
+This scenario is where you have a finetuned model which supports LoRA and want to get optimized model. To use adapter model you have to pass the path of the model as input
+
+path_to_local_folder_on_disk --> location where adapter_config.json and adapter weights are present
+
+```
+# From wheel:
+python3 -m onnxruntime_genai.models.builder -i path_to_local_folder_on_disk -o path_to_output_folder -p fp16 -e execution_provider -c cache_dir_to_store_temp_files
+
+# From source:
+python3 builder.py -i path_to_local_folder_on_disk -o path_to_output_folder -p fp16 -e execution_provider -c cache_dir_to_store_temp_files
+```
+
 ### GGUF Model
 
 This scenario is where your float16/float32 GGUF model is already on disk.
--- a/src/python/py/models/README.md
+++ b/src/python/py/models/README.md
@@ -20,6 +20,7 @@ This folder contains the model builder for quickly creating optimized and quanti
     - [Use 8 Bits Quantization in QMoE](#use-8-bits-quantization-in-qmoe)
     - [Hugging Face Authentication](#hugging-face-authentication)
     - [Use QDQ Pattern for Quantization](#use-qdq-pattern-for-quantization)
+    - [LoRA Models](#lora-models)
   - [Unit Testing Models](#unit-testing-models)
     - [Option 1: Use the model builder directly](#option-1-use-the-model-builder-directly)
     - [Option 2: Edit the config.json file](#option-2-edit-the-configjson-file-on-disk-and-then-run-the-model-builder)
@@ -98,20 +99,6 @@ python3 -m onnxruntime_genai.models.builder -i path_to_local_folder_on_disk -o p
 
 # From source:
 python3 builder.py -i path_to_local_folder_on_disk -o path_to_output_folder -p int4 -e execution_provider -c cache_dir_to_store_temp_files
-```
-
-### LoRA Models
-
-This scenario is where you have a finetuned model which supports LoRA and want to get optimized model. To use adapter model you have to pass the path of the model as input
-
-path_to_local_folder_on_disk --> location where adapter_config.json and adapter weights are present
-
-```
-# From wheel:
-python3 -m onnxruntime_genai.models.builder -o path_to_output_folder -p fp16 -e execution_provider -c cache_dir_to_store_temp_files --extra_options adapter_path=path_to_adapter_files
-
-# From source:
-python3 builder.py -o path_to_output_folder -p fp16 -e execution_provider -c cache_dir_to_store_temp_files --extra_options adapter_path=path_to_adapter_files
 ```
 
 ### GGUF Model
@@ -229,6 +216,20 @@ python3 -m onnxruntime_genai.models.builder -i path_to_local_folder_on_disk -o p
 
 # From source:
 python3 builder.py -i path_to_local_folder_on_disk -o path_to_output_folder -p precision -e execution_provider -c cache_dir_to_store_temp_files --extra_options use_qdq=1
+```
+
+#### LoRA Models
+
+This scenario is where you have a finetuned model with LoRA adapters and your model can be loaded in the Hugging Face style via [PEFT](https://github.com/huggingface/peft).
+
+- path_to_local_folder_on_disk = location where base_model's weights are present
+
+```
+# From wheel:
+python3 -m onnxruntime_genai.models.builder -i path_to_local_folder_on_disk -o path_to_output_folder -p fp16 -e execution_provider -c cache_dir_to_store_temp_files --extra_options adapter_path=path_to_adapter_files
+
+# From source:
+python3 builder.py -i path_to_local_folder_on_disk -o path_to_output_folder -p fp16 -e execution_provider -c cache_dir_to_store_temp_files --extra_options adapter_path=path_to_adapter_files
 ```
 
 ### Unit Testing Models
--- a/src/python/py/models/README.md
+++ b/src/python/py/models/README.md
@@ -108,10 +108,10 @@ path_to_local_folder_on_disk --> location where adapter_config.json and adapter 
 
 ```
 # From wheel:
-python3 -m onnxruntime_genai.models.builder -i path_to_local_folder_on_disk -o path_to_output_folder -p fp16 -e execution_provider -c cache_dir_to_store_temp_files
+python3 -m onnxruntime_genai.models.builder -o path_to_output_folder -p fp16 -e execution_provider -c cache_dir_to_store_temp_files --extra_options adapter_path=path_to_adapter_files
 
 # From source:
-python3 builder.py -i path_to_local_folder_on_disk -o path_to_output_folder -p fp16 -e execution_provider -c cache_dir_to_store_temp_files
+python3 builder.py -o path_to_output_folder -p fp16 -e execution_provider -c cache_dir_to_store_temp_files --extra_options adapter_path=path_to_adapter_files
 ```
 
 ### GGUF Model
--- a/src/python/py/models/builder.py
+++ b/src/python/py/models/builder.py
@@ -677,13 +677,9 @@ class Model:
         self.make_value_info(output, dtype, shape=shape)
 
     def make_matmul(self, matmul, basename, root_input, **kwargs):
-        print("Matmul attr = ", matmul)
         if hasattr(matmul, "base_layer"):
             # For LoRA `MatMul`
-            print("Making lora matmul")
             return self.make_matmul_lora(matmul, basename, root_input, **kwargs)
-        # elif self.matmul_attrs["use_qdq"]:
-        #     return self.make_matmul_qdq(matmul, basename, root_input, **kwargs)
         else:
             # For regular `MatMul`
             return self.make_matmul_op(matmul, basename, root_input, **kwargs)
@@ -700,9 +696,7 @@ class Model:
             raise NotImplementedError(f"The {self.onnx_dtype} precision is not currently supported.")
 
     def make_matmul_fp16_or_fp32(self, matmul, name, root_input, **kwargs):
-        end_position = name.rfind('/')
-        sliced_name = name[:end_position]
-        weight = sliced_name[1:].replace("/", ".") + ".weight"
+        weight = name[1:].replace("/", ".") + ".weight"
         self.make_external_tensor(matmul.weight.detach().numpy().transpose().astype(self.to_numpy_dtype[self.io_dtype]), weight)
 
         last_dim = matmul.weight.shape[0]
@@ -1241,7 +1235,7 @@ class Model:
         #                Transpose
         #                    |
         #                 Reshape
-        basename = f"/model/layers.{layer_id}/self_attn/{'k_proj' if past_kv.endswith('key') else 'v_proj'}/repeat_kv"
+        basename = f"/model/layers.{layer_id}/attn/{'k_proj' if past_kv.endswith('key') else 'v_proj'}/repeat_kv"
 
         # Make the initial subgraph
         #
@@ -1444,17 +1438,17 @@ class Model:
         # Make MatMul nodes
         if self.attention_attrs["use_packed_matmul"]:
             # Combine 3 MatMuls into 1 packed MatMul
-            qkv_matmul_basename = f"/model/layers.{layer_id}/self_attn/qkv_proj/MatMul"
+            qkv_matmul_basename = f"/model/layers.{layer_id}/attn/qkv_proj/MatMul"
             qkv_matmul_name = self.make_packed_matmul(attention.q_proj, attention.k_proj, attention.v_proj, qkv_matmul_basename, root_input)
             self.attention_attrs["q_path"] = f"{qkv_matmul_name}/output_0"
         else:
-            q_matmul_basename = f"/model/layers.{layer_id}/self_attn/q_proj/MatMul"
+            q_matmul_basename = f"/model/layers.{layer_id}/attn/q_proj/MatMul"
             q_matmul_name = self.make_matmul(attention.q_proj, q_matmul_basename, root_input)
             self.attention_attrs["q_path"] = f"{q_matmul_name}/output_0"
-            k_matmul_basename = f"/model/layers.{layer_id}/self_attn/k_proj/MatMul"
+            k_matmul_basename = f"/model/layers.{layer_id}/attn/k_proj/MatMul"
             k_matmul_name = self.make_matmul(attention.k_proj, k_matmul_basename, root_input)
             self.attention_attrs["k_path"] = f"{k_matmul_name}/output_0"
-            v_matmul_basename = f"/model/layers.{layer_id}/self_attn/v_proj/MatMul"
+            v_matmul_basename = f"/model/layers.{layer_id}/attn/v_proj/MatMul"
             v_matmul_name = self.make_matmul(attention.v_proj, v_matmul_basename, root_input)
             self.attention_attrs["v_path"] = f"{v_matmul_name}/output_0"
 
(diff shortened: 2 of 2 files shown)


src/python/py/models/builder.py
```diff
@@ -2799,7 +2963,16 @@ def create_model(model_name, input_path, output_dir, precision, execution_provid
     extra_kwargs = {} if os.path.isdir(input_path) else {"cache_dir": cache_dir}
     hf_name = input_path if os.path.isdir(input_path) else model_name
     hf_token = parse_hf_token(extra_options.get("hf_token", "true"))
-    config = AutoConfig.from_pretrained(hf_name, token=hf_token, trust_remote_code=True, **extra_kwargs)
+
+    is_peft = os.path.isdir(adapter_path) and "adapter_config.json" in os.listdir(adapter_path)
```
jambayk, line 2967: bring this discussion online: the check for `isdir` and `adapter_config.json` requires that the `adapter_path` is local. Is this check required? 
I think the `isdir` check in line 2965 for `input_path` is present to differentiate between a local hf checkpoint (dir) and a local gguf checkpoint (file). 

Since we don't have such issues for adapter_path, checking for `not adapter_path` might be enough? this way it can also support adapters on hf-hub.

src/python/py/models/builder.py
```diff
@@ -2799,7 +2963,16 @@ def create_model(model_name, input_path, output_dir, precision, execution_provid
...
+    peft_model_name = adapter_path if is_peft else None
+    peft_config = None
+    if is_peft:
+        from peft import PeftConfig
+        peft_config = PeftConfig.from_pretrained(peft_model_name, use_auth_token=True, trust_remote_code=True, **extra_kwargs)
```
jambayk, line 2972: use_auth_token was replaced with `token=hf_token` in the deleted line 2802

natke (commented):
Can you please update the README with any new options and usage of this feature? https://github.com/microsoft/onnxruntime-genai/blob/main/src/python/py/models/README.md

src/python/py/models/builder.py
```diff
@@ -157,6 +157,13 @@ def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
             "add_offset": 0,            # Offset value for LayerNorm weight
         }
 
+        # MatMul-specific variables
+        is_lora = hasattr(config, "peft_type") and config.peft_type == "LORA"
+        self.matmul_attrs = {
+            "use_lora": is_lora,        # Use LoRA/QLoRA format
+            "use_qdq": False,           # Use QDQ format
```
kunal-vaishnavi, line 164: Originally, I added `use_qdq` in `self.matmul_attrs` to prepare for the QDQ work and support both QDQ and LoRA/QLoRA. However, the QDQ work was eventually added in [this PR](https://github.com/microsoft/onnxruntime-genai/pull/900).

Given this, can we replace any references to `self.use_qdq` with `self.matmul_attrs["use_qdq"]` and change the `use_qdq` entry to the below?

https://github.com/microsoft/onnxruntime-genai/blob/b3dfed65da067f96be5d9b464a3e0b26496b8621/src/python/py/models/builder.py#L41

Then, we can remove `self.use_qdq`.

src/python/py/models/builder.py
```diff
@@ -665,6 +676,18 @@ def make_tanh(self, name, root_input, dtype, shape):
         self.make_value_info(output, dtype, shape=shape)
 
     def make_matmul(self, matmul, basename, root_input, **kwargs):
+        print("Matmul attr = ", matmul)
```
kunal-vaishnavi, line 679:
```suggestion
```

src/python/py/models/builder.py
```diff
@@ -665,6 +676,18 @@ def make_tanh(self, name, root_input, dtype, shape):
...
+        if hasattr(matmul, "base_layer"):
+            # For LoRA `MatMul`
+            print("Making lora matmul")
```
kunal-vaishnavi, line 682:
```suggestion
```

src/python/py/models/builder.py
```diff
@@ -665,6 +676,18 @@ def make_tanh(self, name, root_input, dtype, shape):
...
+            return self.make_matmul_lora(matmul, basename, root_input, **kwargs)
+        # elif self.matmul_attrs["use_qdq"]:
+        #     return self.make_matmul_qdq(matmul, basename, root_input, **kwargs)
```
kunal-vaishnavi, line 685: Can we remove the commented out code for QDQ since it's supported here?

https://github.com/microsoft/onnxruntime-genai/blob/0f59a900568cf65d74554cc34027648989f4ab1c/src/python/py/models/builder.py#L667-L676

src/python/py/models/builder.py
```diff
@@ -1177,7 +1240,7 @@ def make_repeat_kv(self, layer_id, root_input, past_kv, present_kv, **kwargs):
         #                Transpose
         #                    |
         #                 Reshape
-        basename = f"/model/layers.{layer_id}/attn/{'k_proj' if past_kv.endswith('key') else 'v_proj'}/repeat_kv"
+        basename = f"/model/layers.{layer_id}/self_attn/{'k_proj' if past_kv.endswith('key') else 'v_proj'}/repeat_kv"
```
kunal-vaishnavi, line 1243: I think we should revert the change from `attn` to `self_attn`. There will be models in the future that use `cross_attn` and this naming would be wrong for those.

src/python/py/models/builder.py
```diff
@@ -1406,87 +1465,82 @@ def make_attention(self, layer_id, attention, root_input, **kwargs):
...
             o_bias = eval(f"attention.{o_proj}.bias.detach().numpy()")
             self.make_add_bias(o_bias, o_add_name, root_input=f"{o_matmul_name}/output_0")
 
         # Assign output 0 of previous output node as skip input to next SkipLayerNorm
         self.layernorm_attrs["skip_input"] = f"{o_matmul_name if not o_bias_exists else o_add_name}/output_0"
 
     def make_attention_unpacked(self, layer_id, attention, root_input, **kwargs):
-        q_size = self.num_attn_heads * self.head_size
-        kv_size = self.num_kv_heads * self.head_size
+        # q_size = self.num_attn_heads * self.head_size
```
kunal-vaishnavi, line 1532:
```suggestion
```

src/python/py/models/builder.py
```diff
@@ -1406,87 +1465,82 @@ def make_attention(self, layer_id, attention, root_input, **kwargs):
...
+        # kv_size = self.num_kv_heads * self.head_size
```
kunal-vaishnavi, line 1533:
```suggestion
```

src/python/py/models/builder.py
```diff
@@ -676,7 +699,9 @@ def make_matmul(self, matmul, basename, root_input, **kwargs):
             raise NotImplementedError(f"The {self.onnx_dtype} precision is not currently supported.")
 
     def make_matmul_fp16_or_fp32(self, matmul, name, root_input, **kwargs):
-        weight = name[1:].replace("/", ".") + ".weight"
+        end_position = name.rfind('/')
+        sliced_name = name[:end_position]
```
kunal-vaishnavi, line 703: Why is this slicing logic needed?

src/python/py/models/builder.py
```diff
@@ -1792,7 +1950,7 @@ def make_layer(self, layer_id, layer):
             # Norm after last decoder layer of model (last layer --> norm)
             self.layernorm_attrs["last_layernorm"] = True
 
-    def make_model(self, input_path):
+    def make_model(self, input_path, adapter_path):
```
kunal-vaishnavi, line 1953: We should remove `adapter_path` because the `input_path` should contain all necessary files (PyTorch weights, JSON files, etc). Otherwise, we would have to provide separate arguments for the paths to many types of files and that is not a scalable approach.

src/python/py/models/builder.py
apsonawane: Yeah I think I did not do the rebase correctly

src/python/py/models/builder.py
apsonawane: rebase did not happen properly. Fixing it

src/python/py/models/builder.py
apsonawane: rebase did not happen properly. Fixing it

src/python/py/models/builder.py
apsonawane: adapter_config.json and adapter safetensors file will be in separate folder right? Also when we are using a quantized model we have to pass input_path so it will be conflict between adapter path and input path?

src/python/py/models/builder.py
apsonawane: Okay, I made this change to make it work with Olive. I think we should do the same change in Olive as well

jambayk: In olive, the names of the nodes are automatically created during torch export which in turn comes from the submodule names from the pytorch model itself. 
I don't think we could always rename self_attn to attn since we don't know what model it might be but maybe we can provide an option to the `convert-adapter` command to do this replacement? https://github.com/microsoft/Olive/blob/main/olive/cli/convert_adapters.py

src/python/py/models/builder.py
jambayk: @kunal-vaishnavi akshay and I had a long discussion about this. 
- The adapter config saves the path/name of whatever the model it was trained on. this means that if an adapter is shared on hf-hub or through some other means the base_model_name_or_path might not be valid anymore.
  - valid case: 
    - was trained on an hf-id model like `meta-llama/Llama-2-7b-hf`
    - trained on a local model "local_path" and ran builder from the same place
  - invalid case:
   - trained on a non hf-id model like a locally download model, azureml registry model, github model and model builder run from a different machine
   
From my point of view, it is safer and easier logic wise to treat the base model and adapters separately. We don't know how or where the adapters were trained on so we cannot reliably get the base model information from the adapter config. This also limits the support for local adapters + the valid case above. 
For example, it would not work for an adapter on the model hub like https://huggingface.co/Mikael110/llama-2-7b-guanaco-qlora/blob/main/adapter_config.json

jambayk: tldr: adapter path is not a special case of the model format like gguf is

kunal-vaishnavi: For PyTorch models, the model builder only works for models that are compatible with Hugging Face's APIs currently. For example, it will fail if you provide `meta-llama/Llama-2-7b` as the model name and it will pass if you provide `meta-llama/Llama-2-7b-hf` as the model name.

Since we assume that the base PyTorch model can be loaded with `AutoModelForCausalLM.from_pretrained()`, I think it is fine to assume that the adapters can be loaded with `AutoPeftModelForCausalLM.from_pretrained()` since it comes from Hugging Face's `peft` package and `AutoPeftModelForCausalLM.from_pretrained()` will anyways invoke `AutoModelForCausalLM.from_pretrained()` when loading the base model. This also lets `AutoPeftModelForCausalLM.from_pretrained()` handle the local path vs. model name differences.

Most Hugging Face repos have the correct `base_model_name_or_path` entry. For the ones that don't, one option is that we can recommend users to download the repo locally, fix the `base_model_name_or_path` entry in the `adapter_config.json` file, and then run the model builder.

Alternatively, we can use the existing `-m/--model_name` and `-i/--input_path` flags to [load a `PeftModel`](https://github.com/huggingface/peft/blob/749b92456218f7dddc8f7a9aa27a41815b3d6c2e/src/peft/peft_model.py#L427) in the [same way that a GGUF model is loaded](https://github.com/microsoft/onnxruntime-genai/blob/main/src/python/py/models/README.md#gguf-model).

```python
from peft import PeftModel

peft_model = PeftModel.from_pretrained(args.model_name, args.input_path)
```

By using both the `-m` and `-`i flags, the `-a/--adapter_path` is not needed.

> For example, it would not work for an adapter on the model hub like https://huggingface.co/Mikael110/llama-2-7b-guanaco-qlora/blob/main/adapter_config.json

Yes, this would anyways fail since the repo is not compatible with the Hugging Face APIs.

src/python/py/models/builder.py
apsonawane: Okay, I will revert this back

src/python/py/models/builder.py
apsonawane: Not required reverting this back

src/python/py/models/builder.py
jambayk: this is not a blocker for the olive-based workflow but just wanted to share the different scenarios for the adapters. 

https://huggingface.co/Mikael110/llama-2-7b-guanaco-qlora/blob/main/adapter_config.json was actually trained on https://huggingface.co/meta-llama/Llama-2-7b-hf (as per their readme and my local usage). the `"/root/llama2-7b-hf"` was the local path where the base model was downloaded before the finetuning.

Maybe the alternative is a good compromise. My main concern is that coupling the base model and adapter model limits the use case scenarios. perhaps it's acceptable trade-off for you.
Another thing I want to note is that the base model can also be already int4 quantized using awq or gptq + adapters in float. In that case, `AutoPeftModelForCausalLM.from_pretrained()` probably won't work since the builder loads quantized models differently.

src/python/py/models/builder.py
apsonawane: Removed the adapter_path arg so no longer required

src/python/py/models/builder.py
jambayk: just a heads up, this is also a use case:
- quantized base model, float adapters

what would happen in that case? both adapter and quantized model use `i`/`input_path`

hanbitmyths: It looks to me that adapter_config.json is adding more special cares. Adding an extra argument is simpler and less confused here. Is there any other concern other than adding new argument?

kunal-vaishnavi: I added the check for `adapter_config.json` while I was adding the LoRA support because adapters created by Hugging Face's `peft` have an associating `adapter_config.json` file. Since we have agreed to add the `adapter_path` in the `extra_options` for now and revisit this design later, I think we can remove the check for `adapter_config.json`.

apsonawane: Yes, I will make the changes

src/python/py/models/README.md
```diff
@@ -100,6 +100,20 @@ python3 -m onnxruntime_genai.models.builder -i path_to_local_folder_on_disk -o p
...
 
+### LoRA Models
+
+This scenario is where you have a finetuned model which supports LoRA and want to get optimized model. To use adapter model you have to pass the path of the model as input
+
+path_to_local_folder_on_disk --> location where adapter_config.json and adapter weights are present
+
+```
+# From wheel:
+python3 -m onnxruntime_genai.models.builder -o path_to_output_folder -p fp16 -e execution_provider -c cache_dir_to_store_temp_files --extra_options adapter_path=path_to_adapter_files
```
kunal-vaishnavi, line 111: Can you add back the input path for the base model's weights?

```suggestion
python3 -m onnxruntime_genai.models.builder -i path_to_local_folder_on_disk -o path_to_output_folder -p fp16 -e execution_provider -c cache_dir_to_store_temp_files --extra_options adapter_path=path_to_adapter_files
```

src/python/py/models/README.md
```diff
@@ -100,6 +100,20 @@ python3 -m onnxruntime_genai.models.builder -i path_to_local_folder_on_disk -o p
...
+
+# From source:
+python3 builder.py -o path_to_output_folder -p fp16 -e execution_provider -c cache_dir_to_store_temp_files --extra_options adapter_path=path_to_adapter_files
```
kunal-vaishnavi, line 114: Can you add back the input path for the base model's weights?

```suggestion
python3 builder.py -i path_to_local_folder_on_disk -o path_to_output_folder -p fp16 -e execution_provider -c cache_dir_to_store_temp_files --extra_options adapter_path=path_to_adapter_files
```

src/python/py/models/README.md
```diff
@@ -100,6 +100,20 @@ python3 -m onnxruntime_genai.models.builder -i path_to_local_folder_on_disk -o p
 python3 builder.py -i path_to_local_folder_on_disk -o path_to_output_folder -p int4 -e execution_provider -c cache_dir_to_store_temp_files
 ```
 
+### LoRA Models
+
+This scenario is where you have a finetuned model which supports LoRA and want to get optimized model. To use adapter model you have to pass the path of the model as input
+
+path_to_local_folder_on_disk --> location where adapter_config.json and adapter weights are present
```
kunal-vaishnavi, line 107:
```suggestion
- path_to_local_folder_on_disk = location where base model's weights are present
- path_to_adapter_files = location where adapter weights are present

```

src/python/py/models/README.md
```diff
@@ -100,6 +100,20 @@ python3 -m onnxruntime_genai.models.builder -i path_to_local_folder_on_disk -o p
 python3 builder.py -i path_to_local_folder_on_disk -o path_to_output_folder -p int4 -e execution_provider -c cache_dir_to_store_temp_files
 ```
 
+### LoRA Models
```
kunal-vaishnavi, line 103: Since this uses `extra_options`, can you move this section to be after "Use QDQ Pattern for Quantization" and add a reference to this section in the table of contents after "Use QDQ Pattern for Quantization"?

https://github.com/microsoft/onnxruntime-genai/blob/dfbe14c39bc0486e1289332bca2003ff66a74fc7/src/python/py/models/README.md?plain=1#L5-L26

Can you also change the title's size so that it matches the title sizes that are sub-sections in the "Extra Options"?

```suggestion
#### LoRA Models
```

src/python/py/models/README.md
```diff
@@ -100,6 +100,20 @@ python3 -m onnxruntime_genai.models.builder -i path_to_local_folder_on_disk -o p
...
+
+This scenario is where you have a finetuned model which supports LoRA and want to get optimized model. To use adapter model you have to pass the path of the model as input
```
kunal-vaishnavi, line 105:
```suggestion
This scenario is where you have a finetuned model with LoRA adapters and your model can be loaded in the Hugging Face style via [PEFT](https://github.com/huggingface/peft).
```

src/python/py/models/builder.py
```diff
@@ -2897,9 +3050,19 @@ def create_model(model_name, input_path, output_dir, precision, execution_provid
 
     # Load model config
     extra_kwargs = {} if os.path.isdir(input_path) else {"cache_dir": cache_dir}
-    hf_name = input_path if os.path.isdir(input_path) else model_name
+    hf_name = input_path if os.path.isdir(input_path) and "config.json" in os.listdir(input_path) else model_name
```
kunal-vaishnavi, line 3053: Can this be reverted now that `adapter_path` is in the `extra_options`?

src/python/py/models/builder.py
```diff
@@ -2897,9 +3050,19 @@ def create_model(model_name, input_path, output_dir, precision, execution_provid
...
     hf_token = parse_hf_token(extra_options.get("hf_token", "true"))
+
+    is_peft = os.path.isdir(input_path) and "adapter_config.json" in os.listdir(input_path)
+    peft_model_name = input_path if is_peft else None
+    peft_config = None
+    if is_peft:
```
kunal-vaishnavi, line 3059: Can this section be simplified to the following?

```python
config = AutoConfig.from_pretrained(hf_name, token=hf_token, trust_remote_code=True, **extra_kwargs)

if "adapter_path" in extra_options:
    from peft import PeftConfig
    peft_config = PeftConfig.from_pretrained(extra_options["adapter_path"], token=hf_token, trust_remote_code=True, **extra_kwargs)
    config.update(peft_config.__dict__)
```

kunal-vaishnavi:
Can you add `adapter_path` and explain its usage after `use_qdq`?

https://github.com/microsoft/onnxruntime-genai/blob/2687e60740e5b57bdd67fdb563b91316988ddbd6/src/python/py/models/builder.py#L3181-L3213

src/python/py/models/builder.py
```diff
@@ -774,6 +793,44 @@ def make_matmul_int4_qdq(self, matmul, matmul_name, root_input, **kwargs):
         self.make_value_info(matmul_output, self.io_dtype, shape=['batch_size', 'sequence_length', matmul.out_features])
 
         return matmul_name
+    def make_matmul_lora(self, matmul, basename, root_input, **kwargs):
```
kunal-vaishnavi, line 796:
```suggestion

    def make_matmul_lora(self, matmul, basename, root_input, **kwargs):
```

src/python/py/models/builder.py
```diff
@@ -1855,6 +2004,10 @@ def make_model(self, input_path):
             extra_kwargs = {"num_hidden_layers": self.num_layers} if "num_hidden_layers" in self.extra_options else {}
             model = AutoModelForCausalLM.from_pretrained(self.model_name_or_path, cache_dir=self.cache_dir, token=self.hf_token, trust_remote_code=True, **extra_kwargs)
 
+        if "adapter_path" in self.extra_options:
+            from peft import PeftModel
+            model = PeftModel.from_pretrained(model, self.extra_options["adapter_path"])
```
kunal-vaishnavi, line 2009: Why are the `cache_dir`, `token`, `trust_remote_code`, and `extra_kwargs` parameters are omitted?

kunal-vaishnavi, line 2009: Can you add the `cache_dir`, `token`, `trust_remote_code`, and `extra_kwargs` parameters?

src/python/py/models/README.md
```diff
@@ -217,6 +217,20 @@ python3 -m onnxruntime_genai.models.builder -i path_to_local_folder_on_disk -o p
 python3 builder.py -i path_to_local_folder_on_disk -o path_to_output_folder -p precision -e execution_provider -c cache_dir_to_store_temp_files --extra_options use_qdq=1
 ```
 
+#### LoRA Models
```
kunal-vaishnavi, line 220: Can you add a [reference](https://github.com/microsoft/onnxruntime-genai/pull/955#discussion_r1807486816) to this new section in the table of contents?

src/python/py/models/builder.py
apsonawane: trust_remote_code is not a parameter to this function skipping that

apsonawane: extra_kwargs is also not required since the parameters for config.json and adapter_config.json are different

apsonawane: These parameters are not required but I still added cache and token

kunal-vaishnavi:
This can be done in another PR but we should add a LoRA model such as [this one](https://huggingface.co/argilla/phi2-lora-distilabel-intel-orca-dpo-pairs) in the CIs.

https://github.com/microsoft/onnxruntime-genai/blob/0f59a900568cf65d74554cc34027648989f4ab1c/test/python/_test_utils.py#L55-L77

The models in `hf_paths` are downloaded from Hugging Face, and the models in `ci_paths` are currently uploaded to `/data/ortgenai_pytorch_models` in one of the CI VMs.

https://github.com/microsoft/onnxruntime-genai/blob/0f59a900568cf65d74554cc34027648989f4ab1c/.github/workflows/linux-gpu-x64-build.yml#L122

kunal-vaishnavi (approved):

PR #956, opened 2024-10-05
Add Get/Set logits functions to Python

RyanUnderhill:
Also fixes an issue with the ToPython function, arrays returned were all the same element due to the stride being set to zero. This calls the explicit constructor.


ajindal1 (approved):

src/python/python.cpp
```diff
@@ -432,6 +443,8 @@ PYBIND11_MODULE(onnxruntime_genai, m) {
       .def("is_done", &PyGenerator::IsDone)
       .def("compute_logits", &PyGenerator::ComputeLogits)
       .def("get_output", &PyGenerator::GetOutput)
+      .def("get_logits", &PyGenerator::GetLogits)
+      .def("set_logits", &PyGenerator::SetLogits)
```
kunal-vaishnavi, line 447: Do `get_logits()` and `set_logits()` work with float16 and float32 precision or only float32 precision?

baijumeswani:
Do we need the C/C++ equivalent functions?

RyanUnderhill:
> Do we need the C/C++ equivalent functions? Could we also add a unit test?

These are needed now for the Whisper work, but we can always add the C APIs later.

src/python/python.cpp
RyanUnderhill: Always float32, as that's what our internal logits processing code uses. We convert whatever the model outputs to this to avoid needing multiple combinations of implementations for each type.

baijumeswani (approved):

PR #958, opened 2024-10-07
Initial support of Model - Nemotron-Mini-4B-Instruct

BLSharda:
Pending -
Some accuracy issue seen with DML, but CPU/CUDA works. 
GenAI tokenizer might need some handling.

### DIFF
--- a/src/models/model.cpp
+++ b/src/models/model.cpp
@@ -501,7 +501,7 @@ std::shared_ptr<Model> CreateModel(OrtEnv& ort_env, const char* config_path) {
 
   if (config->model.type == "gpt2")
     return std::make_shared<Gpt_Model>(std::move(config), ort_env);
-  if (config->model.type == "llama" || config->model.type == "gemma" || config->model.type == "gemma2" || config->model.type == "mistral" || config->model.type == "phi" || config->model.type == "phi3" || config->model.type == "phi3small" || config->model.type == "phimoe" || config->model.type == "qwen2")
+  if (config->model.type == "llama" || config->model.type == "gemma" || config->model.type == "gemma2" || config->model.type == "mistral" || config->model.type == "phi" || config->model.type == "phi3" || config->model.type == "phi3small" || config->model.type == "phimoe" || config->model.type == "qwen2" || config->model.type == "nemotron")
     return std::make_shared<DecoderOnly_Model>(std::move(config), ort_env);
   if (config->model.type == "whisper")
     return std::make_shared<Whisper_Model>(std::move(config), ort_env);
--- a/src/python/py/models/builder.py
+++ b/src/python/py/models/builder.py
@@ -1018,104 +1018,7 @@ class Model:
 
         return cos_cache_name, sin_cache_name
 
-
-    def make_rotary_embedding_partial(self, rotemb, name, root_input, **kwargs):
-        cos_cache_name, sin_cache_name = self.make_rotary_embedding_caches(rotemb)
-
-        #         root_input
-        #             |
-        #          Reshape
-        #             |
-        #           Split
-        #           /   \
-        #       Reshape  |
-        #          |     |
-        #     RoteryEmb  |
-        #          |     |
-        #       Reshape  |
-        #           \   /
-        #           Concat
-        #             |
-        #           Reshape
-
-        num_heads = (self.num_kv_heads if "k_rotary" in name else self.num_attn_heads)
-        dim_size = self.head_size * num_heads
-        const_prefix =  ("k_" if "k_rotary" in name else "v_")
-
-        # Reshape 1
-        reshape_name = f"{name}/Reshape/"
-        reshape_output_0 = f"{name}/Reshape/output_0"
-        rotary_reshape_dim = [0, 0, num_heads, int(dim_size / num_heads)]
-        constant_name = f"{const_prefix}RotaryEmb/Reshape/Const_input_dim_2"
-        constant_value = numpy_helper.from_array(np.array(rotary_reshape_dim, dtype="int64"))
-        self.make_node("Constant", inputs=[], outputs=[f"{constant_name}\constant_0"], name=constant_name, value=constant_value)
-        self.make_node("Reshape", inputs=[root_input, f"{constant_name}\constant_0"], outputs=[reshape_output_0], name=reshape_name)
-        self.make_value_info(reshape_output_0, self.io_dtype, shape=['batch_size', 'sequence_length', num_heads, int(dim_size / num_heads)])
-
-        # Split
-        split_name = f"{name}/Split/"
-        split_output_0 = f"{name}/Split/output_0"
-        split_output_1 = f"{name}/Split/output_1"
-        split_size = int(self.rotemb_attrs["partial_rotary_factor"] * int(dim_size/num_heads))
-        split_dim = [split_size, int(dim_size/num_heads) - split_size]
-        constant_split_name = f"{const_prefix}RotaryEmb/Split/Const_input_shape"
-        constant_split_value = numpy_helper.from_array(np.array(split_dim, dtype="int64"))
-        self.make_node("Constant", inputs=[], outputs=[f"{constant_split_name}\constant_0"], name=constant_split_name, value=constant_split_value)
-        self.make_node("Split", inputs=[reshape_output_0, f"{constant_split_name}\constant_0"], outputs=[split_output_0, split_output_1], name=split_name, axis=3)
-        self.make_value_info(split_output_0, self.io_dtype, shape=['batch_size', 'sequence_length', num_heads, split_dim[0]])
-        self.make_value_info(split_output_1, self.io_dtype, shape=['batch_size', 'sequence_length', num_heads, split_dim[1]])
-
-        # Reshape 2
-        reshape2_name = f"{name}/Reshape2/"
-        reshape2_output_0 =f"{name}/Reshape2/output_0"
-        rotary_reshape2_dim = [0, 0, num_heads * split_dim[0]]
-        constant2_name = f"{const_prefix}RotaryEmb/Reshape/Const2_input_dim_2"
-        constant2_value = numpy_helper.from_array(np.array(rotary_reshape2_dim, dtype="int64"))
-        self.make_node("Constant", inputs=[], outputs=[f"{constant2_name}\constant_0"], name=constant2_name, value=constant2_value)
-        self.make_node("Reshape", inputs=[split_output_0, f"{constant2_name}\constant_0"], outputs=[reshape2_output_0], name=reshape2_name)
-        self.make_value_info(reshape2_output_0, self.io_dtype, shape=['batch_size', 'sequence_length', num_heads * split_dim[0]])
-
-        # Rotary Embedding
-        inputs = [reshape2_output_0, kwargs.pop("position_ids"), cos_cache_name, sin_cache_name]
-        rot_output = f"{name}/Rotary/output_0"
-        # Removing duplicate kwargs 
-        kwargs.pop("num_heads", "")
-        kwargs.pop("rotary_embedding_dim", "")
-        self.make_node("RotaryEmbedding", inputs=inputs, outputs=[rot_output], name=name, domain="com.microsoft", interleaved=self.rotemb_attrs["interleaved"], num_heads=num_heads, rotary_embedding_dim=split_dim[0], **kwargs)
-        self.make_value_info(rot_output, self.io_dtype, shape=['batch_size', 'sequence_length', num_heads * split_dim[0]])
-
-        # Reshape 3
-        reshape3_name = f"{name}/Reshape3/"
-        reshape3_output_0 =f"{name}/Reshape3/output_0"
-        rotary_reshape3_dim = [0, 0, num_heads, split_dim[0]]
-        constant3_name = f"{const_prefix}RotaryEmb/Reshape/Const3_input_dim_2"
-        constant3_value = numpy_helper.from_array(np.array(rotary_reshape3_dim, dtype="int64"))
-        self.make_node("Constant", inputs=[], outputs=[f"{constant3_name}\constant_0"], name=constant3_name, value=constant3_value)
-        self.make_node("Reshape", inputs=[rot_output, f"{constant3_name}\constant_0"], outputs=[reshape3_output_0], name=reshape3_name)
-        self.make_value_info(reshape3_output_0, self.io_dtype, shape=['batch_size', 'sequence_length', num_heads, split_dim[0]])
-
-        # Concat
-        concat_name = f"{name}/Concat"
-        concat_inputs = [reshape3_output_0, split_output_1]
-        concat_output = f"{name}/Concat/output_0"
-        self.make_node("Concat", inputs=concat_inputs, outputs=[concat_output], name=concat_name, axis=3)
-        self.make_value_info(concat_output, self.io_dtype, shape=['batch_size', 'sequence_length', num_heads, split_dim[0] + split_dim[1]])
-
-        # Reshape 4
-        reshape4_name = f"{name}/Reshape4/"
-        reshape4_output_0 =f"{name}/output_0"
-        rotary_reshape4_dim = [0, 0, dim_size]
-        constant4_name = f"{const_prefix}RotaryEmb/Reshape/Const4_input_dim_2"
-        constant4_value = numpy_helper.from_array(np.array(rotary_reshape4_dim, dtype="int64"))
-        self.make_node("Constant", inputs=[], outputs=[f"{constant4_name}\constant_0"], name=constant4_name, value=constant4_value)
-        self.make_node("Reshape", inputs=[concat_output, f"{constant4_name}\constant_0"], outputs=[reshape4_output_0], name=reshape4_name)
-        self.make_value_info(reshape4_output_0, self.io_dtype, shape=['batch_size', 'sequence_length', dim_size])
-
     def make_rotary_embedding(self, rotemb, name, root_input, **kwargs):
-        if self.rotemb_attrs["partial_rotary_factor"] > 0:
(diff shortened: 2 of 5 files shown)


natke:
Thank you @BLSharda. We should have support for the tokenizer now too.

BLSharda:
> * By signing below, the defined term “You” includes me and my employer.

BLSharda:
> Thank you @BLSharda. We should have support for the tokenizer now too.

Thanks, I verified that the tokenizer code works out of the box now.

src/models/model.cpp
```diff
@@ -495,7 +495,7 @@ std::shared_ptr<Model> CreateModel(OrtEnv& ort_env, const char* config_path) {
 
   if (config->model.type == "gpt2")
     return std::make_shared<Gpt_Model>(std::move(config), ort_env);
-  if (config->model.type == "llama" || config->model.type == "gemma" || config->model.type == "gemma2" || config->model.type == "mistral" || config->model.type == "phi" || config->model.type == "phi3" || config->model.type == "phi3small" || config->model.type == "phimoe" || config->model.type == "qwen2")
+  if (config->model.type == "llama" || config->model.type == "gemma" || config->model.type == "gemma2" || config->model.type == "mistral" || config->model.type == "phi" || config->model.type == "phi3" || config->model.type == "phi3small" || config->model.type == "phimoe" || config->model.type == "qwen2" || config->model.type == "nemotron")
```
pranavsharma, line 498: This looks like something that _must_ be refactored. @RyanUnderhill - please resolve this separately from this PR. A simple option is to have a hash map that maps the model type to the class you want to instantiate.

src/python/py/models/builder.py
```diff
@@ -1018,7 +1018,104 @@ def make_rotary_embedding_caches(self, rotemb, **kwargs):
 
         return cos_cache_name, sin_cache_name
 
+
+    def make_rotary_embedding_partial(self, rotemb, name, root_input, **kwargs):
+        cos_cache_name, sin_cache_name = self.make_rotary_embedding_caches(rotemb)
```
kunal-vaishnavi, line 1023: Partial rotary embeddings are already supported in the `RotaryEmbedding` op by using the `num_heads` and `rotary_embedding_dim` [attributes](https://github.com/microsoft/onnxruntime/blob/main/docs/ContribOperators.md#com.microsoft.RotaryEmbedding).

From reading the Hugging Face implementation of [Nemotron's rotary embeddings](https://github.com/huggingface/transformers/blob/70b07d97cf2c5f61fff55700b65528a1b6845cd2/src/transformers/models/nemotron/modeling_nemotron.py#L180-L187), the logic looks to be the same as Hugging Face's implementation of [Phi-2's rotary embeddings](https://github.com/huggingface/transformers/blob/70b07d97cf2c5f61fff55700b65528a1b6845cd2/src/transformers/models/phi/modeling_phi.py#L324-L338).

Similar to how the Phi-2 model class is created, can we enable these two attributes in the Nemotron model class to use the partial rotary embeddings instead of creating `make_rotary_embedding_partial`?

https://github.com/microsoft/onnxruntime-genai/blob/0f59a900568cf65d74554cc34027648989f4ab1c/src/python/py/models/builder.py#L2327-L2333

src/python/py/models/builder.py
```diff
@@ -1018,7 +1018,104 @@ def make_rotary_embedding_caches(self, rotemb, **kwargs):
...
+        dim_size = self.head_size * num_heads
+        const_prefix =  ("k_" if "k_rotary" in name else "v_")
+
+        # Reshape 1
+        reshape_name = f"{name}/Reshape/"
+        reshape_output_0 = f"{name}/Reshape/output_0"
+        rotary_reshape_dim = [0, 0, num_heads, int(dim_size / num_heads)]
+        constant_name = f"{const_prefix}RotaryEmb/Reshape/Const_input_dim_2"
+        constant_value = numpy_helper.from_array(np.array(rotary_reshape_dim, dtype="int64"))
+        self.make_node("Constant", inputs=[], outputs=[f"{constant_name}\constant_0"], name=constant_name, value=constant_value)
```
kunal-vaishnavi, line 1051: Code blocks like this can be simplified. Constant nodes are [auto-created](https://github.com/microsoft/onnxruntime-genai/blob/main/src/python/py/models/DESIGN.md#constants) when they are detected as a node's input. Nodes such as Reshape can also be created with `make_{op_type}` to simplify adding it to the model.

Here's an example that uses an array for the constant.

https://github.com/microsoft/onnxruntime-genai/blob/0f59a900568cf65d74554cc34027648989f4ab1c/src/python/py/models/builder.py#L1191-L1193

src/python/py/models/builder.py
```diff
@@ -1735,6 +1834,25 @@ def make_gelu(self, layer_id, root_input, activation):
...
+        self.make_value_info(output, self.io_dtype, shape=['batch_size', 'sequence_length', self.intermediate_size])
+        return relu_name
+
+    def make_relu_squared(self, layer_id, root_input, activation):
+        relu_name = self.make_relu(layer_id, root_input, "Relu")
+        basename = f"/model/layers.{layer_id}/mlp/square/{activation}"
+        pow_name = f"{basename}/pow"
+        output = f"{pow_name}/output_0"
+        constant_shape_name = "static_2"
+        const_value_2 = np.array([2])
```
kunal-vaishnavi, line 1850: As mentioned [here](https://github.com/microsoft/onnxruntime-genai/pull/958/files#r1796153035), Constant ops are auto-created. I think we can simplify this to the following.

```python
# At the top where all the `make_{onnx_op_name}` functions such as `make_reshape` are defined
def make_pow(self, name, inputs, dtype, shape):
    output = f"{name}/output_0"
    self.make_node("Pow", inputs=inputs, outputs=[output], name=name)
    self.make_value_info(output, dtype, shape=shape)

def make_relu_squared(self, layer_id, root_input, activation):
    relu_name = self.make_relu(layer_id, root_input, activation="Relu")
    pow_name = f"/model/layers.{layer_id}/mlp/act_fn/Pow"
    pow_inputs = [f"{relu_name}/output_0", "/model/constants/TensorProto.INT64/1D/2"]
    pow_shape = ['batch_size', 'sequence_length', self.intermediate_size]
    self.make_pow(pow_name, pow_inputs, self.io_dtype, pow_shape)
    return pow_name
```

src/python/py/models/builder.py
```diff
@@ -1735,6 +1834,25 @@ def make_gelu(self, layer_id, root_input, activation):
 
         return gelu_name
 
+    def make_relu(self, layer_id, root_input, activation):
+        relu_name = f"/model/layers.{layer_id}/mlp/act_fn/{activation}"
+        output = f"{relu_name}/output_0"
+        self.make_node(activation, inputs=[root_input], outputs=[output], name=relu_name, domain="")
+        self.make_value_info(output, self.io_dtype, shape=['batch_size', 'sequence_length', self.intermediate_size])
+        return relu_name
+
```
kunal-vaishnavi, line 1843: As mentioned [here](https://github.com/microsoft/onnxruntime-genai/pull/958/files#r1796153035), Constant ops are auto-created. I think we can simplify this to the following.

```python
def make_pow(self, name, inputs, dtype, shape):
    output = f"{name}/output_0"
    self.make_node("Pow", inputs=inputs, outputs=[output], name=name)
    self.make_value_info(output, dtype, shape=shape)

def make_relu_squared(self, layer_id, root_input, activation):
    relu_name = self.make_relu(layer_id, root_input, activation="Relu")
    pow_name = f"/model/layers.{layer_id}/mlp/act_fn/Pow"
    pow_inputs = [f"{relu_name}/output_0", "/model/constants/TensorProto.INT64/1D/2"]
    pow_shape = ['batch_size', 'sequence_length', self.intermediate_size]
    self.make_pow(pow_name, pow_inputs, self.io_dtype, pow_shape)
    return pow_name
```

src/python/py/models/builder.py
```diff
@@ -1839,7 +1961,14 @@ def make_model(self, input_path):
             elif module.__class__.__name__.endswith("DecoderLayer") and self.layer_id < self.num_layers:
                 # Each decoder layer of model
                 print(f"Reading decoder layer {self.layer_id}")
-                self.make_layer(self.layer_id, module)
+
+                # Some model stores rotary_emb at model class instead of decoder layer.
+                model_rotary_emb = None
+                if hasattr(model, "model"):
+                    if hasattr(model.model, "rotary_emb"):
+                        model_rotary_emb = model.model.rotary_emb
```
kunal-vaishnavi, line 1969: Let's define this in the `Nemotron` class instead of using `kwargs` in `make_layer`

src/python/py/models/builder.py
BLSharda: I see the attributes (num_heads and rotary_embedding_dim) of RotaryEmbedding op for partial rotary factor seems to be set correctly but the output of the **phi2** and **Nemotron-Mini** both are garbage without the workaround in this change.

Since, GenAI generated RotaryEmbedding op or the implementation (DML EP), doesn't seem to be handling the partial rotary factor correctly, can we submit this as a workaround to unblock for now? And this part can be removed once a proper fix is available.

src/python/py/models/builder.py
BLSharda: I didn't get you; the 'model' object is created & destroyed locally in make_model() function call. So, accessing 'model.model.rotary_emb' in Nemotron is not possible or I am missing something?

src/python/py/models/builder.py
BLSharda: I have updated code to use constant via names with data now, this looks really clean. Thanks!

src/python/py/models/builder.py
BLSharda: In latest change, I have moved rotary partial implementation workaround to member function make_rotary_embedding() in Nemotron class, to minimize changes/impact. So, phi2 model behavior will not change (will be in broken state until fix is available).

src/python/py/models/builder.py
```diff
@@ -1018,7 +1018,104 @@ def make_rotary_embedding_caches(self, rotemb, **kwargs):
...
+        rotary_reshape4_dim = [0, 0, dim_size]
+        constant4_name = f"{const_prefix}RotaryEmb/Reshape/Const4_input_dim_2"
+        constant4_value = numpy_helper.from_array(np.array(rotary_reshape4_dim, dtype="int64"))
+        self.make_node("Constant", inputs=[], outputs=[f"{constant4_name}\constant_0"], name=constant4_name, value=constant4_value)
+        self.make_node("Reshape", inputs=[concat_output, f"{constant4_name}\constant_0"], outputs=[reshape4_output_0], name=reshape4_name)
+        self.make_value_info(reshape4_output_0, self.io_dtype, shape=['batch_size', 'sequence_length', dim_size])
+
     def make_rotary_embedding(self, rotemb, name, root_input, **kwargs):
+        if self.rotemb_attrs["partial_rotary_factor"] > 0:
+            self.make_rotary_embedding_partial(rotemb, name, root_input, **kwargs)
```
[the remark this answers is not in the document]
BLSharda: Patrice has raised PR that fixes RotaryEmbedding for partial factor in ORT so the workaround is not needed anymore. I have now removed make_rotary_embedding_partial() altogether.

https://github.com/microsoft/onnxruntime/pull/22417

src/python/py/models/builder.py
BLSharda: Patrice has raised PR that fixes RotaryEmbedding for partial factor in ORT, so the workaround is not needed anymore. I have now removed make_rotary_embedding override for partial factor.

https://github.com/microsoft/onnxruntime/pull/22417

src/models/model.cpp
BLSharda: Marking this as resolved conversation for this PR and I hope that is okay.

src/python/py/models/builder.py
kunal-vaishnavi: In the `Nemotron` class, you can override the `make_attention` method to add the `rotary_emb` attribute (i.e. `attention.rotary_emb = model.model.rotary_emb`) and then call the parent method. Here's an example.

https://github.com/microsoft/onnxruntime-genai/blob/25636b434bfb93564be304b3218931a68fb6ac9f/src/python/py/models/builder.py#L2646-L2651

Then, the `kwargs` and `model_rotary_emb` changes can be reverted.

BLSharda: That makes sense, I have fixed the code now.

src/python/py/models/builder.py
```diff
@@ -1839,6 +1859,12 @@ def make_model(self, input_path):
             elif module.__class__.__name__.endswith("DecoderLayer") and self.layer_id < self.num_layers:
                 # Each decoder layer of model
                 print(f"Reading decoder layer {self.layer_id}")
+
+                # Some model stores rotary_emb at model class instead of attention objects.
+                if hasattr(model, "model"):
+                    if hasattr(model.model, "rotary_emb"):
```
kunal-vaishnavi, line 1865: After re-reading the [modeling code](https://github.com/huggingface/transformers/blob/3f06f95ebe617b192251ef756518690f5bc7ff76/src/transformers/models/nemotron/modeling_nemotron.py#L85), I think we can remove this altogether and remove the need to override the `make_attention` method in the Nemotron class. The cos/sin caches for the rotary embeddings in Nemotron are calculated in the same way as LLaMA.

https://github.com/microsoft/onnxruntime-genai/blob/921e0f47eb078c53ecd9d0dbd5c8e7e1930ebc9c/src/python/py/models/builder.py#L995-L1000

Since there isn't a `cos_cached` attribute, the cos/sin caches should get calculated correctly for the user.

src/python/py/models/builder.py
```diff
@@ -1744,12 +1760,16 @@ def make_activation(self, layer_id, root_input):
...
+            output_name = self.make_relu(layer_id, root_input, activation="Relu")
+        elif self.activation in {"relu2"}:
+            output_name = self.make_relu_squared(layer_id, root_input, activation="Relu2")
         else:
             raise NotImplementedError(f"The {self.activation} activation function is not currently supported.")
         return output_name
 
     def make_lm_head(self, lm_head):
-        bias_exists = lm_head.bias is not None
+        bias_exists = lm_head.bias is not None if hasattr(lm_head, 'bias') else False
```
kunal-vaishnavi, line 1772: This can be reverted. The LM head is [saved as](https://github.com/huggingface/transformers/blob/3f06f95ebe617b192251ef756518690f5bc7ff76/src/transformers/models/nemotron/modeling_nemotron.py#L989) `nn.Linear(config.hidden_size, config.vocab_size, bias=False)`. This means that `lm_head.bias = None` so `bias_exists = False` already.

Here's an example in a Python shell.

```python
>>> import torch
>>> a = torch.nn.Linear(2, 2, bias=False)
>>> a.weight
Parameter containing:
tensor([[ 0.3340, -0.0193],
        [ 0.0852,  0.0335]], requires_grad=True)
>>> a.bias
>>> a.bias is None
True
>>> hasattr(a, "bias")
True
```

BLSharda: Makes sense, I have reverted this part.

src/python/py/models/builder.py
BLSharda: Not passing a valid rotary embedding leads to crashes in make_attention(), it expects "attention.rotary_emb" to be valid object. So, overriding make_attention feels necessary to patch rotary_emb as suggested in review earlier.

`                   # Make RotaryEmbedding nodes
        cos_cache_name, sin_cache_name = "", ""
        if self.attention_attrs["use_rotemb_in_attn"]:
            cos_cache_name, sin_cache_name = self.make_rotary_embedding_caches(attention.rotary_emb)
        else:
            q_rotary_name = f"/model/layers.{layer_id}/attn/q_rotary/RotaryEmbedding"
            self.make_rotary_embedding(attention.rotary_emb, q_rotary_name, root_input=q_input_to_attention, position_ids=kwargs.get("position_ids", "position_ids"))
            q_input_to_attention = f"{q_rotary_name}/output_0"
            k_rotary_name = f"/model/layers.{layer_id}/attn/k_rotary/RotaryEmbedding"
            self.make_rotary_embedding(attention.rotary_emb, k_rotary_name, root_input=k_input_to_attention, position_ids=kwargs.get("position_ids", "position_ids"))
            k_input_to_attention = f"{k_rotary_name}/output_0"

`

kunal-vaishnavi: I think we can make a dummy rotary embedding attribute for `attention.rotary_emb` instead.

```python
    attention.rotary_emb = type("RotaryEmbedding", (object,), {'content':{}})()
    return super().make_attention(layer_id, attention, root_input, **kwargs)
```
[This PR](https://github.com/microsoft/onnxruntime-genai/pull/921) has an example of how this is done.

https://github.com/microsoft/onnxruntime-genai/blob/25636b434bfb93564be304b3218931a68fb6ac9f/src/python/py/models/builder.py#L2646-L2651

With this dummy attribute, the logic for `model_rotary_emb` can be removed and `attention.rotary_emb` is still defined.

kunal-vaishnavi:
Can you add the Nemotron model to the CIs?

https://github.com/microsoft/onnxruntime-genai/blob/0f59a900568cf65d74554cc34027648989f4ab1c/test/python/_test_utils.py#L55-L77

The models in `hf_paths` are downloaded from Hugging Face, and the models in `ci_paths` are currently uploaded to `/data/ortgenai_pytorch_models` in one of the CI VMs.

https://github.com/microsoft/onnxruntime-genai/blob/0f59a900568cf65d74554cc34027648989f4ab1c/.github/workflows/linux-gpu-x64-build.yml#L122

You can add it to `hf_paths` for now.

src/python/py/models/builder.py
BLSharda: Sure, that makes sense. Fixed in the latest code now.

kunal-vaishnavi (approved):

kunal-vaishnavi:
Can you merge main into your branch to see if that resolves the CI failures?

PR #959, opened 2024-10-07
Update issueLabeler.yml to fix label removal logic

sophies927:
Issue labeler currently adds back label if it is manually removed. These changes should fix this logic using the "labelRemoved" condition.


natke:
Looks good! Also we don't need the transformer label in this repo. Can we remove that one in this PR too please?

natke (approved):

PR #962, opened 2024-10-08
Rm unused ORT download GHA steps

skyline75489:


baijumeswani:
The build process downloads a version of ort (currently set at 1.19.2). During testing time, we should use a newer ort version. Ideally, the compile time ort and the test time ort should be different (similar to how we manage the python ort versions).

skyline75489:
Yeah sure. I'll close this then.

Issue #935, opened 2024-09-26
Build from source on Windows with default ONNX Runtime fails to run unit tests from visual studio


liqunfu:
**Describe the bug**
follow instructions at
https://onnxruntime.ai/docs/genai/howto/build-from-source.html
the build fails to run unit tests

**To Reproduce**
1.
git clone https://github.com/microsoft/onnxruntime-genai
cd onnxruntime-genai

2.
curl -L https://github.com/microsoft/onnxruntime/releases/download/v1.19.2/onnxruntime-win-x64-1.19.2.zip -o onnxruntime-win-x64-1.19.2.zip
tar xvf onnxruntime-win-x64-1.19.2.zip
move onnxruntime-win-x64-1.19.2 ort 

3.
python build.py --config Release

4. open Generator.sln with vs studio, set unit_tests as the startup program and run.
got "The given version [19] is not supported, only version 1 to 10 is supported in this build."

 
**Expected behavior**
The test shall run and pass

**Screenshots**

**Desktop (please complete the following information):**
 - OS: Windows

**Smartphone (please complete the following information):**

**Additional context**

baijumeswani:
@liqunfu could you try copying the onnxruntime dll to the same directory as the sln file or the unit_test.exe file?

liqunfu:
Thanks @baijumeswani, it works if I copy over onnxruntime.dll to the exe's folder. The issues is caused by having a onnxruntime.dll at "C:\Windows\WinSxS\wow64_microsoft-windows-ai-machinelearning_31bf3856ad364e35_10.0.22621.1_none_7056cab181ea1d5c\onnxruntime.dll". it must be a very old version. It blocks use and develop ort-genai on windows machines. This need to be fixed.

I see it is also confusing work with ort-genai's python api. It used to having a onnxruntime.dll at the genai install location. Now it also uses onnxruntime.dll at onnxrutime's install location. It is hard to know which one is being used.

baijumeswani:
Ideally, the python build script will try to add the onnxruntime dlls to the same path as the exe: https://github.com/microsoft/onnxruntime-genai/blob/53e3ac93c5e49e751308bcbe558c535555304349/build.py#L565-L567

But I think on windows, we need another <config> folder at the end. If you have bandwidth, would you consider submitting a PR for this? If not, I can get to it.

For python, we load onnxruntime from inside onnxruntime's python package. Two reasons for this:

- Minimize the onnxruntime-genai package size
[...]

PR #963, opened 2024-10-08
Fix windows test build issue

baijumeswani:
Addresses https://github.com/microsoft/onnxruntime-genai/issues/935

### DIFF
--- a/build.py
+++ b/build.py
@@ -602,6 +602,8 @@ def test(args: argparse.Namespace, env: dict[str, str]):
     Run the tests.
     """
     lib_dir = args.build_dir / "test"
+    if util.is_windows():
+        lib_dir = lib_dir / args.config
     if not args.ort_home:
         _ = util.download_dependencies(args.use_cuda, args.use_rocm, args.use_dml, lib_dir)
     else:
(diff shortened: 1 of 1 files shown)


RyanUnderhill (commented):
Should there be a comment, that this is needed because on Windows it's Debug\Debug vs just Debug?
Does this change only work on windows or would it work on the other platforms too? Just wondering if it can always work this way for all platforms vs just windows.

kunal-vaishnavi (approved):

RyanUnderhill (approved):

baijumeswani:
> Should there be a comment, that this is needed because on Windows it's Debug\Debug vs just Debug? Does this change only work on windows or would it work on the other platforms too? Just wondering if it can always work this way for all platforms vs just windows.

I added a comment.

On non-windows platforms, the unit_tests executable is inside the test folder inside the build directory (for example build/Linux/RelWithDebInfo/tests) whereas on Windows, the unit tests executable is inside the config directory inside the test directory (build/Windows/Debug/tests/Debug). With this PR, both scenarios will work.

PR #973, opened 2024-10-13
Move Cuda usage into a shared library similar to onnxruntime

RyanUnderhill:
This separates out the cuda code into a /src/cuda folder and builds onnxruntime_genai_cuda.dll/.so.
It is only this shared library that depends on the cuda runtimes.
There is an interface inside the core genai code to route functions to this cuda library, but my next step is to refactor the code to separate out provider specific, so this interface will shrink.

There is a new type to manage provider specific memory, DeviceMemory<T> and DeviceMemorySpan<T>, this will replace RoamingArray once the refactoring is complete.


aciddelgado (approved):

src/cuda/symbols.def
```diff
@@ -0,0 +1,2 @@
+EXPORTS
```
baijumeswani, line 1: Could we use `__declspec(dllexport)` instead similar to how we handle the ort-genai C API?

src/cuda/version_script.lds
```diff
@@ -0,0 +1,9 @@
+#_init and _fini should be local
+VERS_1.0 {
+  global:
+    GetInterface;
```
baijumeswani, line 4: Would we ever need to expose any other functions in the future? Are we able to deal with the visibility from inside the code using ```_attribute__((visibility("default")))```

src/cuda/symbols.def
RyanUnderhill: It won't work across all platforms, that's why I switched to using the separate file in cmake.

src/cuda/version_script.lds
RyanUnderhill: We need to update for all platforms each time. For the providers it should only ever be one function, since it returns an interface and we can change that interface arbitrarily. So this should be fine.

baijumeswani (approved):

PR #984, opened 2024-10-15
Integration nvidia modelopt quantization in GenAI builder

anujj:
[TensorRT-Model-Optimizer](https://github.com/NVIDIA/TensorRT-Model-Optimizer) have a implementation for INT4 AWQ. Adding the support in GenAI onnxruntime builder to quantized the models with TensorRT-Model-Optimizer

### DIFF
--- a/src/python/py/models/README.md
+++ b/src/python/py/models/README.md
@@ -219,11 +219,11 @@ python3 builder.py -i path_to_local_folder_on_disk -o path_to_output_folder -p p
 
 
 ### Quant Provider
-Provides different quantization techniques using option quant_provider
+Set different quantization techniques using option quant_provider
 
-  nvidia_awq: NVIDIA's TensorRT ModelOpt. It supports 2 calibration modes:
-
-  nvidia_awq_calibration: AWQ implementation and weight clipping. Choices: {awq, awq_clip}. awq is a default option for nvidia_awq
+TensorRT-Model-Optimizer's AWQ INT4 
+  1. nvidia_awq: NVIDIA's TensorRT ModelOpt awq implementation. set quant_provider=nvidia_awq. It supports 2 calibration modes.
+  2. nvidia_awq_calibration: AWQ implementation and weight clipping. Choices: {awq, awq_clip}. awq is a default option for nvidia_awq
 
 ```
 # From wheel:
--- a/src/python/py/models/README.md
+++ b/src/python/py/models/README.md
@@ -21,7 +21,6 @@ This folder contains the model builder for quickly creating optimized and quanti
     - [Hugging Face Authentication](#hugging-face-authentication)
     - [Use QDQ Pattern for Quantization](#use-qdq-pattern-for-quantization)
     - [Quant Provider](#quant-provider)
-    - [nvidia-awq calibration](#nvidia-awq-calibration)
   - [Unit Testing Models](#unit-testing-models)
     - [Option 1: Use the model builder directly](#option-1-use-the-model-builder-directly)
     - [Option 2: Edit the config.json file](#option-2-edit-the-configjson-file-on-disk-and-then-run-the-model-builder)
--- a/src/python/py/models/README.md
+++ b/src/python/py/models/README.md
@@ -220,25 +220,24 @@ python3 builder.py -i path_to_local_folder_on_disk -o path_to_output_folder -p p
 
 
 ### Quant Provider
-This option is to select the specific INT4 quantization algo to use. Option avalable are nvidia_awq(Nvidia modelopt awq implimentation) and default
+Provides different quantization techniques.
+
+ nvidia_awq: NVIDIA's TensorRT ModelOpt. It supports 2 calibration options:
+
+  nvidia_awq_calibration: AWQ implementation and weight clipping. Choices: {awq, awq_clip}. awq is a default option for nvidia_awq
 
 ```
 # From wheel:
 python3 -m onnxruntime_genai.models.builder -i path_to_local_folder_on_disk -o path_to_output_folder -p int4 -e execution_provider -c cache_dir_to_store_temp_files --extra_options quant_provider=nvidia_awq use_qdq=True
 # From source:
 python3 builder.py -i path_to_local_folder_on_disk -o path_to_output_folder -p precision -e execution_provider -c cache_dir_to_store_temp_files --extra_options quant_provider=nvidia_awq use_qdq=True
-```
 
-### nvidia-awq calibration
-nvidia_awq support two options, awq implementation and weight clipping. choices {awq, awq_clip}
-
-```
+# nvidia_awq_calibration option
 # From wheel:
 python3 -m onnxruntime_genai.models.builder -i path_to_local_folder_on_disk -o path_to_output_folder -p int4 -e execution_provider -c cache_dir_to_store_temp_files --extra_options quant_provider=nvidia_awq use_qdq=True nvidia_awq_calibration=awq
 # From source:
 python3 builder.py -i path_to_local_folder_on_disk -o path_to_output_folder -p precision -e execution_provider -c cache_dir_to_store_temp_files --extra_options quant_provider=nvidia_awq use_qdq=True nvidia_awq_calibration=awq
 ```
-
 ### Unit Testing Models
 
 This scenario is where your PyTorch model is already downloaded locally (either in the default Hugging Face cache directory or in a local folder on disk). If it is not already downloaded locally, here is an example of how you can download it.
--- a/src/python/py/models/README.md
+++ b/src/python/py/models/README.md
@@ -219,9 +219,9 @@ python3 builder.py -i path_to_local_folder_on_disk -o path_to_output_folder -p p
 
 
 ### Quant Provider
-Provides different quantization techniques.
+Provides different quantization techniques using option quant_provider
 
- nvidia_awq: NVIDIA's TensorRT ModelOpt. It supports 2 calibration options:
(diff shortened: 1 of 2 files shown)


anujj:
@kunal-vaishnavi : that's is a old PR for the same tool integration, i will get that PR closed

src/python/py/models/README.md
```diff
@@ -216,6 +218,27 @@ python3 -m onnxruntime_genai.models.builder -i path_to_local_folder_on_disk -o p
...
 
+
+### Quant Provider
+This option is to select the specific INT4 quantization algo to use. Option avalable are nvidia_awq(Nvidia modelopt awq implimentation) and default
+
+```
+# From wheel:
+python3 -m onnxruntime_genai.models.builder -i path_to_local_folder_on_disk -o path_to_output_folder -p int4 -e execution_provider -c cache_dir_to_store_temp_files --extra_options quant_provider=nvidia_awq use_qdq=True
+# From source:
+python3 builder.py -i path_to_local_folder_on_disk -o path_to_output_folder -p precision -e execution_provider -c cache_dir_to_store_temp_files --extra_options quant_provider=nvidia_awq use_qdq=True
```
kunal-vaishnavi, line 229: How is `use_qdq` used with `nvidia_awq`? I only see `use_qdq` used for the default `quant_provider` case.

src/python/py/models/README.md
```diff
@@ -216,6 +218,27 @@ python3 -m onnxruntime_genai.models.builder -i path_to_local_folder_on_disk -o p
...
+```
+
+### nvidia-awq calibration
```
kunal-vaishnavi, line 232: Rather than have a separate section for `nvidia_awq_calibration`, can we include it in the `quant_provider` section since both flags should be used at the same time? We can mention that both flags need to be provided when using `nvidia_awq`.

src/python/py/models/builder.py
```diff
@@ -45,6 +45,8 @@ def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
         self.filename = extra_options["filename"] if "filename" in extra_options else "model.onnx"
         self.hf_token = parse_hf_token(extra_options.get("hf_token", "true"))
         self.extra_options = extra_options
+        self.quant_provider = extra_options["quant_provider"] if "quant_provider" in extra_options else "default"
```
kunal-vaishnavi, line 48: Can we add `quant_provider` and `calibration_method` to `self.quant_attrs["int4"]` since these are INT4-specific settings?

https://github.com/microsoft/onnxruntime-genai/blob/7998f13e25f31caaba803db4197f737df99a0bb9/src/python/py/models/builder.py#L277-L283

src/python/py/models/builder.py
```diff
@@ -45,6 +45,8 @@ def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
...
+        self.nvidia_awq_calibration = extra_options["nvidia_awq_calibration"] if "nvidia_awq_calibration" in extra_options else "awq"
```
kunal-vaishnavi, line 49: Since other AWQ tools have a calibration option, can we generalize this name?

```suggestion
        self.calibration_method = extra_options["calibration_method"] if "calibration_method" in extra_options else "awq"
```

src/python/py/models/builder.py
```diff
@@ -413,16 +415,74 @@ def save_model(self, out_dir):
...
-        quant = MatMul4BitsQuantizer(
+        if (self.quant_provider == "nvidia_awq"):
+            # Nvidia ModelOpt only accepts an ONNX model path as input so process is
+            # hf --> fp16 model.onnx file (store at temp) --> pass fp16 model.onnx to modelopt
+            import tempfile
+            from onnxruntime.quantization.matmul_4bits_quantizer import NVAWQWeightOnlyQuantConfig
+            out_dir = os.path.join(tempfile.gettempdir(), "temp")
+            os.makedirs(out_dir, exist_ok=True)
+
+            out_path = os.path.join(out_dir, self.filename)
```
kunal-vaishnavi, line 426: Since saving the ONNX model has to be done twice with `nvidia_awq`, we should move the saving logic to its own method. For example, we can move

https://github.com/microsoft/onnxruntime-genai/blob/7998f13e25f31caaba803db4197f737df99a0bb9/src/python/py/models/builder.py#L395-L413

to something like this.

```python
def to_disk(self, model, out_dir):
    # Save ONNX model with only one external data file and delete any existing duplicate copies
    out_path = os.path.join(out_dir, self.filename)
    data_path = os.path.join(out_dir, os.path.basename(out_path) + ".data")
    
    if os.path.exists(out_path):
        print(f"Overwriting {out_path}")
        os.remove(out_path)
    if os.path.exists(data_path):
        print(f"Overwriting {data_path}")
        os.remove(data_path)

    save_model(
        model,
        out_path,
        save_as_external_data=True,
        all_tensors_to_one_file=True,
        location=os.path.basename(data_path),
        size_threshold=0,
        convert_attribute=False,
    )
```

Then, we can re-use this method here as well.

src/python/py/models/builder.py
```diff
@@ -413,16 +415,74 @@ def save_model(self, out_dir):
...
+                else:
+                    assert False, f"Unexpected calibration method: {self.nvidia_awq_calibration}"
+                print("nvidia_awq quantization works with block_size 128")
+                quant_config = NVAWQWeightOnlyQuantConfig(
+                    tokenizer_dir=self.model_name_or_path,
+                    calibration_method=calibration_method,
+                )
+
+                quant = MatMul4BitsQuantizer(
+                    model=out_path,
```
kunal-vaishnavi, line 460: We should simplify this method's logic since both paths will create a `MatMul4BitsQuantizer`, run `quant.process()`, and return `quant.model.model`.

src/python/py/models/builder.py
```diff
@@ -413,16 +415,74 @@ def save_model(self, out_dir):
...
+            finally:
+                if os.path.exists(out_path):
+                    print(f"Deleting temporary file {out_path}")
+                    os.remove(out_path)
+
+                if os.path.exists(data_path):
+                    print(f"Deleting temporary file {data_path}")
+                    os.remove(data_path)
+        else:
+            quant = MatMul4BitsQuantizer(
```
kunal-vaishnavi, line 476: The indentation appears to be incorrect here.

https://github.com/microsoft/onnxruntime-genai/blob/a60762276b0bb41e56261b37cd1c6fe827e596c0/src/python/py/models/builder.py#L475-L485

src/python/py/models/README.md
anujj: current implementation of the doesn't take this parameter as input but in internal implementation quant_format is of type qdq. More importantly use_qdq export the fp16 model with opt 21 which is required in this case. so i specified in the command line

src/python/py/models/README.md
anujj: This might cause problem if some other quant_provider get added in future and support different calibrations

src/python/py/models/builder.py
anujj: Done

src/python/py/models/builder.py
anujj: Done

src/python/py/models/builder.py
```diff
@@ -413,16 +415,74 @@ def save_model(self, out_dir):
...
+                if self.nvidia_awq_calibration == "awq":
+                    calibration_method = "awq_lite"
+                elif self.nvidia_awq_calibration == "awq_clip":
+                    calibration_method = "awq_clip"
+                else:
+                    assert False, f"Unexpected calibration method: {self.nvidia_awq_calibration}"
+                print("nvidia_awq quantization works with block_size 128")
+                quant_config = NVAWQWeightOnlyQuantConfig(
+                    tokenizer_dir=self.model_name_or_path,
+                    calibration_method=calibration_method,
```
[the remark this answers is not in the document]
anujj: Actually this is done internationally, the command line parameters are more aligned with the awq paper so that user can relate.

src/python/py/models/builder.py
anujj: Done

src/python/py/models/builder.py
anujj: This might cause problem if some other quant_provider get added future that supports some algo other then awq. In that case we might have different calibration methods

pranavsharma:
@kunal-vaishnavi any reason we are adding 3rd party quantization tool invocations in model builder?

kunal-vaishnavi:
> @kunal-vaishnavi any reason we are adding 3rd party quantization tool invocations in model builder?

There are multiple third-party invocations [already available](https://github.com/microsoft/onnxruntime/blob/bf604428aa4b9faf84ce99f735ad6402a6d6f886/onnxruntime/python/tools/quantization/matmul_4bits_quantizer.py#L1240) in ONNX Runtime's 4-bit quantization tool. This PR is making one of those invocations optionally available to users via the model builder.

pranavsharma:
> > @kunal-vaishnavi any reason we are adding 3rd party quantization tool invocations in model builder?
> 
> There are multiple third-party invocations [already available](https://github.com/microsoft/onnxruntime/blob/bf604428aa4b9faf84ce99f735ad6402a6d6f886/onnxruntime/python/tools/quantization/matmul_4bits_quantizer.py#L1240) in ONNX Runtime's 4-bit quantization tool. This PR is making one of those invocations optionally available to users via the model builder.

In a recent meeting we have decided to not add any more 3rd party quant tools here. They belong in Olive.

PR #970, opened 2024-10-11
Make Microsoft.ML.OnnxRuntimeGenAI.Tokenizer a Microsoft.ML.Tokenizers.Tokenizer

stephentoub:
This enables an ONNX Runtime GenAI tokenizer instance to be used anywhere a Microsoft.ML.Tokenizers tokenizer is accepted. If we'd prefer, rather than having Tokenizer be a base class for the ONNX Runtime one, we could instead expose some sort of `public Microsoft.ML.Tokenizer.Tokenizer AsTokenizer()` conversion method that returns a wrapper object (though that's a bit confusing given the names of the type are the same, just different namespaces).

cc: @luisquintanilla, @tarekgh

PR #987, opened 2024-10-16
Add an IChatClient implementation to OnnxRuntimeGenAI

stephentoub:


src/csharp/ChatClient.cs
```diff
@@ -0,0 +1,206 @@
...
+        StringBuilder prompt = new();
+
+        foreach (var message in messages)
+        {
+            foreach (var content in message.Contents)
+            {
+                switch (content)
+                {
+                    case TextContent tc when !string.IsNullOrWhiteSpace(tc.Text):
+                        prompt.Append("<|").Append(message.Role.Value).Append("|>\n").Append(tc.Text);
```
nmetulev, line 159: Should make this more general to support non phi models. Mistral and llama for example have a different prompt template

stephentoub: We can parameterize the client with the template to use. Is that information encoded anywhere in a programmatic way? Or developers with ONNX Runtime GenAI today are supposed to look up the format somewhere?

src/csharp/ChatClient.cs
```diff
@@ -0,0 +1,206 @@
...
+using System.Reflection;
+using System.Runtime.CompilerServices;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.ML.OnnxRuntimeGenAI;
+
+/// <summary>An <see cref="IChatClient"/> implementation based on ONNX Runtime GenAI.</summary>
+public sealed class ChatClient : IChatClient, IDisposable
```
azchohfi, line 13: Any reasons as to why this class would be sealed? Native AOT? It would be nice to be able to make the CreatePrompt method virtual, or generalizing it somehow.

stephentoub: > Any reasons as to why this class would be sealed?

I generally opt to until we have good reason to unseal it, especially in a case like this where interfaces are being implemented, as you then have to design the whole type for the fact that it might be extended.

If there are good reasons to unseal it and virtuals that would be warranted, then it can be.

src/csharp/ChatClient.cs
nmetulev: Yeah, unfortunately, there isn't a programmatic way. But I believe you are working on handling this with @azchohfi offline already

azchohfi (commented):
Then only issue I found is that when ChatOptions is null, it fails for Phi 3.5. This works fine for Phi3.
Phi 3.5 requires max_length to be provided. Seems to be a bug on Phi3.5, but I do not see a way around it, other than always providing max_length.

natke:
@stephentoub can you please merge main into your branch? This should enable the CIs to pass

stephentoub:
> @stephentoub can you please merge main into your branch? This should enable the CIs to pass

Thanks, but it's already based off of the latest commit in main. It seems like whatever feed is being used may not contain some of the necessary packages, like System.Text.Json?

pranavsharma:
Can you add a test?

azchohfi:
@pranavsharma Seems like the package needs to be added to the feed, from the upstream (nuget.org).

stephentoub:
> Can you add a test?

Yes

azchohfi:
Build is still failing due to the package not being available on the ORT-Nightly feed.
It needs to be manually added. @pranavsharma I do not have access to do that.

stephentoub:
@pranavsharma, I will update the PR with tests, but I first need your help in getting the feed updated with the necessary dependencies. Otherwise, nothing here even builds.

baijumeswani:
I can help with that. Is System.Text.Json the only package needed. Will try to have it resolved soon.

stephentoub:
> I can help with that. Is System.Text.Json the only package needed. Will try to have it resolved soon.

I'm not sure what's currently available in the feed, but at a minimum you'd need to add Microsoft.Extensions.AI.Abstractions. That itself has a System.Text.Json dependency downlevel.

stephentoub:
@baijumeswani, also https://github.com/microsoft/onnxruntime-genai/pull/970 needs Microsoft.ML.Tokenizers

baijumeswani:
Thanks for the info. A few of our engineering systems team members are out of office this week. And I am trying to find out how to upload nuget packages to our feed. It might take until next week to have this setup.

baijumeswani:
I have (hopefully) added all the dependencies. Please see the latest [windows cpu build](https://github.com/microsoft/onnxruntime-genai/actions/runs/12014174904/job/33569911133?pr=987).

luisquintanilla:
Hi folks, 

Just checking back in on this. Let us know how we can help. Thanks!

cc: @baijumeswani

azchohfi:
@baijumeswani The package version got updated, so we need the new packages from upstream saved into ort's feed again. Thanks in advance!

luisquintanilla:
@baijumeswani is this okay to merge?

azchohfi (approved):

baijumeswani (approved):

PR #992, opened 2024-10-17
initial webgpu support

guschmue:
minimal changes to run genai on top of webgpu.
There is no onnxruntime webgpu build yet so I can't turn on any tests that use webgpu or do a release build.
Will do that in a 2nd pr once the webgpu landed in onnxruntime,

### DIFF
--- a/src/models/model.cpp
+++ b/src/models/model.cpp
@@ -280,14 +280,23 @@ void Model::InitDeviceAllocator([[maybe_unused]] OrtSession& session) {
   if (device_type_ == DeviceType::CUDA) {
     allocator_device_ = GetCudaAllocator(session);
   }
-#elif USE_DML
+#endif
+#if USE_DML
   if (device_type_ == DeviceType::DML) {
     memory_info_device_ = OrtMemoryInfo::Create("DML", OrtAllocatorType::OrtDeviceAllocator, 0, OrtMemType::OrtMemTypeDefault);
     dml_owned_allocator_ = Ort::Allocator::Create(session, *memory_info_device_);
     allocator_device_ = dml_owned_allocator_.get();
   }
 #endif
-
+  allocator_kvcache_ = allocator_device_;
+#if USE_WEBGPU
+  if (device_type_ == DeviceType::WEBGPU) {
+    // for webgpu we only use device memory for kv_cache
+    memory_info_device_ = OrtMemoryInfo::Create("WebGPU_Buffer", OrtAllocatorType::OrtDeviceAllocator, 0, OrtMemType::OrtMemTypeDefault);
+    webgpu_owned_allocator_ = Ort::Allocator::Create(session, *memory_info_device_);
+    allocator_kvcache_ = webgpu_owned_allocator_.get();
+  }
+#endif
   session_info_ = std::make_unique<SessionInfo>(session);
   captured_graph_pool_ = std::make_shared<CapturedGraphPool>(config_.get(), session_info_.get(), allocator_device_);
 }
@@ -463,8 +472,11 @@ void Model::CreateSessionOptionsFromConfig(const Config::SessionOptions& config_
       for (auto& option : provider_options.options) {
         opts.emplace(option.first, option.second);
       }
-
       session_options.AppendExecutionProvider("QNN", opts);
+    } else if (provider_options.name == "web") {
+      device_type_ = DeviceType::WEBGPU;
+      std::unordered_map<std::string, std::string> opts;
+      session_options.AppendExecutionProvider("WebGPU", opts);
     } else
       throw std::runtime_error("Unknown provider type: " + provider_options.name);
   }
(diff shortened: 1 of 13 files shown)


guschmue:
gee, I tried to remove some switch cases and put them into the default path. gcc was happy with that but msvc is not,
hope I now reverted all of it.

src/models/model.h
```diff
@@ -175,9 +176,14 @@ struct Model : std::enable_shared_from_this<Model>, LeakChecked<Model> {
   std::unique_ptr<DmlReadbackHeap> dml_readback_heap_;
   ComPtr<IDMLDevice> dml_device_;
   std::unique_ptr<Ort::Allocator> dml_owned_allocator_;
+#endif
+#if USE_WEBGPU
+  std::unique_ptr<Ort::Allocator> webgpu_owned_allocator_;
```
pranavsharma, line 181: So, for every EP, we'll have separate variables for the allocator? @RyanUnderhill there should be a better way to do this.

pranavsharma (approved):

src/models/model.cpp
```diff
@@ -463,8 +472,11 @@ void Model::CreateSessionOptionsFromConfig(const Config::SessionOptions& config_
       for (auto& option : provider_options.options) {
         opts.emplace(option.first, option.second);
       }
-
       session_options.AppendExecutionProvider("QNN", opts);
+    } else if (provider_options.name == "web") {
```
skottmckay, line 476: nit: can we use "webgpu"? using 'web' as the provider in a mobile scenario would be somewhat confusing

guschmue: I was wondering too if we should use webgpu.
'web' is burned into model builder when you use '-e web' in model builder to create models for ort-web (fp16/int4 with MHA). It is stored in genai_config.json. But ort-web doesn't use genai so it does not look at that config file and we could change it.
But for most llm's there are models with that config file on huggingface for ort-web and we can just use them with genai if we look for 'web'.
Maybe look for 'web' and 'webgpu' and change model builder to insert 'webgpu' going forward ?

But there will be a new issue soon:
users would love to use the same model that are there for directml - working on that.
But their config file has 'directml' in it so we would need to be able to override this via api.

skottmckay: Sounds like 'web' is more a preset (selects attention type and data sizes) than an EP selector and perhaps isn't 1:1 with any EP info that gets baked into genai_config.json. And I assume once you add GQA support the model builder's notion of what 'web' is will change accordingly and now just be a data type selector.

Maybe the genai config should have a set of preferred EPs and their settings. User selects an EP from that list at runtime. If there's only one EP (current state) the selection is automatic.

guschmue: yes, something like "would work on these ep's"

kunal-vaishnavi: > But ort-web doesn't use genai so it does not look at that config file and we could change it.
> Maybe look for 'web' and 'webgpu' and change model builder to insert 'webgpu' going forward ?

Since the model builder uses the execution provider flag to decide more specific information, I think we should replace `web` with `webgpu` to be more precise. Since ONNX Runtime GenAI isn't currently used for ONNX Runtime Web, I think it is fine to update any published `genai_config.json` files so that they use `webgpu` instead of `web`.

kunal-vaishnavi: > But ort-web doesn't use genai so it does not look at that config file and we could change it.
> Maybe look for 'web' and 'webgpu' and change model builder to insert 'webgpu' going forward ?

Since the model builder uses the execution provider flag to decide more specific information, I think we should replace `web` with the execution provider's name used in ONNX Runtime to be more precise. The execution provider's options are set [here](https://github.com/microsoft/onnxruntime-genai/blob/dfbe14c39bc0486e1289332bca2003ff66a74fc7/src/python/py/models/builder.py#L69) based on the name passed. Since ONNX Runtime GenAI isn't currently used for ONNX Runtime Web, I think it is fine to update any published `genai_config.json` files so that they use a different name instead of `web`.

kunal-vaishnavi: > But ort-web doesn't use genai so it does not look at that config file and we could change it.
> Maybe look for 'web' and 'webgpu' and change model builder to insert 'webgpu' going forward ?

Since the model builder uses the execution provider flag to decide more specific information, I think we should replace `web` with the execution provider's name used in ONNX Runtime to be more precise. The execution provider's options are set [here](https://github.com/microsoft/onnxruntime-genai/blob/dfbe14c39bc0486e1289332bca2003ff66a74fc7/src/python/py/models/builder.py#L69) based on the name passed. Since ONNX Runtime GenAI isn't currently used for ONNX Runtime Web, I think it is fine to update any published `genai_config.json` files so that they use a different name instead of `web`.

> users would love to use the same model that are there for directml - working on that.
> But their config file has 'directml' in it so we would need to be able to override this via api.
> Maybe the genai config should have a set of preferred EPs and their settings. User selects an EP from that list at runtime. If there's only one EP (current state) the selection is automatic.

An ordered list of execution providers with their settings can be passed in the `genai_config.json` [here](https://github.com/microsoft/onnxruntime-genai/blob/dfbe14c39bc0486e1289332bca2003ff66a74fc7/src/python/py/models/builder.py#L308). ONNX Runtime GenAI will process the execution providers in the specified order and move to the next execution provider in the list if the previous one fails just like ONNX Runtime does.

guschmue:
I resolved the conversation - can address the general questions in a new PR

PR #993, opened 2024-10-18
Update Java Tensor

edgchen1:
- Save reference to ByteBuffer that owns the data so it won't get garbage collected while the Tensor is around.
- Verify the passed in ByteBuffer has native byte order.
- Change default nativeHandle value to 0.

### DIFF
--- a/src/java/src/main/java/ai/onnxruntime/genai/Tensor.java
+++ b/src/java/src/main/java/ai/onnxruntime/genai/Tensor.java
@@ -57,7 +57,7 @@ public final class Tensor implements AutoCloseable {
 
     // for now, require native byte order as the bytes will be used directly.
     if (data.order() != ByteOrder.nativeOrder()) {
-      throw new GenAIException("Tensor data must have native byte order.")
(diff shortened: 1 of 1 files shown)


src/java/src/main/java/ai/onnxruntime/genai/Tensor.java
```diff
@@ -5,12 +5,16 @@
 package ai.onnxruntime.genai;
 
 import java.nio.ByteBuffer;
+import java.nio.ByteOrder;
 
 public final class Tensor implements AutoCloseable {
-  private long nativeHandle = -1;
+  private long nativeHandle = 0;
```
edgchen1, line 11: not too sure about this change, but close() checks if it's 0.

skottmckay: I would expect `close()` is only called after successful construction of the object, and nativeHandle is set to the return value from createTensor if that occurs.

skottmckay (approved):

skottmckay (approved):

PR #997, opened 2024-10-18
allow setting runtime handles

fs-eire:
Sometimes we need to set runtime data (which cannot be written in a config file) into ORT's session options. This change introduces this RuntimeSettings type, together with internal implementation and C-API, C++-API.

### DIFF
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 #include "generators.h"
+#include "runtime_settings.h"
 #include "json.h"
 #include <fstream>
 #include <sstream>
@@ -659,7 +660,7 @@ void ParseConfig(const fs::path& filename, Config& config) {
   }
 }
 
-Config::Config(const fs::path& path) : config_path{path} {
+Config::Config(const fs::path& path, const RuntimeSettings* settings) : config_path{path} {
   ParseConfig(path / "genai_config.json", *this);
 
   if (model.context_length == 0)
@@ -667,6 +668,27 @@ Config::Config(const fs::path& path) : config_path{path} {
 
   if (search.max_length == 0)
     search.max_length = model.context_length;
+
+  if (settings) {
+    // Enable the following code after #992 is merged
+    // https://github.com/microsoft/onnxruntime-genai/pull/992
+
+    // #if USE_WEBGPU
+    auto& provider_options = model.decoder.session_options.provider_options;
+    auto maybe_webgpu_options = std::find_if(provider_options.begin(),
+                                             provider_options.end(),
+                                             [](Generators::Config::ProviderOptions& po) {
+                                               return po.name == "webgpu";
+                                             });
+    if (maybe_webgpu_options != provider_options.end()) {
+      auto it = settings->handles_.find("dawnProcTable");
+      if (it != settings->handles_.end()) {
+        void* dawn_proc_table_handle = it->second;
+        maybe_webgpu_options->options.emplace_back("dawnProcTable", std::to_string((size_t)(dawn_proc_table_handle)));
+      }
+    }
+    // #endif
+  }
 }
 
 void Config::AddMapping(const std::string& nominal_name, const std::string& graph_name) {
--- a/src/models/model.cpp
+++ b/src/models/model.cpp
@@ -496,8 +496,8 @@ std::shared_ptr<MultiModalProcessor> Model::CreateMultiModalProcessor() const {
   return std::make_shared<MultiModalProcessor>(*config_, *session_info_);
 }
 
-std::shared_ptr<Model> CreateModel(OrtEnv& ort_env, const char* config_path) {
-  auto config = std::make_unique<Config>(fs::path(config_path));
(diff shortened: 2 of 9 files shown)


src/ort_genai_c.h
```diff
@@ -50,6 +50,7 @@ typedef enum OgaElementType {
 typedef struct OgaResult OgaResult;
 typedef struct OgaGeneratorParams OgaGeneratorParams;
 typedef struct OgaGenerator OgaGenerator;
+typedef struct OgaRuntimeSettings OgaRuntimeSettings;
```
guschmue, line 53: add the struct for OgaRuntimeSettings ?

We eventually need a way of selecting the EP via API if we want to use the same model for say webgpu and dml. Guess we can use OgaRuntimeSettings  for that too.

guschmue:
Lgtm.

src/ort_genai_c.h
fs-eire: There will be a struct RuntimeSettings, but in C-API OgaRuntimeSettings is just an opaque type.

src/config.cpp
```diff
@@ -659,14 +660,35 @@ void ParseConfig(const fs::path& filename, Config& config) {
...
-Config::Config(const fs::path& path) : config_path{path} {
+Config::Config(const fs::path& path, const RuntimeSettings* settings) : config_path{path} {
   ParseConfig(path / "genai_config.json", *this);
 
   if (model.context_length == 0)
     throw std::runtime_error("model context_length is 0 or was not set. It must be greater than 0");
 
   if (search.max_length == 0)
     search.max_length = model.context_length;
+
```
RyanUnderhill, line 671: From how it looks, RuntimeSettings is really just extra settings that can overwrite the genai_config.json settings?
Why can't this be passed as a json string that gets parsed after the original file does? Then it should "just work" and any new options would be available in both automatically.

RyanUnderhill (changes requested):

src/config.cpp
fs-eire: The data "dawn proc table" that we want to use is a runtime data. It is a pointer to the Dawn's API table, which is quite similar to what `OrtApiBase::GetApi()` returns.

The actual use case is, user's application already loads the dawn DLL and can get the proc table. This memory address need to be passed to ORT so that ORT WebGPU EP can use the loaded instance of dawn.

RyanUnderhill: Right, I was meaning stuff that was passed at runtime, not written to the config. It looks like the table is just some strings (or am I misunderstanding the to_string part of the parsing?)

Can you show an example of what the data that is being passed in?

fs-eire: The actual data "dawn proc table" is an address to a set of API function pointers, pretty much similar to what `OrtApiBase::GetApi()` does. Any other Dawn API requires this to work.

Specifically, User's application need to call [`dawn::native::GetProcs()`](https://github.com/google/dawn/blob/main/src/dawn/native/DawnNative.cpp#L52-L54) to get the proc table, pass the pointer to onnxruntime which uses [`dawnProcSetProcs()`](https://github.com/google/dawn/blob/main/include/dawn/dawn_proc.h#L44) to set the value.

fs-eire: The `std::to_string` part is because there is no way in to pass EP specific non-string data to ORT using existing ORT APIs. `OrtAppendExecutionProvider()` supports only a string-to-string kvp as EP specific data.

RyanUnderhill: I see, it's just that the config structure should directly map to the json, and this breaks that. So trying to figure out if there's a way to work within this design vs adding exceptions.

src/config.cpp
```diff
@@ -659,14 +660,35 @@ void ParseConfig(const fs::path& filename, Config& config) {
   }
 }
 
-Config::Config(const fs::path& path) : config_path{path} {
+Config::Config(const fs::path& path, const RuntimeSettings* settings) : config_path{path} {
```
fs-eire, line 663: Config(const fs::path& path, std::string_view json_content)

src/models/model.cpp
```diff
@@ -496,8 +496,8 @@ std::shared_ptr<MultiModalProcessor> Model::CreateMultiModalProcessor() const {
   return std::make_shared<MultiModalProcessor>(*config_, *session_info_);
 }
 
-std::shared_ptr<Model> CreateModel(OrtEnv& ort_env, const char* config_path) {
-  auto config = std::make_unique<Config>(fs::path(config_path));
+std::shared_ptr<Model> CreateModel(OrtEnv& ort_env, const char* config_path, const RuntimeSettings* settings /*= nullptr*/) {
+  auto config = std::make_unique<Config>(fs::path(config_path), settings);
```
fs-eire, line 500: const overlay = settings ? BuildOverlay(settings) : "";
  auto config = std::make_unique<Config>(fs::path(config_path), overlay);

src/config.cpp
```diff
@@ -659,14 +660,35 @@ void ParseConfig(const fs::path& filename, Config& config) {
...
   ParseConfig(path / "genai_config.json", *this);
```
fs-eire, line 664: ParseConfig(overlay, *this);

fs-eire:
close this. replaced by #1002

PR #998, opened 2024-10-19
[C#] Expose Adapters API and add tests

yuslepukhin:


src/csharp/Adapters.cs
```diff
@@ -0,0 +1,63 @@
+﻿using Microsoft.ML.OnnxRuntimeGenAI;
```
pranavsharma, line 1: License headers?

src/csharp/Adapters.cs
```diff
@@ -0,0 +1,66 @@
...
+// Licensed under the MIT License.
+
+using Microsoft.ML.OnnxRuntimeGenAI;
+using System;
+using System.Runtime.InteropServices;
+
+/// <summary>
+/// A container of adapters.
+/// </summary>
+public class Adapters : SafeHandle
```
baijumeswani, line 11: When should one use SafeHandle over IDisposable?

src/csharp/Adapters.cs
```diff
@@ -0,0 +1,66 @@
...
+    /// used to load, unload and hold them.
+    /// Throws on error.
+    /// </summary>
+    /// <param name="model">Reference to a loaded model</param>
+    /// <returns>new Adapters object</returns>
+    public static Adapters Create(Model model)
+    {
+        Result.VerifySuccess(NativeMethods.OgaCreateAdapters(model.Handle, out IntPtr handle));
+        return new Adapters(handle);
+    }
```
baijumeswani, line 29: Should we use the constructor instead of a static `Create` method?

```suggestion
    /// <summary>
    /// Creates a container for adapters
    /// used to load, unload and hold them.
    /// Throws on error.
    /// </summary>
    /// <param name="model">Reference to a loaded model</param>
    /// <returns>new Adapters object</returns>
    public Adapters(Model model)
        {
            Result.VerifySuccess(NativeMethods.OgaCreateAdapters(model.Handle, out _handle));
        }
```

test/c_api_tests.cpp
```diff
@@ -4,6 +4,8 @@
 #include <models/model.h>
 #include <iostream>
 #include <ort_genai.h>
+#include "../src/span.h"
```
baijumeswani, line 7: The idea in this test file is that we only use the publicly available header file and not rely on the internal span functionality.
Could we avoid using this include?

test/python/test_onnxruntime_genai_api.py
```diff
@@ -432,6 +432,26 @@ def _prepare_adapter_model(test_data_path):
...
+
+        adapter_weight_b = np.zeros([vocab_size], dtype=(np.float32 if device == "cpu" else np.float16))
+        adapter_weight_b_tensor = onnx.helper.make_tensor(
+            "adapter_b",
+            onnx.TensorProto.FLOAT if device == "cpu" else onnx.TensorProto.FLOAT16,
+            [vocab_size],
+            adapter_weight_b.flatten()
+        )
+
+        model.graph.initializer.extend([adapter_weight_a_tensor, adapter_weight_b_tensor])
```
baijumeswani, line 453:
```suggestion
        for adapter_name in ["adapter_a", "adapter_b"]:
            adapter_weight = np.zeros([vocab_size], dtype=(np.float32 if device == "cpu" else np.float16))
            adapter_weight_tensor = onnx.helper.make_tensor(
                adapter_name,
                onnx.TensorProto.FLOAT if device == "cpu" else onnx.TensorProto.FLOAT16,
                [vocab_size],
                adapter_weight.flatten()
            )
            model.graph.initializer.append(adapter_weight_tensor)
```

src/csharp/Adapters.cs
yuslepukhin: It's complicated. The latest I heard it is OK to use it.

test/c_api_tests.cpp
yuslepukhin: This is a header only feature that substitutes for standard header for non-C++20 environments.
In C++20 it reduces to a standard header.
It creates no binary dependency, and therefore it is OK to use it.

src/csharp/Adapters.cs
yuslepukhin: One issue I am facing is that C# API is not consistent with C++. It tends to use constructors instead of static methods like on C++. I think this should be consistent.

baijumeswani: The reason for the C++ API needing the Create method is because it essentially does not hold any state and is used as a pass-through object with the same name as the underlying C opaque object.
C# does not have any such problems it needs to address. And so, I think it would be ok to make C# and C++ api inconsistent. Considering other API use constructors, I would be in favor of using the Adapters constructor instead of a dedicated static Create method.

baijumeswani (approved):

PR #1000, opened 2024-10-21
Fix some releasing pipeline issue

skyline75489:
Successful run: https://aiinfra.visualstudio.com/ONNX%20Runtime/_build/results?buildId=585119&view=results

### DIFF
--- a/.pipelines/codeql.yaml
+++ b/.pipelines/codeql.yaml
@@ -41,7 +41,6 @@ stages:
     - task: onebranch.pipeline.tsaoptions@1
       displayName: 'OneBranch TSAOptions'
       inputs:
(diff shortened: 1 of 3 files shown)


.pipelines/codeql.yaml
```diff
@@ -41,7 +41,6 @@ stages:
     - task: onebranch.pipeline.tsaoptions@1
       displayName: 'OneBranch TSAOptions'
       inputs:
-        tsaConfigFilePath: '$(Build.Repository.LocalPath)\.config\tsaoptions.json'
```
skyline75489, line 44: This is the default path, but on Linux the `/` is not recognized

snnn (approved):

PR #1002, opened 2024-10-22
allow setting runtime handles (alternative)

fs-eire:
This PR is an alternative way of https://github.com/microsoft/onnxruntime-genai/pull/997 to implement allowing users to set runtime handles.

The basic idea is to create a JSON string representing the `Config` class from the `RuntimeSettings`, and the `Config` class will merge them.

Currently this PR is not working because of the following:
- The existing JSON parser does not support multiple entries in `provider_options`
  The following JSON will fail to parse.
  ```json
  {
    "model": {
        "decoder": {
            "session_options": {
                "provider_options": [
                    {
                        "webgpu": { }
                    },
                    {
                        "dml": {}
                    }
                ]
            },
        },
     },
  }

  ```

- When trying to specify JSON overlay, two "webgpu" items does not merge.
  genai_config.json:
  ```json
  {
    "model": {
        "decoder": {
            "session_options": {
                "provider_options": [
                    {
                        "webgpu": { "abc": "123" }
                    }
                ]
            },
        },
     },
  }

  ```

  generated overlay config:
  ```
  {
    "model": {
      "decoder": {
        "session_options": {
          "provider_options": [
            {
              "webgpu": {
                "dawnProcTable": "12345678"
              }
            }
          ]
        }
      }
    }
  }
  ```

  Result:
  
  image

This PR depends on a code change to the Config class to support the expected parsing behaviors.

### DIFF
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -641,7 +641,7 @@ struct RootObject_Element : JSON::Element {
   JSON::Element& t_;
 };
 
-void ParseConfig(const fs::path& filename, Config& config) {
+void ParseConfig(const fs::path& filename, std::string_view json_overlay, Config& config) {
   std::ifstream file = filename.open(std::ios::binary | std::ios::ate);
   if (!file.is_open()) {
     throw std::runtime_error("Error opening " + filename.string());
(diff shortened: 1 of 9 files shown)


fs-eire:
After merged to latest main (takes #1003), it should work now.

guschmue (approved):
tested, works well.

src/config.cpp
```diff
@@ -664,8 +665,23 @@ void ParseConfig(const fs::path& filename, Config& config) {
   }
 }
 
-Config::Config(const fs::path& path) : config_path{path} {
+void ParseConfig(std::string_view json, Config& config) {
```
RyanUnderhill, line 668: I think you should add the JSON string as a second parameter to ParseConfig, then add this as a second block after it's try/catch. Also change the string slightly:
`Error encountered while parsing config overlay: JSON Error: (message) at line ## index ##"`
Since it will be confusing the way it is currently worded:
`Error encountered while parsing JSON JSON Error: (message) at line ## index ##"`

src/runtime_settings.cpp
```diff
@@ -0,0 +1,43 @@
...
+    }
+  }
+}
+)";
+
+  auto it = handles_.find("dawnProcTable");
+  if (it != handles_.end()) {
+    void* dawn_proc_table_handle = it->second;
+    std::string overlay;
+    overlay.reserve(webgpu_overlay_pre.size() + webgpu_overlay_post.size() + 20);
```
RyanUnderhill, line 33:
```suggestion
    overlay.reserve(webgpu_overlay_pre.size() + webgpu_overlay_post.size() + 20);  // Optional small optimization of buffer size
```

RyanUnderhill (approved):

PR #1022, opened 2024-10-31
Extend builder support for quantized lm_head

BowenBao:
As title. This enables support for further reduced quantized model size and improved runtime efficiency, within acceptable range of accuracy degradation.

Orthogonal to #940. This PR targets already quantized models in autoawq/autogptq format that **has** lmhead quantized.

### DIFF
--- a/src/python/py/models/quantized_model.py
+++ b/src/python/py/models/quantized_model.py
@@ -622,13 +622,12 @@ class AWQModel(QuantizedModel):
 
                     # Set `g_idx` to None since it's not used in `MatMulNBits`
                     q_tensors.g_idx = None
-        if isinstance(self.lm_head, QuantizedTensorModule):
-            if self.lm_head.qweight is not None:
-                self.unpack(self.lm_head)
-                self.repack(self.lm_head)
+        if isinstance(self.lm_head, QuantizedTensorModule) and self.lm_head.qweight is not None:
+            self.unpack(self.lm_head)
+            self.repack(self.lm_head)
 
-                # Set `g_idx` to None since it's not used in `MatMulNBits`
-                self.lm_head.g_idx = None
+            # Set `g_idx` to None since it's not used in `MatMulNBits`
+            self.lm_head.g_idx = None
 
     def unpack_qweight(self, module):
         """
--- a/src/python/py/models/quantized_model.py
+++ b/src/python/py/models/quantized_model.py
@@ -373,17 +373,7 @@ class QuantizedModel:
             self.lm_head.out_features = self.lm_head.scales.shape[1]
             self.lm_head.in_features = self.lm_head.qweight.shape[0]
             # Set g_idx if not already set
-            self.lm_head.g_idx = (
-                self.lm_head.g_idx
-                if self.lm_head.g_idx is not None
-                else torch.tensor(
-                    [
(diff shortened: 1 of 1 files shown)


BowenBao:
@kunal-vaishnavi ptal, thanks.

src/python/py/models/quantized_model.py
```diff
@@ -339,10 +351,39 @@ def __init__(self, quant_type, input_path, bits, group_size, q_size, kv_size, in
         # Set properties of each layer based on quantization type
         self.set_properties()
 
+    def _initialize_quantized_lm_head(self, bits, group_size):
+        """
+        Initialize `QuantizedTensorModule` for LM head if not already set
+        """
```
kunal-vaishnavi, line 357: Should we consider replacing `TensorModule` with `QuantizedTensorModule`, moving `self.weight` and `self.bias` to `QuantizedTensorModule`, and just use `QuantizedTensorModule`? Then, we wouldn't need to initialize and we can still check between quantized LM head vs. unquantized LM head with `self.lm.qweight is not None`.

src/python/py/models/quantized_model.py
```diff
@@ -581,6 +622,12 @@ def __init__(self, quant_type, input_path, bits, group_size, q_size, kv_size, in
 
                     # Set `g_idx` to None since it's not used in `MatMulNBits`
                     q_tensors.g_idx = None
+        if isinstance(self.lm_head, QuantizedTensorModule) and self.lm_head.qweight is not None:
```
kunal-vaishnavi, line 625:
```suggestion

        if isinstance(self.lm_head, QuantizedTensorModule) and self.lm_head.qweight is not None:
```

src/python/py/models/quantized_model.py
```diff
@@ -339,10 +351,39 @@ def __init__(self, quant_type, input_path, bits, group_size, q_size, kv_size, in
...
+                raise NotImplementedError(
+                    "lm_head quantization is only supported for awq."
+                )
+            self.lm_head.out_features = self.lm_head.scales.shape[1]
+            self.lm_head.in_features = self.lm_head.qweight.shape[0]
+            # Set g_idx if not already set
+            self.lm_head.g_idx = (
+                self.lm_head.g_idx
+                if self.lm_head.g_idx is not None
+                else torch.tensor(
```
kunal-vaishnavi, line 379: Can this be in one line?

src/python/py/models/quantized_model.py
```diff
@@ -339,10 +351,39 @@ def __init__(self, quant_type, input_path, bits, group_size, q_size, kv_size, in
...
+            self.lm_head = QuantizedTensorModule(bits, group_size)
+
     def set_properties(self):
         """
         Set in_features, out_features, and g_idx based on quantization type
         """
+        if isinstance(self.lm_head, QuantizedTensorModule):
+            if self.quant_type != "awq":
+                raise NotImplementedError(
+                    "lm_head quantization is only supported for awq."
```
kunal-vaishnavi, line 371: Can we set the properties for GPTQ-quantized LM head as well? It should be similar to setting the properties for AWQ-quantized LM head.

kunal-vaishnavi (approved):

PR #1023, opened 2024-10-31
constrained decoding fixes

mmoskal:
@Taka152

### DIFF
--- a/src/models/logits_processor.cpp
+++ b/src/models/logits_processor.cpp
@@ -11,6 +11,19 @@
 
 namespace Generators {
 
+std::vector<int32_t> tokenize_partial(const Tokenizer* tokenizer, const uint8_t* bytes,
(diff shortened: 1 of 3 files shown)


mmoskal:
So this makes the test pass (provided you also add the grammar files - they are not in the branch, seem to be .gitignored)

I think the `tokenize_partial()` method should become `Tokenizer::EncodePartial()` - it would be also a good place to cache the prefix_len. At some point, maybe it can migrate to OrtxTokenizer.

For non-llama-32k tokenizers, the approach of looking for `<0xFE>` kind of tokens will not work - the tokenizer will itself contain many partial UTF8 sequences. I think this applies to newer llama models, gemma etc.

I guess I could write a simple function that extracts the relevant info from HF tokenizer.json file - is it something you'll have always available?

PR #1030, opened 2024-11-04
Added Java Bindings for Adapters API

skyline75489:
* Added Java Bindings for Adapters API and test cases.
* Enable Java CI tests on win-x64, win-arm64, linux-x64 and osx-arm64.
* Fix existing Java test cases.
* Fix some style issues reported by spotless.

Internal work item: https://task.ms/aii/52363


src/models/onnxruntime_api.h
```diff
@@ -101,6 +101,8 @@ p_session_->Run(nullptr, input_names, inputs, std::size(inputs), output_names, o
 #define PATH_MAX (4096)
 #endif
 
+#if !defined(__ANDROID__)
```
skyline75489, line 104: This is to fix "redefine macros" warning

src/java/src/main/java/ai/onnxruntime/genai/Adapters.java
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  /**
+   * Load an adapter from the specified path.
+   *
+   * @param adapterFilePath The path of the adapter.
+   * @param adapterName A unique user supplied adapter identifier.
+   * @throws GenAIException If the call to the GenAI native API fails.
+   */
+  public void loadAdapters(String adapterFilePath, String adapterName) throws GenAIException {
```
skottmckay, line 32: Is there an API that allows passing in bytes for the adapter so we can read from the app's assets and not need to copy to the app's filesDir so it has a valid path?

src/java/src/main/java/ai/onnxruntime/genai/Adapters.java
skottmckay, line 1: Are there automated tests that exercise this code?

src/java/src/main/java/ai/onnxruntime/genai/Adapters.java
skyline75489: I'll have to ask @baijumeswani do we have that? I don't think so. Reading from bytes seems a reasonable option.

src/java/src/main/java/ai/onnxruntime/genai/Adapters.java
skyline75489: Right now, the adapter models are generated using Python script and will need to be copied to Android app's assets. I don't know if this is the "expected" process. It would require more changes that handles the generation outside Android environment and copy the forementioned adapter models to the phone(emulator). I was planning to add it in the future and not block this PR before 0.5.0.

skottmckay: I would have expected the non-Android Java tests should be simple to add new tests to: https://github.com/microsoft/onnxruntime-genai/tree/main/src/java/src/test/java/ai/onnxruntime/genai

For the Android based tests you should be able to leverage the existing logic that copies all the genai model files from the app assets to the files directory. Basically if you include the adapter file in that directory it will be copied and able to be accessed by path.

https://github.com/microsoft/onnxruntime-genai/blob/c9ffcb9d8a3fb645b0fdc6c9b71d645c74fe973a/src/java/src/test/android/app/src/androidTest/java/ai/onnxruntime_genai/example/javavalidator/SimpleTest.kt#L59

For reference, this is where the test model etc. is copied to the test app for inclusion in the assets.

https://github.com/microsoft/onnxruntime-genai/blob/c9ffcb9d8a3fb645b0fdc6c9b71d645c74fe973a/src/java/CMakeLists.txt#L185-L188

The current Android test app model was chosen based on it being used in other tests and having relatively simple input. I assume there is a test model that is used to validate the adapter usage elsewhere that the Android test app could be updated to use.

skyline75489: Thanks @skottmckay . Per today's meeting, we target this PR in the next release. I'll try to add the test and improve this PR.

skyline75489:
I just realize that the Java tests have not been actually run for a very long time and a lots of things are broken...this is gonna take a while.

build.py
```diff
@@ -229,11 +229,11 @@ def _validate_build_dir(args: argparse.Namespace):
             # also tweak build directory name for mac builds
             target_sys = "macOS"
 
-        args.build_dir = Path("build") / target_sys
+        args.build_dir = Path("build") / target_sys / args.config
```
skyline75489, line 232: This makes it easier for `build.py` reuse the same output dir as `cmake --preset`.

src/java/src/main/native/ai_onnxruntime_genai_Generator.cpp
```diff
@@ -57,9 +57,9 @@ Java_ai_onnxruntime_genai_Generator_getSequenceNative(JNIEnv* env, jobject thiz,
   // as there's no 'destroy' function in GenAI C API for the tokens we assume the OgaGenerator owns the memory.
   // copy the tokens so there's no potential for Java code to write to it (values should be treated as const)
   // or attempt to access the memory after the OgaGenerator is destroyed.
-  jintArray java_int_array = env->NewIntArray(num_tokens);
+  jintArray java_int_array = env->NewIntArray(static_cast<jsize>(num_tokens));
```
skyline75489, line 60: This is to pass the type checking on Windows where we treat warnings as errors.

src/java/src/test/java/ai/onnxruntime/genai/GenAITestExecutionListener.java
```diff
@@ -0,0 +1,10 @@
+package ai.onnxruntime.genai;
+
+import org.junit.platform.launcher.TestExecutionListener;
+import org.junit.platform.launcher.TestPlan;
+
+public class GenAITestExecutionListener implements TestExecutionListener {
+  public void testPlanExecutionFinished(TestPlan testPlan) {
+    GenAI.shutdown();
```
skyline75489, line 8: Global shutdown is needed. Otherwise leaded resources complain

build.py
```diff
@@ -453,7 +453,7 @@ def update(args: argparse.Namespace, env: dict[str, str]):
 
             is_x64_host = platform.machine() == "AMD64"
             if is_x64_host:
-                toolset_options += ["host=x64"]
+                pass
```
skyline75489, line 456: Again, this makes `build.py` align with `cmake --preset`.

src/java/src/main/java/ai/onnxruntime/genai/Tensor.java
```diff
@@ -62,11 +62,40 @@ public Tensor(ByteBuffer data, long[] shape, ElementType elementType) throws Gen
...
 
     nativeHandle = createTensor(data, shape, elementType.ordinal());
   }
 
+  /**
+   * Construct a Tensor from native handle.
+   *
+   * @param handle The native tensor handle.
+   */
+  public Tensor(long handle) {
```
skottmckay, line 75: How do you get a nativeHandle for a Tensor outside of this class? Naively I would have expected the handle to remain an internal implementation detail and at the Java level the user would only deal with the Tensor class.

src/java/src/main/native/ai_onnxruntime_genai_Generator.cpp
```diff
@@ -78,3 +78,21 @@ Java_ai_onnxruntime_genai_Generator_getSequenceLastToken(JNIEnv* env, jobject th
...
+                                        name));
+}
+
+JNIEXPORT jlong JNICALL
+Java_ai_onnxruntime_genai_Generator_getOutputNative(JNIEnv* env, jobject thiz, jlong native_handle,
+                                                    jstring output_name) {
+  OgaTensor* tensor = nullptr;
+  CString name{env, output_name};
+  ThrowIfError(env, OgaGenerator_GetOutput(reinterpret_cast<OgaGenerator*>(native_handle), name, &tensor));
+  return reinterpret_cast<jlong>(tensor);
```
skottmckay, line 97: IIRC you need to explicitly check the return type from ThrowIfError and return as it's doesn't quite behave as expected. The 'throw' is queued up but does not cause the function to return.

src/java/src/test/java/ai/onnxruntime/genai/GenerationTest.java
```diff
@@ -25,79 +25,132 @@ public class GenerationTest {
...
+
   @Test
   @EnabledIf("havePhi2")
   public void testUsageNoListener() throws GenAIException {
-    SimpleGenAI generator = new SimpleGenAI(phi2ModelPath());
-    GeneratorParams params = generator.createGeneratorParams("What's 6 times 7?");
-
-    String result = generator.generate(params, null);
-    logger.info("Result: " + result);
-    assertTrue(result.indexOf("Answer: 42") != -1);
```
skottmckay, line 54: Are we not able to validate that the answer makes sense?

src/java/src/test/java/ai/onnxruntime/genai/GenerationTest.java
```diff
@@ -25,79 +25,132 @@ public class GenerationTest {
...
+      String result = generator.generate(params, listener);
+
+      logger.info("Result: " + result);
+      assertNotNull(result);
+    }
+  }
+
+  @Test
+  @EnabledIf("haveAdapters")
+  public void testUsageWithAdapters() throws GenAIException {
```
skottmckay, line 75: Would be great if the test had before/after responses to demonstrate the adapters changing the output. Testing that you get a response with a shape is good, but it won't catch an issue where the response is invalid.

src/java/src/test/java/ai/onnxruntime/genai/GeneratorParamsTest.java
```diff
@@ -13,17 +13,19 @@ public class GeneratorParamsTest {
   @Test
   public void testValidSearchOption() throws GenAIException {
     // test setting an invalid search option throws a GenAIException
-    SimpleGenAI generator = new SimpleGenAI(TestUtils.testModelPath());
-    GeneratorParams params = generator.createGeneratorParams("Ignoed");
-    params.setSearchOption("early_stopping", true); // boolean
-    params.setSearchOption("max_length", 20); // number
+    try (SimpleGenAI generator = new SimpleGenAI(TestUtils.tinyGpt2ModelPath());
+        GeneratorParams params = generator.createGeneratorParams("Ignoed"); ) {
```
skottmckay, line 17:
```suggestion
        GeneratorParams params = generator.createGeneratorParams("Ignored"); ) {
```

src/java/src/test/java/ai/onnxruntime/genai/MultiModalProcessorTest.java
```diff
@@ -4,34 +4,28 @@
...
-                generator.generateNextToken();
-                 
-                int token = generator.getLastTokenInSequence(0);
-                 
-                fullAnswer += stream.decode(token);
-            }
-        }
+  @Test
+  public void testBatchEncodeDecode() throws GenAIException {
+    try (Model model = new Model(TestUtils.testVisioniModelPath());
```
skottmckay, line 18:
```suggestion
    try (Model model = new Model(TestUtils.testVisionModelPath());
```

src/java/src/test/java/ai/onnxruntime/genai/MultiModalProcessorTest.java
```diff
@@ -4,34 +4,28 @@
...
+        MultiModalProcessor multiModalProcessor = new MultiModalProcessor(model);
+        TokenizerStream stream = multiModalProcessor.createStream();
+        GeneratorParams generatorParams = model.createGeneratorParams()) {
+      String inputs =
+          new String(
+              "<|user|>\n<|image_1|>\n Can you convert the table to markdown format?\n<|end|>\n<|assistant|>\n");
+      try (Images image = new Images(TestUtils.getFilePathFromResource("/images/sheet.png"));
+          NamedTensors processed = multiModalProcessor.processImages(inputs, image); ) {
+        assertNotNull(processed);
```
skottmckay, line 27: Can we check the response is good instead of just not null?

src/java/src/test/java/ai/onnxruntime/genai/TestUtils.java
```diff
@@ -11,16 +11,20 @@
...
-    return f.getPath();
+  public static final String testAdapterTestAdaptersPath() {
+    return getFilePathFromResource("/adapters/adapters.onnx_adapter");
+  }
+
+  public static final String tinyGpt2ModelPath() {
+    return getFilePathFromResource("/hf-internal-testing/tiny-random-gpt2-fp32");
+  }
+
+  public static final String testVisioniModelPath() {
```
skottmckay, line 26:
```suggestion
  public static final String testVisionModelPath() {
```

src/java/src/test/resources/META-INF/services/org.junit.platform.launcher.TestExecutionListener
skottmckay, line 1: Does this need to be checked in?

src/java/src/test/java/ai/onnxruntime/genai/GenAITestExecutionListener.java
skottmckay: Does the GenAI documentation call this requirement out? Or alternatively is there a way to automate the call to shutdown?

skyline75489: Currently this is needed. A better solution from @RyanUnderhill is still WIP.

build.py
skyline75489: Yes. For our build environment, x64 is the default. I'm guessing this is from ORT when x86 is still a target. For GenAI we only ever target x64.

build.py
```diff
@@ -229,11 +229,11 @@ def _validate_build_dir(args: argparse.Namespace):
             # also tweak build directory name for mac builds
             target_sys = "macOS"
 
-        args.build_dir = Path("build") / target_sys
+        args.build_dir = Path("build") / target_sys / args.config
 
     # set to a config specific build dir. it should exist unless we're creating the cmake setup
     is_strict = not args.update
-    args.build_dir = args.build_dir.resolve(strict=is_strict) / args.config
+    args.build_dir = args.build_dir.resolve(strict=is_strict)
```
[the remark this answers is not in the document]
skyline75489: I'm leaning towards migrating presets to `build.py` and then removing presets entirely. But this still needs discussion. I'll add more comments.

src/java/src/test/java/ai/onnxruntime/genai/GenerationTest.java
skyline75489: Yes. Due to the phi-2 model being quite naive, the produced answer is garbage.

src/java/src/test/resources/META-INF/services/org.junit.platform.launcher.TestExecutionListener
skyline75489: Yes this is needed for the global shutdown listener.

.github/workflows/linux-cpu-x64-build.yml
```diff
@@ -27,6 +27,18 @@ jobs:
         with:
           dotnet-version: '8.0.x'
 
+      - name: Setup Java 17
+        uses: actions/setup-java@v4
+        with:
+          java-version: '17'
```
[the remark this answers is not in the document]
skyline75489: I think it should be fine, because we also have the separate CI for android, that should catch the features not available there.

src/java/src/main/java/ai/onnxruntime/genai/Tensor.java
skyline75489: You're right. This should be internal.

src/java/src/main/native/ai_onnxruntime_genai_Generator.cpp
skyline75489: Should I check the return value of `ThrowIfError` everywhere? There's a lot of usage in the existing code that do not check it. What's the downside for not checking it specifically? It'll still throw the exception, right? Will the return value become garbage?

src/java/src/test/java/ai/onnxruntime/genai/MultiModalProcessorTest.java
skyline75489: Currently no checking against the response is done in Python or C#. If we gonna add the check, we might add it in all languages. I have another PR adding validation against vision models. After that I think we can see if it's worth it to check the response here.

src/java/src/main/java/ai/onnxruntime/genai/SimpleGenAI.java
```diff
@@ -55,6 +55,27 @@ public GeneratorParams createGeneratorParams(String prompt) throws GenAIExceptio
...
+  /**
+   * Create the generator parameters and add a batch of prompt texts. The user can set other search
+   * options via the GeneratorParams object prior to running `generate`.
+   *
+   * @param prompts The batch of prompt texts to encode.
+   * @return The generator parameters.
+   * @throws GenAIException on failure
+   */
+  public GeneratorParams createGeneratorParams(String[] prompts) throws GenAIException {
+    GeneratorParams generatorParams = model.createGeneratorParams();
```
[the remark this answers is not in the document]
skyline75489: done

src/java/src/main/java/ai/onnxruntime/genai/SimpleGenAI.java
```diff
@@ -124,4 +146,14 @@ public String generate(GeneratorParams generatorParams, Consumer<String> listene
...
   }
+
+  @Override
+  public void close() {
+    if (tokenizer != null) {
+      tokenizer.close();
+    }
+    if (model != null) {
+      model.close();
+    }
```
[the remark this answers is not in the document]
skyline75489: yes for sure.

src/java/src/main/native/ai_onnxruntime_genai_Generator.cpp
skyline75489: I've added the check of return value. I think there's other places where the return value of `ThrowIfError` is not correctly checked. I'll try to fix them later.

src/java/src/test/java/ai/onnxruntime/genai/GenerationTest.java
skottmckay: Not quite following. Based on the code we can load and set the adapters and get the output. What interfaces are missing to be able to check the response is valid?

src/java/src/test/java/ai/onnxruntime/genai/TensorTest.java
```diff
@@ -15,20 +16,23 @@ public class TensorTest {
...
-    // no error on setting.
-    // assuming there's an error on execution if an invalid input has been provided so the user is
-    // aware of the issue
-    params.setInput("unknown_value", tensor);
+    try (SimpleGenAI generator = new SimpleGenAI(TestUtils.tinyGpt2ModelPath());
+        GeneratorParams params = generator.createGeneratorParams(); ) {
+      long[] shape = {2, 2};
+      Tensor.ElementType elementType = Tensor.ElementType.float32;
+      ByteBuffer data = ByteBuffer.allocateDirect(4 * Float.BYTES).order(ByteOrder.nativeOrder());
+      ;
```
skottmckay, line 24:
```suggestion
    try (SimpleGenAI generator = new SimpleGenAI(TestUtils.tinyGpt2ModelPath());
         GeneratorParams params = generator.createGeneratorParams(); ) {
      long[] shape = {2, 2};
      Tensor.ElementType elementType = Tensor.ElementType.float32;
      ByteBuffer data = ByteBuffer.allocateDirect(4 * Float.BYTES).order(ByteOrder.nativeOrder());
```

src/java/src/test/java/ai/onnxruntime/genai/TensorTest.java
```diff
@@ -15,20 +16,23 @@ public class TensorTest {
...
+
+      FloatBuffer floatBuffer = data.asFloatBuffer();
+      floatBuffer.put(new float[] {1.0f, 2.0f, 3.0f, 4.0f});
+      try (Tensor tensor = new Tensor(data, shape, elementType)) {
+        // no error on setting.
+        // assuming there's an error on execution if an invalid input has been provided so the user
+        // is
+        // aware of the issue
```
skottmckay, line 32:
```suggestion
        // is aware of the issue
```

src/java/src/test/java/ai/onnxruntime/genai/GenerationTest.java
skyline75489: `OrtValue` itself and related functions like `AllocatorFree` are not exposed by GenAI header, which means a lot of things in https://github.com/microsoft/onnxruntime/blob/main/java/src/main/native/ai_onnxruntime_OnnxTensor.c  cannot be done easily in the GenAI JNI wrapper code.

The current implementation of `GetData` is not complete. For example, the current `GetData` interface lacks the ability to fetch float16 type of data with proper size calculation. I was also bothered by this in #1084.

skyline75489: Currently, only limited subset of `GetTensorTypeAndShapeInfo()` is exposed by GenAI header like this:

https://github.com/microsoft/onnxruntime-genai/blob/c5745fd6d91a2ba2d94c6a52f6255652dde02508/src/ort_genai_c.cpp#L477

I believe at least I need to add `GetTensorTypeAndShapeInfo().GetDimensionsCount()` and `GetTensorTypeAndShapeInfo().GetElementCount()` to copy the logic here for proper size calculation:

https://github.com/microsoft/onnxruntime/blob/1128882bfd2a97c20f8a2a5ddb26cb0d42d9ebba/java/src/main/native/OrtJniUtil.c#L257

.gitignore
```diff
@@ -35,3 +35,5 @@ examples/csharp/HelloPhi/models
 /src/java/.gradle
 /src/java/local.properties
 /src/java/build
+/src/java/CMakeFiles
+/src/java/CMakeCache.txt
```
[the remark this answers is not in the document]
skyline75489: This is defined here:

https://github.com/microsoft/onnxruntime-genai/blob/c5745fd6d91a2ba2d94c6a52f6255652dde02508/src/java/CMakeLists.txt#L15

I *think* it's because `src/java/build` contains gradle build artifacts. This would help differentiate those two.

src/java/src/main/java/ai/onnxruntime/genai/Adapters.java
```diff
@@ -0,0 +1,81 @@
...
+      throw new RuntimeException("Failed to load onnxruntime-genai native libraries", e);
+    }
+  }
+
+  private native long createAdapters(long modelHandle) throws GenAIException;
+
+  private native void destroyAdapters(long nativeHandle);
+
+  private native void loadAdapter(long nativeHandle, String adapterFilePath, String adapterName)
+      throws GenAIException;
```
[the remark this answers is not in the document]
skyline75489: This actually violates spotlessJava rules. The current result is what spotlessJava prefers.

src/java/src/test/java/ai/onnxruntime/genai/GenerationTest.java
skyline75489: >Is there potentially a tokenizer or config issue?

Not that I know of. I've checked with @baijumeswani to confirm this is OK. In fact, test cases in other language bindings exhibits the same behaviour of producing garbage output.

src/java/src/main/native/ai_onnxruntime_genai_Tensor.cpp
```diff
@@ -39,3 +41,25 @@ JNIEXPORT void JNICALL
 Java_ai_onnxruntime_genai_Tensor_destroyTensor(JNIEnv* env, jobject thiz, jlong native_handle) {
   OgaDestroyTensor(reinterpret_cast<OgaTensor*>(native_handle));
 }
+
+JNIEXPORT jint JNICALL
+Java_ai_onnxruntime_genai_Tensor_getTensorType(JNIEnv* env, jobject thiz, jlong native_handle) {
+  OgaElementType type;
+  ThrowIfError(env, OgaTensorGetType(reinterpret_cast<OgaTensor*>(native_handle), &type));
+  return static_cast<int>(type);
```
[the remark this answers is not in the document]
skyline75489: done.

skottmckay (approved):

src/java/src/test/java/ai/onnxruntime/genai/GenerationTest.java
skyline75489: I'll resolve this for now and try to seek a better solution in the future.

PR #1036, opened 2024-11-05
Add Multi-Image Support with phi-3.5 in chat app demo

sayanshaw24:
Changes:
- Switch to Gradio File object from Image object to support multiple images
- Auto-populate image tags based on user input to avoid image tag mismatch: https://github.com/microsoft/onnxruntime-genai/blob/main/src/models/prompt_image_processor.cpp#L47
- Increase maximum tokens

Validation:
- [x] Tested demo with UI locally
- [x] Ran python unit test for multi-image support in GenAI here: https://github.com/microsoft/onnxruntime-genai/pull/796/files#diff-3d2f537fb637f846715772cb3ac1a9b32e1289bdcc3be8e36c868c55cc016e9dR284 within chat app

Sample:
image

### DIFF
--- a/examples/chat_app/app.py
+++ b/examples/chat_app/app.py
@@ -119,16 +119,16 @@ def launch_chat_app(expose_locally: bool = False, model_name: str = "", model_pa
                 )
                 max_length_tokens = gr.Slider(
                     minimum=0,
-                    maximum=4096,
-                    value=2048,
+                    maximum=8192,
+                    value=4096,
                     step=8,
                     interactive=True,
                     label="Max Token Length",
                 )
                 max_context_length_tokens = gr.Slider(
                     minimum=0,
-                    maximum=4096,
-                    value=2048,
+                    maximum=8192,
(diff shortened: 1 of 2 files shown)


examples/chat_app/app.py
```diff
@@ -119,16 +119,16 @@ def launch_chat_app(expose_locally: bool = False, model_name: str = "", model_pa
                 )
                 max_length_tokens = gr.Slider(
                     minimum=0,
-                    maximum=4096,
-                    value=2048,
+                    maximum=8192,
```
kunal-vaishnavi, line 122: The context length of the vision models is around [128K](https://huggingface.co/microsoft/Phi-3.5-vision-instruct/blob/4a0d683eba9f1d0cbfb6151705d1ee73c25a80ca/config.json#L31). Should we set the maximum value for `max_length_tokens` to that?

kunal-vaishnavi (approved):

PR #1038, opened 2024-11-06
Support constrained decoding

Taka152:
This PR adds constrained decoding feature that user can provide regex or json schema to guide the decoding process. This can enable function calls or further increase response quality.

The API looks like the following code, currently it only accepts regex and json_schema type, though context free grammar could be considered in the future.
```python
# python
params.set_guidance("json_schema", json.dumps(json_schema))
```
```cpp
// cpp
params->SetGuidance("json_schema", json_str);
```

The overhead brought by this feature is about 3-5% latency increase and could further optimized after continuous decoding feature jump forward steps with only one candidate.


.github/workflows/linux-cpu-arm64-build.yml
```diff
@@ -74,13 +74,19 @@ jobs:
         run: |
           docker run --rm \
           --volume $GITHUB_WORKSPACE:/onnxruntime_src \
-          -w /onnxruntime_src ort_genai_linux_arm64_gha bash -c "/usr/bin/cmake --preset linux_gcc_cpu_release"
+          -w /onnxruntime_src ort_genai_linux_arm64_gha bash -c " \
+            export CARGO_HOME=/onnxruntime_src/.cargo RUSTUP_HOME=/onnxruntime_src/.rustup && \
```
skyline75489, line 78: The installation and env setup should be moved to dockerfile like `tools/ci_build/github/linux/docker/inference/aarch64/default/cpu/Dockerfile`

.github/workflows/win-cpu-arm64-build.yml
```diff
@@ -52,6 +52,13 @@ jobs:
         move ${{ env.ORT_PACKAGE_NAME }}/build/native/include ort/
         move ${{ env.ORT_PACKAGE_NAME }}/runtimes/win-arm64/native/* ort/lib/
 
+    - name: Install Rust
+      run: |
+        $exePath = "$env:TEMP\rustup-init.exe"
```
skyline75489, line 57: There is no GHA action for this?

CMakeLists.txt
```diff
@@ -103,6 +110,14 @@ target_include_directories(onnxruntime-genai-static PRIVATE ${onnxruntime_extens
 target_include_directories(onnxruntime-genai-static PUBLIC ${onnxruntime_extensions_SOURCE_DIR}/shared/api/)
 target_link_libraries(onnxruntime-genai PRIVATE onnxruntime_extensions)
 target_link_libraries(onnxruntime-genai-static PUBLIC onnxruntime_extensions)
+target_include_directories(onnxruntime-genai PUBLIC ${llguidance_SOURCE_DIR}/parser/)
+target_include_directories(onnxruntime-genai-static PUBLIC ${llguidance_SOURCE_DIR}/parser/)
+target_link_libraries(onnxruntime-genai PRIVATE llguidance_parser)
+target_link_libraries(onnxruntime-genai-static PUBLIC llguidance_parser)
+if (WIN32)
+  target_link_libraries(onnxruntime-genai PRIVATE bcrypt)
```
skyline75489, line 118: Why this only needed on WIN32? Maybe add some comments.

src/models/logits_processor.h
```diff
@@ -0,0 +1,74 @@
+#pragma once
```
skyline75489, line 1: LICENSE header is needed.

build.py
```diff
@@ -130,6 +130,8 @@ class HelpFormatter(argparse.ArgumentDefaultsHelpFormatter, argparse.RawDescript
 
     parser.add_argument("--use_dml", action="store_true", help="Whether to use DML. Default is to not use DML.")
 
+    parser.add_argument("--disable_guidance", action="store_true", help="Whether to add guidance support. Default is True.")
```
skyline75489, line 133: Maybe just `use_guidance` and make the default to `true`. There's no `disable_sth` argument until now.

.github/workflows/win-cpu-arm64-build.yml
Taka152: Seems no official action for this, and some 3rd party actions only support linux and macos.

CMakeLists.txt
Taka152: rust std lib depends on it https://github.com/corrosion-rs/corrosion/issues/119. I'll add comment later

RyanUnderhill:
This feature requires the rust toolchain to be installed in order to build?

Taka152:
> This feature requires the rust toolchain to be installed in order to build?

Yes, the parser [llguidance ](https://github.com/microsoft/llguidance)is written in rust by MSR folks and used in [guidance](https://github.com/guidance-ai/guidance).

tools/ci_build/github/linux/docker/inference/aarch64/default/cpu/scripts/install_deps.sh
```diff
@@ -65,5 +65,12 @@ fi
 GetFile https://nodejs.org/dist/v18.17.1/node-v18.17.1-linux-${NODEJS_ARCH}.tar.gz /tmp/src/node-v18.17.1-linux-${NODEJS_ARCH}.tar.gz
 tar --strip 1 -xf /tmp/src/node-v18.17.1-linux-${NODEJS_ARCH}.tar.gz -C /usr
 
+# Install Rust
+export RUSTUP_HOME=/usr/.rustup 
+export CARGO_HOME=/usr/.cargo 
+curl --proto '=https' --tlsv1.2 https://sh.rustup.rs -sSf | sh -s -- -y
```
skyline75489, line 71: This doesn't enforce a specific version of Rust, does it? We might need to specify a version for the stability of the build.

.github/workflows/win-cuda-x64-build.yml
```diff
@@ -59,7 +59,14 @@ jobs:
       run: |
         mkdir ort/lib
         move ${{ env.ORT_PACKAGE_NAME }}/buildTransitive/native/include ort/
-        move ${{ env.ORT_PACKAGE_NAME }}/runtimes/win-x64/native/* ort/lib/    
+        move ${{ env.ORT_PACKAGE_NAME }}/runtimes/win-x64/native/* ort/lib/
+        
+    - name: Install Rust
+      run: |
+        $exePath = "$env:TEMP\rustup-init.exe"
+        (New-Object Net.WebClient).DownloadFile('https://static.rust-lang.org/rustup/dist/x86_64-pc-windows-msvc/rustup-init.exe', $exePath)
```
skyline75489, line 67: Same. Does this specify a Rust version?

.github/workflows/android-build.yml
```diff
@@ -82,6 +82,13 @@ jobs:
           unzip microsoft.ml.onnxruntime/${{ env.ORT_NIGHTLY_VERSION }}/runtimes/android/native/onnxruntime.aar -d ort
           ls -lR ort
 
+      - name: Install Rust
+        uses: dtolnay/rust-toolchain@stable
```
skyline75489, line 86: Can this Action specify Rust version?

src/models/logits.cpp
```diff
@@ -188,19 +236,41 @@ DeviceSpan<float> Logits::Get() {
...
 }
 
 #pragma warning(pop)
 
-void Logits::Update() {
+void Logits::Update(DeviceSpan<int32_t> next_tokens_unk) {
+  if (!logits_processors_.empty() && logits_processors_.at(0)) {
+    auto next_tokens = next_tokens_unk.CopyDeviceToCpu();
+    for (int i = 0; i < next_tokens.size(); i++) {
+      logits_processors_[i]->CommitTokens(static_cast<uint32_t>(next_tokens[i]));
```
skyline75489, line 261: `CommitTokens` can throw. Should the exception be handle here?

src/models/logits.cpp
```diff
@@ -188,19 +236,41 @@ DeviceSpan<float> Logits::Get() {
...
+    }
+    mask_future_ = std::async(std::launch::async, [&]() {
```
skyline75489, line 263: Should previous `future` be released/stopeed before assigning it a new `future`?

src/models/logits_processor.cpp
```diff
@@ -0,0 +1,144 @@
...
+  } else {
+    mask.reserve((vocab_size_ - 1) / 32 + 1);
+    for (int i = 0; i < (vocab_size_ - 1) / 32 + 1; i++) {
+      mask.push_back(mask_result.sample_mask[i]);
+    }
+  }
+  return mask;
+}
+
+void GuidanceLogitsProcessor::CommitTokens(uint32_t token) {
```
skyline75489, line 111: This should be `CommitToken` without the `s`.

src/models/logits.cpp
```diff
@@ -231,6 +301,27 @@ void Logits::HandleEOSArray(std::span<float> batched_logits) {
...
 
+void Logits::AddMask(std::span<float> logits, std::vector<std::vector<uint32_t>>& masks) {
+  size_t vocab_size = shape_[2];
+  size_t vocab_index = 0;
+
+  for (int index = 0; index < shape_[0]; index++) {
+    auto logits_span = logits.subspan(vocab_index, vocab_size);
+    auto& mask = masks[index];
+    for (size_t i = 0; i < vocab_size; i++) {
+      logits_span[i] = mask[i / 32] & (1 << (i % 32)) ? logits_span[i] : std::numeric_limits<float>::lowest();
```
skyline75489, line 312: Maybe explain the magic number 32.

src/models/logits_processor.cpp
```diff
@@ -0,0 +1,144 @@
...
+  std::ifstream json_file(json_path.string());
+  std::stringstream json_buffer;
+  json_buffer << json_file.rdbuf();
+  std::string json_data = json_buffer.str();
+  auto prefix_len = tokenizer_->Encode(kTokenizePrefixStr).size();
+  tokenize_data_ = {tokenizer_.get(), prefix_len};
+  LlgTokenizerInit tokenizer_init = {
+      static_cast<uint32_t>(vocab_size_),
+      eos_token,
+      nullptr,
```
skyline75489, line 55: Add comments to describe the arguments.

src/models/model.h
```diff
@@ -51,6 +51,8 @@ struct State {
 
   std::vector<const char*> input_names_, output_names_;
   std::vector<std::string> adapter_names_;
+  std::string guided_data_;
+  std::string guided_type_;
```
skyline75489, line 55: These doesn't seem to be used anywhere?

src/models/logits.cpp
Taka152: future.get() will wait future to finish. And it will also throw the exception in aysnc running. So I think it may not need to release here.

src/models/logits_processor.cpp
Taka152: Maybe we could add arg names, leave more details in llguidance.h?

src/models/model.h
Taka152: Oh, Mistake

src/models/logits_processor.cpp
skyline75489: Arg names should suffice, thanks.

src/models/logits.cpp
Taka152: I would like to let the exception block the whole execution for now. Because validated(done in the parser) grammar should work fine in one generation, leave it to throw exception should be good to optimize constrained decoding. And if this exception occurs often, we might consider to handle exception here and roll back to normal decoding in the future.

src/models/logits.cpp
skyline75489: good. thanks for the explaination.

src/ort_genai_c.cpp
```diff
@@ -7,6 +7,7 @@
 #include "span.h"
 #include "ort_genai_c.h"
 #include "generators.h"
+#include "logits_processor.h"
```
RyanUnderhill, line 10: Is this needed?

src/smartptrs.h
```diff
@@ -50,7 +50,7 @@ struct DeviceSpan {
   }
 
   // Copy device memory to CPU memory and return the CPU accessible memory
-  std::span<T> CopyDeviceToCpu() {
+  std::span<T> CopyDeviceToCpu() const {
```
RyanUnderhill, line 53: How is this const? p_device_memory_ is a non const std::shared_ptr and CopyDeviceToCpu is a non const method on it?

src/logits_processor.cpp
```diff
@@ -0,0 +1,240 @@
...
+  std::vector<std::vector<uint32_t>> masks;
+  for (int batch_idx = 0; batch_idx < batch_size_; batch_idx++) {  // renamed 'i' to 'batch_idx'
+    LlgMaskResult mask_result;
+    auto error = llg_compute_mask(llg_constraints_[batch_idx].get(), &mask_result);
+    if (error != 0) {
+      std::string error_message = llg_get_error(llg_constraints_[batch_idx].get());
+      throw std::runtime_error("Error computing mask: " + error_message);
+    }
+
+    std::vector<uint32_t> mask;
```
RyanUnderhill, line 123: Maybe put a comment here briefly describing how the two branches work?

src/logits_processor.cpp
```diff
@@ -0,0 +1,240 @@
...
+      }
+    }
+    masks.push_back(mask);
+  }
+  return masks;
+}
+
+void GuidanceLogitsProcessor::CommitTokens(std::span<int32_t> tokens) {
+  for (int i = 0; i < batch_size_; i++) {
+    LlgCommitResult commit_result;
```
RyanUnderhill, line 141: What is commit_result used for, if llg_commit_token returns an error?

src/logits_processor.cpp
```diff
@@ -0,0 +1,240 @@
...
+      cudaMemcpyAsync(cuda_logits_mask_ptr_.Span().data() + (i * vocab_size_ / 32), masks.at(i).data(),
+                      static_cast<int>(masks.at(i).size() * sizeof(uint32_t)), ::cudaMemcpyHostToDevice, cuda_stream_);
+    }
+    cuda::LaunchAddLogitsMask(logits.Span().data(), batch_size_, vocab_size_, cuda_logits_mask_ptr_.Span().data(), cuda_stream_);
+    return;
+  }
+#endif
+  size_t vocab_index = 0;
+
+  auto logits_span = logits.Span();
```
RyanUnderhill, line 176: Should this be CpuSpan()? If the device is dml or webgpu for example?

src/logits_processor.cpp
```diff
@@ -0,0 +1,240 @@
...
+namespace Generators {
+
+#if USE_GUIDANCE
+GuidanceLogitsProcessor::GuidanceLogitsProcessor(const State& state)
+    : vocab_size_(state.params_->config.model.vocab_size),
+      eos_token_(state.params_->config.model.eos_token_id),
+      batch_size_(state.params_->search.batch_size),
+      device_type_(state.params_->device_type) {
+  guidance_type_ = state.params_->guidance_type;
+  guidance_data_ = state.params_->guidance_data;
```
RyanUnderhill, line 32: Why are these not in the initializer list?

src/ort_genai_c.cpp
Taka152: Yeah, Cause it now use same forward declaration as search, model. We need to include their full definition to build with unique_ptr<LogitsProcessor> in generators.h

src/logits_processor.cpp
Taka152: commit_result is useless for now. Jump forward feature will return multiple tokens that is fixed in next steps, so we can call append_tokens(commit_result.tokens) to accelerate constrianed decoding. I will add a new PR to support this.

RyanUnderhill (dismissed):

.pipelines/stages/jobs/steps/capi-win-step.yml
```diff
@@ -46,6 +46,13 @@ steps:
   condition: eq(variables['ep'], 'cuda')
   workingDirectory: '$(Build.Repository.LocalPath)'
 
+- powershell: |
+    $exePath = "$env:TEMP\rustup-init.exe"
+    (New-Object Net.WebClient).DownloadFile('https://static.rust-lang.org/rustup/dist/x86_64-pc-windows-msvc/rustup-init.exe', $exePath)
```
skyline75489, line 51: This assumes `x86_64`. Windows on ARM will not work.

Taka152: I'll revert all the azure pipeline code since it might need more discussion to decide if we want to add this to release

natke (changes requested):
We will postpone merging this until we resolve the rust dependency

Issue #1029, opened 2024-11-03
How to select NPU instead of GPU on Intel laptop?


mmaitre314:
Where could I find a code sample showing how to run genai on NPU instead of GPU using `DmlExecutionProvider` in Python?

Conda environment:
```yaml
name: phi3-onnx-dml
channels:
  - conda-forge
  - nodefaults
dependencies:
  - python=3.12
  - vs2015_runtime=14.40.33810
  - pip
  - pip:
    - ipykernel==6.29.5
    - numpy==1.26.4
    - huggingface-hub[cli]==0.26.2
    - onnxruntime-directml==1.20.0
    - onnxruntime-genai-directml==0.4.0
```

PR #1041, opened 2024-11-06
Added functionality to choose dml adapter by luid

DavitGrigoryan132:
**Description**: This PR adds functionality to allow users to specify a "DML" (DirectML) execution provider device adapter by providing the adapter's LUID (Locally Unique Identifier). Previously, the system would automatically select the most performant adapter by default. With this enhancement, users can now select a specific adapter by passing the luid_high_part and luid_low_part options, giving them greater control over device selection.

**Key Changes**:

Added support for selecting a DML execution provider device adapter using luid_high_part and luid_low_part.
Default behavior remains the same, choosing the most performant adapter unless these options are specified.

**Motivation**: This update provides more flexibility for users needing to target specific adapters, enhancing customizability in environments with multiple available adapters.

#1029

### DIFF
--- a/src/dml/dml_helpers.cpp
+++ b/src/dml/dml_helpers.cpp
@@ -75,8 +75,7 @@ static std::vector<ComPtr<IDXGIAdapter1>> EnumerateAdapters(PLUID device_luid = 
           adapter_infos.emplace_back(std::move(adapter));
           break;
         }
-      }
-      else if (d3d12_device) {
+      } else if (d3d12_device) {
         adapter_infos.emplace_back(std::move(adapter));
       }
     }
--- a/src/dml/dml_helpers.cpp
+++ b/src/dml/dml_helpers.cpp
@@ -21,14 +21,14 @@ static bool IsSoftwareAdapter(IDXGIAdapter1* adapter) {
   return desc.Flags & DXGI_ADAPTER_FLAG_SOFTWARE || (is_basic_render_driver_vendor_id && is_basic_render_driver_device_id);
 };
 
-static std::vector<ComPtr<IDXGIAdapter1>> EnumerateAdapters() {
+static std::vector<ComPtr<IDXGIAdapter1>> EnumerateAdapters(PLUID device_luid = nullptr) {
   ComPtr<IDXGIFactory4> dxgi_factory;
   THROW_IF_FAILED(CreateDXGIFactory(IID_PPV_ARGS(&dxgi_factory)));
 
   std::vector<ComPtr<IDXGIAdapter1>> adapter_infos;
 
   ComPtr<IDXGIFactory6> dxgi_factory6;
-  if (SUCCEEDED(dxgi_factory.As(&dxgi_factory6))) {
+  if (SUCCEEDED(dxgi_factory.As(&dxgi_factory6)) && !device_luid) {
     // Enumerate adapters by performance. This only works in Windows 10 Version 1803 and later.
     ComPtr<IDXGIAdapter1> adapter;
     for (uint32_t adapter_index = 0;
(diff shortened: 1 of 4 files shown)


natke:
Thank you for this contribution. Can you please add a test/tests for the new functionality?

pranavsharma:
This should be reviewed by the DML team.

src/dml/dml_helpers.cpp
```diff
@@ -52,6 +52,31 @@ static std::vector<ComPtr<IDXGIAdapter1>> EnumerateAdapters() {
...
+    for (uint32_t adapter_index = 0; dxgi_factory->EnumAdapters1(adapter_index, &adapter) != DXGI_ERROR_NOT_FOUND; adapter_index++) {
+      // We can't assume the ordering of hardware and software adapters, so keep looping. This path should only execute on Windows 10
+      // version 1709 or earlier; IDD (e.g. remote desktop) adapters do not exist when taking this code path.
+      if (IsSoftwareAdapter(adapter.Get())) {
+        continue;
+      }
+
+      // Make sure that we are able to create the device
+      ComPtr<ID3D12Device> d3d12_device;
+      THROW_IF_FAILED(D3D12CreateDevice(adapter.Get(), D3D_FEATURE_LEVEL_11_0, IID_PPV_ARGS(&d3d12_device)));
```
PatriceVignola, line 67: This enumeration code is the same as the code below. I think It's better to have a single `else` block, and to only have an `if (device_luid)` when it's time to actually push it to the `adapter_infos` vector.

DavitGrigoryan132: Thanks, I  updated it

src/models/model.cpp
```diff
@@ -437,7 +437,24 @@ void Model::CreateSessionOptionsFromConfig(const Config::SessionOptions& config_
...
+        bool contains_device_luid = false;
+        LUID device_luid{};
+        for (const auto& [name, value] : provider_options.options) {
+          if (name == "luid_high_part") {
+            device_luid.HighPart = std::stol(value);
+            contains_device_luid = true;
+          } else if (name == "luid_low_part") {
+            device_luid.LowPart = std::stol(value);
+            contains_device_luid = true;
+          }
```
PatriceVignola, line 450: What happens if the user only specifies the low or high part of the luid? Alternatively, can't we simply parse the entire LUID from a single string?

DavitGrigoryan132: If the user only specifies the low or high part of the luid, the program will crush. Yes, I can change it and specify luid by a single string, where high part and low part will be separated by ":" symbol

src/models/model.cpp
```diff
@@ -497,8 +514,8 @@ void Model::CreateSessionOptionsFromConfig(const Config::SessionOptions& config_
 
   // If no device gets set, default to CPU
   if (!p_device_) {
-    assert(device_type_ == DeviceType::CPU);
-    p_device_ = GetDeviceInterface(device_type_);
+    assert(device_type_ == DeviceType::CPU || device_type_ == DeviceType::DML || device_type_ == DeviceType::WEBGPU);
```
DavitGrigoryan132, line 517: This line throws an error when `device_type_` is set to `DML`. I believe both `DML` and `WEBGPU` should use the `CPU` `DeviceInterface` to avoid this issue.

src/models/model.cpp
DavitGrigoryan132: Updated

src/models/model.cpp
DavitGrigoryan132: Already fixed in main branch

DavitGrigoryan132:
@PatriceVignola Can you review it, please?

PatriceVignola (approved):

baijumeswani (approved):

baijumeswani:
Thanks for your contribution @DavitGrigoryan132 :)

PR #1042, opened 2024-11-06
Add example for set terminate

ajindal1:
Add an example for the set terminate feature using the Phi3 model


examples/c/src/phi3_terminate.cpp
```diff
@@ -0,0 +1,121 @@
...
+      tokenizer->Encode(prompt.c_str(), *sequences);
+
+      std::cout << "Generating response..." << std::endl;
+      auto params = OgaGeneratorParams::Create(*model);
+      params->SetSearchOption("max_length", 1024);
+      params->SetInputSequences(*sequences);
+
+      auto generator = OgaGenerator::Create(*model, *params);
+
+      std::vector<std::thread> threads;
```
RyanUnderhill, line 76: Are there only just two threads?

examples/c/src/phi3_terminate.cpp
```diff
@@ -0,0 +1,121 @@
...
+
+      threads.push_back(std::thread(Generate_Output_CXX, generator.get(), std::move(tokenizer_stream)));
+
+      if (runTerminate){
+        threads.push_back(std::thread(Generator_SetTerminate_Call, generator.get()));
```
RyanUnderhill, line 81: Should this be a separate thread or just do this on the main thread to cause the background thread to terminate?

examples/c/src/phi3_terminate.cpp
ajindal1: Yes, in this example there are only two threads.

examples/c/src/phi3_terminate.cpp
ajindal1: I think doing on the main thread would also work fine, separate thread is not necessary.

baijumeswani:
Instead of creating a new example, could we re-use the current C++ example and modify it so that it can catch a Cntrl+C keyboard signal, and then terminate the generation, before moving on to the next user input?

ajindal1:
> Instead of creating a new example, could we re-use the current C++ example and modify it so that it can catch a Cntrl+C keyboard signal, and then terminate the generation, before moving on to the next user input? 

Done, thanks.

documents/Runtime_option.md
```diff
@@ -4,12 +4,14 @@ This file will provide details on the usage of SetRuntimeOption API. It will lis
 
 ## Set Terminate
 
-Set Terminate is a runtime option to terminate the current session or continue/restart an already terminated session. There are two valid ways to call Set Terminate.
+Set Terminate is a runtime option to terminate the current session or continue/restart an already terminated session. The current session will crash when the terminate option is enabled and the user will need to handle that scenario, examples/c/src/phi3_terminate.cpp contains an example for this.
```
baijumeswani, line 7:
```suggestion
Set Terminate is a runtime option to terminate the current session or continue/restart an already terminated session. The current session will throw an exception when the terminate option is enabled and the user will need to handle that scenario, examples/c/src/phi3_terminate.cpp contains an example for this.
```

documents/Runtime_option.md
```diff
@@ -4,12 +4,14 @@ This file will provide details on the usage of SetRuntimeOption API. It will lis
...
 
-To enable terminate, the valid pair is: ("set_terminate", "1")
+There are two valid ways to call Set Terminate.
 
-To disable terminate, the valid pair is: ("set_terminate", "0")
+To enable terminate, the valid pair is: ("terminate_session", "1")
 
-Key: "set_terminate"
+To disable terminate, the valid pair is: ("terminate_session", "0")
```
baijumeswani, line 13:
```suggestion
To terminate the generation, use this key value pair: ("terminate_session", "1") 

To recover from a terminated state, use this key value pair: ("terminate_session", "0") 
```

examples/c/src/phi3.cpp
```diff
@@ -97,31 +142,22 @@ void CXX_API(const char* model_path) {
 
     auto generator = OgaGenerator::Create(*model, *params);
 
-    while (!generator->IsDone()) {
-      generator->ComputeLogits();
-      generator->GenerateNextToken();
+    std::thread th(Generate_Output_CXX, generator.get(), std::move(tokenizer_stream), is_first_token, std::ref(timing));
```
baijumeswani, line 145: Can you reverse this? Make the main thread do the generation, while the newly created thread checks for the interrupt? That will reduce the changes in this example and users who do not care about termination, can still look at the main function code.

examples/c/src/phi3.cpp
```diff
@@ -65,16 +68,58 @@ class Timing {
...
+}
+
+std::atomic<bool> stopFlag(false);
+
+void signalHandler(int signum) {
+    std::cout << "Interrupt signal received. Terminating current session...\n";
+    stopFlag = true;
+}
+
 void CXX_API(const char* model_path) {
```
baijumeswani, line 112: Can you also do the same with the C API?

examples/c/src/phi3.cpp
ajindal1: Thanks, this is a good idea.

examples/c/src/phi3.cpp
ajindal1: Added, thanks.

examples/c/src/phi3.cpp
```diff
@@ -96,46 +121,69 @@ void CXX_API(const char* model_path) {
...
 // C API Example
 
+void Generator_SetTerminate_Call_C(OgaGenerator* generator) {
+  while (!OgaGenerator_IsDone(generator)) {
+    if (stopFlag) {
+      OgaGenerator_SetRuntimeOption(generator, "terminate_session", "1");
+      stopFlag = false;
+      break;
+    }
+    std::this_thread::sleep_for(std::chrono::milliseconds(100)); // Check every 100 ms
```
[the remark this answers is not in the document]
ajindal1: I was also thinking about it while implementing, thanks for mentioning this. I have added a signaling mechanism for catching this faster.

natke (approved):
Let's get a dev review too

examples/c/src/phi3.cpp
```diff
@@ -7,6 +7,13 @@
 #include <iostream>
 #include <string>
 #include "ort_genai.h"
+#include <thread>
+#include <csignal>
+#include <atomic>
+#include <setjmp.h>
```
baijumeswani, line 13:
```suggestion
```

baijumeswani (approved):

baijumeswani (approved):

PR #1043, opened 2024-11-07
Add Quantized_model + float LoRA model scenario to model builder

apsonawane:
Add Quantized_model + float LoRA model scenario to model builder

### DIFF
--- a/src/python/py/models/builder.py
+++ b/src/python/py/models/builder.py
@@ -437,7 +437,7 @@ class Model:
         # Quantize ONNX model to desired precision
         # TODO: Replace by quantizing the MatMuls as they are created
         already_quantized_in_qdq_format = self.quant_type is not None and self.quant_attrs["use_qdq"]  # Skip quantizing `MatMul` in `DequantizeLinear --> Transpose --> MatMul` path
-        if self.onnx_dtype == "int4" and not already_quantized_in_qdq_format:
+        if self.onnx_dtype == "int4" and not already_quantized_in_qdq_format and not self.matmul_attrs["use_lora"]:
             model = self.to_int4(model)
 
         # Save ONNX model with only one external data file and delete any existing duplicate copies
@@ -714,7 +714,7 @@ class Model:
         self.make_value_info(output, dtype, shape=shape)
 
     def make_matmul(self, matmul, basename, root_input, **kwargs):
-        if hasattr(matmul, "base_layer"):
+        if hasattr(matmul, "lora_A"):
             # For LoRA `MatMul`
             return self.make_matmul_lora(matmul, basename, root_input, **kwargs)
         else:
@@ -849,18 +849,35 @@ class Model:
         basename_parts = basename.split("/")
 
         # Make LoRA MatMul path
-        matmul_A_basename = "/".join(basename_parts[:-1] + ["lora_A"] + basename_parts[-1:])
-        matmul_A_name = self.make_matmul_op(matmul.lora_A.default, matmul_A_basename, root_input=root_input)
-        lora_A = f"{matmul_A_name}/output_0"
+        if hasattr(matmul.lora_A, "default"):
+            matmul_A_basename = "/".join(basename_parts[:-1] + ["lora_A"] + basename_parts[-1:])
+            matmul_A_name = self.make_matmul_op(matmul.lora_A.default, matmul_A_basename, root_input=root_input)
+            lora_A = f"{matmul_A_name}/output_0"
 
-        matmul.lora_B.default.weight *= matmul.scaling["default"]
-        matmul_B_basename = "/".join(basename_parts[:-1] + ["lora_B"] + basename_parts[-1:])
-        matmul_B_name = self.make_matmul_op(matmul.lora_B.default, matmul_B_basename, root_input=lora_A)
-        lora_B = f"{matmul_B_name}/output_0"
+            matmul.lora_B.default.weight *= matmul.scaling["default"]
+            matmul_B_basename = "/".join(basename_parts[:-1] + ["lora_B"] + basename_parts[-1:])
+            matmul_B_name = self.make_matmul_op(matmul.lora_B.default, matmul_B_basename, root_input=lora_A)
+            lora_B = f"{matmul_B_name}/output_0"
+        else:
+            matmul_A_basename = "/".join(basename_parts[:-1] + ["lora_A"] + basename_parts[-1:])
+            matmul_A_name = self.make_matmul_op(matmul.lora_A, matmul_A_basename, root_input=root_input)
+            lora_A = f"{matmul_A_name}/output_0"
+
+            # matmul.lora_B.weight *= matmul.scaling["default"]
+            matmul_B_basename = "/".join(basename_parts[:-1] + ["lora_B"] + basename_parts[-1:])
+            matmul_B_name = self.make_matmul_op(matmul.lora_B, matmul_B_basename, root_input=lora_A)
+            lora_B = f"{matmul_B_name}/output_0"    
 
         # Make regular MatMul path
-        last_dim = matmul.base_layer.weight.shape[0]
-        matmul_name = self.make_matmul_op(matmul.base_layer, basename, root_input, **kwargs)
+        if hasattr(matmul, "base_layer"):
+            last_dim = matmul.base_layer.weight.shape[0]
+            matmul_name = self.make_matmul_op(matmul.base_layer, basename, root_input, **kwargs)
+        elif hasattr(matmul, "qweight"):
+            last_dim = matmul.qweight.shape[0]
+            matmul_name = self.make_matmul_op(matmul, basename, root_input, **kwargs)
+        else:
+            last_dim = matmul.weight.shape[0]
+            matmul_name = self.make_matmul_op(matmul, basename, root_input, **kwargs)
 
         # Make LoRA Add node
         add_name = "/".join(basename_parts[:-1] + ["lora", "Add"])
@@ -2036,13 +2053,14 @@ class Model:
                 kv_size,
                 self.intermediate_size,
                 self.num_layers,
+                self.extra_options["adapter_path"] if "adapter_path" in self.extra_options else None,
             )
         else:
             # Load PyTorch model
             extra_kwargs = {"num_hidden_layers": self.num_layers} if "num_hidden_layers" in self.extra_options else {}
             model = AutoModelForCausalLM.from_pretrained(self.model_name_or_path, cache_dir=self.cache_dir, token=self.hf_token, trust_remote_code=True, **extra_kwargs)
 
-        if "adapter_path" in self.extra_options:
+        if "adapter_path" in self.extra_options and self.quant_type is None:
             from peft import PeftModel
             model = PeftModel.from_pretrained(model, self.extra_options["adapter_path"], cache_dir=self.cache_dir, token=self.hf_token)
 
--- a/src/python/py/models/quantized_model.py
+++ b/src/python/py/models/quantized_model.py
@@ -46,6 +46,11 @@ class QuantizedTensorModule:
 
         return qweight + qzeros + scales + g_idx + in_feats + out_feats + bits + group_size
 
+class QuantizedLoraModule(QuantizedTensorModule):
+    def __init__(self, bits, group_size):
+        super().__init__(bits, group_size)
+        self.lora_A = TensorModule()
+        self.lora_B = TensorModule()
 
 class TensorModule:
     def __init__(self):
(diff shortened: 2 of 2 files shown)


src/python/py/models/builder.py
```diff
@@ -853,14 +853,26 @@ def make_matmul_lora(self, matmul, basename, root_input, **kwargs):
         matmul_A_name = self.make_matmul_op(matmul.lora_A.default, matmul_A_basename, root_input=root_input)
         lora_A = f"{matmul_A_name}/output_0"
 
-        matmul.lora_B.default.weight *= matmul.scaling["default"]
+        from peft import PeftConfig
+        peft_config = PeftConfig.from_pretrained(extra_options["adapter_path"], trust_remote_code=True)
```
kunal-vaishnavi, line 857: Can we create a `lora` section in `self.matmul_attrs` and add the attributes that we need there? The default values can be 0 when LoRA isn't used.

```python
self.matmul_attrs = {
    "use_lora": is_lora,
    "lora": {
        "lora_alpha": config.lora_alpha if is_lora else 0,
        "r": config.r if is_lora else 0
    }
}
```

kunal-vaishnavi, line 857: Can we create a `lora` section in `self.matmul_attrs` and add the attributes that we need there? The default values can be 0 when LoRA isn't used.

```python
self.matmul_attrs = {
    "use_lora": is_lora,
    "lora": {
        "alpha": config.lora_alpha if is_lora else 0,
        "r": config.r if is_lora else 0
    }
}
```

src/python/py/models/builder.py
```diff
@@ -2036,13 +2048,14 @@ def make_model(self, input_path):
                 kv_size,
                 self.intermediate_size,
                 self.num_layers,
+                self.extra_options["adapter_path"] if "adapter_path" in self.extra_options else None,
```
kunal-vaishnavi, line 2051: Can we put all of the parameters passed to `QuantModel.from_pretrained` back in one line?

src/python/py/models/quantized_model.py
```diff
@@ -52,51 +52,86 @@ def __init__(self):
...
+class QuantizedLoRAModule(QuantizedTensorModule):
     def __init__(self, bits, group_size):
-        self.q_proj = QuantizedTensorModule(bits, group_size)
-        self.k_proj = QuantizedTensorModule(bits, group_size)
-        self.v_proj = QuantizedTensorModule(bits, group_size)
-        self.o_proj = QuantizedTensorModule(bits, group_size)
+        super().__init__(bits, group_size)
+        self.lora_A = TensorModule()
+        self.lora_B = TensorModule()
+        self.scaling = {"default": 1}
```
kunal-vaishnavi, line 65: If `scaling` will be calculated from the config since it cannot be determined from the `.safetensors` files, can this be removed?

src/python/py/models/quantized_model.py
```diff
@@ -334,23 +371,37 @@ def __init__(self, quant_type, input_path, bits, group_size, q_size, kv_size, in
...
+                            module.mlp.up_proj.lora_A.weight = tensor
+                        elif bool(re.match(r"^model.layers\.\d+\.mlp.up_proj.lora_B\.weight$", name)):
+                            module.mlp.up_proj.lora_B.weight = tensor
+                        elif bool(re.match(r"^model.layers\.\d+\.mlp.gate_proj.lora_A\.weight$", name)):
+                            module.mlp.gate_proj.lora_A.weight = tensor
+                        elif bool(re.match(r"^model.layers\.\d+\.mlp.gate_proj.lora_B\.weight$", name)):
+                            module.mlp.gate_proj.lora_B.weight = tensor
+                        elif bool(re.match(r"^model.layers\.\d+\.self_attn.k_proj.lora_A\.weight$", name)):
+                            module.self_attn.k_proj.lora_A.weight = tensor
+                        elif bool(re.match(r"^model.layers\.\d+\.self_attn.k_proj.lora_B\.weight$", name)):
```
kunal-vaishnavi, line 388: Can we re-order the cases to the following based on appearance in the ONNX model?

1. `q_proj`
2. `k_proj`
3. `v_proj`
4. `o_proj`
5. `gate_proj`
6. `up_proj`
7. `down_proj`

Can we also add a comment in each case that specifies what the pattern matching is attempting to find (for example, `model.layers.layer_id.mlp.down_proj.lora_A.weight` for the first one)?

src/python/py/models/quantized_model.py
```diff
@@ -729,16 +841,16 @@ def __init__(self, module):
 
 class QuantModel:
     @staticmethod
-    def from_pretrained(quant_type, input_path, bits, group_size, use_g_idx, q_size, kv_size, intermediate_size, num_layers):
+    def from_pretrained(quant_type, input_path, bits, group_size, use_g_idx, q_size, kv_size, intermediate_size, num_layers, adapter_path=None):
```
kunal-vaishnavi, line 844: `None` can be removed since you already pass `None` in for when `adapter_path` isn't used.

```suggestion
    def from_pretrained(quant_type, input_path, bits, group_size, use_g_idx, q_size, kv_size, intermediate_size, num_layers, adapter_path):
```

src/python/py/models/builder.py
kunal-vaishnavi, line 2051:
```suggestion
                self.extra_options.get("adapter_path", None),
```

src/python/py/models/builder.py
```diff
@@ -2036,13 +2048,14 @@ def make_model(self, input_path):
...
             )
         else:
             # Load PyTorch model
             extra_kwargs = {"num_hidden_layers": self.num_layers} if "num_hidden_layers" in self.extra_options else {}
             model = AutoModelForCausalLM.from_pretrained(self.model_name_or_path, cache_dir=self.cache_dir, token=self.hf_token, trust_remote_code=True, **extra_kwargs)
 
-        if "adapter_path" in self.extra_options:
+        if "adapter_path" in self.extra_options and self.quant_type is None:
             from peft import PeftModel
```
kunal-vaishnavi, line 2059: Can we add a comment specifying which case this is (since this is not part of the above `if-elif-else` block)?

src/python/py/models/quantized_model.py
```diff
@@ -52,51 +52,86 @@ def __init__(self):
...
-    def __init__(self, quant_type, input_path, bits, group_size, q_size, kv_size, intermediate_size, num_layers):
+    def __init__(self, quant_type, input_path, bits, group_size, q_size, kv_size, intermediate_size, num_layers, adapter_path):
         self.quant_type = quant_type
         self.embedding = TensorModule()
         self.final_norm = TensorModule()
         self.lm_head = TensorModule()
         self.layers = {}
         self.num_layers = num_layers
 
+        is_lora = adapter_path is not None
```
kunal-vaishnavi, line 107:
```suggestion
        self.is_lora = adapter_path is not None
```

src/python/py/models/quantized_model.py
```diff
@@ -361,7 +412,7 @@ def _initialize_quantized_lm_head(self, bits, group_size):
         if not isinstance(self.lm_head, QuantizedTensorModule):
             self.lm_head = QuantizedTensorModule(bits, group_size)
 
-    def set_properties(self):
+    def set_properties(self, is_lora):
```
kunal-vaishnavi, line 415:
```suggestion
    def set_properties(self):
```

src/python/py/models/quantized_model.py
```diff
@@ -394,6 +446,37 @@ def set_properties(self):
                 module.mlp.down_proj.out_features = module.mlp.down_proj.scales.shape[1]
                 module.mlp.down_proj.in_features = module.mlp.down_proj.qweight.shape[0]
 
+                if is_lora:
```
kunal-vaishnavi, line 449:
```suggestion
                if self.is_lora:
```

src/python/py/models/quantized_model.py
```diff
@@ -52,51 +52,86 @@ def __init__(self):
...
         self.final_norm = TensorModule()
         self.lm_head = TensorModule()
         self.layers = {}
         self.num_layers = num_layers
 
+        is_lora = adapter_path is not None
+
+        self.map_to_modules(quant_type, input_path, bits, group_size, q_size, kv_size, intermediate_size, adapter_path)
+
+        if adapter_path is not None:
```
kunal-vaishnavi, line 111:
```suggestion
        if self.is_lora:
```

src/python/py/models/quantized_model.py
```diff
@@ -52,51 +52,86 @@ def __init__(self):
...
+            self.map_to_modules(quant_type, adapter_path, bits, group_size, q_size, kv_size, intermediate_size, adapter_path)
```
kunal-vaishnavi, line 112: Why does `adapter_path` have to be passed twice?

src/python/py/models/quantized_model.py
```diff
@@ -52,51 +52,86 @@ def __init__(self):
...
+        self.layers = list(self.layers.values())
+        self.layers.sort(key=lambda m: m.layer_id)
+
+        # Set properties of each layer based on quantization type
+        self.set_properties(is_lora)
+
+    def map_to_modules(self, quant_type, file_path, bits, group_size, q_size, kv_size, intermediate_size, adapter_path):
         layer_id = 0
-        for weight_file in os.listdir(input_path):
+        is_lora = adapter_path is not None
```
kunal-vaishnavi, line 130: If `self.is_lora` is used, then this can be removed and any references to `is_lora` can be replaced with `self.is_lora`.

src/python/py/models/quantized_model.py
apsonawane: We are checking for is_lora in map_to_modules function thats why it was added. But now since we have self.is_lora we do not need that

src/python/py/models/quantized_model.py
```diff
@@ -52,51 +52,86 @@ def __init__(self):
...
         self.quant_type = quant_type
         self.embedding = TensorModule()
         self.final_norm = TensorModule()
         self.lm_head = TensorModule()
         self.layers = {}
         self.num_layers = num_layers
 
+        is_lora = adapter_path is not None
+
+        self.map_to_modules(quant_type, input_path, bits, group_size, q_size, kv_size, intermediate_size, adapter_path)
```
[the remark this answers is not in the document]
apsonawane: I did not understand this comment. What is not needed? the is_lora condition or the map_to_modules call?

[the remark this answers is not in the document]
kunal-vaishnavi: Is `adapter_path` needed in the first call to `self.map_to_modules` since there is a second call to `self.map_to_modules` for `adapter_path`? Can this look instead like the following?

```python
self.map_to_modules(quant_type, input_path, bits, group_size, q_size, kv_size, intermediate_size)
if self.is_lora:
    self.map_to_modules(quant_type, adapter_path, bits, group_size, q_size, kv_size, intermediate_size)
```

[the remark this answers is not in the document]
apsonawane: Not needed since we use self.is_lora now

src/python/py/models/builder.py
```diff
@@ -437,7 +441,7 @@ def save_model(self, out_dir):
         # Quantize ONNX model to desired precision
         # TODO: Replace by quantizing the MatMuls as they are created
         already_quantized_in_qdq_format = self.quant_type is not None and self.quant_attrs["use_qdq"]  # Skip quantizing `MatMul` in `DequantizeLinear --> Transpose --> MatMul` path
-        if self.onnx_dtype == "int4" and not already_quantized_in_qdq_format:
+        if self.onnx_dtype == "int4" and not already_quantized_in_qdq_format and not self.matmul_attrs["use_lora"]:
```
jambayk, line 444: MatMul4bits quantizer has an option to excludes nodes for quantization https://github.com/microsoft/onnxruntime/blob/e7987a6b0ba429c0bec248c4a471e1782da4be6c/onnxruntime/python/tools/quantization/matmul_4bits_quantizer.py#L1342 

Maybe instead of a flag, you can keep a set of the lora matmul names and provide it to the quantizer? Otherwise, if the user provides float base + float adapters with `int4` as precision, the output model will be fully float. But you might want to quantize the base model? 
also for a quantized base model + float adapters, you might want to quantize the lm head like https://github.com/microsoft/onnxruntime-genai/pull/940? Not sure what effect always quantizing the lm head on accuracy though.

src/python/py/models/quantized_model.py
```diff
@@ -334,23 +369,51 @@ def __init__(self, quant_type, input_path, bits, group_size, q_size, kv_size, in
                             # model.layers.layer_id.mlp.dense_h_to_4h.bias
                             module.mlp.gate_proj.bias = tensor[: intermediate_size]
                             module.mlp.down_proj.bias = tensor[intermediate_size: ]
+                        elif bool(re.match(r"^model.layers\.\d+\.self_attn.q_proj.lora_A\.weight$", name)):
```
jambayk, line 372: these only cover llama type models. phi3 has qkv_proj and gate_up_proj instead of q,k,v,gate_proj,up_proj

src/python/py/models/quantized_model.py
```diff
@@ -334,23 +369,51 @@ def __init__(self, quant_type, input_path, bits, group_size, q_size, kv_size, in
                             # model.layers.layer_id.mlp.dense_h_to_4h.bias
                             module.mlp.gate_proj.bias = tensor[: intermediate_size]
                             module.mlp.down_proj.bias = tensor[intermediate_size: ]
```
jambayk, line 371: btw, looks like this is a bug? shouldn't it be up_proj?

src/python/py/models/quantized_model.py
apsonawane: Yes, Kunal also mentioned about that. We can support for that as well

src/python/py/models/quantized_model.py
apsonawane: Updated

PR #1048, opened 2024-11-07
Add ability to modify provider options at runtime

RyanUnderhill:
Now a provider can be chosen and configured purely at runtime, to override what the genai_config.json specifies.
`
  auto config = OgaConfig::Create("path\to\model");
  config->SetProviderOption("cuda", "enable_quantum_drive", "1");
`
vs this if no runtime changes are needed (this is also the current API):
`
  auto model = OgaModel::Create("path\to\model");
`

This will solve the following use cases:
* User wants to change a runtime option for a provider already specified in the genai_config.json
* User wants to switch which provider is used vs what the genai_config.json specifies
* The genai_config.json specifies no provider, and one is completely chosen at runtime.

The OgaConfig provides these methods:
* ClearProviders() - Remove all providers currently specified (will default to cpu if nothing else happens)
* SetProvider("provider_name") - If provider is already exists, does nothing, otherwise adds to end of provider list
* SetProviderOption("provider_name", "option_name", "option_value") - If provider doesn't exist, adds it to end of list. Then sets an option on the provider.

### DIFF
--- a/src/java/src/main/native/ai_onnxruntime_genai_Config.cpp
+++ b/src/java/src/main/native/ai_onnxruntime_genai_Config.cpp
@@ -2,7 +2,7 @@
  * Copyright (c) Microsoft Corporation. All rights reserved.
  * Licensed under the MIT License.
  */
-#include "ai_onnxruntime_genai_Model.h"
+#include "ai_onnxruntime_genai_Config.h"
 
 #include "ort_genai_c.h"
 #include "utils.h"
--- a/src/java/src/main/native/ai_onnxruntime_genai_Config.cpp
+++ b/src/java/src/main/native/ai_onnxruntime_genai_Config.cpp
@@ -29,14 +29,14 @@ Java_ai_onnxruntime_genai_Config_destroyConfig(JNIEnv* env, jobject thiz, jlong 
 
 JNIEXPORT void JNICALL
 Java_ai_onnxruntime_genai_Config_clearProviders(JNIEnv* env, jobject thiz, jlong native_handle) {
-  OgaConfig* config = reinterpret_cast<OgaConfig*>(config_handle);
+  OgaConfig* config = reinterpret_cast<OgaConfig*>(native_handle);
   ThrowIfError(env, OgaConfigClearProviders(config));
 }
 
 JNIEXPORT void JNICALL
 Java_ai_onnxruntime_genai_Config_setProvider(JNIEnv* env, jobject thiz, jlong native_handle, jstring provider_name) {
   CString c_provider_name{env, provider_name};
-  OgaConfig* config = reinterpret_cast<OgaConfig*>(config_handle);
+  OgaConfig* config = reinterpret_cast<OgaConfig*>(native_handle);
 
   ThrowIfError(env, OgaConfigSetProvider(config, c_provider_name));
 }
@@ -46,7 +46,7 @@ Java_ai_onnxruntime_genai_Config_setProvider(JNIEnv* env, jobject thiz, jlong na
   CString c_provider_name{env, provider_name};
   CString c_option_name{env, option_name};
   CString c_option_value{env, option_value};
-  OgaConfig* config = reinterpret_cast<OgaConfig*>(config_handle);
+  OgaConfig* config = reinterpret_cast<OgaConfig*>(native_handle);
 
   ThrowIfError(env, OgaConfigSetProviderOption(config, c_provider_name, c_option_name, c_option_value));
 }
--- a/src/java/src/main/native/ai_onnxruntime_genai_Config.cpp
+++ b/src/java/src/main/native/ai_onnxruntime_genai_Config.cpp
@@ -1,0 +1,52 @@
+/*
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+#include "ai_onnxruntime_genai_Model.h"
+
+#include "ort_genai_c.h"
+#include "utils.h"
+
+using namespace Helpers;
+
+JNIEXPORT jlong JNICALL
+Java_ai_onnxruntime_genai_Config_createConfig(JNIEnv* env, jobject thiz, jstring model_path) {
+  CString path{env, model_path};
+
+  OgaConfig* config = nullptr;
+  if (ThrowIfError(env, OgaCreateConfig(path, &config))) {
+    return 0;
+  }
+
+  return reinterpret_cast<jlong>(config);
+}
+
+JNIEXPORT void JNICALL
+Java_ai_onnxruntime_genai_Config_destroyConfig(JNIEnv* env, jobject thiz, jlong native_handle) {
+  OgaConfig* config = reinterpret_cast<OgaConfig*>(native_handle);
+  OgaDestroyConfig(config);
+}
+
+JNIEXPORT void JNICALL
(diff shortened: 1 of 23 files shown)


src/java/src/main/native/ai_onnxruntime_genai_Config.cpp
```diff
@@ -0,0 +1,52 @@
+/*
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+#include "ai_onnxruntime_genai_Model.h"
```
skyline75489, line 5: This should be `ai_onnxruntime_genai_Config.h"

RyanUnderhill: I just noticed that too.

src/ort_genai_c.h
```diff
@@ -172,14 +173,56 @@ OGA_EXPORT void OGA_API_CALL OgaDestroyRuntimeSettings(OgaRuntimeSettings* setti
...
+ */
+OGA_EXPORT OgaResult* OGA_API_CALL OgaConfigClearProviders(OgaConfig* config);
+
+/*
+ * \brief Add the provider at the end of the list of providers in the given config if it doesn't already exist
+ * \param[in] config The config to set the provider on.
+ * \param[in] provider The provider to set on the config.
+ * \return OgaResult containing the error message if the setting of the provider failed.
+ */
+OGA_EXPORT OgaResult* OGA_API_CALL OgaConfigSetProvider(OgaConfig* config, const char* provider);
```
pranavsharma, line 196: 1. So, this is the one that applies to all models in the config (if there are more than 1 model)?
2. I suppose you'll add the model specific set provider later?

src/ort_genai_c.h
```diff
@@ -172,14 +173,56 @@ OGA_EXPORT void OGA_API_CALL OgaDestroyRuntimeSettings(OgaRuntimeSettings* setti
...
+ * \brief Creates a model from the given configuration directory.
  * \param[in] config_path The path to the model configuration directory. The path is expected to be encoded in UTF-8.
  * \param[in] device_type The device type to use for the model.
  * \param[out] out The created model.
  * \return OgaResult containing the error message if the model creation failed.
  */
 OGA_EXPORT OgaResult* OGA_API_CALL OgaCreateModel(const char* config_path, OgaModel** out);
 
+/*
+ * \brief Creates a model from the given configuration directory and device type.
```
pranavsharma, line 218: Documentation needs revision.

RyanUnderhill: Good catch, I caught and removed the 'and device type.' from the existing OgaCreateModel function but didn't realize I copied it here first.

src/ort_genai_c.h
pranavsharma: Documentation says that you are appending, but API says Set. What is the API doing?

src/python/python.cpp
```diff
@@ -5,6 +5,7 @@
 #include <pybind11/numpy.h>
 #include <iostream>
 #include "../generators.h"
+#include "../ort_genai.h"
```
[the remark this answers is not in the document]
RyanUnderhill: They do, the config object in python goes through the C API. I did it as a first test to start switching away from an invasive python bindings.

Python should use the same API that all of the other language bindings use.

src/models/debugging.cpp
```diff
@@ -19,7 +19,7 @@ static constexpr size_t c_value_count = 10;  // Dump this many values from the s
 template <typename... Types>
 const char* TypeToString(ONNXTensorElementDataType type, Ort::TypeList<Types...>) {
   const char* name = "(please add type to list)";
-  ((type == Ort::TypeToTensorType<Types> ? name = typeid(Types).name(), true : false) || ...);
+  (void)((type == Ort::TypeToTensorType<Types> ? name = typeid(Types).name(), true : false) || ...);
```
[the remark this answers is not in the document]
RyanUnderhill: To silence a compiler warning on Android that warns of an expression with an unused result.

src/python/python.cpp
```diff
@@ -433,7 +434,18 @@ PYBIND11_MODULE(onnxruntime_genai, m) {
...
 
+  // Pybind class using the C API for OgaConfig
+  pybind11::class_<OgaConfig>(m, "Config")
+      .def(pybind11::init([](const std::string& config_path) { return OgaConfig::Create(config_path.c_str()); }))
+      .def("set_provider", &OgaConfig::SetProvider)
+      .def("set_provider_option", &OgaConfig::SetProviderOption)
+      .def("clear_providers", &OgaConfig::ClearProviders);
+
   pybind11::class_<Model, std::shared_ptr<Model>>(m, "Model")
+      .def(pybind11::init([](const OgaConfig& config) {
```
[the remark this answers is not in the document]
RyanUnderhill: See the #include "../ort_genai.h" comment below

test/c_api_tests.cpp
```diff
@@ -18,9 +18,24 @@
 #define PHI2_PATH MODEL_PATH "phi-2/int4/cpu"
 #endif
 #endif
+TEST(CAPITests, Config) {
+#if TEST_PHI2
```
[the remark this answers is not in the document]
RyanUnderhill: Not really. But we want to remove the gpt2 models or at least convert them to something new (they're the only ones using the KV_Cache_Combined stuff.

test/python/test_onnxruntime_genai_api.py
```diff
@@ -45,7 +54,8 @@
 def test_greedy_search(test_data_path, relative_model_path):
     model_path = os.fspath(Path(test_data_path) / relative_model_path)
 
-    model = og.Model(model_path)
+    config = og.Config(model_path) # Test using config vs path directly
+    model = og.Model(config)
```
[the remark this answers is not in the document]
RyanUnderhill: It only needs to be done once somewhere, I don't want to double the test run time for that.

test/python/test_onnxruntime_genai_api.py
```diff
@@ -29,6 +29,15 @@
 if og.is_rocm_available():
     devices.append("rocm")
 
+def config_test(test_data_path):
```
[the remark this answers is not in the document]
RyanUnderhill: Hmm, good point

natke (commented):
Can you add a code sample in the PR for the scenario that is driving this change ie to set the default provider? Did we lose the SetDefaultProvider() API?

RyanUnderhill:
> Can you add a code sample in the PR for the scenario that is driving this change ie to set the default provider? Did we lose the SetDefaultProvider() API?

The initial 'SetDefaultProvider' proposed API would rearrange the providers so the one you mention is first in the list. This isn't really what we want since it doesn't match the onnxruntime API for setting providers.

Where should the sample go? This is how it will look:

`auto config = OgaConfig::Create("path\to\model");
config.ClearProviders(); // Only needed if config already specifies a provider, but safe to do in both cases
config.SetProvider("dml");
`

src/ort_genai_c.h
RyanUnderhill: It's both. If it's already in the list, it does nothing. If not, it appends it to the end.

Should I rename it to AppendProvider and have it be an error if it already exists?

pranavsharma: Discussed offline. Let's call it Append.

pranavsharma (approved):
Approving for the C/C++ API changes.

PR #1050, opened 2024-11-08
Fix C# phi-2 test

skyline75489:
The C# phi-2 test was skipped due to the wrong model path. This PR intends to fix it.

### DIFF
--- a/.github/workflows/mac-cpu-arm64-build.yml
+++ b/.github/workflows/mac-cpu-arm64-build.yml
@@ -82,7 +82,7 @@ jobs:
         run: |
           export ORTGENAI_LOG_ORT_LIB=1
           cd test/csharp
-          dotnet test /p:Configuration=Release /p:NativeBuildOutputDir="../../build/cpu/osx-arm64"
+          dotnet test /p:Configuration=Release /p:NativeBuildOutputDir="../../build/cpu/osx-arm64" --verbosity normal
 
       - name: Run tests
         run: |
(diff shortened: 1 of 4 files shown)


skyline75489:
Note that there's still resource leaks in the C# tests. Will try to fix it later.

test/csharp/TestOnnxRuntimeGenAIAPI.cs
```diff
@@ -5,28 +5,24 @@
 using System.IO;
 using System.Linq;
 using System.Runtime.InteropServices;
+using System.Runtime.CompilerServices;
 using Xunit;
 using Xunit.Abstractions;
 
 namespace Microsoft.ML.OnnxRuntimeGenAI.Tests
 {
-    public class TestsFixture : IDisposable
```
skyline75489, line 13: This test-case level fixture doesn't work. Because after `OgaShutdown` is called, it can't recover and the entire test crashes.

.github/workflows/mac-cpu-arm64-build.yml
```diff
@@ -82,7 +82,7 @@ jobs:
         run: |
           export ORTGENAI_LOG_ORT_LIB=1
           cd test/csharp
-          dotnet test /p:Configuration=Release /p:NativeBuildOutputDir="../../build/cpu/osx-arm64"
+          dotnet test /p:Configuration=Release /p:NativeBuildOutputDir="../../build/cpu/osx-arm64" --verbosity normal
```
skyline75489, line 85: This produces much better results in terms of test cases.

RyanUnderhill (approved):

baijumeswani (approved):

Issue #980, opened 2024-10-14
OnnxRuntimeGenAIException: 'bad allocation' when option `max_length` is not set


f2bo:
**Describe the bug**

I've been using the `Phi-3-mini-128k-instruct-onnx` model successfully in an application with the `Microsoft.ML.OnnxRuntimeGenAI.DirectML` package for some time. Recently, I began exploring using Semantic Kernel and its ONNX connector instead and was surprised to find that it crashed with a `bad allocation` exception when using the same model given that Semantic Kernel also depends on the `OnnxRuntimeGenAI` package. In contrast, replacing the model with the `Phi-3-mini-4k-instruct-onnx` model, it ran successfully. 

After comparing what the Semantic Kernel connector was doing with my original code, I managed to narrow it down to having missed setting a value for the `max_length` option when configuring the connector. In SK, this is named `MaxTokens` in the `OnnxRuntimeGenAIPromptExecutionSettings` class.

**To Reproduce**
The issue can be reproduced with the OnnxRuntimeGenAI package directly using the following code and commenting out the line that sets `max_length`.

``` C#
// download from https://huggingface.co/microsoft/Phi-3-mini-128k-instruct-onnx and set modelPath
string modelPath = "..... \\microsoft\\Phi-3-mini-128k-instruct-onnx\\cpu_and_mobile\\cpu-int4-rtn-block-32-acc-level-4";

using Model model = new Model(modelPath);
using Tokenizer tokenizer = new Tokenizer(model);

string prompt = "Tell me a joke";
var sequences = tokenizer.Encode($"<|user|>{prompt}<|end|><|assistant|>");

using GeneratorParams generatorParams = new GeneratorParams(model);
generatorParams.SetSearchOption("max_length", 8192);			// <==== COMMENT OUT THIS LINE TO TRIGGER THE EXCEPTION
generatorParams.SetInputSequences(sequences);

using var tokenizerStream = tokenizer.CreateStream();
using var generator = new Generator(model, generatorParams);
while (!generator.IsDone())
{
    generator.ComputeLogits();
    generator.GenerateNextToken();
    Console.Write(tokenizerStream.Decode(generator.GetSequence(0)[^1]));
}
```

**Expected behavior**
No exception. Alternatively, if the `max_length` parameter isn't optional, then a clearer exception than `bad allocation`.

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Version: 22H2

baijumeswani:
[...]

PR #1058, opened 2024-11-11
Catch ort exception when trying to allocate key value cache

baijumeswani:
Addresses https://github.com/microsoft/onnxruntime-genai/issues/980

### DIFF
--- a/src/models/kv_cache.cpp
+++ b/src/models/kv_cache.cpp
@@ -184,7 +184,7 @@ KV_Cache::KV_Cache(State& state)
         }
       }
     }
-  } catch (const Ort::Exception& e) {
+  } catch ([[maybe_unused]] const Ort::Exception&) {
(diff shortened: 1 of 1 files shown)


RyanUnderhill (approved):
The diff in github looks bigger than it really is due to the indents?

Since it looks like it's just adding the try/catch unless I'm missing something.

baijumeswani:
Yes, the indents make it look bigger than it really is.

f2bo:
Hi. I originally reported https://github.com/microsoft/onnxruntime-genai/issues/980. While the updated error message is better, it doesn't identify the root cause unless you are familiar with the runtime's internals, especially when raised following a call to a higher level component. Perhaps, it would be more actionable if it made the association with the 'max_length' option.

baijumeswani:
Thanks for the review. :)

PR #1066, opened 2024-11-15
Added Phi-3.5-vision model validation in pipeline

skyline75489:
* Restructure Phi-3.5-vision C# and Python examples to make them more CI-friendly.
* Added Phi-3.5-vision validation steps to the existing validation pipeline.

Python run:

* https://aiinfra.visualstudio.com/ONNX%20Runtime/_build/results?buildId=609855&view=results

NuGet run:

* https://aiinfra.visualstudio.com/ONNX%20Runtime/_build/results?buildId=609849&view=results

---


examples/python/phi3v.py
```diff
@@ -83,5 +107,14 @@ def run(args: argparse.Namespace):
...
         "-p", "--provider", type=str, required=True, help="Provider to run model"
     )
+    parser.add_argument(
+        "--image_paths", nargs='*', type=str, required=False, help="Path to the images"
+    )
+    parser.add_argument(
+        '-pr', '--prompt', required=False, help='Input prompts to generate tokens from.'
+    )
+    parser.add_argument(
+        '--interactive', default=False, required=False, help='Interactive mode'
```
kunal-vaishnavi, line 117: Can we keep the existing behavior as the default? Users will be referring to this script when running the new vision models and our published examples on [the ORT website](https://onnxruntime.ai/docs/genai/tutorials/phi3-v.html) and on [GitHub](https://github.com/microsoft/onnxruntime-genai/blob/main/examples/python/phi-3.5-vision.md#4-run-phi-35-vision-onnx-models) refer to `-m` and `p` as the only required flags.

kunal-vaishnavi, line 117: Can we keep the existing behavior as the default? Users will be referring to this script when running the new vision models and our published examples on [the ORT website](https://onnxruntime.ai/docs/genai/tutorials/phi3-v.html) and on [GitHub](https://github.com/microsoft/onnxruntime-genai/blob/main/examples/python/phi-3.5-vision.md#4-run-phi-35-vision-onnx-models) refer to `-m` and `-p` as the only required flags.

kunal-vaishnavi: Yes, that's fine. Since the extra arguments are for CI testing, we should note that information in the `help` section for each of the new arguments. Otherwise, users may try to use the new arguments when seeing which arguments are available with `python phi3v.py --help` when the new arguments don't need to be used for the default behavior.

This comment is mainly referring to the current usage of `--interactive` as it changes the script's default behavior.

https://github.com/microsoft/onnxruntime-genai/blob/d9db677ce5ac4d8d23257c1402ada1b581609913/examples/python/phi3v.py#L40-L58

https://github.com/microsoft/onnxruntime-genai/blob/d9db677ce5ac4d8d23257c1402ada1b581609913/examples/python/phi3v.py#L75-L81

Since the `if` cases are what's originally in the script, the script's default behavior will be non-interactive and will require `--interactive` to be passed for interactive mode. We should make interactive mode the default for users and use non-interactive mode with a flag for CI testing.

skyline75489: Got it. I'll reverse the default of `interactive` to `True` and change the CI script accordingly

examples/csharp/HelloPhi3V/HelloPhi3V.csproj
```diff
@@ -9,9 +9,9 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.ML.OnnxRuntimeGenAI" Version="0.3.0" Condition=" '$(Configuration)' == 'Debug' OR '$(Configuration)' == 'Release' " />
-    <PackageReference Include="Microsoft.ML.OnnxRuntimeGenAI.Cuda" Version="0.3.0" Condition=" '$(Configuration)' == 'Debug_Cuda' OR '$(Configuration)' == 'Release_Cuda' " />
-    <PackageReference Include="Microsoft.ML.OnnxRuntimeGenAI.DirectML" Version="0.3.0" Condition=" '$(Configuration)' == 'Debug_DirectML' OR '$(Configuration)' == 'Release_DirectML' " />
+    <PackageReference Include="Microsoft.ML.OnnxRuntimeGenAI" Version="0.5.0-dev" Condition=" '$(Configuration)' == 'Debug' OR '$(Configuration)' == 'Release' " />
+    <PackageReference Include="Microsoft.ML.OnnxRuntimeGenAI.Cuda" Version="0.5.0-dev" Condition=" '$(Configuration)' == 'Debug_Cuda' OR '$(Configuration)' == 'Release_Cuda' " />
```
kunal-vaishnavi, line 13: Should this be 0.5.1 instead?

examples/python/phi3v.py
```diff
@@ -83,5 +125,14 @@ def run(args: argparse.Namespace):
     parser.add_argument(
         "-p", "--provider", type=str, required=True, help="Provider to run model"
     )
+    parser.add_argument(
+        "--image_paths", nargs='*', type=str, required=False, help="Path to the images"
```
kunal-vaishnavi, line 129:
```suggestion
        "--image_paths", nargs='*', type=str, required=False, help="Path to the images, mainly for CI usage"
```

examples/python/phi3v.py
```diff
@@ -83,5 +125,14 @@ def run(args: argparse.Namespace):
...
+    )
+    parser.add_argument(
+        '-pr', '--prompt', required=False, help='Input prompts to generate tokens from.'
```
kunal-vaishnavi, line 132:
```suggestion
        '-pr', '--prompt', required=False, help='Input prompts to generate tokens from, mainly for CI usage'
```

examples/csharp/HelloPhi3V/HelloPhi3V.csproj
skyline75489: Yeah...this is awkward. Ideally, this should be whatever in `VERSION_INFO`. It's better to have something with `-dev` suffix, because this will differ from what's in official NuGet. We always want to test locally built packages with `-dev` suffix in the CI pipeline. But after each release, we'll update `VERSION_INFO` and sometimes it will not have `-dev` suffix like right now. I'll update this to 0.5.1 later but for the next release we still need to tweak this, in order to always test the latest version.

kunal-vaishnavi (approved):

PR #1080, opened 2024-11-20
Allow OgaShutdown to be called multiple times

RyanUnderhill:
It's nice to be able to call OgaShutdown() then call more Oga* APIs and call OgaShutdown() again, vs just one OgaShutdown() being the permanent end of Oga for a process.

This makes it easier to implement some tests and language bindings.

### DIFF
--- a/src/generators.cpp
+++ b/src/generators.cpp
@@ -61,11 +61,21 @@ struct ValidateShutdown {
   }
 };
 
+static std::unique_ptr<OrtGlobals> g_globals;
+static std::mutex g_globals_mutex;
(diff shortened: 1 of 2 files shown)


skyline75489:
Would a ref-count approach better suits our need? Like:

https://github.com/microsoft/onnxruntime/blob/ac9c135b9543ad0374fe335bc3dc5feb0f24f010/onnxruntime/core/session/ort_env.cc#L29C13-L29C24

`OrtGlobals` feels like our version of `OrtEnv`, which exist globally and pretty much everything relies on it.

RyanUnderhill:
Refcounting has a drawback that we'd now need an OgaStartup to correspond with the OgaShutdown.

PR #1084, opened 2024-11-21
Fix C# Win-Cuda-x64 GHA pipeline

skyline75489:


test/csharp/TestOnnxRuntimeGenAIAPI.cs
```diff
@@ -524,12 +579,19 @@ public void TestAdapters()
                     generator.GenerateNextToken();
                 }
                 using var logits = generator.GetOutput("logits");
-                Assert.Equal(ElementType.float32, logits.Type());
+                if (_useCudaModel)
+                {
+                    Assert.Equal(ElementType.float16, logits.Type());
+                    // TODO: GetData with float16?
```
skyline75489, line 585: I don't think we support this in GenAI yet. I have encountered the same issue when I was trying to copy this test case to Java.

@baijumeswani Is it OK to merge this first, and handle the `GetData` with float16 later?

.github/workflows/win-cuda-x64-build.yml
```diff
@@ -100,6 +95,12 @@ jobs:
         
         Get-ChildItem -Path $env:GITHUB_WORKSPACE\$env:binaryDir -Recurse
 
+    - name: Build the C# API and Run the C# Tests
+      run: |
+        $env:PATH = "${{ env.cuda_dir }}\\v${{ env.cuda_version }}\\bin;" + $env:PATH
```
baijumeswani, line 100: Do we need this? We have a `Add CUDA to PATH` item on line 72, would that not work?

baijumeswani (approved):

.github/workflows/win-cuda-x64-build.yml
skyline75489: Yes this is needed. The env for each run is separated. The code in line 72 has no effect here.

PR #1091, opened 2024-11-22
Add pre-generated prompts option for benchmark

omer-demir:
During benchmarking, we wanted to have pre-generated prompts that have been prepared for better benchmark result. Hence, It can be handy during benchmarking. In our test, we wanted to focus only token generation and sampling on SLM.


benchmark/python/prompts.json
```diff
@@ -0,0 +1,7 @@
+{
+    "16": "How are astronauts launched into space quickly on those rockets? ",
+    "64": "",
```
RyanUnderhill, line 3: Should there be prompts for the larger sizes?

omer-demir: I left that part not committed, I can add sample prompts as well.

natke: added

benchmark/python/benchmark_e2e.py
```diff
@@ -83,6 +83,14 @@ def generate_prompt(model, tokenizer, prompt_length, use_graph_capture) -> str:
         generator.generate_next_token()
     return tokenizer.decode(generator.get_sequence(0))
 
+# Use prompt length to get pre-defined prompt
+def get_prompt_by_length(prompt_length):
+    json_path = "prompts.json"
```
kunal-vaishnavi, line 88: Instead of uploading another copy of `prompts.json`, can we download it from [here](https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/python/tools/transformers/models/llama/prompts.json) and save it to disk using `requests` or `urllib` instead? That way, only one location has to be updated when adding other prompt lengths.

benchmark/python/benchmark_e2e.py
```diff
@@ -232,6 +240,9 @@ def run_benchmark(args, batch_size, prompt_length, generation_length, max_length
         # use random tokens instead of generating a prompt using the model and then tokenizing it
         tokens = np.random.randint(100, size=(batch_size, prompt_length))
         prompt = [tokenizer.decode(tokens[0])] * batch_size
+    elif args.use_prompt_set:
+        prompt = get_prompt_by_length(prompt_length)
+        tokens = tokenizer.encode_batch(prompt)
```
kunal-vaishnavi, line 245: Different tokenizers can encode prompts into different prompt lengths. Some additional work is needed to get the desired prompt length. You can see an example of how to do this [here](https://github.com/microsoft/onnxruntime/blob/00b262dbb4328173bf0e11cbcd1d62e18df3d686/onnxruntime/python/tools/transformers/models/llama/llama_inputs.py#L424-L441).

benchmark/python/benchmark_e2e.py
kunal-vaishnavi, line 88: Instead of uploading another copy of `prompts.json`, can we download it from [here](https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/python/tools/transformers/models/llama/prompts.json) and save it to disk using `requests` or `urllib` instead? That way, only one location has to be updated when adding other prompts.

omer-demir: would it make sense to rely on external file? If that changes, we may have problem with this. Besides, in the benchmark environment, it seems logical to use local files instead of relying on internet connectivity. What do you think?

benchmark/python/benchmark_e2e.py
omer-demir: so basically, we will check the `tokens` length against requested `prompt_length` and we will add/trim if it is needed. Is that correct?

kunal-vaishnavi: Yes, that's correct.

benchmark/python/benchmark_e2e.py
kunal-vaishnavi: If connectivity is a concern in your benchmark environment, then I think it's fine to include a copy of the file.

benchmark/python/benchmark_e2e.py
omer-demir: I introduced check based on requested length to fill or remove from tokens.

kunal-vaishnavi (approved):

PR #1094, opened 2024-11-22
Make tests discoverable at build time

baijumeswani:
`python build.py` will now run tests if tests are enabled.


RyanUnderhill (approved):

baijumeswani:
Fixed in https://github.com/microsoft/onnxruntime-genai/pull/1615

PR #1105, opened 2024-11-28
Add SetLogits and GetLogits C API

Taka152:
C API should be the same as Python API. It only supports the last token's logits, and this is the difference with GetOutput("logits") which returns all prompt tokens' logits.

### DIFF
--- a/src/ort_genai_c.cpp
+++ b/src/ort_genai_c.cpp
@@ -372,13 +372,13 @@ OgaResult* OGA_API_CALL OgaGenerator_GetOutput(const OgaGenerator* oga_generator
 
 OgaResult* OGA_API_CALL OgaGenerator_GetLogits(OgaGenerator* oga_generator, OgaTensor** out) {
   OGA_TRY
-  // Run ComputeLogits to get the logits if computed_logits_ is false
   auto logits_span = reinterpret_cast<Generators::Generator*>(oga_generator)->GetLogits();
   auto& generator = *reinterpret_cast<const Generators::Generator*>(oga_generator);
   std::vector<int64_t> shape{generator.state_->params_->search.batch_size, 1, generator.model_->config_->model.vocab_size};
   logits_span.CopyDeviceToCpu();
   std::span<const float> cpu_logits_span = logits_span.CpuSpan();
 
+  // Copy logits to cpu tensor
   std::unique_ptr<OrtValue> ortvalue_clone = OrtValue::CreateTensor(generator.model_->allocator_cpu_,
                                                                     shape,
                                                                     ONNX_TENSOR_ELEMENT_DATA_TYPE_FLOAT);
--- a/src/ort_genai_c.cpp
+++ b/src/ort_genai_c.cpp
@@ -375,13 +375,14 @@ OgaResult* OGA_API_CALL OgaGenerator_GetLogits(OgaGenerator* oga_generator, OgaT
   auto logits_span = reinterpret_cast<Generators::Generator*>(oga_generator)->GetLogits();
   auto& generator = *reinterpret_cast<const Generators::Generator*>(oga_generator);
   const std::array<int64_t, 3> shape{generator.state_->params_->search.batch_size, 1, generator.model_->config_->model.vocab_size};
-  auto cpu_logits_span = logits_span.CopyDeviceToCpu();
+  std::span<const float> cpu_logits_span = logits_span.CopyDeviceToCpu();
 
   // Copy logits to cpu tensor
   std::unique_ptr<OrtValue> ortvalue_clone = OrtValue::CreateTensor(generator.model_->allocator_cpu_,
                                                                     shape,
                                                                     ONNX_TENSOR_ELEMENT_DATA_TYPE_FLOAT);
   auto clone_span = std::span<float>(ortvalue_clone->GetTensorMutableData<float>(), cpu_logits_span.size());
+  Generators::copy(cpu_logits_span, clone_span);
   auto tensor = std::make_shared<Generators::Tensor>(std::move(ortvalue_clone));
   tensor->external_owner_ = tensor;
   *out = reinterpret_cast<OgaTensor*>(tensor.get());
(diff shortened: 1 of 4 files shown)


src/ort_genai_c.cpp
```diff
@@ -370,6 +370,48 @@ OgaResult* OGA_API_CALL OgaGenerator_GetOutput(const OgaGenerator* oga_generator
...
+
+OgaResult* OGA_API_CALL OgaGenerator_SetLogits(OgaGenerator* oga_generator, OgaTensor* tensor) {
+  OGA_TRY
+  auto generator = reinterpret_cast<Generators::Generator*>(oga_generator);
+  if (!generator->computed_logits_) {
+    throw std::runtime_error("logits are not computed yet. Please call GenerateNextToken or AppendTokens before calling SetLogits.");
+  }
+  auto logits_tensor = reinterpret_cast<Generators::Tensor*>(tensor);
+  size_t element_count = logits_tensor->ort_tensor_->GetTensorTypeAndShapeInfo()->GetElementCount();
+  auto new_logits_span = std::span<const float>(reinterpret_cast<float*>(logits_tensor->ort_tensor_->GetTensorMutableRawData()), element_count);
```
RyanUnderhill, line 402:
```suggestion
  auto new_logits_span = std::span<const float>(logits_tensor->ort_tensor_->GetTensorData<float>()), element_count);
```

src/ort_genai_c.cpp
```diff
@@ -370,6 +370,48 @@ OgaResult* OGA_API_CALL OgaGenerator_GetOutput(const OgaGenerator* oga_generator
...
+  auto& generator = *reinterpret_cast<const Generators::Generator*>(oga_generator);
+  std::vector<int64_t> shape{generator.state_->params_->search.batch_size, 1, generator.model_->config_->model.vocab_size};
+  logits_span.CopyDeviceToCpu();
+  std::span<const float> cpu_logits_span = logits_span.CpuSpan();
+
+  // Copy logits to cpu tensor
+  std::unique_ptr<OrtValue> ortvalue_clone = OrtValue::CreateTensor(generator.model_->allocator_cpu_,
+                                                                    shape,
+                                                                    ONNX_TENSOR_ELEMENT_DATA_TYPE_FLOAT);
+  auto clone_span = std::span<float>(reinterpret_cast<float*>(ortvalue_clone->GetTensorMutableRawData()), cpu_logits_span.size());
```
RyanUnderhill, line 385:
```suggestion
  auto clone_span = std::span<float>(ortvalue_clone->GetTensorMutableData<float>()), cpu_logits_span.size());
```

src/ort_genai_c.h
```diff
@@ -341,6 +341,25 @@ OGA_EXPORT OgaResult* OGA_API_CALL OgaGenerator_RewindTo(OgaGenerator* generator
  */
 OGA_EXPORT OgaResult* OGA_API_CALL OgaGenerator_GetOutput(const OgaGenerator* oga_generator, const char* name, OgaTensor** out);
 
+/*
+ * \brief Returns a copy of the logits from the model as an OgaTensor on CPU. The buffer is owned by returned OgaTensor
+ *        and will be released when the OgaTensor is destroyed
+ * \param[in] generator The generator to run the GetLogits on
```
RyanUnderhill, line 347:
```suggestion
 * \param[in] generator The generator get the logits from
```

'run the GetLogits' sounds like it might run the model, this should only get them if available?

Taka152: Yes, I think it might be better to raise error if logits are not ready. But since we only have GenerateNextToken() now, so we can only call GetLogits() to run logits if it's not ready and get current step's logits?

src/ort_genai_c.cpp
```diff
@@ -370,6 +370,47 @@ OgaResult* OGA_API_CALL OgaGenerator_GetOutput(const OgaGenerator* oga_generator
...
+  OGA_TRY
+  auto logits_span = reinterpret_cast<Generators::Generator*>(oga_generator)->GetLogits();
+  auto& generator = *reinterpret_cast<const Generators::Generator*>(oga_generator);
+  const std::array<int64_t, 3> shape{generator.state_->params_->search.batch_size, 1, generator.model_->config_->model.vocab_size};
+  std::span<const float> cpu_logits_span = logits_span.CopyDeviceToCpu();
+
+  // Copy logits to cpu tensor
+  std::unique_ptr<OrtValue> ortvalue_clone = OrtValue::CreateTensor(generator.model_->allocator_cpu_,
+                                                                    shape,
+                                                                    ONNX_TENSOR_ELEMENT_DATA_TYPE_FLOAT);
```
RyanUnderhill, line 383: Thought of one more:

```suggestion
  std::unique_ptr<OrtValue> ortvalue_clone = OrtValue::CreateTensor<float>(generator.model_->allocator_cpu_, shape);
```

src/ort_genai_c.cpp
```diff
@@ -370,6 +370,47 @@ OgaResult* OGA_API_CALL OgaGenerator_GetOutput(const OgaGenerator* oga_generator
   OGA_CATCH
 }
 
+OgaResult* OGA_API_CALL OgaGenerator_GetLogits(OgaGenerator* oga_generator, OgaTensor** out) {
+  OGA_TRY
+  auto logits_span = reinterpret_cast<Generators::Generator*>(oga_generator)->GetLogits();
+  auto& generator = *reinterpret_cast<const Generators::Generator*>(oga_generator);
```
RyanUnderhill, line 376:
```suggestion
  auto& generator = *reinterpret_cast<const Generators::Generator*>(oga_generator);
  auto logits_span = generator->GetLogits();
```

src/ort_genai_c.cpp
```diff
@@ -372,13 +372,13 @@ OgaResult* OGA_API_CALL OgaGenerator_GetOutput(const OgaGenerator* oga_generator
 
 OgaResult* OGA_API_CALL OgaGenerator_GetLogits(OgaGenerator* oga_generator, OgaTensor** out) {
   OGA_TRY
-  auto& generator = *reinterpret_cast<Generators::Generator*>(oga_generator);
-  auto logits_span = generator.GetLogits();
-  const std::array<int64_t, 3> shape{generator.state_->params_->search.batch_size, 1, generator.model_->config_->model.vocab_size};
+  auto generator = reinterpret_cast<Generators::Generator*>(oga_generator);
```
RyanUnderhill, line 375: Just curious why you switched to generator being a pointer vs a reference?

Taka152: Just to be consistent with SetLogits(), which do you prefer?

RyanUnderhill: Slight preference for references when the value can't be nullptr, but it's up to you. There's plenty of other pointers there.

RyanUnderhill (approved):

PR #1110, opened 2024-12-02
Fix past_present_share_buffer=false kv cache bug and asserts

aciddelgado:
This PR fixes a bug where, when using continuous decoding, the output would differ depending on whether past_present_share_buffer was on or off. The bug was actually deeper than this, as the KV-Cache was missing the last generated token when using continuous decoding, regardless of share buffer.

It also fixes a few assert statements.

### DIFF
--- a/src/generators.cpp
+++ b/src/generators.cpp
@@ -283,6 +283,12 @@ void Generator::AppendTokens(const cpu_span<int32_t>& input_ids) {
   if (search_->GetSequenceLength() != 0 && state_->params_->search.batch_size > 1)
     throw std::runtime_error("AppendTokens can only be called once for batch_size > 1. To call AppendTokens again, use RewindToLength(0)");
 
+  if (just_generated_) {
+    just_generated_ = false;
+    auto next_tokens = search_->GetNextTokens();
+    ComputeLogits(next_tokens);
+  }
+
   auto input_ids_device = AllocateInputIdsOnDevice(input_ids);
   search_->AppendTokens(input_ids_device);
 
@@ -370,6 +376,7 @@ void Generator::GenerateNextToken() {
            << std::endl;
   }
 
+  just_generated_ = true;
   if (!search.do_sample || search.top_k == 1) {
     search_->SelectTop();
     return;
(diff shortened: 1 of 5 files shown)


src/generators.cpp
```diff
@@ -283,6 +283,12 @@ void Generator::AppendTokens(const cpu_span<int32_t>& input_ids) {
   if (search_->GetSequenceLength() != 0 && state_->params_->search.batch_size > 1)
     throw std::runtime_error("AppendTokens can only be called once for batch_size > 1. To call AppendTokens again, use RewindToLength(0)");
 
+  if (just_generated_) {
+    just_generated_ = false;
+    auto next_tokens = search_->GetNextTokens();
+    ComputeLogits(next_tokens);
```
RyanUnderhill, line 289:
```suggestion
    ComputeLogits(search_->GetNextTokens());
```

src/generators.h
```diff
@@ -129,7 +129,8 @@ struct Generator : LeakChecked<Generator> {
  private:
   DeviceSpan<int32_t> AllocateInputIdsOnDevice(const cpu_span<int32_t>& input_ids);
   void ComputeLogits(DeviceSpan<int32_t>& next_tokens);
-  bool just_rewinded_{false};
+  bool just_rewinded_{};
```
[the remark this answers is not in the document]
RyanUnderhill: Then would it make more sense to have an enum of the states?
enum State { normal, rewound, generated };

(you can probably think of a better name for normal)

guschmue:
tested with webgpu and dynamic kv_cache ... super, that works now

guschmue (approved):

src/generators.h
```diff
@@ -127,9 +127,12 @@ struct Generator : LeakChecked<Generator> {
   bool computed_logits_{};  // Set to true in ComputeLogits() and false after appending a token to ensure a 1 to 1 call ratio
 
  private:
-  DeviceSpan<int32_t> AllocateInputIdsOnDevice(const cpu_span<int32_t>& input_ids);
-  void ComputeLogits(DeviceSpan<int32_t>& next_tokens);
-  bool just_rewinded_{false};
+  DeviceSpan<int32_t> AllocateInputIdsOnDevice(const cpu_span<int32_t> input_ids);
+  void ComputeLogits(DeviceSpan<int32_t> next_tokens);
+  enum Action { standard,
```
RyanUnderhill, line 132: Looks good, can you add some one line comments explaining each state?

PR #1116, opened 2024-12-03
Add Debug Pipelines

aciddelgado:
Issues with asserts were previously missed because all pipelines ran in Release mode. This PR adds Debug pipelines.


baijumeswani:
Do we need both Release and Debug? Can we instead convert the existing pipelines to build Debug?

natke:
> Do we need both Release and Debug? Can we instead convert the existing pipelines to build Debug?

We may not need both for every PR, but we will definitely want to the pipelines in Release mode and ensure they are successful before doing a release

skyline75489:
So... I was thinking maybe we should remove presets once and for all? Using presets is causing the switch between `Debug` and `Release` even more difficult. It also causes the build time to be doubled for Java as I mentioned in the group chat.

PR #1118, opened 2024-12-04
Fix comments and function names/arguments for rewind and continuous api's

aciddelgado:
Comments and function names / arguments were confusing. This PR fixes them by aligning them with each other.


skyline75489:
Thanks! This shows why we need something like #1107 that can make sure both the function name & parameters name are equal across different language bindings.

skyline75489 (approved):

PR #1122, opened 2024-12-05
Update phi3v.py

ivo-andreev:
Proposed fix for an error running on CPU: 
"  File "...\phi3v.py", line 140, in <module>  run(args)  
  File "...\phi3v.py", line 103, in run generator.generate_next_token()  
RuntimeError: Must call ComputeLogits before GenerateNextToken"

and 

RuntimeError: IsDone() can't be called in the middle of processing logits

### DIFF
--- a/examples/python/phi3v.py
+++ b/examples/python/phi3v.py
@@ -99,13 +99,22 @@ def run(args: argparse.Namespace):
         generator = og.Generator(model, params)
         start_time = time.time()
 
-        while not generator.is_done():
-            generator.generate_next_token()
-
-            new_token = generator.get_next_tokens()[0]
-            print(tokenizer_stream.decode(new_token), end="", flush=True)
-
-        print()
+        generated_text = ""   
+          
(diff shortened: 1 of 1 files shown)


ivo-andreev:

baijumeswani:
@ivo-andreev this example was updated as part of a bigger change for continuous decoding. You can find the pull-request that introduced these changes here: https://github.com/microsoft/onnxruntime-genai/pull/867.

Unfortunately, the updated example from that pull-request no longer works with the currently published onnxruntime-genai packages. Apologies for the inconvenience. If the user were to build the library from source, this example would work.

If you would like to refer to the example for the currently published packages, I would suggest using the example from the github tag associated with the release you want it to work with. For example, use https://github.com/microsoft/onnxruntime-genai/blob/v0.5.2/examples/python/phi3v.py for 0.5.2.

ivo-andreev:
@baijumeswani Thanks, I can confirm that the linked version works well.

PR #1129, opened 2024-12-09
Improve the documentation of C API header

skyline75489:
* Fix the doc string format
* Normalize the doc strings & parameter names to pass doxygen check.


baijumeswani (approved):

baijumeswani:
Should we consider grouping the API correctly from the beginning?

baijumeswani:
We can do it as part of a subsequent PR.

PR #1132, opened 2024-12-09
Avoid potential desynchronization of cpu and device memory

aciddelgado:
A possible desynchronization of cpu and device memory of new_tokens could result in bugs. This PR guards against any potential bugs, although no issues have been reported thus far to the best of my knowledge.


ambroser53:
Could this potentially be a fix for when encountering the following error when using the dml EP?
```
System.AccessViolationException: Attempted to read or write protected memory. This is often an indication that other memory is corrupt.
```

PR #1134, opened 2024-12-10
Clean up logits in phi3 C++ example

daijh:
Remove the logits demo code as it contains errors and lacks necessary C language implementation

### DIFF
--- a/examples/c/src/phi3.cpp
+++ b/examples/c/src/phi3.cpp
@@ -164,19 +164,6 @@ void CXX_API(const char* model_path) {
           is_first_token = false;
         }
 
-        // Show usage of GetOutput
-        std::unique_ptr<OgaTensor> output_logits = generator->GetOutput("logits");
-
-        // Assuming output_logits.Type() is float as it's logits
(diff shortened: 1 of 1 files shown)


daijh:
Sorry not provide more detail about the PR.

Build genai with 4b12d57dfbc0097bf3f1b9c7660e2e79f02ff003.

> Creating tokenizer...
Prompt: (Use quit() to exit) Or (To terminate current output generation, press Ctrl+C)
some prompt
Generating response...
**Session Terminated: Unsupported Device type: 4**

It happens in 
```c++
std::unique_ptr<OgaTensor> output_logits = generator->GetOutput("logits");
```

As `output_logits` is not actually used, so simply removing it to avoid exception.

baijumeswani (approved):

baijumeswani:
Since that part of the code is not really doing anything, I think it makes sense to remove it from the example. @ajindal1 Maybe we can document in another way.

[daijh](https://github.com/daijh) Thanks for the contribution.

PR #1136, opened 2024-12-10
Added missing javadoc to improve the doc quality

skyline75489:
Internal work item: https://task.ms/aii/53908

### DIFF
--- a/src/java/src/main/java/ai/onnxruntime/genai/Adapters.java
+++ b/src/java/src/main/java/ai/onnxruntime/genai/Adapters.java
@@ -4,6 +4,9 @@
  */
 package ai.onnxruntime.genai;
 
+/**
+ * A container of adapters.
+ */
 public final class Adapters implements AutoCloseable {
   private long nativeHandle = 0;
 
@@ -22,13 +25,13 @@ public final class Adapters implements AutoCloseable {
   }
 
   /**
-   * Load an adapter from the specified path.
+   * Loads the model adapter from the given adapter file path and adapter name.
    *
    * @param adapterFilePath The path of the adapter.
    * @param adapterName A unique user supplied adapter identifier.
    * @throws GenAIException If the call to the GenAI native API fails.
    */
-  public void loadAdapters(String adapterFilePath, String adapterName) throws GenAIException {
+  public void loadAdapter(String adapterFilePath, String adapterName) throws GenAIException {
     if (nativeHandle == 0) {
       throw new IllegalStateException("Instance has been freed and is invalid");
     }
@@ -37,12 +40,13 @@ public final class Adapters implements AutoCloseable {
   }
 
   /**
-   * Unload an adapter.
+   * Unloads the adapter with the given identifier from the previosly loaded adapters.
+   * If the adapter is not found, or if it cannot be unloaded (when it is in use), an error is returned.
    *
    * @param adapterName A unique user supplied adapter identifier.
    * @throws GenAIException If the call to the GenAI native API fails.
    */
-  public void unloadAdapters(String adapterName) throws GenAIException {
+  public void unloadAdapter(String adapterName) throws GenAIException {
     if (nativeHandle == 0) {
       throw new IllegalStateException("Instance has been freed and is invalid");
     }
--- a/src/java/src/main/java/ai/onnxruntime/genai/Adapters.java
+++ b/src/java/src/main/java/ai/onnxruntime/genai/Adapters.java
@@ -4,9 +4,7 @@
  */
 package ai.onnxruntime.genai;
 
-/**
- * A container of adapters.
- */
+/** A container of adapters. */
 public final class Adapters implements AutoCloseable {
   private long nativeHandle = 0;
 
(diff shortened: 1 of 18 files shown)


src/java/src/main/java/ai/onnxruntime/genai/NamedTensors.java
```diff
@@ -3,9 +3,18 @@
  */
 package ai.onnxruntime.genai;
 
+/**
+ * This class is an intermediate storage class that bridges the output of preprocessing and the
```
skyline75489, line 7: This is just my understanding.

src/java/src/main/java/ai/onnxruntime/genai/Adapters.java
```diff
@@ -22,13 +23,13 @@ public Adapters(Model model) throws GenAIException {
...
   /**
-   * Load an adapter from the specified path.
+   * Loads the model adapter from the given adapter file path and adapter name.
    *
    * @param adapterFilePath The path of the adapter.
    * @param adapterName A unique user supplied adapter identifier.
    * @throws GenAIException If the call to the GenAI native API fails.
    */
-  public void loadAdapters(String adapterFilePath, String adapterName) throws GenAIException {
+  public void loadAdapter(String adapterFilePath, String adapterName) throws GenAIException {
```
skyline75489, line 32: Drop the `-s` to match other languages.

src/java/src/main/java/ai/onnxruntime/genai/GenAI.java
```diff
@@ -34,7 +34,7 @@ final class GenAI {
   /** The short name of the ONNX runtime shared library */
   static final String ONNXRUNTIME_LIBRARY_NAME = "onnxruntime";
 
-  /** The value of the {@link #GENAI_NATIVE_PATH} system property */
```
skyline75489, line 37: Doxygen doesn't support `@link` unfortunately.

src/java/src/main/java/ai/onnxruntime/genai/Tensor.java
```diff
@@ -7,6 +7,7 @@
 import java.nio.ByteBuffer;
 import java.nio.ByteOrder;
 
+/** Wraps ORT native Tensor. */
```
skyline75489, line 10: In need of a better description.

src/java/src/main/java/ai/onnxruntime/genai/Model.java
```diff
@@ -27,39 +27,6 @@ public Model(Config config) throws GenAIException {
...
-   */
-  public Tokenizer createTokenizer() throws GenAIException {
-    if (nativeHandle == 0) {
-      throw new IllegalStateException("Instance has been freed and is invalid");
-    }
-
-    return new Tokenizer(this);
-  }
-
-  // NOTE: Having model.createGeneratorParams is still under discussion.
```
skyline75489, line 32: I'm guessing the discussion ended with the conclusion that neither `createTokenizer` or `createGeneratorParams` is needed.

src/java/src/main/java/ai/onnxruntime/genai/NamedTensors.java
RyanUnderhill: Yeah, the named tensors are just a way to have a list of tensors with names that match up with model input names. Perhaps:

"NamedTensors is a list of tensors with names that match up with model input names."

src/java/src/main/java/ai/onnxruntime/genai/Tensor.java
RyanUnderhill: Really just a Tensor. It might not necessarily wrap an ORT tensor in the future.

RyanUnderhill (approved):

src/java/src/main/java/ai/onnxruntime/genai/Model.java
skyline75489: Do you want to keep `createTokenizer` and `createGeneratorParams`? I don't see these two in other language bindings. besides, I don't quite get the difference between `createGeneratorParams` and `new GeneratorParams(model)`.

src/java/src/main/java/ai/onnxruntime/genai/Model.java
```diff
@@ -3,48 +3,28 @@
...
+   * Construct a Model from the given Config.
    *
-   * @return The GeneratorParams instance.
+   * @param config The config to use.
    * @throws GenAIException If the call to the GenAI native API fails.
    */
-  public GeneratorParams createGeneratorParams() throws GenAIException {
-    if (nativeHandle == 0) {
-      throw new IllegalStateException("Instance has been freed and is invalid");
-    }
```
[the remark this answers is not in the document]
skyline75489: I don't see this in C# so I don't quite understand what you mean

src/java/src/main/java/ai/onnxruntime/genai/Tensor.java
skyline75489: Should I change the description to something else?

RyanUnderhill: Maybe just "Currently wraps an ORT Tensor" since that implies it could change. But I'm not sure if we should say it's an ORT Tensor as that's an implementation detail. Up to you.

src/java/src/main/java/ai/onnxruntime/genai/Model.java
RyanUnderhill: model.createGeneratorParams and new GeneratorParams(model) is arbitrary. You should do what makes the most sense in Java.

But, we want to remove 'createTokenizer' in the future, as you can just get the tokenizer from the model internally already. You can remove createTokenizer :)

src/java/src/main/java/ai/onnxruntime/genai/Tensor.java
skyline75489: "Currently wraps an ORT Tensor" feels right to me. Thanks!

PR #1138, opened 2024-12-10
Support for Qualcomm AI Hub sliding window models

baijumeswani:
Language models sourced from [Qualcomm AI Hub](https://aihub.qualcomm.com/compute/models) have a sliding window structure. This pull-request adds support for sliding window models to the decoder-pipeline.

Working with the model sourced from the Qualcomm AI Hub, the model execution goes through two stages:

- Context (prompt) processing: The model runs through several iterations of the model processing `window_size` tokens at a time to compute the key-value cache and generate the first token. In each iteration, we slide the window (by `window_size` over the key-value cache buffer so the new key-value cache can be computed.
- Token generation: After the input prompt has been processed, the execution switches over to the token generation phase where new tokens are processed one at a time. In this phase, we slide the key-value cache buffer by 1 at each iteration.

For example, let's assume the following inputs:

- Prompt: ["What", "is", "the", "square", "root", "of", "16", "?"]
- Window size: 5
- Pad token: 0
- context length: 12

Given the prompt has 8 tokens (assuming a word is a token), to process the entire prompt, 2 iterations will be required.

- Iteration 1:
  - input_ids: [0, 0, "What", "is", "the"]
  - position_ids: [0, 0, 0, 1, 2]
  - attention_mask: [0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 1, 1]

- Iteration 2:
  - input_ids: ["square", "root", "of", "16", "?"]
  - position_ids: [3, 4, 5, 6, 7]
  - attention_mask: [[0, 0, 0, 0, 1, 1, 1, 1, 1, 1, 1, 1]

The above two iterations complete the prompt processing phase. Subsequently, the token generation phase begins. The token generation phase processes 1 token at a time.

At each iteration, the key-value cache is updated by sliding the key-value cache buffer by window size tokens during the prompt processing phase and by 1 token during the token generation phase.

This pull-request introduces the above changes to the decoder-pipeline to support the sliding window models from the Qualcomm AI Hub

In addition to the above changes, the pull request also introduces the `QNN_WITH_SHARED_MEMORY` device type so memory can be allocated using the qnn allocator. Thanks to @edgchen1 for making these changes.

Open items:
- Continuous decoding support for these model types.
- Incorrect results when the generated tokens size becomes large enough.
- Performance.
- Tests


src/models/model.cpp
```diff
@@ -500,6 +513,15 @@ void Model::CreateSessionOptionsFromConfig(const Config::SessionOptions& config_
       for (auto& option : provider_options.options) {
         opts.emplace(option.first, option.second);
       }
+
+      // TODO set device_type_ in a less hacky way.
```
edgchen1, line 517: I wrote this - any idea on how we can do this better?

src/models/threadpool.cpp
```diff
@@ -0,0 +1,20 @@
...
+
+#include "threadpool.h"
+
+namespace Generators {
+
+ThreadPool::ThreadPool(size_t num_threads) : num_threads_{num_threads} {}
+
+void ThreadPool::Compute(const std::function<void(size_t)>& func) {
+  for (size_t i = 0; i < num_threads_; ++i) {
+    threads_.emplace_back([&, i] { func(i); });
```
edgchen1, line 12: I wonder if the overhead of thread creation/joining is significant relative to the actual computation?

src/models/threadpool.cpp
```diff
@@ -0,0 +1,20 @@
...
+  }
+
+  for (auto& thread : threads_) {
+    thread.join();
```
edgchen1, line 16: should threads_ also be cleared? otherwise a later call to Compute() would call join again on an already joined thread.

src/models/input_ids.h
```diff
@@ -4,19 +4,26 @@
 
 namespace Generators {
 
-struct InputIDs {
+struct InputIDsInterface {
```
RyanUnderhill, line 7: Can we have this base class just be 'InputIDs' and then we'd have two types of them, the non sliding window and the InputIDsSlidingWindow. How do we refer to the non sliding window ones?

src/models/input_ids.h
```diff
@@ -45,4 +52,29 @@ struct InputIDs {
...
+
+  void Add() override;
+  void Update(DeviceSpan<int32_t>& next_tokens) override;
+  std::array<int64_t, 2> GetShape() const override { return shape_; }
+
+ private:
+  State& state_;
+  const Model& model_{state_.model_};
+  size_t input_index_{~0U};
+  size_t window_size_{0};
```
RyanUnderhill, line 68:
```suggestion
  size_t window_size_{};
```

src/models/input_ids.h
```diff
@@ -45,4 +52,29 @@ struct InputIDs {
...
+  size_t num_windows_{1};
+  size_t window_index_{0};
```
RyanUnderhill, line 70:
```suggestion
  size_t window_index_{};
```

src/models/kv_cache.cpp
```diff
@@ -435,4 +443,241 @@ void Cross_Cache::AddInputs() {
...
+  for (int i = 0; i < layer_count_; ++i) {
+    input_name_strings_.emplace_back(ComposeKeyValueName(model_.config_->model.decoder.inputs.past_key_names, i));
+    input_name_strings_.emplace_back(ComposeKeyValueName(model_.config_->model.decoder.inputs.past_value_names, i));
+
+    output_name_strings_.emplace_back(ComposeKeyValueName(model_.config_->model.decoder.outputs.present_key_names, i));
+    output_name_strings_.emplace_back(ComposeKeyValueName(model_.config_->model.decoder.outputs.present_value_names, i));
+  }
+
+  type_ = model_.session_info_->GetInputDataType(input_name_strings_[0]);
+  if (type_ != ONNX_TENSOR_ELEMENT_DATA_TYPE_UINT8) {
```
RyanUnderhill, line 467:
```suggestion
  if (type_ != Ort::TypeToTensorType<uint8_t>) {
```

src/models/kv_cache.h
```diff
@@ -89,4 +98,41 @@ struct Cross_Cache {
...
+  void RewindTo(size_t index) override {
+    throw std::runtime_error("SlidingWindowKeyValueCache does not support RewindTo.");
+  }
+
+ private:
+  void Slide();
+
+  State& state_;
+  const Model& model_{state_.model_};
+  int layer_count_{0};
```
RyanUnderhill, line 119:
```suggestion
  int layer_count_{};
```

src/models/kv_cache.h
```diff
@@ -89,4 +98,41 @@ struct Cross_Cache {
...
+  int window_size_{0};
```
RyanUnderhill, line 120:
```suggestion
  int window_size_{};
```

src/models/kv_cache.h
```diff
@@ -89,4 +98,41 @@ struct Cross_Cache {
...
+  size_t num_windows_{1};
+  size_t window_index_{0};
```
RyanUnderhill, line 122:
```suggestion
  size_t window_index_{};
```

src/models/model.cpp
RyanUnderhill: Yeah, my PR has this function a lot cleaner. When I merge with this, it's really easy to add QnnHtpShared

src/models/position_inputs.h
```diff
@@ -86,4 +93,40 @@ struct PositionInputs {
...
+  std::unique_ptr<OrtValue> position_ids_;
+  std::array<int64_t, 2> attention_mask_shape_{};
+  ONNXTensorElementDataType attention_mask_type_{};
+  std::unique_ptr<OrtValue> attention_mask_;
+  size_t attention_mask_backward_offset_{~0U};
+
+  size_t attention_mask_index_{~0U};
+  size_t position_ids_index_{~0U};
+
+  size_t window_size_{0};
```
RyanUnderhill, line 125:
```suggestion
  size_t window_size_{};
```

src/models/position_inputs.h
```diff
@@ -86,4 +93,40 @@ struct PositionInputs {
...
+  size_t num_windows_{1};
+  size_t window_index_{0};
```
RyanUnderhill, line 127:
```suggestion
  size_t window_index_{};
```

src/models/threadpool.cpp
baijumeswani: I am quite certain it is significant. But when I ran some tests on my windows arm64 machine, I saw a speedup (~0.8 tokens per second) relative to not running with the threads.
This implementation is quite rudimentary and will need to be enhanced. It would be great if we can leverage the ort threadpool from across the C API interface.

cmake/cxx_standard.cmake
```diff
@@ -12,3 +12,7 @@ else ()
     message("Test is using C++20")
     set(CMAKE_CXX_STANDARD 20)
 endif ()
+
+if (CMAKE_SYSTEM_NAME STREQUAL "Linux")
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -pthread")
```
edgchen1, line 17: CMake has a FindThreads module that might be more robust. we could also scope the `Threads::Threads` dependency to a particular target.
https://stackoverflow.com/a/29871891
https://cmake.org/cmake/help/latest/module/FindThreads.html

src/models/position_inputs.cpp
```diff
@@ -458,4 +457,149 @@ void PositionInputs::RewindMask(size_t index) {
...
+
+    if (has_mask_input_) {
+      attention_mask_ = OrtValue::CreateTensor(model_.allocator_cpu_, attention_mask_shape_, attention_mask_type_);
+
+      // next_tokens will always be padded so that it's size is a multiple of window_size_
+      // next_tokens -> [0, a, b, c, d, e]
+      // window_size = 3, num_windows = 2, pad_token = 0
+      // window_index = 0, attention_mask_ -> ([0] * context_length - window_size_) + [0, 1, 1]
+      auto* attention_mask_data = attention_mask_->GetTensorMutableData<int32_t>();
+      std::fill(attention_mask_data, attention_mask_data + attention_mask_shape_[1], 0);
```
edgchen1, line 531: nit: does this only need to fill `attention_mask_shape_[1] - window_size_` zeros? it looks like the next for loop will handle the rest.

src/config.h
```diff
@@ -107,6 +107,12 @@ struct Config {
       int num_hidden_layers{};
       int head_size{};
 
+      struct SlidingWindow {
```
edgchen1, line 110: can we document the new config values somewhere?

src/generators.cpp
```diff
@@ -266,15 +275,23 @@ Generator::Generator(const Model& model, const GeneratorParams& params) : model_
...
+  if (model_->config_->model.decoder.sliding_window.has_value()) {
+    // If the model has a sliding window, pad the input_ids to the next multiple of the window size
+    // so that the input_ids can be divided into window size chunks.
+    const auto window_size = model_->config_->model.decoder.sliding_window->window_size;
+    input_ids_size = ((input_ids.size() + window_size - 1) / window_size) * window_size;
+  }
+  auto input_ids_device = state_->params_->p_device->Allocate<int32_t>(input_ids_size);
   auto cpu_span = input_ids_device.CpuSpan();
-  std::copy(input_ids.begin(), input_ids.end(), cpu_span.begin());
+  std::fill_n(cpu_span.begin(), input_ids_size, model_->config_->model.pad_token_id);
```
edgchen1, line 288: nit: filling more than we need to
```suggestion
  std::fill_n(cpu_span.begin(), input_ids_size - input_ids.size(), model_->config_->model.pad_token_id);
```
also maybe rename `input_ids_size` to `padded_input_ids_size`

src/ort_genai_c.cpp
```diff
@@ -292,15 +292,15 @@ OgaResult* OGA_API_CALL OgaGenerator_AppendTokenSequences(OgaGenerator* oga_gene
   }
 
   auto input_ids = Generators::PadInputs(span_sequences, generator.model_->config_->model.pad_token_id);
-  generator.AppendTokens(input_ids);
+  generator.AppendTokens(Generators::cpu_span<const int32_t>(input_ids.data(), input_ids.size()));
```
edgchen1, line 295: do we need an explicit conversion from `std::vector<int32_t>` to `cpu_span<const int32_t>`? might be nicer if it was automatic.

src/models/input_ids.cpp
```diff
@@ -5,7 +5,7 @@
 
 namespace Generators {
 
-InputIDs::InputIDs(State& state)
+InputIDsDefault::InputIDsDefault(State& state)
```
kunal-vaishnavi, line 8: Do we need to add `Default` to the end of the names? There's already separate class names for these models (e.g. `WindowedInputIDs`).

src/generators.cpp
```diff
@@ -45,8 +46,14 @@ void OnCudaError(cudaError_t error) { assert(false); }
 
 static bool _ = (Ort::InitApi(), false);
 
+static OrtLoggingLevel GetDefaultOrtLoggingLevel() {
+  bool ort_verbose_logging = false;
+  GetEnvironmentVariable("ORTGENAI_ORT_VERBOSE_LOGGING", ort_verbose_logging);
```
kunal-vaishnavi, line 51: Can the existing ORT environment variable names be used or do the environment variable names have to be ORT GenAI specific?

src/generators.h
```diff
@@ -60,6 +60,7 @@ enum struct DeviceType {
   CUDA,
   DML,
   WEBGPU,
+  QNN_WITH_SHARED_MEMORY,
```
kunal-vaishnavi, line 63: Will there be a future distinction between `QNN` and `QNN_WITH_SHARED_MEMORY`? If not, should we consider renaming this to just `QNN`?

baijumeswani: I think we could. @edgchen1 should we rename to QNN?

src/models/input_ids.cpp
baijumeswani: There is `InputIds` (base class), and `InputIdsDefault` and `WindowedInputIds` derived classes.

I initially had `InputIdsInterface` as the base and `InputIds` and `SlidingWindowInputIds` as derived.

Do you have any suggestions on the names?

src/models/kv_cache.h
```diff
@@ -4,8 +4,18 @@
...
+  virtual ~KeyValueCache() = default;
+  virtual void Add() = 0;
+  virtual void AddEncoder() = 0;
+  virtual void Update(DeviceSpan<int32_t> beam_indices, int total_length) = 0;
+  virtual void RewindTo(size_t index) = 0;
+
+  static bool IsCacheNeeded(const Model& model);
+};
+
+struct KeyValueCacheDefault_Combined {
```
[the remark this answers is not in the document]
baijumeswani: Their usage is limited to files that are not touched by this PR and so I didn't update them. But I'll do that now since it might get ugly to do this later.

src/models/model.cpp
```diff
@@ -288,29 +288,42 @@ Model::Model(std::unique_ptr<Config> config) : config_{std::move(config)} {
...
-    dml_owned_allocator_ = Ort::Allocator::Create(session, *memory_info_device_);
-    allocator_device_ = dml_owned_allocator_.get();
+    owned_allocator_device_ = Ort::Allocator::Create(session, *memory_info_device_);
+    allocator_device_ = owned_allocator_device_.get();
+    allocator_kvcache_ = allocator_device_;
   }
 #endif
-  allocator_kvcache_ = allocator_device_;
+
 #if USE_WEBGPU
```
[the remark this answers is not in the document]
baijumeswani: I think Ryan has made changes to remove majority of the #ifdefs on device types in his branch.

src/models/kv_cache.h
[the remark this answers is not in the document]
baijumeswani: Adding to the above:

- KeyValueCacheDefault_Combined is only used in the gpt model type. This type of key-value cache should be deprecated going forward. Not sure when we will remove the code references, but we should not be using that. But, I will make the changes to that struct to inherit from KeyValuCache
- Cross_Cache is only used in whisper. This type of key-value cache is not used as a standalone cache and is a supplemental cache for the whisper model needed in some scenarios. It does not behave like the other KeyValue caches. I don't think I will update this one since it might have some implications on how whisper model uses it.

src/generators.h
edgchen1: sure

src/generators.cpp
edgchen1: I'm not aware of an existing general ORT environment variable that controls the log level. if there is one, maybe we don't need this at all.

src/models/input_ids.cpp
edgchen1: one suggestion/opinion is that `DefaultInputIDs` sounds more consistent with `WindowedInputIDs`.

edgchen1: similar suggestion for other names
`PositionIDsDefault` -> `DefaultPositionIDs`
`KeyValueCacheDefault` -> `DefaultKeyValueCache`
`KeyValueCacheDefault_Combined` -> `CombinedKeyValueCache`

edgchen1 (approved):

kunal-vaishnavi (approved):

baijumeswani:
Thank you for the review. :)

PR #1154, opened 2024-12-17
Improve C# documentation

skyline75489:
Added missing C# documentation according to the Javadoc (#1136).

Internal work item: https://task.ms/aii/53908


src/csharp/Images.cs
```diff
@@ -18,7 +21,7 @@ private Images(IntPtr imagesHandle)
 
         internal IntPtr Handle { get { return _imagesHandle; } }
 
-        public static Images Load(string[] imagePaths)
```
skyline75489, line 21: This is bothering me. Why do we use a static method `Images.Load()` instead of `new Images(string[] imagePaths)` just like in java:

https://github.com/microsoft/onnxruntime-genai/blob/d12927416fa791802ce73fa795129f1e14f3d42a/src/java/src/main/java/ai/onnxruntime/genai/Images.java#L16

src/csharp/Utils.cs
```diff
@@ -36,7 +39,10 @@ protected virtual void Dispose(bool disposing)
         }
     }
 
-    public class Utils
+    /// <summary>
+    /// Wraps some utility functions that are available globally
+    /// <summary>
+    internal class Utils
```
skyline75489, line 45: In Java there's a `GenAI` class that handles the global configurations:

https://github.com/microsoft/onnxruntime-genai/blob/d12927416fa791802ce73fa795129f1e14f3d42a/src/java/src/main/java/ai/onnxruntime/genai/GenAI.java#L19

Should we do the same in C#?

natke: @baijumeswani

natke (approved):

src/csharp/Images.cs
natke: Created an internal workitem to address this

src/csharp/Utils.cs
natke: Created an internal workitem to do this

PR #1155, opened 2024-12-17
Switch to self-hosted pool to fix Android CI

skyline75489:
* Switch to internal pool to avoid being hit by GHA image updates.
* Add script to install Android emulator if needed.

### DIFF
--- a/.github/workflows/android-build.yml
+++ b/.github/workflows/android-build.yml
@@ -28,6 +28,10 @@ jobs:
         with:
           submodules: true
 
+      - uses: actions/setup-dotnet@v4
+        with:
+          dotnet-version: '8.0.x'
+
       - name: Setup Java 17
         uses: actions/setup-java@v4
         with:
@@ -53,10 +57,6 @@ jobs:
       - name: Install jq
         run: |
           sudo apt-get install jq
-
-      - uses: actions/setup-dotnet@v4
-        with:
-          dotnet-version: '8.0.x'
 
       - name: Checkout OnnxRuntime GenAI repo
         uses: actions/checkout@v4
--- a/.github/workflows/android-build.yml
+++ b/.github/workflows/android-build.yml
@@ -40,7 +40,7 @@ jobs:
         with:
           gradle-version: '8.6'
 
-      # Check the NDK that we're using
+      # Check the Android SDK and NDK that we're using
       - name: Check Android SDK and NDK
         run: |
           set -e -x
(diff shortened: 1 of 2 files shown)


.github/workflows/android-build.yml
```diff
@@ -21,7 +21,7 @@ jobs:
     #       it doesn't work on macos-14.
     #         HVF error: HV_UNSUPPORTED
     #       it works on macos-13 but with macos-15 being released soon that isn't a long term solution.
-    runs-on: ubuntu-latest
+    runs-on: [ "self-hosted", "1ES.Pool=onnxruntime-genai-Ubuntu2204-AMD-CPU" ]
```
skyline75489, line 24: This pool is also used in Linux x64 CI

.github/workflows/android-build.yml
```diff
@@ -40,48 +40,38 @@ jobs:
...
           echo "ANDROID_NDK_HOME=$ANDROID_NDK_HOME"
           echo "ANDROID_NDK_LATEST_HOME=$ANDROID_NDK_LATEST_HOME"
           ls -l $ANDROID_HOME/ndk
 
+      - name: Check if emulator is installed and install it if needed
+        run: |
+          if [ -d "${ANDROID_SDK_ROOT}/emulator" ]; then
+              echo "${ANDROID_SDK_ROOT}/emulator exists"
+          else
+              ${ANDROID_SDK_ROOT}/cmdline-tools/latest/bin/sdkmanager --install "emulator"
```
skyline75489, line 62: This is from https://github.com/microsoft/onnxruntime/blob/main/tools/ci_build/github/azure-pipelines/templates/use-android-emulator.yml

.github/workflows/android-build.yml
```diff
@@ -40,48 +40,38 @@ jobs:
...
+          fi
+
       # Needed for linux
       - name: Install jq
         run: |
           sudo apt-get install jq
 
-      - uses: actions/setup-dotnet@v4
```
skyline75489, line 57: For some unknown reason, `dotnet install` failed with permission denied.

See https://github.com/microsoft/onnxruntime-genai/actions/runs/12385746487/job/34573153027

.github/workflows/android-build.yml
```diff
@@ -40,48 +40,38 @@ jobs:
...
-        with:
-          dotnet-version: '8.0.x'
-
       - name: Checkout OnnxRuntime GenAI repo
         uses: actions/checkout@v4
         with:
           submodules: true
 
-      - name: Get the Latest OnnxRuntime Nightly Version
```
skyline75489, line 66: Removing these steps has no effect on either building or testing GenAI on Android platform.

For building, without explicitly pass `--ort_home`, the version in [ortlib.cmake](https://github.com/microsoft/onnxruntime-genai/blob/main/cmake/ortlib.cmake) will be used.

For testing, the latest release version of ORT is used:

https://github.com/microsoft/onnxruntime-genai/blob/c77e525243dc233b88edaefabbfc5f19facf4044/src/java/src/test/android/app/build.gradle#L44

I remember @baijumeswani  mentions that the nightly version of ORT should be used for testing, but I'm not suer that's available for Android.

baijumeswani (approved):

baijumeswani:
Thanks 👍🏽

PR #1161, opened 2024-12-20
Recompute KV cache for Phi3 when switching from short to long factor

ajindal1:
Recompute KV cache for Phi3 when switching from short to long factor.

Verified that this PR fixes the issue for:
1. Phi3.5 mini
2. Phi3 mini 128K
3. Phi3 small
4. Phi3 medium

### DIFF
--- a/src/generators.cpp
+++ b/src/generators.cpp
@@ -357,8 +357,7 @@ void Generator::GenerateNextToken() {
   // Phi3 model switches from short factor to long factor at 4097 (original_max_position_embeddings+1) token, needs Recomputation of Position IDs and KV Cache
   // at this stage which is achieved by rewinding to zero and appending the current sequence
   if (model_->config_->model.type == "phi3" && search_->params_->search.batch_size == 1 && search_->params_->search.num_beams == 1) {
-    if (search_->GetSequenceLength() == 4097 && first_switch) {
-      first_switch = false;
+    if (search_->GetSequenceLength() == 4097) {
       auto current_seq = cpu_span<int32_t>(GetSequence(0).CpuSpan());
       RewindToLength(0);
       AppendTokens(current_seq);
--- a/src/generators.cpp
+++ b/src/generators.cpp
@@ -356,8 +356,13 @@ void Generator::GenerateNextToken() {
   // TODO: Extend the solution to make it work for batch size > 1 and num beams > 1
   // Phi3 model switches from short factor to long factor at 4097 (original_max_position_embeddings+1) token, needs Recomputation of Position IDs and KV Cache
   // at this stage which is achieved by rewinding to zero and appending the current sequence
-  if (model_->config_->model.type == "phi3" && search_->params_->search.batch_size == 1 && search_->params_->search.num_beams == 1) {
-    if (search_->GetSequenceLength() == 4097) {
+  if (search_->params_->search.batch_size == 1 && search_->params_->search.num_beams == 1) {
+    if ((search_->GetSequenceLength() == 4097) && (model_->config_->model.type == "phi3" || model_->config_->model.type == "phimoe")) {
+      auto current_seq = cpu_span<int32_t>(GetSequence(0).CpuSpan());
+      RewindToLength(0);
+      AppendTokens(current_seq);
+    }
+    if ((search_->GetSequenceLength() == 8197) && (model_->config_->model.type == "phi3small")) {
       auto current_seq = cpu_span<int32_t>(GetSequence(0).CpuSpan());
       RewindToLength(0);
       AppendTokens(current_seq);
--- a/src/generators.cpp
+++ b/src/generators.cpp
@@ -356,7 +356,7 @@ void Generator::GenerateNextToken() {
   // TODO: Extend the solution to make it work for batch size > 1 and num beams > 1
   // Phi3 model switches from short factor to long factor at 4097 (original_max_position_embeddings+1) token, needs Recomputation of Position IDs and KV Cache
   // at this stage which is achieved by rewinding to zero and appending the current sequence
-  if (model_->config_->model.type == "phi3" && search_->params_->search.batch_size == 1 && params.search.num_beams == 1) {
+  if (model_->config_->model.type == "phi3" && search_->params_->search.batch_size == 1 && search_->params_->search.num_beams == 1) {
     if (search_->GetSequenceLength() == 4097 && first_switch) {
       first_switch = false;
       auto current_seq = cpu_span<int32_t>(GetSequence(0).CpuSpan());
--- a/src/generators.cpp
+++ b/src/generators.cpp
@@ -353,17 +353,15 @@ void Generator::GenerateNextToken() {
   if (search_->GetSequenceLength() == 0 && !computed_logits_)
     throw std::runtime_error("GenerateNextToken called with no prior state. Please call AppendTokens, SetLogits, or params.SetInputs before calling GenerateNextToken.");
 
-  // TODO: Extend the solution to make it work for batch size > 1 and num beams > 1
+  // TODO: Extend the solution to make it work for batch size > 1, num beams > 1, multimodal and DML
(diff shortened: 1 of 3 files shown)


src/generators.cpp
```diff
@@ -352,6 +352,19 @@ void Generator::GenerateNextToken() {
   ThrowErrorIfSessionTerminated(state_->session_terminated_);
   if (search_->GetSequenceLength() == 0 && !computed_logits_)
     throw std::runtime_error("GenerateNextToken called with no prior state. Please call AppendTokens, SetLogits, or params.SetInputs before calling GenerateNextToken.");
+
+  // TODO: Extend the solution to make it work for batch size > 1 and num beams > 1
+  // Phi3 model switches from short factor to long factor at 4097 (original_max_position_embeddings+1) token, needs Recomputation of Position IDs and KV Cache
+  // at this stage which is achieved by rewinding to zero and appending the current sequence
+  if (model_->config_->model.type == "phi3" && search_->params_->search.batch_size == 1 && search_->params_->search.num_beams == 1) {
```
kunal-vaishnavi, line 359: The model types should be `phi3`, `phi3small`, `phi3v`, and `phimoe`. For `phi3v`, rewinding is not currently supported.

src/generators.cpp
```diff
@@ -352,6 +352,19 @@ void Generator::GenerateNextToken() {
...
+    if (search_->GetSequenceLength() == 4097 && first_switch) {
```
kunal-vaishnavi, line 360: For `phi3small`, the switch value is at 8K and not 4K.

baijumeswani, line 360: What is the purpose of the variable `first_switch`?

src/generators.cpp
```diff
@@ -352,6 +352,19 @@ void Generator::GenerateNextToken() {
...
+      first_switch = false;
+      auto current_seq = cpu_span<int32_t>(GetSequence(0).CpuSpan());
```
baijumeswani, line 362: Do you intend to copy the span over to CPU? You might want to use `CopyDeviceToCpu` instead of `CpuSpan`.

src/generators.cpp
```diff
@@ -352,6 +352,19 @@ void Generator::GenerateNextToken() {
...
+      RewindToLength(0);
+      AppendTokens(current_seq);
```
baijumeswani, line 364: Could you also add a comment highlighting that this solution works only for scenarios that support continuous decoding.

- [x] Batch size == 1
- [x] Num beams == 1
- [ ] Model must be a decoder model (not a multimodal model)
- [ ] EP in {CPU, CUDA}. DML won't work.

src/generators.cpp
```diff
@@ -352,6 +352,19 @@ void Generator::GenerateNextToken() {
...
     throw std::runtime_error("GenerateNextToken called with no prior state. Please call AppendTokens, SetLogits, or params.SetInputs before calling GenerateNextToken.");
+
+  // TODO: Extend the solution to make it work for batch size > 1 and num beams > 1
+  // Phi3 model switches from short factor to long factor at 4097 (original_max_position_embeddings+1) token, needs Recomputation of Position IDs and KV Cache
+  // at this stage which is achieved by rewinding to zero and appending the current sequence
+  if (model_->config_->model.type == "phi3" && search_->params_->search.batch_size == 1 && search_->params_->search.num_beams == 1) {
+    if (search_->GetSequenceLength() == 4097 && first_switch) {
+      first_switch = false;
+      auto current_seq = cpu_span<int32_t>(GetSequence(0).CpuSpan());
+      RewindToLength(0);
```
baijumeswani, line 363: Do you think a more general purpose solution might be to have a new API that resets the generator?

```
Reset()
```

Then we don't need to limit ourselves to only supporting scenarios that support RewindToLength(0) (i.e. continuous decoding).

Maybe we can do it in another PR.

src/generators.cpp
baijumeswani: Could we record `original_max_position_embeddings` in the config and avoid hardcoding this value here?

ajindal1: It is not needed anymore, thanks for catching this. Previously, I was doing this in COmputeLogits, where it became a circular loop as AppendTokens calls ComputeLogits again so had added this.

src/generators.cpp
ajindal1: Thanks, updated the model names.

src/generators.cpp
ajindal1: From what I understand is that we cannot add this to the existing models since we don't have `original_max_position_embeddings` in the config. We can add `original_max_position_embeddings` in the config going forward for new models and this can avoid the issue of hardcoding in future.

src/generators.cpp
ajindal1: Yes, replaced it, thanks.

src/generators.cpp
ajindal1: I have added a similar comment, thanks.

src/generators.cpp
ajindal1: Yes, that is a great idea. It might be better to do it in a separate PR.

src/generators.cpp
```diff
@@ -352,6 +352,21 @@ void Generator::GenerateNextToken() {
   ThrowErrorIfSessionTerminated(state_->session_terminated_);
   if (search_->GetSequenceLength() == 0 && !computed_logits_)
     throw std::runtime_error("GenerateNextToken called with no prior state. Please call AppendTokens, SetLogits, or params.SetInputs before calling GenerateNextToken.");
+
+  // TODO: Extend the solution to make it work for batch size > 1, num beams > 1, multimodal and DML
+  // Phi3 model switches from short factor to long factor at 4097 (original_max_position_embeddings+1) token, needs Recomputation of Position IDs and KV Cache
+  // at this stage which is achieved by rewinding to zero and appending the current sequence
+  // Scenarios where this solution works: Batch size = 1, Num beams = 1, decoder model, EP is either CPU or CUDA
+  // Scenarios where it doesn't work: Batch size > 1 OR Num beams > 1 OR Multimodal model (like phi3 vision) OR EP is DML
+  if (search_->params_->search.batch_size == 1 && search_->params_->search.num_beams == 1) {
```
baijumeswani, line 361:
```suggestion
  if (search_->params_->BatchBeamSize() == 1) {
```

src/generators.cpp
```diff
@@ -352,6 +352,21 @@ void Generator::GenerateNextToken() {
...
+    if (((search_->GetSequenceLength() == 4097) && (model_->config_->model.type == "phi3" || model_->config_->model.type == "phimoe")) || ((search_->GetSequenceLength() == 8197) && (model_->config_->model.type == "phi3small"))) {
+      // auto current_seq = cpu_span<int32_t>(GetSequence(0).CpuSpan());
```
baijumeswani, line 363:
```suggestion
```

baijumeswani (approved):

src/generators.cpp
ajindal1: Added in the genai_config creation code, so we don't need to hardcode for newer models.

src/python/py/models/builder.py
```diff
@@ -347,6 +347,7 @@ def make_genai_config(self, model_name_or_path, extra_kwargs, out_dir):
                     "num_key_value_heads": self.num_kv_heads,
                 },
                 "eos_token_id": config.eos_token_id,
+                "original_context_length": self.original_context_length,
```
baijumeswani, line 350: Should we call it `original_max_position_embeddings` in out genai config file as well?

ajindal1: Not sure why we changed the name to original_context_length, it seems to be used in more scenarios than just short and long factor. Maybe @kunal-vaishnavi can provide more context?

ajindal1: I am reverting this now, will discuss this with Kunal once he is back. Will merge the PR for now.

PR #1162, opened 2024-12-22
[Model builder] Add option to exclude cache in inputs and outputs

xenova:
In certain cases (e.g., single-round conversations), it is not necessary to require `past_key_values` as inputs and `present` outputs, like with https://huggingface.co/livekit/turn-detector (and its usage [here](https://github.com/livekit/agents/blob/main/livekit-plugins/livekit-plugins-turn-detector/livekit/plugins/turn_detector/eou.py)).

So, this PR adds an option to exclude these inputs and outputs from the graph.

Example usage:
```
python builder.py -m livekit/turn-detector -o converted -e cpu -p fp32 --extra_options exclude_cache=True
```

Output graph signature:
image

For comparison, graph signature w/ cache IO:
image

### DIFF
--- a/src/python/py/models/builder.py
+++ b/src/python/py/models/builder.py
@@ -111,6 +111,8 @@ class Model:
         elif self.include_hidden_states:
             self.output_names = ["hidden_states"] + self.output_names
 
+        self.exclude_cache = "exclude_cache" in extra_options
+
         # Store names of nodes already created
         self.node_names = set()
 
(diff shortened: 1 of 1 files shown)


xenova:
@microsoft-github-policy-service agree [company="{Hugging Face}"]

xenova:
@microsoft-github-policy-service agree company="Hugging Face"

xenova:
This modification only works for models w/ GQA. Maybe someone with a bit more experience with the model builder could help get it working for models w/ MHA? 😇

ambroser53:
Does this give a memory overhead improvement?

xenova:
> Does this give a memory overhead improvement?

Indeed - not having to store and pass these values back has improved execution time in my tests.

src/python/py/models/builder.py
```diff
@@ -3267,6 +3275,8 @@ def get_args():
                 exclude_lm_head = Remove language modeling head from your ONNX model.
                     Use this option when you want to remove the language modeling head from within your ONNX model.
                     Instead of `logits`, you will have `hidden_states` as the output to your ONNX model.
+                exclude_cache = Remove cache inputs and outputs from your ONNX model.
```
kunal-vaishnavi, line 3278:
```suggestion
                exclude_kv_cache = Remove KV cache inputs and outputs from your ONNX model.
```

src/python/py/models/builder.py
```diff
@@ -3267,6 +3275,8 @@ def get_args():
...
+                    Use this option when you want to remove the `past_key_values` inputs and `present` outputs from within your ONNX model.
```
kunal-vaishnavi, line 3279:
```suggestion
                    Use this option when you want to remove the `past_key_values` inputs and `present` outputs from within your ONNX model.
                    Note that this should be used when you want to run ONNX models with ONNX Runtime only. ONNX Runtime GenAI requires the KV cache inputs and outputs for inference.
```

src/python/py/models/builder.py
```diff
@@ -111,6 +111,8 @@ def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
         elif self.include_hidden_states:
             self.output_names = ["hidden_states"] + self.output_names
 
+        self.exclude_cache = "exclude_cache" in extra_options
```
kunal-vaishnavi, line 114:
```suggestion
        self.exclude_cache = extra_options.get("exclude_cache", False)
```

src/python/py/models/builder.py
```diff
@@ -1536,15 +1539,20 @@ def make_attention(self, layer_id, attention, root_input, **kwargs):
...
+            past_v = f"past_key_values.{layer_id}.value"
+            present_k = f"present.{layer_id}.key"
+            present_v = f"present.{layer_id}.value"
+            if self.num_attn_heads != self.num_kv_heads and self.attention_attrs["op_type"] == "MultiHeadAttention":
+                self.attention_attrs["k_path"] = self.make_repeat_kv(layer_id, root_input=self.attention_attrs["k_path"], past_kv=past_k, present_kv=present_k)
+                self.attention_attrs["v_path"] = self.make_repeat_kv(layer_id, root_input=self.attention_attrs["v_path"], past_kv=past_v, present_kv=present_v)
+                past_k, past_v, present_k, present_v = "", "", "", ""
+        else:
+            past_k, past_v = "", ""
+            present_k = f"present.{layer_id}.key"
```
kunal-vaishnavi, line 1554: Shouldn't `present_k` and `present_v` be empty strings since the KV cache inputs and outputs are not in the model?

kunal-vaishnavi, line 1554: I think `present_k` and `present_v` should be empty strings since the KV cache inputs and outputs are not in the model.

kunal-vaishnavi:
Thanks for the contribution! Can you also update the following places?

1. Add the option in the validation of the boolean extra options

https://github.com/microsoft/onnxruntime-genai/blob/44e541e80393be3a84fca5bb7756a900ed1c4c47/src/python/py/models/builder.py#L3133-L3141

2. Add the option in the [README](https://github.com/microsoft/onnxruntime-genai/blob/main/src/python/py/models/README.md) and provide a usage example

For `MultiHeadAttention` where the repeat KV operation occurs, you may need to remove the `Concat` node that combines `past_kv` and `curr_kv` to produce `present_kv`.

https://github.com/microsoft/onnxruntime-genai/blob/44e541e80393be3a84fca5bb7756a900ed1c4c47/src/python/py/models/builder.py#L1287-L1304

xenova:
Update: This is a very niche use-case (for cases where only a single forward pass is needed; i.e., no generation), and after further consideration, doesn't make much sense to be included in the model builder script. If someone wants to achieve the same behaviour, they can simply remove these nodes in the model produced using the current process 👍

PR #1167, opened 2025-01-07
Allow some known extra inputs in the model

baijumeswani:
This pull-request adds some extra inputs to a registry, so users need not be aware of such extra inputs.

In particular, this pull-request adds the `num_logits_to_keep` input to the registry.

Models from huggingface/transformers that are exported using `torch.onnx.export`, can have an extra input called `num_logits_to_keep` that should be default be set to `0`. Users should not need to know the input is required because it is an optimization in the torch model that does not impact the onnx model.

Sometimes, the exporter will rename the input to `onnx::Neg_xyz`. This pull-request also handles such scenarios.

### DIFF
--- a/src/models/extra_inputs.cpp
+++ b/src/models/extra_inputs.cpp
@@ -5,6 +5,46 @@
 
 namespace Generators {
 
+PresetExtraInputs::PresetExtraInputs(State& state)
+    : state_(state),
+      registry_{
+          {"num_logits_to_keep", [&state = state_]() -> std::unique_ptr<OrtValue> {
+             std::vector<int64_t> shape{1};
+             auto num_logits_to_keep = OrtValue::CreateTensor(state.model_.allocator_cpu_, shape, ONNX_TENSOR_ELEMENT_DATA_TYPE_INT64);
+             *num_logits_to_keep->GetTensorMutableData<int32_t>() = 0;
+             return num_logits_to_keep;
+           }}} {}
+
+void PresetExtraInputs::Add() {
+  const auto input_names_vector = state_.model_.session_info_->GetInputNames();
+  const std::unordered_set<std::string> input_names(state_.input_names_.begin(), state_.input_names_.end());
+  std::vector<std::string> unclaimed_input_names;
+  // Add any model input for which we don't have a corresponding input in the state to the unclaimed_input_names
+  for (const auto& input_name : input_names_vector) {
+    if (input_names.find(input_name) == input_names.end()) {
+      unclaimed_input_names.push_back(input_name);
+    }
+  }
+
+  // Try to claim the unclaimed inputs from the registry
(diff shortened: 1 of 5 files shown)


src/models/extra_inputs.cpp
```diff
@@ -5,6 +5,46 @@
 
 namespace Generators {
 
+PresetExtraInputs::PresetExtraInputs(State& state)
+    : state_(state),
+      registry_{
+          {"num_logits_to_keep", [&state = state_]() -> std::unique_ptr<OrtValue> {
+             std::vector<int64_t> shape{1};
+             auto num_logits_to_keep = OrtValue::CreateTensor(state.model_.allocator_cpu_, shape, ONNX_TENSOR_ELEMENT_DATA_TYPE_INT64);
```
RyanUnderhill, line 13:
```suggestion
             auto num_logits_to_keep = OrtValue::CreateTensor<int64_t>(state.model_.allocator_cpu_, shape);
```

src/models/extra_inputs.cpp
```diff
@@ -5,6 +5,46 @@
...
+             *num_logits_to_keep->GetTensorMutableData<int32_t>() = 0;
```
RyanUnderhill, line 14: The tensor is created as int64_t, but this sets it as int32_t?

src/models/extra_inputs.cpp
```diff
@@ -5,6 +5,46 @@
...
+             return num_logits_to_keep;
+           }}} {}
+
+void PresetExtraInputs::Add() {
+  const auto input_names_vector = state_.model_.session_info_->GetInputNames();
```
RyanUnderhill, line 19: Would it be simpler to just do a 'session_info_->HasInput("num_logits_to_keep")' and add it? There's already a special case for onnx::Neg_ prefixes.

src/models/extra_inputs.cpp
```diff
@@ -5,6 +5,46 @@
 
 namespace Generators {
 
+PresetExtraInputs::PresetExtraInputs(State& state)
+    : state_(state),
+      registry_{
```
RyanUnderhill, line 10: Is the registry always the same? Should this be a global that's created once vs per state?

baijumeswani: I think creating the registry is inexpensive and creating one over and over again per State should be ok for now.

I can make it global but would like to avoid making things global if possible.

src/models/extra_inputs.cpp
baijumeswani: I think the logic currently added is more general purpose since it checks for all inputs that have not been populated already. For the next `preset extra input`, we would only need to add it to the registry.

Also, I don't want to add num_logits_to_keep if the user has supplied this input using params.set_input.

RyanUnderhill (approved):

PR #1190, opened 2025-01-16
Refactor providers into separate libraries

RyanUnderhill:
This removes most of the #if USE_CUDA and #if USE_DML blocks for the model handling code. Device memory management is also handled through the DeviceSpan structure and now all data copying is done in a device independent manner.

It's a huge change, and there will be some rough edges when submitted. Goal is to unblock other people needing the changes and then to make larger improvements in future prs.


src/cuda/interface.cpp
```diff
@@ -66,37 +45,47 @@ struct GpuMemory final : DeviceBuffer {
 
   void CopyDeviceToCpu() override {
     AllocateCpu();
-    ::cudaMemcpy(p_cpu_, p_device_, size_in_bytes_, ::cudaMemcpyDeviceToHost);
```
aciddelgado, line 69: Why make this change? just curious

src/cuda/interface.cpp
```diff
@@ -66,37 +45,47 @@ struct GpuMemory final : DeviceBuffer {
...
+    ::cudaMemcpyAsync(p_cpu_, p_device_, size_in_bytes_, ::cudaMemcpyDeviceToHost, GetStream());
+    ::cudaStreamSynchronize(GetStream());
   }
 
   void CopyCpuToDevice() override {
     assert(p_cpu_);
-    ::cudaMemcpy(p_device_, p_cpu_, size_in_bytes_, ::cudaMemcpyHostToDevice);
```
aciddelgado, line 74: Why would making this async not cause any problems?

src/cuda/interface.cpp
```diff
@@ -113,95 +102,79 @@ struct CudaInterfaceImpl : CudaInterface {
   }
 
   void Synchronize() override {
-    ::cudaStreamSynchronize(cuda_stream_.get());
+    ::cudaStreamSynchronize(GetStream());
   }
 
-  cudaStream_t GetCudaStream() override {
-    return cuda_stream_.get();
+  void* GetCudaStream() override {
```
aciddelgado, line 108: Why not cudaStream_t?

src/cuda/interface.cpp
```diff
@@ -113,95 +102,79 @@ struct CudaInterfaceImpl : CudaInterface {
...
 
-  void Fp32ToFp16(const float* input, uint16_t* output, int count, cudaStream_t stream) override {
-    cuda::LaunchFp32ToFp16(input, output, count, stream);
-  }
-
-  void LaunchExpandAndInt32ToInt64(const int32_t* src, int64_t* dst, int num_beams, int batch_size, int sequence_length, cudaStream_t stream) override {
-    cuda::LaunchExpandAndInt32ToInt64(src, dst, num_beams, batch_size, sequence_length, stream);
-  }
-
-  void LaunchExpand(const int32_t* src, int32_t* dst, int num_beams, int batch_size, int sequence_length, cudaStream_t stream) override {
```
aciddelgado, line 139: Where does Expand functionality go?

src/dml/interface.cpp
```diff
@@ -0,0 +1,227 @@
...
+#include "dml_provider_factory.h"
+#include "../dml/dml_helpers.h"
+#include "../dml/dml_execution_context.h"
+#include "../dml/dml_pooled_upload_heap.h"
+#include "../dml/dml_readback_heap.h"
+
+std::string CurrentModulePath();
+
+namespace Generators {
+namespace Dml {  // If this was in a shared library it wouldn't need to be in its own namespace
```
aciddelgado, line 20: Why isn't the CUDA one in it's own namespace? If we build with both DML and CUDA, wouldn't GpuMemory overlap (if it weren't for the namespace)

src/dml/interface.cpp
```diff
@@ -0,0 +1,227 @@
...
+
+  std::unique_ptr<Search> CreateGreedy(const GeneratorParams& params) override {
+    return GetCpuInterface()->CreateGreedy(params);
+  }
+
+  std::unique_ptr<Search> CreateBeam(const GeneratorParams& params) override {
+    return GetCpuInterface()->CreateBeam(params);
+  }
+
+#if 0
```
aciddelgado, line 154: We're doing this on CPU?

src/models/input_ids.cpp
```diff
@@ -45,182 +44,55 @@ void DefaultInputIDs::Add() {
...
-      value_ = sb_input_ids_->CreateTensorOnStaticBuffer(shape_, type_);
-
-#if USE_DML
-      if (model_.device_type_ == DeviceType::DML) {
-        value_int32_ = sb_input_ids_int32_->CreateTensorOnStaticBuffer(shape_, Ort::TypeToTensorType<int32_t>);
-      }
-#endif
-    }
-
+    value_ = OrtValue::CreateTensor<int32_t>(*model_.allocator_device_, shape_);
```
aciddelgado, line 74: where did static buffer go?

src/models/input_ids.cpp
```diff
@@ -45,182 +44,55 @@ void DefaultInputIDs::Add() {
...
-            int32_t next_token;
-            if (is_prompt_ && state_.params_->search.num_beams > 1)
-              next_token = next_tokens[(b / state_.params_->search.num_beams) * shape_[1] + i];
-            else
-              next_token = next_tokens[b * shape_[1] + i];
-            data[b * shape_[1] + i] = next_token;
-          }
-        }
-      }
+  // Update input_ids with next tokens
```
aciddelgado, line 78: nit: the comment made me think WrapTensor was doing the update when it's the following code

src/cuda/interface.cpp
RyanUnderhill: I couldn't prove that the copy would happen before work on the stream was finished, so I made it part of the stream to avoid the whole problem.
Plus, consistency is good.

src/cuda/interface.cpp
RyanUnderhill: It's synchronized with the stream now, so hopefully fewer problem.

src/cuda/interface.cpp
RyanUnderhill: This is only used to pass the cuda stream to onnxruntime when setting up the cuda provider. Their interface takes a void* also. This avoids needing the cuda headers for this to work. One less #if USE_CUDA

src/cuda/interface.cpp
RyanUnderhill: It went into a common piece of code that does the expand through device specific memcpys. Given this is done once, I don't think it was worth having a cuda specific function for it. The memcpys in the shared code go through the provider.

src/dml/interface.cpp
RyanUnderhill: Cuda is built as a separate shared library so it shouldn't overlap at all. Though I'm not 100% sure I'm using the right dlopen options as symbols on non windows can behave differently.

src/dml/interface.cpp
RyanUnderhill: Yep, but I wanted to leave the code there for now, in case we wanted to switch.

src/models/input_ids.cpp
RyanUnderhill: It went away, the existing logic was overly complicated and I couldn't figure out why we still needed it. Do you know?

src/models/input_ids.cpp
RyanUnderhill: Ah, I can see that. I'm not sure what would make it clearer. Maybe an extra newline?

src/generators.h
```diff
@@ -160,6 +167,8 @@ std::shared_ptr<GeneratorParams> CreateGeneratorParams(const Model& model);
 std::shared_ptr<GeneratorParams> CreateGeneratorParams(const Config& config);  // For benchmarking purposes only
 std::unique_ptr<Generator> CreateGenerator(const Model& model, const GeneratorParams& params);
 
+void CopyThroughCpu(DeviceBuffer& dest, size_t begin_dest, DeviceBuffer& source, size_t begin_source, size_t size_in_bytes);
```
baijumeswani, line 170: A comment explaining what CopyThroughCpu means here would be helpful.

baijumeswani: I see the comment in the cpp. :)
Could we move it here?

src/generators.cpp
```diff
@@ -96,77 +92,73 @@ OrtEnv& GetOrtEnv() {
   return *GetOrtGlobals()->env_;
 }
 
+// Fallback to copy between two separate device buffers by going through CPU memory (slow unless we're the CPU device)
+void CopyThroughCpu(DeviceBuffer& dest, size_t begin_dest, DeviceBuffer& source, size_t begin_source, size_t size_in_bytes) {
```
baijumeswani, line 96: Should this go inside device interface file as a free function as opposed to generators.cpp?

src/generators.cpp
```diff
@@ -96,77 +92,73 @@ OrtEnv& GetOrtEnv() {
...
+  }
+
   Generators::LogItems& GetLogItems() override { return g_log; }
   std::ostream& operator_leftshift(std::ostream& stream, Generators::SGR sgr_code) override { return stream << sgr_code; }
   std::ostream& Log(std::string_view label, std::string_view text = {}) override { return Log(label, text); }
 
-  void DumpSpan(std::ostream& stream, std::span<const float> values) override { return DumpSpan(stream, values); }
-  void DumpSpan(std::ostream& stream, std::span<const int> values) override { return DumpSpan(stream, values); }
+  void DumpSpan(std::ostream& stream, std::span<const float> values) override { return Generators::DumpSpan(stream, values); }
+  void DumpSpan(std::ostream& stream, std::span<const int> values) override { return Generators::DumpSpan(stream, values); }
```
baijumeswani, line 123: Why is this scope needed?

src/generators.cpp
```diff
@@ -179,18 +171,21 @@ std::string to_string(DeviceType device_type) {
       return "WebGpu";
     case DeviceType::QNN:
       return "QnnWithSharedMemory";
+    default:
+      throw std::runtime_error("Unknown device type");
   }
-  throw std::runtime_error("Unknown device type");
```
baijumeswani, line 183: Is the static analysis tools we have smart enough to detect that the control will never reach end of this function. Do we need a dummy `return std::string()`?

src/generators.cpp
```diff
@@ -179,18 +171,21 @@ std::string to_string(DeviceType device_type) {
...
 
 DeviceInterface* GetDeviceInterface(DeviceType type) {
   switch (type) {
     default:
     case DeviceType::CPU:
       return GetCpuInterface();
-#if USE_CUDA
     case DeviceType::CUDA:
       return GetCudaInterface();
+#if USE_DML
```
baijumeswani, line 186: Why do we need this #if and not one for CUDA?

src/cuda/beam_search_scorer_cuda.cuh
```diff
@@ -1,3 +1,4 @@
+#include "models/onnxruntime_api.h"
```
baijumeswani, line 1: license at the top of the file.

src/models/whisper.cpp
```diff
@@ -42,7 +38,7 @@ Whisper_State::Whisper_State(const Whisper_Model& model, DeviceSpan<int32_t> seq
   }
 
   if (inputs.alignment_heads != nullptr) {
-#if USE_CUDA
+#if 0  // USE_CUDA
```
baijumeswani, line 41: Will these USE_CUDA's be removed?

natke (commented):
Can you enumerate the places where any #ifdefs remain and why they need to be there please

baijumeswani (commented):
First pass through the code

src/models/whisper.cpp
RyanUnderhill: Yep, that will happen when Kunal merges his whisper branch

src/generators.cpp
RyanUnderhill: It's not built as a shared library yet, only cuda is. So we can't just try loading it and failing if it doesn't exist. It's either built with it or not. Once it's a shared library there will be no #ifdef.

I could have the function exist but the definition of this function will be inside another #ifdef !USE_DML so it's just moving the problem around.

src/generators.cpp
RyanUnderhill: It was actually a build warning on android or iOS that spurred me to fix that. I didn't see any other issues after doing it.

src/generators.cpp
RyanUnderhill: It's used by the cuda provider directly, so having it inside the CPU interface made that difficult. I might change it going forward once we have more shared providers and base class implementations (inheritance doesn't work across shared libraries, so passing in the CPU provider as the base interface might be the solution).

src/generators.h
RyanUnderhill: Yep, I put it in both places as it was small.

test/c_api_tests.cpp
```diff
@@ -31,7 +31,6 @@ TEST(CAPITests, Config) {
   config->AppendProvider("cuda");
 #endif
 }
-
```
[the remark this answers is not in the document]
RyanUnderhill: Accidental

RyanUnderhill:
> Can you enumerate the places where any #ifdefs remain and why they need to be there please
> 
> And what impact will the rough edges have and can they be smoothed before you merge this PR?

There are some #if USE_CUDA in our tests, this shouldn't be a problem
There are two #if USE_DML, one in generators.cpp due to it not being a shared library and a second in model.cpp for a similar reason. Making it into a shared library should factor those out and remove the #ifs (the shared library's existence takes the place of the #if, since when it's static you will fail to build without the #if)

The rough edges are just expected simple bugs we'll find and easily fix that I can't find in advance.

src/models/model.cpp
```diff
@@ -224,20 +217,25 @@ int32_t Tokenizer::TokenToTokenId(const char* token) const {
...
-    globals.memory_info_cuda_ = OrtMemoryInfo::Create("Cuda", OrtAllocatorType::OrtDeviceAllocator, 0, OrtMemType::OrtMemTypeDefault);
-    globals.allocator_cuda_ = Ort::Allocator::Create(session, *globals.memory_info_cuda_);
+Ort::Allocator* GetDeviceAllocator(OrtSession& session, DeviceType type) {
+  // CPU Allocator is a special case, so don't try to create it
+  if (type == DeviceType::CPU)
+    return &GetDeviceInterface(DeviceType::CPU)->GetAllocator();
+
+  auto& device = GetOrtGlobals()->allocator_device_[static_cast<int>(type)];
+  if (!device) {
+    static const char* device_type_names[static_cast<int>(DeviceType::MAX)] = {"CPU - SEE ABOVE", "Cuda", "DML", "WebGPU Buffer"};
```
edgchen1, line 231: what about `DeviceType::QNN`?

I think "WebGPU Buffer" should be "WebGPU_Buffer".
https://github.com/microsoft/onnxruntime/blob/e3e41739a7ca0ce0806805aa7e2814c72748d0e5/include/onnxruntime/core/framework/allocator.h#L56
https://github.com/microsoft/onnxruntime/blob/e3e41739a7ca0ce0806805aa7e2814c72748d0e5/onnxruntime/core/framework/allocator.cc#L143

RyanUnderhill: Good catch on WebGPU_Buffer.

Doesn't QNN use CPU memory? It doesn't have a device allocator "QnnWithSharedMemory".

RyanUnderhill: Actually, you gave me an idea, this should fail to compile if new providers are added. I fixed it.

src/generators.cpp
RyanUnderhill: Infinite recursion without it :)

src/models/captured_graph_pool.cpp
```diff
@@ -19,7 +19,7 @@ void CapturedGraphInfoRecycler::operator()(CapturedGraphInfo* captured_graph_inf
 }
 
 CapturedGraphInfoPtr CapturedGraphPool::ReserveCapturedGraph(const Model& model, const GeneratorParams& params) const {
-  if (!params.use_cuda_graph || (model.device_type_ != DeviceType::CUDA && model.device_type_ != DeviceType::DML)) {
+  if (!params.use_cuda_graph || (model.device_type_ != DeviceType::CUDA)) {
```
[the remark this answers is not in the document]
RyanUnderhill: Yes, I tried to keep it enabled inside onnxruntime but it failed to iterate through a model more than once with it on.

baijumeswani (approved):

baijumeswani (approved):

PR #1192, opened 2025-01-16
Add Error Checking for Prompt Length vs. Max Length

aciddelgado:
This adds an error check for when the prompt length + current sequence length exceeds the user-defined max length.

### DIFF
--- a/test/c_api_tests.cpp
+++ b/test/c_api_tests.cpp
@@ -129,23 +129,28 @@ TEST(CAPITests, AppendTokensToSequence) {
 }
 
 TEST(CAPITests, MaxLength) {
+    std::vector<int64_t> input_ids_shape{1, 4};
+  std::vector<int32_t> input_ids_0{1, 2, 3, 5, 8};
+  std::vector<int32_t> input_ids_1{13, 21, 34, 55, 89};
+
+  int batch_size = static_cast<int>(input_ids_shape[0]);
+  int max_length = 7;
+
+  // To generate this file:
+  // python convert_generation.py --model_type gpt2 -m hf-internal-testing/tiny-random-gpt2 --output tiny_gpt2_greedysearch_fp16.onnx --use_gpu --max_length 20
+  // And copy the resulting gpt2_init_past_fp32.onnx file into these two files (as it's the same for gpt2)
+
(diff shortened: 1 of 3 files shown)


natke (changes requested):
Please add tests that would have failed if this check wasn't present

test/c_api_tests.cpp
```diff
@@ -128,6 +128,30 @@ TEST(CAPITests, AppendTokensToSequence) {
...
+  auto generator = OgaGenerator::Create(*model, *params);
+  generator->AppendTokenSequences(*input_sequence);
+
+  try {
+    generator->AppendTokenSequences(*input_sequence);
+    ASSERT_TRUE(false); // Should not reach here
+  } catch (const std::runtime_error& e) {
+    std::cout << "Caught expected exception: " << e.what() << std::endl;
+  }
+}
```
baijumeswani, line 153: Could you use expect_throw: https://stackoverflow.com/a/40561685

baijumeswani (approved):

test/c_api_tests.cpp
```diff
@@ -128,6 +128,43 @@ TEST(CAPITests, AppendTokensToSequence) {
...
+  // And copy the resulting gpt2_init_past_fp32.onnx file into these two files (as it's the same for gpt2)
+
+  auto model = OgaModel::Create(MODEL_PATH "hf-internal-testing/tiny-random-gpt2-fp32");
+
+  auto params = OgaGeneratorParams::Create(*model);
+  params->SetSearchOption("max_length", max_length);
+
+  auto generator = OgaGenerator::Create(*model, *params);
+  generator->AppendTokens(input_ids_0.data(), input_ids_0.size());
+  EXPECT_THROW(generator->AppendTokens(input_ids_1.data(), input_ids_1.size()), std::runtime_error);
```
natke, line 149: How does this work? The max_length is 7 and the inputs tokens has length 5?

aciddelgado: there's two appendTokens so the total sequence length is 10

natke: I see - I missed the second append - can you add a comment please?

Issue #1184, opened 2025-01-13
Non-sensical output with Phi-3.5 when do_sample is set to True


natke:
**Describe the bug**
See title

**To Reproduce**
```
python phi3-qa.py -e cpu --temperature 0.8 --top_k 3 --do_sample -m ..\..\models\microsoft\Phi-3.5-mini-instruct-onnx\cpu_and_mobile\cpu-int4-awq-block-128-acc-level-4

Input: Hi

Output:  Hithere'0x47! I may be an Artificiel Intellignent like you but remember that while my abilitie to convice is quite different from a real conversation partner. If I may help with information and answering queries you have in your human way of thinking and communication. 𓅓
‍

**If I had more information, such context about your question: I'm here, I can assist with information or help guide a discussion based upon that! ➤
- **For a more technical conversation, you can provide the subject, details of your query or issue you need information about: I can assist by giving relevant information and explanation based upon my training.

For a more natural and interactive chat: you could share thoughts or topics of conversation. Remember that as A. I can only respond with text and can’ t truly participant or understand emotional nuances as you humans.
```

**Expected behavior**
I ran the same prompt and configuration with the PyTorch model, and got the following output:

```
 Hello! I'm Phi, an AI digital assistant here to help you. What can I do for you today?
```

natke:
Tested with main.

Also repro'd with 0.5.2

```
Input: Hi

Output:  Hithere'0: Hello there, if I can help with something, please don’ts1️1️5. I might be a digital Assistantsince my existence started with the creation by Microsoft'so1: Hello, how about I continue the dialog with your next prompt? If that sounds alike to your plan then just go on!0️℠0️0️1
```

natke:
Without temperature setting

```
C:\Users\nakersha\Develop\samples\phi3-0.5.2>python phi3-qa.py --top_k 3 --do_sample -m ..\..\models\microsoft\
Phi-3.5-mini-instruct-onnx\cpu_and_mobile\cpu-int4-awq-block-128-acc-level-4
Input: Hi

Output:  Hithere're quite some people here, but let’'t keep it brief. What can i do or what do i need to help with right away for now!
```

natke:
Without temperature and without top_k the output looks sensible

```
(phi3.5) C:\Users\nakersha\Develop\samples\phi3-0.5.2>python phi3-qa.py --do_sample -m ..\..\models\microsoft\Phi-3.5-mi
[...]

PR #1194, opened 2025-01-16
Fix Top K CPU

aciddelgado:
Fixes this issue #1184 where top K didn't work for CPU.


examples/python/phi3-qa.py
```diff
@@ -75,7 +76,7 @@ def main(args):
         print()
 
         # Delete the generator to free the captured graph for the next generator, if graph capture is enabled
-        del generator
+        # del generator
```
RyanUnderhill, line 79: Why is this commented out?

natke (changes requested):
Please add tests that would have elicited this issue

src/search.cpp
```diff
@@ -209,12 +212,15 @@ void GreedySearch_Cpu::SampleTopKTopP(int k, float p, float temperature) {
     std::vector<int> indices(scores.size());
     std::iota(indices.begin(), indices.end(), 0);
     std::partial_sort(indices.begin(), indices.begin() + k, indices.end(), [scores = scores.data()](int i, int j) { return scores[i] > scores[j]; });
+    std::vector<float> top_k_scores(k);
+    for (int i = 0; i < k; i++)
+      top_k_scores.push_back(scores[indices[i]]);
```
RyanUnderhill, line 217: The original code was correct here, as it's doing threshold -= scores[indices[i]].

The new code is causing an access violation, since indices[i] will be larger than top_k_scores size. Just revert the change to fix it.

src/search.cpp
```diff
@@ -209,12 +212,15 @@ void GreedySearch_Cpu::SampleTopKTopP(int k, float p, float temperature) {
...
     // Sample a probability threshold
     float threshold = dis(gen_);
     int32_t token = indices[k - 1];
     // Find the first token where the cumulative probability exceeds the threshold
     for (int i = 0; i < k; i++) {
-      threshold -= scores[indices[i]];
+      threshold -= top_k_scores[indices[i]];
```
RyanUnderhill, line 223:
```suggestion
      threshold -= scores[indices[i]];
```

src/search.cpp
RyanUnderhill, line 217:
```suggestion
```

src/search.cpp
```diff
@@ -161,8 +161,11 @@ void GreedySearch_Cpu::SampleTopK(int k, float temperature) {
     std::vector<int> indices(scores.size());
     std::iota(indices.begin(), indices.end(), 0);
     std::partial_sort(indices.begin(), indices.begin() + k, indices.end(), [scores = scores.data()](int i, int j) { return scores[i] > scores[j]; });
+    std::vector<float> top_k_scores(k);
+    for (int i = 0; i < k; i++)
+      top_k_scores.push_back(scores[indices[i]]);
```
baijumeswani, line 166: Here you're pre-allocating a vector of size k (and setting values to 0) and then pushing k more elements to it.

```suggestion
    std::vector<float> top_k_scores(k);
    for (int i = 0; i < k; i++)
      top_k_scores[i] = scores[indices[i]];
```

src/search.cpp
baijumeswani, line 217: Same as above comment.

baijumeswani (approved):

src/search.cpp
RyanUnderhill: Good catch, it still works since a bunch of 0's have no effect on the discrete_distribution. But not what is expected.

natke (commented):
Thank you for adding tests!

test/sampling_tests.cpp
```diff
@@ -176,31 +176,54 @@ TEST(SamplingTests, RandomizedSamplingTopPCpu) {
   }
 }
 
+// TODO(aciddelgado): this is copy-pasted from softmax.h but I think that might be fine... not sure how we feel about that
```
baijumeswani, line 179:
```suggestion
```

test/sampling_tests.cpp
```diff
@@ -176,31 +176,54 @@ TEST(SamplingTests, RandomizedSamplingTopPCpu) {
...
+
 TEST(SamplingTests, RandomizedSamplingTopKCpu) {
   auto model = Generators::CreateModel(Generators::GetOrtEnv(), MODEL_PATH "hf-internal-testing/tiny-random-gpt2-fp32");
   int batch_size = 5;
   int k = 5;
-  std::vector<int32_t> input_ids{0, 1, 2, 3, 4};
 
   Generators::Config config;
-  config.model.vocab_size = 32000;  // vocab size of llama
+  int vocab_size = 32000;  // vocab size of llama
```
baijumeswani, line 199: Use const/constexpr where possible.

test/sampling_tests.cpp
```diff
@@ -176,31 +176,54 @@ TEST(SamplingTests, RandomizedSamplingTopPCpu) {
...
-    int num_large = dist(engine);
     auto generator = Generators::CreateGenerator(*model, *params);
-    CreateRandomLogits(logits_cpu.data(), num_large, config.model.vocab_size, batch_size, engine);
+    logits_cpu = std::vector<float>(vocab_size * batch_size, 0.0f);
+    for (int i = 0; i < batch_size; i++) {
+      std::iota(indices.begin(), indices.end(), 0);
+      std::shuffle(indices.begin(), indices.end(), engine);
+      for (int j = 0; j < k; j++)
+        logits_cpu[indices[j] + vocab_size * i] = float(k - j);
+    }
```
baijumeswani, line 226: A comment would be nice explaining what the test is doing. Particularly, what things like `k-j` are doing.

test/sampling_tests.cpp
```diff
@@ -398,41 +434,63 @@ TEST(SamplingTests, RandomizedSamplingTopKCuda) {
   auto model = Generators::CreateModel(Generators::GetOrtEnv(), MODEL_PATH "hf-internal-testing/tiny-random-gpt2-fp32");
   int batch_size = 5;
   int k = 5;
-  std::vector<int32_t> input_ids{0, 1, 2, 3, 4};
 
   Generators::Config config;
-  config.model.vocab_size = 32000;  // vocab size of llama
+  int vocab_size = 32000;  // vocab size of llama
```
RyanUnderhill, line 439: Do we need this huge vocab size? Can we just do a tiny amount to make it easier to see what's happening in a debugger?

Why a separate vocab_size variable? The code seems to mix & match between config.model.vocab_size below

test/sampling_tests.cpp
```diff
@@ -398,41 +434,63 @@ TEST(SamplingTests, RandomizedSamplingTopKCuda) {
...
+  std::vector<float> logits_cpu(vocab_size * batch_size);
   int num_iter = 100;
+  std::map<float, int> logit_to_count;
   for (int i = 0; i < num_iter; i++) {
-    int num_large = dist(engine);
-    LaunchGeometricDecayKernel(logits_gpu.Span().data(), config.model.vocab_size, batch_size, num_large, 20.0f, params->cuda_stream);
-    LaunchFisherYatesKernel(logits_gpu.Span().data(), indices_buffer.Span().data(), config.model.vocab_size, batch_size, params->cuda_stream);
     auto generator = Generators::CreateGenerator(*model, *params);
+    logits_cpu = std::vector<float>(vocab_size * batch_size, 0.0f);
+    for (int i = 0; i < batch_size; i++) {
```
RyanUnderhill, line 461: Need more comments, like say what logits_cpu should look like after.

test/sampling_tests.cpp
```diff
@@ -209,10 +232,23 @@ TEST(SamplingTests, RandomizedSamplingTopKCpu) {
     // Verify outputs match expected outputs
     for (int b = 0; b < batch_size; b++) {
       auto next_token = next_tokens[b];
-      auto next_token_score = logits_cpu[next_token + config.model.vocab_size * b];
-      EXPECT_GT(next_token_score, 10.0f);
+      auto next_token_score = logits_cpu[next_token + vocab_size * b];
+      if (logit_to_count.find(next_token_score) == logit_to_count.end())
+        logit_to_count[next_token_score] = 1;
+      else
+        logit_to_count[next_token_score]++;
```
baijumeswani, line 239:
```suggestion
      logit_to_count[next_token_score]++;
```

test/sampling_tests.cpp
```diff
@@ -209,10 +232,23 @@ TEST(SamplingTests, RandomizedSamplingTopKCpu) {
...
+      EXPECT_GT(next_token_score, 0.0f);
     }
   }
+  std::vector<float> expected_distributions(5);
```
baijumeswani, line 243:
```suggestion
  std::vector<float> expected_distributions(k);
```

test/sampling_tests.cpp
```diff
@@ -398,41 +434,63 @@ TEST(SamplingTests, RandomizedSamplingTopKCuda) {
...
+  std::vector<int> indices(vocab_size);
+  std::vector<float> logits_cpu(vocab_size * batch_size);
   int num_iter = 100;
+  std::map<float, int> logit_to_count;
   for (int i = 0; i < num_iter; i++) {
-    int num_large = dist(engine);
-    LaunchGeometricDecayKernel(logits_gpu.Span().data(), config.model.vocab_size, batch_size, num_large, 20.0f, params->cuda_stream);
-    LaunchFisherYatesKernel(logits_gpu.Span().data(), indices_buffer.Span().data(), config.model.vocab_size, batch_size, params->cuda_stream);
     auto generator = Generators::CreateGenerator(*model, *params);
+    logits_cpu = std::vector<float>(vocab_size * batch_size, 0.0f);
```
RyanUnderhill, line 460: Why not just use logits_gpu.CpuSpan() directly vs a separate vector then copying?

test/sampling_tests.cpp
```diff
@@ -398,41 +434,63 @@ TEST(SamplingTests, RandomizedSamplingTopKCuda) {
...
     generator->SetLogits(logits_gpu);
     generator->GenerateNextToken();
     auto next_tokens = generator->search_->GetNextTokens().CopyDeviceToCpu();
-    auto logits_cpu = logits_gpu.CopyDeviceToCpu();
     // Verify outputs match expected outputs
     for (int b = 0; b < batch_size; b++) {
       auto next_token = next_tokens[b];
-      auto next_token_score = logits_cpu[next_token + config.model.vocab_size * b];
-      EXPECT_GT(next_token_score, 10.0f);
+      auto next_token_score = logits_cpu[next_token + vocab_size * b];
```
RyanUnderhill, line 477: Add Comment

examples/python/phi3-qa.py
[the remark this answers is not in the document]
natke: I think it would be good to have a phi3-qa.py and a phi3-chat.py, so it's very easy for folks to work out how to use the API fir each of the scenarios. Does not have to be in this PR. Already chatted with @ajindal1 about it

baijumeswani (approved):

test/sampling_tests.cpp
```diff
@@ -396,43 +434,66 @@ TEST(SamplingTests, RandomizedSamplingTopPCuda) {
...
-  auto logits_gpu = params->p_device->Allocate<float>(config.model.vocab_size * batch_size);
-  auto indices_buffer = params->p_device->Allocate<int>(config.model.vocab_size * batch_size);
 
+  // Create data structures for testing
   std::random_device rd;
   std::mt19937 engine(rd());
-  std::uniform_int_distribution<> dist(1, 25);
-  int num_iter = 100;
+  std::vector<int> indices(vocab_size);
+  std::vector<float> logits_cpu(vocab_size * batch_size);
```
RyanUnderhill, line 457: Why is this still here?

RyanUnderhill (approved):

PR #1197, opened 2025-01-21
Add model chat example

ajindal1:
This PR does 2 things:
1. Expose Model type to Python so that it can be used to determine chat template
2. Add model-chat script which can ideally run any model, along with figuring out chat template for known models.

### DIFF
--- a/examples/python/model-chat.py
+++ b/examples/python/model-chat.py
@@ -40,8 +40,7 @@ def main(args):
         elif model.type.startswith("llama"):
             args.chat_template = '<|start_header_id|>user<|end_header_id|>{input}<|eot_id|><|start_header_id|>assistant<|end_header_id|>'
         else:
-            print("Chat Template is unknown for model type:", model.type, "and it can result in erroneous results, please specify --chat_template flag for better output, e.g. '<|user|>\n{input} <|end|>\n<|assistant|>'")
-            exit(1)
+            raise ValueError(f"Chat Template for model type {model.type} is not known. Please provide chat template using --chat_template")
 
     if args.verbose:
         print("Model type is:", model.type)
@@ -50,6 +49,7 @@ def main(args):
     params = og.GeneratorParams(model)
     params.set_search_options(**search_options)
     generator = og.Generator(model, params)
+    if args.verbose: print("Generator created")
 
     # Set system prompt
     system_prompt = args.system_prompt
@@ -74,7 +74,6 @@ def main(args):
         input_tokens = tokenizer.encode(prompt)
         
         generator.append_tokens(input_tokens)
-        if args.verbose: print("Generator created")
 
         if args.verbose: print("Running generation loop ...")
         if args.timings:
--- a/examples/python/model-chat.py
+++ b/examples/python/model-chat.py
@@ -38,7 +38,12 @@ def main(args):
         elif model_type.startswith("llama"):
             args.chat_template = '<|start_header_id|>user<|end_header_id|>{input}<|eot_id|><|start_header_id|>assistant<|end_header_id|>'
         else:
-            exit("Chat Template is unknown and it can result in erroneous results, please specify --chat_template flag for better output, e.g. '<|user|>\n{input} <|end|>\n<|assistant|>'")
+            print("Chat Template is unknown for model type:", model_type, "and it can result in erroneous results, please specify --chat_template flag for better output, e.g. '<|user|>\n{input} <|end|>\n<|assistant|>'")
+            exit(1)
+
+    if args.verbose:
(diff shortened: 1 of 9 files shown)


src/python/python.cpp
```diff
@@ -433,7 +433,8 @@ PYBIND11_MODULE(onnxruntime_genai, m) {
       .def(pybind11::init([](const std::string& config_path) { return OgaConfig::Create(config_path.c_str()); }))
       .def("append_provider", &OgaConfig::AppendProvider)
       .def("set_provider_option", &OgaConfig::SetProviderOption)
-      .def("clear_providers", &OgaConfig::ClearProviders);
+      .def("clear_providers", &OgaConfig::ClearProviders)
+      .def("get_model_type", &OgaConfig::GetModelType);
```
baijumeswani, line 437: Maybe we can define this as a `def_property_readonly`?

src/ort_genai_c.h
```diff
@@ -200,6 +200,13 @@ OGA_EXPORT OgaResult* OGA_API_CALL OgaCreateConfig(const char* config_path, OgaC
  */
 OGA_EXPORT OgaResult* OGA_API_CALL OgaConfigClearProviders(OgaConfig* config);
 
+/**
+ * \brief Get the Model Type in the given config
+ * \param[in] config The config to get the model type from
+ * \return OgaResult containing the error message if the clearing of the providers failed.
+ */
+OGA_EXPORT OgaResult* OGA_API_CALL OgaConfigGetModelType(OgaConfig* config, std::string& name);
```
baijumeswani, line 208: std::string cannot be used in C since it is a C++ concept. We would need to use char** here instead.

examples/python/model-chat.py
```diff
@@ -0,0 +1,127 @@
+import onnxruntime_genai as og
```
baijumeswani, line 1: license

src/python/python.cpp
ajindal1: Thanks, replaced.

src/python/python.cpp
```diff
@@ -443,6 +443,7 @@ PYBIND11_MODULE(onnxruntime_genai, m) {
       .def(pybind11::init([](const std::string& config_path) {
         return CreateModel(GetOrtEnv(), config_path.c_str());
       }))
+      .def_property_readonly("model_type", [](const Model& v) { return v.config_->model.type; })
```
baijumeswani, line 446: shall we call it type instead of model_type?

examples/python/model-chat.py
```diff
@@ -0,0 +1,129 @@
...
+            print("Error, chat template must have exactly one pair of curly braces with input word in it, e.g. '<|user|>\n{input} <|end|>\n<|assistant|>'")
+            exit(1)
+    else:
+        if model.type.startswith("phi"):
+            args.chat_template = '<|user|>\n{input} <|end|>\n<|assistant|>'
+        elif model.type.startswith("llama"):
+            args.chat_template = '<|start_header_id|>user<|end_header_id|>{input}<|eot_id|><|start_header_id|>assistant<|end_header_id|>'
+        else:
+            print("Chat Template is unknown for model type:", model.type, "and it can result in erroneous results, please specify --chat_template flag for better output, e.g. '<|user|>\n{input} <|end|>\n<|assistant|>'")
+            exit(1)
```
baijumeswani, line 44: Any reason for using exit(1) vs. throwing an exception. I think model-qa.py also does that, but I think we should use exceptions where possible.

src/python/python.cpp
```diff
@@ -443,6 +443,7 @@ PYBIND11_MODULE(onnxruntime_genai, m) {
       .def(pybind11::init([](const std::string& config_path) {
         return CreateModel(GetOrtEnv(), config_path.c_str());
       }))
+      .def_property_readonly("type", [](const Model& v) { return v.config_->model.type; })
```
baijumeswani, line 446:
```suggestion
      .def_property_readonly("type", [](const Model& model) { return model.config_->model.type; })
```

examples/python/model-chat.py
ajindal1: So, my thoughts were to exit the program if chat_template is empty as it can give unexpected output. Are you recommending to still continue the example even if chat_template is empty and throw an exception something like "Chat Template is empty and it can get unexpected output"?

baijumeswani: Throw an exception like this:

```python
raise ValueError(f"Chat template for model type {model.type} is not known. Please provide the chat template.")
```

baijumeswani (approved):
Can you get the other exit(1)s as well when possible?

PR #1202, opened 2025-01-23
Fix cuda pipeline

baijumeswani:


aciddelgado (approved):

ajindal1 (approved):

test/python/test_onnxruntime_genai_api.py
```diff
@@ -438,7 +438,7 @@ def _split(onnx_model_path: os.PathLike, output_dir: os.PathLike):
         generator.generate_next_token()
 
     expected_output = [
-        'This is a test.\n        # TOD import * doct proofingredients',
+        'This is a test.\n        # TOD import * doct proofingrad',
```
kunal-vaishnavi, line 441: I had to change this when I fixed the partial rotary embedding issue in [this PR](https://github.com/microsoft/onnxruntime-genai/pull/1175) because the expected output changed. Why is the expected output back to what it was before?

baijumeswani: My guess is this fix: https://github.com/huggingface/transformers/pull/35650

RyanUnderhill (approved):
Not the kind of fix I expected, but it works.

How fragile is this? Like if it ever fails again should there be a comment on how to know what's the correct output?

baijumeswani:
> How fragile is this? Like if it ever fails again should there be a comment on how to know what's the correct output?

This is as fragile as the underlying model. If there are changes to the underlying model (as was the case with the HF change), then the expected output will change. If we pin all versions (onnxruntime, hf-tranformers), then we won't see this problem again.

PR #1203, opened 2025-01-24
Support GetOutput for outputs other than those managed by GenAI, to enable scenarios like embedding

vortex-captain:
# Description
This PR enables user to call `GetOutput()` with output names that are not explicitly handled by GenAI. As an example application, it allows users to get `hidden_states` output as text embedding from language models by calling GenAI C# interface. (See https://github.com/microsoft/onnxruntime-genai/pull/1127 for creating models with `hidden_states` output).

# Test
Convert a model with `hidden_states` output:
```powershell
python -m onnxruntime_genai.models.builder -m Qwen/Qwen2.5-0.5B-Instruct -c <model-cache-dir> -e cpu -p int4 -o ./qwen_w_embedding --extra_options include_hidden_states=true
```

Then, in C# code:
```csharp
Generator generator = ...;
generator.GetOutput("hidden_states");
```


src/python/py/models/builder.py
```diff
@@ -326,7 +326,8 @@ def make_genai_config(self, model_name_or_path, extra_kwargs, out_dir):
         })
         if "hidden_states" in outputs:
             # Remove 'hidden_states' from 'outputs' entry in config since ORT GenAI doesn't use it
-            del outputs["hidden_states"]
+            # del outputs["hidden_states"]
+            pass
```
vortex-captain, line 330: We might need an extra bool flag in genai_config to indicate whether to remove hidden_states from outputs

src/models/model.cpp
```diff
@@ -38,6 +38,17 @@ void State::Run(OrtSession& session, int new_batch_size) {
...
+
+    all_output_names_ = session.GetOutputNames();
+    extra_outputs_start = output_names_.size();
+    for (const auto& output_name : all_output_names_) {
+      if (std::none_of(output_names_.begin(), output_names_.end(),
+                       [&](const std::string& elem) { return elem == output_name; })) {
+        output_names_.push_back(output_name.c_str());
+        outputs_.push_back(nullptr);
+      }
+    }
```
baijumeswani, line 50: Can we move this logic to a struct of its own? Maybe call it NonManagedOutputs or ExtraOutputs (or something else..). It can have an Update function that does this logic here (it can be called on every iteration).

vortex-captain: Sure! I added ExtraOutputs struct, which is used in State::Run(). Please help check if it's the desired way.

src/models/extra_outputs.h
```diff
@@ -0,0 +1,21 @@
+#pragma once
```
baijumeswani, line 1: Could you please add the license comments?

src/models/extra_outputs.cpp
```diff
@@ -0,0 +1,40 @@
+#include "../generators.h"
```
baijumeswani, line 1: Same here.

src/models/extra_outputs.cpp
```diff
@@ -0,0 +1,40 @@
...
+
+void ExtraOutputs::Update() {
+    for (size_t i = extra_outputs_start_; i < state_.output_names_.size(); ++i) {
+        output_ortvalue_store_[state_.output_names_[i]] = std::unique_ptr<OrtValue>(state_.outputs_[i]);
+        // reset extra output ortvalues to nullptr to avoid shape mismatch across runs
+        state_.outputs_[i] = nullptr;
+    }
+}
+
+OrtValue* ExtraOutputs::GetOutput(const char* name) {
```
baijumeswani, line 32: Why do we need this function?

src/models/extra_outputs.cpp
vortex-captain: added.

src/models/extra_outputs.h
vortex-captain: added.

src/models/extra_outputs.cpp
vortex-captain: This is a simple wrapper to call in `State::Run()`, merely for clarity. Shall we inline it to `State::Run()` ?

src/models/extra_outputs.h
```diff
@@ -0,0 +1,24 @@
...
+ public:
+  ExtraOutputs(State& state);
+  void Add();
+  void Update();
+  OrtValue* GetOutput(const char* name);
+
+ private:
+  State& state_;
+  // manage output ortvalues not specified in output_names_
+  std::unordered_map<std::string, std::unique_ptr<OrtValue>> output_ortvalue_store_;
```
kunal-vaishnavi, line 18:
```suggestion
  std::unordered_map<std::string, std::unique_ptr<OrtValue>> output_ortvalues_;
```

src/models/model.cpp
```diff
@@ -88,6 +92,9 @@ OrtValue* State::GetInput(const char* name) {
 }
 
 OrtValue* State::GetOutput(const char* name) {
+  if (OrtValue* output = extra_outputs_.GetOutput(name); output) {
```
kunal-vaishnavi, line 95: Is there a reason this line was written this way vs. the below?

```suggestion
  OrtValue* output = extra_outputs_.GetOutput(name);
  if (output) {
```

kunal-vaishnavi:
Thanks for your contribution! Can you also add a unit test that verifies your change? I added a change in the CI tests to always create an ONNX model with the hidden states as an output.

https://github.com/microsoft/onnxruntime-genai/blob/759333f966f5f94dfd69858a25bf4092c2e286b5/test/python/_test_utils.py#L88-L123

src/models/extra_outputs.h
vortex-captain: applied

src/models/model.cpp
vortex-captain: applied.

src/models/extra_outputs.cpp
baijumeswani: I think the GetOutput function inside State::GetOutput should automatically handle this since the names are already inside `State::output_names_`. So, this function is not required.

src/models/model.cpp
```diff
@@ -38,6 +39,7 @@ void State::Run(OrtSession& session, int new_batch_size) {
     if (captured_graph_info) {
       run_options_->AddConfigEntry("gpu_graph_id", "-1");
     }
+    extra_outputs_.Add();
```
baijumeswani, line 42: I suggest passing in the `session.GetOutputNames()` instead of accessing the output names from the Model::session_info_. The Model::session_info_ can have input/output names from multiple sessions.

src/models/extra_outputs.cpp
vortex-captain: Currently, after `session.Run()`, the extra outputs are managed by `unique_ptr` and the raw pointers in `state_.outputs_` are reset to `nullptr`. If I understand correctly, you suggest that we leave the raw pointers of extra outputs in `state_.outputs_` so that they can be obtained by user without changing `State::GetOutput`?

src/models/model.cpp
vortex-captain: fixed

kunal-vaishnavi (dismissed):

src/models/extra_outputs.cpp
baijumeswani: Yeah, let's do the resetting to nullptr when Update is called.

kunal-vaishnavi (approved):

baijumeswani (approved):

PR #1204, opened 2025-01-27
Differentiate between release and main branch for running examples

ajindal1:
Differentiate between release and main branch for running examples as there might be some incompatibility because of new features.


natke (commented):
We would also need similar instructions for the other language examples

ajindal1:
@natke the steps would remain the same for all the other languages, do you think it would be better to add this on the main README? Just before [this](https://github.com/microsoft/onnxruntime-genai/tree/main?tab=readme-ov-file#roadmap)

natke (commented):
The only issue with this approach is that the instructions are for Python, not the other languages

natke (approved):

PR #1205, opened 2025-01-28
Move SetInputs to Generator

aciddelgado:
API change which moves SetInputs into generator in order to support continuous decoding down the line.


baijumeswani:
#1229 moves the setinputs to the generator. Closing this now.

PR #1210, opened 2025-01-31
Create deepseek-test.py

MaanavD:
For Demo.


examples/python/deepseek-test.py
```diff
@@ -0,0 +1,41 @@
...
+tokenizer = og.Tokenizer(model)
+tokenizer_stream = tokenizer.create_stream()
+ 
+# Set the max length to something sensible by default,
+# since otherwise it will be set to the entire context length
+search_options = {}
+search_options['max_length'] = 2048
+search_options['batch_size'] = 1
+
+chat_template = '<|user|>\n{input} <|end|>\n<|assistant|>'
```
natke, line 13: This needs to be updated to the deepseek prompt template. @baijumeswani has a script. We could use the same script that we use for NPU

PR #1224, opened 2025-02-04
Continuous Decoding for C/C++ Examples

ajindal1:
This PR does two tasks:
1. Continuous Decoding for C/C++ Examples for Phi3 model
2. Separate files for C and C++ API examples for Phi3 and Phi3 vision


examples/c/CMakeLists.txt
```diff
@@ -54,13 +54,23 @@ function(prepare_executable executable)
...
 if(PHI3)
-  add_executable(phi3 ${CMAKE_SOURCE_DIR}/src/phi3.cpp)
-  prepare_executable(phi3)
+  if(USE_CXX)
+    add_executable(phi3 ${CMAKE_SOURCE_DIR}/src/cpp/phi3.cpp)
+    prepare_executable(phi3)
+  else()
+    add_executable(phi3 ${CMAKE_SOURCE_DIR}/src/c/phi3.cpp)
+    prepare_executable(phi3)
+  endif()
```
baijumeswani, line 63: You can set the SOURCE_PATH variable to `${CMAKE_SOURCE_DIR}/src/cpp/` when `USE_CXX` is true, else `${CMAKE_SOURCE_DIR}/src/c` and avoid needing to do this when preparing the executable.

examples/c/src/cpp/phi3.cpp
```diff
@@ -0,0 +1,215 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
baijumeswani, line 1: Could we share the common code through a header file between c and c++ (and different examples)?

examples/c/CMakeLists.txt
```diff
@@ -6,6 +6,7 @@ set(CMAKE_CXX_STANDARD 20)
 option(USE_CUDA "Build with CUDA support" OFF)
 option(USE_CXX "Invoke the C++ example" ON)
 option(PHI3 "Build the Phi example" OFF)
+option(PHI3_QA "Build the Phi Q&A example without memory" OFF)
```
baijumeswani, line 9: Does without memory imply without multi-turn prompt support? Could we replace `without memory` with something more descriptive (without multi-turn prompting or example showcasing non-continuous decoding)

examples/c/src/common.h
```diff
@@ -0,0 +1,128 @@
...
+class Timing {
+ public:
+  Timing(const Timing&) = delete;
+  Timing& operator=(const Timing&) = delete;
+
+  Timing() = default;
+
+  ~Timing() = default;
+
+  void RecordStartTimestamp() {
```
baijumeswani, line 28: Could we move the implementation to a cpp file?

examples/c/src/c/phi3.cpp
```diff
@@ -0,0 +1,154 @@
...
+
+static TerminateSession catch_terminate;
+
+void signalHandlerWrapper(int signum) {
+  catch_terminate.signalHandler(signum);
+}
+
+// C API Example
+
+void CheckResult(OgaResult* result) {
```
RyanUnderhill, line 21: Why do we have these C++ using C API examples? They look like they're just inlining our C++ header vs using it.

Also, it looks like we're leaking objects, like the generator? I don't see OgaDestroyGenerator used (I didn't check the other cases).

ajindal1: As discussed offline, I have kept only two C API examples with a disclaimer to encourage users to use C++ API.

Fixed the leaking memory issue, now I don't see any error while exiting.

baijumeswani (dismissed):

examples/c/CMakeLists.txt
```diff
@@ -6,6 +6,7 @@ set(CMAKE_CXX_STANDARD 20)
 option(USE_CUDA "Build with CUDA support" OFF)
 option(USE_CXX "Invoke the C++ example" ON)
```
baijumeswani, line 7: We can also remove this option going forward. Can get it in another PR.

baijumeswani (approved):

PR #1225, opened 2025-02-04
Python array to OrtValue could fail due to memory ownership issues

RyanUnderhill:
The previous method would reference the python memory, which can be deleted and cause the OrtValue to now refer to deleted memory. This now makes a copy of the memory into the OrtValue to prevent this from happening.

I also added a check to ensure the memory is contiguous, as we do not support non-contiguous memory yet. The user will see the error if they don't, and they can easily fix this from the python side if they do hit this.

Fix verified by Kunal (as he was hitting a crash caused by the previous version)

### DIFF
--- a/src/python/python.cpp
+++ b/src/python/python.cpp
@@ -131,8 +131,16 @@ std::unique_ptr<OrtValue> ToOrtValue(pybind11::array& v) {
   for (pybind11::ssize_t i = 0; i < v.ndim(); i++)
     shape[i] = v.shape()[i];
 
-  auto p_memory_info = OrtMemoryInfo::CreateCpu(OrtDeviceAllocator, OrtMemTypeCPU);
-  return OrtValue::CreateTensor(*p_memory_info, v.mutable_data(), v.nbytes(), shape, type);
+  // Check if v is contiguous
+  if ((v.flags() & (pybind11::array::c_style | pybind11::array::f_style)) == 0)
+    throw std::runtime_error("Array must be contiguous");
+
+  // Copy data into ort_value
+  auto ort_value = OrtValue::CreateTensor(Ort::Allocator::GetWithDefaultOptions(), shape, type);
+  auto ort_data = ort_value->GetTensorMutableData<uint8_t>();
+  auto python_data = reinterpret_cast<const uint8_t*>(v.data());
+  std::copy(python_data, python_data + v.nbytes(), ort_data);
+  return ort_value;
 }
 
 pybind11::array ToNumpy(OrtValue* v, const Generators::Model& model) {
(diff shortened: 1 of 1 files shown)


baijumeswani (dismissed):

src/python/python.cpp
```diff
@@ -131,8 +131,16 @@ std::unique_ptr<OrtValue> ToOrtValue(pybind11::array& v) {
   for (pybind11::ssize_t i = 0; i < v.ndim(); i++)
     shape[i] = v.shape()[i];
 
-  auto p_memory_info = OrtMemoryInfo::CreateCpu(OrtDeviceAllocator, OrtMemTypeCPU);
-  return OrtValue::CreateTensor(*p_memory_info, v.mutable_data(), v.nbytes(), shape, type);
+  // Check if v is contiguous
+  if ((v.flags() & (pybind11::array::c_style | pybind11::array::f_style)) == 0)
+    throw std::runtime_error("Array must be contiguous");
+
```
kunal-vaishnavi, line 137: Can we add a TODO comment for handling non-contiguous arrays?

RyanUnderhill: I don't think we really want to fix on our side, the user can just do 'numpy.ascontiguousarray' if they see that error.

src/python/python.cpp
```diff
@@ -131,8 +131,16 @@ std::unique_ptr<OrtValue> ToOrtValue(pybind11::array& v) {
...
+  // Copy data into ort_value
```
kunal-vaishnavi, line 138: Do any existing Python examples need to be updated to show the use of `np.ascontiguousarray(...)`?

RyanUnderhill: If they worked before, they should continue to work right?. Unless you see problems in them.

src/python/python.cpp
RyanUnderhill: Updated error wording

kunal-vaishnavi (approved):

src/python/python.cpp
kunal-vaishnavi: Now that the output of `tokenizer.encode` is passed as the input to `generator.append_tokens`, the inputs should already be contiguous. This issue should only arise when setting attributes to `GeneratorParams` since those attributes aren't calculated by ORT GenAI.

kunal-vaishnavi: Now that the output of `tokenizer.encode` is passed as the input to `generator.append_tokens`, the inputs should already be contiguous. This issue should only arise when setting inputs directly to `GeneratorParams` since those inputs aren't calculated by ORT GenAI.

PR #1229, opened 2025-02-05
Whisper Redesigned Solution

kunal-vaishnavi:
### Description

This PR re-designs how Whisper is created and supported in ONNX Runtime GenAI. The new solution is designed to be used in conjunction with [this work](https://github.com/microsoft/onnxruntime/pull/23549) in ONNX Runtime.

Some of the added changes include:
- Re-designed GenAI config that separates the encoder model and decoder model
  - Removes the encoder-decoder-init section
  - Creates a new encoder section
  - Separates session options, EP options, and model properties to be per-model instead of re-using the decoder's options for all components
  - Re-assigns pre-computed cross-attention KV caches as outputs to encoder model instead of inputs to decoder model
- Re-designed runtime support that makes the states and steps much clearer
  - Creates `AudioEncoder`, `WhisperDecoder` (i.e. `TextDecoder`), and `WhisperState` as separate states
  - Creates `AudioFeatures` class that can be re-used for other speech models
  - Adds generic support for FP32 CPU, FP32 CUDA, FP16 CUDA, and any quantized versions
  - Removes temporary workarounds for past-present buffer sharing due to restrictions from both the exported ONNX model and ONNX Runtime
  - Handles models with and without the following: buffer sharing, `DecoderMaskedMultiHeadAttention`, and alignment heads

### Known Issues

- [ ] This branch still has to be synced with the latest changes in the main branch of ONNX Runtime GenAI and active dev branches that will materially change this PR.
- [ ] The cross QK kernels do not have parity with the alternative, more-accurate approach to compute the cross QKs as a separate inference pass. Currently, it is recommended to use the alternative approach for calculating word-level timestamps.
- [ ] The cross QK kernels are only supported for CUDA.
- [ ] The end-to-end working example is still under development [here](https://github.com/openai/whisper/compare/main...kunal-vaishnavi:whisper:kvaishnavi/onnxruntime-genai). Once working, a copy of those scripts will be added as a sub-folder in the Python examples.

### Motivation and Context

The original implementation of Whisper was added in ONNX Runtime GenAI to create an initial foundation. This new approach is more flexible and more customizable for users. It also introduces an encoder-decoder architecture setup that can be used for other encoder-decoder models or other speech models.


test/python/_test_utils.py
```diff
@@ -52,6 +52,14 @@ def run_subprocess(
     return completed_process
 
 
+def get_ci_path():
```
baijumeswani: nit: use a more meaningful name.

```suggestion
def get_ci_data_path():
```

src/smartptrs.h
```diff
@@ -69,8 +69,8 @@ struct DeviceSpan {
   // Zero out the device memory
   void Zero() { p_device_memory_->Zero(); }
 
-  void CopyFrom(const DeviceSpan<const T>& source) {
-    assert(source.size() == size());  // Spans must be the same size to copy
+  void CopyFrom(const DeviceSpan<const T>& source, bool same_size = true) {
+    if (same_size) assert(source.size() == size());  // Spans must be the same size to copy
```
baijumeswani: When would same_size = false? In theory, you should only copy if the size of source and destination is the same.

src/python/python.cpp
```diff
@@ -507,6 +506,27 @@ PYBIND11_MODULE(onnxruntime_genai, m) {
             return processor.ProcessImagesAndAudios(prompt.value_or("").c_str(), images, audios);
           },
           pybind11::arg("prompt") = pybind11::none())
+      .def("__call__", [](OgaMultiModalProcessor& processor, const std::vector<std::string>& prompts, const pybind11::kwargs& kwargs) {
+        size_t count;
+        if (kwargs.contains("count"))
+          count = kwargs["count"].cast<size_t>();
+        else
+          throw std::runtime_error("Number of prompts must be provided. Use `processor(prompts, count=num_prompts)` to provide it.");
```
baijumeswani: Could you extract the count with `prompts.size()`?

Also, instead of creating this overload, shall we change the signature of the function above to support multiple prompts? You can do something like this:

```cpp
"__call__", [](OgaMultiModalProcessor& processor, pybind11::object prompts, const pybind11::kwargs& kwargs) {
    if (prompts.is_none()) {
        // no prompt was provided
    } else if (pybind11::isinstance<pybind11::str>(prompts) {
        // a single string was provided
    } else if (pybind11::isinstance<pybind11::list>(prompts) {
        // a list of strings was provided
    } else {
        // not supported argument type
    }
}, pybind11::arg("prompts") = pybind11::none(),
```

src/ort_genai_c.h
```diff
@@ -410,22 +397,38 @@ OGA_EXPORT void OGA_API_CALL OgaDestroyGenerator(OgaGenerator* generator);
...
 OGA_EXPORT bool OGA_API_CALL OgaGenerator_IsSessionTerminated(const OgaGenerator* generator);
 
+/**
+ * \brief For additional model inputs that genai does not handle, this lets the user set their values. For example LoRA models handle
+ * fine tuning through model inputs. This lets the user supply the fine tuning inputs, while genai handles the standard inputs.
+ * \param[in] generator The generator to add the inputs to.
+ * \param[in] name Name of the model input (this must match the model's input name)
+ * \param[in] tensor The OgaTensor of the input data
+ */
+OGA_EXPORT OgaResult* OGA_API_CALL OgaGenerator_SetModelInput(OgaGenerator* generator, const char* name, OgaTensor* tensor);
```
baijumeswani, line 407: Shall we deprecate this function? And ask users to use the NamedTensor function instead?

src/ort_genai_c.h
```diff
@@ -410,22 +397,38 @@ OGA_EXPORT void OGA_API_CALL OgaDestroyGenerator(OgaGenerator* generator);
...
+
+/**
+ * \brief For additional model inputs that genai does not handle, this lets the user set their values.
+ * \param[in] generator The generator to add the inputs to.
+ * \param[in] named_tensors The named tensors to set the inputs as.
+ */
+OGA_EXPORT OgaResult* OGA_API_CALL OgaGenerator_SetInputs(OgaGenerator* generator, const OgaNamedTensors* named_tensors);
```
baijumeswani, line 414: Avoid the `_` in the name?

```suggestion
OGA_EXPORT OgaResult* OGA_API_CALL OgaGeneratorSetInputs(OgaGenerator* generator, const OgaNamedTensors* named_tensors);
```

src/ort_genai_c.h
```diff
@@ -454,6 +457,16 @@ OGA_EXPORT OgaResult* OGA_API_CALL OgaGenerator_SetRuntimeOption(OgaGenerator* g
...
 
+/**
+ * \brief Returns a copy of the model input identified by the given name as an OgaTensor on CPU. The buffer is owned by returned OgaTensor
+ *       and will be released when the OgaTensor is destroyed
+ * \param[in] generator The generator to run the GetInput on the name provided and the out pointer to store the input.
+ * \param[in] name The name of the input tensor.
+ * \param[out] out The returned OgaTensor.
+ * \return OgaResult containing the error message if the computation failed.
+ */
+OGA_EXPORT OgaResult* OGA_API_CALL OgaGenerator_GetInput(const OgaGenerator* generator, const char* name, OgaTensor** out);
```
baijumeswani, line 468:
```suggestion
OGA_EXPORT OgaResult* OGA_API_CALL OgaGeneratorGetInput(const OgaGenerator* generator, const char* name, OgaTensor** out);
```

src/ort_genai_c.h
```diff
@@ -464,6 +477,17 @@ OGA_EXPORT OgaResult* OGA_API_CALL OgaGenerator_RewindTo(OgaGenerator* generator
  */
 OGA_EXPORT OgaResult* OGA_API_CALL OgaGenerator_GetOutput(const OgaGenerator* generator, const char* name, OgaTensor** out);
 
+/**
+ * \brief Returns a copy of the model output identified by the given name as an OgaTensor on CPU. The buffer is owned by returned OgaTensor
```
baijumeswani: Is this function returning the input or the output?

Can we separate them out?

src/ort_genai_c.h
```diff
@@ -464,6 +477,17 @@ OGA_EXPORT OgaResult* OGA_API_CALL OgaGenerator_RewindTo(OgaGenerator* generator
...
+ *       and will be released when the OgaTensor is destroyed
+ * \param[in] oga_generator The generator to run the GetInput or GetOutput method on the name provided and the out pointer to store the output.
+ * \param[in] name The name of the tensor.
+ * \param[in] is_input Whether the tensor name is for an input or not.
+ * \param[out] out The returned OgaTensor.
+ * \return OgaResult containing the error message if the computation failed.
+ */
+OGA_EXPORT OgaResult* OGA_API_CALL OgaGenerator_GetInputOutput(const OgaGenerator* oga_generator, const char* name, bool is_input, OgaTensor** out);
```
baijumeswani:
```suggestion
OGA_EXPORT OgaResult* OGA_API_CALL OgaGeneratorGetInput(const OgaGenerator* oga_generator, const char* name, OgaTensor** out);
```

examples/python/whisper.py
```diff
@@ -20,15 +20,25 @@ class Format:
...
     tokenizer = og.Tokenizer(model)
 
     while True:
         readline.set_completer_delims(" \t\n;")
         readline.parse_and_bind("tab: complete")
         readline.set_completer(_complete)
-        audio_paths = [audio_path.strip() for audio_path in input("Audio Paths (comma separated): ").split(",")]
+
+        if args.ci_test:
+            audio_paths = [args.audio]
```
baijumeswani: Can we use a similar flag as other examples `--non-interactive` instead of ci_test?

examples/c/src/whisper.cpp
```diff
@@ -182,8 +163,7 @@ void C_API(const char* model_path, int32_t num_beams) {
 
     OgaDestroyGenerator(generator);
     OgaDestroyGeneratorParams(params);
-    OgaDestroySequences(input_ids);
-    OgaDestroyNamedTensors(mel);
+    OgaDestroyNamedTensors(inputs);
```
baijumeswani, line 169: Shall we only keep the C++ example and remove the C one?

src/models/whisper_processor.cpp
```diff
@@ -23,20 +24,24 @@ std::unique_ptr<NamedTensors> WhisperProcessor::Process([[maybe_unused]] const T
   Ort::Allocator& allocator{Ort::Allocator::GetWithDefaultOptions()};
   auto named_tensors = std::make_unique<NamedTensors>();
 
+  // TODO: this is not working again...
   ort_extensions::OrtxObjectPtr<OrtxTensorResult> result;
   CheckResult(OrtxSpeechLogMel(processor_.get(), audios->audios_.get(), result.ToBeAssigned()));
```
sayanshaw24: could you use `OrtxFeatureExtraction` instead of `OrtxSpeechLogMel` here: https://github.com/microsoft/onnxruntime-extensions/blob/main/shared/api/c_api_feature_extraction.cc#L100?

it's more up to date and cleaner.

kunal-vaishnavi: Fixed

kunal-vaishnavi:
> General comments after first pass review
> 
> * is it possible to document all of the breaking changes? I couldn't quite tell from the description whether the config will have a breaking change. (Might be able to use copilot for this?)
> * with "Currently, it is recommended to use the alternative approach for calculating word-level timestamps," it is clear how a user would use the alternative approach. Would it be cleaner to say the word level timestamps are not supported yet?
> * with the incomplete end to end example from openai, what is remaining to do there?

I updated the PR description with more information to answer these questions.

test/python/_test_utils.py
kunal-vaishnavi: Fixed

examples/python/whisper.py
kunal-vaishnavi: Fixed

src/smartptrs.h
kunal-vaishnavi: For the `DecoderMaskedMultiHeadAttention` kernel, we need to transpose all K caches after they are computed for the first time. There are two sets of K caches, self attention's K caches and cross attention's K caches, and they are different sizes.

https://github.com/microsoft/onnxruntime-genai/blob/9bcc6812f741496e89919e2b09c6528d9b7a744f/src/models/whisper.cpp#L218-L226

We re-use the same temporary buffer to do the copies. The temporary buffer is the same size as the cross attention K caches so there will be copies for the self attention K caches where the sizes differ.

https://github.com/microsoft/onnxruntime-genai/blob/9bcc6812f741496e89919e2b09c6528d9b7a744f/src/models/whisper.cpp#L239-L241

src/python/python.cpp
kunal-vaishnavi: This can work for Python but not other languages where the input is a `const char**`. I created a new signature with `count` so that there would be consistency across language bindings.

src/ort_genai_c.h
kunal-vaishnavi: It may still be useful for Whisper with rewind or decoding with temperature fallback. For example, you can update the decoder's input ids without needing to set the other inputs and then re-generate the tokens with the same audio features, encoder hidden states, and cross-attention KV caches.

src/ort_genai_c.h
kunal-vaishnavi: I added the underscore to match the existing `OgaGenerator` API naming conventions.

src/ort_genai_c.h
kunal-vaishnavi: It is an internal function. The `GetInput` and `GetOutput` APIs use this one under the hood. The logic in both cases is almost identical so it is refactored into `GetInputOutput` to handle it.

examples/c/src/whisper.cpp
kunal-vaishnavi: I have been asked for C examples to run Whisper before so I think we can keep it for now.

src/smartptrs.h
baijumeswani: We probably should use subspan of the temp_span to help aligning the sizes:

```
temp_span.subspan(being, end).CopyFrom(dest_span)
```

Would that solve the problem? Copies should happen on the same data sizes, or it might result in other issues.

src/python/python.cpp
baijumeswani: I think you would certainly need a separate C API. I am suggesting changes only to the Python API so we don't redo the work in multiple calling sites.

src/models/whisper_processor.cpp
```diff
@@ -23,20 +24,24 @@ std::unique_ptr<NamedTensors> WhisperProcessor::Process([[maybe_unused]] const T
   Ort::Allocator& allocator{Ort::Allocator::GetWithDefaultOptions()};
   auto named_tensors = std::make_unique<NamedTensors>();
 
+  // TODO: this is not working again...
```
sayanshaw24: Fixed now in https://github.com/microsoft/onnxruntime-extensions/pull/969.

kunal-vaishnavi: Updated

src/models/decoder_only_pipeline.h
```diff
@@ -95,7 +94,6 @@ struct DecoderOnlyPipelineState : State {
   std::optional<WorkerThread> key_value_cache_update_worker_thread_{};
 
   std::unique_ptr<PositionInputs> position_inputs_;
-  ExtraInputs extra_inputs_{*this};
```
apsonawane: Do not need extra inputs here?

kunal-vaishnavi: Fixed

examples/python/whisper.py
```diff
@@ -20,15 +20,25 @@ class Format:
...
         readline.set_completer_delims(" \t\n;")
         readline.parse_and_bind("tab: complete")
         readline.set_completer(_complete)
-        audio_paths = [audio_path.strip() for audio_path in input("Audio Paths (comma separated): ").split(",")]
+
+        if args.non_interactive:
+            audio_paths = [args.audio]
+        else:
+            audio_paths = [audio_path.strip() for audio_path in input("Audio Paths (comma separated): ").split(",")]
         if len(audio_paths) == 0:
```
ajindal1, line 42: We need to check whether the `audio_paths[0]` is not empty, as len(audio_paths) would be 1 in both the scenarios if the input is empty strings.

```
>>> audio_paths = [audio_path.strip() for audio_path in input("Audio Paths (comma separated): ").split(",")]
Audio Paths (comma separated):
>>> len(audio_paths)
1
```

ajindal1: THinking more, it will be caught in the next step when the input string is empty but that would make this step redundant. Not a big deal, but wanted to let you know.

src/python/python.cpp
kunal-vaishnavi: I rewrote the batched preprocessing APIs to use `const OgaStringArray*` instead. This allows us to add the bindings for all languages (C, C++, C#, Java, Objective-C, Python) and mirror the existing preprocessing APIs.

test/python/test_onnxruntime_genai_e2e.py
```diff
@@ -38,6 +39,40 @@ def run_model(model_path: str | bytes | os.PathLike):
         assert generator.get_sequence(i) is not None
 
 
+# TODO: enable once ORT extensions fixes the audio features calculation
```
baijumeswani: Is this comment outdated?

src/ort_genai_c.h
baijumeswani: From a user facing api standpoint, I think it makes more sense to have explicit separate functions to get either input or output. Having the same function do both things may make things harder in the future if the underlying implementation changes. Or if we need to extend the internals of the function.

If possible, let's separate out the two.

src/models/processor.h
```diff
@@ -40,6 +40,7 @@ std::unique_ptr<Audios> LoadAudiosFromBuffers(std::span<const void*> audio_data,
 
 struct Payload {
   const std::string& prompt;
+  const std::span<const char*> prompts;
```
baijumeswani:
```suggestion
  std::span<const char*> prompts;
```

src/models/model.h
```diff
@@ -166,4 +167,41 @@ struct Model : std::enable_shared_from_this<Model>, LeakChecked<Model>, External
   std::map<std::string, std::unique_ptr<OrtSessionOptions>> pipeline_session_options_;
 };
 
+struct ModelType {
```
baijumeswani: Can this be a part of an unnamed namespace in model.cpp?

We could make the, constexpr std::array<const char*, NumModels> language_models = {
    "model_name",
    ...
}

And then use std::find. That way, we can avoid static initialization and make these compile time constants.

src/models/model.cpp
```diff
@@ -882,23 +882,18 @@ std::shared_ptr<Model> CreateModel(OrtEnv& ort_env, const char* config_path, con
 }
 
 std::shared_ptr<Model> CreateModel(OrtEnv& ort_env, std::unique_ptr<Config> config) {
-  const std::set<std::string> llm_types = {"chatglm", "decoder", "gemma", "gemma2", "gemma3_text",
-                                           "granite", "llama", "mistral", "nemotron", "olmo",
-                                           "phi", "phimoe", "phi3", "phi3small", "qwen2", "qwen3"};
   if (config->model.type == "gpt2")
     return std::make_shared<Gpt_Model>(std::move(config), ort_env);
-  if (llm_types.find(config->model.type) != llm_types.end())
+  if (ModelType::IsLLM(config->model.type))
```
baijumeswani, line 887: Shall we use full names instead `IsLanguageModel(...)` and `IsAudioLanguageModel(...)`?

src/models/kv_cache.h
```diff
@@ -69,10 +66,13 @@ struct CombinedKeyValueCache : KeyValueCache {
...
 
-  void AddEncoder() override;  // If model has an initial encoder step, this is used
-  // Register input_ids as ORT session input.
-  // Called only once during initialization of state.
+  static bool IsCacheNeeded(const Model& model);
+
   void Add() override;
+  auto& GetShape() const { return shape_; }
+  auto& GetType() const { return type_; }
+  auto& GetPresents() { return presents_; }
```
baijumeswani, line 72: Are these new functions needed as public methods? It becomes hard to use runtime polymorphism if they are only defined on some derived classes and not others.

Also, do these values need to be changed from outside? Shall we return copies instead of returning by reference?

src/java/src/main/native/utils.h
```diff
@@ -6,8 +6,13 @@
 
 #include <jni.h>
 #include <stdlib.h>
+#include <string.h>
```
baijumeswani, line 9: I could be mistaken, but i think this is a c header file. We probably cannot include string.h here.

src/generators.cpp
```diff
@@ -387,6 +353,36 @@ void Generator::AppendTokens(cpu_span<const int32_t> input_ids) {
   ComputeLogits(input_ids_device);
 }
 
+void Generator::SetInputs(const NamedTensors& named_tensors) {
+  if (ModelType::IsLLM(model_->config_->model.type) || ModelType::IsPipe(model_->config_->model.type)) {
+    throw std::runtime_error("Please use generator.AppendTokens for " + model_->config_->model.type + ". SetInputs is not supported for this model type.");
+  }
```
baijumeswani, line 357: Why should this not be supported? We should be able to support in case the model has extra inputs, right?

src/generators.cpp
```diff
@@ -364,8 +324,8 @@ void Generator::AppendTokens(cpu_span<const int32_t> input_ids) {
     throw std::runtime_error("input_ids is empty");
   if ((input_ids.size() / state_->params_->search.batch_size) + search_->GetSequenceLength() > state_->params_->search.max_length)
     throw std::runtime_error("input_ids size (" + std::to_string(input_ids.size()) + ") + current sequence length (" + std::to_string(search_->GetSequenceLength()) + ") exceeds max length (" + std::to_string(state_->params_->search.max_length) + ")");
-  if (model_->config_->model.type == "whisper" || model_->config_->model.type == "phi3v")
-    throw std::runtime_error("Please use params.SetInputs for " + model_->config_->model.type + ". AppendTokens is not supported for this model type.");
+  if (!ModelType::IsLLM(model_->config_->model.type) && !ModelType::IsALM(model_->config_->model.type) && !ModelType::IsPipe(model_->config_->model.type))
```
baijumeswani: Should we check for inclusion instead of exclusion?

Only check for MultiModalModels? Since whisper is a audio language model, would that make this function not available for whisper? It was previously marked as being valid for whisper.

src/config.h
```diff
@@ -233,24 +243,26 @@ struct Config {
 
   } model;
 
+  // clang-format off
```
baijumeswani: Avoid setting this formatting off is possible. Could you werite the comments so they look cleaner. Maybe move them to the top of the member name instead of it being after the member name

src/generators.cpp
aciddelgado: With this API, the idea would be that SetInputs would be used for Multimodal models which have extra inputs and AppendTokens would be used for text-only models.

src/generators.cpp
```diff
@@ -387,6 +353,36 @@ void Generator::AppendTokens(cpu_span<const int32_t> input_ids) {
...
+
+  // Set any extra inputs (those defined in extra_inputs and those defined in the PresetExtraInputs registry)
+  if (set_extra_inputs_) {
+    state_->SetExtraInputs(extra_inputs);
+    set_extra_inputs_ = false;
+  }
+
+  // Append tokens and run ComputeLogits after setting all other possible inputs
+  if (input_ids.size() > 0) {
+    AppendTokens(input_ids);
```
aciddelgado, line 380: Won't this throw an error when AppendTokens checks for model type?

src/generators.cpp
kunal-vaishnavi: Yes, exactly. The `AppendTokens` API will be for decoder-only models and the `SetInputs` API will be for anything else.

src/smartptrs.h
kunal-vaishnavi: Updated to use `subspan` and removed the changes in `CopyFrom`.

src/ort_genai_c.h
kunal-vaishnavi: I moved the `GetInputOutput` method into its own namespace and removed it from the C API header.

examples/python/whisper.py
kunal-vaishnavi: Thanks for letting me know. This step is mainly to separate the CI run and the user run.

src/models/processor.h
kunal-vaishnavi: Fixed

src/models/model.h
kunal-vaishnavi: Changed this and used `std::array<std::string_view>` because `find` with `const char*` will only compare if the pointers point to the exact same memory.

src/models/model.cpp
kunal-vaishnavi: I was keeping this as acronyms since the conditional expressions can get quite long otherwise. The full names are also noted in the comments where the arrays of model types are defined.

src/models/kv_cache.h
kunal-vaishnavi: They are all used in Whisper. The type is constant but the shapes can be modified and the KV caches in `presents_` are modified.

src/generators.cpp
kunal-vaishnavi: I now removed the check instead because multi-modal models such as Whisper and Phi vision use `SetInputs`, which will also call `AppendTokens` internally.

Previously, Whisper required the user to call both `SetInputs` and `AppendTokenSequences` in that order. This can be confusing to a user who expects all inputs to be set with `SetInputs`. Now, Whisper only needs to use `SetInputs`.

src/config.h
kunal-vaishnavi: Fixed

src/generators.cpp
kunal-vaishnavi: I now removed the check inside `AppendTokens` for model type because Whisper and Phi vision use `SetInputs`, which will also call `AppendTokens` internally.

src/java/src/main/native/utils.h
kunal-vaishnavi: This is a C-only header file. Both `strdup` and `memcpy` are imported from the [`<string.h>` C header file](https://en.cppreference.com/w/c/header/string.html).

test/python/test_onnxruntime_genai_e2e.py
kunal-vaishnavi: Activated the CI tests now

src/generators.h
```diff
@@ -115,15 +100,20 @@ struct Generator : LeakChecked<Generator> {
 
   DeviceSpan<int32_t> GetSequence(size_t index) const;
 
+  // A list of extra model inputs that will be matched at runtime based on name
+  std::vector<ExtraInput> extra_inputs;
```
baijumeswani:
```suggestion
  std::vector<ExtraInput> extra_inputs_;
```

src/models/kv_cache.h
```diff
@@ -7,13 +7,13 @@
 
 namespace Generators {
 
+std::string ComposeKeyValueName(const std::string& template_string, int index);
```
baijumeswani: Why is this needed? Is the function not already declared at the end of this file?

src/models/model_type.h
```diff
@@ -0,0 +1,49 @@
...
+#include <algorithm>
+#include <array>
+#include <string>
+#include <string_view>
+
+namespace Generators {
+
+struct ModelType {
+  // Large-language model (LLM)
+  static constexpr std::array<std::string_view, 17> LLM = {"chatglm", "decoder", "gemma", "gemma2", "gemma3_text", "gpt2", "granite", "llama", "mistral", "nemotron", "olmo", "phi", "phimoe", "phi3", "phi3small", "qwen2", "qwen3"};
```
baijumeswani: Shall we define these arrays inside the function, so they're not exposed?

kunal-vaishnavi: Fixed

src/models/kv_cache.h
kunal-vaishnavi: Looks like it got duplicated while resolving merge conflicts. Removed it.

src/generators.h
kunal-vaishnavi: Fixed

baijumeswani (dismissed):

src/models/audio_features.cpp
```diff
@@ -0,0 +1,34 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+#include "../generators.h"
+#include "model.h"
+#include "audio_features.h"
+
+namespace Generators {
+
+AudioFeatures::AudioFeatures(State& state, const std::string& name, const std::vector<ExtraInput> extra_inputs)
```
hanbitmyths: Can extra_inputs be a reference?

kunal-vaishnavi: Fixed

hanbitmyths (approved):

PR #1236, opened 2025-02-09
Update benchmark e2e test to use chat template

apsonawane:
Currently benchmark e2e did not use chat template, update it to use it


benchmark/python/benchmark_e2e.py
```diff
@@ -283,7 +296,7 @@ def run_benchmark(args, batch_size, prompt_length, generation_length, max_length
 
         # Measure tokenization
         tokenize_start_time = time.perf_counter()
-        tokens = tokenizer.encode_batch(prompt)
+        tokens = tokenizer.encode(prompt)
```
kunal-vaishnavi, line 299: Why use `encode` instead of `encode_batch`?

apsonawane: Because prompt is a string and encode_batch takes List[str] as input

kunal-vaishnavi (approved):

PR #1250, opened 2025-02-13
Pin NDK version

baijumeswani:
Android NDK version 28.x uses LLVM 19+ which is not compatible with dlib. dlib is an ort-extensions dependency that gets built when building ort-genai.

As a workaround, pinning ndk version to 27.x.

More details here: https://github.com/davisking/dlib/issues/3045


snnn (approved):
ORT is also using NDK 27 right now.

snnn:
It seems like there is  an issue in android_x64's yaml file.

baijumeswani:
> It seems like there is an issue in android_x64's yaml file.

The runner.temp is not available before the run starts. So, had to move it to later during the run

snnn (approved):

PR #1256, opened 2025-02-15
Update version to 0.7.0

baijumeswani:


natke (commented):
Do we want the samples to update to main? I just added a PR to update them to the release. Then a user can clone the repo and build and run the sample

ajindal1:
> Do we want the samples to update to main? I just added a PR to update them to the release. Then a user can clone the repo and build and run the sample

I believe the main branch should have the mapping to the current nightly version and the release branch should have the version mapped to the release version. So, I think what Baiju is doing is correct.

ajindal1 (approved):

PR #1257, opened 2025-02-15
Add chat example in C#

natke:


examples/csharp/HelloPhi/Program.cs
```diff
@@ -143,4 +144,23 @@ void PrintUsage()
...
+            while (!generator.IsDone())
+            {
+                generator.GenerateNextToken();
+                Console.Write(tokenizerStream.Decode(generator.GetSequence(0)[^1]));
+            }
+            Console.WriteLine();
+            Console.WriteLine("Next prompt:");
+            prompt = Console.ReadLine();
+            sequences = tokenizer.Encode($"<|user|>{prompt}<|end|><|assistant|>");
+        } while (prompt != null);
```
baijumeswani, line 163: Could we avoid writing a while loop within another while loop? To get out of the while loop, we need to exit both the while loops. And it seems like the purpose for both the while loops is the same.

natke: Actually the outer while loop is a new conversation

natke:
Superseded by #1266

PR #1263, opened 2025-02-19
Avoid downloading HF models in CI pipelines

baijumeswani:
This pull-request makes changes to the tests so that the HuggingFace models are not downloaded if these are already available on the machine under some directory. The directory it expects the models to be located in is

- Windows: C:\data\models\ortgenai\pytorch
- Linux: /data/ortdenai/pytorch

If the models are not found in that location, the test script will download from huggingface_hub.

The models are not made available on macos as we don't have a way to add those models to the macos vm yet.

This pull-request also changes how many models are converted in the CI pipeline to onnx to being only phi-2, and qwen-2.5 for now. I'll create a separate pull-request to enable more models for the nightly build.


ajindal1 (approved):

test/python/_test_utils.py
```diff
@@ -53,33 +53,40 @@ def run_subprocess(
 
 
 def get_model_paths():
+    # TODO: Uncomment the following models as needed in the CI pipeline.
+
```
kunal-vaishnavi, line 57: How will we run these tests in a nightly CI if the models are commented out here?

baijumeswani: Will submit another PR for that e2e pipeline (which is the nightly pipeline will include more models). But the CI that run on every PR should not run all these models. Plus there are other issues with the models being copied in the VM (disk space being the main one).

RyanUnderhill:
:shipit:

RyanUnderhill (approved):

PR #1264, opened 2025-02-19
Remove onnxruntime-genai-static library (non trivial change)

RyanUnderhill:
Rewrites the python bindings and our unit tests to purely go through the public C API so that we no longer need to statically link with onnxruntime-genai for these cases.

Adds new C APIs to fill in the API gaps, and fixes a few issues along the way.


src/ort_genai.h
```diff
@@ -64,6 +65,80 @@ inline void OgaCheckResult(OgaResult* result) {
   }
 }
 
+struct OgaFloat16_t;
+struct OgaBFloat16_t;
+
+// Variable templates to convert a C++ type into it's ONNXTensorElementDataType
```
aciddelgado, line 71:
```suggestion
// Variable templates to convert a C++ type into it's OgaElementType
```

RyanUnderhill: True, they are the same numbers but we don't want people to need to include the onnxruntime headers to use our headers.

aciddelgado (approved):

aciddelgado (approved):

src/ort_genai.h
```diff
@@ -9,6 +9,7 @@
 
 #if __cplusplus >= 202002L
 #include <span>
+#define OGA_USE_SPAN 1
```
baijumeswani, line 12: Should this always be 1?

RyanUnderhill: It needs to be conditional in the public header, as users won't have our "span.h" if their compiler doesn't have it.

baijumeswani (approved):

PR #1267, opened 2025-02-20
Add possibility to change graph optimization level

benHeid:
This PR adds the option to set the graph optimization level when running a model.

See discussion in https://github.com/microsoft/onnxruntime-genai/discussions/1260#discussioncomment-12247562

(Disclaimer: I am not very familiar with cpp, so any hints for improving my changes are very very welcome)


benHeid:
@microsoft-github-policy-service agree [company="Mercedes-Benz Tech Innovation"]

PR #1286, opened 2025-02-27
support audio model

rui-ren:
###  Chat_app for Omni model support


examples/chat_app/app.py
```diff
@@ -9,14 +9,15 @@
 from app_modules.utils import cancel_outputing, delete_last_conversation, reset_state, reset_textbox, transfer_input
 from interface.hddr_llm_onnx_interface import ONNXModel
 from interface.multimodal_onnx_interface import MultiModal_ONNXModel
+from interface.omnimodel_onnx_interface import Omni_ONNXModel
```
baijumeswani, line 12: Let's call is MultiModal instead.

PR #1305, opened 2025-03-07
Add support for DML Graph Capture to improve speed

aciddelgado:
Graph Capture support was removed in a previous PR. This adds it back for decoder-only and multi-modal models, while also making the implementation more flexible and readable. These changes provide a massive speed-up for DML.

Benchmarks on RTX4090 using the following command `python benchmark_e2e.py -i ~\test_models\phi-3.5-mini-12_05_24\gpu\gpu-int4-awq-block-128 -e dml -r 100` yielded the following results:

Before the change:
`Average Prompt Processing Throughput (per token): 108.91014399406912 tps`
`Average Token Generation Throughput (per token): 29.935862983984 tps`

After the change:
`Average Prompt Processing Throughput (per token): 177.03018977643373 tps`
`Average Token Generation Throughput (per token): 204.9705547619224 tps`

The improvements mostly affect token generation, improving it by almost 10x.

This PR removes the TryCudaGraphCapture() API, as it is redundant. It also temporarily disables CUDA Graph Capture as it is currently nonfunctional.


src/dml/interface.cpp
```diff
@@ -207,6 +208,118 @@ struct InterfaceImpl : DeviceInterface {
 
   void Synchronize() override {
   }
+
+  // TODO: perform operation directly on DML
```
RyanUnderhill, line 212: Is this a copy of the CPU code? The default implementation for this method just returns false, which means 'Do it on CPU'.

src/dml/interface.cpp
```diff
@@ -207,6 +208,118 @@ struct InterfaceImpl : DeviceInterface {
...
+        data[i] = i + total_length - new_kv_length;
+    } else {
+      // For batch size > 1 we increment position ids by 1... continuous decoding is not supported
+      for (int i = 0; i < batch_beam_size; i++)
+        data[i]++;
+    }
+    data_span.CopyCpuToDevice();
+  }
+
+  void UpdatePositionIds(void* position_ids, int batch_beam_size, int total_length, int new_kv_length, ONNXTensorElementDataType type) override {
```
RyanUnderhill, line 268: Perhaps this should be like Copy, just return false to mean 'do it on CPU'. We want to avoid duplicating the CPU implementations.

src/generators.h
```diff
@@ -49,7 +50,14 @@ DeviceSpan<T> WrapTensor(DeviceInterface& device, OrtValue& value) {
   return device.WrapMemory(std::span<T>{value.GetTensorMutableData<T>(), info->GetElementCount()});
 }
 
+template <typename T>
```
RyanUnderhill, line 53: Would it be better if OgaValue had a 'GetDeviceSpan' method on it?

RyanUnderhill (commented):
Some initial comments

src/generators.h
aciddelgado: This is a good idea

src/dml/interface.cpp
aciddelgado: this copies the CPU code, the issue is that we can't just do it on CPU, we need to copy memory from device, back to cpu, back to device. It may be possible to refactor and do it within position_inputs (like before), but it would similarly use DeviceSpans to handle the copying (on CPU the copy calls would be non-ops which might be misleading to read but maybe who cares)... thoughts?

src/oga_value.h
```diff
@@ -0,0 +1,62 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+#pragma once
+#include "models/onnxruntime_api.h"
+#include "smartptrs.h"
+
+namespace Generators {
+
+struct OgaValue : LeakChecked<OgaValue> {
```
baijumeswani, line 9: Could you add some comments to explain why ogavalue is needed (how it is different from ortvalue)?

Also, not related to this PR, could you add some comments somewhere to help people understand differences between OgaValue, OrtValue, Generators::Tensor? They all seem to represent a tensor, still we need three different types. It might be nice to see an explanation on why we need different types.

baijumeswani, line 9: Why OgaValue? Can we call it something else?

baijumeswani (approved):

src/oga_value.h
aciddelgado: I have folded OgaValue into Generators::Tensor for simplifaction purposes. Internally, we need Generators::Tensor to support static buffers and to link a tensor to its corresponding device; these enable graph capture and simplify the code respectively.

baijumeswani (approved):

PR #1338, opened 2025-03-21
deprecate graph cap

aciddelgado:
Deprecate graph cap instead of outright removing it.

### DIFF
--- a/.pipelines/android-publishing.yml
+++ b/.pipelines/android-publishing.yml
@@ -14,9 +14,26 @@ parameters:
   - 'Release'
   - 'MinSizeRel'
 
+resources:
+  repositories:
+  - repository: 1esPipelines
+    type: git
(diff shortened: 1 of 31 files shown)


RyanUnderhill (commented):
This adds it back fully? It doesn't look like a stub that just reports an error/warning

RyanUnderhill:
I mean, it's fine either way. I just wasn't sure if we were trying to have things run with warnings vs errors.

natke:
> I mean, it's fine either way. I just wasn't sure if we were trying to have things run with warnings vs errors due to missing functions.

I think we agreed it would be a stub

natke (commented):
Was there a Java API too?

baijumeswani (dismissed):

baijumeswani (approved):

PR #1343, opened 2025-03-24
Update torch to 2.5.1

baijumeswani:


test/python/cpu/torch/requirements.txt
```diff
@@ -1,2 +1,2 @@
 -f https://download.pytorch.org/whl/torch_stable.html
```
ajindal1, line 1: You need to provide index_url for newer versions:
For example, for cpu:
--index-url https://download.pytorch.org/whl/cpu
For CUDA 124:
--index-url https://download.pytorch.org/whl/cu124

ajindal1 (approved):

PR #1345, opened 2025-03-25
Windows: Show more useful DLL load errors to say exactly what DLL is missing

RyanUnderhill:
### Description
When we fail to load a provider shared DLL in windows, the error is not very specific. Users have to figure out if the onnxruntime file is missing, a cuda file, or cudnn is not installed (and perhaps others). And this is just the cuda provider. It would be far more useful if it would say exactly what file is missing so the user can fix the actual problem.

Plus, this will likely result in many fewer github issues regarding this problem, but if they do, they will be much easier to fix.

This fix adds a function that will try loading a dll and its dependencies recursively to figure out which file is missing. It uses the OS dbghelp library to do it and is not very complex.

Almost the same as what I added for onnxruntime recently: 

https://github.com/microsoft/onnxruntime/pull/24053


src/dll_load_error.h
```diff
@@ -0,0 +1,3 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+std::string DetermineLoadLibraryError(const char* filename);
```
baijumeswani, line 3: Should this be inside the `Generators` namespace?

baijumeswani (approved):

src/dll_load_error.h
RyanUnderhill: Hmm, good point. It's not really Generators code, and now that we're shared library only the namespace is actually redundant (just makes all of our symbol names longer while debugging).

RyanUnderhill: The JSON code isn't in the Generators namespace either, so I'll leave it outside.

PR #1346, opened 2025-03-25
Support load/unload of models to avoid QNN errors on deepseek r1 1.5B

baijumeswani:
The deepseek r1 1.5B model will fail to run on NPU due to memory issues. This PR adds support for loading and unloading models to avoid QNN errors if the genai config requests it.

The alternative is to have the driver update for the QNN NPU but I am not sure when that will be rolled out.

### DIFF
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -230,6 +230,8 @@ struct PipelineModel_Element : JSON::Element {
       v_.run_on_prompt = JSON::Get<bool>(value);
     } else if (name == "run_on_token_gen") {
       v_.run_on_token_gen = JSON::Get<bool>(value);
+    } else if (name == "reset_session_idx") {
+      v_.reset_session_idx = static_cast<int>(JSON::Get<double>(value));
     } else
       throw JSON::unknown_value_error{};
   }
--- a/src/config.h
+++ b/src/config.h
@@ -192,6 +192,7 @@ struct Config {
         std::unordered_map<std::string, std::string> output_names_forwarder;
         bool run_on_prompt{true};
         bool run_on_token_gen{true};
+        int32_t reset_session_idx{-1};
       };
 
       std::vector<PipelineModel> pipeline;
--- a/src/config.h
+++ b/src/config.h
@@ -192,8 +192,10 @@ struct Config {
         std::unordered_map<std::string, std::string> output_names_forwarder;
         bool run_on_prompt{true};
         bool run_on_token_gen{true};
-        int reset_session_idx{-1};  // Some models cannot keeo all the ort sessions in memory at once due to memory constraints.
+        int reset_session_idx{-1};  // Some models cannot keep all the ort sessions in memory at once due to memory constraints.
                                     // This is the index of the session that needs to be reset during the execution of the current session.
+                                    // This is a temporary solution until the QNN driver updates are available.
+                                    // Once the driver updates are available, this option will be deprecated.
       };
 
       std::vector<PipelineModel> pipeline;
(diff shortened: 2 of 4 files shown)


src/config.h
```diff
@@ -192,6 +192,7 @@ struct Config {
         std::unordered_map<std::string, std::string> output_names_forwarder;
         bool run_on_prompt{true};
         bool run_on_token_gen{true};
+        int32_t reset_session_idx{-1};
```
RyanUnderhill, line 195: Why is it specifically an int32_t? (other places use int).

src/config.cpp
```diff
@@ -230,6 +230,8 @@ struct PipelineModel_Element : JSON::Element {
       v_.run_on_prompt = JSON::Get<bool>(value);
     } else if (name == "run_on_token_gen") {
       v_.run_on_token_gen = JSON::Get<bool>(value);
+    } else if (name == "reset_session_idx") {
```
RyanUnderhill, line 233: Is 'reset_session_idx' intuitive enough? I always get confused by the smart pointer reset/empty/clear naming. Would something like 'deallocate_session_index' be clear that the goal is to free memory?

src/models/decoder_only_pipeline.cpp
```diff
@@ -83,6 +83,11 @@ bool IntermediatePipelineState::SupportsPrimaryDevice() const {
 
 DeviceSpan<float> IntermediatePipelineState::Run(int total_length, DeviceSpan<int32_t>& next_tokens,
                                                  DeviceSpan<int32_t> next_indices) {
+  if (!model_.sessions_[id_]) {
+    const_cast<DecoderOnlyPipelineModel*>(&model_)->sessions_[id_] =
+        OrtSession::Create(model_.ort_env_, (model_.config_->config_path / fs::path(model_.config_->model.decoder.pipeline[id_].filename)).c_str(),
+                           model_.GetSessionOptions(model_.config_->model.decoder.pipeline[id_].model_id));
+  }
```
RyanUnderhill, line 90: How long is the OrtSession lifetime? Is it just freed after a single State::Run() call here after RunPipeline?

baijumeswani: There are two models called `context` (runs on prompt/prefill stage) model and the `iterator` (runs on token generation stage) model.

Here is the lifetime of the two sessions:

- `Model` gets created. All sessions are created.
- `Context` session gets invoked. It has the `reset_session_idx` set to the `iterator` session. So, the iterator session gets dropped.
- `Iterator` session gets invoked. It was previously dropped, so a new session gets created for the run.
- `Iterator` session defines the `reset_session_idx` as the one from the context stage. So, the `context` sessino gets dropped.

That outlines the lifetime of the sessions for 1 generation. It is quite hacky, which is why I didn't push the code earlier. But it seems that this would facilitate running the model until the QC driver fix goes in.

src/config.cpp
baijumeswani: I would actually like to not have this exposed at all, and would like it to be a temporary solution (with the intention of deprecating the config in the future). But for now, some models are already using this config. So, probably not a good idea to change the name. I can add some comments in the config to explain what this option means.

src/config.h
baijumeswani: Will update.

src/config.h
```diff
@@ -192,7 +192,8 @@ struct Config {
         std::unordered_map<std::string, std::string> output_names_forwarder;
         bool run_on_prompt{true};
         bool run_on_token_gen{true};
-        int32_t reset_session_idx{-1};
+        int reset_session_idx{-1};  // Some models cannot keeo all the ort sessions in memory at once due to memory constraints.
```
RyanUnderhill, line 195: Small typo: keeo -> keep

Should you put a comment saying this is a temporary workaround and will be ignored once the proper fix in QNN is in?

PR #1348, opened 2025-03-25
re-enable debugging for dml

aciddelgado:
this was removed in a prior change.

### DIFF
--- a/src/models/debugging.cpp
+++ b/src/models/debugging.cpp
@@ -88,11 +88,23 @@ void DumpTensor(const Model& model, std::ostream& stream, OrtValue* value, bool 
     if (dump_value) {
       DumpValues(stream, type_info->GetElementType(), value->GetTensorRawData(), element_count);
     }
-  } else if (memory_info.GetDeviceType() == OrtMemoryInfoDeviceType_GPU || memory_info.GetDeviceType() == 4) {
-    if (memory_info.GetDeviceType() == OrtMemoryInfoDeviceType_GPU)
-      stream << "GPU\r\n";
-    else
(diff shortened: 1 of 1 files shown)


baijumeswani (dismissed):

src/models/debugging.cpp
```diff
@@ -88,11 +88,23 @@ void DumpTensor(const Model& model, std::ostream& stream, OrtValue* value, bool
...
-  } else if (memory_info.GetDeviceType() == OrtMemoryInfoDeviceType_GPU || memory_info.GetDeviceType() == 4) {
-    if (memory_info.GetDeviceType() == OrtMemoryInfoDeviceType_GPU)
-      stream << "GPU\r\n";
-    else
-      stream << "DML\r\n";
+    // Internally there are 5 device types defined in onnxruntime but only 3 are exposed in the public API
+    // https://github.com/microsoft/onnxruntime/blob/9dbfee91ca9c2ba2074d19805bb6dedccedbcfe3/include/onnxruntime/core/framework/ortdevice.h#L15
+  } else if (memory_info.GetDeviceType() < 5) {
+    switch (model.p_device_->GetType()) {
+      case DeviceType::CUDA:
```
baijumeswani, line 95: Shall we try to define a `model.p_device_->GetType<std::string>()` for each device type? We can avoid the switch statement here then.

RyanUnderhill: I think we can fairly safely say that if the tensor is not on CPU, it's on the model.p_device_

baijumeswani:
i believe this change might have been reverted in ort. closing this.

PR #1351, opened 2025-03-27
Update Nuget pkg name

ajindal1:
Update Nuget pkg name as the package name is same for two scenarios

### DIFF
--- a/.pipelines/stages/jobs/nuget-packaging-job.yml
+++ b/.pipelines/stages/jobs/nuget-packaging-job.yml
@@ -87,7 +87,7 @@ jobs:
       value: 'MinSizeRel'
 
   - name: artifactName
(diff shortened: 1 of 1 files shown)


baijumeswani:
Does this change the name of the packages (*.zip and *.nuget) by any chance?

ajindal1:
> Does this change the name of the packages (*.zip and *.nuget) by any chance?

I am not sure about this, but it is possible because this PR changes the artifactName variable which is used to publish in the Publish Pipeline Artifact [step](https://github.com/microsoft/onnxruntime-genai/blob/1c3573525fce874a446239a6d3013bdfeca82e7d/.pipelines/stages/jobs/nuget-packaging-job.yml#L325). If this name is used to generate the name of the packages, then definitely they will be affected.

baijumeswani (approved):

PR #1352, opened 2025-03-27
Add value_stats logging option to show statistical information about …

RyanUnderhill:
Adds a new logging option "value_stats" that is on by default. This will show 'value_stats' after any array of values is also logged.

So when tensor inputs/output data is shown, the 'value_stats' will be shown after them. Also for things like 'model_logits' as below:

```
  model_logits   Values[ 2.51172 4.83594 6.32422 1.30469 1.97949 ... -1.28418 -1.28418 -1.28711 -1.28516 -1.28223 ]
  value_stats    Min: -5.35156 Max: 8.35156 Mean: 0.933448 StdDev: 1.1297
```

image

This will also detect non-finite values and say the index of the first value similar to this image: (was renamed to just 'value_stats' but the rest is accurate)

image

This also generalizes the Float16ToFloat32 code to be templatized on the number of bits per field. This makes the implementation easier to read (no more magic numbers) and also adds bfloat16 support trivially.

### DIFF
--- a/src/models/debugging.cpp
+++ b/src/models/debugging.cpp
@@ -6,7 +6,59 @@
 #include "model.h"
 
 namespace Generators {
-static constexpr size_t c_value_count = 10;  // Dump this many values from the start of a tensor
+static constexpr size_t c_value_count = 10;  // Dump this many values at the start & end of a tensor (with '...' in between)
+
+// Tensor value statistics to help easily eyeball if the values in a tensor are reasonable.
+struct Stats {
+  float min = std::numeric_limits<float>::max();
+  float max = std::numeric_limits<float>::lowest();
+  float sum{};
+  float sum_of_squares{};
+  size_t count{};
+
+  bool found_non_finite {};
+  size_t first_non_finite_index{};
+  float non_finite_value{};
+  size_t non_finite_count{};
+
+  void Dump(std::ostream& stream) const {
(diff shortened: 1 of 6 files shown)


src/models/debugging.cpp
```diff
@@ -61,6 +113,30 @@ void DumpValues(std::ostream& stream, ONNXTensorElementDataType type, const void
...
+  if (g_log.value_stats) {
+    Stats stats;
+    if (type == Ort::TypeToTensorType<float>) {
+      auto p_values = reinterpret_cast<const float*>(p_values_raw);
+      for (size_t i = 0; i < count; i++)
+        stats << p_values[i];
+    } else if (type == Ort::TypeToTensorType<double>) {
+      auto p_values = reinterpret_cast<const double*>(p_values_raw);
+      for (size_t i = 0; i < count; i++)
+        stats << static_cast<float>(p_values[i]);
```
aciddelgado, line 126: possible to introduce inf values when making this conversion?

RyanUnderhill: Yep, and then we'll report that we did in the stat class. We convert to float for our scoring, so this is good.

RyanUnderhill: (also, I don't think any models use doubles for logits)

aciddelgado (approved):

PR #1354, opened 2025-03-27
Switch the order of softmax on CPU Top K

aciddelgado:
May provide a slight efficiency boost.


src/search.cpp
```diff
@@ -167,6 +166,7 @@ void GreedySearch_Cpu::SampleTopK(int k, float temperature) {
     for (int i = 0; i < k; i++)
       top_k_scores[i] = scores[indices[i]];
     // Sample a token from the top K
+    SoftMax(top_k_scores, temperature);
```
tianleiwu, line 169: Softmax

tianleiwu (approved):

PR #1358, opened 2025-03-28
Convert tokens to list for concat

ajindal1:
Convert tokens to list for concat in Q&A example

### DIFF
--- a/examples/python/model-qa.py
+++ b/examples/python/model-qa.py
@@ -99,7 +99,7 @@ def main(args):
         if args.verbose: print("Generator created")
 
         # Append system and input tokens to the generator
-        generator.append_tokens(system_tokens + input_tokens)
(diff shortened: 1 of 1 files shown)


examples/python/model-qa.py
```diff
@@ -99,7 +99,7 @@ def main(args):
         if args.verbose: print("Generator created")
 
         # Append system and input tokens to the generator
-        generator.append_tokens(system_tokens + input_tokens)
+        generator.append_tokens(list(system_tokens) + list(input_tokens))
```
RyanUnderhill, line 102: why not `np.concatenate([system_tokens, input_tokens])`?

ajindal1: yep, makes more sense.

RyanUnderhill (approved):

baijumeswani (approved):

PR #1365, opened 2025-04-01
Simplify the C API definitions and prevent any type mismatches going forward

RyanUnderhill:
(no mismatches found in this change, but it was easy to accidentally do without error before this)

### DIFF
--- a/src/ort_genai_c.cpp
+++ b/src/ort_genai_c.cpp
@@ -29,13 +29,46 @@ const char* AllocOgaString(const std::string& string) {
   return cstr_buffer.release();
 }
 
+// As the Oga* types are just typedefs, we can use them as the actual types in the C API.
+// We still need to cast from internal types to the external ones, but these definitions ensure that the types are correct.
+// But do not use reinterpret_cast!
+struct OgaAdapters : Generators::Adapters {};
+struct OgaAudios : Generators::Audios {};
+struct OgaConfig : Generators::Config {};
+struct OgaGenerator : Generators::Generator {};
+struct OgaGeneratorParams : Generators::GeneratorParams {};
+struct OgaImages : Generators::Images {};
+struct OgaModel : Generators::Model {};
+struct OgaMultiModalProcessor : Generators::MultiModalProcessor {};
+struct OgaNamedTensors : Generators::NamedTensors {};
+struct OgaResult : Generators::Result {};
+struct OgaRuntimeSettings : Generators::RuntimeSettings {};
+struct OgaSequences : Generators::TokenSequences {};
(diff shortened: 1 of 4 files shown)


src/ort_genai_c.cpp
```diff
@@ -29,21 +29,54 @@ const char* AllocOgaString(const std::string& string) {
...
+struct OgaModel : Generators::Model {};
+struct OgaMultiModalProcessor : Generators::MultiModalProcessor {};
+struct OgaNamedTensors : Generators::NamedTensors {};
+struct OgaResult : Generators::Result {};
+struct OgaRuntimeSettings : Generators::RuntimeSettings {};
+struct OgaSequences : Generators::TokenSequences {};
+struct OgaStringArray : std::vector<std::string> {};
+struct OgaTensor : Generators::Tensor {};
+struct OgaTokenizer : Generators::Tokenizer {};
+struct OgaTokenizerStream : Generators::TokenizerStream {};
```
skottmckay, line 50: will this be fragile if any of these structs change to require proper initialization and no longer have a default ctor. 

maybe one day the struct internals will be private so they can control their invariants and make the code safer, and when doing that a default ctor that doesn't fully initialize the struct into a valid state might go away. :P

RyanUnderhill: None of these types are constructed in this API, and if someone tried you'd surely get a compile error. They're just to tell the compiler that a pointer to one is really a pointer to the others, and that we can static_cast the other way.

This is the basis of the trick here, the Oga* types don't exist, they're really the internal types, and this lets us tell the compiler this in one single place.

src/ort_genai_c.h
```diff
@@ -547,7 +547,7 @@ OGA_EXPORT OgaResult* OGA_API_CALL OgaCreateNamedTensors(OgaNamedTensors** out);
  * \param[out] out The tensor with the given name
  * \return OgaResult containing the error message if the tensor with the given name could not be found.
  */
-OGA_EXPORT OgaResult* OGA_API_CALL OgaNamedTensorsGet(const OgaNamedTensors* named_tensors, const char* name, OgaTensor** out);
+OGA_EXPORT OgaResult* OGA_API_CALL OgaNamedTensorsGet(OgaNamedTensors* named_tensors, const char* name, OgaTensor** out);
```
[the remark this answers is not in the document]
RyanUnderhill: I don't think anyone is using this function externally. I added it in the recent python bindings only use the c api change. The new strong types caught the const being wrong as it's odd to get back a writable OgaTensor from a const OgaNamedTensors.

skottmckay (approved):

PR #1370, opened 2025-04-02
Initial support for VitisAI EP

AnanyaA-9:
- Initial support for VitisAI execution provider.
- Updated config session options to register custom ops library (```custom_ops_library```).

### DIFF
--- a/src/models/model.cpp
+++ b/src/models/model.cpp
@@ -537,6 +537,11 @@ void Model::CreateSessionOptionsFromConfig(const Config::SessionOptions& config_
       else if (provider_options.name == "WebGPU")
         p_device_ = GetDeviceInterface(DeviceType::WEBGPU);
 
+      else if (provider_options.name == "VitisAI") {
+        session_options.AddConfigEntry("session.inter_op.allow_spinning", "0");
+        session_options.AddConfigEntry("session.intra_op.allow_spinning", "0");
+      }
+
       std::vector<const char*> keys, values;
       for (auto& option : provider_options.options) {
         keys.emplace_back(option.first.c_str());


src/models/onnxruntime_inline.h
```diff
@@ -675,6 +676,26 @@ inline OrtSessionOptions& OrtSessionOptions::AppendExecutionProvider_CANN(const
...
+      values.push_back(entry.second.c_str());
+    }
+  }
+
+  Ort::ThrowOnError(
+      Ort::api->SessionOptionsAppendExecutionProvider_VitisAI(this, keys.data(), values.data(), num_entries));
+
+  return *this;
+}
+
```
RyanUnderhill, line 698: This API is meant to be zero overhead for the ORT API, so it can't take the unordered_map here, it must just take the keys/value pointers.

It looks like we're building an unordered_map in the model.cpp file from an array of values only to turn it back into an array here?

src/models/model.cpp
```diff
@@ -431,6 +431,19 @@ void Model::CreateSessionOptionsFromConfig(const Config::SessionOptions& config_
     session_options.AddConfigEntry("session.use_env_allocators", "1");
   }
 
+  if (config_session_options.intra_op_allow_spinning.has_value()) {
+    session_options.AddConfigEntry("session.intra_op.allow_spinning", std::to_string(config_session_options.intra_op_allow_spinning.value()).c_str());
+  }
+
+  if (config_session_options.inter_op_allow_spinning.has_value()) {
+    session_options.AddConfigEntry("session.inter_op.allow_spinning", std::to_string(config_session_options.inter_op_allow_spinning.value()).c_str());
+  }
```
RyanUnderhill, line 440: This is following the pattern from our own code, but I realized we should probably expose the 'AddConfigEntry' as an array of name/value pairs in our config vs naming them specifically. This is fine here, but we might change this soon.

src/models/onnxruntime_inline.h
VishalX: We can use `AppendExecutionProvider("VitisAI, provider_options_map)` instead and remove `AppendExecutionProvider_VitisAI`.

VishalX: Updated to use `AppendExecutionProvider`

VishalX:
@RyanUnderhill pls review again.

baijumeswani (approved):

src/models/model.cpp
```diff
@@ -518,6 +531,17 @@ void Model::CreateSessionOptionsFromConfig(const Config::SessionOptions& config_
...
+      // Add Provider options
+      std::unordered_map<std::string, std::string> provider_options_map;
+      for (const auto& option : provider_options.options) {
+        provider_options_map[option.first] = option.second;
+      }
+      try {
+        session_options.AppendExecutionProvider("VitisAI", provider_options_map);
+      } catch (std::exception& e) {
+        throw std::runtime_error(e.what());
+      }
```
RyanUnderhill, line 544: A new PR was just merged that should remove the need for this entirely :) Please sync again.

(https://github.com/microsoft/onnxruntime-genai/pull/1384)

src/models/onnxruntime_inline.h
AnanyaA-9: Removed ```AppendExecutionProvider_VitisAI```, instead using the extensible ```AppendExecutionProvider```

src/models/model.cpp
AnanyaA-9: Removed the new config session options ```intra_op_allow_spinning``` and ```inter_op_allow_spinning```.

src/models/model.cpp
AnanyaA-9: Rebased with latest main. Removed this section based on https://github.com/microsoft/onnxruntime-genai/commit/b47b4e73d4083a1d5c069566559c59b55da8f2f2.

baijumeswani (approved):

VishalX (approved):

PR #1371, opened 2025-04-02
CUDA Top K / Top P Fixes

aciddelgado:
Top P and Top K on CUDA were not outputting expected results due to incorrect math. Specifically, we were:

1. Softmaxing vocab_size logits _before_ getting the top k subset when it should've been after.
2. Performing top p first, then top k. We should instead perform top p on the top k subset.


src/search.cpp
```diff
@@ -217,11 +216,7 @@ void GreedySearch_Cpu::SampleTopKTopP(int k, float p, float temperature) {
     for (int i = 0; i < k; i++) {
       scores_top_k[i] = scores[indices[i]];
     }
-    if (k > 1) {
-      SoftmaxWithMax(scores_top_k, temperature, scores_top_k[0]);
-    } else {
-      scores_top_k[0] = 1.0f;
-    }
+    SoftmaxWithMax(scores_top_k, temperature, scores_top_k[0]);
```
RyanUnderhill, line 219: If k happens to be 1, does this still work fine?

RyanUnderhill (commented):
What tests did you try to check the changes?

aciddelgado:
> What tests did you try to check the changes?

Updated the tests to be more rigorous and have correct expected distributions

RyanUnderhill (approved):

PR #1378, opened 2025-04-04
Changes how the device OrtAllocators work, use a global OrtSession instead

RyanUnderhill:
Previously the code would store the OrtAllocator from the first OrtSession created. This apparently didn't work as expected. Even though the API to get an allocator from Ort is called CreateAllocator, it's really just returning a proxy to the OrtSession's allocator. When the OrtSession is destroyed, the OrtAllocator is invalid.


RyanUnderhill:
@baijumeswani See the #TODO in decoder_only_pipeline.cpp, is it okay to just delete the #if 0 code?

src/models/model.cpp
```diff
@@ -249,32 +249,159 @@ int32_t Tokenizer::TokenToTokenId(const char* token) const {
...
+      break;
+    default:
+      throw std::runtime_error("Unknown device type");
+  }
+
+  std::vector<Config::ProviderOptions> provider_options_list;
+  provider_options_list.emplace_back(Config::ProviderOptions{device_name, {}});
+
+  SetProviderSessionOptions(*session_options, provider_options_list, true, false);
+  allocator.session_ = OrtSession::Create(GetOrtEnv(), g_trivial_model, sizeof(g_trivial_model), session_options.get());
```
skottmckay, line 395: A comment explaining the setup to the new reader would be great so they don't have to infer 'why' we're doing all this. 

We create a session with the required EP. We put that session into the globals so it remains valid for the lifetime of GenAI. Due to that the allocator we get from that session is guaranteed to remain valid and can be used for any other models we load.

src/models/model.cpp
```diff
@@ -249,32 +249,159 @@ int32_t Tokenizer::TokenToTokenId(const char* token) const {
...
   static const char* device_type_names[] = {"CPU (Not used, see above)", "Cuda", "DML", "WebGPU_Buffer", "QnnHtpShared"};
   static_assert(std::size(device_type_names) == static_cast<size_t>(DeviceType::MAX));
 
   auto name = device_type_names[static_cast<int>(type)];
   auto memory_info = OrtMemoryInfo::Create(name, OrtAllocatorType::OrtDeviceAllocator, 0, OrtMemType::OrtMemTypeDefault);
-  device = Ort::Allocator::Create(session, *memory_info);
-  if (!device)
+  allocator.allocator_ = Ort::Allocator::Create(*allocator.session_, *memory_info);
+  if (!allocator.allocator_)
```
skottmckay, line 402: if this is going to throw should we reset allocator.session_?

skottmckay:
Can you add a unit test to validate the allocator in the globals is still valid after unloading a Model?

src/models/model.cpp
RyanUnderhill: It creates the OrtSession a few lines earlier. If someone resets the session_ the allocator is now invalid. Why would that happen?

src/models/decoder_only_pipeline.cpp
```diff
@@ -13,6 +13,10 @@ DecoderOnlyPipelineModel::DecoderOnlyPipelineModel(std::unique_ptr<Config> confi
   for (const auto& model : config_->model.decoder.pipeline) {
     sessions_.emplace_back(OrtSession::Create(ort_env, (config_->config_path / fs::path(model.filename)).c_str(),
                                               GetSessionOptions(model.model_id)));
+  }
+
+#if 0
+  // TODO: Baiju, is this still needed?
```
baijumeswani, line 19: We probably don't need this anymore since we can assume the device allocator has already been created.

src/models/model.cpp
```diff
@@ -249,32 +249,159 @@ int32_t Tokenizer::TokenToTokenId(const char* token) const {
...
+    } else
+      throw std::runtime_error("Unknown provider type: " + provider_options.name);
+  }
+  return p_device;
+}
+
+// Trivial ONNX model that just returns a single float constant. Used below to create an OrtSession that
+// lets us get a device Ort::Allocator for each device type. This is necessary because the Ort::Allocator
+// needs to persist and is valid for the lifetime of this OrtSession.
+static const uint8_t g_trivial_model[] = {
```
baijumeswani, line 350: One of the concerns with this trivial model approach is how it impacts QNN EP. The QNN execution provider will try to compile the model down to a context binary every time the model is loaded through the ort session. Given the triviality of the model, it should be not too bad for the binary generation. But my concern is that this may increase model load time as a result.

baijumeswani: Just tried it. This does impact the model load time a bit for QNN execution provider. Could we embed the context binary inside the ort-genai package for qnn?
Another thing I found is that the session options seem to enable verbose level logging, so my screen looks really scary when I run a model.

baijumeswani: ok, after doing some more testing. I think the model load time might also not be as bad as I thought it was. Seems fine. Sorry for the series of non-meaningful comments. :)

baijumeswani (approved):

baijumeswani (dismissed):

src/models/model.cpp
```diff
@@ -249,32 +249,149 @@ int32_t Tokenizer::TokenToTokenId(const char* token) const {
...
+  static const char* device_type_names[] = {"CPU (Not used, see above)", "cuda", "dml", "webgpu", "qnn"};
   static_assert(std::size(device_type_names) == static_cast<size_t>(DeviceType::MAX));
 
-  auto name = device_type_names[static_cast<int>(type)];
+  // Create an OrtSessionOptions and set the options to use the DeviceType we're using here
+  auto session_options = OrtSessionOptions::Create();
+  std::vector<Config::ProviderOptions> provider_options_list;
+  provider_options_list.emplace_back(Config::ProviderOptions{device_type_names[static_cast<int>(type)], {}});
+  SetProviderSessionOptions(*session_options, provider_options_list, true, false);
+  session_options->SetLogSeverityLevel(ORT_LOGGING_LEVEL_ERROR);  // Errors only here, as warnings are not useful to the user
```
skottmckay, line 380: would be good if this was configurable. for FL we might need to get GenAI to provide a custom logger to ORT so all output goes to the same file, with a configurable severity level so we can debug issues as getting native stdout/stderr from C# when asp.net is involved is non-trivial.

completely separate change though so FYI and not expecting this PR to change.

src/models/model.cpp
skottmckay: If the allocator was not successfully created and this code is going to throw, do we want allocator.session_ to remain? I would assume it's bad for some reason. It will get replaced on the next call to EnsureDeviceOrtInit but why wait?

src/models/model.cpp
skottmckay: Curious as to why there's any impact. I believe the trivial model has a Constant node and nothing else, which we convert to an initializer in the ORT Graph ctor leaving no actual nodes.

src/models/model.cpp
```diff
@@ -249,32 +249,149 @@ int32_t Tokenizer::TokenToTokenId(const char* token) const {
...
     return;
 
-  auto& device = GetOrtGlobals()->allocator_device_[static_cast<int>(type)];
-  if (device)
+  auto& allocator = GetOrtGlobals()->device_allocators_[static_cast<int>(type)];
+  if (allocator.allocator_)
     return;
 
-  static const char* device_type_names[] = {"CPU (Not used, see above)", "Cuda", "DML", "WebGPU_Buffer", "QnnHtpShared"};
+  // Names for the device types used by 'SetProviderSessionOptions'
```
skottmckay, line 371:
```suggestion
  // Allocator lifetime is tied to the execution provider lifetime, which is typically per Session.
  // We create a global dummy Session using a trivial model with the required EP in order to get the allocator so we can
  // re-use it for all models.
  // This ensures memory allocated on-device for model inputs/outputs is valid for the lifetime of GenAI.

  // Names for the device types used by 'SetProviderSessionOptions'
```

skottmckay (approved):

src/models/model.cpp
RyanUnderhill: Changed to reset everything on failure.

RyanUnderhill:
Waiting on this PR in onnxruntime to fix the crashing Linux & Windows FlashAttention kernel issue this exposed:

https://github.com/microsoft/onnxruntime/pull/24447

RyanUnderhill:
> Can you add a unit test to validate the allocator in the globals is still valid after unloading a Model?

We have multiple tests in a row that will each load a model then release it. So we should be good there. Plus.. this change already flushed out a memory issue inside onnxruntime :)

baijumeswani (approved):

baijumeswani (approved):

baijumeswani (approved):

PR #1389, opened 2025-04-09
[OVEP] feat: Adding OpenVINO EP in ORT-GenAI

ankitm3k:
This PR enables an initial integration & inference with `OpenVINOExecutionProvider` with ORT GenAI Extension ecosystem.

To enable the build using `OpenVINOExecutionProvider` , introducing a new build flag called `--use_openvino` which can be built using below -

Pre-requisites - 
- OpenVINO Toolkit Download - https://www.intel.com/content/www/us/en/developer/tools/openvino-toolkit/overview.html
- Latest ORT OpenVINOExecutionProvider (ORT OVEP) Build Binaries

python build.py --config Release --parallel --skip_tests **--use_openvino**  --ort_home path/to/ort_binaries_directory

There will be more changes hosted in the upcoming PRs ahead.


ankitm3k:
Kindly review & merge my first initial PR
@natke , @RyanUnderhill , @baijumeswani

src/models/model.cpp
```diff
@@ -262,7 +262,7 @@ void EnsureDeviceOrtInit(OrtSession& session, DeviceType type) {
   if (device)
     return;
 
-  static const char* device_type_names[] = {"CPU (Not used, see above)", "Cuda", "DML", "WebGPU_Buffer", "QnnHtpShared"};
+  static const char* device_type_names[] = {"CPU (Not used, see above)", "Cuda", "DML", "WebGPU_Buffer", "QnnHtpShared", "Cpu"};
```
RyanUnderhill, line 265: Is OpenVINO just using the CPU allocator? If so, it should just be using the CPU device type. This is only for when it has a separate Ort::Allocator.

src/models/model.cpp
```diff
@@ -533,6 +532,15 @@ void Model::CreateSessionOptionsFromConfig(const Config::SessionOptions& config_
...
       session_options.AppendExecutionProvider("WebGPU", opts);
+    } else if (provider_options.name == "openvino") {
+      std::unordered_map<std::string, std::string> ov_options;
+      for (auto& option : provider_options.options) {
+        ov_options.emplace(option.first, option.second);
+      }      
+      session_options.AppendExecutionProvider("OpenVINO", ov_options);
+      
+      if (is_primary_session_options)
+        p_device_ = GetDeviceInterface(DeviceType::OPENVINO);
```
RyanUnderhill, line 543: This PR should be merged soon and will simplify this: https://github.com/microsoft/onnxruntime-genai/pull/1384/files

Also, we need to have the provider_options.name match the ORT provider name, so it should be "OpenVINO" not "openvino". Going forward we'll have them match and the PR above has a backwards compatibility fix for the existing ones.

src/openvino/interface.cpp
```diff
@@ -0,0 +1,81 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
RyanUnderhill, line 1: Is there anything OpenVINO device specific? Or is it all just CPU?

src/python/py/models/builder.py
```diff
@@ -66,6 +66,7 @@ def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
             },
             "dml": {},
             "webgpu": {},
+            "openvino": {},
```
RyanUnderhill, line 69: Should be "OpenVINO" (and the other two should be changed, but that's not your fault)

src/smartptrs.h
```diff
@@ -153,6 +154,7 @@ struct ExternalRefCounted {
 namespace Location {
 struct CPU {};
 struct GPU {};
+struct NPU {};
```
RyanUnderhill, line 157: Is this used anywhere? We've been removing this usage as it's been replaced by the DeviceSpan struct for more specific device usage.

src/models/model.cpp
RyanUnderhill: The mentioned PR is now merged, please sync again and merge this with the simplified version.

build.py
```diff
@@ -125,6 +125,8 @@ class HelpFormatter(argparse.ArgumentDefaultsHelpFormatter, argparse.RawDescript
     )
 
     parser.add_argument("--use_rocm", action="store_true", help="Whether to use ROCm. Default is to not use rocm.")
+    
+    parser.add_argument("--use_openvino", action="store_true", help="Whether to use OpenVINO. Default is to not use OpenVINO.")
```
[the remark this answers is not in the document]
RyanMetcalfeInt8: I think @ankitm3k just followed the same pattern here as we observed for other EPs. It's true that there is no compile-time dependency on OpenVINO for these changes. So to double check my understanding -- your recommendation would be to remove all changes in build.py, as well as all CMake files, except for the change in [cmake/global_variables.cmake](https://github.com/microsoft/onnxruntime-genai/pull/1389/files#diff-4ad56a5f310ebf5d72bed8d95a20cb0068daab148db6b232a6fa7a66f0ed9d2d) where we add the required files to the build?

i.e.
```
  "${GENERATORS_ROOT}/openvino/*.h"
  "${GENERATORS_ROOT}/openvino/*.cpp"
```

src/openvino/interface.cpp
RyanMetcalfeInt8: Right now it is all just CPU. Eventually we may add something more OpenVINO-specific.

src/smartptrs.h
RyanMetcalfeInt8: This is not used anywhere. We can remove it.

src/python/py/models/builder.py
```diff
@@ -3458,7 +3459,7 @@ def get_args():
         "-e",
         "--execution_provider",
         required=True,
-        choices=["cpu", "cuda", "rocm", "dml", "webgpu"],
+        choices=["cpu", "cuda", "rocm", "dml", "webgpu","openvino"],
```
hanbitmyths, line 3462: A model builder generates an ONNX model containing contrib ops optimized for specific EPs. Do you have a plan to support contrib ops in OpenVINO to leverage model builder generated models?

build.py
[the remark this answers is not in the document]
baijumeswani: Yes, avoid adding a new build flag if the code does not need new compile time dependencies. As you've identified, you probably only need to add the new files under `${GENERATORS_ROOT}/openvino/` to the build srcs.

src/openvino/interface.cpp
RyanUnderhill: Is it from a separate allocator or is it the CPU allocator?

src/models/model.cpp
RyanMetcalfeInt8: Can you elaborate a bit here? Are you saying that we should have something other than `"Cpu"` for index 5 (when device=OpenVINO)?

src/openvino/interface.cpp
RyanMetcalfeInt8: Could you elaborate?

RyanUnderhill: Is the memory used by OpenVINO from the default CPU allocator? Or does it have a separate OpenVINO OrtAllocator?

RyanMetcalfeInt8: Ah, yes we are using the default CPU allocator.

RyanUnderhill: Ah, then there shouldn't be an OpenVINOMemory object yet, as this would only be if you had your own OrtAllocator. CPU memory is already managed by the CpuMemory object.

src/python/py/models/builder.py
gblong1: We do expect that OpenVINO supports a subset of the contrib ops.

gblong1: @RyanUnderhill does this line also require OpenVINO instead of openvino?

RyanUnderhill: I'm not sure, @kunal-vaishnavi can say for sure. It's only the strings passed to the OrtSessionOptions::AppendExecutionProvider directly that need to match case with Ort.

src/openvino/interface.cpp
RyanMetcalfeInt8: So then I guess you're saying that our Generators::OpenVINO::DeviceInterface should implement AllocateBase / WrapMemoryBase by creating CpuMemory objects?

src/python/py/models/builder.py
kunal-vaishnavi: > We do expect that OpenVINO supports a subset of the contrib ops.

Which contrib ops are currently supported? You can search for `com.microsoft` in the model builder to find all contrib ops. We want to ensure that we are generating ONNX models that can run on the OpenVINO EP. If most contrib ops are not supported, then we can wait for the support to arrive before enabling it as an option here.

> I'm not sure, @kunal-vaishnavi can say for sure. It's only the strings passed to the OrtSessionOptions::AppendExecutionProvider directly that need to match case with Ort.

We can have a mapping inside ORT GenAI for the EP abbreviations here to the actual EP names used in ORT.

```cpp
std::unordered_map<string, string> mapping;
mapping["cpu"] = "CPUExecutionProvider";
mapping["cuda"] = "CUDAExecutionProvider";
...
```

With this approach, using `openvino` instead of `OpenVINO` is fine.

src/smartptrs.h
RyanMetcalfeInt8: Removed in latest push

src/python/py/models/builder.py
RyanMetcalfeInt8: I think we will just remove our changes to builder.py for now, as we don't yet support *all* contrib ops.

ankitm3k: @kunal-vaishnavi  Can you kindly share the primary or upcoming list of Contrib Ops you would recommend which will be used in the ORT GenAI Ecosystem going forward viz. GQA, Flash Attention, SDPA, etc. Basically we want to target enabling these Ops to future proof the requirements coming ahead in GenAI space. Also kindly mention the quantization formats too.

src/openvino/interface.cpp
RyanMetcalfeInt8: okay, this should be resolved in latest push.

RyanMetcalfeInt8:
Okay, changes to address all comments (so far) have been pushed. Please take a look. Thanks!

src/models/model.cpp
```diff
@@ -255,14 +255,14 @@ int32_t Tokenizer::TokenToTokenId(const char* token) const {
 // arena already being destroyed.
 void EnsureDeviceOrtInit(OrtSession& session, DeviceType type) {
   // CPU Allocator is a special case, it's not in the owned 'allocator_device_' table below so we handle it separately
-  if (type == DeviceType::CPU)
+  if (type == DeviceType::CPU || type == DeviceType::OpenVINO)
```
RyanUnderhill, line 258: Can add a comment above to explain:

```
// OpenVINO delegates to the CPU device allocator
```

src/openvino/interface.cpp
```diff
@@ -0,0 +1,50 @@
...
+
+#include "../generators.h"
+#include "../search.h"
+#include "interface.h"
+#include "../cpu/interface.h"
+
+namespace Generators {
+namespace OpenVINO {
+
+const char* device_label = "OpenVINO";
```
RyanUnderhill, line 12: Is this used anywhere now? I think it was only for the DeviceSpan

src/python/python.cpp
```diff
@@ -484,6 +484,7 @@ PYBIND11_MODULE(onnxruntime_genai, m) {
   m.def("is_rocm_available", []() { return USE_ROCM != 0; });
   m.def("is_webgpu_available", []() { return true; });
   m.def("is_qnn_available", []() { return true; });
+  m.def("is_openvino_available", []() { return true; });
```
RyanUnderhill, line 487: Will it always be available? Or does Onnxruntime have to be built with support for it?

RyanMetcalfeInt8: Onnxruntime needs to be built with OpenVINO support. I think in the original changeset, it did something similar to rocm case,  `return USE_OPENVINO != 0;`, where `USE_OPENVINO` was introduced by some CMake checks. Perhaps we need a subset of these changes back?

src/models/model.cpp
RyanMetcalfeInt8: okay, added.

src/openvino/interface.cpp
RyanMetcalfeInt8: You're right, this is not used anymore. I removed it.

src/python/python.cpp
RyanMetcalfeInt8: alright, please see latest commit. Basically, CMake will check for existence of OpenVINO EP library, and set USE_OPENVINO compile flag accordingly. This is then used within python.cpp to give appropriate return value for `is_openvino_available`

RyanMetcalfeInt8:
Alright, I think we've addressed 2nd round of feedback. Let us know if there any any additional questions. Thanks!

ankitm3k:
> Alright, I think we've addressed 2nd round of feedback. Let us know if there any any additional questions. Thanks!

Hey @natke can you please kindly review & merge this PR as we have addressed all the comments.

src/python/python.cpp
RyanUnderhill: Does this mean that it's based on if the machine it was built on has OpenVINO, not the machine it is run on?

baijumeswani (commented):
Minor comment. Otherwise, PR looks good.

src/python/python.cpp
baijumeswani: I think it should always return true. We don't have any compile time dependency and unless we have a runtime way to check if the onnxruntime dll is an openvino one, we should always return true here. It is also how qnn and webgpu is setup.

RyanMetcalfeInt8: So I think we have 2 options:
1. Implement `is_openvino_available()` and always return True
2. Don't implement it at all

I'm not sure what the intent is of this function, but OpenVINO is available if the following conditions are met :
* OpenVINO onnxruntime EP library is installed into user's env.
* OpenVINO Toolkit environment is installed into user's env.

Perhaps eventually we can implement this function by checking if other packages are installed (onnxruntime-openvino, openvino, etc.) -- but for now, my opinion is to *not* implement it at all (option 2).

cmake/check_openvino.cmake
```diff
@@ -0,0 +1,13 @@
...
+# (global_variables.cmake only sets it for supported platforms)
+if(DEFINED ONNXRUNTIME_PROVIDERS_OPENVINO_LIB)
+  # Check for existence of OpenVINO EP library. This will determine
+  #  whether onnxruntime has been compiled with OpenVINO EP support.
+  if(EXISTS "${ORT_LIB_DIR}/${ONNXRUNTIME_PROVIDERS_OPENVINO_LIB}")
+    set(OPENVINO_EP_FOUND 1)
+  endif()
+endif()
+
+add_compile_definitions(USE_OPENVINO=${OPENVINO_EP_FOUND})
```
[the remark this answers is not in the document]
RyanMetcalfeInt8: right, in this case it was only to generate return value of python.cpp function.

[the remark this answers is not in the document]
baijumeswani: ok, let's avoid adding this compile definition and return true in the python.cpp. We can modify that in a later PR if we need to.

RyanMetcalfeInt8:
Alright, I've reverted the commit that added the changes to CMake which checked for existence of OpenVINO. `is_openvino_available` returns `True` now.

baijumeswani (approved):

src/python/py/models/builder.py
baijumeswani: Since the above comments are non-blocking, we can follow up in an offline thread.

build.py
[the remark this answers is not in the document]
ZhangWei125521: > 've removed al

could you share gudiance about how to infer with openvino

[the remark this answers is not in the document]
ZhangWei125521: > okay, as discussed, we've removed all changes to build.py, as well as all CMake files, except for adding the new openvino sources.

I dont now how to infer with openvino without no onnxruntime openvino provider dll .could you share guidance on this?

PR #1398, opened 2025-04-14
Add Chat Template API Changes

sayanshaw24:
Integrates ORT Extensions chat template implementation into ORT GenAI.

### DIFF
--- a/src/models/model.cpp
+++ b/src/models/model.cpp
@@ -211,7 +211,7 @@ std::string Tokenizer::ApplyChatTemplate(const char* template_str, const char* m
   ort_extensions::OrtxObjectPtr<OrtxTensor> tensor;
   CheckResult(OrtxTensorResultGetAt(templated_text.get(), 0, tensor.ToBeAssigned()));
 
-  const char* text_ptr = nullptr;
+  const char* text_ptr{};
   CheckResult(OrtxGetTensorData(tensor.get(), reinterpret_cast<const void**>(&text_ptr), nullptr, nullptr));
 
   return text_ptr;
--- a/src/models/model.cpp
+++ b/src/models/model.cpp
@@ -206,7 +206,7 @@ std::string Tokenizer::Decode(std::span<const int32_t> tokens) const {
 
 std::string Tokenizer::ApplyChatTemplate(const char* template_str, const char* messages, bool add_generation_prompt) const {
   const char* output;
-  CheckResult(OrtxApplyChatTemplate(tokenizer_, template_str, messages, output, add_generation_prompt, false)); // bool tokenize = false
+  CheckResult(OrtxApplyChatTemplate(tokenizer_, template_str, messages, output, add_generation_prompt, false));  // bool tokenize = false
   return output;
 }
 
--- a/src/models/model.cpp
+++ b/src/models/model.cpp
@@ -206,7 +206,7 @@ std::string Tokenizer::Decode(std::span<const int32_t> tokens) const {
 
 std::string Tokenizer::ApplyChatTemplate(const char* template_str, const char* messages, bool add_generation_prompt) const {
   ort_extensions::OrtxObjectPtr<OrtxTensorResult> templated_text;
-  CheckResult(OrtxApplyChatTemplate(tokenizer_, template_str, messages, templated_text.ToBeAssigned(), add_generation_prompt, false));  // Note: tokenize is set to false
+  CheckResult(OrtxApplyChatTemplate(tokenizer_, template_str, messages, templated_text.ToBeAssigned(), add_generation_prompt, false /*tokenize*/));
 
   ort_extensions::OrtxObjectPtr<OrtxTensor> tensor;
   CheckResult(OrtxTensorResultGetAt(templated_text.get(), 0, tensor.ToBeAssigned()));
--- a/src/models/model.cpp
+++ b/src/models/model.cpp
@@ -204,6 +204,12 @@ std::string Tokenizer::Decode(std::span<const int32_t> tokens) const {
   return string;
 }
 
+std::string Tokenizer::ApplyChatTemplate(const char* template_str, const char* messages, bool add_generation_prompt) const {
+  const char* output;
+  CheckResult(OrtxApplyChatTemplate(tokenizer_, template_str, messages, output, add_generation_prompt, false)); // bool tokenize = false
+  return output;
+}
+
 std::vector<int32_t> Tokenizer::EncodeBatch(std::span<const std::string> strings) const {
   std::vector<std::vector<int32_t>> sequences;
   std::vector<std::span<const int32_t>> span_sequences;
--- a/src/models/model.cpp
+++ b/src/models/model.cpp
@@ -205,9 +205,16 @@ std::string Tokenizer::Decode(std::span<const int32_t> tokens) const {
 }
 
 std::string Tokenizer::ApplyChatTemplate(const char* template_str, const char* messages, bool add_generation_prompt) const {
-  const char* output;
-  CheckResult(OrtxApplyChatTemplate(tokenizer_, template_str, messages, output, add_generation_prompt, false));  // bool tokenize = false
-  return output;
+  ort_extensions::OrtxObjectPtr<OrtxTensorResult> templated_text;
+  CheckResult(OrtxApplyChatTemplate(tokenizer_, template_str, messages, templated_text.ToBeAssigned(), add_generation_prompt, false));  // Note: tokenize is set to false
+
+  OrtxObjectPtr<OrtxTensor> tensor;
+  CheckResult(OrtxTensorResultGetAt(templated_text.get(), 0, tensor.ToBeAssigned()));
+  
+  const char* text_ptr = nullptr;
+  CheckResult(OrtxGetTensorData(tensor.get(), reinterpret_cast<const void**>(&text_ptr), nullptr, nullptr));
+  
+  return text_ptr;
 }
 
 std::vector<int32_t> Tokenizer::EncodeBatch(std::span<const std::string> strings) const {
--- a/src/models/model.cpp
+++ b/src/models/model.cpp
@@ -208,7 +208,7 @@ std::string Tokenizer::ApplyChatTemplate(const char* template_str, const char* m
   ort_extensions::OrtxObjectPtr<OrtxTensorResult> templated_text;
   CheckResult(OrtxApplyChatTemplate(tokenizer_, template_str, messages, templated_text.ToBeAssigned(), add_generation_prompt, false));  // Note: tokenize is set to false
 
-  OrtxObjectPtr<OrtxTensor> tensor;
+  ort_extensions::OrtxObjectPtr<OrtxTensor> tensor;
   CheckResult(OrtxTensorResultGetAt(templated_text.get(), 0, tensor.ToBeAssigned()));
   
   const char* text_ptr = nullptr;
--- a/src/models/model.cpp
+++ b/src/models/model.cpp
@@ -210,10 +210,10 @@ std::string Tokenizer::ApplyChatTemplate(const char* template_str, const char* m
 
   ort_extensions::OrtxObjectPtr<OrtxTensor> tensor;
   CheckResult(OrtxTensorResultGetAt(templated_text.get(), 0, tensor.ToBeAssigned()));
-  
+
   const char* text_ptr = nullptr;
   CheckResult(OrtxGetTensorData(tensor.get(), reinterpret_cast<const void**>(&text_ptr), nullptr, nullptr));
-  
+
   return text_ptr;
 }
 
--- a/src/ort_genai.h
+++ b/src/ort_genai.h
@@ -286,7 +286,7 @@ struct OgaTokenizer : OgaAbstract {
   }
 
   OgaString ApplyChatTemplate(const char* template_str, const char* messages, bool add_generation_prompt) const {
-    const char* p;
+    const char* p{};
     OgaCheckResult(OgaTokenizerApplyChatTemplate(this, template_str, messages, add_generation_prompt, &p));
     return p;
   }
--- a/src/ort_genai.h
+++ b/src/ort_genai.h
@@ -285,6 +285,12 @@ struct OgaTokenizer : OgaAbstract {
     return p;
   }
 
+  OgaString ApplyChatTemplate(const char* template_str, const char* messages, bool add_generation_prompt) const {
+    const char* p;
+    OgaCheckResult(OgaTokenizerApplyChatTemplate(this, template_str, messages, add_generation_prompt, &p));
+    return p;
+  }
+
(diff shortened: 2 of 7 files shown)


src/ort_genai_c.h
```diff
@@ -492,6 +492,10 @@ OGA_EXPORT OgaResult* OGA_API_CALL OgaProcessorProcessImagesAndAudios(const OgaM
 OGA_EXPORT OgaResult* OGA_API_CALL OgaTokenizerDecode(const OgaTokenizer*, const int32_t* tokens, size_t token_count, const char** out_string);
 OGA_EXPORT OgaResult* OGA_API_CALL OgaProcessorDecode(const OgaMultiModalProcessor*, const int32_t* tokens, size_t token_count, const char** out_string);
 
+/** Applies chat template to messages.
+ */
```
RyanUnderhill, line 496: If you can, describe the individual parameters like some of the more documented other functions.

src/models/model.cpp
```diff
@@ -204,6 +204,12 @@ std::string Tokenizer::Decode(std::span<const int32_t> tokens) const {
   return string;
 }
 
+std::string Tokenizer::ApplyChatTemplate(const char* template_str, const char* messages, bool add_generation_prompt) const {
+  const char* output;
+  CheckResult(OrtxApplyChatTemplate(tokenizer_, template_str, messages, output, add_generation_prompt, false));  // bool tokenize = false
```
RyanUnderhill, line 209: What is the commented out bool tokenize for?

src/models/model.cpp
```diff
@@ -204,6 +204,12 @@ std::string Tokenizer::Decode(std::span<const int32_t> tokens) const {
   return string;
 }
 
+std::string Tokenizer::ApplyChatTemplate(const char* template_str, const char* messages, bool add_generation_prompt) const {
+  const char* output;
```
RyanUnderhill, line 208:
```suggestion
  const char* output{};
```
Zero init for safety.

RyanUnderhill:
Is the C++ wrapper function still to come?

src/models/model.cpp
sayanshaw24: oh it's not a commented out parameter, it was to clarify what the last parameter was for (false) , but I can remove it if it's unclear, or change the comment to state that more explicitly.

src/ort_genai_c.h
sayanshaw24: done!

src/models/model.cpp
```diff
@@ -204,6 +204,19 @@ std::string Tokenizer::Decode(std::span<const int32_t> tokens) const {
...
 }
 
+std::string Tokenizer::ApplyChatTemplate(const char* template_str, const char* messages, bool add_generation_prompt) const {
+  ort_extensions::OrtxObjectPtr<OrtxTensorResult> templated_text;
+  CheckResult(OrtxApplyChatTemplate(tokenizer_, template_str, messages, templated_text.ToBeAssigned(), add_generation_prompt, false));  // Note: tokenize is set to false
+
+  ort_extensions::OrtxObjectPtr<OrtxTensor> tensor;
+  CheckResult(OrtxTensorResultGetAt(templated_text.get(), 0, tensor.ToBeAssigned()));
+
+  const char* text_ptr = nullptr;
```
RyanUnderhill, line 214:
```suggestion
  const char* text_ptr{};
```
Simpler

test/c_api_tests.cpp
```diff
@@ -94,6 +94,43 @@ TEST(CAPITests, TokenizerCAPI) {
 #endif
 }
 
+TEST(CAPITests, ChatTemplate) {
+#if TEST_PHI2
+  auto config = OgaConfig::Create(PHI2_PATH);  // just to create tokenizer (phi-2 does not have a chat template)
+  auto model = OgaModel::Create(*config);
```
RyanUnderhill, line 100:
```suggestion
  auto tokenizer = OgaTokenizer::Create(*OgaModel::Create(PHI2_PATH));
```
See if this works, just to make it even simpler.

test/c_api_tests.cpp
```diff
@@ -94,6 +94,43 @@ TEST(CAPITests, TokenizerCAPI) {
...
+  auto tokenizer = OgaTokenizer::Create(*model);
+
+  // Testing phi-4 chat template
+  {
+    std::string messages_json = R"(
```
RyanUnderhill, line 105: Why is this a std::string if you're just doing c_str() on it? chat_template is a regular C string right after it. Same with expected_output

sayanshaw24: good point, changed it to a regular C string

src/models/model.cpp
```diff
@@ -204,6 +204,19 @@ std::string Tokenizer::Decode(std::span<const int32_t> tokens) const {
   return string;
 }
 
+std::string Tokenizer::ApplyChatTemplate(const char* template_str, const char* messages, bool add_generation_prompt) const {
+  ort_extensions::OrtxObjectPtr<OrtxTensorResult> templated_text;
+  CheckResult(OrtxApplyChatTemplate(tokenizer_, template_str, messages, templated_text.ToBeAssigned(), add_generation_prompt, false));  // Note: tokenize is set to false
```
RyanUnderhill, line 209:
```suggestion
  CheckResult(OrtxApplyChatTemplate(tokenizer_, template_str, messages, templated_text.ToBeAssigned(), add_generation_prompt, false /*tokenize*/));
```

Recommended way to document bool values in parameters

src/ort_genai.h
```diff
@@ -285,6 +285,12 @@ struct OgaTokenizer : OgaAbstract {
     return p;
   }
 
+  OgaString ApplyChatTemplate(const char* template_str, const char* messages, bool add_generation_prompt) const {
+    const char* p;
```
RyanUnderhill, line 289:
```suggestion
    const char* p{};
```
Just in case

test/c_api_tests.cpp
```diff
@@ -94,6 +94,42 @@ TEST(CAPITests, TokenizerCAPI) {
 #endif
 }
 
+TEST(CAPITests, ChatTemplate) {
+#if TEST_PHI2
+  // just to create tokenizer (phi-2 does not have a chat template)
```
RyanUnderhill, line 99:
```suggestion
  // We load the phi-2 model just to get a tokenizer (phi-2 does not have a chat template)
```

test/c_api_tests.cpp
```diff
@@ -94,6 +94,42 @@ TEST(CAPITests, TokenizerCAPI) {
...
+  auto tokenizer = OgaTokenizer::Create(*OgaModel::Create(PHI2_PATH));
+
+  // Testing phi-4 chat template
+  {
```
RyanUnderhill, line 103: Unnecessary scope?

test/c_api_tests.cpp
```diff
@@ -94,6 +94,42 @@ TEST(CAPITests, TokenizerCAPI) {
...
+        },
+        {
+          "role": "assistant",
+          "content": "You can add numbers by using the '+' operator."
+        }
+      ])";
+    const char* chat_template = R"({% for message in messages %}{% if message['role'] == 'system' and 'tools' in message and message['tools'] is not none %}{{ '<|' + message['role'] + '|>' + message['content'] + '<|tool|>' + message['tools'] + '<|/tool|>' + '<|end|>' }}{% else %}{{ '<|' + message['role'] + '|>' + message['content'] + '<|end|>' }}{% endif %}{% endfor %}{% if add_generation_prompt %}{{ '<|assistant|>' }}{% else %}{{ eos_token }}{% endif %})";
+    
+    // From HuggingFace Python output for 'microsoft/Phi-4-multimodal-instruct'
+    std::string expected_output = "<|system|>You are a helpful assistant.<|tool|>Calculator<|/tool|><|end|><|user|>"
```
RyanUnderhill, line 123: Why is this one a std::string?

sayanshaw24: updated!

src/ort_genai_c.h
```diff
@@ -492,6 +492,22 @@ OGA_EXPORT OgaResult* OGA_API_CALL OgaProcessorProcessImagesAndAudios(const OgaM
...
+ * This function processes the specified template with the provided input using the
+ * tokenizer, and outputs the resulting string. Optionally, it can include a
+ * generation prompt in the output.
+ *
+ * @param tokenizer OgaTokenizer used for template processing.
+ * @param template_str Null-terminated string representing the chat template
+ * @param messages Null-terminated string containing the input messages to be processed.
+ * @param add_generation_prompt Indicates whether to add a generation prompt to the output.
+ * @param out_string Pointer to where the output will be stored.
+ * @return OgaResult* Returns a pointer to the result.
```
RyanUnderhill, line 507:
```suggestion
 * \param[in] tokenizer OgaTokenizer used for template processing.
 * \param[in] template_str Null-terminated string representing the chat template
 * \param[in] messages Null-terminated string containing the input messages to be processed.
 * \param[in] add_generation_prompt Indicates whether to add a generation prompt to the output.
 * \param[out] out_string Pointer to where the output will be stored.
 * \return OgaResult* containing the error message if the function fails
```

src/ort_genai_c.h
```diff
@@ -492,6 +492,22 @@ OGA_EXPORT OgaResult* OGA_API_CALL OgaProcessorProcessImagesAndAudios(const OgaM
...
 
+/**
+ * @brief Applies a chat template to the given input.
+ *
+ * This function processes the specified template with the provided input using the
+ * tokenizer, and outputs the resulting string. Optionally, it can include a
+ * generation prompt in the output.
+ *
+ * \param[in] tokenizer OgaTokenizer used for template processing.
+ * \param[in] template_str Null-terminated string representing the chat template
```
baijumeswani, line 503: Should the user be responsible to pass in the chat template? Can it read from the tokenizer config or from chat template config? Is this an optional input that overrides what is found in the tokenizer config?

baijumeswani: In most cases, the user does not know the chat template and would rather prefer we auto set it from the tokenizer config. Having said that, there could be some scenarios where the user does want to override the template.

But it should not be a required input.

baijumeswani: ok, can we mention it here that it is optional? It wasn't clear to me that this was an optional input.
Also, we probably should avoid adding c api utilities just for testing if possible.

Would users want to override this template str?

src/ort_genai_c.h
```diff
@@ -492,6 +492,22 @@ OGA_EXPORT OgaResult* OGA_API_CALL OgaProcessorProcessImagesAndAudios(const OgaM
...
+ * \param[in] messages Null-terminated string containing the input messages to be processed.
+ * \param[in] add_generation_prompt Indicates whether to add a generation prompt to the output.
```
baijumeswani, line 505: Under what circumstances would a user want to use a generation prompt?

src/ort_genai_c.h
```diff
@@ -492,6 +492,22 @@ OGA_EXPORT OgaResult* OGA_API_CALL OgaProcessorProcessImagesAndAudios(const OgaM
...
+ * \param[out] out_string Pointer to where the output will be stored.
+ * \return OgaResult* containing the error message if the function fails
+ */
+OGA_EXPORT OgaResult* OGA_API_CALL OgaTokenizerApplyChatTemplate(const OgaTokenizer*, const char* template_str, const char* messages, bool add_generation_prompt, const char** out_string);
```
baijumeswani, line 509: Seems like this a two-step process:

- ApplyChatTemplate -> string
- Tokenize -> int array

Mostly, I see users would want to get the int tokens. Could we add another API that returns the int tokens when ApplyChatTemplate?

src/ort_genai_c.h
sayanshaw24: it's mentioned in the brief:
> Optionally, it can include a generation prompt in the output.

https://github.com/microsoft/onnxruntime-genai/pull/1398/files#diff-4d2efedaefd309b86072862c893606033ca895483d661fb936541bc1774ca3f9R499

src/ort_genai_c.h
sayanshaw24: an end-user wouldn't care about it, it's for the API user; you would use it whenever you're generating a model response — it tells the model, "Now it's your turn to speak." (when set to true, it will append the correct assistant prompt tag or marker at the end of the input)

src/ort_genai_c.h
sayanshaw24: This is a good point to discuss; in Extensions we combine the two as we can have variable types in a tensor - but since we want to have the data type (string/int vector) being returned in GenAI, Ryan and I decided on the latter, but it's up to the GenAI folks how we want to handle this, we can bring it up in the next meeting with the team.

src/ort_genai_c.h
baijumeswani: Sorry, I meant can we document the template str as optional? The documentation doesn't do that at the moment.

src/ort_genai_c.h
baijumeswani: OK that makes sense. Could we add a note to that effect in the documentation?

src/ort_genai_c.h
sayanshaw24: the reason i want to avoid that is because it cannot just be left in this case; say with the LITE api Randy made for instance we could actually have optional params, but in this case, if it's not being used, the user of this API needs to pass in a nullptr like this: https://github.com/microsoft/onnxruntime-extensions/blob/d944f38acec75a4ff73dfe40ad4c082e7bb33c53/test/pp_api_test/test_tokenizer_chat.cc#L39

baijumeswani:
```suggestion
 * \param[in] template_str Null-terminated string representing the chat template. Use nullptr to fall back to the default chat template from the tokenizer config.
```

sayanshaw24: gotcha, sounds good, thanks - just committed!

baijumeswani (dismissed):

src/ort_genai_c.h
```diff
@@ -492,6 +492,22 @@ OGA_EXPORT OgaResult* OGA_API_CALL OgaProcessorProcessImagesAndAudios(const OgaM
 OGA_EXPORT OgaResult* OGA_API_CALL OgaTokenizerDecode(const OgaTokenizer*, const int32_t* tokens, size_t token_count, const char** out_string);
 OGA_EXPORT OgaResult* OGA_API_CALL OgaProcessorDecode(const OgaMultiModalProcessor*, const int32_t* tokens, size_t token_count, const char** out_string);
 
+/**
+ * @brief Applies a chat template to the given input. out_string must be freed with OgaDestroyString
```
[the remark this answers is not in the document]
sayanshaw24: oh yeah i just meant through multiple runs, hence messages, not message; just semantics

RyanUnderhill (approved):

PR #1399, opened 2025-04-14
OpenVINO: Model Managed KVCache

RyanMetcalfeInt8:
This PR introduces a *Model Managed* KVCache type, and is used for OpenVINO execution provider when the *enable_causallm* provider option is set. 

*enable_causallm* is expected to be set within the *genai_config.json* for models that support this flow.

The *ModelManagedKeyValueCache* class is mostly a no-op. This is because the KVCache tensors are internally managed by the underlying OpenVINO infer_request, which is tied to the *session*.

For *Rewind*, we make use of ORT's *SetEpDynamicOptions* to trigger the trimming of internal KVCache state tensors.


RyanMetcalfeInt8:
FYI -- @natke , @RyanUnderhill , @baijumeswani

RyanMetcalfeInt8:
The CI failures here seem to show:
```
Error: D:\a\_work\onnxruntime-genai\onnxruntime-genai\src\models\onnxruntime_inline.h(855,31): error C2039: 'SetEpDynamicOptions': is not a member of 'OrtApi'
```

I'm a little bit confused on this one, as it's not a problem in our builds -- doesn't this API exist in ORT since 1.20?
[onnxruntime_c_api.h 1.21](https://github.com/microsoft/onnxruntime/blob/v1.21.0/include/onnxruntime/core/session/onnxruntime_c_api.h#L4787)
[onnxruntime_c_api.h main](https://github.com/microsoft/onnxruntime/blob/main/include/onnxruntime/core/session/onnxruntime_c_api.h#L4835)

baijumeswani:
> The CI failures here seem to show:
> 
> ```
> Error: D:\a\_work\onnxruntime-genai\onnxruntime-genai\src\models\onnxruntime_inline.h(855,31): error C2039: 'SetEpDynamicOptions': is not a member of 'OrtApi'
> ```
> 
> I'm a little bit confused on this one, as it's not a problem in our builds -- doesn't this API exist in ORT since 1.20? [onnxruntime_c_api.h 1.21](https://github.com/microsoft/onnxruntime/blob/v1.21.0/include/onnxruntime/core/session/onnxruntime_c_api.h#L4787) [onnxruntime_c_api.h main](https://github.com/microsoft/onnxruntime/blob/main/include/onnxruntime/core/session/onnxruntime_c_api.h#L4835)

The version the CI pipeline uses to download the onnxruntime version is hard coded here: https://github.com/microsoft/onnxruntime-genai/blob/00a5fb62a36f9635a476039fb60b40b2901c5e71/cmake/ortlib.cmake#L24

I am not sure if this version has the necessary API. Could you update the version to one that should have such support?

RyanMetcalfeInt8:
> > The CI failures here seem to show:
> > ```
> > Error: D:\a\_work\onnxruntime-genai\onnxruntime-genai\src\models\onnxruntime_inline.h(855,31): error C2039: 'SetEpDynamicOptions': is not a member of 'OrtApi'
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I'm a little bit confused on this one, as it's not a problem in our builds -- doesn't this API exist in ORT since 1.20? [onnxruntime_c_api.h 1.21](https://github.com/microsoft/onnxruntime/blob/v1.21.0/include/onnxruntime/core/session/onnxruntime_c_api.h#L4787) [onnxruntime_c_api.h main](https://github.com/microsoft/onnxruntime/blob/main/include/onnxruntime/core/session/onnxruntime_c_api.h#L4835)
> 
> The version the CI pipeline uses to download the onnxruntime is hard coded here:
> 
> https://github.com/microsoft/onnxruntime-genai/blob/00a5fb62a36f9635a476039fb60b40b2901c5e71/cmake/ortlib.cmake#L24
> 
> I am not sure if this version has the necessary API. Could you update the version to one that should have such support?

Sure, but where can I get the list of available package names (like `1.20.0-dev-20241007-1101-407c1ab2e2`)? Also, it seems like slight different variants are used for CUDA, DML, ROCM, etc. I should change all of these, or is there some functional reason why they differ here?

RyanMetcalfeInt8:
> > > The CI failures here seem to show:
> > > ```
> > > Error: D:\a\_work\onnxruntime-genai\onnxruntime-genai\src\models\onnxruntime_inline.h(855,31): error C2039: 'SetEpDynamicOptions': is not a member of 'OrtApi'
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > I'm a little bit confused on this one, as it's not a problem in our builds -- doesn't this API exist in ORT since 1.20? [onnxruntime_c_api.h 1.21](https://github.com/microsoft/onnxruntime/blob/v1.21.0/include/onnxruntime/core/session/onnxruntime_c_api.h#L4787) [onnxruntime_c_api.h main](https://github.com/microsoft/onnxruntime/blob/main/include/onnxruntime/core/session/onnxruntime_c_api.h#L4835)
> > 
> > 
> > The version the CI pipeline uses to download the onnxruntime is hard coded here:
> > https://github.com/microsoft/onnxruntime-genai/blob/00a5fb62a36f9635a476039fb60b40b2901c5e71/cmake/ortlib.cmake#L24
> > 
> > I am not sure if this version has the necessary API. Could you update the version to one that should have such support?
> 
> Sure, but where can I get the list of available package names (like `1.20.0-dev-20241007-1101-407c1ab2e2`)? Also, it seems like different variants are used for CUDA, DML, ROCM, etc. I should change all of these, or is there some functional reason why they differ here?

I'll try with the latest formal release, "1.21.0"

baijumeswani:
We get them from the ort-nightly public feed. Hopefully, you can access this: https://aiinfra.visualstudio.com/PublicPackages/_artifacts/feed/ORT-Nightly/NuGet/Microsoft.ML.OnnxRuntime/versions/1.22.0-dev-20250409-1104-89f8206ba4 since it is a public feed.

These ort dlls are compile time dependencies. Meaning that the ort API version used during compilation is the minimum required at runtime. Ideally, we want to support as many ort-runtime versions as possible at runtime. Needing new API should be the only reason to use newer compile time ort versions.

RyanMetcalfeInt8:
> We get them from the ort-nightly public feed. Hopefully, you can access this: https://aiinfra.visualstudio.com/PublicPackages/_artifacts/feed/ORT-Nightly/NuGet/Microsoft.ML.OnnxRuntime/versions/1.22.0-dev-20250409-1104-89f8206ba4 since it is a public feed.
> 
> These ort dlls are compile time dependencies. Meaning that the ort API version used during compilation is the minimum required at runtime. Ideally, we want to support as many ort-runtime versions as possible at runtime. Needing new API should be the only reason to use newer compile time ort versions.

Thanks, I do have access to that. I just pushed a new commit which sets ORT_VERSION to 1.20.0. It looks like the CI issue with 1.21.0, is that onnxruntime packages installed for python are 1.20.X -- Hopefully this will resolve these failures.

RyanMetcalfeInt8:
> > We get them from the ort-nightly public feed. Hopefully, you can access this: https://aiinfra.visualstudio.com/PublicPackages/_artifacts/feed/ORT-Nightly/NuGet/Microsoft.ML.OnnxRuntime/versions/1.22.0-dev-20250409-1104-89f8206ba4 since it is a public feed.
> > These ort dlls are compile time dependencies. Meaning that the ort API version used during compilation is the minimum required at runtime. Ideally, we want to support as many ort-runtime versions as possible at runtime. Needing new API should be the only reason to use newer compile time ort versions.
> 
> Thanks, I do have access to that. I just pushed a new commit which sets ORT_VERSION to 1.20.0. It looks like the CI issue with 1.21.0, is that onnxruntime packages installed for python are 1.20.X -- Hopefully this will resolve these failures.
> 
> As the `SetEPDynamicOptions` API was introduced in 1.20.0 release, this would become the new minimum ORT version needed to compile.

src/models/kv_cache.cpp
```diff
@@ -376,6 +376,33 @@ std::string ComposeKeyValueName(const std::string& template_string, int index) {
...
+    : state_{state} {
+  // This fixes with _qa sample where a new instance of ModelManagedKeyValueCache is
+  // created for each new prompt from user. In this case, we need to trigger a KVCache
+  // reset on the session before the first Session::Run.
+  RewindTo(0);
+}
+
+void ModelManagedKeyValueCache::Add() {
+  // NOOP
+}
```
baijumeswani, line 389: nit: unnecessary comment.
```suggestion
void ModelManagedKeyValueCache::Add() {}
```

src/models/kv_cache.cpp
```diff
@@ -376,6 +376,33 @@ std::string ComposeKeyValueName(const std::string& template_string, int index) {
...
+
+void ModelManagedKeyValueCache::AddEncoder() {
+  // NOOP
+}
```
baijumeswani, line 393: nit: unnecessary comment.

```suggestion
void ModelManagedKeyValueCache::AddEncoder() {}
```

src/models/kv_cache.cpp
```diff
@@ -376,6 +376,33 @@ std::string ComposeKeyValueName(const std::string& template_string, int index) {
...
+
+void ModelManagedKeyValueCache::Update(DeviceSpan<int32_t> beam_indices, int total_length) {
+  // NO-OP for now.
```
baijumeswani, line 396:
```suggestion
```

src/models/kv_cache.cpp
```diff
@@ -376,6 +376,33 @@ std::string ComposeKeyValueName(const std::string& template_string, int index) {
   return std::string(key_value_name);
 }
 
+ModelManagedKeyValueCache::ModelManagedKeyValueCache(State& state)
+    : state_{state} {
+  // This fixes with _qa sample where a new instance of ModelManagedKeyValueCache is
+  // created for each new prompt from user. In this case, we need to trigger a KVCache
+  // reset on the session before the first Session::Run.
```
baijumeswani, line 383: A comment explaining the coupling of the kv cache state with the ort session state and its implications would be helpful.

```suggestion
  // A new instance of ModelManagedKeyValueCache is created for each Generator.
  // In this case, we need to trigger a KVCache reset on the session before the first Session::Run.
  // This implies that the key-value cache state is coupled with the ONNX Runtime Session and
  // that only 1 Generator can be active for the Model at any given time.
```

In theory, the Model:Generator relationship is 1:many since the ort session is mostly stateless. But for openvino, the ort-session is stateful. Which means the Model:Generator relationship is 1:1 at any given time.

src/models/kv_cache.h
```diff
@@ -123,6 +123,22 @@ struct CrossCache {
...
+
+  virtual void Add() override;
+  virtual void AddEncoder() override;
+  virtual void Update(DeviceSpan<int32_t> beam_indices, int total_length) override;
+  virtual void RewindTo(size_t index) override;
+
+ private:
+  State& state_;
+  const Model& model_{state_.model_};
+  Ort::Allocator& Allocator() { return model_.p_device_kvcache_->GetAllocator(); }
```
baijumeswani, line 139: Is this function needed?

src/models/model.h
```diff
@@ -46,6 +46,8 @@ struct State {
   std::vector<std::string> adapter_names_;
   std::vector<OrtValue*> inputs_, outputs_;
 
+  std::vector<std::pair<std::string, std::string>> _ep_dynamic_options_next_run;
```
baijumeswani, line 49: nit: consistent with other members.

```suggestion
  std::vector<std::pair<std::string, std::string>> ep_dynamic_options_next_run_;
```

src/models/logits.cpp
```diff
@@ -20,6 +20,27 @@ Logits::Logits(State& state)
...
+    const auto& provider_options = state.model_.config_->model.decoder.session_options.provider_options;
+    for (auto& po : provider_options) {
+      if (po.name == "OpenVINO") {
+        const auto& openvino_options = po.options;
+        for (auto& option : openvino_options) {
+          // When 'enable_causallm' option is set, models are patched in a way so that only return the
+          // sliced logits needed for sampling. For example, given 43 prompt tokens, instead of returning
+          // logits of the shape:  [1,43,<vocab_size>]
+          // they will have shape: [1, 1,<vocab_size>].
+          if (option.first == "enable_causallm" && option.second == "True") {
```
baijumeswani, line 34: This check seems to be in multiple places, could we make a utility function for it?

baijumeswani (commented):
Some minor comments but otherwise looks good.

src/models/kv_cache.cpp
RyanMetcalfeInt8: Yes, correct. We are aware of this, and have some ideas about how to address it. Within the session, we can have multiple ov::InferRequest objects, each with their own states -- we just need a way to tie them to a particular instance of a Generator within ORT GenAI. Anyway, absolutely agree that supporting 1:many is needed later on.

src/models/kv_cache.h
RyanMetcalfeInt8: Ah, you're right, it's not used. I'll remove it, thanks.

src/models/logits.cpp
RyanMetcalfeInt8: Take a look at latest commit, https://github.com/microsoft/onnxruntime-genai/pull/1399/commits/14ac497988815867d0003dd723d79fa1e6c70844

src/models/model.h
RyanMetcalfeInt8: Done.

src/models/kv_cache.h
RyanMetcalfeInt8: Done.

RyanMetcalfeInt8:
Hi @baijumeswani -- I think I've addressed all comments so far -- can you take a look? Thanks!

baijumeswani (approved):

RyanUnderhill (approved):

PR #1404, opened 2025-04-16
Initial support for NV EP

BLSharda:
This is initial change to support 'nv' execution provider in GenAI. We have ongoing changes for model builder and inference code. 

Notes:
- This had dependency on changes in ORT - NV EP APIs (coming soon)
- Currently NV EP code is tied to CUDA EP for code compilation 'USE_CUDA' and just requires EP name 'nv'.
- Onnx exporter for 'nv' EP tries to implement primitive op instead of contrib ops.


RyanUnderhill:
Is it possible for NV to extend OrtSessionAppendExecutionProvider with a new name? This fits with the other new providers being added and is general purpose.

src/models/model.cpp
```diff
@@ -480,6 +480,23 @@ void Model::CreateSessionOptionsFromConfig(const Config::SessionOptions& config_
...
+      auto ort_provider_options = OrtNvProviderOptions::Create();
+      // TODO: missing handling in ORT-> NV-EP, need to verify and enable.
+      //ort_provider_options->Update(keys.data(), values.data(), keys.size());
+      session_options.AppendExecutionProvider_Nv(*ort_provider_options);
+
+      // Enable CUDA EP for contrib ops
+      auto ort_provider_options_cuda = OrtCUDAProviderOptionsV2::Create();
+      ort_provider_options_cuda->Update(keys.data(), values.data(), keys.size());
+
+      session_options.AppendExecutionProvider_CUDA_V2(*ort_provider_options_cuda);
```
RyanUnderhill, line 499: Why isn't this just using the existing session_options.AppendExecutionProvider API? Then it's just passing "nv" as the provider name to the one function, the keys/values are shared with all provider types there.

I see the change in Ort, I'll leave a comment there too.

RyanUnderhill (changes requested):

src/python/py/models/builder.py
```diff
@@ -1045,8 +1064,22 @@ def make_layernorm(self, layer_id, layernorm, skip, simple, location):
             output_0 = "hidden_states"
         outputs = [output_0, "", "", output_3] if skip and not self.layernorm_attrs["last_layernorm"] else [output_0]
 
-        self.make_node(op_type, inputs=inputs, outputs=outputs, name=name, domain=("com.microsoft" if skip else None), **kwargs)
-        self.make_value_info(output_0, self.io_dtype, shape=['batch_size', 'sequence_length', self.hidden_size])
+        # NV EP doesn't support Skip/SimplifiedLayerNormalization, so we fallback to primitive ops
```
hanbitmyths, line 1067: This is not the design principle of model builder. If primitive ops are preferrable, it goes to exporter plus optimization using ONNX Script, which can be done well with Olive tool.

BLSharda: I believe model builder is expected to produce functionally supported onnx model for specified EP, right?

Since 'nv' EP is not going to support any contrib-ops, we require official ops conversion. There is new opset (23) going to make most of these contrib-ops as official and until it is available I am using available primitive official ops. 

> it goes to exporter plus optimization using ONNX Script, which can be done well with Olive tool.

Introducing olive dependency for producing just the functional model will make it harder for the app developers to use 'nv' EP.

shishirgoyal85: NVIDIA EP will only support standard ONNX Ops and GQA contrib Op until ONNX Attention Op is supported. We don't intend to support any contribOp since they aren't ONNX spec complaint. Our preference is:

1. ModelBuilder should be able to generate spec complaint model for NVIDIA EP i.e. generate model with Primitive Op + GQA Op only (and later ONNX attention Op). We can do this change ourselves but need MSFT help with the review. Not having this in Modelbuilder would mean that the model builder doesn't support NVIDIA EP.
2. Olive recipe : Using Olive to convert the the model from model builder (like convert ContribOp to ONNX standard Op) will become a 2 step process for the developers that we don't desire. Also, the Olive recipe to generate onnx spec complaint model is something that we are expecting MSFT to provide. Olive being a separate tool with separate release timelines will also require that future changes are synchronized.

BLSharda:
> Is it possible for NV to extend OrtSessionAppendExecutionProvider with a new name? This fits with the other new providers being added and is general purpose.

Sure! I have now updated code to use this.

src/generators.cpp
```diff
@@ -222,6 +224,7 @@ DeviceInterface* GetDeviceInterface(DeviceType type) {
     default:
     case DeviceType::CPU:
       return GetCpuInterface();
+    case DeviceType::NvTensorRtRtx: // Fallback to CUDA interface.
```
RyanUnderhill, line 227: NvTensorRtRtx will gets its own allocator but this says for it to act like the cuda device. Should it share an allocator with cuda?

BLSharda: We intend to reuse CUDA interface & device for this EP. To simplify, I have removed separate DeviceType for NvTensorRtRtx and it will use CUDA device now.

src/models/model.cpp
```diff
@@ -562,12 +545,23 @@ void Model::CreateSessionOptionsFromConfig(const Config::SessionOptions& config_
...
       for (auto& option : provider_options.options) {
         keys.emplace_back(option.first.c_str());
         values.emplace_back(option.second.c_str());
       }
       session_options.AppendExecutionProvider(provider_options.name.c_str(), keys.data(), values.data(), keys.size());
+
+      if (provider_options.name == "NvTensorRtRtx") {
+        // Enable CUDA EP for contrib ops
+        auto ort_provider_options_cuda = OrtCUDAProviderOptionsV2::Create();
+        ort_provider_options_cuda->Update(keys.data(), values.data(), keys.size());
```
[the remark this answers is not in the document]
BLSharda: This code is removed now with replacement as Generic key/values; there might be slightly different options for each EP (primarily prefix).

src/python/py/models/builder.py
```diff
@@ -1045,8 +1064,31 @@
...
+                self.make_simplified_layer_norm(layer_norn_basename, root_input, weight, shape=['batch_size', 'sequence_length', self.hidden_size])
+            elif op_type == "SkipSimplifiedLayerNormalization":
+                output_0 = f"{layer_norn_basename}/skip_simplified_layer_norm/simplified_layer_norm/Mul_1/output_0"
+                output_3 = f"{layer_norn_basename}/skip_simplified_layer_norm/Add/output_0"
+                outputs = [output_0, "", "", output_3]
+                self.make_skip_simplified_layer_norm(layer_norn_basename, root_input, skip_input, weight, shape=['batch_size', 'sequence_length', self.hidden_size])
+            elif op_type == "SkipLayerNormalization":
+                output_0 = f"{layer_norn_basename}/skip_layer_norm/LayerNormalization/output_0"
+                output_3 = f"{layer_norn_basename}/skip_layer_norm/Add/output_0"
+                outputs = [output_0, "", "", output_3]
```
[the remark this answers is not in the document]
BLSharda: This seems a false positive error from CodeQL.

src/python/py/models/builder.py
```diff
@@ -1045,8 +1064,22 @@
...
+        if self.ep == "nv":
+            layer_norn_basename = f"/model/layers.{layer_id}/{location}_layernorm"
+            if op_type == "SimplifiedLayerNormalization":
+                output_0 = f"{layer_norn_basename}/simplified_layer_norm/Mul_1/output_0"
+                outputs = [output_0]
+                self.make_simplified_layer_norm(layer_norn_basename, root_input, weight, shape=['batch_size', 'sequence_length', self.hidden_size])
+            elif op_type == "SkipSimplifiedLayerNormalization":
+                output_0 = f"{layer_norn_basename}/skip_simplified_layer_norm/simplified_layer_norm/Mul_1/output_0"
+                output_3 = f"{layer_norn_basename}/skip_simplified_layer_norm/Add/output_0"
+                outputs = [output_0, "", "", output_3]
```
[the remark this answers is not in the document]
BLSharda: This seems a false positive error from CodeQL.

src/models/model.cpp
```diff
@@ -562,12 +545,23 @@ void Model::CreateSessionOptionsFromConfig(const Config::SessionOptions& config_
         session_options.AddConfigEntry("session.intra_op.allow_spinning", "0");
       }
 
+      else if (provider_options.name == "NvTensorRtRtx")
+        p_device_ = GetDeviceInterface(DeviceType::NvTensorRtRtx);
+
```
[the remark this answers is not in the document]
BLSharda: In latest code, I have removed most of these checks for `if(provider_options.name == "NvTensorRtRtx").` There is single check just before generic `AppendExecutionProvider()` to create interface and enabling some options.

src/python/py/models/builder.py
hanbitmyths: Since ORT team has a strong opinion not to add decomposition of contrib op with primitive ops, I suggest removing builder.py changes from this PR, so we can check in nvep support at ort genai.

src/models/onnxruntime_api.h
```diff
@@ -592,7 +592,8 @@ struct OrtSessionOptions {
   OrtSessionOptions& AppendExecutionProvider_OpenVINO(const OrtOpenVINOProviderOptions& provider_options);       ///< Wraps OrtApi::SessionOptionsAppendExecutionProvider_OpenVINO
   OrtSessionOptions& AppendExecutionProvider_TensorRT(const OrtTensorRTProviderOptions& provider_options);       ///< Wraps OrtApi::SessionOptionsAppendExecutionProvider_TensorRT
   OrtSessionOptions& AppendExecutionProvider_TensorRT_V2(const OrtTensorRTProviderOptionsV2& provider_options);  ///< Wraps OrtApi::SessionOptionsAppendExecutionProvider_TensorRT
-  OrtSessionOptions& AppendExecutionProvider_MIGraphX(const OrtMIGraphXProviderOptions& provider_options);       ///< Wraps OrtApi::SessionOptionsAppendExecutionProvider_MIGraphX
+
+  OrtSessionOptions& AppendExecutionProvider_MIGraphX(const OrtMIGraphXProviderOptions& provider_options);  ///< Wraps OrtApi::SessionOptionsAppendExecutionProvider_MIGraphX
```
baijumeswani, line 596: Was this just a formatting change?

hanbitmyths (changes requested):
Since ORT team has a strong opinion not to add decomposition of contrib op with primitive ops, I suggest removing builder.py changes from this PR, so we can check in nvep support at ort genai.

src/models/model.cpp
```diff
@@ -545,6 +545,11 @@ void Model::CreateSessionOptionsFromConfig(const Config::SessionOptions& config_
         session_options.AddConfigEntry("session.intra_op.allow_spinning", "0");
       }
 
+      else if (provider_options.name == "NvTensorRtRtx") {
+        session_options.AddConfigEntry("ep.nvtensorrtrtxexecutionprovider.nv_cuda_graph_enable", "1");
+        p_device_ = GetDeviceInterface(DeviceType::CUDA);
```
RyanUnderhill, line 550:
```suggestion
        // After setting the NvTensorRtRtx provider in Onnxruntime, GenAI will then treat it as the cuda device.
        session_options.AddConfigEntry("ep.nvtensorrtrtxexecutionprovider.nv_cuda_graph_enable", "1");
        p_device_ = GetDeviceInterface(DeviceType::CUDA);
```

src/models/onnxruntime_api.h
BLSharda: Actually, there were separate NV EP specific APIs added, and now removed by using generic AppendExecutionProvider. I reverted the change but was hitting Clang-Format error in GitHub CICD so kept it with formatting change by clang-format tool.

BLSharda: Just reverted onnxruntime_api.h now

src/python/py/models/builder.py
BLSharda: Removed changes to builder.py for now from this PR, we can discuss it separately.

hanbitmyths (approved):

RyanUnderhill (approved):

PR #1405, opened 2025-04-16
Add logging of ORT version and build info.

edgchen1:
It is useful to be able to see the onnxruntime version and build info when switching between multiple onnxruntime shared libraries.


src/logging.cpp
```diff
@@ -10,7 +10,12 @@
 namespace Generators {
 
 LogItems g_log;
-static std::ostream* gp_stream{&std::cerr};
+
+static std::ostream*& GlobalLogStreamPtr() {
+  static std::ostream* stream = &std::cerr;
```
edgchen1, line 15: avoid static initialization order race by making this a function. on Windows, I encountered the case where gp_stream was nullptr.

RyanUnderhill: The root problem is that our onnxruntime header was not supposed to have any dependencies and is now calling genai code, I forget what calls Ort:: functions first but if there's a way to clean that up, we should do that.

baijumeswani (approved):

PR #1411, opened 2025-04-19
Add Python and C# bindings for Chat Template API

sayanshaw24:
- Adds Python and C# bindings for Chat Template API
- Adds second method to apply chat template _and_ tokenize (since the GenAI C API does not support variable Tensor return types)

### DIFF
--- a/src/ort_genai_c.h
+++ b/src/ort_genai_c.h
@@ -508,6 +508,22 @@ OGA_EXPORT OgaResult* OGA_API_CALL OgaProcessorDecode(const OgaMultiModalProcess
  */
 OGA_EXPORT OgaResult* OGA_API_CALL OgaTokenizerApplyChatTemplate(const OgaTokenizer*, const char* template_str, const char* messages, bool add_generation_prompt, const char** out_string);
 
+/**
+ * @brief Applies a chat template to input messages and tokenizes them
+ *
+ * This function processes the specified template with the provided input using the
+ * tokenizer, and outputs the tokenized input ids of the resulting string. Optionally,
+ * it can include a generation prompt in the output.
+ *
+ * \param[in] tokenizer OgaTokenizer used for template processing.
+ * \param[in] template_str Null-terminated string representing the chat template. Use nullptr to fall back to the default chat template from the tokenizer config.
+ * \param[in] messages Null-terminated string containing the input messages to be processed.
+ * \param[in] add_generation_prompt Indicates whether to add a generation prompt to the output.
+ * \param[out] out_ids Pointer to where the output will be stored. The returned pointer must be freed with OgaDestroyString
+ * \return OgaResult* containing the error message if the function fails
+ */
+OGA_EXPORT OgaResult* OGA_API_CALL OgaTokenizerApplyChatTemplateTokenize(const OgaTokenizer*, const char* template_str, const char* messages, bool add_generation_prompt, const char** out_ids);
+
 /** OgaTokenizerStream is to decoded token strings incrementally, one token at a time.
  */
 OGA_EXPORT OgaResult* OGA_API_CALL OgaCreateTokenizerStream(const OgaTokenizer*, OgaTokenizerStream** out);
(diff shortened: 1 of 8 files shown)


src/ort_genai_c.h
```diff
@@ -508,6 +508,22 @@ OGA_EXPORT OgaResult* OGA_API_CALL OgaProcessorDecode(const OgaMultiModalProcess
...
+ *
+ * \param[in] tokenizer OgaTokenizer used for template processing.
+ * \param[in] template_str Null-terminated string representing the chat template. Use nullptr to fall back to the default chat template from the tokenizer config.
+ * \param[in] messages Null-terminated string containing the input messages to be processed.
+ * \param[in] add_generation_prompt Indicates whether to add a generation prompt to the output.
+ * \param[out] out_ids Pointer to where the output will be stored. The returned pointer must be freed with OgaDestroyString
+ * \return OgaResult* containing the error message if the function fails
+ */
+OGA_EXPORT OgaResult* OGA_API_CALL OgaTokenizerApplyChatTemplateTokenize(const OgaTokenizer*, const char* template_str, const char* messages, bool add_generation_prompt, OgaSequences* sequences);
+
```
RyanUnderhill, line 526: As we discussed offline, this can be removed. Just need the one to apply the chat template. Tokenization will use the existing API to keep it simple (just one way to do it vs two)

sayanshaw24: yep, done!

src/python/python.cpp
```diff
@@ -353,6 +353,11 @@ PYBIND11_MODULE(onnxruntime_genai, m) {
       })
       .def("to_token_id", &OgaTokenizer::ToTokenId)
       .def("decode", [](const OgaTokenizer& t, pybind11::array_t<int32_t> tokens) -> std::string { return t.Decode(ToSpan(tokens)).p_; })
+      .def(
+          "apply_chat_template", [](const OgaTokenizer& t, const char* template_str, const char* messages, pybind11::bool_ add_generation_prompt) -> std::string {
+            return t.ApplyChatTemplate(template_str, messages, static_cast<bool>(add_generation_prompt)).p_;
+          },
+          pybind11::arg("template_str") = nullptr, pybind11::arg("messages"), pybind11::arg("add_generation_prompt"))
```
RyanUnderhill, line 360: Why does it need pybind11::bool_? Is it for the pybind11::arg?
Why do "messages" and "add_generation_prompt" need pybind11::arg params, as they have no defaults set?

sayanshaw24: ah i just figured that was the norm, like using pybind11::array_t<int32_t> instead of std::vector as is done above in other methods, just updated it!

and you're right it's unnecessary i think, fixed!

RyanUnderhill: Ah, makes sense. std::vector maps to a python list, vs a numpy array. So that's why that one is explicit (I had that wrong in the tokenizer case). But otherwise we just leave as C++ types when possible.

RyanUnderhill (dismissed):

src/python/python.cpp
sayanshaw24: update: "messages" and "add_generation_prompt" need pybind11::arg params as otherwise it fails with: `error: static assertion failed: The number of argument annotations does not match the number of function arguments`

so if you add one optional param, you have to explicitly state the other params too (kind of like in method calls); added them back.

src/python/python.cpp
```diff
@@ -353,6 +353,11 @@ PYBIND11_MODULE(onnxruntime_genai, m) {
       })
       .def("to_token_id", &OgaTokenizer::ToTokenId)
       .def("decode", [](const OgaTokenizer& t, pybind11::array_t<int32_t> tokens) -> std::string { return t.Decode(ToSpan(tokens)).p_; })
+      .def(
+          "apply_chat_template", [](const OgaTokenizer& t, const char* template_str, const char* messages, bool add_generation_prompt) -> std::string {
+            return t.ApplyChatTemplate(template_str, messages, static_cast<bool>(add_generation_prompt)).p_;
```
RyanUnderhill, line 358: You still have the static_cast<bool> for a bool

RyanUnderhill (approved):

PR #1413, opened 2025-04-22
C++ and python samples: follow_config support

RyanMetcalfeInt8:
This PR adds the ability for `follow_config` to be given as the execution_provider option for C++ (Phi) example, (most) python samples, and benchmark_e2e.py tool.

When `follow_config` is set as the EP, the sample won't perform the `config.clear_providers()` operation, which was problematic in cases where important provider options, etc. are specified in `genai_config.json`

### DIFF
--- a/benchmark/python/benchmark_e2e.py
+++ b/benchmark/python/benchmark_e2e.py
@@ -212,11 +212,12 @@ def run_benchmark(args, batch_size, prompt_length, generation_length, max_length
     # Get tokenizer, and model
     if args.verbose: print("Getting config")
     config = og.Config(f'{args.input_folder}')
-    config.clear_providers()
+    if args.execution_provider != "follow_config":
+        config.clear_providers()
+        if args.execution_provider != "cpu":
+            if args.verbose: print(f"Setting model to {args.execution_provider}")
+            config.append_provider(args.execution_provider)
     if args.verbose: print("Loading model... ")
-    if args.execution_provider != "cpu":
-        if args.verbose: print(f"Setting model to {args.execution_provider}")
-        config.append_provider(args.execution_provider)
     model = og.Model(config)
     if args.verbose: print("Model loaded")
     tokenizer = og.Tokenizer(model)
@@ -446,7 +447,7 @@ if __name__ == "__main__":
     parser.add_argument('--use_random_tokens', action='store_true', help='Use random tokens instead of generating a prompt')
     parser.add_argument('--use_prompt_set', action='store_true', help='Use pre-generated prompt set instead of generating a prompt')
     parser.add_argument('--chat_template', type=str, default='', help='Chat template to use for the prompt. User input will be injected into {input}')
-    parser.add_argument('-e', '--execution_provider', type=str, required=True, choices=["cpu", "cuda", "dml"], help='Execution provider to run ONNX model with')
+    parser.add_argument('-e', '--execution_provider', type=str, required=False, default='follow_config', choices=["cpu", "cuda", "dml", "follow_config"], help="Execution provider to run ONNX model with")
     args = parser.parse_args()
 
     # check max_lengths
--- a/benchmark/python/benchmark_e2e.py
+++ b/benchmark/python/benchmark_e2e.py
@@ -447,7 +447,7 @@ if __name__ == "__main__":
     parser.add_argument('--use_random_tokens', action='store_true', help='Use random tokens instead of generating a prompt')
     parser.add_argument('--use_prompt_set', action='store_true', help='Use pre-generated prompt set instead of generating a prompt')
     parser.add_argument('--chat_template', type=str, default='', help='Chat template to use for the prompt. User input will be injected into {input}')
-    parser.add_argument('-e', '--execution_provider', type=str, required=False, default='follow_config', choices=["cpu", "cuda", "dml", "follow_config"], help="Execution provider to run ONNX model with")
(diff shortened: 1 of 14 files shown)


RyanMetcalfeInt8:
Hi @RyanUnderhill , @baijumeswani -- could you review? Happy to make any recommended changes, as usual. Thanks!

examples/c/src/phi3v.cpp
```diff
@@ -12,12 +12,14 @@ void CXX_API(const char* model_path, const char* execution_provider) {
...
-    config->AppendProvider(execution_provider);
-    if (provider.compare("cuda") == 0) {
-      config->SetProviderOption(execution_provider, "enable_cuda_graph", "0");
+  if (provider.compare("follow_config") != 0) {
+    config->ClearProviders();
+    if (provider.compare("cpu") != 0) {
+      config->AppendProvider(execution_provider);
+      if (provider.compare("cuda") == 0) {
+        config->SetProviderOption(execution_provider, "enable_cuda_graph", "0");
+      }
```
RyanUnderhill, line 22: Is it easy to move this into common.cpp? It appears to be the same in every cpp example

examples/c/src/common.cpp
```diff
@@ -88,7 +88,5 @@ std::string trim(const std::string& str) {
 }
 
 void print_usage(int /*argc*/, char** argv) {
-  std::cerr << "usage: " << argv[0] << std::endl;
-  std::cerr << "model_path = " << argv[1] << std::endl;
-  std::cerr << "execution_provider = " << argv[2] << std::endl;
+  std::cerr << "usage: " << argv[0] << " <model_path> <execution_provider>" << std::endl;
```
RyanUnderhill, line 91: What did we lose here, why is argv[1] and argv[2] not being printed anymore?

examples/c/src/phi3v.cpp
RyanMetcalfeInt8: Yeah I think this one is easy. I can do that.

examples/c/src/common.cpp
RyanMetcalfeInt8: ah, so `print_usage` here is called when user doesn't (potentially) pass model_path and/or execution provider. So the issue was that this was triggering a crash (as argv[1] / argv[2] were NULL)

RyanMetcalfeInt8: (so it's outside the scope of these `follow_config` changes.. I just threw it in here to fix the crash.

RyanUnderhill: Interesting, I pinged another teammate about it to see if they know how it was originally supposed to work.

examples/c/src/phi3v.cpp
RyanMetcalfeInt8: Alright, done -- also moved some of the argument parsing logic to common as well. Take a look when you get a chance.

examples/c/src/common.cpp
RyanMetcalfeInt8: FYI -- In one of the commits I pushed this morning, I also added more details to the printed usage message.

RyanMetcalfeInt8:
Hi all, any further updates required on this one?

benchmark/python/benchmark_e2e.py
```diff
@@ -446,7 +447,7 @@ def str2strlist(value):
     parser.add_argument('--use_random_tokens', action='store_true', help='Use random tokens instead of generating a prompt')
     parser.add_argument('--use_prompt_set', action='store_true', help='Use pre-generated prompt set instead of generating a prompt')
     parser.add_argument('--chat_template', type=str, default='', help='Chat template to use for the prompt. User input will be injected into {input}')
-    parser.add_argument('-e', '--execution_provider', type=str, required=True, choices=["cpu", "cuda", "dml"], help='Execution provider to run ONNX model with')
+    parser.add_argument('-e', '--execution_provider', type=str, required=False, default='follow_config', choices=["cpu", "cuda", "dml", "follow_config"], help="Execution provider to run ONNX model with")
```
baijumeswani, line 450:
```suggestion
    parser.add_argument('-e', '--execution_provider', type=str, required=False, default='follow_config', choices=["cpu", "cuda", "dml", "follow_config"], help="Execution provider to run the ONNX Runtime session with. Defaults to follow_config that uses the execution provider listed in the genai_config.json instead.")
```

examples/python/model-chat.py
```diff
@@ -141,7 +142,7 @@ def main(args):
 if __name__ == "__main__":
     parser = argparse.ArgumentParser(argument_default=argparse.SUPPRESS, description="End-to-end AI Question/Answer example for gen-ai")
     parser.add_argument('-m', '--model_path', type=str, required=True, help='Onnx model folder path (must contain genai_config.json and model.onnx)')
-    parser.add_argument('-e', '--execution_provider', type=str, required=True, choices=["cpu", "cuda", "dml"], help="Execution provider to run ONNX model with")
+    parser.add_argument('-e', '--execution_provider', type=str, required=False, default='follow_config', choices=["cpu", "cuda", "dml", "follow_config"], help="Execution provider to run ONNX model with")
```
baijumeswani, line 145: Same as the other comment about explaining what `follow_config` means.

baijumeswani (commented):
a couple of small comments, otherwise looks good.

benchmark/python/benchmark_e2e.py
RyanMetcalfeInt8: done, please see latest commit.

examples/python/model-chat.py
RyanMetcalfeInt8: done, please see latest commit.

baijumeswani:
Looks like there were some clang-format errors. Could you please address those?

baijumeswani (approved):

Issue #707, opened 2024-07-17
Nodes are not topologically sorted from models generated by model builder


BowenBao:
**Describe the bug**
As title. By onnx spec the nodes in graph should be topologically sorted.

Discovered while loading the model in onnxscript ir.

**To Reproduce**
1. Generate an Llama2 model with MHA using model builder.
2. Load the model in onnxscript ir

```python
import onnx
from onnxscript import ir

model_path = "model.onnx"

onnx_model = onnx.load(model_path)
model = ir.serde.deserialize_model(onnx_model)

model.graph.display(page=False)
```

Observe
```
Input '/model/attn_mask_reformat/input_ids_subgraph/Gather_2/output_0' of node '/model/attn_mask_reformat/input_ids_subgraph/Add_1(::Add:)' not found in any scope. The graph may be unsorted. Creating a new input (current depth: 1) .
...
      3 |  # /model/attn_mask_reformat/input_ids_subgraph/Add_1
           %"/model/attn_mask_reformat/input_ids_subgraph/Add_1/output_0"<INT64,[]> ⬅️ ::Add(%"/model/attn_mask_reformat/input_ids_subgraph/Gather_2/output_0", %"/model/attn_mask_reformat/past_key_subgraph/Gather/output_0")
...
     35 |  # /model/attn_mask_reformat/input_ids_subgraph/Gather_2
           %"/model/attn_mask_reformat/input_ids_subgraph/Gather_2/output_0"<INT64,[]> ⬅️ ::Gather(%"/model/attn_mask_reformat/input_ids_subgraph/Shape_2/output_0", %"/model/constants/TensorProto.INT64/0D/1") {axis=0}
```

cc @kunal-vaishnavi , @justinchuby

justinchuby:
Yeah that’s known. @kunal-vaishnavi perhaps this can be improved?

We can also create a topo sort pass for this

justinchuby:
@BowenBao note that the graph in IR is still representing what's in the file and the connections should be correct. You may sort the graph in IR to fix any issues.

xenova:
Just ran into this again today - is there any fix planned? Thanks!

Issue #675, opened 2024-07-04
`models/builder.py` creates empty output directory


mlinke-ai:
I have downloaded the `microsoft/phi-3-mini-128k-instruct` model from huggingface using the `huggingface-cli` script.

When I try to convert the model to ONNX format, the directory specified with the `-o` flag stays empty.

I use the following command:

```text
python -m onnxruntime_genai.models.builder -m phi-3-mini-128k-instruct-onnx-cpu -i phi-3-mini-128k-instruct -o phi-3-mini-128k-instruct-onnx-cpu -p int4 -e cpu -c cache --extra_options int4_accuracy_level=1 filename=phi3-mini-128k-instruct-onnx-cpu.onnx
```

- python: 3.10.8
- onnxruntime: 1.18.1
- onnxruntime-genai: 0.3.0

kunal-vaishnavi:
https://github.com/microsoft/onnxruntime-genai/blob/4b0dd43064287dfbd6f13f16593d36b5525c0cdc/src/python/py/models/builder.py#L2412-L2430

Since you already downloaded the model onto disk and you are providing an input path to a directory, you don't need the `-m phi-3-mini-128k-instruct-onnx-cpu` part in your command. Can you omit that part and try again?

mlinke-ai:
Omitting the `-m phi-3-mini-128k-instruct-onnx-cpu` did not work. The directory is still empty.

I have some warnings about missing `flash_attn`. But I don't think this could be the problem.

natke:
Hi @mlinke-ai, can you share the full output of the command here. And what are the specs of the machine you are running on?

mlinke-ai:
The complete output of the command is as following (slightly shortened to remove some clutter):

```text
Valid precision + execution provider combinations are: FP32 CPU, FP32 CUDA, FP16 CUDA, FP16 DML, INT4 CPU, INT4 CUDA, INT4 DML
Extra options: {'int4_accuracy_level': '1', 'filename': 'phi3-mini-128k-instruct-onnx-cpu.onnx'}
C:\Users\mlinke\AppData\Roaming\Python\Python310\site-packages\transformers\models\auto\configuration_auto.py:950: FutureWarning: The `use_auth_token` argument is deprecated and will be removed in v5 of Transformers. Please use `token` instead.
  warnings.warn(
GroupQueryAttention (GQA) is used in this model.
[...]

Issue #319, opened 2024-04-24
cannot build phi2 model on system with 32 GB ram 


liqunfu:
python -m onnxruntime_genai.models.builder -m microsoft/phi-2 -e cpu -p int4 --extra_options int4_accuracy_level=4  -o ./example-models/phi2-int4-cpu

the system memory usage keep increasing up to 30 GB and it exit without any message.

natke:
Hi @liqunfu, can you share which version of onnxruntime-genai you are using and your O/S and machine specs please.

PR #1416, opened 2025-04-23
Use ONNX IR for model builder

justinchuby:
Use ONNX IR for constructing model for model builder. ONNX IR provides a more ergonomic, efficient and robust set of APIs for model construction. imports from `onnx` are removed as a result.

- Refactored and streamlined int4 handling
- Improved torch tensor to numpy conversion calls
- Avoid the use of eval: https://stackoverflow.com/questions/1832940/why-is-using-eval-a-bad-practice
- Models are topologically sorted. Fixes https://github.com/microsoft/onnxruntime-genai/issues/707

Memory reduction:
(left: with IR - 7.3GiB, right: current main - 24.9 GiB)

`memray run src/python/py/models/builder.py -m meta-llama/Llama-3.2-1
B-Instruct -o path_to_output_folder -p fp32 -c cache_dir_to_save_hf_files -e cpu`

image

image (1)

This change will help address the following issues:
- https://github.com/microsoft/onnxruntime-genai/issues/675
- https://github.com/microsoft/onnxruntime-genai/discussions/489
- https://github.com/microsoft/onnxruntime-genai/issues/319


justinchuby:
Working on in/out types

src/python/py/models/builder.py
```diff
@@ -551,65 +638,68 @@ def make_node(self, op_type, inputs, outputs, name=None, doc_string=None, domain
...
+            outputs.append(self._get_or_create_value(value_name, dtype=self.output_types["present.value"], shape=self.output_shapes["present.value"]))
 
     def make_constant(self, name):
         # Make constant ops for 0, 1, 2, 3, etc.
         # Format of name is "/model/constants/{dtype}/{shape}/{num}"
+
         path = name.split("/")
-        onnx_dtype, dims, num = eval(path[-3]), path[-2], eval(path[-1])
-        np_dtype = self.to_numpy_dtype[onnx_dtype]
-        value = numpy_helper.from_array(np.array(num if dims == "0D" else list(num) if type(num) == tuple else [num], dtype=np_dtype), name=name.replace("constants", "numpy_helper"))
```
justinchuby, line 602: I made this more save

justinchuby:
@kunal-vaishnavi tests looking good! Will run again when the other test models are added.

justinchuby:
The reason loading from hf was taking memory was there was casting going on (and it needs to hold onto the casted tensors):

image

If we can delay this process and do the cast ourselves when it's time to save model, we will be able to reduce the bulk of memory consumption _and_ save the need for an additional casting.

src/python/py/models/builder.py
```diff
@@ -2571,23 +2575,23 @@ def make_input_ids_subgraph(self, basename, past_key_gather_name):
...
-        self.make_concat(concat_3_name, concat_3_inputs, dtype=TensorProto.INT64, shape=[2], axis=0)
+        self.make_concat(concat_3_name, concat_3_inputs, dtype=ir.DataType.INT64, shape=[2], axis=0)
 
         # Bottom path
         shape_5_name = f"{basename}/Shape_5"
         self.make_shape(shape_5_name, f"{constant_shape_name}/output_0", shape=[2])
         slice_2_name = f"{basename}/Slice_2"
-        slice_2_inputs = [f"{shape_5_name}/output_0", "/model/constants/TensorProto.INT64/1D/-1", f"/model/constants/TensorProto.INT64/1D/{torch.iinfo(torch.int64).max}", "/model/constants/TensorProto.INT64/1D/0"]
-        self.make_slice(slice_2_name, slice_2_inputs, dtype=TensorProto.INT64, shape=[1])
+        slice_2_inputs = [f"{shape_5_name}/output_0", "/model/constants/TensorProto.INT64/1D/-1", f"/model/constants/TensorProto.INT64/1D/{np.iinfo(np.int64).max}", "/model/constants/TensorProto.INT64/1D/0"]
```
justinchuby, line 2593: update

src/python/py/models/builder.py
```diff
@@ -2878,8 +2882,18 @@ def make_attention_init(self):
         super().make_attention_init()
 
 
+class Qwen3Model(QwenModel):
```
justinchuby, line 2885: revert?

src/python/py/models/builder.py
```diff
@@ -3029,6 +3043,50 @@ def make_position_ids_reformatting(self):
 
         return add_1_name
 
+    def make_rotary_embedding_caches(self, **kwargs):
```
justinchuby, line 3046: revert?

src/python/py/models/builder.py
```diff
@@ -3512,25 +3549,30 @@ def create_model(model_name, input_path, output_dir, precision, execution_provid
...
             config.hidden_act = "swiglu"
-            onnx_model = ChatGLMModel(config, io_dtype, precision, execution_provider, cache_dir, extra_options)
+            onnx_model = ChatGLMModel(config, io_dtype, onnx_dtype, execution_provider, cache_dir, extra_options)
         elif config.architectures[0] == "GemmaForCausalLM":
-            onnx_model = GemmaModel(config, io_dtype, precision, execution_provider, cache_dir, extra_options)
+            onnx_model = GemmaModel(config, io_dtype, onnx_dtype, execution_provider, cache_dir, extra_options)
         elif config.architectures[0] == "Gemma2ForCausalLM":
             if precision == "fp16":
                 print("WARNING: This model loses accuracy with float16 precision. Setting `--precision bf16` by default.")
                 precision = "bf16"
```
justinchuby, line 3569: Update

src/python/py/models/builder.py
```diff
@@ -3512,25 +3549,30 @@ def create_model(model_name, input_path, output_dir, precision, execution_provid
...
-            onnx_model = Gemma2Model(config, io_dtype, precision, execution_provider, cache_dir, extra_options)
+            onnx_model = Gemma2Model(config, io_dtype, onnx_dtype, execution_provider, cache_dir, extra_options)
         elif config.architectures[0] == "Gemma3ForCausalLM":
             if precision == "fp16":
                 print("WARNING: This model loses accuracy with float16 precision. Setting `--precision bf16` by default.")
                 precision = "bf16"
```
justinchuby, line 3574: Update here and below

src/python/py/models/builder.py
```diff
@@ -443,44 +467,74 @@ def save_processing(self, model_name_or_path, extra_kwargs, out_dir):
...
-        )
-
-        # Load external data into ONNX model
-        external_data_helper.load_external_data_for_model(model, self.cache_dir)
 
-        # Delete external data files on disk before re-saving
-        for path in os.listdir(self.cache_dir):
-            if path.endswith(".bin"):
-                os.remove(os.path.join(self.cache_dir, path))
+    def _get_prompt_templates(self, hf_name, extra_kwargs):
```
justinchuby, line 471: Remove

src/python/py/models/builder.py
```diff
@@ -321,11 +327,30 @@ def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
...
+            nodes=(),
+            opset_imports={
+                "": 21 if self.quant_attrs["use_qdq"] else 14,
+                "com.microsoft": 1,
+            },
+            name="main_graph",
+        )
+        # A tracer for recording the nodes added
+        # TODO(justinchuby): Bump IR version to 10
+        self._model = ir.Model(graph, ir_version=7, producer_name="onnxruntime-genai")
```
kunal-vaishnavi: Move the location of the removed inputs

kunal-vaishnavi: Make the class attribute names without a `_` prefix

src/python/py/models/builder.py
```diff
@@ -445,44 +469,50 @@ def save_processing(self, model_name_or_path, extra_kwargs, out_dir):
...
+        return self._values.setdefault(name, ir.Value(name=name))
 
+    def as_ir_model(self) -> ir.Model:
+        """Return the IR model."""
         # Quantize ONNX model to desired precision
         # TODO: Replace by quantizing the MatMuls as they are created
         already_quantized_in_qdq_format = self.quant_type is not None and self.quant_attrs["use_qdq"]  # Skip quantizing `MatMul` in `DequantizeLinear --> Transpose --> MatMul` path
-        if self.onnx_dtype == "int4" and not already_quantized_in_qdq_format:
-            model = self.to_int4(model)
+        if self.onnx_dtype == ir.DataType.UINT4 and not already_quantized_in_qdq_format:
```
kunal-vaishnavi: Let's keep this logic during saving and omit `as_ir_model`

src/python/py/models/builder.py
```diff
@@ -493,17 +523,27 @@ def save_model(self, out_dir):
...
 
+        # TODO(justinchuby): This makes the model not savable again when the weight
+        # files are deleted.
+        # Delete external data files on disk before re-saving
+        for path in os.listdir(self.cache_dir):
+            if path.endswith(".bin"):
+                try:
+                    os.remove(os.path.join(self.cache_dir, path))
+                except OSError:
+                    logger.debug("Error deleting file", stack_info=True)
```
kunal-vaishnavi: Use print instead of logger to keep it simple for now

src/python/py/models/builder.py
```diff
@@ -532,80 +572,33 @@ def make_int4_algo_config(self, quant_method):
...
-
-        if kwargs.get("unpack_int4", False) and self.onnx_dtype == 'int4':
-            tensor_proto.data_type = TensorProto.UINT4
-            tensor_proto.dims[-1] *= 2
-
-        self.initializers.append(tensor_proto)
+        (ir_tensor,) = ir.external_data.convert_tensors_to_external(
+            [ir_tensor], self.cache_dir, filename
+        )
+        initializer = self.make_value(name)
```
```suggestion
        value = self.make_value(name)
```

src/python/py/models/builder.py
```diff
@@ -445,44 +469,50 @@ def save_processing(self, model_name_or_path, extra_kwargs, out_dir):
...
+            name: The name of the value.
+            output: Whether the value is an output value.
+        """
+        if name == "":
+            if not output:
+                # ONNX IR uses None to represent an empty input
+                return None
+            else:
+                # Use empty string to represent an empty output because all outputs
+                # need to be of type ir.Value
```
kunal-vaishnavi: Can the `else` condition always be returned whether an input or an output is used? That would make this simpler to read and understand

src/python/py/models/builder.py
```diff
@@ -532,80 +572,33 @@ def make_int4_algo_config(self, quant_method):
...
-        if name not in self.node_names:
-            node = helper.make_node(op_type, inputs, outputs, name, doc_string, domain, **kwargs)
-            if doc_string == '':
-                node.doc_string = ''
-            self.order_repeated_field(node.attribute, 'name', kwargs.keys())
-            self.nodes.append(node)
-            self.node_names.add(name)
+        # Resolve values from names
+        input_values = [self.make_value(name) for name in inputs]
+        output_values = [self.make_value(name, output=True) for name in outputs]
```
kunal-vaishnavi: We should be able to use `self.make_value(name)` for both input and output values. See [this comment](https://github.com/microsoft/onnxruntime-genai/pull/1416#discussion_r2138779123) for more details.

src/python/py/models/builder.py
```diff
@@ -617,67 +610,72 @@ def make_node(self, op_type, inputs, outputs, name=None, doc_string=None, domain
...
     def make_inputs_and_outputs(self):
         # Add model-specific inputs to list of model inputs
-        inputs = []
+        inputs = self._model.graph.inputs
         for name in self.input_names:
             dtype = self.input_types[name]
             shape = self.input_shapes[name]
-            inputs.append(helper.make_tensor_value_info(name, dtype, shape=shape))
+            inputs.append(self.make_value(name))
+            self.make_value_info(name, dtype=dtype, shape=shape)
```
kunal-vaishnavi: Let's remove `make_value` for now and use `make_value_info` everywhere instead. Having both can be confusing. This will keep the PR changes smaller, and we can rename `make_value_info` as needed in a subsequent PR.

src/python/py/models/builder.py
```diff
@@ -867,9 +865,9 @@ def make_matmul_int4(self, matmul, basename, root_input, **kwargs):
 
         # Input weights are quantized, save quantized MatMul weights for onnx model
         weight = name[1:].replace("/", ".") + ".qweight"
-        self.make_external_tensor(matmul.qweight.detach().cpu().contiguous(), weight)
+        self.make_external_tensor(matmul.qweight, weight)
         scales = name[1:].replace("/", ".") + ".scales"
-        self.make_external_tensor(matmul.scales.detach().cpu().to(self.to_torch_dtype[self.io_dtype]).contiguous(), scales)
+        self.make_external_tensor(matmul.scales.detach().to(self.to_torch_dtype[self.io_dtype]).contiguous(), scales)
```
kunal-vaishnavi, line 870: Since the ONNX IR does not care if the input PyTorch tensor is on CPU or GPU and does not care whether the input tensor is contiguous or not, let's remove any references to `.cpu()` and `.contiguous()` to make the external tensor lines shorter.

kunal-vaishnavi: Since the ONNX IR does not care if the input tensor is on CPU or GPU and does not care whether the input tensor is contiguous or not, let's remove any references to `.cpu()` and `.contiguous()` to make the external tensor lines shorter.

src/python/py/models/builder.py
```diff
@@ -896,22 +894,24 @@ def make_matmul_int4(self, matmul, basename, root_input, **kwargs):
...
 
         dequantize_inputs = [qweight, scales]
 
         if hasattr(quantized_op, "qzeros") and quantized_op.qzeros is not None:
             zeros = dequantize_name[1:].replace("/", ".") + ".qzeros"
             zeros_pt = quantized_op.qzeros.detach().cpu()
-            zeros_pt = zeros_pt.reshape(*qweight_pt.shape[:-1], qweight_pt.shape[-1] // quantized_op.group_size)
-            self.make_external_tensor(zeros_pt.contiguous(), zeros, unpack_int4=True)
+            zeros_pt = zeros_pt.reshape(*qweight_npy.shape[:-1], qweight_npy.shape[-1] // quantized_op.group_size)
+            # TODO(justinchuby): Do we need to uncpack the zeros tensor?
```
```suggestion
            # TODO(justinchuby): Do we need to unpack the zeros tensor?
```

src/python/py/models/builder.py
```diff
@@ -896,22 +894,24 @@ def make_matmul_int4(self, matmul, basename, root_input, **kwargs):
...
+        scales_pt = scales_pt.reshape(*qweight_npy.shape[:-1], qweight_npy.shape[-1] * 2 // quantized_op.group_size)
         self.make_external_tensor(scales_pt.contiguous(), scales)
 
         dequantize_inputs = [qweight, scales]
 
         if hasattr(quantized_op, "qzeros") and quantized_op.qzeros is not None:
             zeros = dequantize_name[1:].replace("/", ".") + ".qzeros"
             zeros_pt = quantized_op.qzeros.detach().cpu()
-            zeros_pt = zeros_pt.reshape(*qweight_pt.shape[:-1], qweight_pt.shape[-1] // quantized_op.group_size)
-            self.make_external_tensor(zeros_pt.contiguous(), zeros, unpack_int4=True)
```
kunal-vaishnavi, line 902: I think we need to unpack the zeros

src/python/py/models/builder.py
```diff
@@ -3737,7 +3756,6 @@ def create_model(model_name, input_path, output_dir, precision, execution_provid
 
     # Set input/output precision of ONNX model
     io_dtype = set_io_dtype(precision, execution_provider, extra_options)
-
```
kunal-vaishnavi, line 3732: Let's have a function that sets the `onnx_dtype` as an `ir.DataType` since that dtype is internal to the ONNX model. Then we can keep the `onnx_dtype` name in the class constructors.

src/python/py/models/builder.py

src/python/py/models/builder.py

src/python/py/models/builder.py

src/python/py/models/builder.py

src/python/py/models/builder.py

src/python/py/models/builder.py

src/python/py/models/builder.py

src/python/py/models/builder.py

src/python/py/models/builder.py

src/python/py/models/builder.py
justinchuby: Done

src/python/py/models/builder.py
justinchuby: Done

src/python/py/models/builder.py
justinchuby: I moved them back because the graph requires self.quant_attrs to decide the opset version, which is created at the end of the init method. Once we universally bump the opset version we can update this, but for now it needs to be at the end.

src/python/py/models/builder.py
```diff
@@ -7,23 +7,85 @@
...
+    AutoModelForCausalLM,
+    AutoTokenizer,
+    GenerationConfig,
+)
+
+
+# NOTE: Avoid importing from onnx helper and numpy_helper. Instead, leverage
+# ONNX IR methods like ir.tensor, ir.DataType.numpy() and other methods for constructing
+# the ONNX graph efficiently.
+
```
kunal-vaishnavi, line 39: Can we keep the `unpack_uint4` methods inside `make_dequantize_linear` instead of global methods since they will eventually be removed once the ONNX IR support is added?

src/python/py/models/builder.py
```diff
@@ -532,140 +529,157 @@ def to_int4(self, model):
...
-        tensor_proto.ClearField("raw_data")
-        tensor_proto.data_location = TensorProto.EXTERNAL
+        already_quantized_in_qdq_format = self.quant_type is not None and self.quant_attrs["use_qdq"]  # Skip quantizing `MatMul` in `DequantizeLinear --> Transpose --> MatMul` path
+        if self.onnx_dtype == ir.DataType.UINT4 and not already_quantized_in_qdq_format:
+            model = self.to_int4()
+        else:
+            model = self.model
+
+        # Make sure all nodes are topologically sorted
+        model.graph.sort()
```
kunal-vaishnavi, line 518: Can you edit the PR description to say that topological sort has been added and fixes [this issue](https://github.com/microsoft/onnxruntime-genai/issues/707)?

src/python/py/models/builder.py
```diff
@@ -3097,7 +3130,7 @@ def make_attention_init(self):
 
 
 class PhiModel(Model):
-    def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
+    def __init__(self, config, io_dtype: int, onnx_dtype: ir.DataType, ep: str, cache_dir, extra_options):
```
kunal-vaishnavi: Some of the model classes are using type hints in the constructor and some aren't. Let's stick to using the type hints on either all of the constructors or none of them.

justinchuby: Will add the rest

src/python/py/models/builder.py
justinchuby: Done

src/python/py/models/builder.py
justinchuby: Removed

src/python/py/models/builder.py
justinchuby: Done

src/python/py/models/builder.py
```diff
@@ -821,11 +804,11 @@ def make_matmul(self, matmul, basename, root_input, **kwargs):
         else:
             # For regular `MatMul`
             return self.make_matmul_op(matmul, basename, root_input, **kwargs)
-    
+
     def make_matmul_op(self, matmul, basename, root_input, **kwargs):
-        if self.onnx_dtype in {"bf16", "fp16", "fp32"}:
+        if self.onnx_dtype in {ir.DataType.FLOAT16, ir.DataType.FLOAT}:
```
kunal-vaishnavi: This is missing bfloat16

src/python/py/models/builder.py
```diff
@@ -3698,20 +3716,35 @@ def parse_hf_token(hf_token):
...
+    if precision == "int4":
+        if extra_options.get("int4_is_symmetric", True):
+            return ir.DataType.INT4
+        else:
+            return ir.DataType.UINT4
+    return {
+        "fp32": ir.DataType.FLOAT,
+        "fp16": ir.DataType.FLOAT16,
+        "bf16": ir.DataType.BFLOAT16,
+        "int4": ir.DataType.UINT4,
```
kunal-vaishnavi: I think this should be `ir.DataType.INT4` by default since `int4_is_symmetric` is true when not specified.

src/python/py/models/builder.py
```suggestion
        if self.onnx_dtype in {ir.DataType.FLOAT16, ir.DataType.BFLOAT16, ir.DataType.FLOAT}:
```

src/python/py/models/builder.py
```diff
@@ -3698,20 +3716,35 @@ def parse_hf_token(hf_token):
...
+    }[precision]
+
 
 
-def create_model(model_name, input_path, output_dir, precision, execution_provider, cache_dir, **extra_options):
+def create_model(model_name, input_path, output_dir, precision: str, execution_provider, cache_dir, **extra_options):
```
```suggestion
def create_model(model_name, input_path, output_dir, precision, execution_provider, cache_dir, **extra_options):
```

src/python/py/models/builder.py
justinchuby: Fixed

src/python/py/models/builder.py
```suggestion
```

justinchuby: Removed because it's handled above

src/python/py/models/builder.py
justinchuby: Partially annotated function is usually better than not annotated function because it gives _some_ hints to devs and tools. The types can be gradually added. I removed it here to save the discussion for future

src/python/py/models/builder.py
```diff
@@ -7,23 +7,54 @@
...
+# ONNX IR methods like ir.tensor, ir.DataType.numpy() and other methods for constructing
+# the ONNX graph efficiently.
 
 
 class Model:
-    def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options): 
+    def __init__(
+        self,
+        config,
+        io_dtype: Literal[ir.DataType.FLOAT, ir.DataType.FLOAT16] | int,
```
kunal-vaishnavi: The `io_dtype` can also be bfloat16

```suggestion
        io_dtype: Literal[
            ir.DataType.FLOAT,
            ir.DataType.FLOAT16,
            ir.DataType.BFLOAT16,
        ]
        | int,
```

kunal-vaishnavi (approved):

justinchuby:
Thanks for reviewing!

src/python/py/models/builder.py
```diff
@@ -1396,35 +1400,53 @@ def make_rotary_embedding_multi_cache(self, **kwargs):
...
+        sin_cache_small_node = ir.node(
+            "Constant", [], outputs=[
+                ir.Value(name=sin_cache_small_name, type=ir.TensorType(self.io_dtype), shape=sin_cache_small.shape)
+            ],
+            name="/small/sin_cache/Constant", attributes=dict(value=ir.tensor(sin_cache_small)))
         self.make_node(
-            "If", inputs=[f"{greater_name}/output_0"], outputs=[cos_cache_name, sin_cache_name], name=if_name,
-            then_branch=self.make_graph(
-                name="large_rotemb_caches_graph",
+            "If", inputs=[f"{greater_name}/output_0"], outputs=[if_cos_cache_output, if_sin_cache_output], name=if_name,
```
jiafatom, line 1425: NameError: name 'if_cos_cache_output' is not defined

jiafatom: Seems it should be `cos_cache_name` and `sin_cache_name`?

justinchuby: https://github.com/microsoft/onnxruntime-genai/pull/1562

PR #1418, opened 2025-04-23
Update the C++ clang-format lint workflow to use clang 20

snnn:
Update the C++ clang-format lint workflow to use clang 20, and format code with the new clang version.

### DIFF
--- a/.github/workflows/clang-format-lint.yml
+++ b/.github/workflows/clang-format-lint.yml
@@ -1,22 +1,34 @@
 name: Clang Format Lint
-concurrency:
-  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
-  cancel-in-progress: true
+
 on:
-  pull_request: { }
+  push:
+    branches:
+      - main
+      - rel-*
+  pull_request:
+    branches:
+      - main
+      - rel-*
+  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event_name == 'pull_request' && github.ref
+    || github.sha }}
+  cancel-in-progress: true
 
 jobs:
-  build:
-    runs-on: ubuntu-latest
-
+  lint-cpp:
+    runs-on:
+      - self-hosted
+      - "1ES.Pool=onnxruntime-genai-Ubuntu2204-AMD-CPU"
     steps:
-      - uses: actions/checkout@v2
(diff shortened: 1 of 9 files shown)


baijumeswani (approved):

.github/workflows/clang-format-lint.yml
```diff
@@ -1,22 +1,34 @@
...
-      - uses: actions/setup-python@v5
+      - name: Checkout code
+        uses: actions/checkout@v4
         with:
-          python-version: '3.11.x'
-          architecture: 'x64'
-      - uses: DoozyX/clang-format-lint-action@v0.17
+          submodules: 'false'
+
+      - name: Check format
```
justinchuby, line 30: Should be configure lintrunner to support auto format and more linters etc?

justinchuby: Just the same config in ort. Developers can run lintrunner f to format locally, a github bot can suggest fixes as well on the github UI.

snnn: I am trying to change the C++ linter to a customized one for both repos. Now it does not have the ability to suggest fixes yet. It can be run locally. I plan to add more checker there, such as enabling clang-tidy, check if any C/C++ has an execution bit, etc.  And we will continue using lintrunner for python code. Is it ok?

justinchuby: Sounds good to me! Question: will the clang-format component be the same? Maybe we can retain clang-format so that the local auto format experience is kept?

snnn: They use the same config file. So it should be the same. However, the one used in ORT's CI pipeline was fetched from pypi.org, and the package didn't seem to be an official package from LLVM.  https://pypi.org/project/clang-format/
And, we want to reduce the usages of 3rd-party Github Actions. That's the main reason why I am doing these changes.

snnn: It's ok to use the unofficial binaries locally, but our build pipelines need to hold a higher security bar.

justinchuby: pytorch uses this https://github.com/pytorch/pytorch/blob/ff075d08150ec060f035b660789ebf1de8deb75b/.lintrunner.toml#L100-L105 to fetch a prebuilt, controlled binary. I wonder if we can do the same?

PR #1429, opened 2025-04-29
GQA bf16 support

nenad1002:


nenad1002:
GQA + Rotary support is added here, merging automatically

PR #1434, opened 2025-04-30
make WebGPU name consistent

guschmue:


baijumeswani (approved):

src/generators.cpp
```diff
@@ -207,7 +207,7 @@ std::string to_string(DeviceType device_type) {
     case DeviceType::DML:
       return "DirectML";
     case DeviceType::WEBGPU:
-      return "WebGpu";
+      return "WebGPU";
```
RyanUnderhill, line 210: Amusing, these ones are just for an error message and the 'GetDeviceType' method on the OgaModel. The 'QnnWithSharedMemory' one implies these are the memory type vs the device type. None are consistent here really.

RyanUnderhill, line 210: Amusing, these ones are just for an error message and the 'GetDeviceType' method on the OgaModel. The 'QnnWithSharedMemory' one implies these are the memory type vs the device type. None are consistent here really.

RyanUnderhill (approved):

baijumeswani (approved):

PR #1441, opened 2025-05-02
Add Gemma3 bf16 related changes

nenad1002:
The Gemma3 model must compute layer norms and residual connections in FP32 to ensure accurate logits. Additionally, all logits are now explicitly cast to FP32. This PR removes the incorrect code paths and refactors the builder script so that it cleanly supports both Gemma and non-Gemma models.


nenad1002:
Automatically merging— the final branch will be reviewed once all multimodal changes have been incorporated.

PR #1442, opened 2025-05-02
Set `add_special_tokens` to false by default in Encode

sayanshaw24:
Sets `add_special_tokens` from `OrtxTokenizeWithOptions` added in https://github.com/microsoft/onnxruntime-extensions/pull/940 to false to solve chat template issue in GenAI with extra BOS tokens.

See https://github.com/huggingface/transformers/issues/37686 for more context.

### DIFF
--- a/src/models/model.cpp
+++ b/src/models/model.cpp
@@ -198,7 +198,7 @@ std::unique_ptr<TokenizerStream> Tokenizer::CreateStream() const {
 
 std::vector<int32_t> Tokenizer::Encode(const char* text) const {
   OrtxPtr<OrtxTokenId2DArray> ids;
-  CheckResult(OrtxTokenizeWithOptions(tokenizer_, &text, 1, ids.Address(), false));
+  CheckResult(OrtxTokenizeWithOptions(tokenizer_, &text, 1, ids.Address(), false /* add_special_tokens */));
 
   const extTokenId_t* tokens;
   size_t count;
--- a/src/models/model.cpp
+++ b/src/models/model.cpp
@@ -206,6 +206,16 @@ std::vector<int32_t> Tokenizer::Encode(const char* text) const {
   return {tokens, tokens + count};
 }
 
+std::vector<int32_t> Tokenizer::EncodeWithOptions(const char* text, bool add_special_tokens) const {
+  OrtxPtr<OrtxTokenId2DArray> ids;
+  CheckResult(OrtxTokenizeWithOptions(tokenizer_, &text, 1, ids.Address(), add_special_tokens));
+
+  const extTokenId_t* tokens;
+  size_t count;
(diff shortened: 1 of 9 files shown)


src/ort_genai_c.h
```diff
@@ -461,6 +461,13 @@ OGA_EXPORT void OGA_API_CALL OgaDestroyMultiModalProcessor(OgaMultiModalProcesso
  */
 OGA_EXPORT OgaResult* OGA_API_CALL OgaTokenizerEncode(const OgaTokenizer*, const char* str, OgaSequences* sequences);
 
+/**
+ * Encodes a single string and adds the encoded sequence of tokens to the OgaSequences, with options such as whether to add or omit special tokens.
+ * The OgaSequences must be freed with OgaDestroySequences when it is no longer needed.
+ * Note: this method may be used to add more options in the future, to leverage OrtxTokenizeWithOptions from ORT Extensions.
+ */
+OGA_EXPORT OgaResult* OGA_API_CALL OgaTokenizerEncodeWithOptions(const OgaTokenizer*, const char* str, OgaSequences* sequences, bool add_special_tokens);
```
RyanUnderhill, line 469: The API is 'WithOptions' but it's just one single hardcoded option? Will we add more in the future?

RyanUnderhill:
Ok, meeting conclusion is that we don't need this API currently as our internal default values will do what users want. This way we avoid exposing an option that nobody knows what value to set to.

src/models/model.cpp
```diff
@@ -198,7 +198,7 @@ std::unique_ptr<TokenizerStream> Tokenizer::CreateStream() const {
 
 std::vector<int32_t> Tokenizer::Encode(const char* text) const {
   OrtxPtr<OrtxTokenId2DArray> ids;
-  CheckResult(OrtxTokenize(tokenizer_, &text, 1, ids.Address()));
+  CheckResult(OrtxTokenizeWithOptions(tokenizer_, &text, 1, ids.Address(), false));
```
RyanUnderhill, line 201:
```suggestion
  CheckResult(OrtxTokenizeWithOptions(tokenizer_, &text, 1, ids.Address(), false /* add_special_tokens */));
```
(label the bool for readability, and double check my guess here)

sayanshaw24: good guess

sayanshaw24:
> Not sure if you want to say in the PR comments why we default to false, just to have some history for it if we forget why we did this in the future.

`add_special_tokens` from `OrtxTokenizeWithOptions` is a tokenizer param we set to false in Encode in GenAI so as to not confuse the user deciding what to set it to - setting it to false omits the extra BOS token added in the case of Gemma-3. See https://github.com/huggingface/transformers/issues/37686 for more context.

PR #1444, opened 2025-05-03
Support k_quant in model builder

jiafatom:


src/python/py/models/builder.py
```diff
@@ -292,6 +292,7 @@ def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
                 "is_symmetric": extra_options.get("int4_is_symmetric", True),
                 "op_types_to_quantize": extra_options.get("int4_op_types_to_quantize", ("MatMul", )),
                 "nodes_to_exclude": extra_options.get("int4_nodes_to_exclude", []),
+                "quant_method": extra_options.get("quant_method", "rtn"),
```
kunal-vaishnavi, line 295: The keys in `self.quant_attrs["int4"]` should match the same names that the constructor for `MatMulNBitsQuantizer` accepts. The keys in `extra_options` are prefixed with `int4` to clarify that they are specifically used for int4 quantization.

```suggestion
                "algo_config": self.make_int4_algo_config(extra_options.get("int4_algo_config", "default")) ,
```

If/when quantization is supported for `n` bits, the `int4` prefixes in the `extra_options` can be removed accordingly.

src/python/py/models/builder.py
```diff
@@ -497,6 +498,39 @@ def save_model(self, out_dir):
         )
 
     def to_int4(self, model):
+        quant_method = self.quant_attrs["int4"]["quant_method"]
```
kunal-vaishnavi, line 501: This is changing the default behavior for the `algo_config`. When `algo_config` is not specified in `MatMulNBitsQuantizer`, the default `algo_config` is [weight-only](https://github.com/microsoft/onnxruntime/blob/6fa8ba107f461d72073ee4842830a9d0a96d2546/onnxruntime/python/tools/quantization/matmul_nbits_quantizer.py#L1185-L1194). There should be three conditions: `RTN`, `k_quant` (mixed or last), and `default`.

Let's also move this logic to create the `algo_config` into its own method.

```python
def make_int4_algo_config(self, quant_method):
    # Your changes here
    ...
```

This will allow us to override this method on a per-model basis if it is deemed necessary.

src/python/py/models/builder.py
```diff
@@ -3542,6 +3577,10 @@ def get_args():
                 enable_cuda_graph = Enable CUDA graph capture during inference. Default is false.
                     If enabled, all nodes being placed on the CUDA EP is the prerequisite for the CUDA graph to be used correctly.
                     It is not guaranteed that CUDA graph be enabled as it depends on the model and the graph structure.
+                quant_method = Method for int4 quantization. Default is 'rtn'.
```
kunal-vaishnavi, line 3580: After renaming, let's move this to be after `int4_nodes_to_exclude` so that all the `int4_*` options are grouped together.

src/python/py/models/builder.py
```diff
@@ -497,6 +498,39 @@ def save_model(self, out_dir):
...
+            int4_algo_config = RTNWeightOnlyQuantConfig()
+        elif quant_method in ["k_quant_mixed", "k_quant_last"]:
+            from onnxruntime.quantization.matmul_nbits_quantizer import KQuantWeightOnlyQuantConfig
+            if quant_method == "k_quant_mixed":
+                node_names = [node.name for node in model.graph.node]
+
+                import re
+
+                pattern = r"/model/layers\.(\d{1,2})/mlp/down_proj/MatMul"
+                layers = set()
```
kunal-vaishnavi, line 513: Let's add some comments here to explain the mixed precision strategy (e.g. why we have the calculation for the layers to exclude, why the QKV proj/down proj/LM head MatMuls are quantized to int8, etc).

src/python/py/models/builder.py
jiafatom: our `algo_config` (k-quant-mixed) is dependent on the model. It needs to know how many layers the model has.

src/python/py/models/builder.py
jiafatom: `algo_config` is the actual RTNWeightOnlyQuantConfig that generated from `quant_method`. The users are providing `quant_method` and we converted it into the actual Config.

src/python/py/models/builder.py
```diff
@@ -3519,6 +3548,10 @@ def get_args():
                 int4_nodes_to_exclude = Specify nodes to exclude from int4 quantization. 
                     Use this option when you want to exclude certain nodes from being quantized.
                     Separate the node names with a ',' when passing them here (e.g. int4_nodes_to_exclude=/lm_head/MatMul,/model/embed_tokens/Gather)
+                int4_algo_config = Method for int4 quantization. Default is 'default'.
+                    Currently supported options are: 'default', 'rtn', 'k_quant_mixed', 'k_quant_last'.
+                    k_quant_mixed is k_quant algorithm with mixed precision.
```
kunal-vaishnavi, line 3553:
```suggestion
                    k_quant_mixed = k_quant algorithm with mixed precision.
```

src/python/py/models/builder.py
```diff
@@ -3519,6 +3548,10 @@ def get_args():
...
+                    k_quant_last is k_quant algorithm where only the last MatMul /lm_head/MatMul is quantized as int8, other MatMuls are quantized as int4.
```
kunal-vaishnavi, line 3554:
```suggestion
                    k_quant_last = k_quant algorithm where only the last MatMul (/lm_head/MatMul) is quantized as int8. Other MatMuls are quantized as int4.
```

src/python/py/models/builder.py
kunal-vaishnavi, line 3553:
```suggestion
                    k_quant_mixed = k_quant algorithm with mixed precision (int4 + int8).
```

kunal-vaishnavi (approved):

kunal-vaishnavi (approved):

PR #1447, opened 2025-05-04
Add bfloat16 support in model builder

kunal-vaishnavi:
### Description

This PR adds [bfloat16](https://en.wikipedia.org/wiki/Bfloat16_floating-point_format) support in the model builder.

### Motivation and Context

Most SLMs and LLMs are trained in bfloat16 precision. Casting from bfloat16 to [float16](https://en.wikipedia.org/wiki/Half-precision_floating-point_format) can cause accuracy loss in models (e.g. Google's Gemma model family).

The NumPy dependency when converting a [torch.Tensor](https://pytorch.org/docs/stable/tensors.html) object to an [ONNX TensorProto](https://onnx.ai/onnx/api/helper.html#onnx.helper.make_tensor) object has been removed. This will allow torch.Tensor objects in [other precisions that are not supported in NumPy](https://numpy.org/doc/stable/user/basics.types.html#relationship-between-numpy-data-types-and-c-data-types) to be converted to ONNX TensorProto objects.

This PR also fixes [this issue](https://github.com/microsoft/onnxruntime-genai/discussions/691).


src/python/py/models/builder.py
```diff
@@ -509,22 +523,36 @@ def order_repeated_field(self, repeated_proto, key_name, order):
...
-        tensor.name = name
+    def make_tensor_proto_from_tensor(self, tensor, name):
+        raw_data = bytes(
+            (ctypes.c_ubyte * tensor.element_size() * tensor.numel()).from_address(
+                tensor.data_ptr()
+            )
+        )
+        tensor_proto = helper.make_tensor(
+            name=name,
+            data_type=torch.onnx._type_utils.JitScalarType.from_dtype(tensor.dtype).onnx_type(),
```
justinchuby, line 534: note: avoid using _type_utils: it is a private module that is scheduled for removal. Instead, consider copying the mapping over. (you will only need to do this temporarily because we have native type conversion support with the IR PR. But pointing it out here just in case so that we don't use an api here that doesn't have any guarantees.)

src/python/py/models/builder.py
```diff
@@ -509,22 +523,36 @@ def order_repeated_field(self, repeated_proto, key_name, order):
         order = list(order)
         repeated_proto.sort(key=lambda x: order.index(getattr(x, key_name)))
 
-    def make_external_tensor(self, np_data, name, unpack_int4=False, **kwargs):
-        tensor = numpy_helper.from_array(np_data)
-        tensor.name = name
+    def make_tensor_proto_from_tensor(self, tensor, name):
+        raw_data = bytes(
+            (ctypes.c_ubyte * tensor.element_size() * tensor.numel()).from_address(
+                tensor.data_ptr()
```
justinchuby, line 529: Do we need to ensure the tensor is on cpu?

src/python/py/models/builder.py
```diff
@@ -509,22 +523,36 @@ def order_repeated_field(self, repeated_proto, key_name, order):
         order = list(order)
         repeated_proto.sort(key=lambda x: order.index(getattr(x, key_name)))
 
-    def make_external_tensor(self, np_data, name, unpack_int4=False, **kwargs):
-        tensor = numpy_helper.from_array(np_data)
-        tensor.name = name
+    def make_tensor_proto_from_tensor(self, tensor, name):
+        raw_data = bytes(
```
justinchuby, line 527: I would also add an assertion on tensor to make sure they are always contiguous (`assert tensor.is_contiguous`) If not call contiguous on them. This way you can avoid having to call contiguous() at the call site.

justinchuby: Ref: https://github.com/microsoft/onnxscript/blob/5283d9de6e932ada11f5358e289d1e6ea06e9436/onnxscript/ir/tensor_adapters.py#L109

baijumeswani, line 527: `bytes(...)` makes copy. And I am guessing helper.make_tensor makes another copy of the provided bytes? Hopefully not. Anyway, I don't think there is any good alternative as of now given this code is all written in Python.

But good work trying to avoid making copies so far.

src/python/py/models/builder.py
```diff
@@ -3321,6 +3307,20 @@ def parse_hf_token(hf_token):
     return hf_token
 
 
+def set_io_dtype(precision, execution_provider, extra_options):
```
baijumeswani, line 3310: nit

```suggestion
def set_io_dtype(precision, execution_provider: str, extra_options: dict[str, str]):
```

baijumeswani (dismissed):

src/python/py/models/builder.py
kunal-vaishnavi: Added a mapping from torch.dtype to TensorProto.dtype

src/python/py/models/builder.py
kunal-vaishnavi: Fixed

src/python/py/models/builder.py
kunal-vaishnavi: Added

src/python/py/models/builder.py
```diff
@@ -509,22 +523,36 @@ def order_repeated_field(self, repeated_proto, key_name, order):
         order = list(order)
         repeated_proto.sort(key=lambda x: order.index(getattr(x, key_name)))
 
-    def make_external_tensor(self, np_data, name, unpack_int4=False, **kwargs):
-        tensor = numpy_helper.from_array(np_data)
-        tensor.name = name
+    def make_tensor_proto_from_tensor(self, tensor, name):
```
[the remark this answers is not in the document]
kunal-vaishnavi: This will be updated in the ONNX IR PR.

src/python/py/models/builder.py
kunal-vaishnavi: I omitted type hints for methods in the model builder since overloading is quite common. For example, not all values in the `extra_options` dictionary are strings.

RyanUnderhill (approved):

RyanUnderhill (approved):

PR #1454, opened 2025-05-06
Persist provider options across ClearProviders, AppendProvider where possible

baijumeswani:
Currently, most applications will ClearProviders and AppendProvider to change the execution provider at runtime to the desired EP. While doing this, the provider options listed in the genai config.json are lost.

Some models, require a provider options to run correctly. So, doing `ClearProviders` and then doing `AppendProvider` does not work since the options are lost in the process.

This pull-request will enable persisting the provider options through a backup in case user tries to add a provider already mentioned in the genai_config.json.

### DIFF
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -54,7 +54,9 @@ struct ProviderOptionsObject_Element : JSON::Element {
 };
 
 struct ProviderOptionsArray_Element : JSON::Element {
-  explicit ProviderOptionsArray_Element(std::vector<Config::ProviderOptions>& v) : v_{v} {}
+  explicit ProviderOptionsArray_Element(std::vector<Config::ProviderOptions>& v,
+                                        std::vector<Config::ProviderOptions>* v_backup = nullptr)
+      : v_{v}, v_backup_{v_backup} {}
 
   JSON::Element& OnObject(std::string_view name) override { return object_; }
 
@@ -69,10 +71,18 @@ struct ProviderOptionsArray_Element : JSON::Element {
         v.name = "DML";
       }
     }
+
+    // Copy the provider options to the backup vector.
+    // The backup is used to restore the original provider options when
+    // AppendProvider is called with the same provider name.
+    if (v_backup_) {
+      *v_backup_ = v_;
+    }
   }
 
  private:
   std::vector<Config::ProviderOptions>& v_;
+  std::vector<Config::ProviderOptions>* v_backup_;
   ProviderOptionsObject_Element object_{v_};
 };
 
@@ -143,7 +153,7 @@ struct SessionOptions_Element : JSON::Element {
 
  private:
   Config::SessionOptions& v_;
-  ProviderOptionsArray_Element provider_options_{v_.provider_options};
+  ProviderOptionsArray_Element provider_options_{v_.provider_options, &v_.provider_options_backup};
   NamedStrings_Element config_entries_{v_.config_entries};
 };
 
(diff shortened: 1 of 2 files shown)


nenad1002 (dismissed):

src/config.h
```diff
@@ -74,6 +74,7 @@ struct Config {
     std::vector<NamedString> config_entries;  // Entries go into OrtSessionOptions::AddConfigEntry
 
     std::vector<ProviderOptions> provider_options;
+    std::vector<ProviderOptions> provider_options_backup;
```
RyanUnderhill, line 77: Do we have other config struct members that aren't part of the json? Currently the struct directly maps to the json.

RyanUnderhill, line 77: I think my main confusion is calling it 'provider_options_backup' it's not really a backup, it's really the provider options. And we've now separated the options from which provider is currently in use.

Is there a way to not duplicate the options but instead add a new field 'providers' that's just an array of names in provider priority order? We can generate this on load with existing configs that omit it based on the provider options they specify.

baijumeswani: Sounds good. I'll make this update.

baijumeswani: I added a new vector<string> that handles the providers.

src/config.h
```diff
@@ -74,6 +74,7 @@ struct Config {
     std::vector<NamedString> config_entries;  // Entries go into OrtSessionOptions::AddConfigEntry
 
     std::vector<ProviderOptions> provider_options;
+    std::vector<std::string> providers;
```
RyanUnderhill, line 77:
```suggestion
    std::vector<std::string> providers;  // List of providers to use at runtime, not persisted in the json currently
```
This comment makes my current understanding more explicit

src/config.cpp
```diff
@@ -68,11 +69,18 @@ struct ProviderOptionsArray_Element : JSON::Element {
       } else if (v.name == "dml") {
         v.name = "DML";
       }
+
+      if (std::find(providers_.begin(), providers_.end(), v.name) == providers_.end()) {
```
RyanUnderhill, line 73: Should we do this here? We could do it in the OnComplete() of the parent object and keep this object simple. Or could do it in the LoadConfig(). I don't think we can do it as late as CreateSessionOptionsFromConfig() as an empty providers list should mean to use cpu.

src/config.cpp
```diff
@@ -711,7 +719,7 @@ void SetSearchBool(Config::Search& search, std::string_view name, bool value) {
 }
 
 void ClearProviders(Config& config) {
-  config.model.decoder.session_options.provider_options.clear();
+  config.model.decoder.session_options.providers.clear();
```
RyanUnderhill, line 722: This is nice

src/models/model.cpp
```diff
@@ -274,12 +274,21 @@ int32_t Tokenizer::TokenToTokenId(const char* token) const {
...
                                            bool is_primary_session_options,
                                            bool disable_graph_capture) {
   DeviceInterface* p_device{};
 
-  for (auto& provider_options : provider_options_list) {
+  for (auto& provider : providers) {
+    auto provider_options_it = std::find_if(provider_options_list.begin(), provider_options_list.end(),
+                                            [&provider](const Config::ProviderOptions& po) { return po.name == provider; });
+
+    if (provider_options_it == provider_options_list.end()) {
```
RyanUnderhill, line 287: Wouldn't a missing provider_options just mean to use defaults for that provider?

baijumeswani: Any provider added to the providers list must also exist in the provider options since that was `SetProviderOption` does. We wouldn't expect there to be a provider string in providers that isn't in the provider options.

RyanUnderhill: Ah, I would, as previously the provider_options was the list of providers + options. But now there's two lists. So you could have a provider with only default options. I don't think it affects the outside world currently, as our API always goes through `SetProviderOption` like you said. So up to you.

baijumeswani: Will leave it as is for now. It won't ever hit this code (I could change to an assert) since the providers and provider options should always be in sync (i.e. provider options would always be a superset of providers).

We can update this code when we decide to change this assumption.

src/config.cpp
baijumeswani: Discussed offline that there are multiple places where this function is called. Doing it here will avoid needing to add the same code in multiple places. We can change this in the future where we avoid keeping the providers and the provider options in sync.

RyanUnderhill (approved):

PR #1461, opened 2025-05-08
Postprocessing to share lm_head weights to embedding

jiafatom:


src/python/py/models/postprocess_lm_head_tied_embeddings.py
```diff
@@ -0,0 +1,313 @@
...
+    )
+    
+    # Multiply by scale
+    dequantized_output = "dequantized_embeddings"
+    mul_node = helper.make_node(
+        'Mul',
+        inputs=["centered_weights", "gathered_scales"],
+        outputs=[dequantized_output],
+        name='/model/embed_tokens/MultiplyByScale'
+    )
```
tianleiwu, line 109: Can we use DequantizeLinear op?

src/python/py/models/postprocess_lm_head_tied_embeddings.py
```diff
@@ -0,0 +1,313 @@
+# -------------------------------------------------------------------------
```
kunal-vaishnavi, line 1: I don't think we need a side script if we are integrating the tied embeddings into the model builder. We should be able to do this in the model builder directly.

src/python/py/models/postprocess_lm_head_tied_embeddings.py
```diff
@@ -0,0 +1,313 @@
...
+
+    return None
+
+
+def convert_gather_to_use_lm_head_weights(model_path, output_path, load_external_data=True):
+    # Load the ONNX model
+    print(f"Loading model from {model_path}...")
+    model_name = "model.onnx"
+    model = onnx.load(model_path + model_name, load_external_data=False)
+    if load_external_data:
```
tianleiwu, line 256: Shall we just use `load_external_data=True` in onnx.load? 
Why need we have an option not loading external data.

src/python/py/models/postprocess_lm_head_tied_embeddings.py
jiafatom: This is a temp solution here. Once GatherBlockQuantized bits=8 is supported, we would put the new postprocessing into model builder.

src/python/py/models/postprocess_lm_head_tied_embeddings.py
jiafatom: Sure

src/python/py/models/postprocess_lm_head_tied_embeddings.py
```diff
@@ -0,0 +1,313 @@
...
+    cast_weights_node = helper.make_node(
+        'Cast',
+        inputs=[gathered_quant_weights],
+        outputs=["casted_quant_weights"],
+        name='/model/embed_tokens/CastWeights',
+        to=scale_value_type
+    )
+    
+    # Create a constant for the zero point (128 for symmetric quantization)
+    zero_point_const = np.array([128], dtype=np.uint8)
```
tianleiwu, line 80: This has assumption of 8 bits quantization. Add some comments.

src/python/py/models/postprocess_lm_head_tied_embeddings.py
kunal-vaishnavi: It should still be in the model builder directly even as a temporary solution. That way, it is still a one-line command to produce a model with tied embeddings.

jiafatom: We already have some model converted, then we need this script to postprocessing. In this case, we don't need run the whole model builder again.

src/python/py/models/postprocess_lm_head_tied_embeddings.py
kunal-vaishnavi: I agree that we should use DequantizeLinear. It is already constructed in the model builder.

https://github.com/microsoft/onnxruntime-genai/blob/8eed73024ca383ad9e5a4cacefe98b3cedfc54d6/src/python/py/models/builder.py#L861-L886

It will also be easier to construct the temporary subgraph for GatherBlockQuantized in the model builder directly.

src/python/py/models/postprocess_lm_head_tied_embeddings.py
tianleiwu: I agree. May also add an option to use GatherBlockQuantized  or DequantizeLinear+Gather for word embedding in model builder.

kunal-vaishnavi: We can already run this script offline for those models since we have it. The ideal solution is to have the tied embeddings in the model from the beginning. It also doesn't take too long to run the model builder to produce the updated ONNX model.

Issue #1412, opened 2025-04-21
phi3.5-mini-instruct: Multi-prompt chat sequences seem sensitive to order of EOS tokens listen in genai_config.json


RyanMetcalfeInt8:
**Describe the bug**
For phi3.4-mini-instruct model (presumably other models as well), the generated responses for a multi-prompt chat sequence appear to be sensitive to the order of EOS tokens listed within the `genai_config.json`

**To Reproduce**
1. Prepare Phi-3.5 mini model, as per the instruction here: https://github.com/microsoft/onnxruntime-genai/tree/main/examples/c#phi-35-mini
2. Run phi3.exe sample, with first prompt as `what color are apples?`, and second prompt as `what about grapes?`

You'll find that for the second prompt, there is a 'good' response, in that it realizes you are asking what the color of grapes are (based on context around the first prompt / response).

This is what I get:
```
>phi3.exe Phi-3.5-mini-instruct-onnx\cpu_and_mobile\cpu-int4-awq-block-128-acc-level-4 cpu
-------------
Hello, Phi-3!
-------------
C++ API
Creating config...
Creating model...
Creating tokenizer...
Prompt: (Use quit() to exit) Or (To terminate current output generation, press Ctrl+C)
what color are apples?
Generating response...
 Apples come in a variety of colors, depending on the species and cultivar. Here are some common apple colors you might encounter:

1. Red: This is the most iconic apple color, often associated with apples in general. Red apples can range from a light pinkish-red to a deep, vibrant red. Examples include Red Delicious, Gala, and Fuji apples.

2. Yellow: Yellow apples are less common than red ones but still quite popular. They can be found in varieties like Golden Delicious, Golden Harvest, and Yellow Transparent.

3. Green: Green apples are often tart and crisp, with a bright, vibrant green color. Examples include Granny Smith and Fuji apples.

4. Pink: Some apple varieties have a pinkish hue, such as the Pink Lady (also known as Envy) and the Pink Pearl.

5. Purple: While not as common, there are purple apple varieties like the Purple Haze.

6. Multicolored: Some apples have a mix of colors, such as the Rainbow Crisp, which has streaks of red, yellow, and green.

It's important to note that the color of an apple can also be influenced by factors like ripeness and growing conditions. For example, a green apple might appear more yellowish when fully ripe.

[...]

PR #1463, opened 2025-05-08
Fix array eos_token_id handling

RyanUnderhill:
The previous approach to tweak the logit values would change the behavior of continuous decoding models by modifying which eos token id was chosen.

This change turns the EOS token id into an array in all places and preserves which EOS token was chosen in the output.

Fixes Issue #1412

### DIFF
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -31,6 +31,18 @@ struct NamedStrings_Element : JSON::Element {
   std::vector<Config::NamedString>& v_;
 };
 
+struct Int_Array_Element : JSON::Element {
+  explicit Int_Array_Element(std::vector<int>& v) : v_{v} {}
+
+  void OnValue(std::string_view name, JSON::Value value) override {
+    v_.emplace_back(static_cast<int>(JSON::Get<double>(value)));
+  }
+
+ private:
+  std::vector<int>& v_;
+};
+
+
 struct ProviderOptionsObject_Element : JSON::Element {
   explicit ProviderOptionsObject_Element(std::vector<Config::ProviderOptions>& v) : v_{v} {}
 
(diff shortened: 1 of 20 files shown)


src/cuda/search_cuda.cpp
```diff
@@ -155,7 +159,8 @@ void GreedySearch_Cuda::SampleTopKTopP(int k, float p, float temperature) {
   // Check for EOS
   assert(next_tokens_.size() == eos_meet_.size());
   // Don't replace EOS with pad for batch_size == 1 for continuous decoding mode
-  cuda::Launch_CheckForEOSAndPad(next_tokens_.data(), static_cast<int>(next_tokens_.size()), eos_meet_.data(), params_->config.model.eos_token_id, params_->search.batch_size > 1 ? params_->config.model.pad_token_id : params_->config.model.eos_token_id, done_cpu_.get(), GetStream());
+  // TODO: Ask Alejandro why this is needed for batch_size==1
```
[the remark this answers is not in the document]
RyanUnderhill: Fixed

src/config.cpp
```diff
@@ -602,7 +590,7 @@ struct Model_Element : JSON::Element {
     } else if (name == "pad_token_id") {
       v_.pad_token_id = static_cast<int>(JSON::Get<double>(value));
     } else if (name == "eos_token_id") {
-      v_.eos_token_id = static_cast<int>(JSON::Get<double>(value));
+      v_.eos_token_id.assign(1, static_cast<int>(JSON::Get<double>(value)));
```
kunal-vaishnavi, line 593: Will these changes support all of the following cases for the EOS token id?

1. EOS token id is not provided (set to pad token id and print a warning to the user)
2. EOS token id is one token id
3. EOS token id is a list of token ids

I think your changes will but I just want to double check.

src/cuda/beam_search_scorer_cuda.cpp
```diff
@@ -10,14 +10,14 @@
 
 namespace Generators {
 
-BeamSearchScorer_Cuda::BeamSearchScorer_Cuda(const GeneratorParams& parameters)
-    : stream_{GetStream()} {
+BeamSearchScorer_Cuda::BeamSearchScorer_Cuda(const GeneratorParams& parameters, std::span<int32_t> cuda_eos_tokens)
```
kunal-vaishnavi, line 13: Should we remove the `cuda` prefix?

```suggestion
BeamSearchScorer_Cuda::BeamSearchScorer_Cuda(const GeneratorParams& parameters, std::span<int32_t> eos_tokens)
```

RyanUnderhill: I added that on purpose to make it clear this was cuda memory, as it's not passed as a DeviceSpan.

src/config.cpp
RyanUnderhill: 1. If it's not provided, the array is empty and the Config constructor sets the pad token_id into it (we could log as a warning if it's ever an issue).
2. The code above is for when it's set to a single value, so it just sets the array of eos_token_ids to a single value.
3. The array version is handled a few lines later in 'OnArray'

baijumeswani (approved):

kunal-vaishnavi (approved):

PR #1464, opened 2025-05-08
Avoid adding providers if not requested

baijumeswani:
In a previous PR: #1454 I added support for the providers to persist the provider options across clearprovider->appendprovider. A bug was introduced where the providers always got built to include every provider in the provider options.

This PR fixes it. So, if the user clears the providers , and then appends a provider, only the new provider should be present in the providers list (not all the ones in the provider options).

### DIFF
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -831,7 +831,7 @@ Config::Config(const fs::path& path, std::string_view json_overlay) : config_pat
     search.max_length = model.context_length;
 
   for (const auto& provider_option : model.decoder.session_options.provider_options) {
-    providers.push_back(provider_option.name);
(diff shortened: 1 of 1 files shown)


RyanUnderhill:
So the issue is that the OnComplete() was adding all providers back to the list from the provider options. The fix is to do it higher up to avoid SetProviderOption from doing it again?

baijumeswani:
Yes, exactly. :)

RyanUnderhill:
Hmm, I think moving it up just moves the problem a little and would still be a surprise if we allow dynamic session options. If you move it into the Config constructor it'll happen once at load time.

RyanUnderhill (approved):

baijumeswani:
Thanks for the review. :)

PR #1470, opened 2025-05-10
Address QNN specific regressions

baijumeswani:
Recent changes broke QNN execution provider. Fixes in this PR:

- When appending providers, normalize the provider string
- Whether or not the device supports continuous decoding must depend on the kv cache device (not on the input device)
- Windowed input ids should be only accessible for CPU accessible devices (QNN and CPU), not for others.
- Providers for pipeline models must always be enabled. This is implicit. We do not expect the pipeline models to be controlled using clearproviders and appendprovider.
- The trivial session for QNN must be created with the `enable_htp_shared_memory_allocator` = `1` since the QNN device allocator is only made available when this option is set.
- Once p_device is set, it must not be overridden. After all attempts are exhausted, if p_device is still not set, set it to CPU. Otherwise, it is the first non-CPU device provided either through the primary providers or through the pipeline providers.
- model-qa.py had unset variables.

### DIFF
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -709,7 +709,7 @@ void ClearProviders(Config& config) {
 
 void SetProviderOption(Config& config, std::string_view provider_name, std::string_view option_name, std::string_view option_value) {
   if (auto normalized_provider = NormalizeProviderName(provider_name); !contains(config.model.decoder.session_options.providers, normalized_provider))
-    config.model.decoder.session_options.providers.push_back(normalized_provider);
+    config.model.decoder.session_options.providers.push_back(std::string(normalized_provider));
 
   std::ostringstream json;
   json << R"({")" << provider_name << R"(":{)";
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -9,7 +9,8 @@
 
 namespace Generators {
 
-std::string NormalizeProviderName(std::string_view name) {
+// Fix casing of certain historical names to match current Onnxruntime names
+std::string_view NormalizeProviderName(std::string_view name) {
   if (name == "qnn") {
     return "QNN";
   } else if (name == "webgpu") {
@@ -17,9 +18,8 @@ std::string NormalizeProviderName(std::string_view name) {
   } else if (name == "dml") {
     return "DML";
   }
-  return std::string(name);
+  return name;  // Return name unchanged
 }
-
 ONNXTensorElementDataType TranslateTensorType(std::string_view value) {
   if (value == "float32") {
     return ONNX_TENSOR_ELEMENT_DATA_TYPE_FLOAT;
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -9,6 +9,17 @@
 
 namespace Generators {
 
+std::string NormalizeProviderName(std::string_view name) {
+  if (name == "qnn") {
+    return "QNN";
+  } else if (name == "webgpu") {
+    return "WebGPU";
+  } else if (name == "dml") {
+    return "DML";
+  }
+  return std::string(name);
+}
+
 ONNXTensorElementDataType TranslateTensorType(std::string_view value) {
   if (value == "float32") {
     return ONNX_TENSOR_ELEMENT_DATA_TYPE_FLOAT;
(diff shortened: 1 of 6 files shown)


src/models/input_ids.cpp
```diff
@@ -104,6 +104,11 @@ void DefaultInputIDs::Update(DeviceSpan<int32_t> new_tokens) {
 }
 
 WindowedInputIDs::WindowedInputIDs(State& state) : state_{state} {
+  if (model_.p_device_inputs_->GetType() != DeviceType::QNN &&
+      model_.p_device_inputs_->GetType() != DeviceType::CPU) {
+    throw std::runtime_error("Sliding a window over input_ids requires works with only QNN and CPU providers.");
```
baijumeswani, line 109:
```suggestion
    throw std::runtime_error("Sliding a window over input_ids works with either the QNN or the CPU provider.");
```

src/config.cpp
```diff
@@ -9,6 +9,17 @@
...
+  if (name == "qnn") {
+    return "QNN";
+  } else if (name == "webgpu") {
+    return "WebGPU";
+  } else if (name == "dml") {
+    return "DML";
+  }
+  return std::string(name);
+}
+
```
RyanUnderhill, line 22:
```suggestion
// Fix casing of certain historical names to match current Onnxruntime names
std::string_view NormalizeProviderName(std::string_view name) {
  if (name == "qnn") {
    return "QNN";
  } else if (name == "webgpu") {
    return "WebGPU";
  } else if (name == "dml") {
    return "DML";
  }
  return name;  // Return name unchanged
}
```

Figured might as well avoid allocating a std::string every time since it seems to be unnecessary.

src/models/model.cpp
```diff
@@ -424,6 +433,11 @@ void EnsureDeviceOrtInit(DeviceInterface& device) {
   auto session_options = OrtSessionOptions::Create();
   std::vector<Config::ProviderOptions> provider_options_list;
   provider_options_list.emplace_back(Config::ProviderOptions{device_type_names[static_cast<int>(type)], {}});
+  // QnnHtpShared is a special case. This allocator is only made available when the provider option
+  // 'enable_htp_shared_memory_allocator' is set to 1.
+  if (type == DeviceType::QNN) {
+    provider_options_list.back().options.emplace_back("enable_htp_shared_memory_allocator", "1");
+  }
```
RyanUnderhill, line 440: We were previously checking for this option based on user settings, now it will always be set?

https://github.com/microsoft/onnxruntime-genai/blob/0d27a3b437c43cc8aa9bae7b8d08fda7fc8ad4e2/src/models/model.cpp#L359

RyanUnderhill: Ah, right. I didn't notice it was the code for the trivial session. All good.

RyanUnderhill (approved):

src/models/input_ids.cpp
```diff
@@ -104,6 +104,11 @@ void DefaultInputIDs::Update(DeviceSpan<int32_t> new_tokens) {
 }
 
 WindowedInputIDs::WindowedInputIDs(State& state) : state_{state} {
+  if (model_.p_device_inputs_->GetType() != DeviceType::QNN &&
+      model_.p_device_inputs_->GetType() != DeviceType::CPU) {
+    throw std::runtime_error("Sliding a window over input_ids works with either the QNN or the CPU provider.");
```
edgchen1, line 109:
```suggestion
    throw std::runtime_error("Sliding a window over input_ids only works with either the QNN or the CPU provider.");
```

src/models/model.cpp
```diff
@@ -280,7 +280,16 @@ DeviceInterface* SetProviderSessionOptions(OrtSessionOptions& session_options,
                                            bool disable_graph_capture) {
   DeviceInterface* p_device{};
 
-  for (auto& provider : providers) {
+  auto providers_list = providers;
+  if (!is_primary_session_options) {
+    // Providers specified in a non-primary provider options list are added
+    // to the primary providers. They are considered immutable and implcitly
```
edgchen1, line 286:
```suggestion
    // to the primary providers. They are considered immutable and implicitly
```

baijumeswani:
Thank you for the review. :)

src/models/model.cpp
```diff
@@ -622,13 +636,13 @@ void Model::CreateSessionOptionsFromConfig(const Config::SessionOptions& config_
...
-                                        config_session_options.provider_options, is_primary_session_options,
-                                        disable_graph_capture);
+  auto session_device = SetProviderSessionOptions(session_options, config_session_options.providers,
+                                                  config_session_options.provider_options, is_primary_session_options,
+                                                  disable_graph_capture);
 
-  // Fallback to CPU if no provider specific interface was set
-  if (!p_device_)
-    p_device_ = GetDeviceInterface(DeviceType::CPU);
+  if (!p_device_) {
```
edgchen1, line 643: would it be worth verifying that if `session_device != nullptr && p_device_ != nullptr`, then `session_device->GetType() == p_device_->GetType()`?

PR #1471, opened 2025-05-10
Logging SetLogCallback + Debugging cleanup

RyanUnderhill:
Remove duplicate SizeOf in utils.cpp in favor of existing Ort::SizeOf. These were exact duplicates, I just forgot to remove one in the past somehow.
Add SetLogCallback to enable runtime arbitrary log redirection Add Python bindings for og.set_log_callback
Add python test for og.set_log_callback


src/logging.cpp
```diff
@@ -12,6 +12,35 @@ namespace Generators {
...
+  };
+
+  CustomBuffer m_buffer;
+} gp_callback_stream;
+
+void SetLogStream() {
+  if (gp_callback)
+    gp_stream = &gp_callback_stream;
+  else if (gp_logfile)
+    gp_stream = gp_logfile.get();
```
edgchen1, line 40: what if someone sets both a log file and a callback? I think it would make sense to have the last one set win, log to both, or be an error. I guess this precedence is fine, but it should be documented.

src/models/utils.cpp
```diff
@@ -7,39 +7,39 @@ namespace Generators {
...
+      return "float64";
     case Ort::TypeToTensorType<Ort::Float16_t>:
-      return sizeof(Ort::Float16_t);
+      return "float16";
     case Ort::TypeToTensorType<Ort::BFloat16_t>:
-      return sizeof(Ort::BFloat16_t);
+      return "bfloat16";
     default:
-      throw std::runtime_error("Unsupported ONNXTensorElementDataType in GetTypeSize");
+      return "(unsupported type, please add)";
```
edgchen1, line 42: would it be better to error out if the type is not supported yet, so it is more likely to be noticed and fixed?

src/ort_genai_c.h
```diff
@@ -97,6 +97,12 @@ OGA_EXPORT const char* OGA_API_CALL OgaResultGetError(const OgaResult* result);
 OGA_EXPORT OgaResult* OGA_API_CALL OgaSetLogBool(const char* name, bool value);
 OGA_EXPORT OgaResult* OGA_API_CALL OgaSetLogString(const char* name, const char* value);
 
+/**
+ * \param[in] callback function pointer to the logging callback function (use nullptr to disable callback)
+ * \return OgaResult containing the error message when the callback could not be set, else nullptr.
+ */
+OGA_EXPORT OgaResult* OGA_API_CALL OgaSetLogCallback(void (*callback)(const char* string, size_t length));
```
edgchen1, line 104: if we are exposing a way to set an onnxruntime-genai log output function, would it make sense to also allow users to do this for the underlying onnxruntime logging? something to consider, not necessarily for this PR.

src/models/utils.cpp
RyanUnderhill: It'll be hit by someone debugging, so I figured it'd be far more annoying to be unable to debug than to see an error message in the debug logs. In both cases someone will be looking at the results, and I saw no reason to make it fatal.

src/logging.cpp
RyanUnderhill: Good point. There's no way to set a non filename log, it's only reset by setting an empty filename. So it makes sense to work through priorities. Where a filename takes precedence over cerr and a callback takes priority over a filename. I'll update the documentation to make it clear.

src/ort_genai_c.h
skottmckay: yes. for foundry local it would be great to use the same logging callback for everything as we can't get the native logs from genai or ort easily.

src/logging.cpp
baijumeswani: Updated the logic so now the most recent logging destination will be selected, and other destinations will be removed/reset.

src/ort_genai_c.h
baijumeswani: I will add this functionality in another pull-request.

src/logging.cpp
```diff
@@ -51,16 +80,23 @@ void SetLogString(std::string_view name, std::string_view value) {
     else {
       fs::path filename{std::string(value)};
       gp_logfile = std::make_unique<std::ofstream>(filename.open_for_write());
+      // If a filename was provided, we log callback will be disabled
```
baijumeswani:
```suggestion
      // If a filename was provided, log callback will be disabled
```

src/logging.cpp
```diff
@@ -51,16 +80,23 @@ void SetLogString(std::string_view name, std::string_view value) {
...
-    else
-      gp_stream = &std::cerr;
+    SetLogStream();
   } else
     throw JSON::unknown_value_error{};
 }
 
+void SetLogCallback(CallbackFn fn) {
+  gp_callback = fn;
+  // If a callback was provided, we disable the file logging
```
baijumeswani:
```suggestion
  // If a callback was provided, file logging will be disabled
```

src/ort_genai_c.h
```diff
@@ -97,12 +97,23 @@ OGA_EXPORT void OGA_API_CALL OgaShutdown();
 OGA_EXPORT const char* OGA_API_CALL OgaResultGetError(const OgaResult* result);
 
 /**
+ * \brief Control the logging behavior of the library.
+ *        If OgaSetLogString is called with name "filename", and value is a valid file path,
+ *        the library will log to that file. This will override any previously set logging destination.
```
edgchen1, line 102: should we also mention the behavior of an empty filename value disabling the file output?

edgchen1 (approved):

src/logging.cpp
```diff
@@ -12,6 +12,35 @@ namespace Generators {
 LogItems g_log;
 static std::ostream* gp_stream{&std::cerr};
 static std::unique_ptr<std::ofstream> gp_logfile;
+static CallbackFn gp_callback{};
+
+// Custom stream that calls gp_callback on every line of output
+struct CallbackStream : std::ostream {
+  CallbackStream() : std::ostream{&m_buffer} {
+    gp_stream = this;
```
edgchen1: I think we don't need to set `gp_stream` here. it just gets overwritten in `SetLogStream()`, right?

src/ort_genai_c.h
```diff
@@ -97,12 +97,25 @@ OGA_EXPORT void OGA_API_CALL OgaShutdown();
...
  * \param[in] name logging option name, see logging.h 'struct LogItems' for the list of available options
  * \param[in] value logging option value.
  */
 OGA_EXPORT OgaResult* OGA_API_CALL OgaSetLogBool(const char* name, bool value);
 OGA_EXPORT OgaResult* OGA_API_CALL OgaSetLogString(const char* name, const char* value);
 
+/**
+ * \brief Register a callback function to receive log messages from the library. If invoked, the callback will override
+ *        the previously set logging destination (e.g. a file or std::cerr).
+ * \param[in] callback function pointer to the logging callback function (use nullptr to disable callback)
```
edgchen1: for completeness, maybe also mention here that specifying `nullptr` will revert to logging to std::cerr

edgchen1 (approved):

edgchen1 (approved):

PR #1476, opened 2025-05-13
Fix how torch tensors are saved

kunal-vaishnavi:
### Description

This PR fixes how tensors are saved by ensuring `is_contiguous` and `cpu` are called as methods on the tensor instead of returning the methods themselves.

### Motivation and Context

This fixes an issue with how the model's weights are saved.

### DIFF
--- a/src/python/py/models/builder.py
+++ b/src/python/py/models/builder.py
@@ -549,12 +549,12 @@ class Model:
         repeated_proto.sort(key=lambda x: order.index(getattr(x, key_name)))
 
     def make_tensor_proto_from_tensor(self, tensor, name):
-        if not tensor.is_contiguous:
-            # Make tensor contiguous
-            tensor = tensor.contiguous()
-        if tensor.get_device != -1:
+        if tensor.get_device() != -1:
             # Tensor must be on CPU (device id = -1)
             tensor = tensor.cpu()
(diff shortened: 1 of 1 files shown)


nenad1002 (approved):

src/python/py/models/builder.py
```diff
@@ -1067,7 +1067,7 @@ def make_layernorm(self, layer_id, layernorm, skip, simple, location):
 
         # Create weight and bias tensors
         weight = f"model.layers.{layer_id}.{location}_layernorm.weight"
-        self.make_external_tensor(layernorm.weight.detach().cpu().to(new_torch_dtype).contiguous() + self.layernorm_attrs["add_offset"], weight)
+        self.make_external_tensor((layernorm.weight.detach().cpu().to(new_torch_dtype) + self.layernorm_attrs["add_offset"]).contiguous(), weight)
```
nenad1002, line 1070: .contigous() not needed after to(new_torch_dtype)?

kunal-vaishnavi: There is a potential add offset to calculate so this will apply `.contiguous()` afterwards to ensure the final calculated tensor is contiguous.

nenad1002 (approved):

PR #1485, opened 2025-05-15
NvTensorRtRtx dependency on CUDA device name removed

BLSharda:
GenAI EP NvTensorRtRtx dependency on using CUDA device removed. This is to align with latest changes in ORT and NV EP.

Also, still requires onnxruntime-genai-cuda.dll but device name is NvTensorRtRtx instead of CUDA. And the memory type is still "Cuda" for ORT API.

### DIFF
--- a/src/cuda/interface.cpp
+++ b/src/cuda/interface.cpp
@@ -68,7 +68,7 @@ struct GpuMemory final : DeviceBuffer {
   bool owned_;  // If we own the memory, we delete it on destruction
 };
 
-struct CudaInterfaceImpl final : DeviceInterface {
+struct CudaInterfaceImpl : DeviceInterface {
   CudaInterfaceImpl() {
     g_stream.Create();
   }
@@ -164,6 +164,10 @@ struct CudaInterfaceImpl final : DeviceInterface {
   }
 };
 
+struct NvTensorRtRtxInterfaceImpl final : CudaInterfaceImpl {
+  DeviceType GetType() const override { return DeviceType::NvTensorRtRtx; }
+};
+
 std::unique_ptr<DeviceInterface> g_cuda_device;
 
 DeviceInterface& GetCudaDeviceInterface() { return *g_cuda_device; }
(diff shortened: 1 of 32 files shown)


src/models/model.cpp
```diff
@@ -445,7 +443,7 @@ void EnsureDeviceOrtInit(DeviceInterface& device) {
   allocator.session_ = OrtSession::Create(GetOrtEnv(), g_trivial_model, sizeof(g_trivial_model), session_options.get());
 
   // Names for the device memory types used by 'OrtMemoryInfo::Create'
-  static const char* device_memory_type_names[] = {"CPU (Not used, see above)", "Cuda", "DML", "WebGPU_Buffer", "QnnHtpShared", "OpenVINO (Not used, see above)"};
+  static const char* device_memory_type_names[] = {"CPU (Not used, see above)", "Cuda", "DML", "WebGPU_Buffer", "QnnHtpShared", "OpenVINO (Not used, see above)", "Cuda"};
```
kunal-vaishnavi, line 446: `Cuda` is already listed as the second entry in this array. This change can be reverted.

src/cuda/interface.cpp
```diff
@@ -9,6 +9,11 @@
 #include "kernels.h"
 #include <cstdarg>
 
+#if defined(_WIN32) || defined(_WIN64)
+#define strcasecmp _stricmp
+#define strncasecmp _strnicmp
```
kunal-vaishnavi, line 14: Can `strncasecmp` be removed? I only see `strcasecmp` used in this file.

BLSharda:
@kunal-vaishnavi and @baijumeswani , can you help review and merge this change?

baijumeswani:
> This is to align with latest changes in ORT and NV EP.

Could you elaborate or share the relevant onnxruntime pull-request so we can track the reason for this chance?

src/cuda/interface.cpp
```diff
@@ -164,6 +166,14 @@ struct CudaInterfaceImpl final : DeviceInterface {
   }
 };
 
+struct CudaInterfaceImpl final : CudaInterfaceImplBase {
+  DeviceType GetType() const override { return DeviceType::CUDA; }
+};
+
+struct NvTensorRtRtxInterfaceImpl final : CudaInterfaceImplBase {
+  DeviceType GetType() const override { return DeviceType::NvTensorRtRtx; }
+};
```
baijumeswani, line 175: Is the only difference between the two the `GetType` return value? Is `DeviceType::NvTensorRtRtx` used anywhere? Should it behave the same way as `DeviceType::CUDA`. If they both should behave the same, then we can avoid making this change.

BLSharda: This change is added to achieve different device name primarily in EnsureDeviceOrtInit that uses device.GetType(). 

static const char* device_type_names[] = {"CPU (Not used, see above)", "cuda", "DML", "WebGPU", "QNN", "OpenVINO (Not used, see above)", **"NvTensorRtRtx"**};

BLSharda:
> > This is to align with latest changes in ORT and NV EP.
> 
> Could you elaborate or share the relevant onnxruntime pull-request so we can track the reason for this chance?

In PR [24656](https://github.com/microsoft/onnxruntime/pull/24656) CUDA dependency is removed (USE_CUDA) for NvTensorRtRtx EP so we need to switch device name from "cuda" to "NvTensorRtRtx".

baijumeswani (approved):

PR #1493, opened 2025-05-19
Enable TRT multi profile option though provider option  

anujj:
- Enable TRT multi profile option though provider option 
- Pass the optimization profile index though the run options 
- Hide both multi profile and cuda graph feature behind the provider option 

Reviewer : @BLSharda @hanbitmyths


src/models/model.cpp
```diff
@@ -35,7 +35,7 @@ State::State(const GeneratorParams& params, const Model& model)
   }
 }
 
-void State::Run(OrtSession& session, bool graph_capture_this_run) {
+void State::Run(OrtSession& session, bool graph_capture_this_run, bool enable_mutiprofile) {
```
kunal-vaishnavi, line 38:
```suggestion
void State::Run(OrtSession& session, bool graph_capture_this_run, bool enable_multi_profile) {
```

src/models/model.cpp
```diff
@@ -45,9 +45,16 @@ void State::Run(OrtSession& session, bool graph_capture_this_run) {
 
   if (first_run_) {
     extra_outputs_.Add(session.GetOutputNames());
+    if(enable_mutiprofile){
```
kunal-vaishnavi, line 48:
```suggestion
    if (enable_multi_profile){
```

src/models/model.cpp
```diff
@@ -369,8 +377,121 @@ DeviceInterface* SetProviderSessionOptions(OrtSessionOptions& session_options,
...
+          const int max_context_len = model_config.context_length;
+          const int opt_context_len = 512;  // Optimal is half of max, capped at 512
+          const int min_seq_len = 1;
+
+          // Extract KV cache name patterns from decoder config
+          std::string past_key_pattern = model_config.decoder.inputs.past_key_names;
+          std::string past_value_pattern = model_config.decoder.inputs.past_value_names;
+
+          // If patterns are empty, use defaults
+          if (past_key_pattern.empty()) past_key_pattern = "past_key_values.%d.key";
```
kunal-vaishnavi, line 403:
```suggestion
          if (past_key_pattern.empty()) past_key_pattern = Config::PastKeyName;
```

src/models/model.cpp
```diff
@@ -369,8 +377,121 @@ DeviceInterface* SetProviderSessionOptions(OrtSessionOptions& session_options,
...
+          if (past_value_pattern.empty()) past_value_pattern = "past_key_values.%d.value";
```
kunal-vaishnavi, line 404:
```suggestion
          if (past_value_pattern.empty()) past_value_pattern = Config::PastValueName;
```

src/models/model.cpp
```diff
@@ -369,8 +377,121 @@ DeviceInterface* SetProviderSessionOptions(OrtSessionOptions& session_options,
...
+
+          std::ostringstream min_shapes, opt_shapes, max_shapes;
+
+          // MIN SHAPES (context phase and first token generation)
+          min_shapes << "input_ids:1x" << min_seq_len << ",attention_mask:1x" << min_seq_len
```
kunal-vaishnavi, line 409: Can we use the config's default string names for all of the strings that are created here?

https://github.com/microsoft/onnxruntime-genai/blob/375f1b61aab14c3931cc732b3c4c1fc5b86eafba/src/config.h#L16-L18

For example:

```suggestion
          min_shapes << Config::InputIdsName << ":1x" << min_seq_len << "," << Config::AttentionMaskName << ":1x" << min_seq_len 
```

src/models/model.cpp
```diff
@@ -369,8 +377,121 @@ DeviceInterface* SetProviderSessionOptions(OrtSessionOptions& session_options,
...
+          if (past_key_pattern.empty()) past_key_pattern = "past_key_values.%d.key";
+          if (past_value_pattern.empty()) past_value_pattern = "past_key_values.%d.value";
+
+          std::ostringstream min_shapes, opt_shapes, max_shapes;
+
+          // MIN SHAPES (context phase and first token generation)
+          min_shapes << "input_ids:1x" << min_seq_len << ",attention_mask:1x" << min_seq_len 
+                   << ",position_ids:1x" << min_seq_len;
+
+          // Add empty KV cache entries for all layers
```
kunal-vaishnavi, line 412: Can you move some of these string constructions into functions? It seems the same logic is repeated multiple times below.

src/models/model.h
```diff
@@ -49,7 +49,7 @@ struct State {
   std::vector<std::pair<std::string, std::string>> ep_dynamic_options_next_run_;
 
  protected:
-  void Run(OrtSession& session, bool graph_capture_this_run = false);  // Uses the inputs below to run
+  void Run(OrtSession& session, bool graph_capture_this_run = false, bool enable_mutiprofile = false);  // Uses the inputs below to run
```
kunal-vaishnavi, line 52:
```suggestion
  void Run(OrtSession& session, bool graph_capture_this_run = false, bool enable_multi_profile = false);  // Uses the inputs below to run
```

src/models/model.cpp
anujj: Done

src/models/model.cpp
anujj: Done

src/models/model.cpp
anujj: Done

src/models/model.h
anujj: Done

anujj:
Thanks for the review @kunal-vaishnavi.  I have addressed all the review comments

src/config.cpp
```diff
@@ -735,6 +735,21 @@ bool IsGraphCaptureEnabled(Config::SessionOptions& session_options) {
...
+{
+  for (const auto& provider_options : session_options.provider_options) {
+    if (provider_options.name == "NvTensorRtRtx") {
+      for (const auto& value : provider_options.options) {
+        if (value.first == "nv_multi_profile_enable" && value.second == "1") {
+          return true;
+        }
+      }
+      return false;
+    }
```
baijumeswani, line 748: Please iterate over the `session_options.providers` instead. See https://github.com/microsoft/onnxruntime-genai/pull/1497

src/models/model.cpp
```diff
@@ -35,7 +35,7 @@ State::State(const GeneratorParams& params, const Model& model)
   }
 }
 
-void State::Run(OrtSession& session, bool graph_capture_this_run) {
+void State::Run(OrtSession& session, bool graph_capture_this_run, bool enable_multi_profile) {
```
baijumeswani, line 38: Can we use `params_->enable_multi_profile` instead of passing it in? We pobably can do the same for use_graph_capure. But I can address the use_graph_capture in another pr so as to avoid <over> complicating this one.

src/models/model.cpp
```diff
@@ -369,8 +377,102 @@ DeviceInterface* SetProviderSessionOptions(OrtSessionOptions& session_options,
...
         session_options.AddConfigEntry("session.intra_op.allow_spinning", "0");
       } else if (provider_options.name == "NvTensorRtRtx") {
-        // After setting the NvTensorRtRtx provider in Onnxruntime, GenAI will then treat it as the cuda device.
+        // Always enable CUDA graph
         session_options.AddConfigEntry("ep.nvtensorrtrtxexecutionprovider.nv_cuda_graph_enable", "1");
+
+        // Check for multi-profile option
+        if (const auto opt_it = std::find_if(provider_options.options.begin(), provider_options.options.end(),
+                                             [](const auto& pair) { return pair.first == "nv_multi_profile_enable"; });
+            opt_it != provider_options.options.end() && opt_it->second == "1") {
```
baijumeswani, line 386: Can you use the utility in config.cpp `IsMultiProfileEnabled`?

Also, can we move the entire logic below into a function of its own?

src/models/model.cpp
```diff
@@ -277,7 +284,8 @@ DeviceInterface* SetProviderSessionOptions(OrtSessionOptions& session_options,
                                            const std::vector<std::string>& providers,
                                            const std::vector<Config::ProviderOptions>& provider_options_list,
                                            bool is_primary_session_options,
-                                           bool disable_graph_capture) {
+                                           bool disable_graph_capture,
+                                           Config::Model& model_config) {
```
baijumeswani, line 288:
```suggestion
                                           const Config::Model& model_config) {
```

Also, shall we pass in the entire Config instead of Config::Model?

src/models/model.cpp
```diff
@@ -369,8 +377,102 @@ DeviceInterface* SetProviderSessionOptions(OrtSessionOptions& session_options,
...
+          const int opt_context_len = model_config.context_length/2; // opt shape is half of max context len
+          const int min_seq_len = 1;
+
+          // Extract KV cache name patterns from decoder config
+          std::string past_key_pattern = model_config.decoder.inputs.past_key_names;
+          std::string past_value_pattern = model_config.decoder.inputs.past_value_names;
+
+          // If patterns are empty, use defaults
+          if (past_key_pattern.empty()) past_key_pattern = Config::Defaults::PastKeyName;
+          if (past_value_pattern.empty()) past_value_pattern = Config::Defaults::PastValueName;
```
baijumeswani, line 404: Why would this be empty? If not specified, the config will use the default values. They will be empty only if the genai_config.json explicitly defines it to be empty. In that case, maybe we should error out in case that is not an acceptable value here.

src/models/model.cpp
```diff
@@ -369,8 +377,102 @@ DeviceInterface* SetProviderSessionOptions(OrtSessionOptions& session_options,
...
+
+          // Helper function to add input shapes (input_ids, attention_mask, position_ids)
+          auto addInputShapes = [](std::ostringstream& shapes, int seq_len, bool append = false) {
```
baijumeswani, line 407:
```suggestion
          const auto add_input_shapes = [](std::ostringstream& shapes, int seq_len, bool append = false) {
```

src/models/model.cpp
```diff
@@ -369,8 +377,102 @@ DeviceInterface* SetProviderSessionOptions(OrtSessionOptions& session_options,
...
+            if (append) shapes << ",";
+            shapes << Config::Defaults::InputIdsName << ":1x" << seq_len << ","
+                   << Config::Defaults::AttentionMaskName << ":1x" << seq_len << ","
+                   << Config::Defaults::PositionIdsName << ":1x" << seq_len;
+          };
+
+          // Helper function to add generation phase input shapes
+          auto addGenerationInputShapes = [](std::ostringstream& shapes, int context_len) {
```
baijumeswani, line 415:
```suggestion
          const auto add_generation_input_shapes = [](std::ostringstream& shapes, int context_len) {
```

src/models/model.cpp
```diff
@@ -369,8 +377,102 @@ DeviceInterface* SetProviderSessionOptions(OrtSessionOptions& session_options,
...
+
+          // Helper function to add KV cache with sequence length
+          auto addKVCacheShapes = [&](std::ostringstream& shapes,
+                                     const std::string& key_pattern,
+                                     const std::string& value_pattern,
+                                     int seq_len) {
+            for (int i = 0; i < num_layers; i++) {
+              char key_name[256], value_name[256];
+              snprintf(key_name, sizeof(key_name), key_pattern.c_str(), i);
+              snprintf(value_name, sizeof(value_name), value_pattern.c_str(), i);
```
baijumeswani, line 443: Can we use `ComposeKeyValueName` instead of duplicating that code here?

src/models/model.cpp
```diff
@@ -369,8 +377,102 @@ DeviceInterface* SetProviderSessionOptions(OrtSessionOptions& session_options,
...
+
+              shapes << "," << key_name << ":1x" << num_kv_heads << "x0x" << head_dim;
+              shapes << "," << value_name << ":1x" << num_kv_heads << "x0x" << head_dim;
+            }
+          };
+
+          // Helper function to add KV cache with sequence length
+          auto addKVCacheShapes = [&](std::ostringstream& shapes,
```
baijumeswani, line 436: same here.
```suggestion
          const auto add_key_value_cache_shapes = [&](std::ostringstream& shapes,
```

src/models/model.cpp
```diff
@@ -400,7 +502,7 @@ static const uint8_t g_trivial_model[] = {
 // the allocator used is not destroyed until last. This keeps the allocator around until exit, after all other memory
 // has been destroyed. Without this, we will crash in the Onnxruntime BFCArena code when deleting tensors due to the
 // arena already being destroyed.
-void EnsureDeviceOrtInit(DeviceInterface& device) {
+void EnsureDeviceOrtInit(DeviceInterface& device, std::unique_ptr<Config>& config) {
```
baijumeswani, line 505: Config& instead of passing in a unique ptr by reference?

anujj:
@baijumeswani : Addressed all the review comments

src/models/model.cpp
```diff
@@ -511,7 +613,7 @@ std::vector<const char*> SessionInfo::GetOutputSymbolicShape(const std::string&
 
 Model::Model(std::unique_ptr<Config> config) : config_{std::move(config)} {
   CreateSessionOptions();
-  EnsureDeviceOrtInit(*p_device_);
+  EnsureDeviceOrtInit(*p_device_, config_);
```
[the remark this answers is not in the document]
anujj: Done

src/models/model.cpp
anujj: Done

src/models/model.cpp
anujj: Done

src/models/model.cpp
```diff
@@ -369,8 +377,102 @@ DeviceInterface* SetProviderSessionOptions(OrtSessionOptions& session_options,
...
+
+          // Helper function to add generation phase input shapes
+          auto addGenerationInputShapes = [](std::ostringstream& shapes, int context_len) {
+            shapes << "," << Config::Defaults::AttentionMaskName << ":1x" << context_len << ","
+                   << Config::Defaults::InputIdsName << ":1x1,"
+                   << Config::Defaults::PositionIdsName << ":1x1";
+          };
+
+          // Helper function to add empty KV cache shapes for all layers
+          auto addEmptyKVCacheShapes = [&](std::ostringstream& shapes,
```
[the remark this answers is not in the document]
anujj: Done

src/models/model.cpp
anujj: Done

anujj: Done

src/models/model.cpp
anujj: Done

src/models/model.cpp
anujj: Done

src/models/model.cpp
anujj: Done

src/models/model.cpp
anujj: Done

src/config.cpp
anujj: Done

src/models/model.cpp
```diff
@@ -273,11 +280,111 @@ int32_t Tokenizer::TokenToTokenId(const char* token) const {
...
+                                             std::string_view key_pattern,
+                                             std::string_view value_pattern,
+                                             int seq_len,
+                                             int num_layers,
+                                             int num_kv_heads,
+                                             int head_dim) {
+    for (int i = 0; i < num_layers; i++) {
+      // Use the existing function to format the key/value names
+      std::string key_name = ComposeKeyValueName(std::string(key_pattern), i);
+      std::string value_name = ComposeKeyValueName(std::string(value_pattern), i);
```
baijumeswani, line 349:
```suggestion
      const std::string key_name = ComposeKeyValueName(std::string(key_pattern), i);
      const std::string value_name = ComposeKeyValueName(std::string(value_pattern), i);
```

baijumeswani (commented):
minor comment. But looks good otherwise.

src/models/model.cpp
anujj: Done

anujj:
@baijumeswani : Thanks for the review, addressed the review comment

baijumeswani (approved):

PR #1498, opened 2025-05-20
Use Phi-3 Instead of Qwen for Chat Template Test

sayanshaw24:


test/python/test_onnxruntime_genai_api.py
```diff
@@ -268,8 +268,8 @@ def test_tokenizer_encode_decode(device, phi2_for, batch):
     reason="Python 3.8 is required for downloading models.",
 )
 @pytest.mark.parametrize("device", devices)
-def test_qwen_chat_template(device, qwen_for):
-    model_path = qwen_for(device)
+def test_phi3_chat_template(device, phi3_for):
+    model_path = phi3_for(device)
```
baijumeswani, line 272: Could you please try to remove the template below?

sayanshaw24: good catch, missed that

baijumeswani (approved):

PR #1496, opened 2025-05-20
[#1508] Automatically install java maven artifact in the local maven repository

asoldano:
The best practice for Java Maven project is that project artifacts (jars, poms, etc.) are added to the local Maven repository as part of the build. This allows for easy consumption / integration /testing of such artifacts as dependencies in other Maven projects.
This PR automatically installs the generated onnxruntime-genai jar as a maven artifact in the local maven repository (if the "--build_java" flag is passed to the usual python build).

https://github.com/microsoft/onnxruntime-genai/issues/1508

PR #1511, opened 2025-05-27
Automatically install java maven artifact in the local maven reposito…

### LINKED ISSUE #1508: Automatically install java maven artifact in the local maven repository
The best practice for Java Maven project is that project artifacts (jars, poms, etc.) are added to the local Maven repository as part of the build. This allows for easy consumption / integration /testing of such artifacts as dependencies in other Maven projects.

asoldano:
…ry, as if a regular java maven project were built

(gradle build version)

Fixes #1508 

This is an alternative to PR #1496 , proposing a solution that follows @Craigacp hint (basically, use Gradle's publishToMavenLocal).

### DIFF
--- a/build.py
+++ b/build.py
@@ -433,7 +433,6 @@ def _run_android_tests(args: argparse.Namespace):
             # util.run([adb, "logcat", "-d", "*:E"])
             raise exception
 
-
 def _get_windows_build_args(args: argparse.Namespace):
     win_args = [
         "-DCMAKE_EXE_LINKER_FLAGS_INIT=/profile /DYNAMICBASE",
@@ -633,6 +632,21 @@ def build(args: argparse.Namespace, env: dict[str, str]):
         csharp_build_command += _get_csharp_properties(args, ort_lib_dir=lib_dir)
         util.run(csharp_build_command, cwd=REPO_ROOT / "src" / "csharp")
         util.run(csharp_build_command, cwd=REPO_ROOT / "test" / "csharp")
+
+    # If Java was built, install it to the local Maven repository
+    if args.build_java:
+        # Use the Gradle wrapper to publish to the local Maven repository
+        log.info("Publishing Java API to local Maven repository using Gradle...")
+        gradle_executable = str(REPO_ROOT / "src" / "java" / ("gradlew.bat" if util.is_windows() else "gradlew"))
+        # Use the src/java directory where build.gradle is located
+        java_build_dir = REPO_ROOT / "src" / "java"
+        # Pass build properties to Gradle.
+        gradle_properties = {
+            "cmakeBuildDir": str(args.build_dir)
+        }
+        # We've updated the Gradle build file to make signing optional, so we don't need to skip it explicitly
+        util.run([gradle_executable, "--no-daemon", "publishToMavenLocal"] + [f"-D{key}={value}" for key, value in gradle_properties.items()],
+                 cwd=java_build_dir)
 
 
 def package(args: argparse.Namespace, env: dict[str, str]):
--- a/src/java/build.gradle
+++ b/src/java/build.gradle
@@ -108,43 +108,104 @@ sourceSets.test {
 	}
 }
 
-if (cmakeNativeLibDir != null) {
-	// generate tasks to be called from cmake
+// generate tasks to be called from cmake
 
-	// Overwrite jar location
-	task allJar(type: Jar) {
-		manifest {
-			attributes('Automatic-Module-Name': "com.microsoft.onnxruntime.genai",
-					'Implementation-Title': 'onnxruntime-genai',
-					'Implementation-Version': project.version)
+// Overwrite jar location
+task allJar(type: Jar) {
+	manifest {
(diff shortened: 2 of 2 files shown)


Craigacp:
I'm a little unclear why this modifies the allJar build task and moves it outside of the cmakeBuildDir guard? Running gradle publishToMavenLocal -DcmakeBuildDir=../../build/<platform>/Release` when in `src/java` seems to execute ok aside from the signing issue. It also seems like it'll run the install even if the tests fail because the tests happen after the build. It's probably best to keep this close to the ORT build.gradle where possible. Also the install should probably be optional and guarded by a flag to `build.py`, as I imagine a lot of builds of this from source are going to be to test ORT-genai rather than for users (e.g. in the CI).

Defaulting to always installing the build to Maven local will cause problems because ORT-genai is not built by default with a `-SNAPSHOT` version number, it uses `-dev`, so it probably wants to rewrite the version string in that case so it gets handled like a regular snapshot and not a concrete version.

Craigacp:
I'm still unclear why there are any changes to the build.gradle beyond an if statement changing the version number from ending with `-dev` to `-SNAPSHOT`, and also allowing the signing to be disabled for the local publish task. Why is the native artifact packaging changed?

Craigacp:
Yeah, that's done by the CMake build system here though - https://github.com/microsoft/onnxruntime-genai/blob/main/src/java/CMakeLists.txt#L102.

asoldano:
Thanks, I was going to comment that, when running on Linux, I was getting the so files in the root of the generated jar.

Craigacp:
The arguments are slightly off in the CMake, it should use this:
```
diff --git a/src/java/CMakeLists.txt b/src/java/CMakeLists.txt
index cf61874..b1b0e35 100644
--- a/src/java/CMakeLists.txt
+++ b/src/java/CMakeLists.txt
@@ -123,7 +123,7 @@ else()
 endif()
 
 # run the build process
-set(GRADLE_ARGS cmakeBuild -DcmakeBuildDir=${JAVA_OUTPUT_DIR} -DnativeLibDir=${JAVA_PACKAGE_LIB_DIR})
+set(GRADLE_ARGS cmakeBuild -DcmakeBuildDir=${JAVA_OUTPUT_DIR} -DnativeLibDir=${JAVA_NATIVE_LIB_DIR})
 add_custom_command(TARGET onnxruntime-genai-jni POST_BUILD
                    COMMAND ${POST_BUILD_GRADLE_EXECUTABLE} ${GRADLE_OPTIONS} ${GRADLE_ARGS}
                    WORKING_DIRECTORY ${JAVA_SRC_ROOT})
@@ -217,7 +217,7 @@ if (ENABLE_TESTS)
   else()
     add_test(NAME onnxruntime-genai4j_test
              COMMAND ${GRADLE_EXECUTABLE} cmakeCheck
-               -DcmakeBuildDir=${JAVA_OUTPUT_DIR} -DnativeLibDir=${JAVA_PACKAGE_LIB_DIR}
+               -DcmakeBuildDir=${JAVA_OUTPUT_DIR} -DnativeLibDir=${JAVA_NATIVE_LIB_DIR}
              WORKING_DIRECTORY ${JAVA_SRC_ROOT})
   endif()
```
That causes the library to appear in the right place in the jar:
```
$ jar tf ./src/java/build/libs/onnxruntime-genai-0.9.0-dev.jar
META-INF/
META-INF/MANIFEST.MF
ai/
ai/onnxruntime/
ai/onnxruntime/genai/
ai/onnxruntime/genai/Adapters.class
ai/onnxruntime/genai/Model.class
ai/onnxruntime/genai/GenAIException.class
ai/onnxruntime/genai/Config.class
ai/onnxruntime/genai/GenAI.class
ai/onnxruntime/genai/Generator$Iterator.class
ai/onnxruntime/genai/Generator$1.class
ai/onnxruntime/genai/Generator.class
ai/onnxruntime/genai/GeneratorParams.class
ai/onnxruntime/genai/Sequences.class
ai/onnxruntime/genai/Tensor$ElementType.class
ai/onnxruntime/genai/Tensor.class
ai/onnxruntime/genai/NamedTensors.class
ai/onnxruntime/genai/Images.class
ai/onnxruntime/genai/MultiModalProcessor.class
ai/onnxruntime/genai/TokenizerStream.class
ai/onnxruntime/genai/SimpleGenAI.class
ai/onnxruntime/genai/Tokenizer.class
ai/onnxruntime/genai/native/
ai/onnxruntime/genai/native/linux-x64/
ai/onnxruntime/genai/native/linux-x64/libonnxruntime-genai.so
ai/onnxruntime/genai/native/linux-x64/libonnxruntime-genai-jni.so
META-INF/maven/
META-INF/maven/com.microsoft.onnxruntime/
META-INF/maven/com.microsoft.onnxruntime/onnxruntime-genai/
META-INF/maven/com.microsoft.onnxruntime/onnxruntime-genai/pom.xml
```

asoldano:
@Craigacp thanks a lot. I've used your patch and cleaned up most of the changes, the result is in https://github.com/microsoft/onnxruntime-genai/pull/1531 . Please have look :-)

PR #1520, opened 2025-06-02
Add Gemma3 Model support for NvTensorRtRtx execution provider

anujj:
## Problem
Gemma3 models failed to build with NvTensorRtRtx execution provider due to:
- Unsupported standalone RotaryEmbedding nodes
- Dual RoPE configuration requirements (local vs global attention layers)
- Missing cos/sin cache inputs to GroupQueryAttention nodes

## Root Cause
1. NvTensorRtRtx doesn't support standalone RotaryEmbedding contrib ops
2. Gemma3 uses mixed attention pattern: 5/6 layers use local attention 
   (sliding window + rope_local_base_freq), 1/6 layers use global attention
3. Previous implementation created incompatible RotaryEmbedding nodes
4. GroupQueryAttention expected cos/sin caches but they weren't provided

## Solution
- **Dual Cache System**: Create separate cos/sin caches for local vs global layers
  - `cos_cache_global/sin_cache_global`
  - `cos_cache_local/sin_cache_local` (theta=rope_local_base_freq)
- **Skip RotaryEmbedding Nodes**: For NvTensorRtRtx, skip standalone node creation
- **Internal RoPE Processing**: Set `use_rope_in_attn=True` to handle RoPE inside GroupQueryAttention
- **Dynamic Layer Configuration**: Automatically select appropriate cache/theta based on layer type using sliding_window_pattern

## Implementation Details
- Layer pattern: `(layer_id + 1) % sliding_window_pattern == 0` → Global, else Local
- NvTensorRtRtx: RoPE handled internally by GroupQueryAttention with proper cache inputs
- Other EPs: Standard RotaryEmbedding nodes with appropriate caches

## Testing
✅ Gemma3-4b-it model builds successfully with NvTensorRtRtx EP
✅ Maintains compatibility with other execution providers

@BLSharda @kunal-vaishnavi @baijumeswani


src/python/py/models/builder.py
```diff
@@ -3641,6 +3654,36 @@
         sin_cache_name = kwargs.get("sin_cache_name", self.sin_cache_global_name if self.window_size == -1 else self.sin_cache_local_name)
         return super().make_rotary_embedding_caches(cos_cache_name=cos_cache_name, sin_cache_name=sin_cache_name)
 
+    def make_rotary_embedding(self, name, **kwargs):
```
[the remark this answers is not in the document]
anujj: fixed

src/python/py/models/builder.py
```diff
@@ -3618,13 +3618,26 @@ def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
...
+        if self.ep == "NvTensorRtRtx":
+            self.original_theta = self.rotemb_attrs.get("theta")
+
         self.make_rotary_embedding_multi_cache()
 
     def make_attention_init(self):
         self.attention_attrs["q_norm"] = True
         self.attention_attrs["k_norm"] = True
+
+        # For NvTensorRtRtx, force RoPE to be handled inside GroupQueryAttention
```
kunal-vaishnavi: We can set the default to use RoPE inside GQA instead and enable this for all EPs where possible.

https://github.com/microsoft/onnxruntime-genai/blob/8530e958f92172e0e76c404288147a7b1c8c82ae/src/python/py/models/builder.py#L363-L368

src/python/py/models/builder.py
```diff
@@ -3618,13 +3618,26 @@ def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
         super().__init__(config, io_dtype, onnx_dtype, ep, cache_dir, extra_options)
         self.is_local = lambda layer_id: bool((layer_id + 1) % config.sliding_window_pattern)
         self.rope_local_theta = config.rope_local_base_freq
+
+        if self.ep == "NvTensorRtRtx":
+            self.original_theta = self.rotemb_attrs.get("theta")
```
kunal-vaishnavi: I think this can be restructured to create the right rotary embedding caches (for either the RotaryEmbedding or GQA ops) on a per-layer basis for all EPs.

src/python/py/models/builder.py
anujj: Done

src/python/py/models/builder.py
anujj: **Addressed by restructuring the implementation:**

1. **RoPE-in-GQA now default** for all supported EPs (except "dml", "webgpu" as suggested)
2. **Per-layer cache creation** - proper cos/sin caches for both local and global attention layers
3. **Cleaner logic** - removed EP-specific conditionals where possible
4. Validated change with NvTensorRtRtx and DML 

hope that was the expectation from the review comment

src/python/py/models/builder.py
```diff
@@ -3618,29 +3618,84 @@ def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
...
+        self.original_theta = self.rotemb_attrs.get("theta", 10000.0)
+
+        # Create multi-cache for all EPs (restructured approach)
         self.make_rotary_embedding_multi_cache()
 
     def make_attention_init(self):
         self.attention_attrs["q_norm"] = True
         self.attention_attrs["k_norm"] = True
+
+        # Set default to use RoPE inside GQA for all EPs where possible
```
kunal-vaishnavi, line 3632: I think this setting should already be handled when calling `super().make_attention_init()`

src/python/py/models/builder.py
```diff
@@ -3618,29 +3618,84 @@ def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
...
+        self.attention_attrs["use_rope_in_attn"] = (
+            self.ep not in ["dml", "webgpu"] and
+            not self.attention_attrs.get("q_norm", False) and
+            not self.attention_attrs.get("k_norm", False)
+        )
+
         super().make_attention_init()
 
+        # Force RoPE-in-GQA for NvTensorRtRtx (always required)
+        if self.ep == "NvTensorRtRtx":
```
kunal-vaishnavi: This should already be handled inside `super().make_attention_init()` since you are presumably calling the NV EP with float16 precision and that is already a valid GQA combination.

https://github.com/microsoft/onnxruntime-genai/blob/d6979cfac314bd92da2509bb379a9dd418e0a8f2/src/python/py/models/builder.py#L338-L371

src/python/py/models/builder.py
```diff
@@ -3618,29 +3618,84 @@ def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
         super().__init__(config, io_dtype, onnx_dtype, ep, cache_dir, extra_options)
         self.is_local = lambda layer_id: bool((layer_id + 1) % config.sliding_window_pattern)
         self.rope_local_theta = config.rope_local_base_freq
+
+        # Store original theta for restoration
+        self.original_theta = self.rotemb_attrs.get("theta", 10000.0)
```
kunal-vaishnavi: The `self.original_theta` variable is not used anywhere. This can be removed.

src/python/py/models/builder.py
```diff
@@ -3618,29 +3618,84 @@ def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
...
+        # Restore original theta
+        self.rotemb_attrs["theta"] = original_theta
+
     def make_rotary_embedding_caches(self, **kwargs):
+        # Use the appropriate cache names based on window size (per-layer basis)
         cos_cache_name = kwargs.get("cos_cache_name", self.cos_cache_global_name if self.window_size == -1 else self.cos_cache_local_name)
         sin_cache_name = kwargs.get("sin_cache_name", self.sin_cache_global_name if self.window_size == -1 else self.sin_cache_local_name)
         return super().make_rotary_embedding_caches(cos_cache_name=cos_cache_name, sin_cache_name=sin_cache_name)
 
+    def make_rotary_embedding(self, name, **kwargs):
```
kunal-vaishnavi: The decision to create or skip creating the `RotaryEmbedding` op is already handled below. I think this can be removed.

https://github.com/microsoft/onnxruntime-genai/blob/7c02fe6f029952912f28ac327e50a499d541ff5d/src/python/py/models/builder.py#L1944-L1954

src/python/py/models/builder.py
```diff
@@ -3618,29 +3618,84 @@ def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
...
+
+        # Create multi-cache for all EPs (restructured approach)
         self.make_rotary_embedding_multi_cache()
 
     def make_attention_init(self):
         self.attention_attrs["q_norm"] = True
         self.attention_attrs["k_norm"] = True
+
+        # Set default to use RoPE inside GQA for all EPs where possible
+        # Some EPs don't support fusing rotary embeddings inside GQA yet
```
kunal-vaishnavi: I think this setting should already be handled when calling `super().make_attention_init()`. To keep the rotary embeddings inside GQA, we can instead re-define the below setting in the parent method as follows.

```python
self.attention_attrs["use_rope_in_attn"] = self.ep not in ["dml", "webgpu"]
```

src/python/py/models/builder.py
```diff
@@ -3618,29 +3618,84 @@ def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
...
+        self.attention_attrs["use_rope_in_attn"] = (
+            self.ep not in ["dml", "webgpu"] and
+            not self.attention_attrs.get("q_norm", False) and
+            not self.attention_attrs.get("k_norm", False)
+        )
```
kunal-vaishnavi: In some precision + EP combinations, this approach will cause an extra `position_ids` graph input to be in the resulting model when it should not be there.

kunal-vaishnavi:
I ran a quick test with the current main branch except

https://github.com/microsoft/onnxruntime-genai/blob/8530e958f92172e0e76c404288147a7b1c8c82ae/src/python/py/models/builder.py#L363-L368

was changed to

```python
self.attention_attrs["use_rope_in_attn"] = self.ep not in ["dml", "webgpu"]
```

This fixes the issue of moving the rotary embeddings inside GQA. For Gemma-3, the cos/sin caches are already calculated and chosen on a per-layer basis correctly. Therefore, this PR just needs to make the above one-line change instead of all of these changes and the resulting model should be correct for the NV EP.

PR #1523, opened 2025-06-03
Clamp KV Cache Size to Sliding Window for NvTensorRtRtx EP

BLSharda:
# Overview
This pull request modifies the ORT GenAI to ensure that the key-value (KV) cache size used by the NvTensorRtRtx Execution Provider (EP) is clamped to the configured sliding window size. 

# Changes
1. Updated GenAI Model Builder to write 'sliding windowed size' config in genai_config.json.
2. Updated GenAI kv cache to limit the kv allocation size when 'sliding windowed size' is set for NvTensorRtRtx EP. 

# Benefits
This leads to required memory footprint reduction for the models that uses sliding window.

# Tests
[x] Locally verified DeepSeek-R1-Distill-Qwen-1.5B and few others for NvTensorRtRtx EP.

@kunal-vaishnavi @baijumeswani

### DIFF
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -425,6 +425,8 @@ struct Decoder_Element : JSON::Element {
       v_.num_hidden_layers = static_cast<int>(JSON::Get<double>(value));
     } else if (name == "head_size") {
       v_.head_size = static_cast<int>(JSON::Get<double>(value));
+    } else if (name == "sliding_window_size") {
+      v_.sliding_window_size = static_cast<int>(JSON::Get<double>(value));
     } else
       throw JSON::unknown_value_error{};
   }
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -367,6 +367,8 @@ struct SlidingWindow_Element : JSON::Element {
       v_->alignment = JSON::Get<std::string_view>(value);
     } else if (name == "slide_key_value_cache") {
       v_->slide_key_value_cache = JSON::Get<bool>(value);
+    } else if (name == "slide_inputs") {
+      v_->slide_inputs = JSON::Get<bool>(value);
     } else
       throw JSON::unknown_value_error{};
   }
@@ -425,8 +427,6 @@ struct Decoder_Element : JSON::Element {
       v_.num_hidden_layers = static_cast<int>(JSON::Get<double>(value));
     } else if (name == "head_size") {
       v_.head_size = static_cast<int>(JSON::Get<double>(value));
-    } else if (name == "sliding_window_size") {
-      v_.sliding_window_size = static_cast<int>(JSON::Get<double>(value));
     } else
       throw JSON::unknown_value_error{};
   }
--- a/src/models/kv_cache.cpp
+++ b/src/models/kv_cache.cpp
@@ -175,7 +175,7 @@ DefaultKeyValueCache::DefaultKeyValueCache(State& state)
   }
 
   // Set the size after empty_past_ has been created with 0 for this field
-  if (past_present_share_buffer_ && model_.config_->model.decoder.sliding_window_size > 0)
+  if (state.model_.p_device_->GetType() == DeviceType::NvTensorRtRtx && model_.config_->model.decoder.sliding_window_size > 0)
     shape_[2] = std::min(state_.params_->search.max_length, model_.config_->model.decoder.sliding_window_size);
   else if (past_present_share_buffer_)
     shape_[2] = state_.params_->search.max_length;
--- a/src/models/kv_cache.cpp
+++ b/src/models/kv_cache.cpp
@@ -175,8 +175,10 @@ DefaultKeyValueCache::DefaultKeyValueCache(State& state)
   }
 
   // Set the size after empty_past_ has been created with 0 for this field
-  if (state.model_.p_device_->GetType() == DeviceType::NvTensorRtRtx && model_.config_->model.decoder.sliding_window_size > 0)
-    shape_[2] = std::min(state_.params_->search.max_length, model_.config_->model.decoder.sliding_window_size);
+  if (state.model_.p_device_->GetType() == DeviceType::NvTensorRtRtx &&
+      model_.config_->model.decoder.sliding_window.has_value() &&
+      model_.config_->model.decoder.sliding_window->window_size > 0)
+    shape_[2] = std::min(state_.params_->search.max_length, model_.config_->model.decoder.sliding_window->window_size);
   else if (past_present_share_buffer_)
     shape_[2] = state_.params_->search.max_length;
 
@@ -424,7 +426,8 @@ std::unique_ptr<KeyValueCache> CreateKeyValueCache(State& state) {
     return nullptr;
   }
 
-  if (state.model_.config_->model.decoder.sliding_window &&
+  if (state.model_.p_device_->GetType() != DeviceType::NvTensorRtRtx && 
+      state.model_.config_->model.decoder.sliding_window && 
       state.model_.config_->model.decoder.sliding_window->slide_key_value_cache) {
     return std::make_unique<WindowedKeyValueCache>(state);
   }
--- a/src/models/kv_cache.cpp
+++ b/src/models/kv_cache.cpp
@@ -177,10 +177,12 @@ DefaultKeyValueCache::DefaultKeyValueCache(State& state)
   // Set the size after empty_past_ has been created with 0 for this field
   if (state.model_.p_device_->GetType() == DeviceType::NvTensorRtRtx &&
       model_.config_->model.decoder.sliding_window.has_value() &&
-      model_.config_->model.decoder.sliding_window->window_size > 0)
-    shape_[2] = std::min(state_.params_->search.max_length, model_.config_->model.decoder.sliding_window->window_size);
-  else if (past_present_share_buffer_)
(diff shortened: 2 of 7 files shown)


src/config.cpp
```diff
@@ -425,6 +425,8 @@ struct Decoder_Element : JSON::Element {
       v_.num_hidden_layers = static_cast<int>(JSON::Get<double>(value));
     } else if (name == "head_size") {
       v_.head_size = static_cast<int>(JSON::Get<double>(value));
+    } else if (name == "sliding_window_size") {
```
baijumeswani: We already have a sliding window config. Could you please re-use that?

kunal-vaishnavi:
You can re-use the sliding window section in the `genai_config.json` file.

https://github.com/microsoft/onnxruntime-genai/blob/7c02fe6f029952912f28ac327e50a499d541ff5d/src/config.cpp#L358-L376

src/config.cpp
BLSharda: I tried to reuse struct "sliding_window" in genai_config.json from [Kunal's comment],(https://github.com/microsoft/onnxruntime-genai/pull/1523#issuecomment-2948336406) but this engages few classes in GenAI runtime like WindowedInputIDs, WindowedPositionInputs and WindowedKeyValueCache not compatible to GPU at the moment. And they seem to fail with expected datatypes of inputs to int8 etc.

https://github.com/microsoft/onnxruntime-genai/blob/7c02fe6f029952912f28ac327e50a499d541ff5d/src/models/windowed_kv_cache.cpp#L44

Can you suggest if there is a way to avoid use of WindowedInputIDs, WindowedPositionInputs and WindowedKeyValueCache while still using existing "sliding_window" in genai_config.json?

baijumeswani: Does NvTensorRtRtx use the decoder-pipeline model? The windowed positions and windowed inputs are only engaged for the decoder-pipeline model. Having said that, ort-genai will pad the input prompt to a multiple of the window size when the sliding window is defined.

What does windowing in the NvTensorRtRtx EP do? It seems it sets the upper bound on the kv cache. What happens when the user's prompt is larger than the window size? Does this case need to be addressed?

gaugarg-nv: > What happens when the user's prompt is larger than the window size? Does this case need to be addressed?

I can try to answer this. This is already handled, @baijumeswani . The pre-fill phase attention is run with the KV values for the whole prompt. However, when writing to the KV cache, we copy only the last "sliding_window" tokens.

I would like to understand how "sliding_window" is handled by CUDA EP. If CUDA EP allows KV cache to be larger than "sliding_window", does it mean some tokens are never used in the KV cache? Or, does it mean, CUDA EP just ignores "sliding window " during attention calculation?

BLSharda: > Does NvTensorRtRtx use the decoder-pipeline model?

Not sure if I understand decoder-pipeline? Currently, we are testing text generation models using ModelBuilder to export onnx, like DeepSeek-R1-Distill-Qwen-7B. Ultimately, we would like to support all the other types of models that GenAI supports.

> The windowed positions and windowed inputs are only engaged for the decoder-pipeline model.

Based on the code, it seems this choice is purely based the `"sliding_window":` config value and doesn't seem tied to model class or anything.

WindowedInputIDs - https://github.com/microsoft/onnxruntime-genai/blob/bfc8027c3635a8bb0abaad95b432d6be44e790c0/src/models/input_ids.cpp#L229

WindowedPositionInputs - https://github.com/microsoft/onnxruntime-genai/blob/bfc8027c3635a8bb0abaad95b432d6be44e790c0/src/models/position_inputs.cpp#L441

baijumeswani: > The pre-fill phase attention is run with the KV values for the whole prompt. However, when writing to the KV cache, we copy only the last "sliding_window" tokens.

So, the actual sliding of the kv cache happens inside ort? If this is the case, this is different from what ort-genai considers sliding window since sliding window from ort-genai's perspective means to actually manage the sliding (which is what windowedkeyvaluecache does).

> I would like to understand how "sliding_window" is handled by CUDA EP. If CUDA EP allows KV cache to be larger than "sliding_window", does it mean some tokens are never used in the KV cache? Or, does it mean, CUDA EP just ignores "sliding window " during attention calculation?

@gaugarg-nv For now, we do not have support for sliding a window over the kv cache for cuda ep from ort-genai's perspective. That means, we assume the entire context is preserved in memory for generation. We might introduce support for sliding window for cuda ep in the future.

> Not sure if I understand decoder-pipeline?

ort-genai supports multiple <"types"> of models. Naming a few:
- decoder only models: most language models with a single decoder onnx file. Handled in this file: https://github.com/microsoft/onnxruntime-genai/blob/main/src/models/decoder_only.cpp
- decoder pipeline models: language models where the model needs to be split into multiple onnx files (output of one onnx model feeds into input of the next one; chaining them together). This is handled in this file: https://github.com/microsoft/onnxruntime-genai/blob/main/src/models/decoder_only_pipeline.cpp. Primarily used for NPU models where models need to be split due to execution constraints.
- Multi-modal models: Serves multiple modalities (speech+vision+decoder). This is handled in https://github.com/microsoft/onnxruntime-genai/blob/main/src/models/multi_modal.cpp.

From my understanding, NVTensorrtrtx only uses decoder_only.cpp which uses DefaultInputIds and DefaultPositionIds: https://github.com/microsoft/onnxruntime-genai/blob/49d9f76fa0ac16835d8d2e92aa950e15e726c58d/src/models/decoder_only.h#L30-L34

I think maybe we should approach this by making the genai_config for the nvtensorrtrtx as follows:

```json
"sliding_window": {
    "window_size": 64,
    "slide_inputs": false, # new config that will need to be introduced
    "slide_key_value_cache": false
},
```

^ `slide_inputs`=false and `slide_key_value_cache`=false would mean that the sliding is taken care of by ort. And the only thing ort-genai needs to maintain is the actual size of the kv cache.

If we go down this route, please add relevant documentation to the kv cache and the inputs creation function.

gaugarg-nv: > So, the actual sliding of the kv cache happens inside ort? If this is the case, this is different from what ort-genai considers sliding window since sliding window from ort-genai's perspective means to actually manage the sliding (which is what windowedkeyvaluecache does)

Yes, that's correct. We handle the sliding window inside the GQA implementation of TRT-RTX. The GQA op has all the required information to implement SWA. **The only constraint it has is that this requires the KV cache size to be the same as min(sliding_window, target_seq_length).** This is what we do-
1. Pre-fill phase: Consider all income tokens during attention calculation, but attend to only "sliding_window" tokens when calculating attention of an individual token in the prompt.  Write only the last "sliding_window" tokens in the KV cache if the prompt length is larger than sliding_window.
2. During the "Gen phase", if the KV cache is already full, use the KV cache as a circular buffer and overwrite the old token with the new incoming token.

I don't have a good understanding of "windowedkeyvaluecache". I will try to understand it better. Does it allow limiting the KV cache size to a sliding window? Does GQA op work with it?

baijumeswani: Thanks. I don't think you need to use the WindowedKeyValueCache since that is used for the QNN ep and it actually does the sliding logic.

The current approach is ok. We can just re-use the sliding window config.

gaugarg-nv: Thanks @baijumeswani . For other EPs, what is your recommended approach - Use `WindowedKeyValueCache` or implement it in GQA like TRT-RTX does? It's important to support the sliding window to stay compliant with the PyTorch implementation and to reduce the memory footprint for large sequence lengths. Another issue is that if other EPs are not compliant with the spec, end users will not see consistent output across EPs, which is not ideal.

BLSharda: @baijumeswani, I have updated code to reuse "sliding_window" config with new "slide_inputs": false option. Can you review this?

anskumar01:
@kunal-vaishnavi , @baijumeswani , can you please help with the review? Are there any open concerns?

src/models/kv_cache.cpp
```diff
@@ -422,7 +428,8 @@ std::unique_ptr<KeyValueCache> CreateKeyValueCache(State& state) {
     return nullptr;
   }
 
-  if (state.model_.config_->model.decoder.sliding_window &&
+  if (state.model_.p_device_->GetType() != DeviceType::NvTensorRtRtx &&
```
baijumeswani, line 431: Can we avoid dependency on the EP here? Can we assume the config will contain slide_key_value_cache=false?

src/models/kv_cache.cpp
```diff
@@ -175,8 +175,14 @@ DefaultKeyValueCache::DefaultKeyValueCache(State& state)
   }
 
   // Set the size after empty_past_ has been created with 0 for this field
-  if (past_present_share_buffer_)
+  if (state.model_.p_device_->GetType() == DeviceType::NvTensorRtRtx &&
```
baijumeswani, line 178: Can we avoid dependence on the DeviceType::NvTensorRtRtx in this check?

BLSharda: I noticed that other EPs (such as CUDA) encounter issues when limiting the kv size to the sliding window size, so I’ve enabled this only for the NV EP for now. Would it be alright if we try enabling this for the other EPs in a separate PR? This might require additional testing and handling.

baijumeswani (approved):

PR #1222, opened 2025-02-04
Overlap KV cache update for WindowedKeyValueCache in DecoderOnlyPipelineState

edgchen1:
# Description

Add support for overlapping the KV cache update for a pipelined model part with the graph execution of other pipelined model parts. It only applies to `DecoderOnlyPipelineState` with `WindowedKeyValueCache`.

For example, consider a model with two parts (graph[1] and graph[2]) that have KV caches.

This is the approach in this PR:
```
iter 1 graph[1] run | -
iter 1 graph[2] run | iter 1 graph[1] KV cache update
iter 2 graph[1] run | iter 1 graph[2] KV cache update
iter 2 graph[2] run | iter 2 graph[1] KV cache update
iter 3 graph[1] run | iter 2 graph[2] KV cache update
```

For comparison, this is the existing approach:
```
iter 1 graph[1] run
iter 1 graph[2] run
iter 1 graph[1] KV cache update
iter 1 graph[2] KV cache update
iter 2 graph[1] run
iter 2 graph[2] run
iter 2 graph[1] KV cache update
iter 2 graph[2] KV cache update
```

# Measurements

Token generation rate with QNN EP 3-part Llama3.2 3B:
Baseline: 15.5328 tokens/sec
Updated: 18.2981 tokens/sec

Prompt processing logic is unchanged.

PR #1526, opened 2025-06-03
Overlap prompt processing KV cache update for WindowedKeyValueCache in DecoderOnlyPipelineState

edgchen1:
# Description
In a pipelined model, we can reduce some latency by updating part of the KV cache while other model parts are running. PR #1222 added support for overlapping the KV cache update during token generation. This change enables this for prompt processing as well.

# Measurements
Baseline: based off of e9656943f328a1304cfbe4f555fd89f0e6b61e71, with additional code for tracing
Updated: based off of ff2b88d9c74ef9ab76e045edb8903c37af594801, with additional code for tracing

Model: Phi-3.5-mini from QC AI hub
Prompt length: 1221
Generated tokens: 20

-|baseline|updated
-|-|-
average prompt processing rate (t/s)|547.344|574.978
average token generation rate (t/s)|8.28566|8.57479

## Traces

Bottom row is closest to the times of the ORT session run calls. Note how in the updated picture there are fewer gaps in the bottom row.

baseline:
image

updated:
image


src/models/windowed_kv_cache.h
```diff
@@ -17,40 +17,61 @@ struct WindowedKeyValueCache : KeyValueCache {
...
+    size_t window_index{0};
+    size_t window_size{};
+    size_t num_windows{};
+    bool is_first_update{true};
+
+    CacheTensorShape key_cache_shape_in{}, key_cache_shape_out{};
+    CacheTensorShape value_cache_shape_in{}, value_cache_shape_out{};
+  };
+
+  static std::vector<LayerState> MakeInitialPerLayerStates(size_t layer_count,
```
baijumeswani, line 42: Should this be static?

baijumeswani (approved):

src/models/windowed_kv_cache.h
edgchen1: it doesn't actually access any data members. however, it does use the private `LayerState` type, so I made it a private static member function instead of a free function.

PR #1531, opened 2025-06-05
Automatically install java maven artifact in the local maven repository

### LINKED ISSUE #1508: Automatically install java maven artifact in the local maven repository

asoldano:
Fix CMakeLists.txt (thanks @Craigacp) and install java maven artifact in the local maven repository when --publish_java_maven flag is used

Fixes #1508 

(this replaces https://github.com/microsoft/onnxruntime-genai/pull/1511)

### DIFF
--- a/build.py
+++ b/build.py
@@ -87,6 +87,7 @@ def _parse_args():
     # Default to not building the language bindings
     parser.add_argument("--build_csharp", action="store_true", help="Build the C# API.")
     parser.add_argument("--build_java", action="store_true", help="Build Java bindings.")
+    parser.add_argument("--publish_java_maven", action="store_true", help="Publish Java bindings to local Maven repository after tests.")
 
     parser.add_argument("--parallel", action="store_true", help="Enable parallel build.")
 
@@ -502,6 +503,7 @@ def update(args: argparse.Namespace, env: dict[str, str]):
         f"-DENABLE_JAVA={'ON' if args.build_java else 'OFF'}",
         f"-DBUILD_WHEEL={build_wheel}",
         f"-DUSE_GUIDANCE={'ON' if args.use_guidance else 'OFF'}",
+        f"-DPUBLISH_JAVA_MAVEN={'ON' if args.publish_java_maven else 'OFF'}",
     ]
 
     if args.ort_home:
--- a/src/java/CMakeLists.txt
+++ b/src/java/CMakeLists.txt
@@ -123,7 +123,11 @@ else()
 endif()
 
 # run the build process
-set(GRADLE_ARGS cmakeBuild publishToMavenLocal -DcmakeBuildDir=${JAVA_OUTPUT_DIR} -DnativeLibDir=${JAVA_NATIVE_LIB_DIR})
+set(GRADLE_ARGS cmakeBuild)
+if(PUBLISH_JAVA_MAVEN)
+  set(GRADLE_ARGS ${GRADLE_ARGS} publishToMavenLocal)
+endif()
+set(GRADLE_ARGS ${GRADLE_ARGS} -DcmakeBuildDir=${JAVA_OUTPUT_DIR} -DnativeLibDir=${JAVA_NATIVE_LIB_DIR})
 add_custom_command(TARGET onnxruntime-genai-jni POST_BUILD
                    COMMAND ${POST_BUILD_GRADLE_EXECUTABLE} ${GRADLE_OPTIONS} ${GRADLE_ARGS}
                    WORKING_DIRECTORY ${JAVA_SRC_ROOT})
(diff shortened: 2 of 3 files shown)


build.py
```diff
@@ -87,6 +87,7 @@ class HelpFormatter(argparse.ArgumentDefaultsHelpFormatter, argparse.RawDescript
     # Default to not building the language bindings
     parser.add_argument("--build_csharp", action="store_true", help="Build the C# API.")
     parser.add_argument("--build_java", action="store_true", help="Build Java bindings.")
+    parser.add_argument("--publish_java_maven", action="store_true", help="Publish Java bindings to local Maven repository after tests.")
```
baijumeswani, line 90: Could we avoid exposing this as an argument to the `build.py` script? Are there good defaults? If this is needed only at the cmake level, shall we ask users to pass it in like `python build.py ... --cmake_extra_defines PUBLISH_JAVA_MAVEN=ON`?

baijumeswani (commented):
Is this pull request ready for review? Does it have any implications on existing java library installation (for those who don't use maven)?

Craigacp (approved):
LGTM.

build.py
Craigacp: It tells the Java build to put it in the user's local Maven repository which will be checked by other Maven & Gradle builds on their system, so it makes it easier for local development. However you don't want it for all the builds, particularly if you're working on a feature or don't plan to use the local build (e.g. in the CI) so defaulting it to be on isn't ideal.

src/java/CMakeLists.txt
```diff
@@ -123,7 +123,11 @@ else()
 endif()
 
 # run the build process
-set(GRADLE_ARGS cmakeBuild -DcmakeBuildDir=${JAVA_OUTPUT_DIR} -DnativeLibDir=${JAVA_PACKAGE_LIB_DIR})
+set(GRADLE_ARGS cmakeBuild)
+if(PUBLISH_JAVA_MAVEN)
+  set(GRADLE_ARGS ${GRADLE_ARGS} publishToMavenLocal)
+endif()
+set(GRADLE_ARGS ${GRADLE_ARGS} -DcmakeBuildDir=${JAVA_OUTPUT_DIR} -DnativeLibDir=${JAVA_NATIVE_LIB_DIR})
```
edgchen1, line 130:
```suggestion
if(PUBLISH_JAVA_MAVEN)
  list(APPEND GRADLE_ARGS publishToMavenLocal)
endif()
list(APPEND GRADLE_ARGS -DcmakeBuildDir=${JAVA_OUTPUT_DIR} -DnativeLibDir=${JAVA_NATIVE_LIB_DIR})
```

build.py
edgchen1, line 90: would it be worth mentioning "local" in the option name? e.g., build.py `--publish_java_maven_local`, CMake `PUBLISH_JAVA_MAVEN_LOCAL`

build.py
```diff
@@ -502,6 +503,7 @@ def update(args: argparse.Namespace, env: dict[str, str]):
         f"-DENABLE_JAVA={'ON' if args.build_java else 'OFF'}",
         f"-DBUILD_WHEEL={build_wheel}",
         f"-DUSE_GUIDANCE={'ON' if args.use_guidance else 'OFF'}",
+        f"-DPUBLISH_JAVA_MAVEN={'ON' if args.publish_java_maven else 'OFF'}",
```
edgchen1, line 506: maybe also define a CMake option here:
https://github.com/microsoft/onnxruntime-genai/blob/0a34c76e024cff85f0018f495e67de0bda417ffe/cmake/options.cmake#L10

could make it dependent on `ENABLE_JAVA`, as I assume this one doesn't make sense if `ENABLE_JAVA` is false.

build.py
asoldano: Thanks @Craigacp , resolving this conversation.

src/java/CMakeLists.txt
asoldano: OK, thanks for the review and sorry for the late follow up (I've been on holiday). I'm addressing this in https://github.com/microsoft/onnxruntime-genai/pull/1570 , which is also rebased against main).

build.py
asoldano: OK, thanks for the review and sorry for the late follow up (I've been on holiday). I'm addressing this in https://github.com/microsoft/onnxruntime-genai/pull/1570 , which is also rebased against main.

build.py
asoldano: OK, thanks for the review and sorry for the late follow up (I've been on holiday). I'm addressing this in https://github.com/microsoft/onnxruntime-genai/pull/1570 , which is also rebased against main).

PR #1533, opened 2025-06-06
Update OnnxRuntimeGenAIChatClient with chat template and guidance

stephentoub:
Adds support for using the model's chat template format as well as applying JSON schema guidance.

### DIFF
--- a/src/csharp/OnnxRuntimeGenAIChatClient.cs
+++ b/src/csharp/OnnxRuntimeGenAIChatClient.cs
@@ -5,6 +5,8 @@ using System;
 using System.Collections.Generic;
 using System.Runtime.CompilerServices;
 using System.Text;
+using System.Text.Json;
+using System.Text.Json.Serialization;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Extensions.AI;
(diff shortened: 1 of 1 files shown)


src/csharp/OnnxRuntimeGenAIChatClient.cs
```diff
@@ -267,15 +269,41 @@ private bool IsStop(string token, ChatOptions? options) =>
...
         {
-            sb.Append(message).AppendLine();
+            if (message.Text is string text)
+            {
+                prompt.Append(separator);
+                separator = ",";
+
+                m.Role = message.Role.Value;
+                m.Content = text;
+                prompt.Append(JsonSerializer.Serialize(m));
```
f2bo: Shouldn't this specify the `JsonTypeInfo`?
```
prompt.Append(JsonSerializer.Serialize(m, OnnxJsonContext.Default.SerializableMessage));
```

stephentoub: It should, yes, not sure how I managed to drop that.

I'll also submit a separate PR to enable the NativeAOT analyzers, which would have also caught this.

PR #1538, opened 2025-06-09
Trim genai version for dev runs in Nuget packaging

ajindal1:
Trim genai version for dev runs in Nuget packaging step and fixes the issue for Nuget packaging output for dev runs.

Currently the genai version for dev runs is something like this "0.9.0-dev-423423-da43-d34*" and because of that it doesn't match the file names. We just need to extract the first two parts in the genai version and this PR does the same. Have started test runs to ensure it works properly.

### DIFF
--- a/.pipelines/stages/jobs/nuget-packaging-job.yml
+++ b/.pipelines/stages/jobs/nuget-packaging-job.yml
@@ -232,25 +232,17 @@ jobs:
       $outputDir = '$(Build.Repository.LocalPath)/$(buildDir)'
       Write-Host "List extracted artifacts"
       Get-ChildItem -Path $nativeBuildOutputDir -Recurse
-      # For non-stable versions, we don't need the complete version, just the first two parts
-      # e.g. "0.9.0-dev-rest" → "0.9.0-dev"
-      $version_parts = "$(genai_version)" -split '-', 3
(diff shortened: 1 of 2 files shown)


.pipelines/stages/jobs/nuget-packaging-job.yml
```diff
@@ -232,17 +232,25 @@ jobs:
...
       Write-Host "List extracted artifacts"
       Get-ChildItem -Path $nativeBuildOutputDir -Recurse
+      # For non-stable versions, we don't need the complete version, just the first two parts
+      # e.g. "0.9.0-dev-rest" → "0.9.0-dev"
+      $version_parts = "$(genai_version)" -split '-', 3
+      if ($version_parts.Count -ge 2) {
+        $req_genai_version = "$($version_parts[0])-$($version_parts[1])"
+      } else {
+        $req_genai_version = "$(genai_version)"
+      }
```
baijumeswani: Instead of extracting here, try setting another variable in this file: https://github.com/microsoft/onnxruntime-genai/blob/31a00fc3cf081b1c9ba5db9efc350f79cf7b6109/.pipelines/stages/jobs/steps/utils/set-genai-version.yml#L28-L36

You can set the variable and call it `genai_version_raw` and set it to the ${version}.

PR #1539, opened 2025-06-09
Default `add_generation_prompt` for Chat Templates to true in Python API

sayanshaw24:
Adds default to make `add_generation_prompt` optional in the Python API (note: because of lack of support for optionals in the C API we can only do this for Python).

### DIFF
--- a/src/python/python.cpp
+++ b/src/python/python.cpp
@@ -358,21 +358,7 @@ PYBIND11_MODULE(onnxruntime_genai, m) {
       })
       .def("to_token_id", &OgaTokenizer::ToTokenId)
       .def("decode", [](const OgaTokenizer& t, pybind11::array_t<int32_t> tokens) -> std::string { return t.Decode(ToSpan(tokens)).p_; })
-      .def(
-    "apply_chat_template",
-    [](const OgaTokenizer& t,
-       const char* messages,
-       const char* template_str,
-       const char* tools,
-       bool add_generation_prompt) -> std::string {
-        return t.ApplyChatTemplate(template_str, messages, tools, add_generation_prompt).p_;
-    },
-    pybind11::arg("messages"),
-    pybind11::kw_only(),
-    pybind11::arg("template_str") = nullptr,
-    pybind11::arg("tools") = nullptr,
-    pybind11::arg("add_generation_prompt") = true
-)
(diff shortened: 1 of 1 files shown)


src/python/python.cpp
```diff
@@ -358,7 +358,7 @@ PYBIND11_MODULE(onnxruntime_genai, m) {
       })
       .def("to_token_id", &OgaTokenizer::ToTokenId)
       .def("decode", [](const OgaTokenizer& t, pybind11::array_t<int32_t> tokens) -> std::string { return t.Decode(ToSpan(tokens)).p_; })
-      .def("apply_chat_template", [](const OgaTokenizer& t, const char* template_str, const char* messages, const char* tools, bool add_generation_prompt) -> std::string { return t.ApplyChatTemplate(template_str, messages, tools, add_generation_prompt).p_; }, pybind11::arg("template_str") = nullptr, pybind11::arg("messages"), pybind11::arg("tools") = nullptr, pybind11::arg("add_generation_prompt"))
+      .def("apply_chat_template", [](const OgaTokenizer& t, const char* template_str, const char* messages, const char* tools, bool add_generation_prompt) -> std::string { return t.ApplyChatTemplate(template_str, messages, tools, add_generation_prompt).p_; }, pybind11::arg("template_str") = nullptr, pybind11::arg("messages"), pybind11::arg("tools") = nullptr, pybind11::arg("add_generation_prompt") = true)
```
baijumeswani: I think we should reorder the arguments to this function, so the required arguments show up first.

Also, we should mark any non-required argument as a kwarg only.
Not sure if this will work, but you can try it.
```suggestion
      .def(
    "apply_chat_template",
    [](const OgaTokenizer& t,
       const char* messages,
       const char* template_str,
       const char* tools,
       bool add_generation_prompt) -> std::string {
        return t.ApplyChatTemplate(template_str, messages, tools, add_generation_prompt).p_;
    },
    pybind11::arg("messages"),
    pybind11::kw_only(),
    pybind11::arg("template_str") = nullptr,
    pybind11::arg("tools") = nullptr,
    pybind11::arg("add_generation_prompt") = true
)
```

sayanshaw24: yep agreed, this is also closer to what we do in Extensions, I was just following the file norm before, but this is way cleaner, thanks

sayanshaw24: unfortunately, this commit breaks formatting so ran clang format and re-committed but it's definitely less readable as 1 line; we might want to change cpp formatting in .clang-format for the long term, perhaps in a separate PR.

baijumeswani (approved):

PR #1546, opened 2025-06-11
Remove scanning for manylinux directory

ajindal1:
Remove scanning for manylinux directory

### DIFF
--- a/.pipelines/pypl-publishing.yml
+++ b/.pipelines/pypl-publishing.yml
@@ -82,9 +82,6 @@ extends:
       name: 'onnxruntime-Win-CPU-2022'  # Name of your hosted pool
       os: windows  # OS of the image. This value cannot be a variable. Allowed values: windows, linux, macOS
     sdl:
-      sourceRepositoriesToScan:
-        include:
-        - repository: manylinux
       policheck:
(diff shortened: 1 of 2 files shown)


.pipelines/pypl-publishing.yml
```diff
@@ -83,7 +83,7 @@ extends:
       os: windows  # OS of the image. This value cannot be a variable. Allowed values: windows, linux, macOS
     sdl:
       sourceRepositoriesToScan:
-        include:
+        exclude:
```
baijumeswani: Can we remove the two lines? Or is this required?

```
exclude:
- repository: manylinux
```

snnn (dismissed):
We probably do not need to use the repo at all. You can remove it from the top of the yaml file.

.pipelines/pypl-publishing.yml
ajindal1: It is required, otherwise the pipeline cannot be run. But let me try the below suggestion for removing the manylinux repo.

snnn:
Remove 
```
--manylinux-src manylinux --multiple_repos 
```

baijumeswani (approved):

PR #1548, opened 2025-06-12
Always use opset21 in builder

justinchuby:


src/python/py/models/builder.py
```diff
@@ -330,10 +330,7 @@ def __init__(
             inputs=(),
             outputs=(),
             nodes=(),
-            opset_imports={
-                "": 21 if self.quant_attrs["use_qdq"] else 14,
-                "com.microsoft": 1,
-            },
+            opset_imports={"": 21, "com.microsoft": 1},
```
kunal-vaishnavi, line 333: There are other node types that need to be updated to opset 21 such as `LayerNorm`

justinchuby: Are there others besides layer norm?

kunal-vaishnavi: There are too many ops to list here that are in the model builder. `Gelu` is another example as it becomes part of the ONNX standard by opset 21.

You will have to check every op to see if its op schema has changed or not and if the op is now part of the ONNX standard or must still be an ORT contrib op.

justinchuby: I updated gelu. Since the int4 option is using opset 21, I think it is verified that the rest of the ops are used correctly. The only major change is Reduce* ops, which I verified.

For LayerNorm it is more about adopting the newly introduced op. Since there's still discussion on whether the implementation was accurate enough, I propose that we merge this first and adopt LayerNormalization-17 in a separate PR. What do you think?

kunal-vaishnavi: I don't think that's enough verification. I checked all of the ops and here are the ones where I see differences between opset 14 and opset 21.

- [`Cast`](https://onnx.ai/onnx/operators/text_diff_Cast_13_21.html)
- [`DequantizeLinear`](https://onnx.ai/onnx/operators/text_diff_DequantizeLinear_13_21.html)
- [`ReduceMax`](https://onnx.ai/onnx/operators/text_diff_ReduceMax_13_20.html)
- [`ReduceMean`](https://onnx.ai/onnx/operators/text_diff_ReduceMean_13_18.html)
- [`Shape`](https://onnx.ai/onnx/operators/text_diff_Shape_13_21.html)

kunal-vaishnavi: I don't think that's enough verification. I checked all of the ops and here are the ones where I see differences between opset 14 and opset 21.

- [`Cast`](https://onnx.ai/onnx/operators/text_diff_Cast_13_21.html)
- [`DequantizeLinear`](https://onnx.ai/onnx/operators/text_diff_DequantizeLinear_13_21.html)
- [`ReduceMax`](https://onnx.ai/onnx/operators/text_diff_ReduceMax_13_20.html)
- [`ReduceMean`](https://onnx.ai/onnx/operators/text_diff_ReduceMean_13_18.html)
- [`Shape`](https://onnx.ai/onnx/operators/text_diff_Shape_13_21.html)

It should be fine to upgrade to LayerNorm-17 now. The kernel implementation used inside ORT should still be the same whether the op's domain is `com.microsoft` or `ai.onnx`.

justinchuby: Cast - dtype addition only; DequantizeLinear - new attribute "block_size" we don't use here; ReduceMax, ReduceMean - we don't provide axes so the optional input is not needed and kept as default; Shape - new ability to specify "start" and "end" I guess this part can be simplified:

image

I will go ahead and simplify the shapes

kunal-vaishnavi: Axes are provided for `ReduceMean`

https://github.com/microsoft/onnxruntime-genai/blob/f63bf208f3b4f430700476223d50074f23dcbfdc/src/python/py/models/builder.py#L736-L746

Let's simplify any graph structures in a different PR after upgrading to opset 21. There are other subgraphs that could also be updated.

justinchuby: Sound good. I agree that graph structure can be a separate PR. I found quite a few places

justinchuby: > Axes are provided for `ReduceMean`
> 
> https://github.com/microsoft/onnxruntime-genai/blob/f63bf208f3b4f430700476223d50074f23dcbfdc/src/python/py/models/builder.py#L736-L746
> 
> Let's simplify any graph structures in a different PR after upgrading to opset 21. There are other subgraphs that could also be updated.

Yeah this is the correct input right?

src/python/py/models/builder.py
```diff
@@ -390,7 +387,7 @@ def make_attention_init(self):
     def make_genai_config(self, model_name_or_path, extra_kwargs, out_dir):
         try:
             config = GenerationConfig.from_pretrained(model_name_or_path, token=self.hf_token, trust_remote_code=True, **extra_kwargs)
-        except:
+        except Exception:
```
```suggestion
        except:
```

PR #1556, opened 2025-06-14
[Builder] Use lazy tensor

justinchuby:
Use lazy tensor to reduce even more memory footprint. Only needed on the tensors that have torch operations on them. 50GB -> 38GB

image

### DIFF
--- a/src/python/py/models/builder.py
+++ b/src/python/py/models/builder.py
@@ -2519,7 +2519,6 @@ class Model:
             # Norm after last decoder layer of model (last layer --> norm)
             self.layernorm_attrs["last_layernorm"] = True
 
-    @torch.no_grad
     def make_model(self, input_path):
         # Make inputs and outputs to ONNX model
         self.make_inputs_and_outputs()
@@ -3722,7 +3721,7 @@ def set_onnx_dtype(precision: str, extra_options: dict[str, Any]) -> ir.DataType
     }[precision]
 
 
-
+@torch.no_grad
 def create_model(model_name, input_path, output_dir, precision, execution_provider, cache_dir, **extra_options):
     # Create cache and output directories
     os.makedirs(output_dir, exist_ok=True)


src/python/py/models/builder.py
```diff
@@ -2521,7 +2595,7 @@ def make_model(self, input_path):
         else:
             # Load PyTorch model
             extra_kwargs = {"num_hidden_layers": self.num_layers} if "num_hidden_layers" in self.extra_options else {}
-            model = AutoModelForCausalLM.from_pretrained(self.model_name_or_path, cache_dir=self.cache_dir, token=self.hf_token, trust_remote_code=True, **extra_kwargs)
+            model = AutoModelForCausalLM.from_pretrained(self.model_name_or_path, cache_dir=self.cache_dir, token=self.hf_token, trust_remote_code=True, torch_dtype="auto", **extra_kwargs)
```
justinchuby: @kunal-vaishnavi I wonder if it's possible to set torch_dtype properly to reduce some need for casting?

src/python/py/models/builder.py
```diff
@@ -800,7 +825,12 @@ def make_matmul_op(self, matmul, basename, root_input, **kwargs):
 
     def make_matmul_float(self, matmul, name, root_input, **kwargs):
         weight = name[1:].replace("/", ".") + ".weight"
-        self.make_initializer(matmul.weight.T.to(self.to_torch_dtype[self.io_dtype]), weight)
+        self.make_initializer(
```
kunal-vaishnavi: Let's keep the `make_initializer` calls as simple as possible. Something like the following should work.

```python
def make_initializer(self, tensor, name, to = self.io_dtype):
   # Lambdas and casts go inside the `make_initializer` method so that the caller does not have to do that work
   func = lambda: tensor.to(self.to_torch_dtype[to])

   # Getting the shape should be as simple as doing `tensor.shape`
   shape = tensor.shape

   ...
```

Then the call can still remain very simple.

```python
# To use self.io_dtype as the precision
self.make_initializer(matmul.weight.T, weight)

# To use a new dtype as the precision
self.make_initializer(matmul.weight.T, weight, to=new_dtype)
```

src/python/py/models/builder.py
kunal-vaishnavi: Currently, I believe the weights are loaded based on the precision stored in the `.safetensors` files. This should be the same behavior as `torch_dtype="auto"`.

src/python/py/models/builder.py
justinchuby: Great idea - updated and simplified.

src/python/py/models/builder.py
```diff
@@ -38,6 +38,10 @@
 # the ONNX graph efficiently.
 
 
+def transpose_shape(shape: Sequence[int]) -> Sequence[int]:
+    return [*shape[:-2], shape[-1], shape[-2]]
+
+
```
```suggestion
```

src/python/py/models/builder.py
```diff
@@ -3783,11 +3811,12 @@ def create_model(model_name, input_path, output_dir, precision, execution_provid
         else:
             raise NotImplementedError(f"The {hf_name} model is not currently supported.")
 
-        # Make ONNX model
-        onnx_model.make_model(input_path)
+        with torch.no_grad():
```
kunal-vaishnavi: Can we add the `@torch.no_grad` decorator to `create_model` instead? I believe that would load the PyTorch model with no grad as well.

kunal-vaishnavi (approved):

PR #1566, opened 2025-06-17
[Builder] Use packed 4bit tensors directly

justinchuby:
Use ir.PackedTensor to simplify handling 4bit weights of QDQ models. Requires onnx-ir 0.1.2.


justinchuby:
@kunal-vaishnavi this path is currently not tested in ci. Maybe just run a few example option combinations in ci?

PR #1570, opened 2025-06-18
Automatically install java maven artifact in the local maven repository

asoldano:
Fix CMakeLists.txt (thanks @Craigacp) and install java maven artifact in the local maven repository when --publish_java_maven_local flag is used

Fixes https://github.com/microsoft/onnxruntime-genai/issues/1508
Adds an additional change related to https://github.com/microsoft/onnxruntime-genai/pull/1510

(this replaces https://github.com/microsoft/onnxruntime-genai/pull/1531 to deal with @edgchen1 requests and rebase against main)


src/java/CMakeLists.txt
```diff
@@ -233,11 +241,21 @@ if (ENABLE_TESTS)
 
   # Copy ORT native libs for Java tests
   foreach(LIB_FILE ${ort_native_libs})
-    add_custom_command(
-            TARGET onnxruntime-genai-jni POST_BUILD
-            COMMAND ${CMAKE_COMMAND} -E copy
-                    ${LIB_FILE}
-                    ${JAVA_PACKAGE_LIB_DIR}/)
+    if(LIB_FILE MATCHES ".*onnxruntime[^/]*$")
```
edgchen1: would you mind providing an example of the files this is intended to match and not match?

baijumeswani: We should add all shared libraries inside `ORT_LIB_DIR` to the ort-specific directory. This may include libraries such as `libonnxruntime.so` and other ort execution provider specific libraries needed by onnxruntime to execute the model. These libraries may or may not include `onnxruntime` in its name.

From what I understand, `ORT_LIB_DIR` contains only ort specific libraries.

asoldano: Well, I just realised the comment in the else clause is misleading (actually a leftover from some experiments I was doing on a different branch a few weeks ago, sorry about that). The regular expression is meant to match any file whose name has "onnxruntime" and then does not contain a slash after "onnxruntime" until the end of the string. Basically the ONNX Runtime library.
Having said that, I'm going to remove regexp and add things into ORT_PACKAGE_LIB_DIR .

asoldano: (clearly, thanks @baijumeswani and @edgchen1 )

baijumeswani (approved):

edgchen1 (approved):

PR #1571, opened 2025-06-18
Support loading models from memory

baijumeswani:
This pull-request introduces support for loading onnx models from memory instead of from filesystem.

Example usage:

#### Python
```python
model_data = get_model_data(...)

# model_path refers to the directory that contains model files
# The *.onnx files needn't be present in the model_path
# The model_data is used to load the *.onnx model file
config = og.Config(model_path)
config.register_model_data("model.onnx", model_data)

# The model_data is expected to be valid at least until the og.Model has been created
model = og.Model(config)
...
```
#### C++

```cpp
const std::vector<uint8_t> model_data = GetModelData(...);
auto config = OgaConfig::Create(model_path);
config->RegisterModelData("model.onnx", model_data);
...
```
#### C#

```csharp
var modelData = GetModelData(...);
using (var config = new Config(modelPath))
{
    config.RegisterModelData("model.onnx", modelData);
    using (var model = new Model(config))
    {
        ...
    }
}
```
</details>

### DIFF
--- a/src/config.h
+++ b/src/config.h
@@ -257,8 +257,8 @@ struct Config {
   // otherwise returns the nominal name and false
   std::pair<std::string, bool> GetGraphName(const std::string& nominal_name) const;
 
-  std::unordered_map<std::string, std::string> nominal_names_to_graph_names_;      // Mapping of nominal input/output names to graph input/output names
-  std::unordered_map<std::string, std::span<const std::byte>> model_datas_spans_;  // Model bytes to support loading a model from memory
+  std::unordered_map<std::string, std::string> nominal_names_to_graph_names_;     // Mapping of nominal input/output names to graph input/output names
+  std::unordered_map<std::string, std::span<const std::byte>> model_data_spans_;  // Model bytes to support loading a model from memory
 };
 
 void SetSearchNumber(Config::Search& search, std::string_view name, double value);
--- a/src/config.h
+++ b/src/config.h
@@ -258,6 +258,7 @@ struct Config {
   std::pair<std::string, bool> GetGraphName(const std::string& nominal_name) const;
 
   std::unordered_map<std::string, std::string> nominal_names_to_graph_names_;  // Mapping of nominal input/output names to graph input/output names
+  std::unordered_map<std::string, std::span<const uint8_t>> model_datas_;      // Model bytes to support loading a model from memory
 };
 
 void SetSearchNumber(Config::Search& search, std::string_view name, double value);
--- a/src/config.h
+++ b/src/config.h
@@ -257,8 +257,8 @@ struct Config {
   // otherwise returns the nominal name and false
   std::pair<std::string, bool> GetGraphName(const std::string& nominal_name) const;
 
-  std::unordered_map<std::string, std::string> nominal_names_to_graph_names_;  // Mapping of nominal input/output names to graph input/output names
-  std::unordered_map<std::string, std::span<const uint8_t>> model_datas_;      // Model bytes to support loading a model from memory
+  std::unordered_map<std::string, std::string> nominal_names_to_graph_names_;      // Mapping of nominal input/output names to graph input/output names
+  std::unordered_map<std::string, std::span<const std::byte>> model_datas_spans_;  // Model bytes to support loading a model from memory
 };
 
 void SetSearchNumber(Config::Search& search, std::string_view name, double value);
--- a/src/models/model.cpp
+++ b/src/models/model.cpp
@@ -854,10 +854,6 @@ std::unique_ptr<OrtSession> Model::CreateSession(OrtEnv& ort_env, const std::str
     dir_guard.ChangeTo(config_->config_path);
     auto session = OrtSession::Create(ort_env, model_data_it->second.data(), model_data_it->second.size(), session_options);
 
-    // Remove the model data from the map after creating the session
-    // The application must free up the memory used for the model data as it deems fit
-    config_->model_data_spans_.erase(model_data_it);
-
     return session;
   }
 
--- a/src/models/model.cpp
+++ b/src/models/model.cpp
@@ -834,8 +834,8 @@ OrtSessionOptions* Model::GetSessionOptions(const std::string& model_id) const {
 }
 
 std::unique_ptr<OrtSession> Model::CreateSession(OrtEnv& ort_env, const std::string& model_filename, OrtSessionOptions* session_options) {
-  if (auto model_data_it = config_->model_datas_.find(model_filename);
-      model_data_it != config_->model_datas_.end()) {
+  if (auto model_data_it = config_->model_datas_spans_.find(model_filename);
+      model_data_it != config_->model_datas_spans_.end()) {
     // If model data was provided, load the model from memory
     if (model_data_it->second.empty()) {
       throw std::runtime_error("Failed to load model data from memory for " + model_filename);
@@ -846,7 +846,7 @@ std::unique_ptr<OrtSession> Model::CreateSession(OrtEnv& ort_env, const std::str
 
     // Remove the model data from the map after creating the session
     // The application must free up the memory used for the model data as it deems fit
-    config_->model_datas_.erase(model_data_it);
+    config_->model_datas_spans_.erase(model_data_it);
 
     return session;
   }
--- a/src/models/model.cpp
+++ b/src/models/model.cpp
@@ -21,7 +21,50 @@
 #include "decoder_only_pipeline.h"
 #include "../dml/interface.h"
 
+#if defined(_WIN32)
+#include <direct.h>
+#define GETCWD _getcwd
+#define CHDIR _chdir
+#include <windows.h>
+#else
+#include <unistd.h>
+#define GETCWD getcwd
+#define CHDIR chdir
+#include <limits.h>
+#endif
+
 namespace Generators {
+
+namespace {
+
+class DirGuard {
+ private:
+  fs::path original_dir_;
+
+ public:
+  DirGuard() {
+    char buffer[PATH_MAX];
+    if (GETCWD(buffer, sizeof(buffer))) {
+      original_dir_ = fs::path(buffer);
+    } else {
+      throw std::runtime_error("Failed to get current working directory");
+    }
+  }
+
+  void ChangeTo(const fs::path& new_dir) {
+    if (CHDIR(new_dir.c_str()) != 0) {
+      throw std::runtime_error("Failed to change directory to: " + new_dir.string());
+    }
+  }
+
+  ~DirGuard() {
+    if (CHDIR(original_dir_.c_str()) != 0) {
+      Log("warning", "Failed to change back to original directory: " + original_dir_.string());
+    }
+  }
+};
+
+}  // namespace
 
 State::State(const GeneratorParams& params, const Model& model)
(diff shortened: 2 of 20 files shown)


src/models/model.cpp
```diff
@@ -790,6 +833,24 @@ OrtSessionOptions* Model::GetSessionOptions(const std::string& model_id) const {
...
 }
 
+std::unique_ptr<OrtSession> Model::CreateSession(OrtEnv& ort_env, const std::string& model_filename, OrtSessionOptions* session_options) {
+  if (auto model_data_it = config_->model_datas_.find(model_filename);
+      model_data_it != config_->model_datas_.end()) {
+    // If model data was provided, load the model from memory
+    if (model_data_it->second.empty()) {
+      throw std::runtime_error("Failed to load model data from memory for " + model_filename);
+    }
+    DirGuard dir_guard;
```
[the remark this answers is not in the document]
baijumeswani: This part of the code is not expected to be thread safe. It is necessary since the model from memory may contain references to external data files. These datafiles are stored relative to the original model file which is expected to be found in the model config folder. Without the chdir, models with external data cannot be loaded.

Having said that, we could potentially shift the onus of chdir to the application layer and avoid handling that in ort-genai.

src/ort_genai_c.cpp
```diff
@@ -259,6 +259,18 @@ OgaResult* OGA_API_CALL OgaConfigOverlay(OgaConfig* config, const char* json) {
   OGA_CATCH
 }
 
+OgaResult* OGA_API_CALL OgaConfigRegisterModelData(OgaConfig* config, const char* model_filename, const uint8_t* model_data, size_t model_data_length) {
+  OGA_TRY
+  if (model_data == nullptr || model_data_length == 0) {
+    throw std::runtime_error("Expected a valid model data pointer and length. Received nullptr or zero length.");
+  }
+
+  config->model_datas_.emplace(model_filename, std::span<const uint8_t>(model_data, model_data_length));
```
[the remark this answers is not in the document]
baijumeswani: Model data can be large, and I intend to avoid making copies of the data if possible. I could remove the model data from the map once the model is loaded unless there is a need to reload the model or load the model again in another context.

src/models/model.cpp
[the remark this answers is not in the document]
edgchen1: I wonder if we should add a way to specify the working directory in ORT when we don't load from an onnx file?

src/ort_genai_c.h
```diff
@@ -261,6 +261,16 @@ OGA_EXPORT OgaResult* OGA_API_CALL OgaConfigAppendProvider(OgaConfig* config, co
...
 
+/**
+ * \brief Register the model data to load the model from memory.
+ * \param[in] config The config to register the model data on.
+ * \param[in] model_filename The name of the model file as defined in the config.
+ * \param[in] model_data The model data to register. The data is expected to be valid at least until the model is created.
+ * \param[in] model_data_length The length of the model data.
+ * \return OgaResult containing the error message if the registration of the model data failed.
+ */
+OGA_EXPORT OgaResult* OGA_API_CALL OgaConfigRegisterModelData(OgaConfig* config, const char* model_filename, const uint8_t* model_data, size_t model_data_length);
```
edgchen1: `const void* model_data` would give the caller more flexibility. e.g., they could also pass in a `const char*`.
```suggestion
OGA_EXPORT OgaResult* OGA_API_CALL OgaConfigRegisterModelData(OgaConfig* config, const char* model_filename, const void* model_data, size_t model_data_length);
```

src/config.h
```diff
@@ -258,6 +258,7 @@ struct Config {
   std::pair<std::string, bool> GetGraphName(const std::string& nominal_name) const;
 
   std::unordered_map<std::string, std::string> nominal_names_to_graph_names_;  // Mapping of nominal input/output names to graph input/output names
+  std::unordered_map<std::string, std::span<const uint8_t>> model_datas_;      // Model bytes to support loading a model from memory
```
edgchen1: "datas" sounds funny. how about `model_data_spans_`?

src/ort_genai.h
```diff
@@ -179,6 +179,20 @@ struct OgaConfig : OgaAbstract {
     OgaCheckResult(OgaConfigOverlay(this, json));
   }
 
+  void RegisterModelData(const std::string& model_filename, const uint8_t* model_data, size_t model_data_length) {
```
edgchen1: general - instead of uint8_t, consider using std::byte

src/models/model.cpp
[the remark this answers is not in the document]
baijumeswani: I think adding this in ort makes sense. As this seems like a necessity when creating a session from an array with external data in a file.

src/config.h
```diff
@@ -257,7 +257,8 @@ struct Config {
   // otherwise returns the nominal name and false
   std::pair<std::string, bool> GetGraphName(const std::string& nominal_name) const;
 
-  std::unordered_map<std::string, std::string> nominal_names_to_graph_names_;  // Mapping of nominal input/output names to graph input/output names
+  std::unordered_map<std::string, std::string> nominal_names_to_graph_names_;      // Mapping of nominal input/output names to graph input/output names
+  std::unordered_map<std::string, std::span<const std::byte>> model_datas_spans_;  // Model bytes to support loading a model from memory
```
edgchen1: nit: `model_datas_spans_` -> `model_data_spans_`

src/models/model.cpp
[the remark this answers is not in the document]
edgchen1: if we plan to do this, maybe add a comment about it?
maybe chdir is fine as a temporary measure, but I think the original comment has a valid point about thread-safety. we don't know what other user code will be doing and the current directory is global shared state.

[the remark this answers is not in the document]
baijumeswani: I agree that this solution is not ideal. I'll add a comment here highlighting the need to support this in onnx / onnx runtime.

src/models/model.cpp
```diff
@@ -790,6 +833,38 @@ OrtSessionOptions* Model::GetSessionOptions(const std::string& model_id) const {
...
+    // This is a temporary solution and can be potentially addressed by exposing means to set a working directory
+    // for the OrtSession through the ONNX Runtime API.
+    // This solution is not ideal since it modifies the global state of the process, and is hence not thread-safe.
+    DirGuard dir_guard;
+    dir_guard.ChangeTo(config_->config_path);
+    auto session = OrtSession::Create(ort_env, model_data_it->second.data(), model_data_it->second.size(), session_options);
+
+    // Remove the model data from the map after creating the session
+    // The application must free up the memory used for the model data as it deems fit
+    config_->model_data_spans_.erase(model_data_it);
```
edgchen1: I wonder if this is the right place to clean it up.

if the user registers some data for a model that doesn't show up in the config file, it will not get cleaned up here. also, if the same model is loaded twice (perhaps in a pipeline, not sure if that's valid), the data will be cleaned up too early. and is it valid to reuse an OgaConfig instance to create another OgaModel?

src/ort_genai_c.cpp
```diff
@@ -259,6 +259,18 @@ OgaResult* OGA_API_CALL OgaConfigOverlay(OgaConfig* config, const char* json) {
   OGA_CATCH
 }
 
+OgaResult* OGA_API_CALL OgaConfigRegisterModelData(OgaConfig* config, const char* model_filename, const void* model_data, size_t model_data_length) {
+  OGA_TRY
+  if (model_data == nullptr || model_data_length == 0) {
+    throw std::runtime_error("Expected a valid model data pointer and length. Received nullptr or zero length.");
+  }
+
+  config->model_data_spans_.emplace(model_filename, std::span<const std::byte>(static_cast<const std::byte*>(model_data), model_data_length));
```
edgchen1: should we allow it to overwrite an existing entry or complain about that?

src/models/model.cpp
baijumeswani: Maybe at the end of Model creation is ok. The problem is that there are instances where we reload models (in decoder-pipeline) after model the model is created, but this is not common and was only introduced as a workaround at the time.

baijumeswani: > and is it valid to reuse an OgaConfig instance to create another OgaModel?

I think this is a valid usecase. We shouldn't disallow it. Perhaps keeping the span around isn't that bad (it may become dangling at some point when the user frees up the memory). But if we don't use it, it shouldn't matter.

Perhaps the cleanest way is for the user to unregister the model data themselves through an API.

src/ort_genai_c.cpp
baijumeswani: Added a new API to remove model data from the map. Now, the user must remove before adding again. Otherwise, we emit an error message.

src/models/model.cpp
baijumeswani: Added a new API to remove the model data from the map.

src/ort_genai_c.h
```diff
@@ -261,6 +261,25 @@ OGA_EXPORT OgaResult* OGA_API_CALL OgaConfigAppendProvider(OgaConfig* config, co
  */
 OGA_EXPORT OgaResult* OGA_API_CALL OgaConfigSetProviderOption(OgaConfig* config, const char* provider, const char* key, const char* value);
 
+/**
+ * \brief Add the model data to load the model from memory. Applications must call OgaConfigRemoveModelData to remove the model data
+ *        when it is no longer needed.
```
edgchen1, line 266: > Applications must call OgaConfigRemoveModelData to remove the model data when it is no longer needed.

they can also just delete the OrtConfig, right? maybe we don't need to say that removal is required.

src/ort_genai_c.h
```diff
@@ -261,6 +261,25 @@ OGA_EXPORT OgaResult* OGA_API_CALL OgaConfigAppendProvider(OgaConfig* config, co
...
+ * \param[in] config The config to add the model data to.
+ * \param[in] model_filename The name of the model file as defined in the config.
+ * \param[in] model_data The model data to add. The data is expected to be valid at least until the model is created.
+ * \param[in] model_data_length The length of the model data.
+ * \return OgaResult containing the error message if the addition of the model data failed.
+ */
+OGA_EXPORT OgaResult* OGA_API_CALL OgaConfigAddModelData(OgaConfig* config, const char* model_filename, const void* model_data, size_t model_data_length);
+
+/**
+ * \brief Cleanup model data previously added to the config.
```
edgchen1:
```suggestion
 * \brief Remove model data previously added to the config.
```

src/models/model.cpp
```diff
@@ -21,8 +21,51 @@
...
+#define GETCWD getcwd
+#define CHDIR chdir
+#include <limits.h>
+#endif
+
 namespace Generators {
 
+namespace {
+
+class DirGuard {
```
edgchen1, line 40: maybe explicitly delete copy/move ctors/assignment operators for clarity

edgchen1 (dismissed):

GiantSox:
Does this API support loading the weights/tokenizer/vocab/etc from memory too? As far as I can tell everything other than the .onnx still needs to be stored on the filesystem?

I have a use case where the model is never stored on disk in an uncompressed/filesystem-readable format, so I was hoping there'd be some way to load everything directly from memory.

Issue #590, opened 2024-06-10
Memory leak during back-to-back inferences


jeremyfowers:
I am experiencing a memory leak while running my application, which is to run an MMLU accuracy test on my Radeon 780M iGPU via DirectML.

Each inference adds tens-hundreds of megabytes to the total system memory and total graphics memory utilized, until it eventually fills up after about 50 inferences and crashes the system.

## My system

* Ryzen "Pheonix" 7940HS with Radeon 780M iGPU
* 32 GB system memory

## Software

* Phi-3-Mini with awq 4 bit weights
* onnxruntime-genai-directml 0.2.0
* onnxruntime-directml 1.18.0

The model is running on the Radeon 780M iGPU,

## My Code

I define a `generate()` function like this, that is meant to return all the response tokens to the input_ids from a prompt.

```python
def generate(
        input_folder,
        input_ids,
        max_new_tokens=512,
        do_sample=True,
        top_k=50,
        top_p=1.0,
        temperature=0.7,
        pad_token_id=None,
    ):
        model = og.Model(input_folder)
        params = og.GeneratorParams(model)

        if pad_token_id:
            params.pad_token_id = pad_token_id

        max_length = len(input_ids) + max_new_tokens

        params.input_ids = input_ids
        params.set_search_options(
            do_sample=do_sample,
            top_k=top_k,
            top_p=top_p,
            temperature=temperature,
            max_length=max_length,
            min_length=max_length,
        )
        params.try_graph_capture_with_max_batch_size(1)

        generator = og.Generator(model, params)

        prompt_start_time = time.perf_counter()
        generator.compute_logits()
        generator.generate_next_token()
        prompt_end_time = time.perf_counter()

        time_to_first_token = prompt_end_time - prompt_start_time

        if max_new_tokens > 1:

            token_gen_times = []
            while not generator.is_done():
                token_gen_start_time = time.perf_counter()
                generator.compute_logits()
                generator.generate_next_token()
                token_gen_end_time = time.perf_counter()

                token_gen_times.append(token_gen_end_time - token_gen_start_time)

            if token_gen_times:
                # List will be empty if we generated 1 or 0 tokens, and we don't
                # want a divide-by-zero error in those cases
                avg_token_gen_latency_s = sum(token_gen_times) / len(
                    token_gen_times
                )
                tokens_per_second = 1 / avg_token_gen_latency_s

        return [generator.get_sequence(0)]
```

[...]

PR #1578, opened 2025-06-21
Fix DML Memory Leak

aciddelgado:
A memory leak where DML would take up more and more memory was found. This should fix it by enabling memory arena.

### DIFF
--- a/src/tensor.cpp
+++ b/src/tensor.cpp
@@ -16,6 +16,7 @@ Tensor::Tensor(std::unique_ptr<OrtValue> ort_tensor) : ort_tensor_{std::move(ort
 Tensor::~Tensor() {
   if (buffer_ != nullptr) {
     p_device_->GetAllocator().Free(buffer_);
+    buffer_ = nullptr;
   }
 }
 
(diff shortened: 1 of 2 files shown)


src/tensor.cpp
```diff
@@ -16,6 +16,7 @@ Tensor::Tensor(std::unique_ptr<OrtValue> ort_tensor) : ort_tensor_{std::move(ort
 Tensor::~Tensor() {
   if (buffer_ != nullptr) {
     p_device_->GetAllocator().Free(buffer_);
+    buffer_ = nullptr;
```
baijumeswani, line 19: Is this required?

aciddelgado: I think it's good to make it nullptr

sheller-ms:
Can you elaborate how enabling the memory arena fixes the issue?
Back in Jun of 2024 a PR was submitted to disable the memory arena to fix a (different) memory leak

[https://github.com/microsoft/onnxruntime-genai/pull/580 ](https://github.com/microsoft/onnxruntime-genai/pull/580 )
[https://github.com/microsoft/onnxruntime-genai/issues/590 ](https://github.com/microsoft/onnxruntime-genai/issues/590 )

PR #1579, opened 2025-06-21
Bump onnx-ir to 0.1.2

jiafatom:
Currently it seems that the command "python builder.py -i" in model builder is not working anymore, because there is an API change that it uses "callback" that onnx-ir 0.1.1 does not have
 ir.save(
model,
out_path,
external_data=os.path.basename(data_path),
size_threshold_bytes=0,
callback=callback,
)

### DIFF
--- a/src/python/py/models/builder.py
+++ b/src/python/py/models/builder.py
@@ -553,6 +553,7 @@ class Model:
             ir.save(
                 model,
                 out_path,
+                format=None,
                 external_data=os.path.basename(data_path),
                 size_threshold_bytes=0,
                 callback=callback,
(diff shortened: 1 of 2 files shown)


justinchuby:
What does this fix?

jiafatom:
> What does this fix?

Currently it seems that the command "python builder.py -i" in model builder is not working anymore, because   (1)          ir.save(
                model,
                out_path,
                external_data=os.path.basename(data_path),
                size_threshold_bytes=0,
                callback=callback,
            )   does not the argument "format", and it uses "callback" that onnx-ir 0.1.1 does not have

PR #1580, opened 2025-06-23
Support for Continuous Batching

baijumeswani:
Batching is a key technique for improving throughput in large language models by allowing multiple requests to be processed simultaneously. By grouping requests, models can leverage parallel processing to handle more queries per unit of time.

Currently, `onnxruntime-genai` performs inference using **static batching** ([see explanation of static vs. dynamic batching here](https://www.anyscale.com/blog/continuous-batching-llm-inference#llm-batching-explained)). In this mode, the batch state is tightly coupled to the `Generator`, which restricts the ability to process multiple requests dynamically or concurrently.

To enable **dynamic batching** in `onnxruntime-genai`, this pull-request introduces several key concepts that decouple the batch state from the generation engine, paving the way for more efficient request handling.

#### OgaRequest

The `OgaRequest` encapsulates the state of the application request. It tracks the progress of the request, including the input tokens, generation parameters, and the current state of the generation process. Requests are processed concurrently by the Engine, which dynamically batches them for efficient model execution.

#### OgaEngine

The `OgaEngine` manages concurrent requests. It is responsible for scheduling them and dynamically batching them in addition to running the core generation logic (model execution).

#### Python

```python

model = og.Model(config)
engine = og.Engine(model)

streaming_tokenizer = tokenizer.create_stream()

request = og.Request(params)
request.add_tokens(tokenizer.encode(tokenizer.apply_chat_template(...)))

engine.add_request(request)

while ready_request := engine.step():
    while ready_request.has_unseen_tokens():
        print(streaming_tokenizer.decode(ready_request.get_unseen_token()), end="", flush=True)

engine.remove_request(request)
```

#### C++

```cpp
#include "ort_genai.h"

auto model = OgaModel::Create(PHI2_PATH);
auto engine = OgaEngine::Create(*model);

auto streaming_tokenizer = OgaTokenizerStream::Create(*tokenizer);

const std::string prompt = GetPrompt(...);
auto sequence = OgaSequences::Create();
tokenizer->Encode(prompt .c_str(), *sequence);

auto request = OgaRequest::Create(*params);
request->AddTokens(*sequence);

engine->Add(*request);

while (auto ready_request = engine->Step()) {
    while (ready_request->HasUnseenTokens()) {
        std::cout << streaming_tokenizer->Decode(ready_request->GetUnseenToken());
    }
}

engine->Remove(*request);
```

### What About the OgaGenerator?

This pull-request decouples the batch state from the generation engine, addressing one of the main limitations of the `OgaGenerator` which is its tight coupling with the batch state. This design constraint restricted the `OgaGenerator` to static batching only. With the introduction of `OgaEngine` and `OgaRequest`, the need for `OgaGenerator` in this role is effectively eliminated.

### OgaEngine v/s. OgaGenerator

| Feature | `OgaEngine` | `OgaGenerator` |
|----------|----------|----------|
| Continuous Static Batching    | ✅ | ❌|
| Continuous Dynamic Batching    | ⏳ | ❌|
| Language Models    | ✅     | ✅ |
| Other Model Types    | ⏳     | ✅ |
| Beam Search    | ⏳    | ✅    |
| Adapters Support    | ⏳ | ✅ |
| DML Support   | ❌ | ✅  |
| Cuda, CPU, WebGPU Support   | ✅ | ✅ |
| Other EPs  | ⏳    | ✅    |


src/config.h
```diff
@@ -23,6 +23,12 @@ struct Config {
     static constexpr std::string_view PresentValueName = "present.%d.value";
     static constexpr std::string_view RnnStatesName = "rnn_states";
     static constexpr std::string_view RnnStatesPrevName = "rnn_states_prev";
+    static constexpr std::string_view CumulativeSequenceLengths = "cumulative_sequence_lengths";
+    static constexpr std::string_view SequenceLengths = "sequence_lengths";
+    static constexpr std::string_view MaxQueryLength = "max_query_length";
```
aciddelgado: max lengths and slot_mapping should be no longer needed. sequence_lengths may be renamed to past_sequence_lengths

src/config.h
```diff
@@ -210,6 +215,12 @@ struct Config {
         std::string encoder_hidden_states{Defaults::EncoderHiddenStatesName};
         std::string rnn_prev_states{Defaults::RnnStatesPrevName};
         std::string encoder_attention_mask{Defaults::EncoderAttentionMaskName};
+        std::string cumulative_sequence_lengths{Defaults::CumulativeSequenceLengths};
+        std::string sequence_lengths{Defaults::SequenceLengths};
+        std::string max_query_length{Defaults::MaxQueryLength};
```
aciddelgado: As above, remove the unnecessary ones

src/engine/cache_manager.cpp
```diff
@@ -0,0 +1,96 @@
...
+  if (!key_value_cache_) {
+    auto request_with_max_total_sequence_length =
+        std::max_element(
+            requests.begin(), requests.end(),
+            [](const std::shared_ptr<Request>& a, const std::shared_ptr<Request>& b) {
+              return a->Params()->search.max_length < b->Params()->search.max_length;
+            });
+
+    params_ = std::make_shared<GeneratorParams>(*model_);
+    params_->search.max_length = (*request_with_max_total_sequence_length)->Params()->search.max_length;
```
aciddelgado: nit: more accurate to say max_max_sequence_length

src/engine/decoders/decoder.h
```diff
@@ -0,0 +1,48 @@
...
+namespace Generators {
+
+struct ModelIO : State {
+  ModelIO(const GeneratorParams& params, const Model& model) : State(params, model) {}
+
+  DeviceSpan<float> Run(int total_length, DeviceSpan<int32_t>& next_tokens, DeviceSpan<int32_t> next_indices = {}) override {
+    return {};
+  }
+
+  void RewindTo(size_t index) override { (void)index; };
```
aciddelgado: Perhaps these functions should return an error rather than be non-op. In the future we could abstract out only the necessary parts of the State struct and inherit from that without these extra functions.

src/engine/model_io.h
```diff
@@ -0,0 +1,38 @@
...
+struct DecoderWithPagedAttention : ModelIO {
+  DecoderWithPagedAttention(std::shared_ptr<DecoderOnly_Model> model,
+                            ScheduledRequests& scheduled_requests)
+      : model_{model},
+        scheduled_requests_{scheduled_requests} {
+    input_names = {
+        model_->config_->model.decoder.inputs.input_ids.c_str(),
+        model_->config_->model.decoder.inputs.position_ids.c_str(),
+        model_->config_->model.decoder.inputs.cumulative_sequence_lengths.c_str(),
+        model_->config_->model.decoder.inputs.sequence_lengths.c_str(),
```
aciddelgado: Update inputs here too

src/engine/request.h
```diff
@@ -0,0 +1,153 @@
...
+  RequestStatus status_{RequestStatus::Unassigned};
+
+  /**
+   * @brief Retrieves the generator parameters associated with this request.
+   * @return Shared pointer to GeneratorParams.
+   */
+  std::shared_ptr<GeneratorParams> Params();
+
+  /**
+   * @brief Sets the opaque data for user-defined purposes.
```
aciddelgado, line 130: Personally would appreciate the comment to describe what is meant by opaque

src/engine/scheduler.cpp
```diff
@@ -0,0 +1,91 @@
...
+
+ScheduledRequests Scheduler::Schedule() {
+  std::vector<std::shared_ptr<Request>> requests_to_schedule;
+  for (auto& request : requests_pool_) {
+    if (request->status_ == RequestStatus::Assigned) {
+      requests_to_schedule.push_back(request);
+    }
+  }
+
+  if (cache_manager_->SupportsDynamicBatching()) {
```
aciddelgado: Maybe it's a good idea to remove these paths since they aren't implemented yet? Not sure though

src/ort_genai.h
```diff
@@ -706,6 +706,74 @@ struct OgaAdapters : OgaAbstract {
   static void operator delete(void* p) { OgaDestroyAdapters(reinterpret_cast<OgaAdapters*>(p)); }
 };
 
+struct OgaRequest : OgaAbstract {
+  static std::unique_ptr<OgaRequest> Create(OgaSequences& tokens, OgaGeneratorParams& params) {
```
aciddelgado: Consider whether we want tokens to be in the creation of ogarequest or in a separate function call like appendtokens

src/engine/model_io.h
baijumeswani: Since this struct is not used in this PR, I'll remove it entirely for now.

src/ort_genai.h
baijumeswani: ok makes sense. let's keep the api so that it doesn't need to pass in the input ids during construction.

aciddelgado (approved):

PR #1587, opened 2025-06-26
[Builder] use dtype conversion helpers from onnx_ir

justinchuby:
Simplify builder with dtype conversion helpers for torch. Also bump onnx-ir to 0.1.3

### DIFF
--- a/src/python/py/models/builder.py
+++ b/src/python/py/models/builder.py
@@ -33,10 +33,6 @@ from transformers import (
     GenerationConfig,
 )
 
-# NOTE: Avoid importing from onnx helper and numpy_helper. Instead, leverage
-# ONNX IR methods like ir.tensor, ir.DataType.numpy() and other methods for constructing
-# the ONNX graph efficiently.
-
 
 class Model:
     def __init__(
--- a/src/python/py/models/builder.py
+++ b/src/python/py/models/builder.py
@@ -16,23 +16,23 @@ import os
 import textwrap
 from typing import Any, Literal, Sequence
 
-from tqdm import tqdm
 import numpy as np
 import onnx_ir as ir
-from onnx_ir import tensor_adapters
 import torch
+from onnx_ir import tensor_adapters
+from onnx_ir.tensor_adapters import from_torch_dtype, to_torch_dtype
 from onnxruntime.quantization.matmul_nbits_quantizer import (
     MatMulNBitsQuantizer,
     QuantFormat,
     RTNWeightOnlyQuantConfig,
 )
+from tqdm import tqdm
 from transformers import (
     AutoConfig,
     AutoModelForCausalLM,
     AutoTokenizer,
     GenerationConfig,
 )
-
 
 # NOTE: Avoid importing from onnx helper and numpy_helper. Instead, leverage
 # ONNX IR methods like ir.tensor, ir.DataType.numpy() and other methods for constructing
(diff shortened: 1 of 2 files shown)


src/python/py/models/builder.py
```diff
@@ -16,24 +16,24 @@
 import textwrap
 from typing import Any, Literal, Sequence
 
-from tqdm import tqdm
 import numpy as np
 import onnx_ir as ir
-from onnx_ir import tensor_adapters
 import torch
+from onnx_ir import tensor_adapters
```
kunal-vaishnavi, line 22: Is there a way to consolidate the ONNX IR imports (similar to how the `transformers` imports are consolidated to `from transformers import XYZ`)?

justinchuby: I cleaned up the imports. Usually it is better to import modules only: https://gist.github.com/justinchuby/9085242a53158f2fd7ae7aa650e55ee3#3-import-only-modules

kunal-vaishnavi: This is not always true. If you do this for `transformers` for example, you will significantly degrade performance in your application because there are too many modules to import with `import transformers`.

src/python/py/models/builder.py
```diff
@@ -76,6 +75,7 @@ def __init__(
         self.model_name_or_path = config._name_or_path
         self.model_type = config.architectures[0]
         self.io_dtype: ir.DataType = ir.DataType(io_dtype)
+        self.io_dtype_torch: torch.dtype = tensor_adapters.to_torch_dtype(self.io_dtype)
```
kunal-vaishnavi: The original implementation with using `from onnx_ir.tensor_adapters import from_torch_dtype, to_torch_dtype` was better. Let's revert back to it. I was just looking to see if there's a way to remove `import onnx_ir as ir` and use `from onnx_ir import XYZ`.

src/python/py/models/builder.py
justinchuby: Yeah unless for performance reasons of course. This is not the case for onnx_ir modules.

- And since importing is a one time cost, if we know we are going to import it at least once we might as well take the cost up front for readability. It's usually a good trade off to favor readability and maintainability over slight impact in performance when the performance is not the bottleneck.

src/python/py/models/builder.py
justinchuby: I would gently push back on idea for the reasons explained. `import onnx_ir as ir` has been a conventional and a clean way to use ir methods. Otherwise we can create quite some ambiguities given there is a lot of similar concepts defined in onnxruntime packages etc. which could harm readability.

justinchuby:
I updated the call sites and removed the io_dtype_torch field, thanks!

src/python/py/models/builder.py
```diff
@@ -16,24 +16,23 @@
...
+from tqdm import tqdm
 from transformers import (
     AutoConfig,
     AutoModelForCausalLM,
     AutoTokenizer,
     GenerationConfig,
 )
 
-
 # NOTE: Avoid importing from onnx helper and numpy_helper. Instead, leverage
```
kunal-vaishnavi: Now that the ONNX IR methods are used everywhere, can this comment be removed?

justinchuby: Was there as a reminder - but if you like we can remove and let reviewers guard this?

kunal-vaishnavi: I think we can remove it and let reviewers guard this.

PR #1594, opened 2025-06-30
NvTensorRTRTx: Enable CUDA graph via config and fix attention_mask shape handling

anujj:
- Add option to enable CUDA graph for NvTensorRTRTx EP through provider config.
- Fix handling of attention_mask shapes when `enable_cuda_graph` is false for NvTensorRTRTx:
    - When `past_present_share_buffer` is enabled, NvTensorRTRTx expects attention_mask shape as `[b, nh, max_seq_len, hd]` with masking applied. Previously, these shapes were only sent when both `past_present_share_buffer` and graph capture were enabled. This PR ensures the correct shape is passed to TRT for in-place KV cache, aligning with expected behavior.

### DIFF
--- a/src/models/position_inputs.cpp
+++ b/src/models/position_inputs.cpp
@@ -141,7 +141,9 @@ void DefaultPositionInputs::UpdatePositionIDs(int total_length, int new_kv_lengt
 }
 
 void DefaultPositionInputs::CreateNextAttentionMaskTensor(int total_length) {
-  if (state_.params_->use_graph_capture)
+  if (state_.params_->use_graph_capture ||
+    (state_.params_->search.past_present_share_buffer &&
+    model_.p_device_->GetType() == DeviceType::NvTensorRtRtx))
     return;
   attention_mask_shape_[1] = total_length;
   attention_mask_next_->CreateTensor(attention_mask_shape_);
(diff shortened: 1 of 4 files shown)


anujj:
adding @kunal-vaishnavi for review

src/models/position_inputs.cpp
```diff
@@ -141,7 +141,9 @@ void DefaultPositionInputs::UpdatePositionIDs(int total_length, int new_kv_lengt
 }
 
 void DefaultPositionInputs::CreateNextAttentionMaskTensor(int total_length) {
-  if (state_.params_->use_graph_capture)
+  if (state_.params_->use_graph_capture ||
+      (state_.params_->search.past_present_share_buffer &&
+       model_.p_device_->GetType() == DeviceType::NvTensorRtRtx))
```
baijumeswani: How is past present buffer sharing related to the shape of the attention mask? In the description, you mention that the shape of the attention mask is expected to be `[b, nh, max_seq_len, hd]`. But from what I am reading in this code, the shape is initialized to `[b, seq_len]` regardless of graph capture. What is the expected shape of the attention mask for nvtensorrtrtx ep model?

Could you abstract this into a function of its own? It seems to be used multiple times.

anujj: Updated the description to show the correct expected shapes for attention_mask when in-place KV is enabled for  NvTensorRTRtx ep. The previous text mistakenly listed the shapes for past_present_share_buffer.

I’ve also added the separate helper function requested in the review in recent commit.

anujj:
Thanks @baijumeswani . We will avoid device specific check going forward. 
Can we merge this PR

PR #1596, opened 2025-07-01
Correct model type as expected by genai CreateModel function for llm …

anujj:
Problem
create_model() in src/models/model.cpp (line [899](https://github.com/microsoft/onnxruntime-genai/blob/main/src/models/model.cpp#L899)) expects the model_type field to be "gemma3_text" when loading a Gemma3ForConditionalGeneration model.
During export, builder.py writes only the top-level family name—"gemma3"—into genai_config.json.
This mismatch prevents onnxruntime-genai from instantiating the model at runtime.

Fix
Update builder.py so that, when exporting a Gemma 3 checkpoint, it writes the correct subtype "gemma3_text" to genai_config.json. The generated config now aligns with create_model(), allowing Gemma 3 LLMs to load and run without manual edits.

Added @baijumeswani to review the changes

### DIFF
--- a/src/python/py/models/builder.py
+++ b/src/python/py/models/builder.py
@@ -3762,6 +3762,7 @@ def create_model(model_name, input_path, output_dir, precision, execution_provid
                 print("WARNING: This model loses accuracy with float16 precision. Setting `--precision bf16` by default.")
                 onnx_dtype = ir.DataType.BFLOAT16
             onnx_model = Gemma3Model(config, io_dtype, onnx_dtype, execution_provider, cache_dir, extra_options)
+            onnx_model.model_type = "gemma3_text"
         elif config.architectures[0] == "GraniteForCausalLM":
             onnx_model = GraniteModel(config, io_dtype, onnx_dtype, execution_provider, cache_dir, extra_options)
         elif config.architectures[0] == "LlamaForCausalLM":


kunal-vaishnavi:
This is not a mistake actually. The Gemma-3 1B model is a text-only model and uses the [`Gemma3ForCausalLM`](https://huggingface.co/google/gemma-3-1b-it/blob/main/config.json#L3) architecture name. In the `genai_config.json` file, the model type is listed as `gemma3_text` to match the Hugging Face name.

The other models in the Gemma-3 family (4B, 12B, 27B) are vision-language models (VLMs) and use the [`Gemma3ForConditionalGeneration`](https://huggingface.co/google/gemma-3-4b-it/blob/main/config.json#L3) architecture name. In the `genai_config.json` file, the model type is listed as `gemma3` to match the Hugging Face name.

The model builder is used to create the text components for the VLMs. Similar to the Phi vision and Phi multimodal releases, there are [additional scripts](https://github.com/microsoft/onnxruntime-genai/blob/025c24554f538db20f940a7086ab9b6e2c7f9b2c/examples/python/gemma-3-vision-tutorial.md?plain=1#L1) to generate the remaining assets needed to run the Gemma-3 VLMs. Once the models pass Microsoft's responsible AI evaluations, we will make them public.

kunal-vaishnavi:
For VLMs whose text components are created by the model builder, we have kept the model type as the VLM's type because the generated text model cannot run standalone. The `exclude_embeds` flag is turned on in those instances to ensure that `inputs_embeds` is the input and not `input_ids`.

This was previously set for Gemma-3 VLMs.

https://github.com/microsoft/onnxruntime-genai/blob/119dfc7e0dc68dd461b27592a142d099ede8f8d6/src/python/py/models/builder.py#L3367-L3374

However, it seems to have been accidentally removed later on.

https://github.com/microsoft/onnxruntime-genai/blob/6b99e0d73976938b0ff25ed940afdc9d3260e99b/src/python/py/models/builder.py#L3755-L3764

This explains why you are able to run the text component of the Gemma-3 VLMs.

This appears to be a regression as the `exclude_embeds` flag does need to be added back and set to true. It is expected that users should run the special `builder.py` file built for a VLM to get the components needed to run with ONNX Runtime GenAI. The special `builder.py` file will still allow you to run text-only inference if desired.

anujj:
okey thanks . I can see the PR also that enable exclude_embeds : https://github.com/microsoft/onnxruntime-genai/pull/1601/files

PR #1614, opened 2025-07-09
NvTensorRtRtx: Pass the dynamic shapes (ISL and batch_size) to the ep at runtime as nv profile. 

anujj:
1. Pass the dynamic shapes (ISL and `batch_size`) to the NV EP.  
2. Update the C++ benchmark:  
   - For the `NvTensorRtRTx` EP,  need to pass `batch_size` at `OgaModel::Create` time, as it is required to build the engine.

### DIFF
--- a/src/runtime_settings.cpp
+++ b/src/runtime_settings.cpp
@@ -25,18 +25,38 @@ std::string RuntimeSettings::GenerateConfigOverlay() const {
 }
 )";
 
-  auto it = handles_.find("dawnProcTable");
-  if (it != handles_.end()) {
-    void* dawn_proc_table_handle = it->second;
-    std::string overlay;
-    overlay.reserve(webgpu_overlay_pre.size() + webgpu_overlay_post.size() + 20);  // Optional small optimization of buffer size
-    overlay += webgpu_overlay_pre;
-    overlay += std::to_string((size_t)(dawn_proc_table_handle));
-    overlay += webgpu_overlay_post;
-    return overlay;
-  }
+    constexpr std::string_view batch_size_overlay = R"({
+  "search": {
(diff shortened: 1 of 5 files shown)


anujj:
@baijumeswani @kunal-vaishnavi : could you please review the changes

src/runtime_settings.cpp
```diff
@@ -25,17 +25,37 @@ std::string RuntimeSettings::GenerateConfigOverlay() const {
 }
 )";
 
-  auto it = handles_.find("dawnProcTable");
-  if (it != handles_.end()) {
-    void* dawn_proc_table_handle = it->second;
+  constexpr std::string_view batch_size_overlay = R"({
```
baijumeswani: Instead of using the runtime settings, would using something like [`OgaConfig::Overlay`](https://github.com/microsoft/onnxruntime-genai/blob/37ec6967ad24e67b42799f149cce4ea5dd570334/src/ort_genai_c.h#L289-L295) help?

src/models/whisper.cpp
```diff
@@ -342,8 +342,6 @@ DeviceSpan<float> WhisperState::Run(int current_length, DeviceSpan<int32_t>& nex
     first_run_ = false;
     return logits;
   }
-  // Not reached
-  return {};
```
kunal-vaishnavi: Can you sync with the main branch? This snippet should already be removed.

anujj: done

src/runtime_settings.cpp
anujj: yeas, switched to Overlay

anujj:
Addressed the review comments

Issue #776, opened 2024-08-09
generator.get_next_token always returns zero


SreenilaRajesh:
generator.get_next_token always returns zero

generator.get_output(“logits”).squeeze also returns an array of zeros

this results in a blank Output everytime
————————-

model used : phi3-mini-128k-instruct-cpu-int4-rtn-block-32.onnx

cc @kunal-vaishnavi

yufenglee:
@SreenilaRajesh , could you please share more information to repro, like the script?

SreenilaRajesh:
@yufenglee @kunal-vaishnavi 

> @SreenilaRajesh , could you please share more information to repro, like the script?
I’m trying the below script with an offline model, passing the model folder in the command line (-m)
**Script** : https://github.com/microsoft/onnxruntime-genai/blob/main/examples/python/phi3-qa.py

**Model** : phi3-mini-128k-instruct-cpu-int4-rtn-block-32.onnx
https://huggingface.co/microsoft/Phi-3-mini-128k-instruct-onnx/tree/main/cpu_and_mobile/cpu-int4-rtn-block-32

The weight matrices seem to be empty when visualizing the model with https://netron.app/


> In addition to a repro script, what was the prompt that you tried and what version of ONNX Runtime GenAI are you using?
 testing with simple prompts like “tell me a joke”, “tell a short story”
ONNX Runtime GenAI - 0.3.0

kunal-vaishnavi:
> I’m trying the below script with an offline model, passing the model folder in the command line (-m)
> **Script** : https://github.com/microsoft/onnxruntime-genai/blob/main/examples/python/phi3-qa.py
> **Model** : phi3-mini-128k-instruct-cpu-int4-rtn-block-32.onnx
> https://huggingface.co/microsoft/Phi-3-mini-128k-instruct-onnx/tree/main/cpu_and_mobile/cpu-int4-rtn-block-32
> testing with simple prompts like “tell me a joke”, “tell a short story”

I am able to use the above links to run the model and get outputs with your prompts with ONNX Runtime GenAI v0.3.0.

```
$ python3 phi3-qa.py -m ./phi3_mini_128k_cpu_int4_rtn_block_32
Input: tell me a joke

Output:  Why don't scientists trust atoms?

Because they make up everything!

Input: tell a short story

[...]

PR #1615, opened 2025-07-09
Make OnnxRuntime GenAI Examples Simpler

baijumeswani:
Changes in this pull-request:

- Build the examples by default through `python build.py`
- Make the instructions simpler
- Make the unit tests discoverable.


examples/c/src/phi3v.cpp
```diff
@@ -31,7 +31,7 @@ void CXX_API(const char* model_path, const char* execution_provider) {
     std::vector<std::string> image_paths;
     for (size_t start = 0, end = 0; end < image_paths_str.size(); start = end + 1) {
```
kunal-vaishnavi, line 32: Can we change this example to `model_vision.cpp`? We have a similar change done for `model-vision.py`.

examples/c/README.md
```diff
@@ -1,362 +1,64 @@
...
-```
-
-#### Run the sample
-
-```bash
-cd build/Release
-./phi3v <path_to_model> <execution_provider>
-```
+> 📝 **Note:** The examples from the main branch of this repository are compatible with the binaries build from the same commit. Therefore, if using the example from `main`, ONNX Runtime GenAI needs to be built from source. If this is your scenario, just build the library and the examples will be auto built along with the library.
+If this is not your scenario, please use prebuilt binaries from the release you're interested in and use the exmaples from the same version tag and follow the steps below.
```
kunal-vaishnavi:
```suggestion
If this is not your scenario, please use prebuilt binaries from the release you're interested in and use the examples from the same version tag and follow the steps below.
```

examples/c/README.md
```diff
@@ -1,362 +1,64 @@
...
+
+## Download the model
+
+1. Download and install [foundry-local](https://github.com/microsoft/Foundry-Local/releases)
+2. List available models: `foundry model list`
+3. Download a model you would like to run. For example: `foundry model download Phi-4-generic-cpu`
+4. Find out where the model is saved on disk: `foundry cache location`
+5. Identify the path to the model on disk. For example: `C:\Users\<user>\.foundry\Microsoft\Phi-4-generic-cpu\cpu-int4-rtn-block-32-acc-level-4`
+
+> 📝 **Note:** FoundryLocal cli is not available on Linux at the moment. Please download the model from a Windows machine and copy it over to your Linux machine if you would like to run on Linux.
```
kunal-vaishnavi:
```suggestion
> 📝 **Note:** Foundry Local CLI is not available on Linux at the moment. Please download the model from a Windows machine and copy it over to your Linux machine if you would like to run on Linux.
```

examples/c/README.md
```diff
@@ -1,362 +1,64 @@
...
-#### Run the sample
-
-```bash
-cd build/Release
-./phi3v <path_to_model> <execution_provider>
-```
+> 📝 **Note:** The examples from the main branch of this repository are compatible with the binaries build from the same commit. Therefore, if using the example from `main`, ONNX Runtime GenAI needs to be built from source. If this is your scenario, just build the library and the examples will be auto built along with the library.
+If this is not your scenario, please use prebuilt binaries from the release you're interested in and use the exmaples from the same version tag and follow the steps below.
+
+## Download the model
```
kunal-vaishnavi, line 6: Can we also list our already-published Hugging Face repos as other places to download the assets?

examples/c/README.md
```diff
@@ -1,362 +1,64 @@
...
+
+## Build the C++ Example
+
+1. Clone the repo: `git clone https://github.com/microsoft/onnxruntime-genai.git`
+   - Use the relevant release tag that aligns with the version of the libraries you're planning to use.
+2. `cd onnxruntime-genai`
+3. Download the ONNX Runtime libraries.
+   - Depending on the execution provider of interest, download one of the following:
+      - [CPU execution provider](https://www.nuget.org/packages/Microsoft.ML.OnnxRuntime)
+      - [Cuda execution provider on Windows](https://www.nuget.org/packages/Microsoft.ML.OnnxRuntime.Gpu.Windows)
```
kunal-vaishnavi:
```suggestion
      - [CUDA execution provider on Windows](https://www.nuget.org/packages/Microsoft.ML.OnnxRuntime.Gpu.Windows)
```

examples/c/README.md
```diff
@@ -1,362 +1,64 @@
...
+      - [Cuda execution provider on Linux](https://www.nuget.org/packages/Microsoft.ML.OnnxRuntime.Gpu.Linux)
```
kunal-vaishnavi:
```suggestion
      - [CUDA execution provider on Linux](https://www.nuget.org/packages/Microsoft.ML.OnnxRuntime.Gpu.Linux)
```

examples/c/README.md
```diff
@@ -1,362 +1,64 @@
...
+      - [QNN execution provider](https://www.nuget.org/packages/Microsoft.ML.OnnxRuntime.QNN)
+      - [DirectML execution provider](https://www.nuget.org/packages/Microsoft.ML.OnnxRuntime.DirectML)
+   - Rename extension from `*.nupkg` to `*.zip`.
+   - Extract the zip and copy over all the files from `<package>\runtimes\<platform>\native` to `examples\c\lib`
+4. Download the [ONNX Runtime GenAI libraries](https://github.com/microsoft/onnxruntime-genai/releases).
+   - Depending on the execution provider of interest, download one of the following:
+      - CPU / QNN execution provider: `onnxruntime-genai-<version>-<platform>.zip/tar.gz`
+      - Cuda execution provider: `onnxruntime-genai-<version>-<platform>-cuda.zip/tar.gz`
```
kunal-vaishnavi:
```suggestion
      - CUDA execution provider: `onnxruntime-genai-<version>-<platform>-cuda.zip/tar.gz`
```

examples/c/README.md
```diff
@@ -1,362 +1,64 @@
...
+      - DirectML execution provider: `onnxruntime-genai-<version>-<platform>-dml.zip/tar.gz`
+   - Extract the zip and copy over all the files from `<package>\lib` to `examples\c\lib`
+   - Copy over the header files from `<package>\include` to `examples\c\include`
+5. Build using cmake
```
kunal-vaishnavi, line 37: What about macOS builds?

kunal-vaishnavi, line 37: Can we show build steps for macOS?

examples/c/src/model_chat.cpp
```diff
@@ -87,18 +104,15 @@ void CXX_API(const char* model_path, const char* execution_provider) {
         std::cout << tokenizer_stream->Decode(new_token) << std::flush;
       }
     } catch (const std::exception& e) {
-      std::cout << "Session Terminated: " << e.what() << std::endl;
+      std::cout << "\n\033[31mTerminating generation: " << e.what() << "\033[0m" << std::endl;
```
kunal-vaishnavi, line 107: ANSI tags don't display well on all OS platforms. Can you enable it for Windows and disable for Linux?

kunal-vaishnavi, line 107: ANSI tags don't display well on all OS platforms. Can you enable/disable it with a flag?

examples/c/src/model_chat.cpp
```diff
@@ -113,12 +127,16 @@ int main(int argc, char** argv) {
   // Responsible for cleaning up the library during shutdown
   OgaHandle handle;
 
-  std::cout << "-------------" << std::endl;
-  std::cout << "Hello, Phi-3!" << std::endl;
-  std::cout << "-------------" << std::endl;
+  std::cout << "---------------------------" << std::endl;
+  std::cout << "Hello, OrtGenAI Model Chat!" << std::endl;
```
kunal-vaishnavi:
```suggestion
  std::cout << "Hello, ORT GenAI Model Chat!" << std::endl;
```

examples/c/src/model_qa.cpp
```diff
@@ -102,12 +119,17 @@ int main(int argc, char** argv) {
   // Responsible for cleaning up the library during shutdown
   OgaHandle handle;
 
-  std::cout << "-------------" << std::endl;
-  std::cout << "Hello, Phi-3!" << std::endl;
-  std::cout << "-------------" << std::endl;
+  std::cout << "-------------------------" << std::endl;
+  std::cout << "Hello, OrtGenAI Model-QA!" << std::endl;
```
kunal-vaishnavi:
```suggestion
  std::cout << "Hello, ORT GenAI Model-QA!" << std::endl;
```

src/models/whisper.cpp
```diff
@@ -319,31 +319,29 @@ DeviceSpan<float> WhisperState::Run(int current_length, DeviceSpan<int32_t>& nex
     }
 
     return logits;
-  } else {
```
kunal-vaishnavi, line 322: I kept the else condition because it was easier to understand when the encoder stage ran vs. when the decoder stage ran (even though the early return means that the else condition isn't necessary).

examples/c/src/common.cpp
```diff
@@ -39,46 +39,11 @@ void Timing::Log(const int prompt_tokens_length, const int new_tokens_length) {
   std::cout << "-------------" << std::endl;
 }
 
-void TerminateSession::signalHandler(int signum) {
```
natke, line 42: DO we not need an example for terminate session?

src/models/whisper.cpp
baijumeswani: Usually with code like that, some compilers will emit error messages that there is non-reachable code. I was getting that error on my machine. Which is why I formatted this logic.

examples/c/src/model_chat.cpp
baijumeswani: Do you have a list of platforms where you have had issues with ANSI tags? It should work on most platforms we use on a regular basis. I can try to make changes, so we disable/introduce workarounds on platforms you've seen issues before. 
But i think it might be ok to leave them in.

examples/c/README.md
baijumeswani: Can I do that in a separate PR since I am not sure these examples build well for macos at the moment. Will introduce macos steps after testing (i don't think we had macos steps before).

examples/c/README.md
baijumeswani: I want to keep the instructions here simple. I feel like downloading a model is probably simpler through foundry-local. But you're right in that not all models on huggingface might be found on foundry local. Shall we create a new readme for instructions on downloading a model?

baijumeswani: Will benefit python as well as c++ / c#.

kunal-vaishnavi: Yes, let's have a new README for that.

examples/c/README.md
kunal-vaishnavi: Sounds good

examples/c/src/model_chat.cpp
kunal-vaishnavi: I have seen it fail on certain Windows and Linux machines. [Here](https://github.com/microsoft/onnxruntime-genai/issues/776#issuecomment-2318567981) is an example on Windows.

baijumeswani: Seems like an older windows cmd terminal. I think it is fine given we are using all sorts of things in foundry local (emojis and so on). And probably there are ways to enable these ansi tags in older terminals as well.

baijumeswani: Unless you feel strongly about it. Then I can add cmake flags to disable this. Want to keep the example as simple as possible (with the least amount of cmakke flags).

examples/c/src/common.cpp
baijumeswani: The termination feature is still shown in the example. But it is implemented differently now. We don't explicitly spawn a background thread to terminate generation. We do it through control+c signal.

examples/c/src/model_chat.cpp
kunal-vaishnavi: I think it's fine to keep it as-is then.

baijumeswani:
Thank you for the review @kunal-vaishnavi @natke :)

PR #1616, opened 2025-07-09
Move data to cpu first before transposing to avoid OOM on gpu

justinchuby:


justinchuby:
Alternatively load the model to cpu directly

PR #1617, opened 2025-07-10
Use two-step Softmax to do cuda sampling

jiafatom:
This algorithm aligns with huggingface samping algorithm, and I think it eventually solves big topk issue that we had before. The previouis sampling algorithm does not align with huggingface.

Huggingface did the logits process: TemperatureLogitsWarper, TopKLogitsWarper, TopPLogitsWarper
See
https://github.com/huggingface/transformers/blob/9f41f67135b0656c428ff2c2b446d8eb15f5a7c5/src/transformers/generation/logits_process.py
so we need do the same logic

In our change,
we start thinking that the case of TopP only is a special case of TopPTopK when k=vocab_size. And the code actually set `k=-1` for TopP case. Therefore we unify the code into
```
  if (k <= 0 || k > vocab_size)
  {
    k = vocab_size;
  }
  GetTopKSubset(data, stream, scores_in, data->scores_sorted.get(), data->indices_sorted.get(), vocab_size, batch_size, k, temperature);
```
rather than the old if-else. 

Once `k` is properly set, the remaining logic is same for all the cases.

`GetTopKKernel` did the score adjustment based on temperature, this is what `TemperatureLogitsWarper` does. In addition, it select top k scores like `TopKLogitsWarper` does. It is followed by `DispatchBlockwiseSoftmaxForward` which did the SoftMax.
`PrefixSumKernelWithActualValue` keep the tokens which survives topP (based on SoftMax that is calculated above) but use the old score, and fills remaining to minus infinity. 
This aligns with the following huggingface code:
```
        # Remove tokens with cumulative top_p above the threshold (token with 0 are kept)
        sorted_indices_to_remove = cumulative_probs <= (1 - self.top_p)
        # Keep at least min_tokens_to_keep
        sorted_indices_to_remove[..., -self.min_tokens_to_keep :] = 0

        # scatter sorted tensors to original indexing
        indices_to_remove = sorted_indices_to_remove.scatter(1, sorted_indices, sorted_indices_to_remove)
        scores_processed = scores.masked_fill(indices_to_remove, self.filter_value)
```
Then we do second round of SoftMax 
`DispatchBlockwiseSoftmaxForward` 
This aligns with 
https://github.com/huggingface/transformers/blob/main/src/transformers/generation/utils.py#L1291
```
            # token selection
            if do_sample:
                probs = nn.functional.softmax(next_token_scores, dim=-1)
```
in `def _sample(`

Then we do random sampling with
`PrefixSumKernel`, `RandomThresholdKernel` (note that we need to do random generation through 0 and 1, not p)
and `SampleKernel`
this aligns with
```
next_tokens = torch.multinomial(probs, num_samples=1).squeeze(1)
```

python benchmark_e2e.py -i my_model_dir -b 5 -k 5 -l 16 -p 0.95 
Args: batch_size = 5, prompt_length = 16, tokens = 256, max_length = 272

Before this PR
Average Tokenization Latency (per token): 0.0014017103239893913 ms
Average Tokenization Throughput (per token): 3567070.823713104 tps
Average Prompt Processing Latency (per token): 0.0740071922305844 ms
Average Prompt Processing Throughput (per token): 67561.00115812375 tps
Average Token Generation Latency (per token): 7.260638480576887 ms
Average Token Generation Throughput (per token): 688.6446713158385 tps
Average Sampling Latency (per token): 0.03105300129391253 ms
Average Sampling Throughput (per token): 161015.0320954701 tps
Average Wall Clock Time: 1.9715723514556884 s
Average Wall Clock Throughput: 945.9455031528506 tps

With this PR
Average Tokenization Latency (per token): 0.0010468188621517685 ms
Average Tokenization Throughput (per token): 4776375.532365117 tps
Average Prompt Processing Latency (per token): 0.07414427689380115 ms
Average Prompt Processing Throughput (per token): 67436.08825212005 tps
Average Token Generation Latency (per token): 7.280274138936554 ms
Average Token Generation Throughput (per token): 686.7873248424628 tps
Average Sampling Latency (per token): 0.03698428627103567 ms
Average Sampling Throughput (per token): 135192.55078651503 tps
Average Wall Clock Time: 1.9773935556411744 s
Average Wall Clock Throughput: 943.1607555711232 tps


src/cuda/cuda_sampling.cu
```diff
@@ -406,9 +410,45 @@ __global__ void PrefixSumKernel(float* scores, float* prefix_sums, int sample_ra
   }
 }
 
+template <int kBlockSize>
+__global__ void PrefixSumKernelWithActualValue(float* scores, float* prefix_sums, float* scores_temp, float* acutal_values, int sample_range, int batch_size, float p) {
```
aciddelgado: typo: actual_values

src/cuda/cuda_sampling.cu
```diff
@@ -406,9 +410,45 @@ __global__ void PrefixSumKernel(float* scores, float* prefix_sums, int sample_ra
...
+      scores_temp[global_index] = prefix_sum;
+    }
+    __syncthreads();
+    if (local_index == 0)
+    {
+      prefix_sums[global_index] = acutal_values[global_index];
+    }
+    else if (local_index < sample_range) {
+      if (scores_temp[global_index - 1] < p) {
+          prefix_sums[global_index] = acutal_values[global_index];
```
aciddelgado: Are the values in actual_values prefix sums? it seems like this function is outputting scores rather than prefix sums; if so, then a different name for the prefix_sums output and the function would be appropriate.

test/sampling_tests.cpp
```diff
@@ -516,6 +516,8 @@ TEST(SamplingTests, RandomizedSamplingTopKCuda) {
     }
   }
   // Calculate expected distribution of tokens by Softmaxing given logits (integers 1 through k)
+  // TODO: need to adjust value here to align with second Softmax call in the C++ code
```
aciddelgado: Would be good to make this change before merging PR

src/cuda/cuda_sampling.cu
jiafatom: Fixed.

test/sampling_tests.cpp
jiafatom: Updated.

aciddelgado (approved):

PR #1622, opened 2025-07-11
whisper: Allow session options to be used for encoder

RyanMetcalfeInt8:
This PR adds the ability to set `session_options` (with `provider_options`) specific to the `encoder` (via `genai_config.json`).

Before this change-set, the `session_options` specified in the `decoder` section within `genai_config.json` is used to initialize both the encoder & decoder ORT sessions.

Note that, this PR still preserves the *default* behavior -- If no `provider_options` are specified by the `encoder` section of `genai_config.json`, then it will still use the `session_options` specified for the decoder. I did this to preserve any existing use-cases, where perhaps the assumption is that the `decoder` session options will also be used for the `encoder`.

In our (OpenVINO EP) case, we would like the flexibility to run encoder/decoder models across separate OpenVINO devices -- hence we need the ability to specify differing `provider_options` between decoder & encoder.

### DIFF
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -969,6 +969,10 @@ Config::Config(const fs::path& path, std::string_view json_overlay) : config_pat
   for (const auto& provider_option : model.decoder.session_options.provider_options) {
     model.decoder.session_options.providers.push_back(provider_option.name);
   }
+
+  for (const auto& provider_option : model.encoder.session_options.provider_options) {
+    model.encoder.session_options.providers.push_back(provider_option.name);
+  }
 }
 
 void Config::AddMapping(const std::string& nominal_name, const std::string& graph_name) {
--- a/src/models/whisper.cpp
+++ b/src/models/whisper.cpp
@@ -7,7 +7,17 @@ namespace Generators {
 
 WhisperModel::WhisperModel(std::unique_ptr<Config> config, OrtEnv& ort_env)
     : Model{std::move(config)} {
-  session_encoder_ = CreateSession(ort_env, config_->model.encoder.filename, session_options_.get());
+  // default encoder session options to decoder session options.
+  OrtSessionOptions* encoder_session_options = session_options_.get();
+
+  // If provider options were explicitly added for the encoder, create session options from this config.
+  if (!config_->model.encoder.session_options.provider_options.empty()) {
+    encoder_session_options_ = OrtSessionOptions::Create();
+    CreateSessionOptionsFromConfig(config_->model.encoder.session_options, *encoder_session_options_, true, false);
+    encoder_session_options = encoder_session_options_.get();
+  }
+
+  session_encoder_ = CreateSession(ort_env, config_->model.encoder.filename, encoder_session_options);
   session_decoder_ = CreateSession(ort_env, config_->model.decoder.filename, session_options_.get());
 
   session_info_.Add(*session_decoder_);
--- a/src/models/whisper.h
+++ b/src/models/whisper.h
@@ -16,6 +16,8 @@ struct WhisperModel : Model {
 
   std::unique_ptr<OrtSession> session_encoder_;  // audio_features -> encoder_hidden_states, cross_kv_cache
   std::unique_ptr<OrtSession> session_decoder_;  // input_ids, self_kv_cache, cross_kv_cache -> logits, self_kv_cache
+
+  std::unique_ptr<OrtSessionOptions> encoder_session_options_;
 };
 
 struct AudioEncoderState : State {


RyanMetcalfeInt8:
FYI -- @baijumeswani, @kunal-vaishnavi

src/config.cpp
```diff
@@ -969,6 +969,10 @@ Config::Config(const fs::path& path, std::string_view json_overlay) : config_pat
   for (const auto& provider_option : model.decoder.session_options.provider_options) {
     model.decoder.session_options.providers.push_back(provider_option.name);
   }
+
+  for (const auto& provider_option : model.encoder.session_options.provider_options) {
+    model.encoder.session_options.providers.push_back(provider_option.name);
+  }
```
baijumeswani, line 975: Should the encoder providers be made configurable at runtime? Or is the purpose for this PR to only allow the encoder session options to be read from the genai_config specifically for the whisper model?

RyanMetcalfeInt8: Hi @baijumeswani -- I think eventually it would be good to expose some APIs that would allow the encoder providers to be configured at runtime. But for this PR, the intention is allow the session options to get read from `genai_config.json`, specifically.

src/models/whisper.cpp
```diff
@@ -7,7 +7,17 @@ namespace Generators {
...
+  OrtSessionOptions* encoder_session_options = session_options_.get();
+
+  // If provider options were explicitly added for the encoder, create session options from this config.
+  if (!config_->model.encoder.session_options.provider_options.empty()) {
+    encoder_session_options_ = OrtSessionOptions::Create();
+    CreateSessionOptionsFromConfig(config_->model.encoder.session_options, *encoder_session_options_, true, false);
+    encoder_session_options = encoder_session_options_.get();
+  }
+
+  session_encoder_ = CreateSession(ort_env, config_->model.encoder.filename, encoder_session_options);
```
baijumeswani: optional: Avoid using raw ptr if possible. I am ok if you prefer to not to apply this suggestoin. 

```suggestion
// If provider options were explicitly added for the encoder, create session options from this config.
  if (!config_->model.encoder.session_options.provider_options.empty()) {
    encoder_session_options_ = OrtSessionOptions::Create();
    CreateSessionOptionsFromConfig(config_->model.encoder.session_options, *encoder_session_options_, true, false);
  }

  session_encoder_ = CreateSession(ort_env, config_->model.encoder.filename,
                                   encoder_session_options_ ? encoder_session_options_.get() : session_options_.get());
```

RyanMetcalfeInt8: thanks! I like it.

baijumeswani (approved):

src/config.cpp
kunal-vaishnavi: I had set up the `genai_config.json` changes to enable using separate session options and execution providers for the encoder and decoder sessions. I was planning to add this support in a later PR so thank you for adding it so quickly!

src/models/whisper.cpp
```diff
@@ -7,7 +7,14 @@ namespace Generators {
 
 WhisperModel::WhisperModel(std::unique_ptr<Config> config, OrtEnv& ort_env)
     : Model{std::move(config)} {
-  session_encoder_ = CreateSession(ort_env, config_->model.encoder.filename, session_options_.get());
+  // If provider options were explicitly added for the encoder, create session options from this config.
+  if (!config_->model.encoder.session_options.provider_options.empty()) {
+    encoder_session_options_ = OrtSessionOptions::Create();
```
kunal-vaishnavi, line 12: Can we add the encoder session options here instead? That way, it can be re-used for other encoder-decoder models.

https://github.com/microsoft/onnxruntime-genai/blob/b199f27611f114eb0da09d2f7f2cb1fdf3c35ee2/src/models/model.h#L133-L169

src/models/whisper.h
```diff
@@ -16,6 +16,8 @@ struct WhisperModel : Model {
 
   std::unique_ptr<OrtSession> session_encoder_;  // audio_features -> encoder_hidden_states, cross_kv_cache
   std::unique_ptr<OrtSession> session_decoder_;  // input_ids, self_kv_cache, cross_kv_cache -> logits, self_kv_cache
+
+  std::unique_ptr<OrtSessionOptions> encoder_session_options_;
```
kunal-vaishnavi, line 20: Can we add the encoder session options here instead? That way, it can be re-used for other encoder-decoder models.

https://github.com/microsoft/onnxruntime-genai/blob/b199f27611f114eb0da09d2f7f2cb1fdf3c35ee2/src/models/model.h#L133-L169

kunal-vaishnavi, line 20: Can we add the encoder session options in the below location instead? That way, it can be re-used for other encoder-decoder models.

https://github.com/microsoft/onnxruntime-genai/blob/b199f27611f114eb0da09d2f7f2cb1fdf3c35ee2/src/models/model.h#L133-L169

baijumeswani: Maybe best to let each model type define its own options (if it needs more than one) since not all model types will require encoder session options. And may potentially require different ways to set up the session options. We've probably already cluttered the Model base class with a lot of variables that not all models need.

src/models/whisper.cpp
kunal-vaishnavi, line 12: Can this initialization be moved to where the `session_options_` are initialized instead? That way, we don't have to re-implement initializing the encoder session options for every encoder-decoder architecture.

src/models/whisper.cpp
```diff
@@ -7,7 +7,14 @@ namespace Generators {
...
+    CreateSessionOptionsFromConfig(config_->model.encoder.session_options, *encoder_session_options_, true, false);
+  }
+
+  session_encoder_ = CreateSession(ort_env, config_->model.encoder.filename,
```
kunal-vaishnavi, line 16: I think the ideal scenario would be the following. The encoder's session options can either be created from the config or set as the decoder's session options by default.

```suggestion
  session_encoder_ = CreateSession(ort_env, config_->model.encoder.filename, encoder_session_options_.get();
```

src/models/whisper.h
kunal-vaishnavi: Can this header file contain the variables and then their initializations can be handled per model type with the default initializations to use the decoder's session options? It's a [similar issue](https://github.com/microsoft/onnxruntime-genai/blob/b199f27611f114eb0da09d2f7f2cb1fdf3c35ee2/src/models/multi_modal.cpp#L65-L81) for multi-modal models as well.

baijumeswani: I think to solve this problem, we should have a single base class called `Model` that only defines simple variables needed for a Decoder model (since decoder will always exist). Then have another class called EncoderDecoderModel that derives from this. This class can define the encoder session options variable and can also contain the implementation for creating the encoder session options. And then finally have models such as Whisper that can inherit from EncoderDecoderModel.
This would avoid unnecessary clutter in the Model base class. We can do this work separately from this PR though.

src/models/whisper.cpp
RyanMetcalfeInt8: So I think one complication here is that `decoder_session_options_` is a unique_ptr, so I can't very easily initialize `encoder_session_options_` to it. Or maybe I am misunderstanding what you are suggesting.

src/models/whisper.h
kunal-vaishnavi: Sounds good

src/models/whisper.cpp
kunal-vaishnavi: The idea was to initialize `encoder_session_options_` in the same way that `decoder_session_options_` gets initialized so that the if-else logic isn't needed here. Given [this conversation](https://github.com/microsoft/onnxruntime-genai/pull/1622#discussion_r2201913495), I think it is fine to keep it as is for now.

PR #1628, opened 2025-07-15
RegisterExecutionProviderLibrary

vortex-captain:


src/ort_genai_c.cpp
```diff
@@ -888,4 +888,12 @@ void OGA_API_CALL OgaDestroyNamedTensors(OgaNamedTensors* p) { delete p; }
 void OGA_API_CALL OgaDestroyAdapters(OgaAdapters* p) { p->ExternalRelease(); }
 void OGA_API_CALL OgaDestroyRuntimeSettings(OgaRuntimeSettings* p) { delete p; }
 
+void OGA_API_CALL OgaRegisterExecutionProviderLibrary(const char* registration_name, const wchar_t* path) {
+  Ort::api->RegisterExecutionProviderLibrary(&(Generators::GetOrtEnv()), registration_name, path);
+}
```
baijumeswani: Add the new function to [this file](https://github.com/microsoft/onnxruntime-genai/blob/main/src/models/onnxruntime_api.h).

```suggestion
void OGA_API_CALL OgaRegisterExecutionProviderLibrary(const char* registration_name, const char* library_path) {
  Ort::RegisterExecutionProviderLibrary(&(Generators::GetOrtEnv()), registration_name, fs::path(library_path));
}
```

src/python/py/_dll_directory.py
```diff
@@ -40,6 +40,14 @@ def add_onnxruntime_dependency(package_id: str):
             import ctypes
             _ = ctypes.CDLL(dml_path)
 
+        # temporary workaround for onnxruntime.dll loading
+        ort_path = os.path.join(ort_package_path, "capi", "onnxruntime.dll")
+        # The dependent onnxruntime package may have multiple execution providers.
+        # Check to see if onnxruntime.dll exists before trying to load it.
+        if os.path.exists(ort_path):
+            import ctypes
+            _ = ctypes.CDLL(ort_path)
```
baijumeswani: Why is this change needed? [This line](https://github.com/microsoft/onnxruntime-genai/blob/850c6d2d1d5fa928e068e878f3e00d789df1ef88/src/python/py/_dll_directory.py#L32) should take care of loading the correct dll on windows.

Unless the user's application code loads another version of onnxruntime.dll first, then the process won't reload onnxruntime.dll and may use the wrong version.

src/python/python.cpp
```diff
@@ -552,4 +552,21 @@ PYBIND11_MODULE(onnxruntime_genai, m) {
 
   m.def("set_current_gpu_device_id", [](int device_id) { Ort::SetCurrentGpuDeviceId(device_id); });
   m.def("get_current_gpu_device_id", []() { return Ort::GetCurrentGpuDeviceId(); });
-}
+
+  m.def("register_execution_provider_library", [](const std::string& provider_name, const std::wstring& path_wstr) {
+    // TODO: input python string to std::wstring conversion performed by pybind11 may not be working properly here: wrong string if including special characters
```
baijumeswani: Does ort also currently only allow this api to run on windows (and not on other platforms)?

src/ort_genai_c.cpp
jiaxuwu2021: added new functions into `onnxruntime_genai.h`

src/python/python.cpp
jiaxuwu2021: not only windows, should be supported to run on all platforms. Updated the code.

src/python/py/_dll_directory.py
jiaxuwu2021: removed. sorry this is for local debug only, we forgot to delete it.

src/ort_genai_c.h
```diff
@@ -23,6 +23,7 @@ extern "C" {
 #define OGA_EXPORT __declspec(dllimport)
 #endif
 #define OGA_API_CALL _stdcall
+#define ORTCHAR_T wchar_t
```
baijumeswani: Do we still need to #define this?

src/ort_genai_c.h
```diff
@@ -31,6 +32,7 @@ extern "C" {
 #define OGA_EXPORT
 #endif
 #define OGA_API_CALL
+#define ORTCHAR_T char
```
baijumeswani: Same as above

src/ort_genai_c.h
jiaxuwu2021: we still need it, because `fs::path(str).c_str()` supports different string according to the platform
see https://github.com/microsoft/onnxruntime-genai/blob/main/src/filesystem.h#L82

src/ort_genai_c.h
jiaxuwu2021: we still need it, because `fs::path(str).c_str()` supports different string according to the platform
see https://github.com/microsoft/onnxruntime-genai/blob/main/src/filesystem.h#L82

src/ort_genai_c.h
jiaxuwu2021: as synced offline, you are right, this macro is no longer used, removed now.

src/ort_genai_c.h
jiaxuwu2021: as synced offline, you are right, this macro is no longer used, removed now.

PR #1629, opened 2025-07-15
Change IO dtype for INT4 CUDA models

kunal-vaishnavi:
### Description

This PR allows a user to set the IO dtype (i.e. the input/output dtype) for an INT4 CUDA ONNX model to be bfloat16 precision instead of float16 precision. This can be done by using `-p/--precision int4`, `-e/--execution_provider cuda`, and then setting `--extra_options use_bf16_cuda=true/True/1`.

### Motivation and Context

Models lose accuracy when converting weights from their native bfloat16 precision to float16 precision. With the [recent support](https://github.com/microsoft/onnxruntime/pull/25161) of bfloat16 precision in `MatMulNBits`, the conversion is not always needed any more.

### DIFF
--- a/src/python/py/models/builder.py
+++ b/src/python/py/models/builder.py
@@ -2409,8 +2409,7 @@ class Model:
 
     def make_relu_squared(self, layer_id, root_input, activation):
         relu_name = self.make_relu(layer_id, root_input, "Relu")
-        basename = f"/model/layers.{layer_id}/mlp/square/{activation}"
-        pow_name = f"{basename}/pow"
+        pow_name = f"/model/layers.{layer_id}/mlp/act_fn/Pow"
         pow_inputs = [f"{relu_name}/output_0", "/model/constants/INT32/[2]"]
         self.make_node("Pow", inputs=pow_inputs, outputs=[f"{pow_name}/output_0"], name=pow_name, domain="")
         self.make_value(f"{pow_name}/output_0", self.io_dtype, shape=['batch_size', 'sequence_length', self.intermediate_size])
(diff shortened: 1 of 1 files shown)


baijumeswani (approved):
Does this require changes to the runtime?

src/python/py/models/builder.py
```diff
@@ -3706,7 +3706,7 @@ def set_io_dtype(precision, execution_provider, extra_options) -> ir.DataType:
         # FP32 precision
         return ir.DataType.FLOAT
 
-    if precision == "bf16":
+    if precision == "bf16" or (precision == "int4" and execution_provider == "cuda"):
```
tianleiwu: I think int4 could also use FP16 for some model (like phi4 mini). Also BF16 only support Ampere or later GPU.

Why don't we distinguish quantized (int4/int8) and output precision (fp16/bf16/fp32) for builder?

kunal-vaishnavi: Most models (such as [Phi-4 mini](https://huggingface.co/microsoft/Phi-4-mini-instruct/blob/main/config.json#L140)) use bfloat16 now. This is to minimize precision loss in ONNX by maintaining the original precision when possible.

I was hoping to avoid that to keep the usage simple, but I think that is something that will need to happen now. Then, older GPUs can use INT4 + FP16 and newer GPUs can use INT4 + BF16.

baijumeswani: Would we publish both types of models then? Would application layers like foundry local then need to check the generation of the gpus before downloading and running with ort/ort-genai?

tianleiwu: fp16 has higher precision than bf16 but limited range (so it is more likely overflow). I think we shall first try int4+fp16 whenever it is possible. If there is overflow during testing, we shall switch to int4+bf16 anyway. So we shall only publish one based on this rule.

kunal-vaishnavi: I moved it to be optional as part of the extra options instead of making it the default.

PR #1634, opened 2025-07-17
Add open bytes functionality for C#

ajindal1:
Add open bytes functionality for C#

There might be a more efficient way to do this using `fixed`. I will explore that option in parallel. But since we have a deadline for this, we can start with this implementation.

### DIFF
--- a/test/csharp/TestOnnxRuntimeGenAIAPI.cs
+++ b/test/csharp/TestOnnxRuntimeGenAIAPI.cs
@@ -234,7 +234,7 @@ namespace Microsoft.ML.OnnxRuntimeGenAI.Tests
         {
             byte[] audioBytes = File.ReadAllBytes(Path.Combine(GetDirectoryInTreeThatContains(Directory.GetCurrentDirectory(), "test"),
                                                             "test_models", "audios", "1272-141231-0002.mp3"));
-            var audios = Audios.OpenBytes(audioBytes);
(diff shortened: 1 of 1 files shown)


src/csharp/Audios.cs
```diff
@@ -30,6 +30,29 @@ public static Audios Load(string[] audioPaths)
             return new Audios(audiosHandle);
         }
 
+        public static Audios OpenBytes(byte[] audioBytesDatas)
+        {
+            if (audioBytesDatas == null || audioBytesDatas.Length == 0)
+            {
+                throw new ArgumentException("Audio byte data cannot be null or empty.");
+            }
```
baijumeswani: I think we should let the native code handle the checks so there is uniformity in the error messages. Could you check if the C implementation already has such a check? If not, can we add it there.

src/csharp/Audios.cs
```diff
@@ -30,6 +30,29 @@ public static Audios Load(string[] audioPaths)
             return new Audios(audiosHandle);
         }
 
+        public static Audios OpenBytes(byte[] audioBytesDatas)
```
baijumeswani: Shall we just call it `Load` to match C++ api?

```suggestion
        public static Audios Load(byte[] audioBytesData)
```

baijumeswani (approved):

kunal-vaishnavi (approved):

PR #1636, opened 2025-07-17
Use last windowed input ids to update logits

baijumeswani:
Bug: When using windowing logic for large sequence length inputs (sequence length > window size), `Logits::Update` was using the `input_ids` associated with the first chunk (always) since it didn't receive any of the subsequent chunks to process.

As a result, for large sequence length inputs (sequence length > window size), incorrect logits were being used to compute the next tokens (always window_size -1). This pull-request addresses this bug by processing the correct windowed input ids.

### DIFF
--- a/src/models/input_ids.h
+++ b/src/models/input_ids.h
@@ -58,7 +58,7 @@ struct WindowedInputIDs : public InputIDs {
   void Add() override;
   void Update(DeviceSpan<int32_t> next_tokens) override;
   std::array<int64_t, 2> GetShape() const override { return shape_; }
-  OrtValue* Get() override { return value_->GetOrtTensor(); }
+  OrtValue* Get() override { return value_.get(); }
 
  private:
   State& state_;


src/models/decoder_only_pipeline.cpp
```diff
@@ -357,6 +357,8 @@ DeviceSpan<float> DecoderOnlyPipelineState::Run(int total_length, DeviceSpan<int
       input_ids_->Update(next_tokens);
       UpdateKeyValueCache(next_indices, total_length);
       position_inputs_->Update(next_tokens, total_length, static_cast<int>(input_ids_->GetShape()[1]));
+      logits_.Update(WrapTensor<int32_t>(*model_.p_device_inputs_, *input_ids_->Get()),
```
apsonawane, line 360: Why do we need to update logits here again after UpdateInputsOutputs?

aciddelgado (approved):

src/models/decoder_only_pipeline.cpp
baijumeswani: This is for windowing logic. For windowing, the prompt is broken down into window chunks. Each window runs through the model. The very last chunk can have input_ids that are padded. In such a case, we want to use the logits of the last valid token in the input ids and not the last one from the window. That's why we need the update here.

PR #1641, opened 2025-07-21
Fixes all packaging pipelines

baijumeswani:
Recent pull-requests broke the packaging pipeline. This pull-request addresses the issues.

I also discovered that the binaries were no longer getting correctly signed. The primary reason is that our `python build.py ` or `cmake --preset ... --build` script does two things:

- Build the binaries
- Package them (into *.zip, *.tar.gz or *.whl)

As a result, there is no room to codesign the binaries (since codesigning is an intermediate step and we cannot yield from a python script before continuing).
In order to codesign the binaries, we need to break down the build process in two phases.

1. Build the biaries
2. -- intermediate step to codesign --
3. Build the package

This pr introduces that change as well.

### DIFF
--- a/.github/workflows/android-build.yml
+++ b/.github/workflows/android-build.yml
@@ -94,12 +94,12 @@ jobs:
         run: |
           set -e -x
           rm -rf build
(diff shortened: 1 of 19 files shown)


snnn (approved):
Or you may build the wheels, then exact the files out, sign them and repack them. Either works.

PR #1642, opened 2025-07-21
Make default top_k=50 in model builder

jiafatom:
According to huggingface, top_k (int, optional, defaults to 50) — The number of highest probability vocabulary tokens to keep for top-k-filtering. This value is set in a model’s generation_config.json file. If it isn’t set, the default value is 50.

Our sampling algorithm has deficiency before so top_k=50 is not working for ort-genai.
Now we merged the new sampling algorithm https://github.com/microsoft/onnxruntime-genai/pull/1617, https://github.com/microsoft/onnxruntime-genai/pull/1631
So we can use top_k=50 to align with hf now.

### DIFF
--- a/src/python/py/models/builder.py
+++ b/src/python/py/models/builder.py
@@ -424,7 +424,7 @@ class Model:
                 "past_present_share_buffer": False if "config_only" in self.extra_options else self.past_present_share_buffer,
                 "repetition_penalty": config.repetition_penalty if hasattr(config, "repetition_penalty") else 1.0,
                 "temperature": config.temperature if hasattr(config, "temperature") else 1.0,
-                "top_k": 1,
+                "top_k": config.top_k if hasattr(config, "top_k") else 50,
                 "top_p": config.top_p if hasattr(config, "top_p") else 1.0,
             },
         }
--- a/test/sampling_tests.cpp
+++ b/test/sampling_tests.cpp
@@ -32,6 +32,7 @@ TEST(SamplingTests, BatchedSamplingTopPCpu) {
   auto params = OgaGeneratorParams::Create(*model);
   params->SetSearchOption("max_length", 10);
   params->SetSearchOptionBool("do_sample", true);
+  params->SetSearchOption("top_k", 1);
   params->SetSearchOption("top_p", 0.25f);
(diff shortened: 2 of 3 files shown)


src/python/py/models/builder.py
```diff
@@ -424,7 +424,7 @@ def make_genai_config(self, model_name_or_path, extra_kwargs, out_dir):
                 "past_present_share_buffer": False if "config_only" in self.extra_options else self.past_present_share_buffer,
                 "repetition_penalty": config.repetition_penalty if hasattr(config, "repetition_penalty") else 1.0,
                 "temperature": config.temperature if hasattr(config, "temperature") else 1.0,
-                "top_k": 1,
+                "top_k": config.top_k if hasattr(config, "top_k") else 50,
```
baijumeswani, line 427: Could you also update the runtime default here: https://github.com/microsoft/onnxruntime-genai/blob/f1a6509defaaf638f9af0fb4aec30ba7493f6d5d/src/config.h#L254? In case the value is not mentioned in the genai_config.json, we can default to using 50.

baijumeswani (dismissed):

kunal-vaishnavi (approved):

jiafatom:
> Can we put the motivation for this change in the PR description? That way, it will be documented in case it needs to be referred to in the future.

updated.

test/sampling_tests.cpp
```diff
@@ -32,6 +32,7 @@ TEST(SamplingTests, BatchedSamplingTopPCpu) {
   auto params = OgaGeneratorParams::Create(*model);
   params->SetSearchOption("max_length", 10);
   params->SetSearchOptionBool("do_sample", true);
+  params->SetSearchOption("top_k", 1);
```
kunal-vaishnavi, line 35: Should this be set to 1 or 50?

kunal-vaishnavi (approved):

baijumeswani (approved):

test/sampling_tests.cpp
jiafatom: This test has `vocab_size=5`, so we cannot use `top_k=50` here.

PR #1645, opened 2025-07-22
Enable cuda graph for LLMs for NvTensorRtRtx EP

anujj:
- Enable cuda graph for LLMs for NvTensorRtRtx EP
- change the flag name from nv_cuda_graph_enable to enable_cuda_graph

### DIFF
--- a/src/python/py/models/builder.py
+++ b/src/python/py/models/builder.py
@@ -94,7 +94,9 @@ class Model:
             },
             "dml": {},
             "webgpu": {},
-            "NvTensorRtRtx": {},
+            "NvTensorRtRtx": {
+                "enable_cuda_graph": "1",
+            },
         }
 
         # Map input names to their types and shapes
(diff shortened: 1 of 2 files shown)


src/python/py/models/builder.py
```diff
@@ -94,7 +94,9 @@ def __init__(
             },
             "dml": {},
             "webgpu": {},
-            "NvTensorRtRtx": {},
+            "NvTensorRtRtx": {
+                **({"enable_cuda_graph": "1"} if extra_options.get("enable_cuda_graph", False) else {}),
```
baijumeswani: Why do you need to pack and then unpack the dict? Are you able to use the dict directly?

anujj: The NV EP in ORT 1.22 doesn’t support the enable_cuda_graph flag, so even passing enable_cuda_graph: 0 fails. The EP validates flags and errors out, which is why I’m avoiding passing it at all to keep things compatible.

src/python/py/models/builder.py
```diff
@@ -94,7 +94,9 @@ def __init__(
             },
             "dml": {},
             "webgpu": {},
-            "NvTensorRtRtx": {},
+            "NvTensorRtRtx": {
```
kunal-vaishnavi: I think it would be simpler to do the following instead.

```suggestion
            "NvTensorRtRtx": {"enable_cuda_graph": "1"} if extra_options.get("enable_cuda_graph", False) else {}
```

anujj: Thanks, done

kunal-vaishnavi (approved):

PR #1646, opened 2025-07-22
Update the benchmark scripts to account for the time spent in sampling

gaugarg-nv:
This adjustment is crucial for the CUDA Execution Provider (EP), which relies on GPU-based sampling. In the case of CUDA EP, the generation time currently only includes the sampling kernels' launch time, not the execution time, due to synchronization occurring during the `generator.is_done()` call. For other EPs that perform sampling on the CPU, the token generation time already encompasses the sampling time.

I added nvtx marker in the generation loop (shown as gray with text "Decode").

Without the fix, there is a gap in decode steps. This is due to `cudaStreamSynchronize` call in `is_done` call.

image

With Fix: There is no gap as  `is_done`  is part of the loop.

image

### DIFF
--- a/benchmark/c/main.cpp
+++ b/benchmark/c/main.cpp
@@ -236,7 +236,8 @@ void RunBenchmark(const benchmark::Options& opts) {
         {
           Timing token_gen_timing{token_gen_times};
           generator->GenerateNextToken();
-          generator_done = generator->IsDone(); // Enforce stream synchronize to compute accurate token generation times
+          // Enforce stream synchronize to compute accurate token generation times
+          generator_done = generator->IsDone();
         }
       }
     }
--- a/benchmark/c/main.cpp
+++ b/benchmark/c/main.cpp
@@ -224,15 +224,19 @@ void RunBenchmark(const benchmark::Options& opts) {
         generator->AppendTokenSequences(*prompt_sequences);
       }
 
+      bool generator_done = false;
+
       {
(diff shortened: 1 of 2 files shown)


gaugarg-nv:
FYI @baijumeswani @kunal-vaishnavi

benchmark/c/main.cpp
```diff
@@ -224,15 +224,19 @@ void RunBenchmark(const benchmark::Options& opts) {
...
         generator->GenerateNextToken();
+        generator_done = generator->IsDone();
       }
 
-      while (!generator->IsDone()) {
+      while (!generator_done) {
         {
           Timing token_gen_timing{token_gen_times};
           generator->GenerateNextToken();
+          generator_done = generator->IsDone();
```
baijumeswani, line 240: Could we add a comment here indicating why this was done this way?

```suggestion
          generator_done = generator->IsDone(); // Enforce stream synchronize to compute accurate token generation times
```

gaugarg-nv: Done.

baijumeswani (approved):

benchmark/c/main.cpp
gaugarg-nv: Had a clang-format error with this change. Now fixed.

baijumeswani (approved):

baijumeswani:
Closing and reopening to trigger pipelines.

PR #1647, opened 2025-07-22
[Builder] Add support for Olive quantized models

jambayk:
- Support new `"olive"` quant type
- Weight and zero-point packings are the same as gptq. No g_idx.
- Similar to the `k_quant` mixed precision `int4_algo`, select matmuls can be in 8 bits.
   - Currently, we ensure that the `q_proj`, `k_proj`, `v_proj` matmuls use the same configuration (bits and group_size) so that they can be merged without issues.
- The modules are generalized to remove the requirement that all matmuls in a layer must have the same bits and group_size.
- `quant_weight` and `dequant_weight` support no `g_idx` by using `repeat_interleave`. Otherwise, we have to create a trivial g_idx like the quark model does.
- `pack_ort_format` supports 8 bit packing.

### DIFF
--- a/src/python/py/models/builder.py
+++ b/src/python/py/models/builder.py
@@ -901,6 +901,9 @@ class Model:
             # print(f"Quantizing to {self.onnx_dtype} on-the-fly is not currently supported.")
             # print(f"Saving as {self.io_dtype} on-the-fly and quantizing to {self.onnx_dtype} at the end.")
             return self.make_matmul_float(matmul, matmul_name, root_input, **kwargs)
+        
+        if matmul.bits != 4:
+            raise NotImplementedError(f"{matmul.bits} bits precision is not currently supported in QDQ format.")
 
         dequantize_output = self.make_dequantize_linear(f"{matmul_name}/DequantizeLinear", matmul)
 
--- a/src/python/py/models/builder.py
+++ b/src/python/py/models/builder.py
@@ -994,6 +994,10 @@ class Model:
         # Create dummy PackedMatMul class
         class PackedMatMul:
             def __init__(self):
+                if q_matmul.bits != k_matmul.bits or q_matmul.bits != v_matmul.bits:
+                    raise ValueError("All MatMuls must have the same bits for packed MatMul.")
+                if q_matmul.group_size != k_matmul.group_size or q_matmul.group_size != v_matmul.group_size:
+                    raise ValueError("All MatMuls must have the same group size for packed MatMul.")
                 self.qweight = torch.cat([q_matmul.qweight, k_matmul.qweight, v_matmul.qweight], dim=0)
                 self.scales = torch.cat([q_matmul.scales, k_matmul.scales, v_matmul.scales], dim=0)
                 self.qzeros = torch.cat([q_matmul.qzeros, k_matmul.qzeros, v_matmul.qzeros], dim=0)
--- a/src/python/py/models/quantized_model.py
+++ b/src/python/py/models/quantized_model.py
@@ -1,7 +1,6 @@
 # -------------------------------------------------------------------------
 # Copyright (c) Microsoft Corporation.  All rights reserved.
 # Licensed under the MIT License.  See License.txt in the project root for
-# license information.
 # --------------------------------------------------------------------------
 # Modifications Copyright(C) 2024 Advanced Micro Devices, Inc. All rights reserved
 """
@@ -21,7 +20,7 @@ import re
 
 
 class QuantizedTensorModule:
-    def __init__(self, bits, group_size):
+    def __init__(self):
         self.qweight = None
         self.scales = None
         self.qzeros = None
@@ -30,12 +29,16 @@ class QuantizedTensorModule:
 
         self.in_features = 0
         self.out_features = 0
-        self.bits = bits
-        self._group_size = group_size
+        self.bits = None
+        self._group_size = None
 
     @property
     def group_size(self):
         return self._group_size if self._group_size != -1 else self.in_features
+
+    @group_size.setter
+    def group_size(self, value):
+        self._group_size = value
 
     def __str__(self):
         qweight = f"qweight = {self.qweight.shape}, {self.qweight}\n"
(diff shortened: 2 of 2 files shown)


src/python/py/models/builder.py
```diff
@@ -901,6 +901,9 @@ def make_matmul_int4_qdq(self, matmul, matmul_name, root_input, **kwargs):
             # print(f"Quantizing to {self.onnx_dtype} on-the-fly is not currently supported.")
             # print(f"Saving as {self.io_dtype} on-the-fly and quantizing to {self.onnx_dtype} at the end.")
             return self.make_matmul_float(matmul, matmul_name, root_input, **kwargs)
+        
+        if matmul.bits != 4:
+            raise NotImplementedError(f"{matmul.bits} bits precision is not currently supported in QDQ format.")
```
justinchuby, line 907: Helpful to also mention what _is_ support in the error message so it's actionable?

src/python/py/models/quantized_model.py
```diff
@@ -339,75 +340,86 @@ def __init__(self, quant_type, input_path, quant_attrs, q_size, kv_size, interme
...
                             # model.layers.layer_id.mlp.gate_up_proj.bias
                             # model.layers.layer_id.mlp.dense_h_to_4h.bias
-                            module.mlp.gate_proj.bias = tensor[: intermediate_size]
-                            module.mlp.up_proj.bias = tensor[intermediate_size: ]
+                            tensor_map["mlp.gate_proj.bias"] = tensor[: intermediate_size]
+                            tensor_map["mlp.up_proj.bias"] = tensor[intermediate_size: ]
                         else:
                             raise NotImplementedError(f"{name} in your quantized model is not recognized.")
 
+                        for tensor_name, tensor_value in tensor_map.items():
```
kunal-vaishnavi, line 412: It seems `tensor_map` is mainly used for setting the `bits` and `group_size` fields per tensor. Since only one key-value pair is added to `tensor_map` per iteration before this line is reached, why is the for-loop needed?

src/python/py/models/quantized_model.py
```diff
@@ -625,25 +665,30 @@ def quant_weight(self, module):
...
+        intweight_T = torch.round((weight + scale_zeros) / scales).to(torch.int)
 
         return intweight_T
 
     def pack_ort_format(self, module, intweight):
         """
         Pack `scales`, `qzeros`, and `qweight` to ORT format
         """
-        if module.bits != 4:
+        if module.bits not in [4, 8]:
```
kunal-vaishnavi: How much work is left to support 2-bits?

kunal-vaishnavi:
Can we uncomment some of the CI models to test the quantized PyTorch to quantized ONNX path?

https://github.com/microsoft/onnxruntime-genai/blob/f9a57f5f1a5b9d5577bda95ec7a5f0ae44484cb7/test/python/_test_utils.py#L63-L95

src/python/py/models/quantized_model.py
```diff
@@ -608,9 +644,13 @@ def dequant_weight(self, module):
 
         # De-quantize weight to higher precision
         scale_zeros = zeros * scales
-        scale_mat = scales[g_idx]
-        scale_zeros_mat = scale_zeros[g_idx]
-        qdq_weight_T = intweight * scale_mat - scale_zeros_mat.half()
+        if g_idx is not None:
+            scales = scales[g_idx]
+            scale_zeros = scale_zeros[g_idx]
+        elif module.group_size != module.in_features:
```
kunal-vaishnavi, line 660: Should the else condition work as-is or should an error be raised?

src/python/py/models/quantized_model.py
```diff
@@ -625,25 +665,30 @@ def quant_weight(self, module):
         g_idx = module.g_idx
 
         scale_zeros = zeros * scales
-        scale_mat = scales[g_idx]
-        scale_zeros_mat = scale_zeros[g_idx]
-        intweight_T = torch.round((weight + scale_zeros_mat) / scale_mat).to(torch.int)
+        if g_idx is not None:
+            scales = scales[g_idx]
+            scale_zeros = scale_zeros[g_idx]
+        elif module.group_size != module.in_features:
```
kunal-vaishnavi, line 681: Should the else condition work as-is or should an error be raised?

src/python/py/models/quantized_model.py
jambayk: some of the tensors like qkv_proj and gate_up proj in the checkpoints for phi models are split into more than one tensor. so I added a map and loop for such cases

src/python/py/models/quantized_model.py
jambayk: not much I think. would need to generalize the if condition into a function of the the bit size.

src/python/py/models/quantized_model.py
jambayk: else condition works as is. that is when group_size == -1  (= in_features and thus per-channel). in that case, the shape is 1 in the relevant demension so it broadcasts.

src/python/py/models/quantized_model.py
jambayk: same as the above comment

src/python/py/models/builder.py
jambayk: I could add a comment about why it's not supported. It's not actionable to the user so not sure an error message to them would be useful. 
this is not supported since the code assumes 4bit. they could export as matmulnbits and use the olive pass matmulnbitstoqdq instead https://github.com/microsoft/Olive/blob/main/olive/passes/onnx/mnb_to_qdq.py#L27 which I plan to generalize to other bits and also has more options.

jambayk:
> Can we uncomment some of the CI models to test the quantized PyTorch to quantized ONNX path?
> 
> https://github.com/microsoft/onnxruntime-genai/blob/f9a57f5f1a5b9d5577bda95ec7a5f0ae44484cb7/test/python/_test_utils.py#L63-L95

I can do it as part of this PR but not sure which ones to uncomment.

src/python/py/models/quantized_model.py
kunal-vaishnavi: Since it is not much, can we add that in this PR since [support for 2-bit quantization is coming soon](https://github.com/microsoft/onnxruntime-genai/pull/1643)?

src/python/py/models/builder.py
jambayk: added a comment

src/python/py/models/quantized_model.py
jambayk: done! I generalized `pack_on_row_for_2_4_8_bits` and reused it for here.

PR #1650, opened 2025-07-23
Enable GPU based sampling for TRT-RTX

gaugarg-nv:
It shows an improvement of about ~6% in throughput for small models those have large vocab.

**Before Gen phase**:

image

**Before pre-fill phase**: The full logits are being copied to host. This screenshot is for 100 ISL. The time increases further for larger seq length as logit size increases.

image

**After Gen phase**: GPU idle time is gone as sampling is done on GPU

image

**After pre-fill phase**: large mem copy is gone

image

### DIFF
--- a/src/models/model.cpp
+++ b/src/models/model.cpp
@@ -720,8 +720,8 @@ Model::Model(std::unique_ptr<Config> config) : config_{std::move(config)} {
   CreateSessionOptions();
   EnsureDeviceOrtInit(*p_device_, *config_);
 
-  // Only CUDA and DML does every input on the device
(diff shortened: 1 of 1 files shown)


gaugarg-nv:
FYI @baijumeswani @kunal-vaishnavi @BLSharda @anujj

baijumeswani:
> @baijumeswani , does it take so long to auto-merge the changes. I see that 2 checks are pending for very long

We've recently had to move some of our agents to a different subscription. As a result, the agents have been overwhelmed with jobs. I'll follow up internally to see if we can get more machines allocated.

PR #1659, opened 2025-07-28
Minor changes in TRT-RTX min and opt profile

gaugarg-nv:
- Keep minimum batch size and seq length as 1 instead of 0
- Change opt profile for attention mask and past KV shape to have larger seq length instead of 1

### DIFF
--- a/src/models/model.cpp
+++ b/src/models/model.cpp
@@ -433,20 +433,25 @@ void ConfigureNvTensorRtRTxProfile(const Config& config, OrtSessionOptions& sess
     session_options.AddConfigEntry("ep.nvtensorrtrtxexecutionprovider.nv_profile_opt_shapes", opt_shapes.str().c_str());
     session_options.AddConfigEntry("ep.nvtensorrtrtxexecutionprovider.nv_profile_max_shapes", max_shapes.str().c_str());
   } else {
-    // Single profile mode: simple shapes with batch_dim=[0,1,1] and seq_dim=[0,1,max_context_len]
+    // Single profile mode: simple shapes with batch_dim=[1,1,batch_size] and seq_dim=[1,1024,max_context_len]
     std::ostringstream min_shapes, opt_shapes, max_shapes;
 
-    // MIN SHAPES: batch_dim=0, seq_dim=0
-    min_shapes << Config::Defaults::InputIdsName << ":0x0,"
-               << Config::Defaults::AttentionMaskName << ":0x0";
-    add_key_value_cache_shapes(min_shapes, 0, past_key_pattern, past_value_pattern, 0, num_layers, num_kv_heads, head_dim);
+    // MIN SHAPES: batch_dim=1, seq_dim=1
+    const int min_context_len = 1;
+    const int min_batch_size = 1;
+    min_shapes << Config::Defaults::InputIdsName << ":" << min_batch_size << "x" << min_context_len << ","
+               << Config::Defaults::AttentionMaskName << ":" << min_batch_size << "x" << min_context_len;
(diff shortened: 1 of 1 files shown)


src/models/model.cpp
```diff
@@ -433,20 +433,25 @@ void ConfigureNvTensorRtRTxProfile(const Config& config, OrtSessionOptions& sess
...
-
-    // OPT SHAPES: batch_dim=1, seq_dim=1
-    opt_shapes << Config::Defaults::InputIdsName << ":1x1,"
-               << Config::Defaults::AttentionMaskName << ":1x1";
-    add_key_value_cache_shapes(opt_shapes, 1, past_key_pattern, past_value_pattern, 1, num_layers, num_kv_heads, head_dim);
-
-    // MAX SHAPES: batch_dim=1, seq_dim=max_context_len
+    // MIN SHAPES: batch_dim=1, seq_dim=1
+    const int min_context_len = 1;
+    const int min_batch_size = 1;
```
```suggestion
    constexpr int min_context_len = 1;
    constexpr int min_batch_size = 1;
```

src/models/model.cpp
```diff
@@ -433,20 +433,25 @@ void ConfigureNvTensorRtRTxProfile(const Config& config, OrtSessionOptions& sess
...
+    min_shapes << Config::Defaults::InputIdsName << ":" << min_batch_size << "x" << min_context_len << ","
+               << Config::Defaults::AttentionMaskName << ":" << min_batch_size << "x" << min_context_len;
+    add_key_value_cache_shapes(min_shapes, min_batch_size, past_key_pattern, past_value_pattern, 0, num_layers, num_kv_heads, head_dim);
+
+    // OPT SHAPES: batch_dim=1, seq_dim=1024
+    const int opt_context_len = std::min(max_context_len / 2, 1024);  // Use a reasonable opt context length
+    const int opt_batch_size = 1;                                     // Use a opt batch size of 1
```
```suggestion
    constexpr int opt_batch_size = 1;                                     // Use a opt batch size of 1
```

src/models/model.cpp
```diff
@@ -433,20 +433,25 @@ void ConfigureNvTensorRtRTxProfile(const Config& config, OrtSessionOptions& sess
...
-    // MAX SHAPES: batch_dim=1, seq_dim=max_context_len
+    // MIN SHAPES: batch_dim=1, seq_dim=1
+    const int min_context_len = 1;
+    const int min_batch_size = 1;
+    min_shapes << Config::Defaults::InputIdsName << ":" << min_batch_size << "x" << min_context_len << ","
+               << Config::Defaults::AttentionMaskName << ":" << min_batch_size << "x" << min_context_len;
+    add_key_value_cache_shapes(min_shapes, min_batch_size, past_key_pattern, past_value_pattern, 0, num_layers, num_kv_heads, head_dim);
+
+    // OPT SHAPES: batch_dim=1, seq_dim=1024
+    const int opt_context_len = std::min(max_context_len / 2, 1024);  // Use a reasonable opt context length
```
baijumeswani, line 447: Could we also make `max_context_len` constexpr? then we can update this line as well.

```suggestion
    constexpr int opt_context_len = std::min(max_context_len / 2, 1024);  // Use a reasonable opt context length
```

src/models/model.cpp
gaugarg-nv: Done

src/models/model.cpp
gaugarg-nv: Done

src/models/model.cpp
gaugarg-nv: `max_context_len` is being read from config, so can't be constexpr.

PR #1662, opened 2025-07-29
Introduce support for dynamic batching

baijumeswani:
The changes in #1580 introduce a new API in `onnxruntime-genai` that enables continuous submission of requests to a generation engine. Internally, the scheduler statically batches these requests, allowing the API to interface seamlessly with models designed for static batching.

This pull request lays the groundwork for dynamic request batching by incorporating concepts from [vLLM](https://docs.vllm.ai/en/v0.9.1/design/kernel/paged_attention.html). Specifically, it implements a block-based key-value cache manager.

To support dynamic batching, the underlying model must handle variable-length inputs (as opposed to fixed-size static batches) and leverage the [`PagedAttention` contrib-op](https://github.com/microsoft/onnxruntime/blob/87f149971f6b47cc82e84e08685193f31bb0cf8e/onnxruntime/core/graph/contrib_ops/bert_defs.cc#L1258).


src/config.h
```diff
@@ -264,10 +264,25 @@ struct Config {
     int no_repeat_ngram_size{};        // Unused param
     float diversity_penalty{};         // Unused param
     float length_penalty{1.0f};        // Exponential penalty to the length that is used with beam-based generation. length_penalty > 0.0 promotes longer sequences, while length_penalty < 0.0 encourages shorter sequences.
-    bool past_present_share_buffer{};  // The past/present kv tensors are shared and allocated once to max_length (cuda only)
+    bool past_present_share_buffer{};  // The past/present kv tensors are shared and  allocated once to max_length (cuda only)
```
```suggestion
    bool past_present_share_buffer{};  // The past/present kv tensors are shared and allocated once to max_length (cuda only)
```

src/config.h
```diff
@@ -264,10 +264,25 @@ struct Config {
...
 
+  struct Engine {
+    struct DynamicBatching {
+      size_t block_size{256};                       // Total number of slots per block.
+      std::optional<size_t> num_blocks;             // Total number of blocks per layer.
+      std::optional<float> gpu_utilization_factor;  // Fraction of free GPU memory to use for key-value cache.
+      size_t max_batch_size{16};                    // Maximum batch size for dynamically batching requests.
+    };
+    std::optional<DynamicBatching> dynamic_batching;  // Dynamic batching settings
+
```
kunal-vaishnavi, line 286: How are these defaults for the batch size determined?

src/engine/block.cpp
```diff
@@ -0,0 +1,100 @@
...
+    : id_(id), size_(slots), capacity_(block_size) {}
+
+size_t Block::Id() const {
+  return id_;
+}
+
+size_t Block::Size() const {
+  return size_;
+}
+
```
kunal-vaishnavi, line 22: Can these one-line functions be initialized in the header file?

src/engine/block.cpp
```diff
@@ -0,0 +1,100 @@
...
+      if (blocks_[i] == nullptr) {
+        blocks_[i] = std::make_shared<Block>(i, num_slots, block_size_);
+        return blocks_[i];
+      }
+    }
+    return std::shared_ptr<Block>();
+  };
+
+  if (BlocksNeeded(num_slots) > AvailableBlocks()) {
+    throw std::runtime_error("Not enough blocks available in the pool.");
```
kunal-vaishnavi: Can we give more memory information when a runtime error is thrown to help the user understand the issue and potentially how to fix it (similar to the out-of-memory errors thrown by ORT)?

src/engine/decoders/static_batch_decoder_io.h
```diff
@@ -0,0 +1,46 @@
...
+/**
+ * @class StaticBatchDecoderIO
+ * @brief Prepares and manages the inputs and outputs for a static batch decoder model.
+ *
+ * A static batch decoder model is one that expects the first dimension of the input tensors
+ * to be the batch size, and all sequences in the batch to have the same length (padded as necessary).
+ * This class handles the preparation of the following input and output tensors:
+ * Inputs:
+ * - Input IDs - int64[batch_size, sequence_length]
+ * - Attention Mask - int32[batch_size, sequence_length]
```
kunal-vaishnavi: Should these be int64 instead?

src/engine/decoders/varlen_decoder_io.h
```diff
@@ -0,0 +1,43 @@
...
+
+/**
+ * @class VarlenDecoderIO
+ * @brief Prepares and manages the inputs and outputs for a variable-length decoder model.
+ *
+ * A variable-length decoder model is one that can handle input sequences of varying lengths
+ * within the same batch. This class handles the preparation of the following input and output tensors:
+ * Inputs:
+ * - Input IDs - int64[total_num_tokens]
+ * - Cumulative Sequence Lengths - int32[batch_size + 1]
```
kunal-vaishnavi, line 19: Same question about the dtypes

src/engine/paged_key_value_cache.cpp
```diff
@@ -0,0 +1,216 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#include "cache_manager.h"
+
+#include <iostream>
```
kunal-vaishnavi: Can this be removed? I don't see any `std::cout` calls.

src/smartptrs.h
```diff
@@ -128,6 +128,7 @@ struct DeviceInterface {
   virtual void CopyCrossQK(Ort::Float16_t* /*cross_qk_buffer_data*/, void** /*qk_layer_pointers*/, int /*token_index*/, int /*batch_beam_size*/, int /*num_layers*/, int /*num_heads*/, int /*num_alignment_heads*/, const int* /*alignment_heads*/, int /*frames*/, int /*max_length*/, int /*sequence_length*/) { assert(false); }
   virtual void FinalizeCrossQK(int /*iteration_number*/, int /*context_decoding_len*/, int /*batch_size*/, int /*num_beams*/, int /*max_length*/, int /*num_alignment_heads*/, int /*frames_of_k*/, const float* /*cross_qk_buffer_data*/, float* /*cross_qk_output*/, int /*num_return_sequences*/, const int* /*cache_indir_data*/) { assert(false); }
   virtual void FinalizeCrossQK(int /*iteration_number*/, int /*context_decoding_len*/, int /*batch_size*/, int /*num_beams*/, int /*max_length*/, int /*num_alignment_heads*/, int /*frames_of_k*/, const Ort::Float16_t* /*cross_qk_buffer_data*/, Ort::Float16_t* /*cross_qk_output*/, int /*num_return_sequences*/, const int* /*cache_indir_data*/) { assert(false); }
+  virtual void GetAvailableMemory(size_t& free_bytes, size_t& total_bytes) { assert(false); }
```
kunal-vaishnavi:
```suggestion
  virtual void GetAvailableMemory(size_t& /*free_bytes*/, size_t& /*total_bytes*/) { assert(false); }
```

src/engine/decoders/varlen_decoder_io.h
baijumeswani: These I believe are supposed to be int32

src/engine/block.cpp
baijumeswani: This particular block of code needn't be in the header file. We could add it there but since it is not performance heavy and nor does it need to be inlined, I think it makes sense for it to be here.

src/config.h
baijumeswani: I initially chose 16 as an estimate for what users might run on edge devices. This value should be revisited and adjusted based on benchmarking results or practical heuristics

baijumeswani:
Thank you for the review. :)

PR #1664, opened 2025-07-29
Update C++, C# and Python Examples

sayanshaw24:
### Updates

Updates examples to now use ApplyChatTemplate, including in multimodal cases as it now supports multimodal templating with a Minja-powered C++ templating engine.

Note that we decided not to use the newly added whisper chat template functionality (https://github.com/microsoft/onnxruntime-extensions/pull/981) for the examples, because specifically for our examples itself, the chat template functionality (just appending message contents) does not make sense - we do not need it.

### Validation
- [x] Native C++, Python and C# testing using phi-4, phi-3-vision, and phi-4-mm models
- [x] Ran C# examples e2e with project solutions

### DIFF
--- a/examples/python/model-vision.py
+++ b/examples/python/model-vision.py
@@ -40,7 +40,7 @@ def run(args: argparse.Namespace):
 
     tokenizer = og.Tokenizer(model)
     processor = model.create_multimodal_processor()
-    tokenizer_stream = processor.create_stream()
+    stream = processor.create_stream()
 
     interactive = not args.non_interactive
 
@@ -118,7 +118,7 @@ def run(args: argparse.Namespace):
             generator.generate_next_token()
 
             new_token = generator.get_next_tokens()[0]
-            print(tokenizer_stream.decode(new_token), end="", flush=True)
+            print(stream.decode(new_token), end="", flush=True)
 
         print()
         total_run_time = time.time() - start_time
--- a/examples/python/model-vision.py
+++ b/examples/python/model-vision.py
@@ -5,6 +5,7 @@ import argparse
 import os
 import glob
 import time
+import json
 from pathlib import Path
 
 import onnxruntime_genai as og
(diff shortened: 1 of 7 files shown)


examples/python/model-vision.py
```diff
@@ -98,7 +98,24 @@ def run(args: argparse.Namespace):
                 text = args.prompt
             else:
                 text = "What is shown in this image?"
-        prompt += f"{text}{end_tag}{assistant_tag}"
```
baijumeswani, line 101: Shall we remove line 70~80? Do we still need it?

sayanshaw24: nope, missed removing that! updated.

baijumeswani (approved):

examples/python/model-vision.py
```diff
@@ -37,6 +38,7 @@ def run(args: argparse.Namespace):
     model = og.Model(config)
     print("Model loaded")
 
+    tokenizer = og.Tokenizer(model)
     processor = model.create_multimodal_processor()
     tokenizer_stream = processor.create_stream()
```
kunal-vaishnavi: Should we update all references of `tokenizer_stream` with `stream`? It seems unintuitive from a customer's perspective to have a tokenizer object and create the tokenizer stream from the processor instead of creating the tokenizer stream from the tokenizer object.

kunal-vaishnavi:
Can we update the Whisper examples to use the [chat template](https://github.com/microsoft/onnxruntime-extensions/pull/981)?

examples/python/model-vision.py
sayanshaw24: good point! just updated them - not all references of `tokenizer_stream`, but all references where the tokenizer stream was created from the processor instead of the tokenizer (so for vision and phi4-mm cases).

sayanshaw24:
> Can we update the Whisper examples to use the [chat template](https://github.com/microsoft/onnxruntime-extensions/pull/981)?

After discussion with @baijumeswani we decided to not use the whisper chat template functionality for the examples, because specifically for our examples itself, the chat template functionality (just appending message contents) does not make sense - we do not need it. Can give you more context offline.

baijumeswani (approved):

PR #1665, opened 2025-07-30
Update builder.py

lnigam:
Olive Recipe FP16 configuration for model "mistralai/Mistral-7B-Instruct-v0.2" failed. Added check in case config.head_dim is None, calculate head_size instead of None

### DIFF
--- a/src/python/py/models/builder.py
+++ b/src/python/py/models/builder.py
@@ -63,7 +63,7 @@ class Model:
         self.hidden_size = config.hidden_size
         self.num_kv_heads = config.num_key_value_heads if hasattr(config, "num_key_value_heads") else config.multi_query_group_num if hasattr(config, "multi_query_group_num") else config.num_attention_heads
         self.num_attn_heads = config.num_attention_heads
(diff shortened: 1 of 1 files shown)


src/python/py/models/builder.py
```diff
@@ -63,7 +63,7 @@ def __init__(
         self.hidden_size = config.hidden_size
         self.num_kv_heads = config.num_key_value_heads if hasattr(config, "num_key_value_heads") else config.multi_query_group_num if hasattr(config, "multi_query_group_num") else config.num_attention_heads
         self.num_attn_heads = config.num_attention_heads
-        self.head_size = config.head_dim if hasattr(config, "head_dim") else config.hidden_size // config.num_attention_heads
+        self.head_size = config.head_dim if (hasattr(config, "head_dim") and config.head_dim is not None) else config.hidden_size // config.num_attention_heads
```
```suggestion
        self.head_size = config.head_dim if hasattr(config, "head_dim") and config.head_dim is not None else config.hidden_size // config.num_attention_heads
```

PR #1666, opened 2025-07-30
Add support for smollm3

xenova:
cc @guschmue

### DIFF
--- a/src/python/py/models/builder.py
+++ b/src/python/py/models/builder.py
@@ -3619,6 +3619,35 @@ class Gemma3Model(Gemma2Model):
         return super().make_rotary_embedding_caches(cos_cache_name=cos_cache_name, sin_cache_name=sin_cache_name)
 
 
+class SmolLM3Model(LlamaModel):
+    def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
+        super().__init__(config, io_dtype, onnx_dtype, ep, cache_dir, extra_options)
+        self.layer_types = config.layer_types
+        self.no_rope_layers = config.no_rope_layers
+
+    def make_attention(self, layer_id, attention, root_input, **kwargs):
+        # SmolLM3 uses per-layer conditional RoPE and Sliding Window Attention.
+        # So, we temporarily modify the model's attributes before calling the
+        # base `make_attention` method, then restore them immediately after.
+        original_use_rope = self.attention_attrs["use_rope_in_attn"]
+        original_window_size = self.window_size
+
+        # Enable/disable RoPE for the current layer.
+        self.attention_attrs["use_rope_in_attn"] = bool(self.no_rope_layers[layer_id])
+
+        # Set the sliding window size for the current layer.
+        assert self.layer_types[layer_id] in {"sliding_attention", "full_attention"}
+        if self.layer_types[layer_id] == "full_attention":
+            self.window_size = -1
+
+        # Call the original `make_attention` with the temporarily-modified settings.
+        super().make_attention(layer_id, attention, root_input, **kwargs)
+
+        # Restore original values
+        self.attention_attrs["use_rope_in_attn"] = original_use_rope
+        self.window_size = original_window_size
+
+
 def check_extra_options(kv_pairs):
     """
     Check key-value pairs and set values correctly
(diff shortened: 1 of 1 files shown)


src/python/py/models/builder.py
```diff
@@ -3619,6 +3619,35 @@ def make_rotary_embedding_caches(self, **kwargs):
         return super().make_rotary_embedding_caches(cos_cache_name=cos_cache_name, sin_cache_name=sin_cache_name)
 
 
+class SmolLM3Model(LlamaModel):
```
kunal-vaishnavi, line 3643: Can we inherit from `MistralModel` to get the `position_ids` [benefit](https://github.com/microsoft/onnxruntime-genai/blob/95d7a334f76aa19f90e3bd84be6f6da7c21c0dad/src/python/py/models/builder.py#L3074)? Eventually, that benefit needs to be moved into the base class for all models to receive it.

src/python/py/models/builder.py
```diff
@@ -3801,6 +3830,8 @@ def create_model(model_name, input_path, output_dir, precision, execution_provid
             onnx_model = QwenModel(config, io_dtype, onnx_dtype, execution_provider, cache_dir, extra_options)
         elif config.architectures[0] == "Qwen3ForCausalLM":
             onnx_model = Qwen3Model(config, io_dtype, onnx_dtype, execution_provider, cache_dir, extra_options)
+        elif config.architectures[0] == "SmolLM3ForCausalLM":
+          onnx_model = SmolLM3Model(config, io_dtype, onnx_dtype, execution_provider, cache_dir, extra_options)
```
```suggestion
            onnx_model = SmolLM3Model(config, io_dtype, onnx_dtype, execution_provider, cache_dir, extra_options)
```

kunal-vaishnavi:
Thanks for your contribution! Can you also make the following additions for SmolLM-3 in alphabetical order?

- Add the model type that gets generated in the `genai_config.json` here:
https://github.com/microsoft/onnxruntime-genai/blob/95d7a334f76aa19f90e3bd84be6f6da7c21c0dad/src/models/model_type.h#L15
- Add the model type to the [model builder README](https://github.com/microsoft/onnxruntime-genai/blob/main/src/python/py/models/README.md)
- Add the model type to the [repo's main README](https://github.com/microsoft/onnxruntime-genai/blob/main/README.md)

src/python/py/models/builder.py
xenova: Sure thing -- will do 👍 Could you maybe explain what that benefit does? And why isn't it in the base class already? 👀

xenova: Also, the model uses rope in attn for certain layers, and not for others. Is this still applicable?

kunal-vaishnavi: If RoPE isn't used at all in certain layers, then let's keep the inheritance from `LlamaModel` then. The benefit is that for certain EPs that don't fuse the rotary embedding op into an attention op, the position ids subgraph can be created to calculate the position ids needed for the rotary embedding op. It is not in the base class because it was originally created just for Mistral. It may be an outdated subgraph now as it has not been used in a while.

LorenRd:
Hi @xenova I built from source from your repo and tried this: 
`python -m onnxruntime_genai.models.builder -m HuggingFaceTB/SmolLM3-3B` 

But I'm getting:
```
NotImplementedError: The HuggingFaceTB/SmolLM3-3B model is not currently supported.
```
Is that expected?

src/models/model_type.h
```diff
@@ -12,7 +12,7 @@ namespace Generators {
 struct ModelType {
   inline static bool IsLLM(const std::string& model_type) {
     // Large-language model (LLM)
-    static constexpr std::array<std::string_view, 18> LLM = {"chatglm", "decoder", "ernie4_5", "gemma", "gemma2", "gemma3_text", "gpt2", "granite", "llama", "mistral", "nemotron", "olmo", "phi", "phimoe", "phi3", "phi3small", "qwen2", "qwen3"};
+    static constexpr std::array<std::string_view, 18> LLM = {"chatglm", "decoder", "ernie4_5", "gemma", "gemma2", "gemma3_text", "gpt2", "granite", "llama", "mistral", "nemotron", "olmo", "phi", "phimoe", "phi3", "phi3small", "qwen2", "qwen3", "smollm3"};
```
LorenRd: @xenova remember also to increase the array size to 19 
```suggestion
    static constexpr std::array<std::string_view, 19> LLM = {"chatglm", "decoder", "ernie4_5", "gemma", "gemma2", "gemma3_text", "gpt2", "granite", "llama", "mistral", "nemotron", "olmo", "phi", "phimoe", "phi3", "phi3small", "qwen2", "qwen3", "smollm3"};
```

xenova: Ah of course! Thanks for that 😅🤦

PR #1684, opened 2025-08-15
 Update pytorch > 2.6.0

sushraja-msft:
Update pytorch to address
https://nvd.nist.gov/vuln/detail/CVE-2025-32434

### DIFF
--- a/tools/python/model_validation/requirements.txt
+++ b/tools/python/model_validation/requirements.txt
@@ -12,4 +12,4 @@ onnxruntime-genai
 sentencepiece
 pandas
 datasets
-torch
+torch>=2.6.0


tools/python/model_validation/requirements.txt
```diff
@@ -12,4 +12,4 @@ onnxruntime-genai
 sentencepiece
 pandas
 datasets
-torch
\ No newline at end of file
+torch>=2.6.0
```
kunal-vaishnavi, line 15: This should already be using the latest version when a version is not specified.

sushraja-msft: Thanks Kunal, aborting this PR then. I thought the behavior was to not update existing pytorch if requirements.txt uses torch without a version.

PR #1686, opened 2025-08-17
Register custom ops library with absolute path

VishalX:
Use absolute path to register custom ops library.

## Motivation

With new ORT ABI, if `RegisterExecutionProviderLibrary` is used to register EP, the custom op library doesn't load with relative path. 

This should ideally be fixed in ORT, but making the paths absolute explicitly in OGA works too.


VishalX:
@baijumeswani pls review

PR #1690, opened 2025-08-18
Tie embedding weight sharing

jiafatom:
Follow up the idea in https://github.com/microsoft/onnxruntime-genai/pull/1461
Now after GatherBlockQuantized implemented in https://github.com/microsoft/onnxruntime/pull/25214/, we can tie embedding here.

Tested on phi-4-mini-instruct, cpu model size reduces from 5.15 GB to 2.69 GB (47.8% drop)

### DIFF
--- a/src/python/py/models/builder.py
+++ b/src/python/py/models/builder.py
@@ -264,10 +264,11 @@ class Model:
 
         # Quantization-specific variables (INT4, INT8, etc.)
         int4_algo_config = self.make_int4_algo_config(extra_options.get("int4_algo_config", "default"))
+        self.int4_block_size = extra_options.get("int4_block_size", 32)
         self.quant_attrs = {
             "int4": {
                 "accuracy_level": int(extra_options.get("int4_accuracy_level", 4 if self.ep in ["cpu", "webgpu"] else 0)),
-                "block_size": int(extra_options.get("int4_block_size", 32)),
+                "block_size": int(self.int4_block_size),
                 "is_symmetric": extra_options.get("int4_is_symmetric", True),
                 "op_types_to_quantize": extra_options.get("int4_op_types_to_quantize", ("MatMul", )),
                 "nodes_to_exclude": extra_options.get("int4_nodes_to_exclude", []),
@@ -279,6 +280,13 @@ class Model:
             # Create quantized attributes from quantization config
             self.quant_attrs["config"] = config.quantization_config
             self.quant_attrs["use_g_idx"] = config.quantization_config["desc_act"] if "desc_act" in config.quantization_config else False
+
+        self.int4_tied_embeddings = config.tie_word_embeddings if hasattr(config, "tie_word_embeddings") and config.tie_word_embeddings is not None else False
+        self.int4_tied_embeddings = extra_options.get("int4_tied_embeddings", self.int4_tied_embeddings)
+        self.int8_lm_head = extra_options.get("int4_algo_config", "default") in {"k_quant_mixed", "k_quant_last"}
+        if not self.int8_lm_head:
+            # matmul_nbits_quantizer.py has a different naming for default quantization, so lm_head.MatMul.weight_Q{}G{} does not match.
+            self.int4_tied_embeddings = False
 
     def to_str_dtype(self, dtype: ir.DataType) -> str:
         return dtype.name
@@ -1050,13 +1058,28 @@ class Model:
         self.make_add_bias(add, name, root_input, **kwargs)
 
     def make_embedding(self, embedding):
-        weight = "model.embed_tokens.weight"
-        self.make_initializer(embedding, weight, to=self.io_dtype)
-
         basename = "/model/embed_tokens"
-        gather_name = f"{basename}/Gather"
-        gather_output = f"{gather_name}/output_0"
-        self.make_node('Gather', inputs=[weight, 'input_ids'], outputs=[gather_output], name=gather_name)
+        if self.int4_tied_embeddings:
+            gather_name = f"{basename}/GatherBlockQuantized"
+            gather_output = f"{gather_name}/output_0"
+
+            weight_reshape_name = f"{basename}/Reshape"
+            bits = 8 if self.int8_lm_head else 4
+            weight_reshape_inputs = [f"lm_head.MatMul.weight_Q{bits}G{self.int4_block_size}", f"/model/constants/INT64/[{self.vocab_size}, {self.hidden_size}]"]
+            weight_reshape_output = f"{weight_reshape_name}/output_0"
+            # quantized weight dtype is uint8, see here
+            # https://github.com/microsoft/onnxruntime/blob/0c9356cb986fd4cd2c5d510909d31186010ba226/onnxruntime/python/tools/quantization/neural_compressor/weight_only.py#L73
+            self.make_reshape(weight_reshape_name, weight_reshape_inputs, dtype=ir.DataType.UINT8, shape=['vocab_size', 'hidden_size'])
+
+            self.make_node('GatherBlockQuantized', inputs=[weight_reshape_output, 'input_ids', 'lm_head.MatMul.weight_scale', 'lm_head.MatMul.weight_zp'], outputs=[gather_output], name=gather_name, domain="com.microsoft", bits=bits, block_size=int(self.int4_block_size))
+        else:
+            weight = "model.embed_tokens.weight"
+            self.make_initializer(embedding, weight, to=self.io_dtype)
+
+            gather_name = f"{basename}/Gather"
+            gather_output = f"{gather_name}/output_0"
+            self.make_node('Gather', inputs=[weight, 'input_ids'], outputs=[gather_output], name=gather_name)
+
         self.make_value(gather_output, self.io_dtype, shape=['batch_size', 'sequence_length', self.hidden_size])
 
         if self.embed_attrs["scale"] != 1:
@@ -4157,7 +4180,7 @@ def check_extra_options(kv_pairs):
     """
     bools = [
         "int4_is_symmetric", "exclude_embeds", "exclude_lm_head", "include_hidden_states", "enable_cuda_graph",
-        "use_8bits_moe", "use_qdq", "use_webgpu_fp32", "use_cuda_bf16",
+        "use_8bits_moe", "use_qdq", "use_webgpu_fp32", "use_cuda_bf16", "int4_tied_embeddings"
     ]
     for key in bools:
         if key in kv_pairs:
@@ -4447,6 +4470,8 @@ def get_args():
                     Currently supported options are: 'default', 'rtn', 'k_quant_mixed', 'k_quant_last'.
                     k_quant_mixed = k_quant algorithm with mixed precision (int4 + int8).
                     k_quant_last = k_quant algorithm where only the last MatMul (/lm_head/MatMul) is quantized as int8. Other MatMuls are quantized as int4.
+                int4_tied_embeddings = Enable weight sharing for quantization. Default is false.
+                    Use this option when you want to share the weights in the embedding and unembedding.
                 num_hidden_layers = Manually specify the number of layers in your ONNX model.
                     Used for unit testing purposes.
                 filename = Filename for ONNX model (default is 'model.onnx').


hanbitmyths:
We need to check "tie_word_embedings" in config.json and tie embedding only when it's true. We can use int4_weight_sharing param for force tying embedding.

kunal-vaishnavi:
Can you update the PR description to show the memory savings with tied embeddings for other model architectures that are supported via the model builder as well (e.g. LLaMA-3.3, Mistral v0.3, Phi-4, Gemma-3, gpt-oss)? It would be great to quantify the memory reduction for popular models (ex: Gemma-3 and gpt-oss use large embeddings).

src/python/py/models/builder.py
```diff
@@ -46,6 +46,7 @@ def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
         self.num_attn_heads = config.num_attention_heads
         self.head_size = config.head_dim if hasattr(config, "head_dim") and config.head_dim is not None else config.hidden_size // config.num_attention_heads
         self.num_layers = int(extra_options["num_hidden_layers"]) if "num_hidden_layers" in extra_options else config.num_hidden_layers if hasattr(config, "num_hidden_layers") else config.num_layers
+        self.tie_word_embeddings = config.tie_word_embeddings if hasattr(config, "tie_word_embeddings") else False
```
kunal-vaishnavi: Should we add `config.tie_word_embeddings is not None` to the check? What is the value for this attribute for models where `tie_word_embeddings` is not in the config and where tied embeddings does not work (e.g. the weights differ for the embedding and unembedding)?

src/python/py/models/builder.py
```diff
@@ -4447,6 +4470,7 @@ def get_args():
                     Currently supported options are: 'default', 'rtn', 'k_quant_mixed', 'k_quant_last'.
                     k_quant_mixed = k_quant algorithm with mixed precision (int4 + int8).
                     k_quant_last = k_quant algorithm where only the last MatMul (/lm_head/MatMul) is quantized as int8. Other MatMuls are quantized as int4.
+                int4_weight_sharing = Enable weight sharing for quantization. Default is false.
```
kunal-vaishnavi: Can we add a new line to describe what this does for users who are unaware? Something like the following should work: "Use this option when you want to share the weights in the embedding and unembedding".

kunal-vaishnavi: Should we rename this to `int4_tied_embeddings` instead?

src/python/py/models/builder.py
jiafatom: Will add `is not None` check. tied embedding is the default setting when we do the quantization. If it is not working, then by default we are safe. The user should not set the weight sharing at this case. The user should only use weight sharing overwrite if he knows what he is doing...

src/python/py/models/builder.py
jiafatom: Will change.

jiafatom: Will do.

src/python/py/models/builder.py
```diff
@@ -280,6 +282,12 @@ def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
             self.quant_attrs["config"] = config.quantization_config
             self.quant_attrs["use_g_idx"] = config.quantization_config["desc_act"] if "desc_act" in config.quantization_config else False
 
+        self.is_lm_head_int8 = True if extra_options.get("int4_algo_config", "default") in {"k_quant_mixed", "k_quant_last"} else False
+        self.int4_weight_sharing = extra_options.get("int4_weight_sharing", self.tie_word_embeddings)
```
kunal-vaishnavi: It seems there are two variables that are doing the same thing, `self.tie_word_embeddings` and `self.int4_weight_sharing`. Should we combine the conditions to set the two variables into setting one variable?

src/python/py/models/builder.py
```diff
@@ -280,6 +282,12 @@ def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
             self.quant_attrs["config"] = config.quantization_config
             self.quant_attrs["use_g_idx"] = config.quantization_config["desc_act"] if "desc_act" in config.quantization_config else False
 
+        self.is_lm_head_int8 = True if extra_options.get("int4_algo_config", "default") in {"k_quant_mixed", "k_quant_last"} else False
+        self.int4_tied_embeddings = extra_options.get("int4_tied_embeddings", self.tie_word_embeddings)
+        if not self.is_lm_head_int8:
+            print("WARNING: lm_head MatMul is not quantized on int8, disabling int4 weight sharing.")
```
kunal-vaishnavi, line 288: We should be able to use tied embeddings for int4 or int8 because both `GatherBlockQuantized` and `MatMulNBits` support both precisions. Can we enable that here?

I believe an update will need to be made in ORT to quantize Gather to GatherBlockQuantized for int8. If I remember correctly, the `matmul_nbits_quantizer.py` file only quantizes a Gather to GatherBlockQuantized for int4.

src/python/py/models/builder.py
```diff
@@ -280,6 +282,12 @@ def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
             self.quant_attrs["config"] = config.quantization_config
             self.quant_attrs["use_g_idx"] = config.quantization_config["desc_act"] if "desc_act" in config.quantization_config else False
 
+        self.is_lm_head_int8 = True if extra_options.get("int4_algo_config", "default") in {"k_quant_mixed", "k_quant_last"} else False
+        self.int4_tied_embeddings = extra_options.get("int4_tied_embeddings", self.tie_word_embeddings)
+        if not self.is_lm_head_int8:
```
kunal-vaishnavi: We should be able to use tied embeddings for when the embedding and unembedding are either both int4 or both int8 because both `GatherBlockQuantized` and `MatMulNBits` support both precisions. Can we enable that here?

I believe an update will need to be made in ORT to quantize Gather to GatherBlockQuantized for int8. If I remember correctly, the `matmul_nbits_quantizer.py` file only quantizes a Gather to GatherBlockQuantized for int4.

src/python/py/models/builder.py
```diff
@@ -1050,13 +1058,28 @@ def make_packed_add(self, q_add, k_add, v_add, name, root_input, **kwargs):
...
-
         basename = "/model/embed_tokens"
-        gather_name = f"{basename}/Gather"
-        gather_output = f"{gather_name}/output_0"
-        self.make_node('Gather', inputs=[weight, 'input_ids'], outputs=[gather_output], name=gather_name)
+        if self.int4_tied_embeddings :
+            gather_name = f"{basename}/GatherBlockQuantized"
+            gather_output = f"{gather_name}/output_0"
+
+            weight_reshape_name = f"{basename}/Gather_Reshape"
```
```suggestion
            weight_reshape_name = f"{basename}/Reshape"
```

src/python/py/models/builder.py
```diff
@@ -1050,13 +1058,28 @@ def make_packed_add(self, q_add, k_add, v_add, name, root_input, **kwargs):
...
 
     def make_embedding(self, embedding):
-        weight = "model.embed_tokens.weight"
-        self.make_initializer(embedding, weight, to=self.io_dtype)
-
         basename = "/model/embed_tokens"
-        gather_name = f"{basename}/Gather"
-        gather_output = f"{gather_name}/output_0"
-        self.make_node('Gather', inputs=[weight, 'input_ids'], outputs=[gather_output], name=gather_name)
+        if self.int4_tied_embeddings :
```
```suggestion
        if self.int4_tied_embeddings:
```

xenova:
Very nice! Definitely a feature I've been looking forward to 🔥

src/python/py/models/builder.py
```diff
@@ -1050,13 +1058,28 @@ def make_packed_add(self, q_add, k_add, v_add, name, root_input, **kwargs):
...
         basename = "/model/embed_tokens"
-        gather_name = f"{basename}/Gather"
-        gather_output = f"{gather_name}/output_0"
-        self.make_node('Gather', inputs=[weight, 'input_ids'], outputs=[gather_output], name=gather_name)
+        if self.int4_tied_embeddings:
+            gather_name = f"{basename}/GatherBlockQuantized"
+            gather_output = f"{gather_name}/output_0"
+
+            weight_reshape_name = f"{basename}/Reshape"
+            bits = 8 if self.is_lm_head_int8 else 4
```
kunal-vaishnavi: The `self.is_lm_head_int8` variable doesn't exist after the latest changes

src/python/py/models/builder.py
```diff
@@ -280,6 +281,13 @@ def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
             self.quant_attrs["config"] = config.quantization_config
             self.quant_attrs["use_g_idx"] = config.quantization_config["desc_act"] if "desc_act" in config.quantization_config else False
 
+        self.int4_tied_embeddings = config.tie_word_embeddings if hasattr(config, "tie_word_embeddings") and config.tie_word_embeddings is not None else False
+        self.int4_tied_embeddings = extra_options.get("int4_tied_embeddings", self.int4_tied_embeddings)
+        is_kquant = True if extra_options.get("int4_algo_config", "default") in {"k_quant_mixed", "k_quant_last"} else False
+        if not is_kquant:
```
kunal-vaishnavi: Other cases where tied embeddings cannot be used include 1) if the embedding is excluded from the decoder (i.e. `inputs_embeds` is the input instead of `input_ids`) or 2) if the unembedding is excluded from the decoder (i.e. `hidden_states` is the output instead of `logits`). We should check for those scenarios.

jiafatom: Is `tie_word_embeddings=True` for these two cases? If not, then by default we will not do tie embedding. If the user wants to overwrite, s/he should only proceed when they know what they are doing.

src/python/py/models/builder.py
```diff
@@ -280,6 +281,14 @@ def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
             self.quant_attrs["config"] = config.quantization_config
             self.quant_attrs["use_g_idx"] = config.quantization_config["desc_act"] if "desc_act" in config.quantization_config else False
 
+        self.int4_tied_embeddings = config.tie_word_embeddings if hasattr(config, "tie_word_embeddings") and config.tie_word_embeddings is not None else False
+        self.int4_tied_embeddings = extra_options.get("int4_tied_embeddings", self.int4_tied_embeddings)
+        is_kquant = True if extra_options.get("int4_algo_config", "default") in {"k_quant_mixed", "k_quant_last"} else False
```
```suggestion
        self.int8_lm_head = True if extra_options.get("int4_algo_config", "default") in {"k_quant_mixed", "k_quant_last"} else False
```

src/python/py/models/builder.py
```diff
@@ -280,6 +281,14 @@ def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
...
+        self.is_lm_head_int8 = is_kquant
+        if not is_kquant:
```
```suggestion
        if not self.int8_lm_head:
```

src/python/py/models/builder.py
```diff
@@ -1050,13 +1059,28 @@ def make_packed_add(self, q_add, k_add, v_add, name, root_input, **kwargs):
...
         basename = "/model/embed_tokens"
-        gather_name = f"{basename}/Gather"
-        gather_output = f"{gather_name}/output_0"
-        self.make_node('Gather', inputs=[weight, 'input_ids'], outputs=[gather_output], name=gather_name)
+        if self.int4_tied_embeddings:
+            gather_name = f"{basename}/GatherBlockQuantized"
+            gather_output = f"{gather_name}/output_0"
+
+            weight_reshape_name = f"{basename}/Reshape"
+            bits = 8 if self.is_lm_head_int8 else 4
```
```suggestion
            bits = 8 if self.int8_lm_head else 4
```

src/python/py/models/builder.py
jiafatom: (1) We only `make_embedding` when ` if not self.exclude_embeds:` so this case is handled. (2)  I checked whether `lm_head` is in the model and then do `            self.int4_tied_embeddings = (self.lm_head_weight_shape is not None
                                         and (embedding.shape == self.lm_head_weight_shape)
            )`

src/python/py/models/builder.py
```diff
@@ -280,6 +281,13 @@ def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
             self.quant_attrs["config"] = config.quantization_config
             self.quant_attrs["use_g_idx"] = config.quantization_config["desc_act"] if "desc_act" in config.quantization_config else False
 
+        self.int4_tied_embeddings = config.tie_word_embeddings if hasattr(config, "tie_word_embeddings") and config.tie_word_embeddings is not None else False
+        self.int4_tied_embeddings = extra_options.get("int4_tied_embeddings", self.int4_tied_embeddings)
+        self.int8_lm_head = True if extra_options.get("int4_algo_config", "default") in {"k_quant_mixed", "k_quant_last"} else False
```
```suggestion
        self.int8_lm_head = extra_options.get("int4_algo_config", "default") in {"k_quant_mixed", "k_quant_last"}
```

kunal-vaishnavi (approved):

PR #1691, opened 2025-08-19
Add support for inference using EP context model

thevishalagarwal:
This PR adds support for inference using compiled EP context model by updating genai config. 

Added a new `ep_context` object inside `model` object in genai config
```json
"ep_context": {
  "enable": false,
}
```

The expected workflow is as follows
* compile EP context model ahead-of-time and use it only during inference by updating genai config
* if `enable` then uses EP context model for inference
* EP context model should be in the same folder as the base model and the name should be  `{filename}_ctx.onnx`. `{filename}` is the name of the base model name set in genai config
* if EP context model not found, inference fallbacks to using original base ONNX at `config_path / model.onnx`
* supports multiple models in a pipeline, e.g. VLMs. Each EP context model should be saved as  `{filename}_ctx.onnx`

### DIFF
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -692,8 +692,6 @@ struct EpContext_Element : JSON::Element {
   void OnValue(std::string_view name, JSON::Value value) override {
     if (name == "enable") {
       v_.enable = JSON::Get<bool>(value);
-    } else if (name == "filename") {
-      v_.filename = JSON::Get<std::string_view>(value);
     } else {
       throw JSON::unknown_value_error{};
     }
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -692,8 +692,8 @@ struct EpContext_Element : JSON::Element {
   void OnValue(std::string_view name, JSON::Value value) override {
     if (name == "enable") {
       v_.enable = JSON::Get<bool>(value);
-    } else if (name == "filepath") {
-      v_.filepath = JSON::Get<std::string_view>(value);
+    } else if (name == "filename") {
+      v_.filename = JSON::Get<std::string_view>(value);
     } else {
       throw JSON::unknown_value_error{};
     }
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -686,6 +686,23 @@ struct Embedding_Element : JSON::Element {
   EmbeddingOutputs_Element outputs_{v_.outputs};
 };
 
+struct EpContext_Element : JSON::Element {
+  explicit EpContext_Element(Config::Model::EpContext& v) : v_(v) {}
+
+  void OnValue(std::string_view name, JSON::Value value) override {
+    if (name == "enable") {
+      v_.enable = JSON::Get<bool>(value);
+    } else if (name == "filepath") {
+      v_.filepath = JSON::Get<std::string_view>(value);
+    } else {
+      throw JSON::unknown_value_error{};
+    }
+  }
+
+  private:
+    Config::Model::EpContext& v_;
+};
+
 struct Model_Element : JSON::Element {
   explicit Model_Element(Config::Model& v) : v_{v} {}
 
@@ -733,6 +750,10 @@ struct Model_Element : JSON::Element {
     if (name == "speech") {
       return speech_;
     }
+    if (name == "ep_context") {
+      return ep_context_;
+    }
+
     throw JSON::unknown_value_error{};
   }
 
@@ -744,6 +765,7 @@ struct Model_Element : JSON::Element {
   Vision_Element vision_{v_.vision};
   Embedding_Element embedding_{v_.embedding};
   Speech_Element speech_{v_.speech};
+  EpContext_Element ep_context_{v_.ep_context};
 };
 
 struct Search_Element : JSON::Element {
--- a/src/config.h
+++ b/src/config.h
@@ -60,9 +60,6 @@ struct Config {
     static constexpr std::string_view EncoderOutputsName = "encoder_outputs";
     static constexpr std::string_view EncoderAttentionMaskName = "encoder_attention_mask";
 
-    // EP context model default file name
-    static constexpr std::string_view EpContextFileName = "model_ctx.onnx";
-
   };
 
   fs::path config_path;  // Path of the config directory
@@ -111,9 +108,8 @@ struct Config {
 
     struct EpContext {
       // use EP context model for inference
+      // it assumes the EP context model is in the same directory as the base model with name as "filename_ctx.onnx"
       bool enable{false};
-      // file name for EP context model. Default is "model_ctx.onnx"
-      std::string filename{""};
     } ep_context;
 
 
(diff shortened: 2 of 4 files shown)


natke:
@thevishalagarwal Can you make use of EpContext enable and filepath in the sesssion options for this?

https://github.com/microsoft/onnxruntime-genai/blob/main/src/config.h#L80

thevishalagarwal:
> @thevishalagarwal Can you make use of EpContext enable and filepath in the sesssion options for this?

In this PR, the main goal to add EP context inference is to reduce model load time. The expectation is to have EP context model compiled AOT and update the config to use it.

Can't use these session options because this path just generates the EP context model. If EP context model is already present, then inference fails. Also, session options path is the old way. The new recommended path is to use [Compile APIs](https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/core/session/compile_api.h) to generate EP context models which was released in ORT 1.22.

src/python/py/models/builder.py
```diff
@@ -415,6 +415,10 @@ def make_genai_config(self, model_name_or_path, extra_kwargs, out_dir):
                 "pad_token_id": config.pad_token_id if hasattr(config, "pad_token_id") and config.pad_token_id is not None else config.eos_token_id[0] if isinstance(config.eos_token_id, list) else config.eos_token_id,
                 "type": self.model_type[ : self.model_type.find("For") if "For" in self.model_type else len(self.model_type)].lower(),
                 "vocab_size": self.vocab_size,
+                "ep_context": {
```
kunal-vaishnavi: This section should be optionally added into the GenAI config. Let's follow the same approach used for adding the sliding window into the config.

https://github.com/microsoft/onnxruntime-genai/blob/a2bbbc662206bdb254feda075bfb231c9a3813b4/src/python/py/models/builder.py#L464-L465

src/config.cpp
```diff
@@ -686,6 +686,23 @@ struct Embedding_Element : JSON::Element {
   EmbeddingOutputs_Element outputs_{v_.outputs};
 };
 
+struct EpContext_Element : JSON::Element {
+  explicit EpContext_Element(Config::Model::EpContext& v) : v_(v) {}
+
+  void OnValue(std::string_view name, JSON::Value value) override {
+    if (name == "enable") {
+      v_.enable = JSON::Get<bool>(value);
+    } else if (name == "filepath") {
```
kunal-vaishnavi: Can the EP context model always be put in the same folder as the other files needed to run with ORT GenAI? Then we can rename this option to `filename` and be consistent with the other sections in the GenAI config.

src/python/py/models/builder.py
thevishalagarwal: We can do this for TRT RTX EP to begin with but this can be a common feature across EPs. I would want other EPs to use EP context as well

src/config.cpp
thevishalagarwal: Make sense. Updated it

src/models/model.cpp
```diff
@@ -920,7 +920,23 @@ std::unique_ptr<OrtSession> Model::CreateSession(OrtEnv& ort_env, const std::str
   }
 
   // Otherwise, load the model from the file system
-  return OrtSession::Create(ort_env, (config_->config_path / fs::path(model_filename)).c_str(), session_options);
+  auto model_path_ = config_->config_path / fs::path(model_filename);
+  
+  if(config_->model.ep_context.enable) {
```
baijumeswani, line 925: Is the presence of this structure enough to determine whether to enable loading the ep context model?

src/config.h
```diff
@@ -105,6 +109,14 @@ struct Config {
     int vocab_size{};
     int context_length{};
 
+    struct EpContext {
+      // use EP context model for inference
+      bool enable{false};
+      // file name for EP context model. Default is "model_ctx.onnx"
+      std::string filename{""};
```
baijumeswani: Should this be empty? Or should it be the default value?

```suggestion
      std::string filename{EpContextFileName};
```

src/models/model.cpp
```diff
@@ -920,7 +920,23 @@ std::unique_ptr<OrtSession> Model::CreateSession(OrtEnv& ort_env, const std::str
...
+    auto model_ctx_path_ = config_->config_path / fs::path(std::string(Config::Defaults::EpContextFileName));
```
baijumeswani: Use local variable naming convention (if this variable is needed).

```suggestion
    const auto model_ctx_path = config_->config_path / fs::path(std::string(Config::Defaults::EpContextFileName));
```

src/config.cpp
```diff
@@ -733,6 +750,10 @@ struct Model_Element : JSON::Element {
     if (name == "speech") {
       return speech_;
     }
+    if (name == "ep_context") {
+      return ep_context_;
+    }
```
baijumeswani, line 753: If the model folder has multiple models, how would each model's ep context variant be determined? Or is that path not supported?

baijumeswani: Alternatively, can we just use the `filename` inside the each struct to point to the ep context model (if the intention is to only support aot compilation)?

src/models/model.cpp
thevishalagarwal: Yes, if `enable` is true

src/config.h
thevishalagarwal: can be default value

src/models/model.cpp
thevishalagarwal: Done

src/config.cpp
thevishalagarwal: Good catch - we should support multiple models case as well. But instead of supporting it at per-model type, we can consider ep context enable for a "pipeline" (`model` struct) and then use the `filename_ctx.onnx` for the context model name. What do you think?

thevishalagarwal:
@baijumeswani Updated the PR to support multiple models usecase. EP context models should be saved as `{filename}_ctx.onnx` in the base model folder. 
EP context cannot be enabled per-model basis. It will be enabled for all models in the pipeline

src/config.cpp
baijumeswani: Can we have a scenario where some models in the model folder need to be ep context models while others don't?

thevishalagarwal: That will be fine. If the EP context model is not present, it will use the base model

thevishalagarwal:
Had a discussion with @baijumeswani for EP context inference in ORT GenAI. We would like to integrate ORT ModelCompilationOptions for model compilation and then do compilation on first run, if ep context is enabled. On subsequent runs, context model will be used for faster load time. This will be planned and integrated soon.

For now, we will support AOT offline compilation and expected it to be inside the genai model folder. For inference, we can simply change the model `filename` to `model_ctx.onnx` in `genai_config`, instead of adding new ep context options. So closing this PR.

PR #1693, opened 2025-08-20
Update how Hugging Face's config files are processed

kunal-vaishnavi:
### Description

This PR updates how Hugging Face's config files are processed while creating the `genai_config.json` file for ONNX Runtime GenAI.

### Motivation and Context

The `config.json` file typically stores model attributes while the `generation_config.json` file typically stores search attributes. Some fields can overlap between the two files, and the latter file typically contains more detailed information than the former file for a particular attribute (e.g. the [`config.json`](https://huggingface.co/google/gemma-3-270m-it/blob/main/config.json#L10) file for `google/gemma-3-270m-it` says that `eos_token_id = 1` but the [`generation_config.json`](https://huggingface.co/google/gemma-3-270m-it/blob/main/generation_config.json#L4) file for the same model says that `eos_token_id = [1, 106]`).

To ensure the right attributes are stored, we compare the values stored in `generation_config.json` with the traditional default values loaded in `config.json`. If they do not match, then the values in `config.json` are rewritten with the `generation_config.json` values.

With this PR, models such as [`google/gemma-3-270m-it`](https://huggingface.co/google/gemma-3-270m-it) and [`janhq/Jan-v1-4B`](https://huggingface.co/janhq/Jan-v1-4B) can be set with their default search values.

### DIFF
--- a/src/python/py/models/builder.py
+++ b/src/python/py/models/builder.py
@@ -374,10 +374,26 @@ class Model:
         self.past_present_share_buffer = self.attention_attrs["op_type"] == "GroupQueryAttention"
 
     def make_genai_config(self, model_name_or_path, extra_kwargs, out_dir):
+        # Create config with attributes from config.json and generation_config.json (if latter file exists)
+        config = AutoConfig.from_pretrained(model_name_or_path, token=self.hf_token, trust_remote_code=True, **extra_kwargs)
         try:
-            config = GenerationConfig.from_pretrained(model_name_or_path, token=self.hf_token, trust_remote_code=True, **extra_kwargs)
+            # Override search attributes in config based on values in generation_config.json
+            gen_config = GenerationConfig.from_pretrained(model_name_or_path, token=self.hf_token, trust_remote_code=True, **extra_kwargs)
+            defaults = {
+                "bos_token_id": None,
+                "do_sample": False,
+                "eos_token_id": None,
+                "pad_token_id": None,
+                "temperature": 1.0,
+                "top_k": 50,
+                "top_p": 1.0,
+            }
+            for key, default_val in defaults.items():
+                val = getattr(gen_config, key)
+                if val != default_val:
+                    setattr(config, key, getattr(gen_config, key))
         except:
-            config = AutoConfig.from_pretrained(model_name_or_path, token=self.hf_token, trust_remote_code=True, **extra_kwargs)
+            pass
 
         inputs = dict(zip(self.input_names, self.input_names))
         inputs.update({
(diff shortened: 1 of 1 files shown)


src/python/py/models/builder.py
```diff
@@ -374,10 +374,26 @@ def make_attention_init(self):
...
     def make_genai_config(self, model_name_or_path, extra_kwargs, out_dir):
+        # Create config with attributes from config.json and generation_config.json (if latter file exists)
+        config = AutoConfig.from_pretrained(model_name_or_path, token=self.hf_token, trust_remote_code=True, **extra_kwargs)
         try:
-            config = GenerationConfig.from_pretrained(model_name_or_path, token=self.hf_token, trust_remote_code=True, **extra_kwargs)
+            # Override search attributes in config based on values in generation_config.json
+            gen_config = GenerationConfig.from_pretrained(model_name_or_path, token=self.hf_token, trust_remote_code=True, **extra_kwargs)
+            defaults = {
+                "bos_token_id": None,
+                "do_sample": False,
```
xiaoyu-work, line 384: If `temperature`, `top_k` and `top_p` will be ignored automatically if `do_sample` is false? Should we check `do_sample` originally from gen_config to decide whether to add those?

src/python/py/models/builder.py
```diff
@@ -4185,10 +4204,6 @@ def check_extra_options(kv_pairs):
         # 'include_hidden_states' is for when 'hidden_states' are outputted and 'logits' are outputted
         raise ValueError("Both 'exclude_lm_head' and 'include_hidden_states' cannot be used together. Please use only one of them at once.")
 
-    # NvTensorRtRtx EP requires Opset 21, so force use_qdq which controls it.
-    if args.execution_provider == "NvTensorRtRtx":
```
xiaoyu-work, line 4189: The PR https://github.com/microsoft/onnxruntime-genai/pull/1692 had been merged.

src/python/py/models/builder.py
kunal-vaishnavi: This will read the attributes from the `generation_config.json` and store them in the `config` object regardless of whether `do_sample` is true or false. We leave it up to ORT GenAI to decide how to use `do_sample`, `temperature`, `top_k`, and `top_p` based on their values (e.g. ignore the latter three if `do_sample = false` vs. use the latter three if `do_sample = true`).

nenad1002: Are all the values overriden?

nenad1002 (approved):

src/python/py/models/builder.py
kunal-vaishnavi: Only the values in the `defaults` dictionary are overridden if necessary.

Issue #1682, opened 2025-08-14
Question about attention_mask / position_ids initialization


RyanMetcalfeInt8:
Wanted to ask a question here if that's okay. I've been trying to track down some accuracy issues we're encountering on my side when we I ORT GenAI (as compared with OpenVINO GenAI) for certain models. 

I've found that the attention_mask is prepared slightly differently between the two frameworks. In ORT GenAI, it initializes the attention_mask here: https://github.com/microsoft/onnxruntime-genai/blob/main/src/models/position_inputs.cpp#L245

Here, the attention_mask entries are set to 1, except for tokens that are equal to the model's pad_token_id -- in which case they're set to 0.

In my case, the input_id's don't have any padding added by the tokenizer but may have some EOS tokens from chat templating, which happen to be equal to the pad_token_id -- and so they are set to 0 which seems to affect accuracy. 

As far as I understand, the attention_mask should be set to 1 for all tokens except for explicit pad tokens that may added by the tokenizer (i.e. to left/right pad to a particular max length, etc.).

For example, with native pytorch transformers, the tokenizer will generate an attention mask of all 1's.

The same kind of logic is also used for initializing position_ids: https://github.com/microsoft/onnxruntime-genai/blob/main/src/models/position_inputs.cpp#L183

I think it may be a bug, but I wanted to get some feedback on perhaps what the intent is of these initialization functions are. Thanks!

kunal-vaishnavi:
Closing this issue since it has been fixed [here](https://github.com/microsoft/onnxruntime-genai/pull/1694)

PR #1694, opened 2025-08-20
fix pad vs. eos token misidentification

aciddelgado:
Addresses this issue: #1682 regarding the misidentification of eos tokens as padding tokens. For models which share EOS and pad token, each EOS token present in the input would be identified as a pad token, leading to incorrect attention_mask and later parity issues. This PR fixes this issue by adding a special case for batch_size 1 (where input padding is unnecessary and assumed to not be there) as well as iterating through the input backwards in batch_size > 1 scenarios, as to not misinterpret any EOS tokens which are not at the end of the sequence.

### DIFF
--- a/test/c_api_tests.cpp
+++ b/test/c_api_tests.cpp
@@ -427,7 +427,7 @@ TEST(CAPITests, EndToEndPhi) {
   auto model = OgaModel::Create(PHI2_PATH);
   auto tokenizer = OgaTokenizer::Create(*model);
 
-  const char* input_string = "This is a test.";
+  const char* input_string = "This is a test.<|endoftext|>";
   auto input_sequence = OgaSequences::Create();
   tokenizer->Encode(input_string, *input_sequence);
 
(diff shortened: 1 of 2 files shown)


test/c_api_tests.cpp
```diff
@@ -447,9 +447,9 @@ TEST(CAPITests, EndToEndPhi) {
 
   // Verify outputs match expected outputs
   std::vector<int32_t> expected_output{
-      1212, 318, 257, 1332, 13, 198, 50280, 2, 16926, 1330, 1635, 10412, 6617, 278,
-      6335, 32994, 21857, 13849, 38665, 82, 21815, 1108, 9557, 40755, 27446, 2417,
-      6381, 6, 7131, 6, 14870, 31314, 21411, 46009, 3974, 82, 1039, 889, 263, 3684};
+      1212, 318, 257, 1332, 13, 50256, 198, 198, 198, 198, 4010, 4420, 43168, 15666,
```
sayanshaw24, line 489: are these from HF? any chance you can add a comment above this with the model id from transformers and why the eos/padding changes affects the input ids in this way (for instance the changes to non-eos/padding ids, those should intuitively not change right) - so in the future we know why this is expected.

test/c_api_tests.cpp
```diff
@@ -427,7 +427,7 @@ TEST(CAPITests, EndToEndPhi) {
   auto model = OgaModel::Create(PHI2_PATH);
   auto tokenizer = OgaTokenizer::Create(*model);
 
-  const char* input_string = "This is a test.";
+  const char* input_string = "This is a test.<|endoftext|>";
```
kunal-vaishnavi, line 469: Is this to get the EOS token id in the tokenized input?

aciddelgado: Yes

test/c_api_tests.cpp
aciddelgado: I had some trouble understanding your comment but basically this should be appending an EOS token at the end which should a) change the output (at a rudimentary level the model can interpret this as a change of turn in conversation) and b) this eos token would be masked to 0 rather than 1 without this pr, which is a mistake as the EOS token is valid and should not be masked out.

sayanshaw24: awesome, thanks! by HF i just meant where are we getting the expected input ids from (for instance from HuggingFace transformers, model id ___) could you just add a quick comment above the specific test along with this context you gave so we know what changed in the future if/when editing this file?

PR #1695, opened 2025-08-21
make custom_ops_library support a file name for VitisAI

vortex-captain:


VishalX:
For Linux, would PR https://github.com/microsoft/onnxruntime-genai/pull/1686 be helpful?

chrisdMSFT (commented):
Great 'point' fix, which we can use to unblock Vitis now, but we need to work on a wholistic fix that supports all EPs that may use the custom_opts feature.

vortex-captain:
> Great 'point' fix, which we can use to unblock Vitis now, but we need to work on a wholistic fix that supports all EPs that may use the custom_opts feature.

Closed this PR as the wholistic fix is in progress.

PR #1705, opened 2025-08-26
Improve TopK CUDA execution path while sampling and when vocab size is sufficiently large

hariharans29:
The 20B GPT OSS model has a vocab size of 200k+ and while sampling (with no batching), the top_k sampling boils down to finding the top_k elements in a large array (batch_size == 1) which is a hard-problem to parallelize and optimally use hardware resources on modern GPUs. 

A profile of the 20B GPT OSS model on A100 with default top_k parameter (top_k = 50) shows the top_k kernel as a huge bottleneck (about 50%) of the forward pass as can be seen from this nsys snapshot:

image

As can be seen from the kernel launch parameters, only one threadblock (since batch_size = 1 in sampling) is used and 256 threads are utilized to search top_k elements from an array of about 200k+elements

image

This leads to severe under-utilization of the hardware.

To mitigate the problem, thje following were tried:

1) Do distributed "top_k" search:
    Use multiple threadblocks to each individually pick top_k elements from their own individual shards of the array and do a final "reduction" tro pick the final top_k elements. This gave about 5% speed-up. The bottleneck was the reduction step.

2) Use sort and pick top_k
    For sufficiently large arrays, sorting using CUB's device fragmented radix sort (which has clever bit manipulation tricks to perform radix sort for `floats` and `halfs`  uses the hardware at hand much more optimally and outperforms any tradional top_k picking algorithms on device. See https://github.com/pytorch/pytorch/issues/22812 and https://github.com/pytorch/pytorch/pull/68632.

Hence, the final approach is to go with option 2 based on a tunable heuristic.

Here are the before and after sampling TPS numbers for 20B GPT OSS on A100 for this optimization for the same prompt (for top_k = 50 which seems to be FoundryLocal's default config) - about **35%** more sampling throughput
 
image

### DIFF
--- a/src/cuda/cuda_sampling.cu
+++ b/src/cuda/cuda_sampling.cu
@@ -548,15 +548,26 @@ void GetTopKSubset(SamplingData* data, cudaStream_t stream, float* scores_in, fl
                              k,                          \
                              temperature);
 
-  if (k <= 4) {
+  bool use_sort_and_pick_topk = false;
+
+  // TODO(hasesh): Tweak this heuristic by performing more benchmarks.
+  // It is likely that the sort and pick top_k approach works well for
+  // smallish batch_sizes on most hardware. For now, we limit it to
+  // sampling (batch_size == 1) and when vocab_size is "pretty large"
+  // (also needs benchmarking to identify cut-off).
+  if (batch_size == 1 && vocab_size >= 100000) {
+    use_sort_and_pick_topk = false;
+  }
+
+  if (!use_sort_and_pick_topk && k <= 4) {
     GetTopK(4);
-  } else if (k <= 8) {
+  } else if (!use_sort_and_pick_topk && k <= 8) {
     GetTopK(8);
-  } else if (k <= 16) {
+  } else if (!use_sort_and_pick_topk && k <= 16) {
     GetTopK(16);
-  } else if (k <= 32) {
+  } else if (!use_sort_and_pick_topk && k <= 32) {
     GetTopK(32);
-  } else if (k <= 64) {
+  } else if (!use_sort_and_pick_topk && k <= 64) {
     GetTopK(64);
   } else {
     // In this case, we need vocab_size as stride for indices_out.
--- a/src/cuda/cuda_sampling.cu
+++ b/src/cuda/cuda_sampling.cu
@@ -556,7 +556,7 @@ void GetTopKSubset(SamplingData* data, cudaStream_t stream, float* scores_in, fl
   // sampling (batch_size == 1) and when vocab_size is "pretty large"
   // (also needs benchmarking to identify cut-off).
   if (batch_size == 1 && vocab_size >= 100000) {
-    use_sort_and_pick_topk = false;
+    use_sort_and_pick_topk = true;
   }
 
   if (!use_sort_and_pick_topk && k <= 4) {
--- a/test/sampling_tests.cpp
+++ b/test/sampling_tests.cpp
@@ -649,4 +649,47 @@ TEST(SamplingTests, RandomizedSamplingSelectTopCuda) {
     }
   }
 }
+
+TEST(SamplingTests, RandomizedSamplingSelectTopCuda_BatchSize1_LargeVocabSize) {
+  // This combination of `batch_size` and `vocab_size` will use the Sort + TopK
+  // algorithm to optimally use the hardware.
+  int batch_size = 1;
+  int vocab_size = 55000;
+  std::vector<int32_t> input_ids{3};
+
+  auto config = OgaConfig::Create(MODEL_PATH "hf-internal-testing/tiny-random-gpt2-fp32");
+  config->Overlay(R"({ "model": { "vocab_size" : 55000 } })");
+  config->ClearProviders();
+  config->AppendProvider("cuda");
+
+  auto model = OgaModel::Create(*config);
+  auto params = OgaGeneratorParams::Create(*model);
+  params->SetSearchOption("max_length", 10);
+  params->SetSearchOptionBool("do_sample", false);
+  params->SetSearchOption("batch_size", batch_size);
+
+  std::vector<float> logits_cpu(vocab_size * batch_size);
+  std::vector<int> indices(vocab_size * batch_size);
+  std::random_device rd;
+  std::mt19937 engine(rd());
+  std::uniform_int_distribution<> dist(1, 25);
+  int num_iter = 5000;
+  for (int i = 0; i < num_iter; i++) {
+    int num_large = dist(engine);
+    CreateRandomLogits(logits_cpu.data(), num_large, vocab_size, batch_size, engine);
+
+    auto generator = OgaGenerator::Create(*model, *params);
+    generator->SetLogits(*OgaTensor::Create(logits_cpu.data(), std::array<int64_t, 2>{batch_size, vocab_size}));
+    generator->GenerateNextToken();
+    auto next_tokens = generator->GetNextTokens();
+
+    // Verify outputs match expected outputs
+    for (int b = 0; b < batch_size; b++) {
+      float max_score = *std::max_element(logits_cpu.begin() + vocab_size * b, logits_cpu.begin() + vocab_size * (b + 1));
+      auto next_token = next_tokens[b];
+      auto next_token_score = logits_cpu[next_token + vocab_size * b];
+      EXPECT_EQ(next_token_score, max_score);
+    }
+  }
+}
 #endif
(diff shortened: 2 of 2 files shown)


src/cuda/cuda_sampling.cu
```diff
@@ -548,15 +548,26 @@ void GetTopKSubset(SamplingData* data, cudaStream_t stream, float* scores_in, fl
...
 
-  if (k <= 4) {
+  bool use_sort_and_pick_topk = false;
+
+  // TODO(hasesh): Tweak this heuristic by performing more benchmarks.
+  // It is likely that the sort and pick top_k approach works well for
+  // smallish batch_sizes on most hardware. For now, we limit it to
+  // sampling (batch_size == 1) and when vocab_size is "pretty large"
+  // (also needs benchmarking to identify cut-off).
+  if (batch_size == 1 && vocab_size >= 100000) {
```
kunal-vaishnavi: Should we make this heuristic accessible outside of the build to allow runtime tuning (e.g. as an env variable)?

src/cuda/cuda_sampling.cu
```diff
@@ -548,15 +548,26 @@ void GetTopKSubset(SamplingData* data, cudaStream_t stream, float* scores_in, fl
                              k,                          \
                              temperature);
 
-  if (k <= 4) {
+  bool use_sort_and_pick_topk = false;
```
jiafatom, line 577: How does this change affect model accuracy in general? It seems changing sampling algorithm? Then we need to measure it on babelbench to make sure the model accuracy does not get affected. Need more information from author to proceed.

hariharans29: Could you elaborate on how it changes the sampling algo ? It just changes the way in which the top_k elements are retrieved. Happy to run it through any quality benchmarks though.

src/cuda/cuda_sampling.cu
hariharans29: Personally I think that is overkill for now.

Ultimately, the heuristic will be something like - 

`(batch_size <= some_batch_threshold && vocab_size >= some_vocab_size_threshold)`

 I think, ultimately, these can be fine-tuned within the build as we keep profiling models ?

kunal-vaishnavi (approved):

tianleiwu:
A partial sort algorithm like cub::DeviceSelect is generally more efficient than a full sort. 

Example code snippet generated by LLM:
```
#include <cub/cub.cuh>

// A simple kernel to partition the original scores and indices based on a threshold
__global__ void PartitionTopKKernel(const float* scores_in, const int* indices_in,
                                    float* scores_out, int* indices_out,
                                    int vocab_size, float threshold, int k) {
    // Use an atomic counter to track the number of elements written to the output
    __shared__ int output_idx_counter;
    if (threadIdx.x == 0) {
        output_idx_counter = 0;
    }
    __syncthreads();

    for (int i = threadIdx.x; i < vocab_size; i += blockDim.x) {
        if (scores_in[i] >= threshold) {
            int write_idx = atomicAdd(&output_idx_counter, 1);
            // Ensure we don't write more than k elements, which can happen with duplicate scores
            if (write_idx < k) {
                scores_out[write_idx] = scores_in[i];
                indices_out[write_idx] = indices_in[i];
            }
        }
    }
}

// A new version of GetTopKSubset using partial sort for the general case
void GetTopKSubset_PartialSort(SamplingData* data, cudaStream_t stream, float* scores_in, float* scores_out, int* indices_out, int vocab_size, int batch_size, int k, float temperature) {
    // This implementation is simplified for batch_size = 1 as per the use case.
    // Extending to batch_size > 1 would require cub::DeviceSegmentedSelect.
    if (batch_size != 1) {
        // Fallback or error for simplicity.
        // The original LaunchSort using CUB Segmented Radix Sort is a good general solution.
        LaunchSort(data, stream, scores_in, data->scores_buffer.get(), indices_out, vocab_size, batch_size);
        DispatchBlockwiseSoftmaxForward<false>(stream, scores_out, const_cast<const float*>(data->scores_buffer.get()), k, vocab_size, k, batch_size);
        return;
    }

    // --- Step 1: Find the value of the k-th largest element ---
    // We want the k-th largest, which is the (vocab_size - k)-th smallest element.
    size_t temp_storage_bytes = 0;
    int rank = vocab_size - k;
    
    // Get temporary storage size
    cub::DeviceSelect::Select(nullptr, temp_storage_bytes, scores_in, data->scores_buffer.get(), vocab_size, &rank);
    data->temp_storage_bytes = temp_storage_bytes; // Assuming temp_buffer can be resized or is large enough
    
    // Run the selection
    cub::DeviceSelect::Select(data->temp_buffer.get(), data->temp_storage_bytes, scores_in, data->scores_buffer.get(), vocab_size, &rank);
    
    // data->scores_buffer now contains a single float: the k-th largest score.
    float threshold = 0.0f;
    cudaMemcpyAsync(&threshold, data->scores_buffer.get(), sizeof(float), cudaMemcpyDeviceToHost, stream);
    cudaStreamSynchronize(stream); // Sync to get the threshold value on the CPU.
                                   // A device-side alternative is preferred for max performance.

    // --- Step 2: Partition the full arrays based on the threshold ---
    // We need original indices [0, 1, 2, ..., vocab_size-1] to partition alongside the scores.
    LaunchPopulateIndices(data->indices_in.get(), vocab_size, 1, stream);
    
    // This custom kernel filters both scores and indices into a compact list of size k.
    // The output goes into the first k slots of scores_buffer and indices_out.
    dim3 grid(1);
    dim3 block(1024);
    PartitionTopKKernel<<<grid, block, 0, stream>>>(
        scores_in, data->indices_in.get(),
        data->scores_buffer.get(), indices_out,
        vocab_size, threshold, k);

    // --- Step 3: Apply softmax to the k selected elements ---
    // The input stride is now k, as PartitionTopKKernel created a compact list.
    DispatchBlockwiseSoftmaxForward<false>(stream, scores_out, const_cast<const float*>(data->scores_buffer.get()), k, k, k, batch_size);
}
```

tianleiwu (approved):

src/cuda/cuda_sampling.cu
```diff
@@ -548,15 +548,26 @@ void GetTopKSubset(SamplingData* data, cudaStream_t stream, float* scores_in, fl
...
+  } else if (!use_sort_and_pick_topk && k <= 8) {
     GetTopK(8);
-  } else if (k <= 16) {
+  } else if (!use_sort_and_pick_topk && k <= 16) {
     GetTopK(16);
-  } else if (k <= 32) {
+  } else if (!use_sort_and_pick_topk && k <= 32) {
     GetTopK(32);
-  } else if (k <= 64) {
+  } else if (!use_sort_and_pick_topk && k <= 64) {
```
jiafatom, line 596: For `GetTopK`, there is a `scores_out[i + batch * k] = top_k_sequence.value[i] / temperature;` in 
 https://github.com/microsoft/onnxruntime-genai/blob/b6d74e2e7efe2ae4d2b4eda7a10bcbd2eb3b1ff3/src/cuda/cuda_sampling.cu#L469, seems we need change in `LaunchSort`

hariharans29: Good catch, thanks - seems like the temperature is not honored in the sort based approach. 
Added the temperature scaling logic to the "sort + top_k" path

hariharans29: Are there any temperature specific tests I can add to cover this case ?

src/cuda/cuda_sampling.cu
```diff
@@ -528,13 +547,20 @@ void LaunchSampleKernel(SamplingData* data, cudaStream_t stream, float* scores,
...
   LaunchPopulateIndices(data->indices_in.get(), vocab_size, batch_size, stream);
   LaunchSortPairs<float>(data->temp_buffer.get(), data->temp_storage_bytes, scores_in, scores_out,
                          data->indices_in.get(), indices_out, vocab_size * batch_size, batch_size, data->offsets.get(),
                          stream, /*is_descending*/ true);
+
+  if (temperature != 1.f) {
+    // TODO: Fuse the temperature scaling with subsequent Softmax kernel - currently it isn't a very costly
+    // kernel, so keep it unfused
+    // We only need to scale the "top_k" elements from the sorted array (which is the output of the above kernel)
+    LaunchTemperatureScaling(scores_out, temperature, k, stream);
```
jiafatom: Can we have same param order to match with `LaunchSort`? like `score_out, stream, k, temperature`?

hariharans29: Sure, fixed the ordering

PR #1707, opened 2025-08-27
Add custom ops library path resolution using EP metadata

psakhamoori:
## Summary
Adds support for automatic custom ops library path resolution using execution provider metadata.

## Changes
- Enhanced custom ops library registration to resolve relative paths using EP metadata
- Added path resolution logic that checks EP devices for library_path metadata
- Improved filesystem utility functions for cross-platform path handling

## Key Features
- Automatically resolves relative custom ops library paths using EP library metadata
- Maintains backward compatibility with absolute paths
- Cross-platform path handling (Windows/Unix)

## Testing
- Build verified on Windows with MSVC

```
1: [----------] Global test environment tear-down
1: [==========] 37 tests from 6 test suites ran. (4785 ms total)
1: [  PASSED  ] 37 tests.
1: Shutting down OnnxRuntime... done
1/1 Test #1: UnitTests ........................   Passed    4.90 sec
```

Fixes custom ops library loading issues for execution providers.

### DIFF
--- a/src/filesystem.h
+++ b/src/filesystem.h
@@ -107,6 +107,30 @@ class path {
 #endif
   }
 
+  bool is_relative() const {
+#ifdef _WIN32
+    // On Windows, check if path starts with drive letter or UNC path
+    if (path_.length() >= 2 && path_[1] == ':') {
+      return false;  // Absolute path with drive letter (e.g., "C:\")
+    }
+    if (path_.length() >= 2 && path_[0] == '\\' && path_[1] == '\\') {
+      return false;  // UNC path (e.g., "\\server\share")
+    }
+    return true;  // Relative path
+#else
+    // On Unix-like systems, absolute paths start with '/'
+    return !path_.empty() && path_[0] != '/';
+#endif
+  }
+
+  path parent_path() const {
+    size_t pos = path_.find_last_of("/\\");
+    if (pos == std::string::npos) {
+      return path();  // No parent directory found
+    }
+    return path(path_.substr(0, pos));
+  }
+
  private:
   std::string path_;
 
@@ -148,5 +172,10 @@ class path {
   }
 #endif  // _WIN32
 };
+
+// Namespace-level functions
+inline bool exists(const path& p) {
+  return p.exists();
+}
 
 }  // namespace fs
--- a/src/models/model.cpp
+++ b/src/models/model.cpp
@@ -21,7 +21,6 @@
 #include "marian.h"
 #include "decoder_only_pipeline.h"
 #include "../dml/interface.h"
-#include "onnxruntime_ep_device_ep_metadata_keys.h"
 
 #if defined(_WIN32)
 #include <direct.h>
@@ -872,7 +871,7 @@ void Model::CreateSessionOptionsFromConfig(const Config::SessionOptions& config_
         for (size_t kvi = 0; kvi < num_entries; kvi++) {
           const std::string key = keys[kvi];
           const std::string val = values[kvi];
-          if (key == kOrtEpDevice_EpMetadataKey_LibraryPath) {
+          if (key == "library_path") {
             fs::path ep_library_dir = fs::path(val).parent_path();
             fs::path resolved_path = ep_library_dir / custom_library_path;
             if (fs::exists(resolved_path)) {
--- a/src/models/model.cpp
+++ b/src/models/model.cpp
@@ -4,6 +4,7 @@
 // Modifications Copyright(C) 2024-2025 Advanced Micro Devices, Inc. All rights reserved.
 #include <algorithm>
 #include <climits>
+#include <iostream>
 #include <random>
 #include <set>
 #include <string>
@@ -20,6 +21,7 @@
 #include "marian.h"
 #include "decoder_only_pipeline.h"
 #include "../dml/interface.h"
+#include "onnxruntime_ep_device_ep_metadata_keys.h"
 
 #if defined(_WIN32)
 #include <direct.h>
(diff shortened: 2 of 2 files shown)


chrisdMSFT (approved):

src/filesystem.h
```diff
@@ -107,6 +107,30 @@ class path {
 #endif
   }
 
+  bool is_relative() const {
+#ifdef _WIN32
+    // On Windows, check if path starts with drive letter or UNC path
+    if (path_.length() >= 2 && path_[1] == ':') {
+      return false;  // Absolute path with drive letter (e.g., "C:\")
```
osamasal2, line 114: Why not use functions, such as `PathIsUNC()` and `PathIsRelative()` to determine this? I believe these are part of `shlwapi.h`.

src/models/model.cpp
```diff
@@ -847,9 +848,45 @@ void Model::CreateSessionOptionsFromConfig(const Config::SessionOptions& config_
...
   if (config_session_options.custom_ops_library.has_value()) {
-    fs::path custom_library_file_prefix{config_session_options.custom_ops_library.value()};
-    session_options.RegisterCustomOpsLibrary(custom_library_file_prefix.c_str());
+    std::string custom_library_file_prefix = config_session_options.custom_ops_library.value();
+
+    // If relative path, try to resolve using EP library path from metadata
+    fs::path custom_library_path{custom_library_file_prefix};
+    if (custom_library_path.is_relative()) {
+      // Get EP devices to check for library_path metadata
+      size_t num_devices;
```
osamasal2: Recommend initializing all local variables.

src/models/model.cpp
```diff
@@ -847,9 +848,45 @@ void Model::CreateSessionOptionsFromConfig(const Config::SessionOptions& config_
...
+      const OrtEpDevice* const* device_ptrs;
+      Ort::api->GetEpDevices(&GetOrtEnv(), &device_ptrs, &num_devices);
+
+      bool resolved = false;
+      for (size_t i = 0; i < num_devices && !resolved; ++i) {
+        const OrtKeyValuePairs* keyvals = Ort::api->EpDevice_EpMetadata(device_ptrs[i]);
+        size_t num_entries;
```
osamasal2: Same with this variable.

osamasal2 (commented):
Added a couple of comments.

src/filesystem.h
psakhamoori: Thanks for the suggestion! You're absolutely right that `PathIsRelativeW()` from `shlwapi.h` would be the proper Windows API approach. However, I opted for the manual string parsing to avoid introducing a new library dependency.

Adding `shlwapi.h` would require linking against `shlwapi.lib` in the CMakeLists.txt, which would be a broader change to the build configuration. The goal is to minimize changes to the build system; I kept the implementation dependency-free.

The current manual approach handles the two main Windows absolute path cases:

Drive letter paths (e.g., C:\path)
UNC paths (e.g., \\server\share)
This covers the expected use cases for custom ops library path resolution while maintaining the existing build dependencies. If there's a strong preference for the Windows API approach, I am happy to discuss adding the shlwapi dependency, but wanted to keep this change minimal.

src/models/model.cpp
```diff
@@ -847,9 +848,45 @@ void Model::CreateSessionOptionsFromConfig(const Config::SessionOptions& config_
...
+        const OrtKeyValuePairs* keyvals = Ort::api->EpDevice_EpMetadata(device_ptrs[i]);
+        size_t num_entries = 0;
+        const char* const* keys = nullptr;
+        const char* const* values = nullptr;
+        Ort::api->GetKeyValuePairs(keyvals, &keys, &values, &num_entries);
+
+        for (size_t kvi = 0; kvi < num_entries; kvi++) {
+          const std::string key = keys[kvi];
+          const std::string val = values[kvi];
+          if (key == "library_path") {
```
chrisdMSFT: Why is this hardcoded here?

chrisdMSFT (changes requested):
We should use the const from the header not the hardcoded name.

Looks like ORT does NOT include the header we need.... SOO let's make a `const` and a comment that points to the header. `include/onnxruntime/core/session/onnxruntime_ep_device_ep_metadata_keys.h` - `kOrtEpDevice_EpMetadataKey_LibraryPath`

src/models/model.cpp
chrisdMSFT:
```c
// From include/onnxruntime/core/session/onnxruntime_ep_device_ep_metadata_keys.h
static const char* const kOrtEpDevice_EpMetadataKey_LibraryPath = "library_path";
```

src/models/model.cpp
```diff
@@ -847,9 +848,48 @@ void Model::CreateSessionOptionsFromConfig(const Config::SessionOptions& config_
...
+
+    std::string custom_library_file_prefix = config_session_options.custom_ops_library.value();
+
+    // If relative path, try to resolve using EP library path from metadata
+    fs::path custom_library_path{custom_library_file_prefix};
+    if (custom_library_path.is_relative()) {
+      // Get EP devices to check for library_path metadata
+      size_t num_devices = 0;
+      const OrtEpDevice* const* device_ptrs = nullptr;
+      Ort::api->GetEpDevices(&GetOrtEnv(), &device_ptrs, &num_devices);
```
baijumeswani, line 864: Could you please make necessary changes in the onnxruntime api wrapper instead of using onnxruntime api directly:

https://github.com/microsoft/onnxruntime-genai/blob/main/src/models/onnxruntime_api.h

src/models/model.cpp
```diff
@@ -4,6 +4,7 @@
 // Modifications Copyright(C) 2024-2025 Advanced Micro Devices, Inc. All rights reserved.
 #include <algorithm>
 #include <climits>
+#include <iostream>
```
baijumeswani, line 7:
```suggestion
```

src/models/model.cpp
```diff
@@ -847,9 +848,48 @@ void Model::CreateSessionOptionsFromConfig(const Config::SessionOptions& config_
     session_options.AddConfigEntry(config_entry.first.c_str(), config_entry.second.c_str());
   }
 
+  // Register custom ops libraries only if explicitly configured
   if (config_session_options.custom_ops_library.has_value()) {
-    fs::path custom_library_file_prefix{config_session_options.custom_ops_library.value()};
-    session_options.RegisterCustomOpsLibrary(custom_library_file_prefix.c_str());
+    // From include/onnxruntime/core/session/onnxruntime_ep_device_ep_metadata_keys.h
+    static const char* const kOrtEpDevice_EpMetadataKey_LibraryPath = "library_path";
```
baijumeswani, line 854: Use compile time const instead of runtime if possible.

```suggestion
    constexpr const char* const library_path_metadata_key_name = "library_path";
```

PR #1708, opened 2025-08-27
Improve top-k sampling CUDA kernel

gaugarg-nv:
Throughput numbers (tok/sec) for Qwen-2.5-1.5B int4 model on RTX 4090 with different top-K values and fixed top-p=0.8 value.

K | Before | After | Speed-up
-- | -- | -- | --
1 | 210 | 217 | 1.03
20 | 178 | 206 | 1.16
50 | 141 | 195 | 1.38

Top-K kernel shows an improvement of ~5.7x for top-k = 50 and 13x for top-k = 1.

Nsight profile before the change for Top-k = 50:

image

Nsight profile after the change for Top-K = 50:

image

Nsight profile before the change for Top-k = 1:
image

Nsight profile after the change for Top-k = 1:

image

### DIFF
--- a/src/cuda/cuda_sampling.cu
+++ b/src/cuda/cuda_sampling.cu
@@ -310,11 +310,11 @@ void DispatchBlockwiseSoftmaxForward(cudaStream_t stream, float* output, const f
   if (is_log_softmax) {
     SoftmaxBlockForward<ILP, float, float, float, LogSoftmaxForwardEpilogue>
         <<<grid, block, block.x * sizeof(float), stream>>>(output, const_cast<float*>(input),
-                                                            softmax_elements, input_stride, output_stride);
+                                                           softmax_elements, input_stride, output_stride);
   } else {
     SoftmaxBlockForward<ILP, float, float, float, SoftmaxForwardEpilogue>
         <<<grid, block, block.x * sizeof(float), stream>>>(output, const_cast<float*>(input),
-                                                            softmax_elements, input_stride, output_stride);
+                                                           softmax_elements, input_stride, output_stride);
   }
 }
 template void DispatchBlockwiseSoftmaxForward<true>(cudaStream_t, float*, const float*, int, int, int, int);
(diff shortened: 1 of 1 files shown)


src/cuda/cuda_sampling.cu
```diff
@@ -431,53 +431,79 @@ __global__ void FilterOnTopP(float* scores, float* prefix_sums, float* scores_te
...
-  if (threadIdx.x == 0) {
-    for (int i = 0; i < k; i++) {
-      scores_out[i + batch * k] = top_k_sequence.value[i] / temperature;
-      indices_out[i + batch * k] = top_k_sequence.key[i];
+    if (tid == 0) {
+      scores_out[ite + batch * k] = top_k_sequence.u / temperature;
+      indices_out[ite + batch * k] = top_k_sequence.p;
+
+      // set the ax value to -MAX_T_VAL so that the value doesn't get picked again
+      scores_in[batch * vocab_size + top_k_sequence.p] = -MAX_T_VAL;
```
hariharans29, line 493: Question: Is this safe without a threadfence ? Can this write be seen by every processing thread in ther CUDA memory model ?

hariharans29: Or does syncthreads() below guarantee that writes have gone through ?

gaugarg-nv: `__syncthreads()` is enough as we need to make memory writes visible within the thread block here, and not across thread blocks.

tianleiwu: How the Race Condition Happens

Here's a step-by-step breakdown of the problem:

Iteration 1:

All 1024 threads in the block scan the scores_in array for the current batch.

They cooperatively find the maximum value.

thread 0 writes this max value's index and score to the output arrays (indices_out, scores_out).

Crucially, thread 0 then overwrites the max value in the global scores_in array with -FLT_MAX to "remove" it from consideration for the next iteration.

The Problem Point:

Modern GPUs have complex memory hierarchies with caches (L1, L2). When thread 0 writes to scores_in, that write goes to a cache. It is not guaranteed to be immediately visible to the other 1023 threads in the same block.

The __syncthreads() call ensures all threads wait at that point, but it only synchronizes execution; it does not guarantee memory visibility.

Iteration 2 (Potential Bug):

The loop continues to the next iteration.

A different thread, say thread 5, might now read from the scores_in array.

Because the write from thread 0 might still be in a cache and not flushed to the memory that thread 5 can see, thread 5 could read the old, original maximum value instead of the -FLT_MAX that thread 0 wrote.

This would cause the kernel to find the same maximum value again in the second iteration, which is incorrect.

The Solution: __threadfence_block() acts as a memory barrier. It forces the write operation from thread 0 to be made visible to all other threads within the same block before any of them are allowed to proceed past the fence.

hariharans29: @tianleiwu  - I think `__syncthreads()` also performs the function of `threadfence_block()` and makes writes visible to all threads within a block. That is what I remember.

hariharans29: @gaugarg-nv  - In my distributed implementation in https://github.com/microsoft/onnxruntime-genai/pull/1710/, I also didn't use `threadfence()` as each TB keeps operating on its own shard of the array and hence any resetting of values to -FLT_MAX is still fully visible to all the threads within the same own block. Please let me know if this understanding is right.

tianleiwu: > `__syncthreads()` also performs the function of `threadfence_block()`

Key difference:
__syncthreads() = execution barrier + shared memory visibility guarantee (but not global memory).
__threadfence_block() = memory visibility guarantee (global, shared, local) within the block, but no execution barrier.
Here is global memory.

gaugarg-nv: Thanks @tianleiwu . I think you are right. 
I have fixed it now.

tianleiwu (approved):

gaugarg-nv:
@tianleiwu the CI/CD failure with MacOS seems like an issue with CI/CD infra. Could you please check?

tianleiwu:
> @tianleiwu the CI/CD failure with MacOS seems like an issue with CI/CD infra. Could you please check?

@gaugarg-nv, 
You need integrate/rebase latest main, which has a commit that fix the build error: https://github.com/microsoft/onnxruntime-genai/commit/e22739fae3c7fbbeba3f485a363f8b80a5ca73c8

gaugarg-nv:
> > @tianleiwu the CI/CD failure with MacOS seems like an issue with CI/CD infra. Could you please check?
> 
> @gaugarg-nv, You need merge/rebase latest main, which has a commit that fix the build error: [e22739f](https://github.com/microsoft/onnxruntime-genai/commit/e22739fae3c7fbbeba3f485a363f8b80a5ca73c8)

Changes are now rebased on top of main.

PR #1709, opened 2025-08-27
[CUDA] TopK+Softmax Fused Kernel based on Bitonic Sort

tianleiwu:
This pull request introduces a new cuda fused kernel for topk + temperature + softmax for performance optimizations for the Top-K sampling. Some bitonic improvement is based on [paper](http://anilshanbhag.in/static/papers/gputopk_sigmod18.pdf).

## 📊 Macro Benchmark Results

Latency in millisecond, and selection sort was the default kernel for k <= 64 before this change. The selection sort is the improved version from https://github.com/microsoft/onnxruntime-genai/pull/1708.

### vocab\_size = 201088, H200 GPU

bitonic sort is faster than selection sort when k is larger than 16. If we use top 50 in sampling, the new algorithm is about 4X faster.

batch_size | k | full_sort | selection_sort | bitonic_sort | selection/bitonic
-- | -- | -- | -- | -- | --
1 | 1 | 1.070 | 0.020 | 0.121 | 0.16
1 | 8 | 1.074 | 0.091 | 0.121 | 0.75
1 | 16 | 1.071 | 0.172 | 0.122 | 1.41
1 | 32 | 1.074 | 0.334 | 0.123 | 2.72
1 | 50 | 1.074 | 0.517 | 0.124 | **4.16**
1 | 64 | 1.075 | 0.659 | 0.124 | 5.31
2 | 1 | 1.168 | 0.020 | 0.135 | 0.15
2 | 8 | 1.168 | 0.091 | 0.135 | 0.68
2 | 16 | 1.169 | 0.173 | 0.135 | 1.28
2 | 32 | 1.170 | 0.336 | 0.136 | 2.47
2 | 50 | 1.170 | 0.522 | 0.137 | 3.81
2 | 64 | 1.171 | 0.665 | 0.137 | 4.84
4 | 1 | 1.169 | 0.020 | 0.149 | 0.13
4 | 8 | 1.169 | 0.091 | 0.150 | 0.61
4 | 16 | 1.170 | 0.173 | 0.150 | 1.15
4 | 32 | 1.171 | 0.337 | 0.151 | 2.23
4 | 50 | 1.172 | 0.521 | 0.152 | 3.44
4 | 64 | 1.172 | 0.665 | 0.152 | 4.36
8 | 1 | 1.187 | 0.020 | 0.199 | 0.1
8 | 8 | 1.188 | 0.092 | 0.199 | 0.46
8 | 16 | 1.189 | 0.174 | 0.200 | 0.87
8 | 32 | 1.190 | 0.339 | 0.200 | 1.69
8 | 50 | 1.189 | 0.525 | 0.201 | 2.61
8 | 64 | 1.191 | 0.669 | 0.202 | 3.31

The core of this change is the introduction of several new map-reduce algorithms and a dynamic benchmarking framework that automatically selects the fastest implementation at runtime for a given problem configuration.

---
### Bitonic Sort

The improvement is still ongoing:
```
Version Batch  Vocab Size  K    Algorithm                   Latency (ms)
-------------------------------------------------------------------------------------------------
v1        1           201088      50   BITONIC (s=2048,p=128)      0.1228
v2        1           201088      50   BITONIC (s=1024,p=256)      0.0985
```

#### Version 1
(Map): The FindBlockTopK_BitonicSort kernel partitions the input vocab_size into many smaller blocks. Each GPU block loads a partition into shared memory, performs a full bitonic sort on it, and writes the top kBitonicSortMaxK (64) results to an intermediate buffer in global memory.

(Reduce): It then runs cub::DeviceSegmentedRadixSort on the much smaller intermediate buffer to find the final global Top-K elements from the local winners.

Analysis: The implementation sorts many small, independent chunks completely. To further improve it, use a recursive reduction strategy that progressively halves the entire candidate set without ever fully sorting the intermediate partitions.

#### Version 2

Replaced FindBlockTopK_BitonicSort Kernel: The old kernel has been completely replaced with a new, optimized version that incorporates the advanced techniques from the paper.

In-Register Sorting (InRegisterBitonicSort8): A new __device__ function now handles the initial stages of the sort for a small chunk of 8 elements. This happens entirely in registers, which dramatically reduces the number of expensive shared memory reads and writes.

Shared Memory Padding: The shared memory arrays (smem_scores and smem_indices) are now declared with extra padding. All indexing into these arrays has been adjusted to account for this padding, which prevents memory bank conflicts and increases parallelism.

Hybrid Sort Logic: The new kernel now uses a hybrid approach. It first sorts small 8-element chunks in registers, writes the results back, and then performs the remaining stages of the bitonic merge across the entire partition in shared memory (using the conflict-free padded layout).

---
### Dynamic Benchmarking Framework
Idea: Inspired by https://github.com/microsoft/onnxruntime-genai/pull/1705, it is hard to know which algorithm is faster unless we benchmark it on current device and target parameters. 

The optimal algorithm and its tuning parameters (like num_partitions) can vary depending on the GPU architecture and problem size (vocab_size, k, batch_size). To ensure the best performance is always used, a one-time benchmarking system was added.

Implementation: On the first call to GetTopKSubset with a new configuration, the framework runs all viable algorithms with a range of tuning parameters, measures their latency, and caches the best-performing configuration. Subsequent calls with the same configuration will directly use the cached optimal choice without re-benchmarking.

---
### Testing

This provides a dedicated tool for collecting macro-benchmark results across all algorithms and configurations, making it easy to validate performance improvements.

### Future Works

* Consolidate softmax kernels in cuda_sampling.cu

---
## Appendix

A full list of ideas I have tried (Most of them does not help much since it is memory-bound problem):

---
### MAP_REDUCE
Idea: To address the GPU underutilization of the DIRECT_KERNEL, this algorithm implements a classic two-stage map-reduce strategy.

Map Stage: A large grid of CUDA blocks is launched. Each block processes a unique partition of the input vocabulary, finding the local Top-K elements within its partition. The results are written to an intermediate global buffer.

Reduce Stage: A second, smaller kernel launch (one block per batch item) reads the intermediate results and performs a final reduction to find the true global Top-K.

Improvement: This approach parallelizes the main workload across all available Streaming Multiprocessors (SMs) on the GPU, dramatically improving performance for large vocabularies.

---
### Optimization: Kernel Fusion (ReduceFinalTopKAndSoftmax)
Idea: The initial MAP_REDUCE algorithm required a third kernel launch to apply temperature scaling and compute the final softmax on the Top-K results. This optimization fuses the temperature scaling and softmax calculation directly into the second (reduce) stage kernel.

Improvement: By eliminating an entire kernel launch and a round-trip to global memory (writing scaled scores then reading them back), this fusion reduces latency and memory bandwidth usage.

---
### Optimization: MAP_REDUCE_SHARED
Idea: This variant of the map-reduce algorithm optimizes the reduce stage by leveraging shared memory. By reducing the number of partitions in the map stage (e.g., from 256 to 64), the intermediate results become small enough to fit entirely within a block's fast shared memory. The reduce kernel first cooperatively loads all intermediate data into shared memory and then performs the reduction.

Improvement: Reading from shared memory is significantly faster than reading from global memory, leading to a much faster final reduction step.

---
### Optimization: SINGLE_KERNEL_MAP_REDUCE (for batch_size=1)
Idea: To achieve the lowest possible latency, this algorithm fuses the map and reduce stages into a single kernel launch. It works as follows:

All blocks in the grid perform the "map" step, writing intermediate results to global memory.

Each block atomically increments a global sync_counter in device memory.

A designated block (block 0) spins, waiting for the sync_counter to equal the number of blocks in the grid. This acts as a device-wide grid synchronization barrier.

Once all blocks are finished, block 0 proceeds to execute the "reduce" stage on the intermediate data.

This approach completely eliminates the overhead and latency of launching a second kernel.

All the above kernels actually do not help much since it is memory-bound problem. I started optimizations targeting memory I/O speed as below:

---
### Optimization: `MAP_REDUCE_VEC`

* **Idea:** This optimization targets the "Map" stage, which is memory-bound by the initial read of the vocabulary. To increase memory throughput, the kernel was modified to read data in `float4` chunks instead of single `float` values. This allows the GPU to fetch four 32-bit floats in a single memory transaction.

* **Improvement:** Vectorized memory access more effectively utilizes the GPU's memory bandwidth, leading to a faster "Map" stage. However, this approach was ultimately limited by high register pressure from the per-thread `TopK` data structure.

---
### Optimization: `MAP_REDUCE_HYBRID_SORT`

* **Idea:** This algorithm combines the best parts of previous approaches. It uses the efficient vectorized "Map" stage from `MAP_REDUCE_VEC` but completely replaces the custom "Reduce" kernel. Instead of a manual reduction, it uses the highly optimized `cub::DeviceSegmentedRadixSort` library function to sort the small intermediate buffer.
* **Improvement:** CUB's radix sort is a "best-in-class" implementation that is significantly faster than a custom reduction kernel for sorting the intermediate candidates. This resulted in a substantial performance gain, making it the fastest algorithm before the final optimization.

---
### Optimization:  `MAP_REDUCE_BITONIC_SORT`

* **Idea:** While `MAP_REDUCE_HYBRID_SORT` was fast, its "Map" stage was still bottlenecked by register pressure. This final version replaces the `TopK` struct-based map kernel with a new one that performs a **Bitonic Sort** directly in shared memory. Each block loads its partition into the fast on-chip memory and performs a full, cooperative sort.

* **Improvement:** Sorting in shared memory is orders of magnitude faster than operating in global memory. This approach dramatically reduces global memory traffic and eliminates the register pressure bottleneck. By shifting the primary sorting work from slow global memory to ultra-fast shared memory, this algorithm achieves the best performance by a large margin, outperforming even the generic `FULL_SORT` baseline.


src/cuda/cuda_sampling.cu
```diff
@@ -537,32 +540,534 @@ void LaunchSort(SamplingData* data, cudaStream_t stream, float* scores_in, float
...
+    if (threadIdx.x == 0) {
+        atomicAdd(sync_counter, 1);
+    }
+
+    // --- STAGE 2: REDUCE (Only block 0 executes this) ---
+    if (blockIdx.x == 0) {
+        // Block 0 waits for all other blocks to report in
+        volatile int* p_counter = sync_counter;
+        while (*p_counter < gridDim.x) {}
+        __threadfence(); // Ensure visibility of writes from other blocks
```
hariharans29: Shouldn't all threadblocks execute threadfence to ensure their writes are visibile ?

src/cuda/cuda_sampling.cu
```diff
@@ -537,32 +540,534 @@ void LaunchSort(SamplingData* data, cudaStream_t stream, float* scores_in, float
...
+    // --- STAGE 1.5: GLOBAL SYNC ---
+    if (threadIdx.x == 0) {
+        atomicAdd(sync_counter, 1);
+    }
+
+    // --- STAGE 2: REDUCE (Only block 0 executes this) ---
+    if (blockIdx.x == 0) {
+        // Block 0 waits for all other blocks to report in
+        volatile int* p_counter = sync_counter;
+        while (*p_counter < gridDim.x) {}
```
hariharans29: I think it is enough that only one thread do the spinning (not every thread in the block needs to poll from global memory and spin ? An elected thread can spin and we can have a barrier (syncthreads()) before the threads in the reducing block start doing the reduction

src/cuda/cuda_sampling.cu
```diff
@@ -537,32 +540,534 @@ void LaunchSort(SamplingData* data, cudaStream_t stream, float* scores_in, float
...
+    }
+
+    float* intermediate_scores = data->scores_buffer.get();
+    int* intermediate_indices = data->indices_in.get();
+
+    cudaMemsetAsync(data->sync_counter.get(), 0, sizeof(int), stream);
+
+    dim3 grid(num_partitions);
+    dim3 block(kBlockSize);
+    SingleKernelMapReduceTopKAndSoftmax<max_k, kBlockSize><<<grid, block, 0, stream>>>(
```
hariharans29: Technically, we need to "reset" the sync counter (either via cudaMemsetAsync outside the kernel or inside the kernel once we cross the synchronization step) - its cost must be accounted for while measuring the E2E latency of this approach

src/cuda/cuda_sampling.cu
```diff
@@ -537,32 +540,534 @@ void LaunchSort(SamplingData* data, cudaStream_t stream, float* scores_in, float
...
+
+        // Now perform the reduction
+        TopK<float, max_k> final_thread_top_k;
+        final_thread_top_k.Init();
+
+        for (int i = threadIdx.x; i < num_intermediate_results_per_batch; i += kBlockSize) {
+            final_thread_top_k.Insert(intermediate_scores[i], intermediate_indices[i]);
+        }
+
+        TopK<float, max_k> final_top_k = BlockReduce(temp_storage).Reduce(final_thread_top_k, reduce_topk_op<float, max_k>);
```
hariharans29: Suggestion for one more variant of this: If k is small enough and the number of partitions are small enough - I think a single thread doing the reduction may work better than again doing multi-thread reduction and incurring that overhead. So, maybe having a variant to do single thread reduction will help ?

src/cuda/cuda_sampling.cu
tianleiwu: That time is small since set an integer async does not have much impact on latency.
BTW, this kernel will be removed based on benchmark results.

src/cuda/cuda_sampling.cu
tianleiwu: The kernel was removed.

src/cuda/cuda_sampling.cu
tianleiwu: I think a single thread reduction might help a little but need benchmark to verify.
BTW, the kernel was removed.

src/cuda/cuda_sampling.cu
tianleiwu: Good question.
BTW, the kernel was removed.

hariharans29:
> 0.149

For BS == 16, I wonder if we should benchmark with MapReduce version of selection sort to make it a fair comparison. But it does look like, when K > 16, bitonic sort is clearly much better.

tianleiwu:
I found a hybrid algorithm that is faster. Close this PR and will add a new one to replace it.

PR #1710, opened 2025-08-27
Provide distributed version of improved TopK kernel 

hariharans29:
Provides an improvement over the selection sort implementation in #1708 by using more threadblocks ("distributed") to do the top k search thereby improving the device utilization. We see non-negligible gains for large `k` values provided the vocabulary size is large enough

image

Summarizing OpenAI GPT OSS 20B model TPS with recent changes (with top_k = 50) on A100:

With ORT-GenAI TopK kernel (current version in the main branch) : 73 TPS (Baseline)
"Full sort" based TopK (https://github.com/microsoft/onnxruntime-genai/pull/1705): 99 TPS
Nvidia's TopK kernel (https://github.com/microsoft/onnxruntime-genai/pull/1708): 109 TPS
With the "distributed" version of Nvidia's TopK kernel (this PR): 115 TPS

CC: @kunal-vaishnavi @gaugarg-nv  @tianleiwu @hanbitmyths


src/cuda/cuda_sampling.cu
```diff
@@ -431,53 +439,178 @@ __global__ void FilterOnTopP(float* scores, float* prefix_sums, float* scores_te
...
+
+      // set the ax value to -MAX_T_VAL so that the value doesn't get picked again
+      scores_in[batch * vocab_size + top_k_sequence.p] = -MAX_T_VAL;
+    }
+
+    __syncthreads();
+  }
+
+  // All TBs flush their data and it should be visible to every other TB
+  __threadfence();
```
hariharans29: The distributed logic starts here

src/cuda/cuda_sampling.cu
```diff
@@ -431,53 +439,178 @@ __global__ void FilterOnTopP(float* scores, float* prefix_sums, float* scores_te
...
 
 // Gets all top K indices and scores from unsorted input
-template <int max_k>
-void LaunchGetTopKSubset(cudaStream_t stream, float* scores_in, float* scores_out, int* indices_out, int vocab_size, int batch_size, int k, float temperature) {
-  dim3 grid(batch_size, 1, 1);
-  dim3 block(256, 1, 1);
-  GetTopKKernel<max_k, 256><<<grid, block, 0, stream>>>(indices_out, scores_in, scores_out, batch_size, vocab_size, k, temperature);
+void LaunchGetTopKSubset(cudaStream_t stream, float* scores_in, float* scores_out, int* indices_out, int vocab_size, int batch_size, int k, float temperature, SamplingData* data = nullptr) {
+  bool enable_distributed_top_k = true;
+  if (batch_size == 1 && k <= 64 && enable_distributed_top_k) {
```
hariharans29: Right now it can only supported k <= 50 (this is a bug) - I will send out a fresh PR based on this once folks are happy with the implementation

src/cuda/cuda_sampling.cu
```diff
@@ -431,53 +439,178 @@ __global__ void FilterOnTopP(float* scores, float* prefix_sums, float* scores_te
...
+        int index = distributed_indices_out[i];
+        reduce_partial.insert(score, index);
+      }
+
+      // reduce in thread block
+      TopK_2 top_k_sequence_reduced = BlockReduce(temp_storage).Reduce(reduce_partial, reduce_topk_op_2);
+
+      if (tid == 0) {
+        scores_out[ite + batch * k] = top_k_sequence_reduced.u / temperature;
+        indices_out[ite + batch * k] = top_k_sequence_reduced.p;
```
hariharans29: This is a bug: We may need to reset the chosen score to -FLT_MAX so that it is not picked again (Perf might drop a bit with a fix)

src/cuda/cuda_sampling.cu
```diff
@@ -20,6 +20,7 @@ namespace cuda {
 
 constexpr int kMaxThreads = 1024;
 constexpr int kGPUWarpSize = 32;
+constexpr int top_k_shards = 32;
```
gaugarg-nv: I think it might be a good idea to decide this number based on the number of SMs and batch size. We want to make sure there is one TB per SM. We can also increase the number of TBs per SM but in that case, we should change the block size to a smaller value.

src/cuda/cuda_sampling.cu
```diff
@@ -431,53 +439,183 @@ __global__ void FilterOnTopP(float* scores, float* prefix_sums, float* scores_te
...
 }
 
 // Gets all top K indices and scores from unsorted input
-template <int max_k>
-void LaunchGetTopKSubset(cudaStream_t stream, float* scores_in, float* scores_out, int* indices_out, int vocab_size, int batch_size, int k, float temperature) {
-  dim3 grid(batch_size, 1, 1);
-  dim3 block(256, 1, 1);
-  GetTopKKernel<max_k, 256><<<grid, block, 0, stream>>>(indices_out, scores_in, scores_out, batch_size, vocab_size, k, temperature);
+void LaunchGetTopKSubset(cudaStream_t stream, float* scores_in, float* scores_out, int* indices_out, int vocab_size, int batch_size, int k, float temperature, SamplingData* data = nullptr) {
+  bool enable_distributed_top_k = true;
```
gaugarg-nv: It will be a good idea to capture perf numbers across different vocab sizes, and K-values to see when to enable `enable_distributed_top_k`. My guess is that for small-K (1 to 4) values, it might not help much. The same for small vocab sizes.

src/cuda/cuda_sampling.cu
hariharans29: I think this is one of those cases where more parallelism does not necessarily yield more gain as it just adds to the reduction overhead. On A100, I did not see any improvement from shards 32 to 64. I think there are more gains possible by using shared memory in the reduction threadblock and fetching all individual shards' results to shared memory (as we are going to read that data `k` times). For now, I think topk_shard of 32 seems like a good approach - but yes, in future we can think of having a dynamic selection of shards.

src/cuda/cuda_sampling.cu
hariharans29: I plan to add gating for vocab_size (and only enable sharding for vocab_size > large_threshold). 

In my benchmark for GPT-OSS (which has vocab_size 200k), I dropped down till k = 1 and it had 1.5-2% gain over your baseline. So, my conclusion is as long as vocab_size is large enough, distributed top_k search is better than or equivalent to 1 TB searching for all `k` values.

src/cuda/cuda_sampling.cu
hariharans29: Fixed

src/cuda/cuda_sampling.cu
hariharans29: Fixed

src/cuda/cuda_sampling.cu
gaugarg-nv: Actually my main concern is not with more parallelism but for devices where SM count < batch size * 32.
Consider RTX 5060 that has SM count = 30. Even with batch size 1, you will end-up having only 2 CTAs in the last wave underutilizing GPU.

hariharans29: Ok. Maybe we can make shard count atmost 32 ? 

Something like: top_k_shards = min(sm_count, 32);

src/cuda/cuda_sampling.cu
```diff
@@ -431,53 +439,185 @@ __global__ void FilterOnTopP(float* scores, float* prefix_sums, float* scores_te
...
+    }
+    // reduce in thread block
+    TopK_2 top_k_sequence = BlockReduce(temp_storage).Reduce(partial, reduce_topk_op_2);
+
+    if (tid == 0) {
+      distributed_scores_out_curr[ite + batch * k] = top_k_sequence.u;
+      distributed_indices_out_curr[ite + batch * k] = top_k_sequence.p;
+
+      // set the ax value to -MAX_T_VAL so that the value doesn't get picked again
+      scores_in[batch * vocab_size + top_k_sequence.p] = -MAX_T_VAL;
```
hariharans29: Needs this commit https://github.com/microsoft/onnxruntime-genai/pull/1708/commits/30dc69aaf27d0670b0a870e8a3faf0426b711954 reflected here and in the reduction logic and need to re-benchmark

src/cuda/cuda_sampling.cu
gaugarg-nv: Should we consider `batch_size` in this calculation, or do you plan to enable this only for batch size 1?

hariharans29: Initially, we can go with BS = 1. I will collect more benchmarks for BS > 1. I suspect it may help upto BS 4. 
Btw - I am not sure if this is still relevant after the bitonic sort introduced in https://github.com/microsoft/onnxruntime-genai/pull/1709 ? Is selection sort still useful upto K =16. If so, I can see if the distributed version of selection sort yields some sort of gains till that BS.

src/cuda/cuda_sampling.cu
hariharans29: Fixed

src/cuda/cuda_sampling.cu
hariharans29: Incorporated the race fix from the original PR. 

Distributed version does better when K is large.

Original selection sort + Top K = 50 -> **108 TPS**
Distributed selection sort + Top K = 50 -> **115 TPS**

Original selection sort + Top K = 8 -> **116 TPS**
Distributed selection sort + Top K = 8 -> **118 TPS**

Smaller the value of K, smaller is the gain over the single TB version

src/cuda/cuda_sampling.cu
```diff
@@ -41,6 +41,17 @@ SamplingData::SamplingData(unsigned long long random_seed, int batch_size, int v
...
   curand_states = CudaMallocArray<curandState>(batch_size);
+
+  top_k_distributed_lock = CudaMallocArray<uint>(1);
+  cudaMemset(top_k_distributed_lock.get(), 0, sizeof(uint));
+
+  cudaDeviceProp deviceProp;
+  cudaGetDeviceProperties(&deviceProp, 0); // Get properties for device 0
+  top_k_shards = std::min(top_k_shards, deviceProp.multiProcessorCount);
+
+  top_k_distirbuted_keys = CudaMallocArray<int>(top_k_shards * 64);  // We support distributed sharding upto top_k 64
```
kunal-vaishnavi:
```suggestion
  top_k_distributed_keys = CudaMallocArray<int>(top_k_shards * 64);  // We support distributed sharding upto top_k 64
```

src/cuda/cuda_sampling.cu
```diff
@@ -41,6 +41,17 @@ SamplingData::SamplingData(unsigned long long random_seed, int batch_size, int v
...
+  top_k_distirbuted_values = CudaMallocArray<float>(top_k_shards * 64); // We support distributed sharding upto top_k 64
```
kunal-vaishnavi:
```suggestion
  top_k_distributed_values = CudaMallocArray<float>(top_k_shards * 64); // We support distributed sharding upto top_k 64
```

src/cuda/cuda_sampling.cu
```diff
@@ -498,13 +509,129 @@ __global__ void GetTopKKernel(int* indices_out, float* scores_in, float* scores_
...
+    }
+
+    // TODO: Is there enough shared memory to bring the top_k shards to shared memory on all GPUs?
+    TopK_2 reduce_partial; 
+    for (int ite = 0; ite < k; ite++) {
+      reduce_partial.init();
+  
+      for (int i = threadIdx.x; i < top_k_shards*k; i += blockDim.x) {
+        float score = distributed_scores_out[i];
+        reduce_partial.insert(score, i);
```
tianleiwu: i is the buffer index instead of vocab index, this will cause tie-breaking error (so the sort result is no longer stable).

hariharans29: Thanks for the feedback and review. You're right - I will think of a solution. If you do have a simple one to fix this (I think we may have to also hold the vocab index in the reduction struct - that is what I can think of), please let me know.

hariharans29: Pushed a fix - hopefully the sort is stable now and tie-breaking happens correctly. 

Also pushed the optimization to use shared memory for the distributed TopK results which is then used during reduction - this approach is to be compared with the parallel bitonic sort based reduction you have in the other branch.

src/cuda/cuda_sampling.cu
```diff
@@ -498,13 +533,144 @@ __global__ void GetTopKKernel(int* indices_out, float* scores_in, float* scores_
...
+  int* distributed_indices_out_curr = distributed_indices_out + blockIdx.z * k;
+  float* distributed_scores_out_curr = distributed_scores_out + blockIdx.z * k;
+
+  int start_i = blockIdx.z * blockDim.x + tid;  
+ 
+  for (int ite = 0; ite < k; ite++) {
+    partial.init();
+
+    for (auto elemId = start_i; elemId < vocab_size; elemId += kBlockSize* gridDim.z) {
+      float elem = scores_in[elemId + batch * vocab_size];
```
tianleiwu: It is better to use coalesced load from global memory to speed up. Current reading pattern is with a stride. For best I/O speed, let a thread block reads from a continuous global memory block.

src/cuda/cuda_sampling.cu
```diff
@@ -498,13 +533,144 @@ __global__ void GetTopKKernel(int* indices_out, float* scores_in, float* scores_
   }
 }
 
+// Distributed Selection Sort based TopK kernel (multiple TB works along vocab space)
+template <int kBlockSize>
+__global__ void GetTopKKernelDistributed(int* indices_out, float* scores_in, float* scores_out, int batch_size,
```
tianleiwu: Remove temperature here since temperature is fused into sampling kernel now.

src/cuda/cuda_sampling.cu
```diff
@@ -498,13 +533,144 @@ __global__ void GetTopKKernel(int* indices_out, float* scores_in, float* scores_
...
+      int count_of_completed_TBs = 0;
+
+      asm volatile("ld.volatile.global.s32 %0, [%1];" : "=r"(count_of_completed_TBs) : "l"(top_k_distributed_lock));
+      while (count_of_completed_TBs < top_k_shards)
+      {
+          asm volatile("ld.volatile.global.s32 %0, [%1];" : "=r"(count_of_completed_TBs) : "l"(top_k_distributed_lock));
+      }
+
+      // Reset the lock for next kernel call
+      atomicExch(top_k_distributed_lock, 0);
```
tianleiwu: It is better to let the host (LaunchGetTopKSubset) reset it before launching the kernel, so that kernel always start with a clean state.

hariharans29: I don't think it matters either way and I think it is cleaner this way - if someone copy/pastes the code, they would have to copy/paste the resetting code and it adds an unwanted dependency - this way all the logic is self-contained. 

Also, it is better to have the kernel do this so that the compiler/runtime might find some overlapping opportunity with other instructions.

src/cuda/cuda_sampling.cu
hariharans29: Let me rebase with latest changes. There will be many changes after that.

src/cuda/cuda_sampling.cu
hariharans29: In the current implementation, adjacent threads of a warp read with unit stride - it is already best setup for coalescing. The stride you are seeing for `elemId` is the stride for a single thread (i.e.) all threads in the warp will get the same stride and still differ only by unit stride => coalescing happens already

src/cuda/cuda_sampling.cu
```diff
@@ -498,13 +533,144 @@ __global__ void GetTopKKernel(int* indices_out, float* scores_in, float* scores_
...
+  // Signal that each threadblock has done its work using elected thread
+  if (threadIdx.x == 0)
+  {
+    atomicAdd(top_k_distributed_lock, 1);
+  }
+
+  // The reduction threadblock
+  if (blockIdx.z == 0) {
+    // Elected thread spins while waiting for other TBs to complete their work
+    if (threadIdx.x == 0)
```
tianleiwu: Suggest to try cooperative group for synchronization (wait for finish of stage 1). That is more robustness, but I am not sure whether it can have better performance so it is better to do some experiments if time is allowed.

hariharans29: Thanks for the suggestion. I am also not sure if that will give tangible gains. I will do those experiments in a separate PR when I get some time. For now, I will leave a TODO to suggest doing that exploration.

src/cuda/cuda_sampling.cu
```diff
@@ -498,13 +533,144 @@ __global__ void GetTopKKernel(int* indices_out, float* scores_in, float* scores_
...
+      {
+          asm volatile("ld.volatile.global.s32 %0, [%1];" : "=r"(count_of_completed_TBs) : "l"(top_k_distributed_lock));
+      }
+
+      // Reset the lock for next kernel call
+      atomicExch(top_k_distributed_lock, 0);
+    }
+
+    // Number of shards is atmost 32 and top_k for this kernel is atmost 64
+    __shared__ float shared_distributed_scores_out[32 * 64];
```
tianleiwu: Add a constexpr for 32 for max shards allowed; add a constexpr for 64 for max k allowed for this algo.

src/cuda/cuda_sampling.cu
hariharans29: Should be fixed now.

src/cuda/cuda_sampling.cu
hariharans29: Done

src/cuda/cuda_sampling.cu
hariharans29: Left a TODO for this. Will experiment later - for now, let's go with perf gains given by this PR.

src/cuda/cuda_topk_distributed_select_sort.cuh
```diff
@@ -0,0 +1,193 @@
...
+                     : "l"(top_k_distributed_lock));
+      }
+
+      // Reset the lock for next kernel call
+      atomicExch(top_k_distributed_lock, 0);
+    }
+
+    // Number of shards is atmost `kTopKDistributedSelectSortMaxShards` and top_k for this kernel is atmost `kTopKDistributedSelectSortMaxTopK`
+    __shared__ float shared_distributed_scores_out[topk_impl_details::kTopKDistributedSelectSortMaxShards * topk_impl_details::kTopKDistributedSelectSortMaxTopK];
+    __shared__ float shared_distributed_indices_out[topk_impl_details::kTopKDistributedSelectSortMaxShards * topk_impl_details::kTopKDistributedSelectSortMaxTopK];
```
tianleiwu, line 118: float => int?

tianleiwu:
Overall, looks good.  Further optimization opportunities:

(1) The final reduction stage on the num_shards * k candidates uses a slow selection sort. Since the number of candidates is small enough to fit in shared memory, we can use a highly efficient parallel sorting algorithm like Bitonic Sort to sort all candidates at once. After sorting, simply take the first k elements. This replaces an O(k^2) loop with a much faster O(M log^2 M) sort, where M = num_shards * k.

(2) Have a build option for unstable sort without tie-breaking. That could streamline the logic and some performance gain since it is not needed to have tie-breaking for sampling use case.

(3) If we want to optimize k=1, it is possible to have a variant that does not write to output. In this way, k=1 does not need copy input scores.

(4) If we want to optimize k<=8, it is possible to use registers to sort up to 256 elements (i.e 32 shards * 8). For example, in a warp of 32 threads, each thread sort 8 elements. Then we can use warp level reduction to find top 8 among all 32 threads. The reduction can be performed in a tree-like fashion.  In the first step, thread 0 merges its list with thread 1's list, thread 2 merges with thread 3's, and so on. This reduces the number of candidate lists by half. This merging process continues for a few steps (log2(32)=5 steps) until thread 0 holds the final top 8 elements for the entire warp. Once the warp-level reduction is complete, a single thread (e.g., thread 0) will have the final top 8 values and their corresponding indices for the original 256 inputs. This thread can then write the results to global memory.
This approach might be optimized for k up to 16. Need some experiments to compare with shared memory bitonic sort.

PR #1714, opened 2025-08-28
Fix benchmark_multimodal

jiafatom:
(1) When I run `python benchmark_multimodal.py -i /sunghcho_data/onnx_models/whisper-tiny-en/cuda/cuda-fp16/ -au /home/jiafa/accuracy/open_asr_leaderboard/whisper/data/20090202-0900-PLENARY-9-en_20090202-17\:20\:18_2.wav -m 448`, the code ` inputs = processor(prompt, images=image, audios=audio)` has core dump because `strings.size()==0` for `auto shape = std::array<int64_t, 2>{static_cast<int64_t>(strings.size()), static_cast<int64_t>(encoded.size() / strings.size())};` in `model.cpp`. This is because ` WhisperProcessor::Process` only goes through `EncodeBatch` whereas `payload.prompts={}` when we set up a single audio there. So for single audio case, we capsulate into `prompts` and then process.
(2) The code refactoring causes `params.set_inputs(inputs)` no longer works.

### DIFF
--- a/benchmark/python/benchmark_multimodal.py
+++ b/benchmark/python/benchmark_multimodal.py
@@ -157,18 +157,19 @@ def run_benchmark(args, model, processor, image, audio, generation_length, max_l
         main_prompt = "What is the meaning of life?"
         prompt = f'{user_prompt}{main_prompt}{prompt_suffix}{assistant_prompt}'        
 
-    inputs = processor(prompt, images=image, audios=audio)
-    prompt_length = inputs['input_ids'].shape[1]
+    prompts = [prompt]
+    inputs = processor(prompts, images=image, audios=audio)
+    prompt_length = inputs['input_ids'].shape()[1]
     if args.verbose: print(f"Prompt used: {prompt}")
 
     params = og.GeneratorParams(model)
-    params.set_inputs(inputs)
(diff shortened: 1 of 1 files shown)


src/models/model.cpp
```diff
@@ -1046,7 +1046,9 @@ MultiModalProcessor::MultiModalProcessor(Config& config, const SessionInfo& sess
 }
 
 std::unique_ptr<NamedTensors> MultiModalProcessor::Process(const std::string& prompt, const Images* images, const Audios* audios) const {
-  Payload payload{prompt, {}, images, audios};
+  const char* prompt_list[] = { prompt.c_str() };
+  std::span<const char*> prompts(prompt_list, 1);
```
kunal-vaishnavi: Instead of changing this, we can change

https://github.com/microsoft/onnxruntime-genai/blob/b6d74e2e7efe2ae4d2b4eda7a10bcbd2eb3b1ff3/benchmark/python/benchmark_multimodal.py#L160

to follow the working example to provide the prompt.

https://github.com/microsoft/onnxruntime-genai/blob/b6d74e2e7efe2ae4d2b4eda7a10bcbd2eb3b1ff3/examples/python/whisper.py#L54-L55

kunal-vaishnavi (approved):

PR #1722, opened 2025-09-04
Use updated Decoder API with `skip_special_tokens`

sayanshaw24:
### Updates

Inside the foundry catalog for [Foundry Local](https://github.com/[microsoft/Foundry-Local](https://github.com/microsoft/Foundry-Local)), all tool call tokens are marked as special such that we can use them with guidance inside [GenAI](https://github.com/microsoft/onnxruntime-genai). Now, in order for decoding to work with these tokens, we need to expose using special tokens as an option in our Decoder API in Extensions. Hence, ORT Extensions now adds `skip_special_tokens` to `OrtxDetokenize1DWithOptions`, which this PR leverages in the GenAI `Decoder`.

Note that we leave `OrtxDetokenize1D` as is for backward compatibility. In the future, we shall be adding an extensible options map API into ORT Extensions that will reduce the need for messy parameter additions to our Encoder/Decoder APIs, and can be easily consumed by GenAI and other partners.

### Validation
- [x] Native C++ testing in ORT Extensions against HuggingFace benchmark.

### DIFF
--- a/src/models/model.cpp
+++ b/src/models/model.cpp
@@ -274,7 +274,7 @@ std::vector<int32_t> Tokenizer::Encode(const char* text) const {
 
 std::string Tokenizer::Decode(std::span<const int32_t> tokens) const {
   OrtxPtr<OrtxStringArray> ortx_string_array;
-  CheckResult(OrtxDetokenize1D(tokenizer_, reinterpret_cast<const uint32_t*>(tokens.data()), tokens.size(), ortx_string_array.Address()));
+  CheckResult(OrtxDetokenize1DWithOptions(tokenizer_, reinterpret_cast<const uint32_t*>(tokens.data()), tokens.size(), ortx_string_array.Address(), true /* skip_special_tokens */));
 
   const char* string;
   CheckResult(OrtxStringArrayGetItem(ortx_string_array, 0, &string));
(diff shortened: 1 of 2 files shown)


src/models/model.cpp
```diff
@@ -274,7 +274,7 @@ std::vector<int32_t> Tokenizer::Encode(const char* text) const {
 
 std::string Tokenizer::Decode(std::span<const int32_t> tokens) const {
```
kunal-vaishnavi, line 275: Can we enable controlling `skip_special_tokens` in the parameters for `Tokenizer.Decode` and `Tokenizer.Encode`?

sayanshaw24: We can, but this will break backward compatibility - maybe let's discuss this with @baijumeswani first thing tomorrow/in our GenAI sync?

baijumeswani: @kunal-vaishnavi is your comment about controlling the behavior via a flag exposed through the public C API? Or are you suggesting that we add in an argument to this function here with default values (as true for now)?

If we plan to expose this flag through the C API, we should consider all potential flags we might want to expose and define a TokenizerOptions struct to encapsulate them.

src/models/model.cpp
```diff
@@ -274,7 +274,7 @@ std::vector<int32_t> Tokenizer::Encode(const char* text) const {
...
   OrtxPtr<OrtxStringArray> ortx_string_array;
-  CheckResult(OrtxDetokenize1D(tokenizer_, reinterpret_cast<const uint32_t*>(tokens.data()), tokens.size(), ortx_string_array.Address()));
+  CheckResult(OrtxDetokenize1DWithOptions(tokenizer_, reinterpret_cast<const uint32_t*>(tokens.data()), tokens.size(), ortx_string_array.Address(), true /* skip_special_tokens */));
```
baijumeswani, line 277: Does using true for all scenarios impact any model negatively?

baijumeswani (commented):
Changes look good to me. @kunal-vaishnavi to approve

src/models/model.cpp
kunal-vaishnavi: I was originally thinking we should add a flag. From the perspective of Foundry Local, special tokens such as tool calling tokens need to be decoded. Then the existing functionality in this PR would be sufficient for Foundry Local.

kunal-vaishnavi: I was originally thinking we should add a flag. From Foundry Local's perspective, special tokens such as tool calling tokens need to be decoded. Then the existing functionality in this PR would be sufficient for Foundry Local.

kunal-vaishnavi (approved):

src/models/model.cpp
sayanshaw24: Nope! It was defaulting to true previously already, we only expose the parameter in this PR.

PR #1729, opened 2025-09-05
Update sampling tests

tianleiwu:
(1) Update sampling benchmark test to show a table friendly.
(2) Update RandomizedSamplingTopPAndKCuda to use random seed, and temperature (the test case  is not enabled since current code base has bug).
(3) Update Config if random seed is out of int32 range. It is because, my previous test failed since I pass int64 random seeds and it silently truncate it and causes test failure.

### DIFF
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -6,6 +6,10 @@
 #include "json.h"
 #include <fstream>
 #include <sstream>
+#include <limits>
+#include <cmath>
+#include <sstream>
+#include <stdexcept>
 
 namespace Generators {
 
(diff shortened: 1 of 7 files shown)


baijumeswani:
> Example sampling benchmark results (CPU is faster than CUDA!):

This is not with the optimizations from your PR, right?

src/config.cpp
```diff
@@ -746,6 +750,30 @@ struct Model_Element : JSON::Element {
   Speech_Element speech_{v_.speech};
 };
 
+int safe_double_to_int(double x, std::string_view name) {
```
baijumeswani:
```suggestion
int SafeDoubleToInt(double x, std::string_view name) {
```

tianleiwu:
> > Example sampling benchmark results (CPU is faster than CUDA!):
> 
> This is not with the optimizations from your PR, right?

Right, this is used to show baseline number.

PR #1731, opened 2025-09-06
[CUDA] TopK and Sampling kernel Improvements

tianleiwu:
This is a combo of TopK and Sampling cuda kernel improvements.

Note that I will split this PR to two: one for sampling kernel, one for new topk kernel.

The new TopK kernel has 6X ~ 10X improvement compared to selection sort.


src/cuda/cuda_topk.cu
```diff
@@ -0,0 +1,75 @@
...
+#include "cuda_topk_hybrid_sort.cuh"
+#include "cuda_topk_select_sort.cuh"
+
+namespace Generators {
+namespace cuda {
+
+// Helper to determine the optimal partition size for the hybrid sort algorithm
+// based on vocabulary and batch size.
+inline int GetHybridSortPartitionSize(int vocab_size, int batch_size) {
+  if (vocab_size >= 147456) {
```
hariharans29, line 17: Could you share commentary around how these numbers were chosen ?

PR #1732, opened 2025-09-06
[CUDA] sampling kernel improvements

tianleiwu:
Key improvements:
(1) Unified Fused Kernel: We have adopted the single FusedSamplingKernel for all k <= 256. This kernel performs all sampling steps (temperature scaling, softmax, CDF calculation, filtering, re-normalization, and final selection) in one pass.
(2) Refactoring to put top-k related code to separated files.


hariharans29:
Could you provide the perf improvement numbers (simply as sampling TPS numbers) for this change alone ?

src/cuda/cuda_topk.h
```diff
@@ -0,0 +1,67 @@
...
+
+#pragma once
+#include <curand_kernel.h>
+
+#include "cuda_common.h"
+
+namespace Generators {
+namespace cuda {
+
+constexpr int kHybridSortMaxK = 64;  // up to 256.
```
hariharans29: Comment out of sync ?

src/cuda/cuda_topk.cu
```diff
@@ -0,0 +1,63 @@
...
+}
+
+void TopkDataCompact::CompactOutput(int batch_size, int vocab_size, cudaStream_t stream, int k) {
+  topk_scores_compact = CudaMallocArray<float>(static_cast<size_t>(batch_size) * k);
+  topk_indices_compact = CudaMallocArray<int>(static_cast<size_t>(batch_size) * k);
+  dim3 grid(batch_size);
+  dim3 block(256);
+  CompactStridedData<float><<<grid, block, 0, stream>>>(topk_scores, topk_scores_compact.get(), k, batch_size, topk_stride);
+  CompactStridedData<int><<<grid, block, 0, stream>>>(topk_indices, topk_indices_compact.get(), k, batch_size, topk_stride);
+  cudaStreamSynchronize(stream);
```
hariharans29: Is this stream sync needed ?

src/cuda/cuda_topk.h
```diff
@@ -0,0 +1,67 @@
...
+
+  // Used to hold initial vocabulary indices for full sort.
+  cuda_unique_ptr<int> intermediate_indices_1;
+
+  // A dedicated "ping-pong" buffer for full sort.
+  cuda_unique_ptr<int> intermediate_indices_2;
+
+  // Primary buffer for holding raw scores.
+  // - Full sort: Holds the fully sorted raw scores.
+  // - Selection sort: Not used directly for output, but reserved.
```
hariharans29: Could you expand on "not used directly for output, but reserved" ?

src/cuda/cuda_topk_softmax.cuh
```diff
@@ -0,0 +1,136 @@
...
+  __shared__ float block_max_val;
+  __shared__ float block_sum_exp;
+
+  // STEP 1: Find max_val in parallel
+  // Each thread (where threadIdx.x < k) loads its score and applies temperature.
+  float thread_score =
+      (threadIdx.x < k) ? (batch_sorted_scores[threadIdx.x] / temperature) : -std::numeric_limits<float>::max();
+
+  // CUB reduces the values, placing the result in thread 0.
+  float max_val_reduced = BlockReduce(temp_storage).Reduce(thread_score, cub::Max());
```
hariharans29: For algorithms like 'Selection sort', the top k values are bound to be in sorted order. Can there be an input flag to provide that hint so that we can avoid the search for max value ?

src/cuda/cuda_topk_full_sort.cuh
```diff
@@ -0,0 +1,104 @@
...
+    offsets[index] = index * vocab_size;
+  }
+}
+
+inline void LaunchPopulateOffsets(int* offsets, int vocab_size, int batch_size, cudaStream_t stream) {
+  // We need batch_size + 1 offsets for CUB's segmented sort.
+  dim3 grid(CeilDiv(batch_size + 1, 256), 1, 1);
+  dim3 block(256, 1, 1);
+  PopulateOffsets<<<grid, block, 0, stream>>>(offsets, vocab_size, batch_size);
+  CUDA_CHECK(cudaGetLastError());
```
hariharans29: Is this needed ? Can this be enabled only on debug builds ?

src/cuda/cuda_topk.h
hariharans29: Is this comment related to the other PR ?

src/cuda/cuda_topk.cu
```diff
@@ -0,0 +1,63 @@
...
+}
+
+void GetTopK(TopkData* topk_data, cudaStream_t stream, const float* scores_in, int vocab_size, int batch_size, int k) {
+  assert(topk_data != nullptr);
+
+  if (k > 64) {
+    LaunchSort(topk_data, stream, scores_in, topk_data->intermediate_scores_1.get(),
+               topk_data->intermediate_indices_1.get(), vocab_size, batch_size);
+  } else {
+    // NOTE: This modifies scores_in in-place
```
hariharans29: Maybe expose a boolean flag to indicate input buffer mutation is allowed and default it to true ? That way future users of this routine will be aware from the parameter list that potentially the input buffer could be modified if they provide `true` for that flag

src/cuda/cuda_topk.h
tianleiwu: removed

src/cuda/cuda_topk_full_sort.cuh
tianleiwu: It's needed since the segment radix need to know the offset of each batch.

src/cuda/cuda_topk.cu
tianleiwu: Changed the kernel to make sure the input buffer is not mutable. It seems that impact on perf is small (2 us).

src/cuda/cuda_topk.cu
tianleiwu: It is used in test code so I moved the stream sync to test code.

src/cuda/cuda_topk.h
tianleiwu: comment updated.

src/cuda/cuda_topk_softmax.cuh
tianleiwu: Right. It is not needed. I've updated the code to use first elements for max value.

src/cuda/cuda_topk.h
tianleiwu: removed

src/cuda/cuda_topk_full_sort.cuh
hariharans29: You mean `cudaGetLastError()` is needed ?

tianleiwu: I updated the check according to your suggestion.

test/CMakeLists.txt
```diff
@@ -1,46 +1,58 @@
...
-source_group(TREE ${PROJECT_SOURCE_DIR} FILES ${all_test_srcs})
-set_property(DIRECTORY ${CMAKE_SOURCE_DIR} PROPERTY VS_STARTUP_PROJECT unit_tests)
-
-# Hide symbols by default, so that shared libraries don't link to our redirected symbols (leads to infinite loops)
-if (NOT MSVC)
-  target_compile_options(unit_tests PRIVATE "-fvisibility=hidden")
-endif()
-
-add_test(NAME UnitTests COMMAND unit_tests)
+include(${CMAKE_SOURCE_DIR}/cmake/cxx_standard.cmake)
```
hariharans29, line 1: Is there a diff in this file ?

test/topk_tests.cpp
```diff
@@ -0,0 +1,134 @@
...
+    Generators::cuda::RunTopKViaSelectionSort(topk_data.get(), stream, scores_in_d.get(),
+                                              params.vocab_size, params.batch_size, params.k);
+  });
+
+  CUDA_CHECK(cudaStreamDestroy(stream));
+}
+
+TEST(TopKTests, ParityTests) {
+  std::vector<TopKTestParams> test_cases = {
+      {1, 10000, 50},
```
hariharans29: Maybe add a few more tests ? 

Larger batch sizes ?
Few more vocab sizes and top_k values - just to improve coverage

src/cuda/cuda_common.h
```diff
@@ -91,4 +102,54 @@ cuda_unique_ptr<T> CudaMallocArray(size_t count, std::span<T>* p_span = nullptr)
...
+      ss << "CUDA error in " << __func__ << " at " << __FILE__    \
+         << ":" << __LINE__ << " - " << cudaGetErrorString(err);   \
+      throw Generators::CudaError(ss.str(), err);                 \
+    }                                                             \
+  } while (0)
+
+#ifdef NDEBUG
+#define CUDA_CHECK_LAUNCH()                                        \
+  do {                                                             \
+    cudaError_t err = cudaPeekAtLastError();                       \
```
hariharans29: Could you please explain this distinction ?

Why is `cudaPeekAtLastError()` better in release builds ?

tianleiwu: cudaPeekAtLastError() is very light weight, just peek at status. 
The cudaGetLastError() need clear last error. That introduces a little more overhead.

test/CMakeLists.txt
tianleiwu: There is change to add cuda related source code to build unit test. That's needed since we are testing internal functions that are not exposed in GenAI library.

You can view it by change the view to hide whitespace.

image

test/topk_tests.cpp
tianleiwu: Good suggestion. I added more test cases.

src/cuda/cuda_common.h
```diff
@@ -1,3 +1,14 @@
+#pragma once
```
hariharans29, line 4: Add Copyright ?

src/cuda/cuda_sampling.h
```diff
@@ -0,0 +1,53 @@
...
+void GetSample(SamplingData* sampling_data, cudaStream_t stream, int32_t* d_next_token, const float* d_scores,
+               int vocab_size, int batch_size, int k, float p, float temperature);
+
+// A general-purpose block-wise softmax implementation, needed by beam search.
+template <bool is_log_softmax>
+void DispatchBlockwiseSoftmaxForward(cudaStream_t stream, float* output, const float* input, int softmax_elements,
+                                     int input_stride, int output_stride, int batch_count);
+
+// The following are for macro benchmark.
+void LaunchFusedSampleKernel(SamplingData* data, cudaStream_t stream, const float* scores, const int* indices,
```
hariharans29, line 46: Nit: Can the 2 methods being used for benchmarks be in a different namespace ?

src/cuda/cuda_topk.h
```diff
@@ -0,0 +1,70 @@
...
+
+  cuda_unique_ptr<float> topk_scores_compact;  // compact [batch_size, k] output scores
+  cuda_unique_ptr<int> topk_indices_compact;   // compact [batch_size, k] output indices
+};
+
+// Main dispatcher for Top-K. Used by the sampling logic. The topk_data will be updated for output pointers and stride.
+void GetTopK(TopkData* topk_data, cudaStream_t stream, const float* scores_in, int vocab_size, int batch_size, int k);
+
+// Top-K algorithm implementations. These are exposed for testing and benchmarking.
+void RunTopKViaSelectionSort(TopkData* data, cudaStream_t stream, const float* scores_in, int vocab_size, int batch_size, int k);
```
hariharans29, line 66: Can these functions be in different namespaces as they are only used for testing and benchmarking ?

src/cuda/cuda_topk.h
```diff
@@ -0,0 +1,70 @@
...
+  cuda_unique_ptr<int> batch_offsets;
+
+  // --- Information of Final Output (Input to Sampling Stage) ---
+  const float* topk_scores = nullptr;  // pointer to the top-k scores data (in either intermediate_scores_1 or intermediate_scores_2)
+  const int* topk_indices = nullptr;   // pointer to the top-k indices data (in either intermediate_indices_1 or intermediate_indices_2)
+  int topk_stride = 0;                 // stride of the top-k output data: k for selection sort, vocab_size for full sort
+};
+
+// For parity test, a derived struct to help compact output buffers.
+struct TopkDataCompact : public TopkData {
```
hariharans29, line 49: Is there use-case for the compaction operations and hence the struct ?

tianleiwu: This is for testing. Make it easy to compare (otherwise, we need a compare kernel with stride input).

src/cuda/cuda_topk.h
tianleiwu: They are also used in GetTopK. If not for testing, they need not be in header file.
What namespaces do you suggest for them?

src/cuda/cuda_sampling.cu
```diff
@@ -1,601 +1,386 @@
...
+
+  typedef cub::BlockReduce<float, kBlockSize> BlockReduce;
+  __shared__ typename BlockReduce::TempStorage reduce_temp_storage;
+  __shared__ float block_max_val;
+  __shared__ float block_sum_exp;
+
+  // --- Stage 1: Initial Softmax with Temperature (for Top-P filtering) ---
+  float thread_val = -FLT_MAX;
+  for (int i = threadIdx.x; i < k; i += kBlockSize) {
+    float logit = batch_scores[i] / temperature;
```
hariharans29: I thought we were going to avoid this linear pass and reduction if we already know that the scores are sorted in descending order ? (i.e.) the first value is the max value ?

src/cuda/cuda_sampling.h
tianleiwu: They are also used internally in GetSample so I think the namespace is fine.

These functions are exposed in header for benchmarking purposes.

src/cuda/cuda_topk.h
hariharans29: namespace `test_and_benchmark` or something similar ?

hariharans29: Oh if they are called from GetTopK() - I guess it is fine to leave as is. Looking at the comment, I accidentally thought they are only used for benchmarking.

src/cuda/cuda_sampling.h
hariharans29: Got it

src/cuda/cuda_topk.h
hariharans29: If this is purely for testing, can they moved to the tests ? The code can remain same (i.e.) inherit from `TopkData` struct - trying to reduce amount of essential code in these files. You can consider these suggestions nits.

test/statistics_helper.h
```diff
@@ -1,6 +1,5 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
-
```
hariharans29: Nit: stray change

src/cuda/cuda_topk_softmax.cuh
```diff
@@ -0,0 +1,135 @@
...
+
+  // This implementation uses CUB for parallel reduction, which is efficient.
+  // The key fix is to correctly broadcast the reduction result (which CUB places in thread 0)
+  // to all other threads in the block using shared memory.
+  typedef cub::BlockReduce<float, kBlockSize> BlockReduce;
+  __shared__ typename BlockReduce::TempStorage temp_storage;
+  __shared__ float block_max_val;
+  __shared__ float block_sum_exp;
+
+  // STEP 1: Find max_val in parallel
```
hariharans29: Nit: comment of finding max_val in parallel is outdated

src/cuda/cuda_sampling.cu
hariharans29: I think it is better to keep both code-paths (i.e.) use first element as max value and parallel find max value as in future we may use a different TopK algorithm that return stop k values unsorted and we may forget to update these logic. So, we can keep both code paths and take a flag for the user to specify if the input values are sorted or not and take appropriate code-path.

hariharans29: Same comment for the "unfused" path (i.e.) using `CopyAndSoftmaxKernel`

hariharans29:
General comment about perf in the PR description: Since we are using the fused approach till k <= 256 - Maybe add a couple of more comparative perf data points - for k = 64, k = 128, k = 192, and k = 256

src/cuda/cuda_topk.h
tianleiwu: There is limitation of unit test project. Current setting does not build cu file, so I put it here.

tianleiwu: Keep it for now until I found a better way to add cu file to unit test folder.

src/cuda/cuda_sampling.cu
tianleiwu: Yes. Updated.

src/cuda/cuda_sampling.cu
```diff
@@ -1,601 +1,392 @@
...
+  float* temp_scaled_logits = smem;
+  float* filtered_logits = smem + kBlockSize;
+
+  typedef cub::BlockReduce<float, kBlockSize> BlockReduce;
+  __shared__ typename BlockReduce::TempStorage reduce_temp_storage;
+  __shared__ float block_max_val;
+  __shared__ float block_sum_exp;
+
+  // --- Stage 1: Initial Softmax with Temperature (for Top-P filtering) ---
+  float thread_val = -FLT_MAX;
```
hariharans29: No use for `thread_val` in lines 68...72 as we no longer do threadblock reduction to find max value ?

src/cuda/cuda_sampling.cu
```diff
@@ -1,601 +1,394 @@
...
-  // Sample kernel
-  int sample_range = k;
-  int indices_stride = (k > 0 && k <= 64) ? k : vocab_size;
-  LaunchSampleKernel(data, stream, data->scores_sorted.get(), data->indices_sorted.get(), next_token_out, sample_range, batch_size, indices_stride, p, k);
+template <bool is_log_softmax>
+void DispatchBlockwiseSoftmaxForward(cudaStream_t stream, float* output, const float* input, int softmax_elements,
+                                     int input_stride, int output_stride, int batch_count) {
+  // This kernel is efficient for large softmax_elements (like vocab_size) where
+  // a single block can cooperatively process one batch item.
+  constexpr int kBlockSize = 256;
```
hariharans29, line 380: Is there potential to increase this from 256 for larger vocab_sizes ?

src/cuda/cuda_topk.h
```diff
@@ -0,0 +1,70 @@
...
+
+  // - Full sort: General-purpose temporary storage for CUB's DeviceSegmentedRadixSort
+  cuda_unique_ptr<unsigned char> cub_temp_storage;
+  size_t cub_temp_storage_bytes = 0;
+
+  // - Full sort: Stores the start offset of each batch segment for CUB's segmented sort
+  cuda_unique_ptr<int> batch_offsets;
+
+  // --- Information of Final Output (Input to Sampling Stage) ---
+  const float* topk_scores = nullptr;  // pointer to the top-k scores data (in either intermediate_scores_1 or intermediate_scores_2)
```
hariharans29, line 43: Wouldn't both sorting approaches point to `intermediate_scores_1` as the final result ?

That is what the comments above seems to suggest:

  // - Full sort: Holds the fully sorted raw scores.
  // - Selection sort: Holds the top-k scores for selection sort.
  cuda_unique_ptr<float> intermediate_scores_1;

  // - Selection sort: Holds a copy of input scores. Will be updated in place by selection sort kernel.
  cuda_unique_ptr<float> intermediate_scores_2;

tianleiwu: Yes for now.
New algorithm (like Hybrid Sort), the top_k scores might point to intermediate_scores_2 (depending on ping-pong rounds). So I keep the comments like this.

test/CMakeLists.txt
```diff
@@ -1,46 +1,66 @@
...
+
+# Add an option to enable/disable CUDA kernel tests. This option is ON by default
+# if building on non-Windows platform with CUDA available, and OFF otherwise.
+cmake_dependent_option(ENABLE_CUDA_KERNEL_TESTS "Build cuda kernel tests" ON "NOT WIN32;USE_CUDA;CMAKE_CUDA_COMPILER" OFF)
```
hariharans29, line 5: Any reason why Windows need not run the CUDA tests ?

tianleiwu: It is disabled in Windows for now since the cuda kernel tests failed to launch in Windows CI machine.
We can enable it later after figuring out the root cause.

hariharans29: Maybe add a TODO there mentioning this reason ?

tianleiwu: Will do.

src/cuda/cuda_sampling.cu
tianleiwu: It is possible. I can do some experiments later.

PR #1733, opened 2025-09-06
Fix CUDA synchronization issue between ORT-GenAI and TRT-RTX inference

anujj:
Fix CUDA synchronization issue between ORT-GenAI and TRT-RTX inference

Problem:
- Race condition between ORT-GenAI CUDA operations 
  and TRT-RTX inference execution
- CUDA operations were not completing before session.Run() was called
- This caused incorrect inference outputs when GPU sampling was enabled

Solution:
- configure user_compute_stream for NvTensorRtRtx provider

### DIFF
--- a/src/models/model.cpp
+++ b/src/models/model.cpp
@@ -734,7 +734,7 @@ Model::Model(std::unique_ptr<Config> config) : config_{std::move(config)} {
   EnsureDeviceOrtInit(*p_device_, *config_);
 
   // Only CUDA, TRT-RTX and DML does every input on the device
-  if (p_device_->GetType() == DeviceType::CUDA || p_device_->GetType() == DeviceType::DML || p_device_->GetType() == DeviceType::NvTensorRtRtx)
+  if (p_device_->GetType() == DeviceType::CUDA || p_device_->GetType() == DeviceType::DML)
     p_device_inputs_ = p_device_;
   else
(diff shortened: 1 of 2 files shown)


anujj:
@gaugarg-nv for review

src/models/model.cpp
```diff
@@ -590,6 +594,18 @@ DeviceInterface* SetProviderSessionOptions(OrtSessionOptions& session_options,
       }
 
       std::vector<const char*> keys, values;
+
+      std::string stream_value_str;
```
baijumeswani, line 597:
```suggestion
```

src/models/model.cpp
```diff
@@ -590,6 +594,18 @@ DeviceInterface* SetProviderSessionOptions(OrtSessionOptions& session_options,
...
+      if (provider_options.name == "NvTensorRtRtx" && is_primary_session_options && p_device) {
+        void* stream_ptr = p_device->GetCudaStream();
+        std::stringstream stream_value;
+        stream_value << reinterpret_cast<uintptr_t>(stream_ptr);
+        stream_value_str = stream_value.str();
+
+        keys.emplace_back("user_compute_stream");
+        values.emplace_back(stream_value_str.c_str());
```
baijumeswani, line 605:
```suggestion
        keys.emplace_back("user_compute_stream");
        values.emplace_back(stream_value.str().c_str());
```

PR #1745, opened 2025-09-10
Add decode with special tokens

sayanshaw24:
This ORT Extensions [PR](https://github.com/microsoft/onnxruntime-extensions/pull/989) adds a `skip_special_tokens` option in the Decoder (the functionality was previously equivalent to it defaulting to true), and then exposes it to GenAI - now we cannot simply set it to false, because it would break backward compatibility. Hence, in this PR, we introduce `DecodeWithSpecial` so users can leverage it to keep special tokens _specifically_ for the cases they need, such as for models with function calling that require special tool tags.

Now, users can do a simple check for model type and then use `Decode` or `DecodeWithSpecial` accordingly (i.e. `skip_special_tokens = false` for models with function calling, and `skip_special_tokens = true` otherwise).

### Validation
- [x] Native C++ testing in ORT Extensions.

Note: we do not test this added primitive locally in GenAI due to HuggingFace rate limiting, due to which we only have certain models in our test database (and currently do not have function calling models such as phi-4 or qwen-2.5).

### DIFF
--- a/src/csharp/NativeMethods.cs
+++ b/src/csharp/NativeMethods.cs
@@ -249,6 +249,13 @@ namespace Microsoft.ML.OnnxRuntimeGenAI
                                                                               int /* int32_t */ token,
                                                                               out IntPtr /* const char** */ outStr);
 
+        // This function is used to decode the given token into a string (including special token ids).
+        // The returned pointer is freed when the OgaTokenizerStream object is destroyed.
+        [DllImport(NativeLib.DllName, CallingConvention = CallingConvention.Winapi)]
+        public static extern IntPtr /* OgaResult* */ OgaTokenizerStreamDecodeWithSpecial(IntPtr /* const OgaTokenizerStream* */ tokenizerStream,
+                                                                                         int /* int32_t */ token,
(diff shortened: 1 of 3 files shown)


src/csharp/NativeMethods.cs
```diff
@@ -241,6 +249,13 @@ public static extern UIntPtr OgaSequencesGetSequenceCount(IntPtr /* const OgaSeq
                                                                               int /* int32_t */ token,
                                                                               out IntPtr /* const char** */ outStr);
 
+        // This function is used to decode the given token into a string (including special token ids).
+        // The returned pointer is freed when the OgaTokenizerStream object is destroyed.
+        [DllImport(NativeLib.DllName, CallingConvention = CallingConvention.Winapi)]
+        public static extern IntPtr /* OgaResult* */ OgaTokenizerStreamDecodeWithSpecial(IntPtr /* const OgaTokenizerStream* */ tokenizerStream,
```
sayanshaw24, line 255: looks great, thanks for adding the streaming methods as well!

PR #1746, opened 2025-09-11
Support block-wise quant in qmoe op

apsonawane:
Add block-wise quant support for QmoE Op

### DIFF
--- a/src/python/py/models/builder.py
+++ b/src/python/py/models/builder.py
@@ -2412,7 +2412,7 @@ class Model:
             normalize_routing_weights=self.moe_attrs["normalize_routing_weights"],
             swiglu_fusion=self.moe_attrs["swiglu_fusion"],
             use_sparse_mixer=self.moe_attrs["use_sparse_mixer"],
-            block_size=self.moe_attrs.get("actual_block_size", 0),
+            block_size=self.moe_attrs["block_size"],
             **extra_kwargs,
         )
         self.make_value(output, self.io_dtype, shape=['batch_size', 'sequence_length', self.hidden_size])
@@ -2428,17 +2428,13 @@ class Model:
         if block_size > 0:
             try:
                 qweight, scales = self._symmetric_blockwise_quantize(weights, block_size)
-                if self.moe_attrs.get("actual_block_size") != 0:
-                    self.moe_attrs["actual_block_size"] = block_size
+                self.moe_attrs["block_size"] = block_size
                 return qweight, scales.to(torch.float16)
             except Exception as e:
-                print(f"WARNING: Block-wise quantization failed: {e}")
-                print("Falling back to tensor-level quantization")
-                self.moe_attrs["actual_block_size"] = 0
-                # Fall through to existing implementation
+                raise RuntimeError(f"Block-wise quantization failed with block_size={block_size}: {e}")
         else:
             # block_size is 0, so we're using tensor-level quantization
-            self.moe_attrs["actual_block_size"] = 0
+            self.moe_attrs["block_size"] = 0
 
         # Existing tensor-level quantization implementation (fallback)
         unsuccessful = True
@@ -2463,13 +2459,10 @@ class Model:
 
     def _symmetric_blockwise_quantize(self, weights, block_size):
         # Ensure weights are on CPU for quantization
-        original_device = weights.device
         weights = weights.cpu().contiguous()
 
         original_shape = weights.shape
         bits = self.moe_attrs["expert_weight_bits"]
-
-        block_size = original_shape[-1]
 
         qmin, qmax = (-7, 7) if bits == 4 else (-127, 127)
 
@@ -4447,6 +4440,7 @@ def create_model(model_name, input_path, output_dir, precision, execution_provid
             extra_options["exclude_embeds"] = True
             onnx_model = Gemma3Model(config, io_dtype, onnx_dtype, execution_provider, cache_dir, extra_options)
         elif config.architectures[0] == "GptOssForCausalLM":
+            print("WARNING: This model only supports symmetric quantization for `QMoE`.")
             delattr(config, "quantization_config")
             onnx_model = GPTOSSModel(config, io_dtype, onnx_dtype, execution_provider, cache_dir, extra_options)
         elif config.architectures[0] == "GraniteForCausalLM":


src/python/py/models/builder.py
```diff
@@ -2420,6 +2421,20 @@ def make_qmoe_weights(self, weights):
...
+        # Get block size from quantization attributes
+        block_size = self.quant_attrs["int4"]["block_size"]
+
+        # Use block-wise quantization if block_size > 0
+        if block_size > 0:
+            try:
+                qweight, scales = self._symmetric_blockwise_quantize(weights, block_size)
+                return qweight, scales.to(torch.float16)
+            except Exception as e:
+                print(f"WARNING: Block-wise quantization failed: {e}")
```
guschmue: would be better to error out instead of falling back to something that is not what was requested

src/python/py/models/builder.py
```diff
@@ -2440,6 +2455,89 @@ def make_qmoe_weights(self, weights):
 
         return qweight, scales.to(torch.float16)
 
+    def _symmetric_blockwise_quantize(self, weights, block_size):
```
guschmue, line 2460: why not use 
quantize_matmul_4bits
from 
from onnxruntime.capi._pybind_state

src/python/py/models/builder.py
```diff
@@ -2412,6 +2412,7 @@ def make_qmoe_op(self, name, **kwargs):
             normalize_routing_weights=self.moe_attrs["normalize_routing_weights"],
             swiglu_fusion=self.moe_attrs["swiglu_fusion"],
             use_sparse_mixer=self.moe_attrs["use_sparse_mixer"],
+            block_size=self.quant_attrs["int4"]["block_size"] if self.quant_attrs["int4"]["block_size"] > 0 else 0,
```
kunal-vaishnavi: The [default is currently 32](https://github.com/microsoft/onnxruntime-genai/blob/6ae0010a02c2406e9086fc1a5a94212952026c35/src/python/py/models/builder.py#L267) so I think this would get set even if block-wise quantization fails.

src/python/py/models/builder.py
```diff
@@ -2440,6 +2455,89 @@ def make_qmoe_weights(self, weights):
...
+        weights = weights.cpu().contiguous()
+
+        original_shape = weights.shape
+        bits = self.moe_attrs["expert_weight_bits"]
+
+        block_size = original_shape[-1]
+
+        if bits == 4:
+            qmax = 7
+            qmin = -7
```
kunal-vaishnavi: Should we rewrite this into one line?

```qmin, qmax = (-7, 7) if bits == 4 else (-127, 127)```

File: src/python/py/models/builder.py [hunk 5]
```diff
@@ -4157,10 +4255,12 @@ def make_moe_fused(self, layer_id, mlp, root_input):
             down_proj_qweight_list, down_proj_scales_list = [], []
 
             for i in range(self.moe_attrs["num_experts"]):
-                qweight1, scales1 = self.make_qmoe_weights(mlp.experts.gate_up_proj[i])
+                gate_up_weight_transposed = mlp.experts.gate_up_proj[i].transpose(-1, -2)
```
kunal-vaishnavi: Since both the `if` and `else` branches are doing `mlp.experts.gate_up_proj[i].transpose(-1, -2)`, can we always apply the transpose before the branches are entered?

`If` branch:
https://github.com/microsoft/onnxruntime-genai/blob/6ae0010a02c2406e9086fc1a5a94212952026c35/src/python/py/models/builder.py#L4250-L4251

`Else` branch:
https://github.com/microsoft/onnxruntime-genai/blob/6ae0010a02c2406e9086fc1a5a94212952026c35/src/python/py/models/builder.py#L4257-L4265

src/python/py/models/builder.py
```diff
@@ -4169,11 +4269,24 @@ def make_moe_fused(self, layer_id, mlp, root_input):
...
-            self.make_initializer(down_proj_qweight_tensor.view(self.moe_attrs["num_experts"], self.hidden_size, self.intermediate_size // pack_size), down_proj_weight)
-            self.make_initializer(down_proj_scales_tensor, down_proj_scales, to=self.io_dtype)
+            # Handle different tensor shapes for block-wise vs tensor-level quantization
+            block_size = self.quant_attrs["int4"]["block_size"]
+            if block_size > 0:
+                gate_up_qweight_shape = gate_up_proj_qweight_tensor.shape
+                down_qweight_shape = down_proj_qweight_tensor.shape
+
+                # Provide 3D scales as specified by the operator definition
+                self.make_initializer(gate_up_proj_qweight_tensor.view(self.moe_attrs["num_experts"], gate_up_qweight_shape[1], gate_up_qweight_shape[2]), gate_up_proj_weight)
```
kunal-vaishnavi: According to the [QMoE op spec](https://github.com/microsoft/onnxruntime/blob/f418a44c40b86676049951b393f69594c84c0ca4/onnxruntime/core/graph/contrib_ops/contrib_defs.cc#L1447), the shape for the `qweight` tensors is the same whether block-wise quantization is used or not. Only the `scales` tensors have different shapes.

Since both branches use the same logic for saving the `scales` tensors, why does the original logic not work for both cases? For example, the shape for the `gate_up_proj_qweight` tensor should be `(num_experts, fusion_size * inter_size, hidden_size / pack_size)`. The original logic is already performing a `view` operation to make that happen.

src/python/py/models/builder.py
apsonawane: Good catch. Updated

src/python/py/models/builder.py
apsonawane: Updated

src/python/py/models/builder.py
apsonawane: Good suggestion, updated

src/python/py/models/builder.py
apsonawane: Updated

src/python/py/models/builder.py
```diff
@@ -2412,6 +2412,7 @@ def make_qmoe_op(self, name, **kwargs):
             normalize_routing_weights=self.moe_attrs["normalize_routing_weights"],
             swiglu_fusion=self.moe_attrs["swiglu_fusion"],
             use_sparse_mixer=self.moe_attrs["use_sparse_mixer"],
+            block_size=self.moe_attrs.get("actual_block_size", 0),
```
kunal-vaishnavi: We can set the right value for the block size during quantization and always pass its assigned value here.

```suggestion
            block_size=self.moe_attrs["block_size"],
```

src/python/py/models/builder.py
```diff
@@ -2420,6 +2421,26 @@ def make_qmoe_weights(self, weights):
...
         qweight, scales = None, None
 
+        # Get block size from quantization attributes
+        block_size = self.quant_attrs["int4"]["block_size"]
+
+        # Use block-wise quantization if block_size > 0
+        if block_size > 0:
+            try:
+                qweight, scales = self._symmetric_blockwise_quantize(weights, block_size)
+                if self.moe_attrs.get("actual_block_size") != 0:
```
kunal-vaishnavi: Can we rewrite the `if` statement to be just `self.moe_attrs["block_size"] = block_size`? The exception and else branches are taking care of setting the block size to 0 as needed.

src/python/py/models/builder.py
```diff
@@ -2440,6 +2461,84 @@ def make_qmoe_weights(self, weights):
 
         return qweight, scales.to(torch.float16)
 
+    def _symmetric_blockwise_quantize(self, weights, block_size):
```
kunal-vaishnavi, line 2460: Will asymmetric quantization also be supported? If not, then we should add a note when creating the model that symmetric quantization is always done on the MoE op. For example:

https://github.com/microsoft/onnxruntime-genai/blob/27c591f9cd47951b2055c66481ef32e19b4ddb69/src/python/py/models/builder.py#L4363-L4364

src/python/py/models/builder.py
```diff
@@ -2440,6 +2461,84 @@ def make_qmoe_weights(self, weights):
...
+        # Ensure weights are on CPU for quantization
+        original_device = weights.device
+        weights = weights.cpu().contiguous()
+
+        original_shape = weights.shape
+        bits = self.moe_attrs["expert_weight_bits"]
+
+        block_size = original_shape[-1]
```
guschmue: i think you do not want to override the block_size that was passed argument, else bs=5880.

src/python/py/models/builder.py
apsonawane: Fixed

src/python/py/models/builder.py
apsonawane: Fixed

src/python/py/models/builder.py
apsonawane: Added warning

src/python/py/models/builder.py
apsonawane: Yes this was a bug, fixed it

src/python/py/models/builder.py
apsonawane: I haven't tried it, will give it a try

src/python/py/models/builder.py
apsonawane: Updated

kunal-vaishnavi (approved):

src/python/py/models/builder.py
apsonawane: will make this change in another PR

PR #1756, opened 2025-09-15
Mpasumarthi/nvtrt test suite

mpasumarthi-git:
Extend GenAI test suite to support NvTensorRtRtx EP by switching to a newer model and adding deterministic tests for API flow, batch size > 1, sampling, and out-of-place KV cache.


anujj:
can you please guard trt-rtx tests under USE_TRT_RTX

this is added with the PR:https://github.com/microsoft/onnxruntime-genai/commit/b236fb82a2002229269c5981daed4d95d434283b, u need tor rebase your change

test/model_tests.cpp
```diff
@@ -249,6 +249,180 @@ TEST(ModelTests, BeamSearchGptCuda) {
 }
 #endif
 
+#if USE_TRT_RTX
```
baijumeswani: Why do you need this ifdef?

test/model_tests.cpp
```diff
@@ -249,6 +249,180 @@ TEST(ModelTests, BeamSearchGptCuda) {
...
+// NvTensorRT test cases using Phi3 models
+static const std::pair<const char*, const char*> c_phi3_nvtrt_model_paths[] = {
+    {MODEL_PATH "hf-internal-testing/phi3-fp16-nvtrt", "fp16"},
+};
+
+void Test_GreedySearch_Phi3_NvTensorRtRtx(const char* model_path, const char* model_label) {
+  std::vector<int64_t> input_ids_shape{1, 19};
+  std::vector<int32_t> input_ids{32006, 887, 526, 263, 8444, 29871, 23869, 20255, 29889, 32007, 32010, 6324, 29892, 1128, 526, 366, 29973, 32007, 32001};
```
baijumeswani:
```suggestion
  const std::vector<int64_t> input_ids_shape{1, 19};
  const std::vector<int32_t> input_ids{32006, 887, 526, 263, 8444, 29871, 23869, 20255, 29889, 32007, 32010, 6324, 29892, 1128, 526, 366, 29973, 32007, 32001};
```

test/model_tests.cpp
```diff
@@ -249,6 +249,180 @@ TEST(ModelTests, BeamSearchGptCuda) {
...
+
+  // Complete expected sequence (input + generated) from model_qa.cpp using the working phi3-fp16-nvtrt model
+  std::vector<int32_t> expected_output{
```
baijumeswani:
```suggestion
  const std::vector<int32_t> expected_output{
```

test/model_tests.cpp
```diff
@@ -249,6 +249,180 @@ TEST(ModelTests, BeamSearchGptCuda) {
...
+    32006, 887, 526, 263, 8444, 29871, 23869, 20255, 29889, 32007, 32010, 6324, 29892, 1128, 526, 366, 29973, 32007, 32001,  // Input tokens (19)
+    15043, 29991, 306, 29915, 29885, 2599 
+  };
+  auto config = OgaConfig::Create(model_path);
+  config->ClearProviders();
+  config->AppendProvider("NvTensorRtRtx");
+  auto model = OgaModel::Create(*config);
+
+  int max_length = 25;
```
baijumeswani:
```suggestion
  constexpr int max_length = 25;
```

test/model_tests.cpp
```diff
@@ -249,6 +249,180 @@ TEST(ModelTests, BeamSearchGptCuda) {
 }
 #endif
 
+#if USE_TRT_RTX
+// NvTensorRT test cases using Phi3 models
+static const std::pair<const char*, const char*> c_phi3_nvtrt_model_paths[] = {
+    {MODEL_PATH "hf-internal-testing/phi3-fp16-nvtrt", "fp16"},
+};
+
+void Test_GreedySearch_Phi3_NvTensorRtRtx(const char* model_path, const char* model_label) {
```
baijumeswani, line 258: Are the test values specific to the model or to the ep?

test/model_tests.cpp
```diff
@@ -249,6 +249,180 @@ TEST(ModelTests, BeamSearchGptCuda) {
...
+  }
+}
+
+TEST(ModelTests, GreedySearchPhi3NvTensorRtRtx) {
+  for (auto model_path : c_phi3_nvtrt_model_paths)
+    Test_GreedySearch_Phi3_NvTensorRtRtx(model_path.first, model_path.second);
+}
+
+void Test_OutOfPlaceKvCache_Phi3_NvTensorRtRtx(const char* model_path, const char* model_label) {
+  std::vector<int64_t> input_ids_shape{1, 19};
```
baijumeswani: Could you please apply the same consts to this function as well?

test/model_tests.cpp
```diff
@@ -249,6 +249,180 @@ TEST(ModelTests, BeamSearchGptCuda) {
...
+  auto sequence_before_rewind = generator1->GetSequence(0);
+  std::cout << "Sequence before rewind (length " << sequence_before_rewind.size() << "): ";
+  for (size_t j = 0; j < sequence_before_rewind.size(); j++) {
+    std::cout << sequence_before_rewind[j];
+    if (j < sequence_before_rewind.size() - 1) std::cout << ", ";
+  }
+  std::cout << std::endl;
+
+  // Test out-of-place KV cache functionality by rewinding to a previous position
+  // This should create new tensor allocations (out-of-place) rather than modifying existing ones
```
baijumeswani: Could you please explain why rewinding tests out-of-place kv cache? what does out-of-place kv cache mean?

test/sampling_benchmark.cpp
```diff
@@ -178,6 +182,9 @@ TEST(SamplingBenchmarks, PerformanceTests) {
 #if USE_CUDA
   device_types.push_back("cuda");
 #endif
+#if USE_TRT_RTX
```
baijumeswani: Could we avoid using this macro anywhere in the code?

test/sampling_tests.cpp
```diff
@@ -544,3 +544,369 @@ TEST(SamplingTests, RandomizedSamplingSelectTopCuda_BatchSize1_LargeVocabSize) {
   }
 }
 #endif
+
+#if USE_TRT_RTX
+TEST(SamplingTests, BatchedSamplingTopPNvTensorRtRtx) {
```
baijumeswani, line 584: Please test based on model availability instead of using build time flags if possible?

Is there anything in this test that is nv-trt ep specific?

test/model_tests.cpp
mpasumarthi-git: Done.

test/model_tests.cpp
mpasumarthi-git: Done.

test/model_tests.cpp
mpasumarthi-git: Done.

test/model_tests.cpp
mpasumarthi-git: Done.

test/model_tests.cpp
mpasumarthi-git: This is to guard the tests and only build when NVTRT EP flag is enabled.

test/model_tests.cpp
mpasumarthi-git: These values are EP specific.

test/model_tests.cpp
mpasumarthi-git: Updated this test case, please have a look at the latest one.
Out-of-place KV cache creates new memory buffers for each generation step instead of reusing the same buffer.

test/sampling_tests.cpp
mpasumarthi-git: We are following the existing pattern used for other EPs throughout the test suite. The USE_TRT_RTX macro is consistent with how USE_CUDA and USE_DML are used. This is to maintain consistent test organization across the codebase and run these tests only if nv-trt EP is available.

>Is there anything in this test that is nv-trt ep specific?

Yes - this test is NV-TRT EP specific because it uses the NvTensorRtRtx provider.

test/model_tests.cpp
baijumeswani: Could we avoid that? I would like to remove the --use_trt and --use_cuda in the future if possible (or at least remove the compile definitions for them). So far, we didn't have any compile definitions for trtrtx ep. Now, with this PR, we are introducing that.

My main point being that there aren't any API differences or real compile time requirements for nvtrt ep (except for perhaps the onnxruntime-genai-cuda.dll being built).

With that in mind, it would be helpful if we can have a single build that can test all EP models. This will help us when we setup our CI/CD pipeline to test models for all EPs.

So instead of conditionally building the source code / test code based on a compile definition, could we enable the test based on the `hf-internal-testing/phi3-fp16-nvtrt` path being a valid model path?

All you need to do is `GTEST_SKIP()` if model path does not exist.

anujj: Just to add, there is one more compile-time definition: when we build with -use_trt_rtx, the build wheel has dependencies on the onnxruntime-trt-rtx EP wheel, and when it’s built with -use_cuda, the wheel has dependencies on onnxruntime-cuda.

mpasumarthi-git: If we remove USE_TRT_RTX, as Anuj also mentioned, we'll have conflicts with wheel dependencies at build time. TRT tests won't compile due to missing library dependencies. It breaks consistency with other EPs that follow the same pattern.

We can add GTEST_SKIP() inside the `#if USE_TRT_RTX` blocks for runtime model availability checks if that helps with CI/CD flexibility.

anujj: I think the tests should compile, and we can use the hf-internal-testing/phi3-fp16-nvtrt path to run the NVTRT tests. However, we still need the --use_trt_rtx flag in GenAI to complement the onnxruntime-genai-trt-rtx wheel. I’ve added this in the PR: https://github.com/microsoft/onnxruntime-genai/pull/1736

mpasumarthi-git: Updated all NvTensorRT tests to use runtime model availability checks `(GTEST_SKIP())` instead of compile-time `#if USE_TRT_RTX` guards.

test/sampling_tests.cpp
```diff
@@ -544,3 +545,395 @@ TEST(SamplingTests, RandomizedSamplingSelectTopCuda_BatchSize1_LargeVocabSize) {
...
+  int batch_size = 4;
+  int vocab_size = 5;
+
+  auto config = OgaConfig::Create(MODEL_PATH "hf-internal-testing/phi3-fp16-nvtrt");
+  config->Overlay(R"({ "model": { "vocab_size" : 5 } })");
+  config->ClearProviders();
+  config->AppendProvider("NvTensorRtRtx");
+
+  auto model = OgaModel::Create(*config);
+  auto params = OgaGeneratorParams::Create(*model);
```
kunal-vaishnavi: Can you refactor the unit tests to have environment set ups for the parts that repeat across the tests? For example, I see `max_length` and `batch_size` set every time.

mpasumarthi-git: Done.

kunal-vaishnavi (dismissed):

PR #1762, opened 2025-09-16
Disable lmhead while prompt processing

qti-ashimaj:
Disable executing lm head model in the decoder-only pipeline while prompt is being processed for all except the last sliding window. This reduces the time taken in the prefill phase and leads to better ttft metric, especially for larger prompts.
To enable this, set `"is_lm_head": true` for lm_head model in the genai_config.json


qti-ashimaj:
@baijumeswani please review. Thanks!

baijumeswani (approved):

PR #1763, opened 2025-09-16
Use abbreviation for TensorRT RTX EP

kunal-vaishnavi:
### Description

This PR replaces any references to `NvTensorRtRtx` with `trt-rtx` except in the GenAI config file.

### Motivation and Context

This abbreviation reduces [potential bugs](https://github.com/microsoft/onnxruntime-genai/pull/1687#discussion_r2286022857) that can be raised while using the EP name in the model builder. It also preserves the original intention to keep EP names short and brief.

### DIFF
--- a/src/python/py/models/builder.py
+++ b/src/python/py/models/builder.py
@@ -82,7 +82,7 @@ class Model:
             },
             "dml": {},
             "webgpu": {},
-            "NvTensorRtRtx": {"enable_cuda_graph": "1"}
+            "trt-rtx": {"enable_cuda_graph": "1"}
         }
 
         # Map input names to their types and shapes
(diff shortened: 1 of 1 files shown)


thevishalagarwal:
Can we change all references of `NvTensorRtRtx` in the code to `trt-rtx`? Making it consistent across all scripts and benchmarks

gaugarg-nv:
@kunal-vaishnavi I think this will also break the inference for TRT-RTX EP. 

The generated `genai_config.json` will have `trt-rtx` as the EP name with this change. But we expect the name to be `NvTensorRtRtx`.

apsonawane (approved):

PR #1765, opened 2025-09-16
Add Prefix Chunking Support for NvTensorRtRtx and Cuda Providers

anujj:
This PR introduces prefix chunking functionality to improve memory efficiency and performance when processing long input sequences during the context/prefill phase. The feature is specifically enabled for NvTensorRtRtx and Cuda execution providers.

Key Changes
🚀 New Features
- Prefix Chunking Implementation: Added support for processing long input sequences in configurable chunks during the context phase
- Device-Specific Enablement: Chunking is only activated for NvTensorRtRtx and CUDA providers 
- Configuration Support: New chunk_size parameter exposed through the config overlay API

Configuration Changes
- Added chunk_size parameter to search configuration (src/config.h, src/config.cpp)
- Default value: -1 (disabled), positive values enable chunking with specified size
- or enable via overlay apis: `config.overlay(f'{{"search": {"chunk_size": {args.chunk_size}}}}}')`

### DIFF
--- a/examples/python/model-generate.py
+++ b/examples/python/model-generate.py
@@ -19,7 +19,8 @@ def main(args):
     batch_size = len(prompts)
 
     config = og.Config(args.model_path)
-    config.overlay(f'{{"search": {{"batch_size": {batch_size}, "num_beams": {3}}}}}')
+    # Example: Configure search parameters including chunk_size for prefix chunking  
+    config.overlay(f'{{"search": {{"batch_size": {batch_size}, "num_beams": {3}, "chunk_size": {args.chunk_size}}}}}')
 
     if args.execution_provider != "follow_config":
         config.clear_providers()
@@ -90,6 +91,7 @@ if __name__ == "__main__":
     parser.add_argument('-v', '--verbose', action='store_true', default=False, help='Print verbose output and timing information. Defaults to false')
     parser.add_argument('-b', '--batch_size_for_cuda_graph', type=int, default=1, help='Max batch size for CUDA graph')
     parser.add_argument('-c', '--chat_template', type=str, default='', help='Chat template to use for the prompt. User input will be injected into {input}. If not set, the prompt is used as is.')
+    parser.add_argument('--chunk_size', type=int, default=-1, help='Chunk size for prefix chunking during context processing (default: -1 = disabled, >0 = enabled)')
     parser.add_argument('--non-interactive', action=argparse.BooleanOptionalAction, required=False, default=False, help='Non-interactive mode, mainly for CI usage')
 
     args = parser.parse_args()
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -868,6 +868,8 @@ struct Search_Element : JSON::Element {
       v_.length_penalty = static_cast<float>(JSON::Get<double>(value));
     } else if (name == "random_seed") {
       v_.random_seed = SafeDoubleToInt(JSON::Get<double>(value), name);
+    } else if (name == "chunk_size") {
+      v_.chunk_size = static_cast<int>(JSON::Get<double>(value));
     } else if (name == "do_sample") {
       v_.do_sample = JSON::Get<bool>(value);
     } else if (name == "past_present_share_buffer") {
--- a/src/config.h
+++ b/src/config.h
@@ -273,6 +273,7 @@ struct Config {
     float length_penalty{1.0f};        // Exponential penalty to the length that is used with beam-based generation. length_penalty > 0.0 promotes longer sequences, while length_penalty < 0.0 encourages shorter sequences.
     bool past_present_share_buffer{};  // The past/present kv tensors are shared and allocated once to max_length (cuda only)
     int random_seed{-1};               // -1 = Seed with random device, otherwise use value to seed RNG
+    int chunk_size{-1};              // Chunk size for prefix chunking during context processing. -1 = disabled, >0 = enabled with specified chunk size.
   } search;
 
   void AddMapping(const std::string& nominal_name, const std::string& graph_name);
--- a/src/models/decoder_only.cpp
+++ b/src/models/decoder_only.cpp
@@ -29,10 +29,10 @@ void DecoderOnly_State::SetExtraInputs(const std::vector<ExtraInput>& extra_inpu
 
 DeviceSpan<float> DecoderOnly_State::Run(int total_length, DeviceSpan<int32_t>& next_tokens, DeviceSpan<int32_t> next_indices) {
   size_t num_tokens = next_tokens.size();
-  const size_t chunk_size = 1024; // Experimental value
+  const size_t chunk_size = static_cast<size_t>(model_.config_->search.chunk_size);
   
-  if (num_tokens > chunk_size) {
-    // Chunking logic for context phase - process in chunks of 512 tokens
+  if (chunk_size > 0 && num_tokens > chunk_size) {
+    // Chunking logic for context phase - process in chunks based on configured chunk_size
     size_t processed_tokens = 0;
     int length = total_length - static_cast<int>(num_tokens);
     while (processed_tokens < num_tokens) {
--- a/src/models/decoder_only.cpp
+++ b/src/models/decoder_only.cpp
@@ -31,7 +31,11 @@ DeviceSpan<float> DecoderOnly_State::Run(int total_length, DeviceSpan<int32_t>& 
   size_t num_tokens = next_tokens.size();
   const size_t chunk_size = static_cast<size_t>(model_.config_->search.chunk_size);
   
-  if (chunk_size > 0 && num_tokens > chunk_size) {
+  // Enable prefill chunking for CUDA and NvTensorRtRtx devices
+  bool is_chunking_supported_device = (model_.p_device_->GetType() == DeviceType::CUDA ||
+                                       model_.p_device_->GetType() == DeviceType::NvTensorRtRtx);
+
+  if (is_chunking_supported_device && chunk_size > 0 && num_tokens > chunk_size) {
     // Chunking logic for context phase - process in chunks based on configured chunk_size
     size_t processed_tokens = 0;
     int length = total_length - static_cast<int>(num_tokens);
@@ -55,7 +59,8 @@ DeviceSpan<float> DecoderOnly_State::Run(int total_length, DeviceSpan<int32_t>& 
     // Return logits from the last chunk for potential sampling
     return logits_.Get();
   } else {
-    // Original logic for tokens <= 512 (generation phase or small context)
+    // Original logic for tokens <= chunk_size (generation phase or small context) 
+    // or chunking disabled due to unsupported device
     UpdateInputsOutputs(next_tokens, next_indices, total_length);
 
     // Graph capture enabled for token generation case, allowing it to repeat the same graph for each token.
--- a/src/models/decoder_only.cpp
+++ b/src/models/decoder_only.cpp
@@ -29,7 +29,7 @@ void DecoderOnly_State::SetExtraInputs(const std::vector<ExtraInput>& extra_inpu
 
 DeviceSpan<float> DecoderOnly_State::Run(int total_length, DeviceSpan<int32_t>& next_tokens, DeviceSpan<int32_t> next_indices) {
   size_t num_tokens = next_tokens.size();
-  const size_t chunk_size = 15;
+  const size_t chunk_size = 1024; // Experimental value
   
   if (num_tokens > chunk_size) {
     // Chunking logic for context phase - process in chunks of 512 tokens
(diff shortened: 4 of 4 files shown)


src/models/decoder_only.cpp
```diff
@@ -28,13 +28,47 @@ void DecoderOnly_State::SetExtraInputs(const std::vector<ExtraInput>& extra_inpu
 }
 
 DeviceSpan<float> DecoderOnly_State::Run(int total_length, DeviceSpan<int32_t>& next_tokens, DeviceSpan<int32_t> next_indices) {
-  UpdateInputsOutputs(next_tokens, next_indices, total_length);
+  size_t num_tokens = next_tokens.size();
+  const size_t chunk_size = static_cast<size_t>(model_.config_->search.chunk_size);
+  
+  // Enable prefill chunking for CUDA and NvTensorRtRtx devices
+  bool is_chunking_supported_device = (model_.p_device_->GetType() == DeviceType::CUDA ||
```
kunal-vaishnavi: Can this be enabled for CPU as well?

anujj: @kunal-vaishnavi :Ideally, this change should work for all devices where continue decoding is enabled. Today, I tried to verify it on CPU, but even without these changes I am not able to run inference on the CPU device due to the following error:

(genai) D:\code\anuj_github\onnxruntime-genai\examples\python>python model-qa.py -m D:\models\llama_3_2_1b_inst\cpu                                                                       
Prompt (Use quit() to exit): hi

Output: How2025-09-23 16:14:15.6961515 [E:onnxruntime:onnxruntime-genai, sequential_executor.cc:572 onnxruntime::ExecuteKernel] Non-zero status code returned while running Cast node. Name:'InsertedPrecisionFreeCast_/model/layers.1/attn/v_proj/repeat_kv/Reshape_4/output_0' Status Message: D:\a\_work\1\s\onnxruntime\core\framework\op_kernel.cc:83 onnxruntime::OpKernelContext::OutputMLValue status.IsOK() was false. Shape mismatch attempting to re-use buffer. {1,1,2048} != {1,44,2048}. Validate usage of dim_value (values should be > 0) and dim_param (all values with the same string should equate to the same size) in shapes in the model.


Not sure if something is broken for CPU.

I have validated this feature for NvTensorRtRtx and CUDA, so we can target these two devices in this PR?

examples/python/model-generate.py
```diff
@@ -19,7 +19,8 @@ def main(args):
     batch_size = len(prompts)
 
     config = og.Config(args.model_path)
-    config.overlay(f'{{"search": {{"batch_size": {batch_size}, "num_beams": {3}}}}}')
+    # Example: Configure search parameters including chunk_size for prefix chunking  
+    config.overlay(f'{{"search": {{"batch_size": {batch_size}, "num_beams": {3}, "chunk_size": {args.chunk_size}}}}}')
```
baijumeswani: Could you add chunking setting conditionally only for trt rtx ep? I don't think all eps may be able to support it for batch size > 1

src/config.h
```diff
@@ -273,6 +273,7 @@ struct Config {
     float length_penalty{1.0f};        // Exponential penalty to the length that is used with beam-based generation. length_penalty > 0.0 promotes longer sequences, while length_penalty < 0.0 encourages shorter sequences.
     bool past_present_share_buffer{};  // The past/present kv tensors are shared and allocated once to max_length (cuda only)
     int random_seed{-1};               // -1 = Seed with random device, otherwise use value to seed RNG
+    int chunk_size{-1};              // Chunk size for prefix chunking during context processing. -1 = disabled, >0 = enabled with specified chunk size.
```
baijumeswani: Would it be cleaner to make this optional rather than using -1 to indicate disabled? If the value isn’t found in the GenAI config, we could default to disabling it.

src/models/decoder_only.cpp
baijumeswani: Let's leave this check out of this PR. The genai_config already provides sufficient control to disable chunking, so no additional logic is needed here.

src/models/decoder_only.cpp
```diff
@@ -28,13 +28,47 @@ void DecoderOnly_State::SetExtraInputs(const std::vector<ExtraInput>& extra_inpu
...
+      // Graph capture is typically disabled during context phase chunking
+      bool graph_capture_this_run = false; // Disable graph capture during chunking
+      State::Run(*model_.session_decoder_, graph_capture_this_run);
+      
+      processed_tokens += current_chunk_size;
+    }
+    
+    // Return logits from the last chunk for potential sampling
+    return logits_.Get();
+  } else {
```
baijumeswani, line 36: Could you refactor the while loop (possibly switching to a for loop) so that the if-else blocks can be unified? The only differences between them seem to be the graph capture and chunk size, so consolidating the logic might improve readability and reduce duplication.

baijumeswani (commented):
Could you please update the title and description to clarify that this PR introduces support for prefill chunking, not prefix chunking?

src/models/decoder_only.cpp
kunal-vaishnavi: An `InsertedPrecisionFreeCast` means that you are likely running a float16 model on the CPU EP. I agree that we can leave this check out of the PR.

examples/python/model-generate.py
```diff
@@ -90,6 +91,7 @@ def main(args):
     parser.add_argument('-v', '--verbose', action='store_true', default=False, help='Print verbose output and timing information. Defaults to false')
     parser.add_argument('-b', '--batch_size_for_cuda_graph', type=int, default=1, help='Max batch size for CUDA graph')
     parser.add_argument('-c', '--chat_template', type=str, default='', help='Chat template to use for the prompt. User input will be injected into {input}. If not set, the prompt is used as is.')
+    parser.add_argument('--chunk_size', type=int, default=-1, help='Chunk size for prefix chunking during context processing (default: -1 = disabled, >0 = enabled)')
```
```suggestion
    parser.add_argument('--chunk_size', type=int, default=-1, help='Chunk size for prefill chunking during context processing (default: -1 = disabled, >0 = enabled)')
```

src/config.h
```suggestion
    int chunk_size{-1};              // Chunk size for prefill chunking during context processing. -1 = disabled, >0 = enabled with specified chunk size.
```

src/models/decoder_only.cpp
```diff
@@ -28,13 +28,47 @@ void DecoderOnly_State::SetExtraInputs(const std::vector<ExtraInput>& extra_inpu
...
+  // Enable prefill chunking for CUDA and NvTensorRtRtx devices
+  bool is_chunking_supported_device = (model_.p_device_->GetType() == DeviceType::CUDA ||
+                                       model_.p_device_->GetType() == DeviceType::NvTensorRtRtx);
 
-  // Graph capture enabled for token generation case, allowing it to repeat the same graph for each token.
-  bool graph_capture_this_run = params_->use_graph_capture && input_ids_.GetShape()[1] == 1;
-  State::Run(*model_.session_decoder_, graph_capture_this_run);
+  if (is_chunking_supported_device && chunk_size > 0 && num_tokens > chunk_size) {
+    // Chunking logic for context phase - process in chunks based on configured chunk_size
+    size_t processed_tokens = 0;
```
kunal-vaishnavi: Can the chunking logic be refactored to a virtual method that all pipelines can override (e.g. decoder-only, multi-modal, etc.)?

examples/python/model-generate.py
anujj: done

src/config.h
anujj: Updated the logic

src/models/decoder_only.cpp
anujj: i have added the sperate function for the chunking logic, I guess that should address your concern  ?

examples/python/model-generate.py
anujj: done

src/config.h
anujj: done

src/models/decoder_only.cpp
anujj: done

anujj:
Addressed the review comments

examples/python/model-generate.py
```diff
@@ -19,7 +20,18 @@ def main(args):
...
     config = og.Config(args.model_path)
-    config.overlay(f'{{"search": {{"batch_size": {batch_size}, "num_beams": {3}}}}}')
+
+    search_config = {
+        "batch_size": batch_size,
+        "num_beams": 3
+    }
+
+    # Add chunk_size only for NvTensorRtRtx execution provider
+    if args.execution_provider == "NvTensorRtRtx" and args.chunk_size > 0:
```
kunal-vaishnavi, line 30: Why is the chunk size not needed for the CUDA EP?

examples/python/model-generate.py
```diff
@@ -90,6 +102,7 @@ def main(args):
     parser.add_argument('-v', '--verbose', action='store_true', default=False, help='Print verbose output and timing information. Defaults to false')
     parser.add_argument('-b', '--batch_size_for_cuda_graph', type=int, default=1, help='Max batch size for CUDA graph')
     parser.add_argument('-c', '--chat_template', type=str, default='', help='Chat template to use for the prompt. User input will be injected into {input}. If not set, the prompt is used as is.')
+    parser.add_argument('--chunk_size', type=int, default=-1, help='Chunk size for prefill chunking during context processing (default: 0 = disabled, >0 = enabled)')
```
```suggestion
    parser.add_argument('--chunk_size', type=int, default=-1, help='Chunk size for prefill chunking during context processing (default: -1 = disabled, >0 = enabled)')
```

src/config.cpp
```diff
@@ -868,6 +868,9 @@ struct Search_Element : JSON::Element {
       v_.length_penalty = static_cast<float>(JSON::Get<double>(value));
     } else if (name == "random_seed") {
       v_.random_seed = SafeDoubleToInt(JSON::Get<double>(value), name);
+    } else if (name == "chunk_size") {
+      // Optional: defaults to 0 (disabled) if not specified
```
kunal-vaishnavi:
```suggestion
      // Optional: defaults to -1 (disabled) if not specified
```

examples/python/model-generate.py
anujj: i enabled this feature for the NvTensorRtRtx only for now

src/models/decoder_only.cpp
baijumeswani: I meant could we possibly combine the logic in the two `if...else` blocks. It seems similar for the most part. As it stands right now, the code in the two blocks is somewhat duplicated.

src/models/decoder_only.cpp
```diff
@@ -50,4 +85,10 @@ void DecoderOnly_State::UpdateInputsOutputs(DeviceSpan<int32_t>& next_tokens, De
   logits_.Update(next_tokens, new_length);
 }
 
+bool DecoderOnly_State::IsChunkingSupported() const {
+  // Enable prefill chunking for CUDA and NvTensorRtRtx devices
+  return (model_.p_device_->GetType() == DeviceType::CUDA ||
+          model_.p_device_->GetType() == DeviceType::NvTensorRtRtx);
```
baijumeswani: Let's avoid using this device check. Instead let's check if the chunking is enabled through the config alone. It should be a sufficient check to avoid doing chunked prefill.

src/config.h
```diff
@@ -273,6 +273,7 @@ struct Config {
     float length_penalty{1.0f};        // Exponential penalty to the length that is used with beam-based generation. length_penalty > 0.0 promotes longer sequences, while length_penalty < 0.0 encourages shorter sequences.
     bool past_present_share_buffer{};  // The past/present kv tensors are shared and allocated once to max_length (cuda only)
     int random_seed{-1};               // -1 = Seed with random device, otherwise use value to seed RNG
+    int chunk_size{-1};                // Chunk size for prefill chunking during context processing. -1 = disabled, >0 = enabled with specified chunk size.
```
baijumeswani: Sorry, I should have been clear in my previous comment. Could we use `std::optional<size_t>` to indicate whether chunked prefill is enabled or not. The size_t part can be the chunk size and the absence of it in the genai_config indicates that chunked prefill is not to be done.

anujj: done

src/models/model.h
```diff
@@ -59,6 +59,12 @@ struct State {
 
   std::unique_ptr<OrtRunOptions> run_options_;
 
+  // Virtual methods for prefill chunking support
+  virtual DeviceSpan<float> RunWithChunking(int total_length, DeviceSpan<int32_t>& next_tokens,
+                                            DeviceSpan<int32_t> next_indices, size_t chunk_size) {
+    return DeviceSpan<float>{};
```
baijumeswani: Maybe throw an exception here since this should be implemented in the derived class.

baijumeswani (dismissed):

anujj:
removed RunWithChunking being virtual as discussed

baijumeswani (approved):

PR #1772, opened 2025-09-18
Fix new/delete override and Enable cuda kernel test in Windows

tianleiwu:
Previously, cuda kernel test was disabled in Windows CI since test failed to start.
It is caused by new/delete override, which has assumption that the genai object is created before any other object. The global initialization order depends on compiler.

In Windows, when new is called, the genai object might not be created yet.

Here we change the implementation of the override to fix such potential crash in Windows.

### DIFF
--- a/test/cuda_kernel/cuda_sampling_tests.cpp
+++ b/test/cuda_kernel/cuda_sampling_tests.cpp
@@ -73,7 +73,7 @@ TEST_P(CudaSamplingTopKTopPTest, StatisticalVerification) {
   const int vocab_size = 512;  // Smaller vocab for faster test execution
   const int k = GetParam();
   const float p = 0.9f;
-  const float temperature = 0.7f;
+  constexpr float temperature = 0.7f;
   const int num_iter = 5000;
   const unsigned long long initial_seed = 42;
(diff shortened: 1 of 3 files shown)


test/cuda_kernel/cuda_sampling_tests.cpp
```diff
@@ -73,7 +73,7 @@ TEST_P(CudaSamplingTopKTopPTest, StatisticalVerification) {
   const int vocab_size = 512;  // Smaller vocab for faster test execution
   const int k = GetParam();
   const float p = 0.9f;
-  const float temperature = 0.7f;
+  constexpr float temperature = 0.7f;
```
kunal-vaishnavi, line 76: Should we change some of these other constants to `constexpr` as well?

kunal-vaishnavi (approved):

test/cuda_kernel/cuda_sampling_tests.cpp
tianleiwu: We can. 
This is to fix build warning and it seems to only warn this line since temperature is used later for if condition.

PR #1775, opened 2025-09-18
Support for decoder pipeline based multimodel

prudhvi-qti:


prudhvi-qti:
@microsoft-github-policy-service agree company="Qualcomm"

src/models/embeddings.cpp
```diff
@@ -65,4 +65,88 @@ void Embeddings::ReuseEmbeddingsBuffer(const Embeddings& other) {
   state_.outputs_[index_] = other.state_.inputs_[other.index_];
 }
 
+WindowedEmbeddings::WindowedEmbeddings(State& state, Embeddings::Mode mode, const std::string& name)
```
qti-kromero, line 68: I think this can be reduced to just the following because it's so similar to the parent class constructor:

```c++
WindowedEmbeddings::WindowedEmbeddings(State& state, Embeddings::Mode mode, const std::string& name) 
    : Embeddings(state, mode, name) {
    window_size_ = model_.config_->model.decoder.sliding_window->window_size;
}
```

src/models/embeddings.cpp
```diff
@@ -65,4 +65,88 @@ void Embeddings::ReuseEmbeddingsBuffer(const Embeddings& other) {
...
+
+  // Assuming batch_size = 1
+
+  int64_t sequence_length = full_shape[1];
+  int64_t hidden_size = full_shape[2];
+
+  const uint16_t* full_data = full_embeddings->GetTensorData<uint16_t>();
+
+  if (window_index_ == 0) {
+    num_windows_ = (sequence_length + window_size_ - 1) / window_size_;
```
qti-kromero, line 109: the first two conditions here share a decent amount of initialization? could we combine these two conditions and remove some duplicate code?

src/models/embeddings.h
```diff
@@ -38,4 +38,29 @@ struct Embeddings {
...
-}  // namespace Generators
+struct WindowedEmbeddings : public Embeddings {
+  WindowedEmbeddings(State& state, Embeddings::Mode mode, const std::string& name);
+  WindowedEmbeddings(const WindowedEmbeddings&) = delete;
+  WindowedEmbeddings& operator=(const WindowedEmbeddings&) = delete;
+
+  void Update(Embeddings& embeddings);
+
+ private:
+  State& state_;
```
qti-kromero, line 52: we could in theory update the parent class variables to be protected and then these don't need to be redeclared in the child class

src/models/model.cpp
```diff
@@ -859,7 +859,7 @@ void Model::CreateSessionOptionsFromConfig(const Config::SessionOptions& config_
   auto session_device = SetProviderSessionOptions(session_options, config_session_options.providers,
                                                   config_session_options.provider_options, is_primary_session_options,
                                                   disable_graph_capture, *config_);
-
+
```
qti-kromero, line 862: nit: whitespace

src/models/multi_decoder_pipeline_modal.h
```diff
@@ -0,0 +1,205 @@
...
+#include "embeddings.h"
+#include "extra_inputs.h"
+#include "logits.h"
+#include "kv_cache.h"
+#include "position_inputs.h"
+#include "decoder_only_pipeline.h"
+
+namespace Generators {
+
+struct MultiModalPipelineLanguageModel : Model {
```
qti-kromero, line 17: would suggest trying to align the name of the class here and the name of the file

src/models/multi_decoder_pipeline_modal.cpp
```diff
@@ -0,0 +1,743 @@
...
+#include "multi_decoder_pipeline_modal.h"
+#include "../logging.h"
+#include "../tracing.h"
+#include "windowed_kv_cache.h"
+
+namespace Generators {
+
+namespace {
+
+int64_t GetNumImageTokens(const std::vector<ExtraInput>& extra_inputs) {
```
qti-kromero, line 14: only usage of this function in file is commented out

src/models/multi_decoder_pipeline_modal.cpp
```diff
@@ -0,0 +1,743 @@
...
+      return std::accumulate(num_image_tokens_data,
+                             num_image_tokens_data + extra_inputs[i].tensor->ort_tensor_->GetTensorTypeAndShapeInfo()->GetElementCount(),
+                             0LL);
+    }
+  }
+
+  return 0;
+}
+
+int64_t GetNumAudioTokens(const std::vector<ExtraInput>& extra_inputs,
```
qti-kromero, line 28: would be nice if we could move these to another file where they can both be accessed by this file and the multi_modal.cpp file

src/models/multi_decoder_pipeline_modal.cpp
```diff
@@ -0,0 +1,743 @@
...
+      return extra_inputs[i].tensor->ort_tensor_->GetTensorTypeAndShapeInfo()->GetShape().front();
+    }
+  }
+
+  return 0;
+}
+
+}  // namespace
+
+MultiModalPipelineLanguageModel::MultiModalPipelineLanguageModel(std::unique_ptr<Config> config, OrtEnv& ort_env, bool vision, bool speech)
```
qti-kromero, line 66: feels like with the rest of this file there is a big opportunity for code reuse with multi_modal.cpp if we organize the classes a little better

baijumeswani:
Is this PR still relevant? Please re-open in case it is.

PR #1785, opened 2025-09-19
Add Support For Tokenizer Options

sayanshaw24:
### Updates
This PR introduces support for ORT Extensions introduced in [this PR](https://github.com/microsoft/onnxruntime-extensions/pull/998), which allows passing an **options map** with `OrtxCreateTokenizerWithOptions` when creating tokenizers, or using a new `OrtxUpdateTokenizerOptions` method which allows updating the options map on an existing tokenizer object (including those created using `OrtxCreateTokenizer`), enabling more flexible configurations.

It additionally removes the previously added `OrtxTokenizeWithOptions` and `OrtxDetokenize1DWithOptions` functions, which are now redundant. With the new design, **options are set once on the tokenizer object itself**, so there’s no longer a need to pass ad-hoc option sets into individual tokenize/detokenize calls — reducing API clutter and simplifying the C interface.

In additions to the C API updates, it also adds bindings for C++, C# and Python.

### Sample Usage

C++
```

// Define tokenizer options as C-style arrays
const char* keys[] = {"add_special_tokens", "skip_special_tokens"};
const char* values[] = {"true", "false"};

// Update tokenizer options
tokenizer->UpdateOptions(keys, values, 2);
```

C#
```
var tokenizer = new Tokenizer(model);

// Update tokenizer options using a dictionary
var options = new Dictionary<string, string>
{
    { "add_special_tokens", "true" },
    { "skip_special_tokens", "false" }
};

tokenizer.UpdateOptions(options);
```

Python
```
tokenizer = Tokenizer(model)

options = {
    "add_special_tokens": "true",
    "skip_special_tokens": "false"
}

tokenizer.update_options(**options)
```

### DIFF
--- a/test/c_api_tests.cpp
+++ b/test/c_api_tests.cpp
@@ -120,12 +120,11 @@ TEST(CAPITests, TokenizerUpdateOptions) {
   auto tokenizer = OgaTokenizer::Create(*model);
 
   // Update tokenizer options
-  // Note: This simply tests the OgaUpdateTokenizerOptions API; these options are already set as default.
+  // Note: This simply tests the UpdateOptions API; these options are already set as default.
   {
     const char* keys[] = {"add_special_tokens", "skip_special_tokens"};
     const char* values[] = {"false", "true"};
-    OgaResult* result = OgaUpdateTokenizerOptions(tokenizer.get(), keys, values, 2);
-    ASSERT_EQ(result, nullptr) << "Failed to update tokenizer options";
+    tokenizer->UpdateOptions(keys, values, 2);
   }
 
   // Encode single decode single


natke (commented):
Can you add tests please?

src/python/python.cpp
```diff
@@ -380,6 +380,17 @@ PYBIND11_MODULE(onnxruntime_genai, m) {
 
   pybind11::class_<OgaTokenizer>(m, "Tokenizer")
       .def(pybind11::init([](const OgaModel& model) { return OgaTokenizer::Create(model); }))
+      .def("update_options", [](OgaTokenizer& t, std::map<std::string, std::string> options) {
```
baijumeswani: Can we take in kwargs instead of a dict?

src/ort_genai_c.h
```diff
@@ -572,6 +572,24 @@ OGA_EXPORT OgaResult* OGA_API_CALL OgaCreateMultiModalProcessor(const OgaModel*
...
+ * @param values Array of corresponding option value strings.
+ * @param num_options Number of key/value pairs provided.
+ *
+ * @return nullptr on success, or an OgaResult* describing the error.
+ *         The returned OgaResult* (if not null) must be freed with OgaDestroyResult.
+ */
+OGA_EXPORT OgaResult* OGA_API_CALL OgaUpdateTokenizerOptions(
+    const OgaTokenizer* tokenizer,
+    const char* keys[],
+    const char* values[],
```
baijumeswani: Should these be `const char* const*` instead of `const char* keys[]`?

src/ort_genai_c.cpp
```diff
@@ -581,6 +581,32 @@ OgaResult* OGA_API_CALL OgaCreateTokenizer(const OgaModel* model, OgaTokenizer**
...
+  if (!tokenizer)
+    throw std::runtime_error("Tokenizer pointer is null");
+
+  // Build std::unordered_map from the arrays
+  std::unordered_map<std::string, std::string> options;
+  for (size_t i = 0; i < num_options; ++i) {
+    if (keys[i] && values[i]) {
+      options[keys[i]] = values[i];
+    }
+  }
```
baijumeswani: What's the rationale for converting the key/value lists into an unordered_map and then back to an array of strings before calling the ORT Extensions tokenizer? Can we avoid the map round‑trip and pass strings directly?

src/models/model.cpp
```diff
@@ -266,9 +266,26 @@ std::unique_ptr<TokenizerStream> Tokenizer::CreateStream() const {
...
+  keys.reserve(options.size());
+  values.reserve(options.size());
+
+  for (const auto& kv : options) {
+    keys.push_back(kv.first.c_str());
+    values.push_back(kv.second.c_str());
+  }
+
+  // Tap into ORT Extensions API
+  CheckResult(OrtxUpdateTokenizerOptions(const_cast<OrtxTokenizer*>(static_cast<const OrtxTokenizer*>(tokenizer_)), keys.data(), values.data(), options.size()));
```
baijumeswani: Why do we need to do this?

```
const_cast<OrtxTokenizer*>(static_cast<const OrtxTokenizer*>(tokenizer_))
```

src/ort_genai_c.h
```diff
@@ -572,6 +572,24 @@ OGA_EXPORT OgaResult* OGA_API_CALL OgaCreateMultiModalProcessor(const OgaModel*
...
+ * @param tokenizer Pointer to the OgaTokenizer whose options will be updated.
+ * @param keys Array of option key strings.
+ * @param values Array of corresponding option value strings.
+ * @param num_options Number of key/value pairs provided.
+ *
+ * @return nullptr on success, or an OgaResult* describing the error.
+ *         The returned OgaResult* (if not null) must be freed with OgaDestroyResult.
+ */
+OGA_EXPORT OgaResult* OGA_API_CALL OgaUpdateTokenizerOptions(
+    const OgaTokenizer* tokenizer,
```
baijumeswani: Since we are making changes to the tokenizer, the input param should be marked non const.

src/ort_genai_c.h
```diff
@@ -572,6 +572,24 @@ OGA_EXPORT OgaResult* OGA_API_CALL OgaCreateMultiModalProcessor(const OgaModel*
...
 OGA_EXPORT void OGA_API_CALL OgaDestroyMultiModalProcessor(OgaMultiModalProcessor* processor);
 
+/**
+ * Updates tokenizer options for the given OgaTokenizer instance.
+ * The provided keys and values must be null-terminated C strings.
+ *
+ * @param tokenizer Pointer to the OgaTokenizer whose options will be updated.
+ * @param keys Array of option key strings.
+ * @param values Array of corresponding option value strings.
+ * @param num_options Number of key/value pairs provided.
```
baijumeswani, line 586: How would the user of this API determine what options are available? Shall we document the support of available options here? Or redirect users to some readme / page where the options are available?

src/ort_genai_c.h
sayanshaw24: good catch, yep they should, since the pointers themselves won't be modified.

src/ort_genai_c.cpp
sayanshaw24: great catch! this is redundant i missed updating this after adding the array UpdateOptions, it was temporary.

src/models/model.cpp
sayanshaw24: because `OrtxUpdateTokenizerOptions` takes in a non const tokenizer in Extensions (which is right in my opinion, as we should only be making updates to a non const tokenizer), and `tokenizer_` here is const.

still, if we want to avoid it we could change this in Extensions but it would require another Ext change, wait for CIs, merge and update on GenAI - i think it's unnecessary for now since we want this in asap for FL testing - and this minor change can be done in the future.

also functionally, nothing changes from a GenAI standpoint.

src/ort_genai_c.h
sayanshaw24: yep, documenting the info there in the latest commit!

test/c_api_tests.cpp
```diff
@@ -65,6 +65,14 @@ TEST(CAPITests, TokenizerCAPI) {
   auto model = OgaModel::Create(*config);
   auto tokenizer = OgaTokenizer::Create(*model);
 
+  // Update tokenizer options
```
baijumeswani, line 122: Could we instead create a single test that only tests updating the tokenizer options? We could also check output of the encoded string / decoded string?

src/models/model.cpp
```diff
@@ -266,9 +266,15 @@ std::unique_ptr<TokenizerStream> Tokenizer::CreateStream() const {
   return std::make_unique<TokenizerStream>(*this);
 }
 
+void Tokenizer::UpdateOptions(const char* const* keys, const char* const* values, size_t num_options) const {
```
```suggestion
void Tokenizer::UpdateOptions(const char* const* keys, const char* const* values, size_t num_options) {
```

src/models/model.cpp
```diff
@@ -266,9 +266,15 @@ std::unique_ptr<TokenizerStream> Tokenizer::CreateStream() const {
...
+  // Tap into ORT Extensions API
+  auto* raw = static_cast<const OrtxObject*>(tokenizer_);
+  CheckResult(OrtxUpdateTokenizerOptions(const_cast<OrtxTokenizer*>(reinterpret_cast<const OrtxTokenizer*>(raw)), const_cast<const char**>(keys), const_cast<const char**>(values), num_options));
+}
```
```suggestion
  CheckResult(OrtxUpdateTokenizerOptions(tokenizer_, const_cast<const char**>(keys), const_cast<const char**>(values), num_options));
}
```

src/models/model.h
```diff
@@ -85,6 +85,7 @@ struct Tokenizer : std::enable_shared_from_this<Tokenizer>, LeakChecked<Tokenize
 
   std::unique_ptr<TokenizerStream> CreateStream() const;
 
+  void UpdateOptions(const char* const* keys, const char* const* values, size_t num_options) const;
```
baijumeswani:
```suggestion
  void UpdateOptions(const char* const* keys, const char* const* values, size_t num_options);
```

src/python/python.cpp
```diff
@@ -380,11 +380,24 @@ PYBIND11_MODULE(onnxruntime_genai, m) {
...
+        std::vector<const char*> keys;
+        std::vector<const char*> values;
+        keys.reserve(kwargs.size());
+        values.reserve(kwargs.size());
+
+        for (auto& item : kwargs) {
+          const std::string key = pybind11::str(item.first);
+          const std::string value = pybind11::str(item.second);
+          keys.push_back(key.c_str());
+          values.push_back(value.c_str());
```
baijumeswani: The underlying c_str() of the string will no longer be valid after the for loop. It will result in an error.

src/models/model.cpp
```diff
@@ -266,9 +266,15 @@ std::unique_ptr<TokenizerStream> Tokenizer::CreateStream() const {
   return std::make_unique<TokenizerStream>(*this);
 }
 
+void Tokenizer::UpdateOptions(const char* const* keys, const char* const* values, size_t num_options) const {
+  // Tap into ORT Extensions API
+  auto* raw = static_cast<const OrtxObject*>(tokenizer_);
+  CheckResult(OrtxUpdateTokenizerOptions(const_cast<OrtxTokenizer*>(reinterpret_cast<const OrtxTokenizer*>(raw)), const_cast<const char**>(keys), const_cast<const char**>(values), num_options));
```
baijumeswani: Also consider updating the tokenizer api to accept const char* const* for keys and values (or const char* const[]).

src/models/model.cpp
sayanshaw24: tried this in an older commit already - it fails because `OrtxUpdateTokenizerOptions` in ORT Extensions expects a non const tokenizer (which is correct as we are modifying it), but in GenAI here `tokenizer_` is const.

baijumeswani: The constness comes from the method being defined as const. The tokenizer itself is not const. It should work with the suggested changes.

sayanshaw24: yeah (was going to say this in response to the next comment) i did miss removing the const in the method header before - so trying it with both right now.

src/models/model.cpp
sayanshaw24: Yeah, for sake of expediting Kunal's FL testing, we can keep it as is for now. But long term, I will update this in Extensions (along with other tool calling changes planned) and we can update the Ext commit.

src/python/python.cpp
sayanshaw24: good catch, need to store them. will resolve in latest commit along with other comments.

test/c_api_tests.cpp
sayanshaw24: So previously, `Encode` set `add_special_tokens = false` by default using `OrtxTokenizeWithOptions`,
and `Decode` set `skip_special_tokens = true` by default using `OrtxDetokenize1DWithOptions`.

Therefore, we need to update these regardless for all the tests.

Encoder/decoder output cannot really be tested well without more model files accessible (hence the conversation with Nat in the chat regarding just having API testing) - but even outputs are well tested, with a variety of input combinations, (not just false/true for add_special_tokens/skip_special_tokens), in Extensions here: https://github.com/microsoft/onnxruntime-extensions/pull/998 since we have tokenzizer files checked into the repo.

src/ort_genai.h
```diff
@@ -305,6 +305,10 @@ struct OgaTokenizer : OgaAbstract {
     return std::unique_ptr<OgaTokenizer>(p);
   }
 
+  void UpdateOptions(const char* const* keys, const char* const* values, size_t num_options) const {
+    OgaCheckResult(OgaUpdateTokenizerOptions(const_cast<OgaTokenizer*>(this), keys, values, num_options));
```
baijumeswani: Does this need a const_cast?

sayanshaw24: it does currently (failed without), but let me update the function header similarly to UpdateTokenizerOptions and try it without it right now.

test/c_api_tests.cpp
baijumeswani: Does the new api change the default behavior? Does it break existing tests if we don't do this?

sayanshaw24: Updated tests and added a new one - kept Encoding and Decoding in the test still to ensure default behavior.

src/ort_genai_c.h
```diff
@@ -572,6 +572,43 @@ OGA_EXPORT OgaResult* OGA_API_CALL OgaCreateMultiModalProcessor(const OgaModel*
...
+ *
+ * - `add_special_tokens`
+ *   - Purpose: Controls whether to add special tokens (e.g., BOS/EOS) during tokenization.
+ *   - Values: `"true"` / `"false"` or `"1"` / `"0"`.
+ *   - Default: `"true"`.
+ *
+ * - `skip_special_tokens`
+ *   - Purpose: Controls whether to remove special tokens during detokenization.
+ *   - Values: `"true"` / `"false"` or `"1"` / `"0"`.
+ *   - Default: `"true"`.
```
baijumeswani: Are the defaults correct?

sayanshaw24: nope, missed updating the comment, just fixed!

test/c_api_tests.cpp
```diff
@@ -124,7 +124,7 @@ TEST(CAPITests, TokenizerUpdateOptions) {
   {
     const char* keys[] = {"add_special_tokens", "skip_special_tokens"};
     const char* values[] = {"false", "true"};
-    OgaResult* result = OgaUpdateTokenizerOptions(tokenizer, keys, values, 2);
+    OgaResult* result = OgaUpdateTokenizerOptions(tokenizer.get(), keys, values, 2);
```
baijumeswani:
```suggestion
    tokenizer->UpdateOptions(keys, values, 2);
```

PR #1788, opened 2025-09-22
Add enable_webgpu_graph in extra_options

qjia7:
This PR adds the enable_webgpu_graph in extra_options. It optimizes attention mask subgraph by removing all unnecessary nodes running on cpu ep. And directly pass position_ids into RotaryEmbedding to further removing the unnecessary nodes running on cpu. With these changes, it makes sure all ops in phi4 can run on webgpu ep and no memcpy ops and control flow ops. This is preparation to support graph capture.

With the changes, it also improves the inference time on high power gpus since it reduces the cpu time.


src/python/py/models/builder.py
```diff
@@ -75,13 +76,25 @@ def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
 
         # EP-specific variables
         self.ep = ep
+
+        # Validate enable_webgpu_graph option
+        if extra_options.get("enable_webgpu_graph", False) and self.ep != "webgpu":
```
kunal-vaishnavi: Validation of the extra options can be done in `check_extra_options`.

https://github.com/microsoft/onnxruntime-genai/blob/83a6e5d2fc97874a7d3b274b6dd00d3ec02cc42e/src/python/py/models/builder.py#L4294-L4297

src/python/py/models/builder.py
```diff
@@ -3205,41 +3223,87 @@ def make_common_mask_reformat_subgraph(self, basename, root_input, unsqueeze_for
...
-        gather_name = f"{attn_mask_basename}/Gather"
-        gather_inputs = [f"{shape_name}/output_0", "/model/constants/INT64/1"]
-        self.make_gather(gather_name, gather_inputs, dtype=ir.DataType.INT64, shape=[], axis=0)
-        cast_2_name = f"{attn_mask_basename}/Gather/Cast"
-        self.make_cast(cast_2_name, f"{gather_name}/output_0", dtype=ir.DataType.INT32, shape=None)
-
-        self.mask_attrs["seqlens_k"] = cast_1_name
-        self.mask_attrs["total_seq_len"] = cast_2_name
+        if self.extra_options.get("enable_webgpu_graph", False):
+            # WebGPU graph mode: Remove right path, calculate total_seq_len from seqlens_k
```
kunal-vaishnavi: What is the benefit of having one path? Can all EPs use this one path instead?

src/python/py/models/builder.py
```diff
@@ -3279,7 +3343,18 @@ def make_attention_mask_reformatting_for_sparse_attn(self):
         self.mask_attrs["total_seq_len"] = cast_2_name
 
     def make_position_ids_reformatting(self):
-        # Make nodes for the position ids reformatting subgraph
```
kunal-vaishnavi, line 3282: The logic used in the position ids reformatting subgraph was originally for reshaping the `position_ids` input tensor to a 2D tensor for the `RotaryEmbedding` contrib op. Since the position ids are now always a 2D tensor with int64 values and its shape always matches the shape of the input ids, I think we can remove `make_position_ids_reformatting` altogether and always provide the `position_ids` input directly to the `RotaryEmbedding` op.

src/python/py/models/builder.py
qjia7: For the two path, the right path gets the total sequence length from shape, which always runs on cpu not webgpu and results a MemcpyFromHost inserted and fails the graph capture path. That's why I remove the right path and calculate total_seq_len from seqlens_k. Previously the two path are parallel. Now I make it to sequential to support graph capture. If you think the impact is minimal and can be ignored, we can make all EPs use this one path.

src/python/py/models/builder.py
qjia7: Done

src/python/py/models/builder.py
```diff
@@ -2876,10 +2884,9 @@ def has_final_norm(self, module, orig_model):
 
     def make_preprocessing_nodes(self):
         self.make_attention_mask_reformatting()
-        # TODO: add make_position_ids_reformatting() here
 
     def make_attention_mask_reformatting(self):
-        if self.extra_options.get("enable_cuda_graph", False) or self.ep == "dml":
+        if self.extra_options.get("enable_cuda_graph", False) or self.extra_options.get("enable_webgpu_graph", False) or self.ep == "dml":
             # ORT does not allow nodes to be placed on mulitple execution providers
             # with cuda graph enabled. We've only verified it works with GQA and with
```
```suggestion
            # with graph capture enabled. We've only verified it works with GQA and with
```

src/python/py/models/builder.py
kunal-vaishnavi: Let's keep both paths for now and we can revisit this later. For the sequential version, we should enable it for all models that use graph capture and not just WebGPU.

src/python/py/models/builder.py
```diff
@@ -4325,8 +4325,18 @@ def check_extra_options(kv_pairs):
         # 'include_hidden_states' is for when 'hidden_states' are outputted and 'logits' are outputted
         raise ValueError("Both 'exclude_lm_head' and 'include_hidden_states' cannot be used together. Please use only one of them at once.")
 
+    # Validate enable_webgpu_graph option
+    if kv_pairs.get("enable_webgpu_graph", False) and execution_provider != "webgpu":
+        warnings.warn(
```
kunal-vaishnavi: All warning messages are currently printed with a simple `print("WARNING: text")` call to keep logging simple.

https://github.com/microsoft/onnxruntime-genai/blob/a502ce31889ec33435935cfa30dccdee77314279/src/python/py/models/builder.py#L2450-L2451

src/python/py/models/builder.py
```diff
@@ -3466,33 +3493,6 @@ def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
         super().__init__(config, io_dtype, onnx_dtype, ep, cache_dir, extra_options)
         self.make_rotary_embedding_multi_cache()
 
-    def make_position_ids_reformatting(self):
-        if self.ep != "dml":
-            position_ids_input_to_rotemb = super().make_position_ids_reformatting()
```
kunal-vaishnavi, line 3471: Upon further consideration, let's bring back `make_position_ids_reformatting` in the base `Model` class but change its implementation to the below. To handle LongRoPE in the Phi models, some EPs combine the two sets of caches in the rotary embeddings together. The cache merging changes the calculation for the position ids that need to be passed to the `RotaryEmbedding` op.

This method is called for EPs that don't support If nodes. We can remove the early if condition and call `make_position_ids_reformatting` in this class for those EPs instead of calling it in the `MistralModel` class.

src/python/py/models/builder.py
```diff
@@ -3325,10 +3342,20 @@ def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
 class MistralModel(Model):
     def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
         super().__init__(config, io_dtype, onnx_dtype, ep, cache_dir, extra_options)
-        self.position_ids_name = f"{self.make_position_ids_reformatting()}/output_0" if not self.attention_attrs["use_rope_in_attn"] else "position_ids"
+
+        # Set position_ids_name based on whether position_ids is available as an input
+        if "position_ids" in self.input_names:
+            self.position_ids_name = "position_ids"
+        else:
```
kunal-vaishnavi: Let's move this logic into the `Phi3MiniLongRoPE` class since the position ids reformatting affects that model and not the `MistralModel` class anymore. Then the `MistralModel` class can always call `super().make_attention(layer_id, attention, root_input, **kwargs)` and not need `self.position_ids_name`.

src/python/py/models/builder.py
```diff
@@ -1461,8 +1464,13 @@ def make_rotary_embedding_multi_cache(self, **kwargs):
         self.rope_attrs["save_caches"] = False
         cos_cache_small, sin_cache_small = self.make_rotary_embedding_caches(cos_cache_name=cos_cache_small_name, sin_cache_name=sin_cache_small_name)
 
-        if self.ep in ["dml", "trt-rtx"]:
-            # Concat small and large cos/sin caches for DML and TRT-RTX EPs
+        # Determine which EPs don't support the If operator
+        eps_without_if_support = ["dml", "trt-rtx"]
```
```suggestion
        self.eps_without_if_support = ["dml", "trt-rtx"]
```

src/python/py/models/builder.py
```diff
@@ -1461,8 +1464,13 @@ def make_rotary_embedding_multi_cache(self, **kwargs):
...
+        if self.extra_options.get("enable_webgpu_graph", False):
+            eps_without_if_support.append("webgpu")
```
```suggestion
            self.eps_without_if_support.append("webgpu")
```

src/python/py/models/builder.py
```diff
@@ -75,13 +76,15 @@ def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
...
         self.ep = ep
+
         self.ep_attrs = {
             "cpu": {},
             "cuda": {
                 "enable_cuda_graph": "1" if extra_options.get("enable_cuda_graph", False) else "0",              # "1" if the model is able to enable cuda graph, "0" otherwise
                 "enable_skip_layer_norm_strict_mode": "1"
             },
             "dml": {},
+            # TODO: Enable graph capture for webgpu once supported both in onnxruntime-genai and onnxruntime.
```
kunal-vaishnavi, line 85: What is missing to support graph capture?

src/python/py/models/builder.py
qjia7: I keep `make_position_ids_reformatting` unchanged in `Phi3MiniLongRoPEModel` since this is the only place to handle LongRoPE. And in the base Model, just simply return "position_ids". 

>This method is called for EPs that don't support If nodes. 

From the current code, it seems only for dml ep. `trt-rtx` didn't go to this path. With this PR, we have DML, TRT-RTX, and WebGPU (when graph enabled) EPs without if op. So I keep `make_position_ids_reformatting` unchanged in this model.

src/python/py/models/builder.py
qjia7: ORT part: https://github.com/microsoft/onnxruntime/pull/25868
ORT-genai part: https://github.com/microsoft/onnxruntime-genai/pull/1720

src/python/py/models/builder.py
qjia7: >we should enable it for all models that use graph capture and not just WebGPU

I added the `dml` but the dml CI failed. So I revert the changes back https://github.com/microsoft/onnxruntime-genai/pull/1788/commits/5c6b84e38bdea431fab696ac7a2a6dcfa7dd9dfb. I think the new changes may not compatible with existed dml path. So I prefer to keep the changes to webgpu.

src/python/py/models/builder.py
```diff
@@ -75,13 +75,15 @@ def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
 
         # EP-specific variables
         self.ep = ep
+
```
```suggestion
```

src/python/py/models/builder.py
```diff
@@ -3205,41 +3211,87 @@ def make_common_mask_reformat_subgraph(self, basename, root_input, unsqueeze_for
...
-        gather_name = f"{attn_mask_basename}/Gather"
-        gather_inputs = [f"{shape_name}/output_0", "/model/constants/INT64/1"]
-        self.make_gather(gather_name, gather_inputs, dtype=ir.DataType.INT64, shape=[], axis=0)
-        cast_2_name = f"{attn_mask_basename}/Gather/Cast"
-        self.make_cast(cast_2_name, f"{gather_name}/output_0", dtype=ir.DataType.INT32, shape=None)
-
-        self.mask_attrs["seqlens_k"] = cast_1_name
-        self.mask_attrs["total_seq_len"] = cast_2_name
+        if self.extra_options.get("enable_webgpu_graph", False):
+            # WebGPU graph mode: Remove right path, calculate total_seq_len from seqlens_k
```
kunal-vaishnavi: Let's move the logic inside the `if` and `else` conditions into their own functions whose names describe their purpose (e.g. `make_attention_mask_graph_capture_reformatting_for_gqa` and `make_attention_mask_standard_reformatting_for_gqa` respectively).

src/python/py/models/builder.py
kunal-vaishnavi: The below logic shows that the cos and sin caches for the rotary embeddings will be combined.

https://github.com/microsoft/onnxruntime-genai/blob/5c6b84e38bdea431fab696ac7a2a6dcfa7dd9dfb/src/python/py/models/builder.py#L1471-L1480

The cache merging requires a change to how the position ids are computed since the `If` node is only created when the caches are not merged. I think we can change the condition to the following to fix this.

```python
def make_position_ids_reformatting(self):
    if self.ep not in self.eps_without_if_support:
        position_ids_input_to_rotemb = super().make_position_ids_reformatting()
        return position_ids_input_to_rotemb
```

qjia7: Done. But still skip Long RoPE scaling subgraph for WebGPU for now. The ops in subgroup for example, ReduceMax/GreaterOrEqual have inputs with type int64 are still running on cpu in webgpu ep.

qjia7: Make webgpu also support long RoPE by position_ids→ Cast(int32) → computation in int32 → Cast(int64). Other EPs graph: Direct computation in int64

src/python/py/models/builder.py
```diff
@@ -3325,10 +3346,9 @@ def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
 class MistralModel(Model):
     def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
         super().__init__(config, io_dtype, onnx_dtype, ep, cache_dir, extra_options)
-        self.position_ids_name = f"{self.make_position_ids_reformatting()}/output_0" if not self.attention_attrs["use_rope_in_attn"] else "position_ids"
 
     def make_attention(self, layer_id, attention, root_input, **kwargs):
-        super().make_attention(layer_id, attention, root_input, position_ids=self.position_ids_name, **kwargs)
+        super().make_attention(layer_id, attention, root_input, **kwargs)
```
kunal-vaishnavi: Since this method has the same signature as its parent, can we remove this method to override the parent implementation of `make_attention`?

src/python/py/models/builder.py
kunal-vaishnavi: Sounds good. Will int64 op support be improved on WebGPU?

src/python/py/models/builder.py
```diff
@@ -3466,32 +3486,86 @@ def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
...
             return position_ids_input_to_rotemb
 
+        # Make Long RoPE scaling subgraph to adjust position_ids for sequences beyond original context length
+        # For WebGPU: use int32 for all ops due to limited int64 support, then cast back to int64
+        # For other EPs: use native int64 throughout
+        #
+        # WebGPU graph:                              Other EPs graph:
+        #   position_ids                               position_ids
+        #        |                                          |
+        #   Cast (int32)                                ReduceMax (int64)
```
```suggestion
        #   Cast (int32)                               ReduceMax (int64)
```

src/python/py/models/builder.py
qjia7: For the long RoPE case, it can be further improved to delete the extra two `Cast` on webgpu since we know which ops with int64 need to be registered. But for the non graph capture case, ops with int64 for webgpu ep will run on cpu since we think they are processing shape related things which can provide better performance on cpu. So the problem is there is no a good way to tell us that which ops with int64 should run on gpu not cpu in normal case.

kunal-vaishnavi (approved):

PR #1791, opened 2025-09-24
Add TRT-RTX EP support, normalize alias, and force QDQ

anujj:
- Accept both "trt-rtx" and "NvTensorRtRtx" in --execution_provider, with
  runtime normalization to "trt-rtx" for internal handling.
- Automatically set extra_options["use_qdq"]=True when using TRT-RTX to
  ensure Q/DQ quantization path is enabled by default.

### DIFF
--- a/src/python/py/models/builder.py
+++ b/src/python/py/models/builder.py
@@ -4325,6 +4325,10 @@ def check_extra_options(kv_pairs):
         # 'include_hidden_states' is for when 'hidden_states' are outputted and 'logits' are outputted
         raise ValueError("Both 'exclude_lm_head' and 'include_hidden_states' cannot be used together. Please use only one of them at once.")
 
+    # force use_qdq for trt-rtx
+    if args.execution_provider == "trt-rtx":
+        kv_pairs["use_qdq"] = True
+
 
 def parse_extra_options(kv_items):
     """
@@ -4547,7 +4551,7 @@ def get_args():
         "-e",
         "--execution_provider",
         required=True,
-        choices=["cpu", "cuda", "dml", "webgpu", "trt-rtx"],
+        choices=["cpu", "cuda", "dml", "webgpu", "trt-rtx", "NvTensorRtRtx"],
         help="Execution provider to target with precision of model (e.g. FP16 CUDA, INT4 CPU, INT4 WebGPU)",
     )
 
@@ -4635,4 +4639,10 @@ def get_args():
 if __name__ == '__main__':
     args = get_args()
     extra_options = parse_extra_options(args.extra_options)
-    create_model(args.model_name, args.input, args.output, args.precision, args.execution_provider, args.cache_dir, **extra_options)
+
+    # Normalize execution provider name: both "NvTensorRtRtx" and "trt-rtx" should be treated as "trt-rtx" internally
+    execution_provider = args.execution_provider
+    if execution_provider.lower() == "nvtensorrtrtx":
+        execution_provider = "trt-rtx"
+
+    create_model(args.model_name, args.input, args.output, args.precision, execution_provider, args.cache_dir, **extra_options)


anujj:
@gaugarg-nv @BLSharda @kunal-vaishnavi @baijumeswani

anujj:
Updated PR to keep user facing name as NvTensorRtRtx

src/python/py/models/builder.py
```diff
@@ -4634,5 +4638,7 @@ def get_args():
 
 if __name__ == '__main__':
     args = get_args()
+    if args.execution_provider == "NvTensorRtRtx":
```
kunal-vaishnavi: We should handle this logic inside `create_model` so that users who import `create_model` can leverage this (e.g. Olive).

anujj: i am forcing use_qdq in check_extra_options() with is called before create_model

    # force use_qdq for trt-rtx
    if args.execution_provider == "trt-rtx":
        kv_pairs["use_qdq"] = True
        
i had to switch the condition to `if args.execution_provider == "NvtensorRtRtx":`, which should be okey i guess ?

src/python/py/models/builder.py
```diff
@@ -4325,6 +4325,10 @@ def check_extra_options(kv_pairs):
         # 'include_hidden_states' is for when 'hidden_states' are outputted and 'logits' are outputted
         raise ValueError("Both 'exclude_lm_head' and 'include_hidden_states' cannot be used together. Please use only one of them at once.")
 
+    # force use_qdq for trt-rtx
+    if args.execution_provider == "trt-rtx":
+        kv_pairs["use_qdq"] = True
```
kunal-vaishnavi: In the original implementation of this before it was removed, there was a comment which said that QDQ was used because opset 21 was desired. The default opset is now 21 but it seems that creating a QDQ model is the true intention.

I also remember there was an issue in Olive which originated from forcing `use_qdq` here. Could there be a future scenario where creating a TRT-RTX model without QDQ is desired?

anujj: we dont support matmulnbit in trt-rtx that's why it is always required, without this int4 path will break.

src/python/py/models/builder.py
kunal-vaishnavi: This is a separate issue from what you are mentioning. This check to change the EP name from `NvTensorRtRtx` to `trt-rtx` will not happen for users who import `create_model`. Here is an example which will fail with your current PR changes.

```python
from onnxruntime_genai.models.builder import create_model

input_dir = ""
output_dir = "./phi-4-mini"
cache_dir = "./cache_dir"

model_name = "microsoft/Phi-4-mini-instruct"
precision = "int4"
ep = "NvTensorRtRtx"

extra_options = {
    "use_qdq": "true"
}

create_model(model_name, input_dir, output_dir, precision, ep, cache_dir, **extra_options)
```

In the above code, `create_model` will not replace `NvTensorRtRtx` with `trt-rtx`. This will cause the incorrect ONNX model to get created. After moving the replacement of the EP name inside `create_model`, the above code should work.

src/python/py/models/builder.py
kunal-vaishnavi: Will `MatMulNBits` be supported in the future?

anujj: No plan for that kunal

anujj: simplify the changes in the latest commit

gaugarg-nv:
@anujj we will also have to revert this PR in Olive: https://github.com/microsoft/Olive/pull/2169

anujj:
Raised a PR for that : https://github.com/microsoft/Olive/pull/2182

src/python/py/models/builder.py
```diff
@@ -4326,6 +4326,7 @@ def check_extra_options(kv_pairs):
         raise ValueError("Both 'exclude_lm_head' and 'include_hidden_states' cannot be used together. Please use only one of them at once.")
 
 
+
```
```suggestion
```

src/python/py/models/builder.py
```diff
@@ -4390,6 +4391,11 @@ def set_onnx_dtype(precision: str, extra_options: dict[str, Any]) -> ir.DataType
 
 @torch.no_grad
 def create_model(model_name, input_path, output_dir, precision, execution_provider, cache_dir, **extra_options):
+
```
```suggestion
```

PR #1795, opened 2025-09-26
Add run options to ONNX Runtime GenAI

kunal-vaishnavi:
### Description

This PR allows users to customize run options per ONNX model that runs in ONNX Runtime GenAI.

### Usage

The run options can be added as key-value pairs in a separate, optional section within the GenAI config.

```json
"session_options": {
    "": "",
    "provider_options": [
        {
            "cuda": {
                "enable_cuda_graph": "1",
                "enable_skip_layer_norm_strict_mode": "1",
                "arena_extend_strategy": "kSameAsRequested",
            }
        }
    ]
},
"run_options": {
    "memory.enable_memory_arena_shrinkage": "cpu:0;gpu:0"
},
```

You can also have separate run options per ONNX model within the GenAI config.

```json
"decoder": {
    "session_options": {
        "provider_options": []
    },
    "run_options": {
        "memory.enable_memory_arena_shrinkage": "cpu:0;gpu:0"
    }
},
"vision": {
    "run_options": {
        "memory.enable_memory_arena_shrinkage": "cpu:0;gpu:0"
    },
    "inputs": {
        "pixel_values": "pixel_values",
        "attention_mask": "image_attention_mask",
        "image_sizes": "image_sizes"
    },
    "outputs": {
        "image_features": "image_features"
    }
},
"speech": {
    "run_options": {
        "memory.enable_memory_arena_shrinkage": "cpu:0;gpu:0"
    },
    "inputs": {
        "audio_embeds": "audio_embeds",
        "attention_mask": "audio_attention_mask",
        "audio_sizes": "audio_sizes",
        "audio_projection_mode": "audio_projection_mode"
    },
    "outputs": {
        "audio_features": "audio_features"
    }
},
"embedding": {
    "run_options": {
        "memory.enable_memory_arena_shrinkage": "cpu:0;gpu:0"
    },
    "inputs": {
        "input_ids": "input_ids",
        "image_features": "image_features",
        "audio_features": "audio_features"
    },
    "outputs": {
        "inputs_embeds": "inputs_embeds"
    }
},
```

#### Session Options

For a full list, please see the list of keys available [here](https://onnxruntime.ai/docs/api/python/api_summary.html#onnxruntime.SessionOptions).

#### Provider Options

For a full list, please see your target execution provider's page inside the [ONNX Runtime docs](https://onnxruntime.ai/docs/execution-providers/).

#### Run Options

For a full list, please see the list of keys available [here](https://github.com/microsoft/onnxruntime/blame/main/include/onnxruntime/core/session/onnxruntime_run_options_config_keys.h).

### Motivation and Context

This PR allows users to use run options such as `memory.enable_memory_arena_shrinkage` to reduce memory usage for memory-constrained environments.


src/config.h
```diff
@@ -89,18 +90,34 @@ struct Config {
...
-    bool use_env_allocators{};
+    std::optional<bool> use_env_allocators;
     std::vector<NamedString> config_entries;  // Entries go into OrtSessionOptions::AddConfigEntry
 
     std::vector<ProviderOptions> provider_options;
     std::vector<std::string> providers;  // List of providers to use at runtime, not persisted in the json currently
     std::optional<GraphOptimizationLevel> graph_optimization_level;
   };
 
+  struct RunOptions {
```
baijumeswani: It seems like all of these options are defined on the ort run-options by doing

```
run_options_->AddConfigEntry(key, value);
```

We could consider passing in values unchecked by defining this struct to contain an unordered_map of std::string -> std::string.


That way, we can avoid needing to keep adding new entries to support new run options as and when they become available.

src/models/model.cpp
```diff
@@ -155,14 +156,47 @@ void State::Run(OrtSession& session, bool graph_capture_this_run) {
...
+      run_options_->UnsetTerminate();
+    } else if (strcmp(value, "1") == 0) {
+      session_terminated_ = true;
+      run_options_->SetTerminate();
+    } else {
+      // Value not expected
+      throw std::runtime_error(std::string("terminate_session key value unexpected: ") + value);
+    }
+  }
+  run_options_->AddConfigEntry(key, value);
```
baijumeswani, line 173: It appears that this line of code will be executed even for `terminate_session` where it shouldn't.

src/models/model.cpp
```diff
@@ -155,14 +156,47 @@ void State::Run(OrtSession& session, bool graph_capture_this_run) {
...
+  }
+  if (config_run_options.qnn_rpc_control_latency.has_value()) {
+    run_options_->AddConfigEntry("qnn.rpc_control_latency", config_run_options.qnn_rpc_control_latency.value().c_str());
+  }
+  if (config_run_options.qnn_lora_config.has_value()) {
+    run_options_->AddConfigEntry("qnn.lora_config", config_run_options.qnn_lora_config.value().c_str());
+  }
+  if (config_run_options.nv_profile_index.has_value()) {
+    run_options_->AddConfigEntry("nv_profile_index", config_run_options.nv_profile_index.value().c_str());
+  }
```
baijumeswani: If we define the run-options map as described in the comment in the config.h file, we can avoid needing this parsing logic.

src/config.h
kunal-vaishnavi: I was following the pattern introduced with the session options. Should we change both then to pass unchecked values?

src/models/model.cpp
kunal-vaishnavi: I was following the pattern introduced with the session options. Should we change both then to pass unchecked values?

src/config.h
baijumeswani: The way we set it up was that for session options (that require explicit API calls), we check the names. However, the provider options within the session options are unchecked: https://github.com/microsoft/onnxruntime-genai/blob/a34c09845110a0471c0c6ede05dfa5377069e0bd/src/config.h#L73-L77

We can try to keep the ort runoptions unchecked if all of them are set with the same `AddConfigEntry(key, value);` function.

src/models/model.cpp
kunal-vaishnavi: Fixed it. There was originally an early return but it must have gotten removed by mistake.

src/config.h
kunal-vaishnavi: [Many](https://github.com/microsoft/onnxruntime/blob/main/include/onnxruntime/core/session/onnxruntime_session_options_config_keys.h) of the session options don't require explicit API calls either. Most are ultimately just wrappers around `AddConfigEntry` as well.

https://github.com/microsoft/onnxruntime-genai/blob/a34c09845110a0471c0c6ede05dfa5377069e0bd/src/models/onnxruntime_inline.h#L575-L609

There are some session options that require explicit API calls.

baijumeswani: That is true. We can define a mix. If the run-option requires explicit API call or if the type of the run option value is not a string, then we can explicitly define it in the runoptions and add the parsing logic. For all others, we can keep them as is. We can do this for the session options in this PR (or another).

src/models/decoder_only.cpp
```diff
@@ -60,6 +63,9 @@ DeviceSpan<float> DecoderOnly_State::RunWithChunking(int total_length, DeviceSpa
 
     // Process this chunk - fills KV cache progressively
     UpdateInputsOutputs(chunk_tokens, next_indices, length);
+    if (model_.config_->model.decoder.run_options.has_value()) {
+      State::SetRunOptions(model_.config_->model.decoder.run_options.value());
+    }
```
baijumeswani: Can we make this call only once in the main run call?

kunal-vaishnavi: Fixed

baijumeswani:
Looks good. One improvement we could make is that we could just set the run options once at session creation. Since run options are changing mid generation. But we can get this in another PR.

baijumeswani (approved):

PR #1803, opened 2025-10-06
Enable guidance ff tokens for faster inference

JC1DA:
### What changed
Add params.set_guidance_ff_tokens(enabled) function to enable fast-forward tokens in llguidance library.

When using constrained grammar, llguidance can `suggest` the next multiple tokens because those tokens are the only valid tokens in the next few steps (e.g., `{"name": ` in json constraint). 

So instead generating one token per generation, we can generate more than one per iter, hence improving tokens/s metric.

### How to use
```
    guidance_type = "lark_grammar"
    guidance_input = f"""start: %json {json.dumps(schema)}\n"""
    params.set_search_options(**search_options)
    params.set_guidance(guidance_type, guidance_input) # set guidance
    params.set_guidance_ff_tokens(True)  # enable guidance feed-forward tokens
```

### DIFF
--- a/src/ort_genai.h
+++ b/src/ort_genai.h
@@ -400,6 +400,10 @@ struct OgaGeneratorParams : OgaAbstract {
     OgaCheckResult(OgaGeneratorParamsSetGuidance(this, type, data));
   }
 
+  void SetGuidanceFFTokens(bool enabled) {
+    OgaCheckResult(OgaGeneratorParamsSetGuidanceFFTokens(this, enabled));
+  }
+
   static void operator delete(void* p) { OgaDestroyGeneratorParams(reinterpret_cast<OgaGeneratorParams*>(p)); }
 };
 
--- a/src/ort_genai_c.h
+++ b/src/ort_genai_c.h
@@ -425,6 +425,14 @@ OGA_EXPORT OgaResult* OGA_API_CALL OgaGeneratorParamsTryGraphCaptureWithMaxBatch
 OGA_EXPORT OgaResult* OGA_API_CALL OgaGeneratorParamsSetGuidance(OgaGeneratorParams* params, const char* type, const char* data);
 
 /**
+ * \brief Sets whether to enable ff_tokens during constrained decoding for the Generator params
+ * \param[in] params The generator params to set the guidance on
+ * \param[in] enabled Whether to enable ff_tokens generation. Only valid when guidance type is set and batch_size is 1.
+ * \return OgaResult containing the error message if the setting of the guidance failed
+ */
+OGA_EXPORT OgaResult* OGA_API_CALL OgaGeneratorParamsSetGuidanceFFTokens(OgaGeneratorParams* params, bool enabled);
+
+/**
  * \brief Creates a generator from the given model and generator params.
  * \param[in] model The model to use for generation.
  * \param[in] params The parameters to use for generation.
--- a/src/python/python.cpp
+++ b/src/python/python.cpp
@@ -200,6 +200,10 @@ struct PyGeneratorParams {
     params_->SetGuidance(type.c_str(), data.c_str());
   }
 
+  void SetGuidanceFFTokens(bool enabled) {
+    params_->SetGuidanceFFTokens(enabled);
+  }
+
   std::vector<pybind11::object> refs_;  // References to data we want to ensure doesn't get garbage collected
 };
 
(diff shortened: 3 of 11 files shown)


src/ort_genai_c.h
```diff
@@ -420,9 +420,10 @@ OGA_EXPORT OgaResult* OGA_API_CALL OgaGeneratorParamsTryGraphCaptureWithMaxBatch
  * \param[in] params The generator params to set the guidance on
  * \param[in] type The type of the guidance. Currently, we support json_schema, regex and lark_grammar
  * \param[in] data The input string, which is the guidance data. Examples are present in test/test_models/grammars folder
+ * \param[in] enable_ff_tokens Whether to enable ff_tokens generation. Only valid when guidance type is set and batch_size is 1.
  * \return OgaResult containing the error message if the setting of the guidance failed
  */
-OGA_EXPORT OgaResult* OGA_API_CALL OgaGeneratorParamsSetGuidance(OgaGeneratorParams* params, const char* type, const char* data);
+OGA_EXPORT OgaResult* OGA_API_CALL OgaGeneratorParamsSetGuidance(OgaGeneratorParams* params, const char* type, const char* data, bool enable_ff_tokens = false);
```
baijumeswani: `C` does not support optional arguments or default values for function arguments. You can assume it is false by default internally, but the C API must not have the default value of the argument `bool enable_ff_tokens = false`

baijumeswani: Optional: If you want to name the argument as `ff_tokens` instead of `enable_ff_tokens`? If you feel it might be confusing, we can keep it `enable_ff_tokens`.

JC1DA: as `ff_tokens` often referred as an array of tokens, it may confuse some users (without reading the docs).
@kunal-vaishnavi what do you think?

kunal-vaishnavi: Either is fine with me. Let's use `enable_ff_tokens` then.

src/ort_genai.h
```diff
@@ -430,8 +430,8 @@ struct OgaGeneratorParams : OgaAbstract {
     printf("TryGraphCaptureWithMaxBatchSize is deprecated and will be removed in a future release\n");
   }
 
-  void SetGuidance(const char* type, const char* data) {
-    OgaCheckResult(OgaGeneratorParamsSetGuidance(this, type, data));
+  void SetGuidance(const char* type, const char* data, bool enable_ff_tokens = false) {
```
kunal-vaishnavi, line 433: Can we keep the API similar to the existing APIs which don't set default values for the boolean parameters?

https://github.com/microsoft/onnxruntime-genai/blob/f0c35a696f550d5fdc7b388df5fa2836d8923ad2/src/ort_genai.h#L364-L368

https://github.com/microsoft/onnxruntime-genai/blob/f0c35a696f550d5fdc7b388df5fa2836d8923ad2/src/ort_genai.h#L425-L427

https://github.com/microsoft/onnxruntime-genai/blob/f0c35a696f550d5fdc7b388df5fa2836d8923ad2/src/ort_genai.h#L849-L851

```suggestion
  void SetGuidance(const char* type, const char* data, bool enable_ff_tokens) {
```

src/python/python.cpp
```diff
@@ -196,8 +196,8 @@ struct PyGeneratorParams {
     std::cerr << "TryGraphCaptureWithMaxBatchSize is deprecated and will be removed in a future release" << std::endl;
   }
 
-  void SetGuidance(const std::string& type, const std::string& data) {
-    params_->SetGuidance(type.c_str(), data.c_str());
+  void SetGuidance(const std::string& type, const std::string& data, bool enable_ff_tokens = false) {
```
kunal-vaishnavi, line 199: Same question here

src/ort_genai_c.h
```diff
@@ -420,9 +420,10 @@ OGA_EXPORT OgaResult* OGA_API_CALL OgaGeneratorParamsTryGraphCaptureWithMaxBatch
  * \param[in] params The generator params to set the guidance on
  * \param[in] type The type of the guidance. Currently, we support json_schema, regex and lark_grammar
  * \param[in] data The input string, which is the guidance data. Examples are present in test/test_models/grammars folder
+ * \param[in] enable_ff_tokens Whether to enable ff_tokens generation. Only valid when guidance type is set and batch_size is 1.
```
kunal-vaishnavi: Can we explain what `ff_tokens` is here for users who may not be familiar?

```suggestion
 * \param[in] enable_ff_tokens Whether to enable ff_tokens generation. Only valid when guidance type is set, batch size is 1, and beam size is 1.
```

src/ort_genai.h
JC1DA: Should it be set default value so people can run existing code without any modification?
I'm fine with removing the default token, just something we should consider

kunal-vaishnavi: Let's keep the default value then.

kunal-vaishnavi (approved):

PR #1807, opened 2025-10-09
Fix Attention Mask Sizing for Sliding Window with NvTensorRtRtx EP

anujj:
## Problem

When using NvTensorRtRtx (TensorRT) execution provider with sliding window models, the following error occurs for models for ISL 4000 and OSL 200
Current shape: {1,4,4096,256}
Requested shape:{1,4,4200,256}

**Why this happens:**
- Sliding window is configured to 4096 tokens
- KV cache buffers are correctly allocated with size 4096 (from PR #1523)
- But attention_mask is sized to runtime `max_length` (e.g., 4200, ISL 4000, OSL 200)
- This creates a mismatch shapes 

## Solution

Clamp the static attention_mask size to the sliding window size for NvTensorRtRtx EP.

### DIFF
--- a/src/csharp/NativeMethods.cs
+++ b/src/csharp/NativeMethods.cs
@@ -225,6 +225,18 @@ namespace Microsoft.ML.OnnxRuntimeGenAI
                              UIntPtr /* size_t */ numOptions);
 
         [DllImport(NativeLib.DllName, CallingConvention = CallingConvention.Winapi)]
+        public static extern IntPtr /* OgaResult* */ OgaTokenizerGetBosTokenId(IntPtr /* const OgaTokenizer* */ tokenizer,
+                                                                               out int /* const int32_t* */ outStr);
+
+        [DllImport(NativeLib.DllName, CallingConvention = CallingConvention.Winapi)]
+        public static extern IntPtr /* OgaResult* */ OgaTokenizerGetEosTokenIds(IntPtr /* const OgaTokenizer* */ tokenizer,
+                                                                                IntPtr /* const OgaSequences* */ outStr);
+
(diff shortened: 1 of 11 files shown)


src/models/position_inputs.cpp
```diff
@@ -234,6 +234,16 @@ void DefaultPositionInputs::InitializeStaticMask(OrtValue& cpu_attention_mask) {
   // Create static tensor of size max_length and expanded by num_beams
   attention_mask_shape_[0] *= state_.params_->search.num_beams;
   attention_mask_shape_[1] = state_.params_->search.max_length;
+
+  // Clamp attention mask size to sliding window for NvTensorRtRtx EP
+  if (state_.model_.p_device_->GetType() == DeviceType::NvTensorRtRtx &&
+      state_.model_.config_->model.decoder.sliding_window.has_value() &&
+      state_.model_.config_->model.decoder.sliding_window->window_size > 0) {
+    attention_mask_shape_[1] = std::min(attention_mask_shape_[1],
+                                        static_cast<int64_t>(state_.model_.config_->model.decoder.sliding_window->window_size));
```
gaugarg-nv: @anujj I don't understand why trt-rtx EP should have such a constraint on the attention mask. The attention mask can have a larger shape than the sliding window value.

The only constraint we should have is regarding the shape of KV cache tensors.

src/models/position_inputs.cpp
```diff
@@ -234,6 +234,16 @@ void DefaultPositionInputs::InitializeStaticMask(OrtValue& cpu_attention_mask) {
...
+    std::cout << "InitializeStaticMask: Clamped attention_mask to sliding window size: " << attention_mask_shape_[1] << std::endl;
```
gaugarg-nv: Please remove console prints

PR #1742, opened 2025-09-09
Generate pyd type info

chemwolf6922:
## Changes
* Generate type info in the wheel. This enables type hints and checks.
* Expose the enum OgaElementType which is used as a return type but not declared in python.

PR #1817, opened 2025-10-11
Add build time python type stub generation

chemwolf6922:
## Changes
* Generate type info in the wheel. This enables type hints and checks.
* Expose the enum OgaElementType which is used as a return type but not declared in python.
## Adjustments compare to https://github.com/microsoft/onnxruntime-genai/pull/1742
The following adjustments were made to avoid breaking the internal ado build pipelines:
* Use onnxruntime python package's included libraries instead of copying libraries to the same folder for stub generation.
* Use mypy instead of pybind11-stubgen for stub generation. Which can generate stub against the whole package. Enabling the use of the existing library loading scripts during stub generation to avoid loading wrong onnxruntime libraries in some build environment.

### DIFF
--- a/src/python/stubgen.py
+++ b/src/python/stubgen.py
@@ -6,6 +6,7 @@ from pathlib import Path
 from argparse import ArgumentParser
 import subprocess
 import sys
+import platform
 import shutil
 
 def install_package(package_name: str, version: str | None = None) -> bool:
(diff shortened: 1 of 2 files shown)


src/python/stubgen.py
```diff
@@ -0,0 +1,83 @@
...
+    fix_stubs(wheel_root)
+    clean_up_pycache(wheel_root)
+
+    if onnxruntime_installed_by_us:
+        try_uninstall_package('onnxruntime')
+    if mypy_installed_by_us:
+        try_uninstall_package('mypy')
+
+if __name__ == '__main__':
+    # Explicitly skip windows ARM64.
```
chemwolf6922, line 77: This is required to properly skip the stubgen for win arm64 builds. Please refer to the ado pipeline runs for details.

PR #1832, opened 2025-10-16
Layer-wise KV Cache Allocation for Models with Alternating Attention Patterns

anujj:
### Background

Gemma2 models use an alternating attention pattern:
- **Even layers (0, 2, 4, ...)**: Full attention (requires full context, e.g., 8K tokens)
- **Odd layers (1, 3, 5, ...)**: Sliding window attention (only needs 4K tokens)

PR #1523 applied a **uniform** sliding window size (4K) to all layers for NvTensorRtRtx EP. This was incorrect

### Approach

1. **Export layer-specific attention types** during model building
2. **Use different dimension names** in ONNX to satisfy TensorRT constraints:
   - `past_sequence_length_full` for full attention layers
   - `past_sequence_length_sliding` for sliding window layers
   - As the shapes for full attention and sliding attention will be diffrent, trt-rtx expects the diff names
3. **Implement per-layer KV cache allocation** in runtime based on attention type
4. **Allocate optimal memory** for each layer type: Below eg for Gemma 
   - Full attention: 8192 tokens
   - Sliding window: 4096 tokens
  
**Memory Savings Examples**:
### Gemma2
- **Pattern**: Alternating (every other layer: full, sliding, full, sliding, ...)
- **Memory**: 8K (full) vs 4K (sliding)
- **Savings**: ~25% reduction (13 full + 13 sliding vs 26 full)

## Benefits for Multiple Models

This approach enables significant memory optimization for various model architectures with mixed attention patterns:

### Gemma2
- **Pattern**: Alternating (every other layer: full, sliding, full, sliding, ...)
- **Memory**: 8K (full) vs 4K (sliding)
- **Savings**: ~25% reduction (13 full + 13 sliding vs 26 full)

### Gemma3-4B
- **Pattern**: Every 6th layer uses global attention (5 sliding + 1 full)
- **Layers**: 34 layers (29 sliding + 5 full)
- **Memory**: 128K (full) vs 1K (sliding)
- **Savings**: **~85% reduction** - huge improvement in KV memory requirement

### GPT-OSS
- **Pattern**: Alternating global and sliding windows
- **Memory**: 131K (full) vs 128 (sliding)
- **Savings**: **~99.9% reduction** for sliding layers - huge improvement in KV memory requirement
  - Enables support for extremely long context windows with manageable memory footprint

### DIFF
--- a/src/models/kv_cache.cpp
+++ b/src/models/kv_cache.cpp
@@ -6,7 +6,7 @@
 #include "kv_cache.h"
 #include "windowed_kv_cache.h"
 #include "../openvino/interface.h"
-#include <unordered_set>
+#include <algorithm>
 
 namespace Generators {
 
@@ -176,33 +176,25 @@ DefaultKeyValueCache::DefaultKeyValueCache(State& state)
   }
 
   // Set the size after empty_past_ has been created with 0 for this field
-  if (state.model_.p_device_->GetType() == DeviceType::NvTensorRtRtx &&
-      model_.config_->model.decoder.sliding_window.has_value() &&
+  if (model_.config_->model.decoder.sliding_window.has_value() &&
       model_.config_->model.decoder.sliding_window->window_size > 0) {
-    int sliding_window_size = model_.config_->model.decoder.sliding_window->window_size;
-    int max_length = state_.params_->search.max_length;
+    const int sliding_window_size = model_.config_->model.decoder.sliding_window->window_size;
+    const int max_length = state_.params_->search.max_length;
 
     // Check if we need per-layer allocation for models with alternating attention patterns
     if (!model_.config_->model.decoder.sliding_window->layers.empty()) {
       // Use per-layer allocation based on sliding window layer indices
-      use_layer_types_ = true;
       layer_shapes_.resize(layer_count_);
 
-      // Create a set of sliding window layer indices for fast lookup
-      std::unordered_set<int> sliding_layers(
-          model_.config_->model.decoder.sliding_window->layers.begin(),
-          model_.config_->model.decoder.sliding_window->layers.end());
-
+      // Initialize all layers with base shape and max_length
       for (int layer_idx = 0; layer_idx < layer_count_; ++layer_idx) {
-        layer_shapes_[layer_idx] = shape_;  // Copy base shape
+        layer_shapes_[layer_idx] = shape_;
+        layer_shapes_[layer_idx][2] = max_length;
+      }
 
-        if (sliding_layers.count(layer_idx) > 0) {
-          // Sliding window layer
-          layer_shapes_[layer_idx][2] = std::min(max_length, sliding_window_size);
-        } else {
-          // Full attention layer
-          layer_shapes_[layer_idx][2] = max_length;
-        }
+      // Update sliding window layers with constrained cache size
+      for (int layer_idx : model_.config_->model.decoder.sliding_window->layers) {
+        layer_shapes_[layer_idx][2] = std::min(max_length, sliding_window_size);
       }
       // Set shape_[2] to max of all layer shapes for RewindTo bounds checking
       shape_[2] = max_length;
@@ -215,31 +207,20 @@ DefaultKeyValueCache::DefaultKeyValueCache(State& state)
   }
 
   try {
-    if (use_layer_types_) {
-      // Allocate per-layer with different shapes
-      for (int layer_idx = 0; layer_idx < layer_count_; ++layer_idx) {
-        // Key tensor
-        presents_.push_back(OrtValue::CreateTensor(Allocator(), layer_shapes_[layer_idx], type_));
-        if (Device().GetType() != DeviceType::WEBGPU) {
-          ByteWrapTensor(Device(), *presents_.back()).Zero();
-        }
-
-        // Value tensor
-        presents_.push_back(OrtValue::CreateTensor(Allocator(), layer_shapes_[layer_idx], type_));
-        if (Device().GetType() != DeviceType::WEBGPU) {
-          ByteWrapTensor(Device(), *presents_.back()).Zero();
-        }
+    // Allocate KV cache tensors - 2 per layer (key and value)
+    // For per-layer shapes: alternates between key and value for each layer
+    // For uniform shape: all tensors use the same shape
+    for (int i = 0; i < layer_count_ * 2; ++i) {
+      std::array<int64_t, 4> tensor_shape = shape_;
+      if (!layer_shapes_.empty()) {
+        // Per-layer allocation: use layer-specific shape
+        // i/2 gives us the layer index since we have 2 tensors per layer
+        tensor_shape = layer_shapes_[i / 2];
       }
-    } else {
-      // Uniform allocation (existing behavior)
-      for (int i = 0; i < layer_count_ * 2; ++i) {
-        presents_.push_back(OrtValue::CreateTensor(Allocator(), shape_, type_));
 
-        // Zero the memory so we don't leak any data from the previous run
-        // WebGPU device has no Zero() implementation yet. Since this zeroing is optional we disable it for WebGPU for now
-        if (Device().GetType() != DeviceType::WEBGPU) {
-          ByteWrapTensor(Device(), *presents_.back()).Zero();
-        }
+      presents_.push_back(OrtValue::CreateTensor(Allocator(), tensor_shape, type_));
+      if (Device().GetType() != DeviceType::WEBGPU) {
+        ByteWrapTensor(Device(), *presents_.back()).Zero();
       }
     }
   } catch (const Ort::Exception&) {
@@ -288,11 +269,11 @@ void DefaultKeyValueCache::Update(DeviceSpan<int32_t> beam_indices, int total_le
     }
   }
 
-  if (use_layer_types_) {
+  if (!layer_shapes_.empty()) {
     // Update per-layer shapes based on total_length, but respect max allocations
     for (int layer_idx = 0; layer_idx < layer_count_; ++layer_idx) {
-      int max_cache_length = static_cast<int>(layer_shapes_[layer_idx][2]);
-      int actual_length = std::min(total_length, max_cache_length);
+      const int max_cache_length = static_cast<int>(layer_shapes_[layer_idx][2]);
+      const int actual_length = std::min(total_length, max_cache_length);
 
       std::array<int64_t, 4> current_shape = layer_shapes_[layer_idx];
       current_shape[2] = actual_length;
@@ -341,7 +322,7 @@ template <typename T>
 void DefaultKeyValueCache::RewindPastTensorsTo(size_t index) {
   assert(index > 0 && !past_present_share_buffer_);
 
-  if (use_layer_types_) {
+  if (!layer_shapes_.empty()) {
     // Handle per-layer shapes
     // First validate that index doesn't exceed the global max_length
     int max_length = static_cast<int>(shape_[2]);  // Set to max_length in constructor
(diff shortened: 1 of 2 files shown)


anujj:
@gaugarg-nv for review

kunal-vaishnavi:
For the model builder changes, there is a simpler approach that can be used.

```py
def make_key_value_cache_shape(self, layer_id, shape):
    if self.ep == "trt-rtx" and hasattr(self, "is_local") and self.is_local(layer_id):
        return [shape[0], shape[1], shape[2].replace("sequence", "sliding"), shape[3]]
    return shape
```

Then the `make_inputs_and_outputs` method can be modified to look as follows.

```py
def make_inputs_and_outputs(self):
    # Add model-specific inputs to list of model inputs
    inputs = self.model.graph.inputs
    for name in self.input_names:
        dtype = self.input_types[name]
        shape = self.input_shapes[name]
        inputs.append(self.make_value(name, dtype=dtype, shape=shape))

    # Add model-specific outputs to list of model outputs
    outputs = self.model.graph.outputs
    for name in self.output_names:
        dtype = self.output_types[name]
        shape = self.output_shapes[name]
        outputs.append(self.make_value(name, dtype=dtype, shape=shape))

    # Add KV cache to inputs and outputs
    for i in range(self.num_layers):
        # Add KV cache to inputs
        key_name = f"past_key_values.{i}.key"
        key_shape = self.make_key_value_cache_shape(i, self.input_shapes["past_key_values.key"])
        inputs.append(self.make_value(key_name, dtype=self.input_types["past_key_values.key"], shape=key_shape))
        value_name = f"past_key_values.{i}.value"
        value_shape = self.make_key_value_cache_shape(i, self.input_shapes["past_key_values.value"])
        inputs.append(self.make_value(value_name, dtype=self.input_types["past_key_values.value"], shape=value_shape))

        # Add KV cache to outputs
        key_name = f"present.{i}.key"
        key_shape = self.make_key_value_cache_shape(i, self.output_shapes["present.key"])
        outputs.append(self.make_value(key_name, dtype=self.output_types["present.key"], shape=key_shape))
        value_name = f"present.{i}.value"
        value_shape = self.make_key_value_cache_shape(i, self.output_shapes["present.value"])
        outputs.append(self.make_value(value_name, dtype=self.output_types["present.value"], shape=value_shape))
```

The list of layer types could be simplified to a list of layer indices per attention type. For example:

```json
"sliding_window": {
    "window_size": 4096,
    "slide_key_value_cache": false,
    "slide_inputs": false,
    "layers": [2, 4, 6, 8]
}
```

src/python/py/models/builder.py
```diff
@@ -474,6 +481,23 @@ def make_genai_config(self, model_name_or_path, extra_kwargs, out_dir):
...
 
+    def get_layer_types(self):
+        """
+        Returns a list of attention types for each layer.
+        Override in subclasses to provide layer-specific attention patterns.
+        Returns None for models with uniform attention across all layers.
+        """
+        return None
+
+    def make_kv_value_cache_shape(self, layer_id, shape):
```
```suggestion
    def make_key_value_cache_shape(self, layer_id, shape):
```

or

```suggestion
    def make_kv_cache_shape(self, layer_id, shape):
```

src/python/py/models/builder.py
```diff
@@ -464,6 +464,13 @@ def make_genai_config(self, model_name_or_path, extra_kwargs, out_dir):
 
         if self.ep == "trt-rtx" and self.window_size is not None and self.window_size > 0:
             genai_config["model"]["decoder"]["sliding_window"] = {"window_size": self.window_size, "slide_key_value_cache": False, "slide_inputs": False}
+            
+            # Add layer indices for sliding window layers if model has alternating attention patterns
```
kunal-vaishnavi: Only the TRT-RTX EP is entering the parent `if` condition. You can add the layer indices directly and remove `get_layer_types`.

```py
layer_idxs = [layer_id for layer_id in range(self.num_layers) if hasattr(self, "is_local") and self.is_local(layer_id)]
genai_config["model"]["decoder"]["sliding_window"]["layers"] = layer_idxs
```

I think it should be fine to always keep `layers` inside the GenAI config when `sliding_window` exists (even if the array is empty).

src/python/py/models/builder.py

src/python/py/models/builder.py
anujj: made the suggested changes

anujj:
@kunal-vaishnavi : I have addressed the review comments, if PR looks in good shape now , can we merge this change ?

src/models/kv_cache.cpp
```diff
@@ -175,23 +176,70 @@ DefaultKeyValueCache::DefaultKeyValueCache(State& state)
...
+
+      if (sliding_layers.count(layer_idx) > 0) {
+        // Sliding window layer
+        layer_shapes_[layer_idx][2] = std::min(max_length, sliding_window_size);
+      } else {
+        // Full attention layer
+        layer_shapes_[layer_idx][2] = max_length;
+      }
+    }
+  } else if (state.model_.p_device_->GetType() == DeviceType::NvTensorRtRtx &&
```
nenad1002: This is repetition of code

src/models/kv_cache.cpp
```diff
@@ -175,23 +176,70 @@ DefaultKeyValueCache::DefaultKeyValueCache(State& state)
...
+             model_.config_->model.decoder.sliding_window.has_value() &&
+             model_.config_->model.decoder.sliding_window->window_size > 0) {
+    // Uniform sliding window allocation (backward compatibility)
     shape_[2] = std::min(state_.params_->search.max_length,
```
nenad1002: Doesn't shape_[2] need to be set in both cases, otherwise RewindTo could throw an exception?

src/models/kv_cache.cpp
anujj: Addressed in latest commit

src/models/kv_cache.cpp
anujj: Addressed in latest commit

src/config.cpp
```diff
@@ -468,8 +479,20 @@ struct SlidingWindow_Element : JSON::Element {
...
+        layers_ = std::make_unique<IntArray_Element>(v_->layers);
+      }
+      return *layers_;
+    }
+    throw JSON::unknown_value_error{};
+  }
+
  private:
   std::optional<Config::Model::Decoder::SlidingWindow>& v_;
+  std::unique_ptr<IntArray_Element> layers_;
```
baijumeswani, line 495: Why does this need to be a unique ptr?

src/models/kv_cache.cpp
```diff
@@ -178,20 +179,67 @@ DefaultKeyValueCache::DefaultKeyValueCache(State& state)
   if (state.model_.p_device_->GetType() == DeviceType::NvTensorRtRtx &&
```
baijumeswani: Could we avoid this check so it can also be enabled for other EPs if needed through the genai config?

src/models/kv_cache.cpp
```diff
@@ -178,20 +179,67 @@ DefaultKeyValueCache::DefaultKeyValueCache(State& state)
...
       model_.config_->model.decoder.sliding_window.has_value() &&
       model_.config_->model.decoder.sliding_window->window_size > 0) {
-    shape_[2] = std::min(state_.params_->search.max_length,
-                         model_.config_->model.decoder.sliding_window->window_size);
+    int sliding_window_size = model_.config_->model.decoder.sliding_window->window_size;
+    int max_length = state_.params_->search.max_length;
```
baijumeswani:
```suggestion
    const int sliding_window_size = model_.config_->model.decoder.sliding_window->window_size;
    const int max_length = state_.params_->search.max_length;
```

src/models/kv_cache.h
```diff
@@ -97,6 +97,10 @@ struct DefaultKeyValueCache : KeyValueCache {
   std::array<int64_t, 4> shape_;
   ONNXTensorElementDataType type_;
 
+  // Support for per-layer KV cache shapes (for models with alternating attention patterns)
+  bool use_layer_types_{false};
+  std::vector<std::array<int64_t, 4>> layer_shapes_;
```
baijumeswani: Instead of these 2 variables, could you check if `layer_shapes_` is empty of not?

src/models/kv_cache.cpp
```diff
@@ -178,20 +179,67 @@ DefaultKeyValueCache::DefaultKeyValueCache(State& state)
...
+    // Check if we need per-layer allocation for models with alternating attention patterns
+    if (!model_.config_->model.decoder.sliding_window->layers.empty()) {
+      // Use per-layer allocation based on sliding window layer indices
+      use_layer_types_ = true;
+      layer_shapes_.resize(layer_count_);
+
+      // Create a set of sliding window layer indices for fast lookup
+      std::unordered_set<int> sliding_layers(
+          model_.config_->model.decoder.sliding_window->layers.begin(),
+          model_.config_->model.decoder.sliding_window->layers.end());
```
baijumeswani: Instead of using an unordered_set here, consider running 2 for loops. Once to copy the base shape and once to iterate over `model_.config_->model.decoder.sliding_window->layers` to update the shape for that index in layer_shapes_.

src/models/kv_cache.cpp
```diff
@@ -178,20 +179,67 @@ DefaultKeyValueCache::DefaultKeyValueCache(State& state)
...
+        presents_.push_back(OrtValue::CreateTensor(Allocator(), layer_shapes_[layer_idx], type_));
+        if (Device().GetType() != DeviceType::WEBGPU) {
+          ByteWrapTensor(Device(), *presents_.back()).Zero();
+        }
+
+        // Value tensor
+        presents_.push_back(OrtValue::CreateTensor(Allocator(), layer_shapes_[layer_idx], type_));
+        if (Device().GetType() != DeviceType::WEBGPU) {
+          ByteWrapTensor(Device(), *presents_.back()).Zero();
+        }
```
baijumeswani: This is the same code as above? Why not do it as is done in the else block where you iterate until `layer_count_ * 2`?

src/models/kv_cache.cpp
```diff
@@ -178,20 +179,67 @@ DefaultKeyValueCache::DefaultKeyValueCache(State& state)
...
+      }
+    } else {
```
baijumeswani: Usually, this kind of `if...else...` logic does not scale well. Since any changes made in the else block may not get applied to the if block. A lot of the code is just duplicated in the if and else blocks.
Is there any way to streamline the code and just use `layer_shapes_` instead of `shape_`?

src/models/kv_cache.cpp
```diff
@@ -271,39 +339,96 @@ void DefaultKeyValueCache::RewindTo(size_t index) {
...
+    int max_length = static_cast<int>(shape_[2]);  // Set to max_length in constructor
+    if (static_cast<int>(index) > max_length) {
+      throw std::runtime_error("Requested rewind length exceeds max_length.");
+    }
 
-    auto past_span = WrapTensor<T>(Device(), *past);
-    auto present_span = WrapTensor<T>(Device(), present);
+    for (int i = 0; i < layer_count_ * 2; i++) {
+      int layer_idx = i / 2;
+      std::array<int64_t, 4> layer_shape = layer_shapes_[layer_idx];
```
baijumeswani: Use consts where possible.

src/models/kv_cache.cpp
```diff
@@ -271,39 +339,96 @@ void DefaultKeyValueCache::RewindTo(size_t index) {
...
-  auto block_size_per_beam = shape_[1] * shape_[2] * shape_[3];
+
+  std::array<int64_t, 4> tensor_shape;
+  if (use_layer_types_) {
+    // Get shape from the actual tensor for per-layer allocation
+    OrtValue& present_value = *presents_[index];
+    auto present_shape = present_value.GetTensorTypeAndShapeInfo()->GetShape();
+    for (size_t i = 0; i < 4; i++) {
+      tensor_shape[i] = present_shape[i];
+    }
```
baijumeswani: std::copy(present_shape.begin(), present_shape.end(), tensor_shape.begin());

src/config.cpp
anujj: The JSON framework requires OnArray() to return a stable reference. Stack variables would create dangling references, and direct members can't be initialized at construction time. unique_ptr provides the required lazy initialization with a stable heap reference
 
 hope my understanding is correct and this make sense ?

src/models/kv_cache.cpp
anujj: done

src/models/kv_cache.cpp
anujj: done

src/models/kv_cache.h
anujj: done

src/models/kv_cache.cpp
anujj: done

src/models/kv_cache.cpp
anujj: done

src/models/kv_cache.cpp
anujj: done

src/models/kv_cache.cpp
anujj: done

anujj:
@baijumeswani : Thanks for the review, i have addressed the review comments

src/config.cpp
baijumeswani: Sorry for the late response. `layers_` can be allocated on the stack. Since it is only required for parsing, it will not create any dangling references.
But i think the issue is that v_ is std::optional and so there is no good way to construct layers_ without the `v_->layers` available during construction time. 

Let's leave it the way it is.

src/models/kv_cache.cpp
```diff
@@ -271,39 +320,94 @@ void DefaultKeyValueCache::RewindTo(size_t index) {
...
+  if (!layer_shapes_.empty()) {
+    // Get shape from the actual tensor for per-layer allocation
+    OrtValue& present_value = *presents_[index];
+    const auto present_shape = present_value.GetTensorTypeAndShapeInfo()->GetShape();
+    std::copy(present_shape.begin(), present_shape.end(), tensor_shape.begin());
+  } else {
+    tensor_shape = shape_;
+  }
+
+  auto block_size_per_beam = tensor_shape[1] * tensor_shape[2] * tensor_shape[3];
```
baijumeswani, line 407: Where is this variable used?

anujj: in the for loop below, there is no change in that part

PR #1833, opened 2025-10-17
Add missing tokenizer methods in java

baijumeswani:
- Add missing tokenizer functions in the java api.
- Update the get eos token ids function to return an array of ints instead of a `Sequences` object.

### DIFF
--- a/src/java/src/test/java/ai/onnxruntime/genai/TokenizerTest.java
+++ b/src/java/src/test/java/ai/onnxruntime/genai/TokenizerTest.java
@@ -105,8 +105,9 @@ public class TokenizerTest {
               + "{{ '<|assistant|>' }}{% else %}{{ eos_token }}{% endif %}";
 
       // From HuggingFace Python output for 'microsoft/Phi-4-multimodal-instruct'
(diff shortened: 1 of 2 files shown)


src/java/src/main/java/ai/onnxruntime/genai/Tokenizer.java
```diff
@@ -79,6 +79,109 @@ public String[] decodeBatch(Sequences sequences) throws GenAIException {
...
+    return tokenizerApplyChatTemplate(
+        nativeHandle, templateStr, messages, tools, addGenerationPrompt);
+  }
+
+  /**
+   * Updates tokenizer options.
+   *
+   * @param options Map of option keys to values.
+   * @throws GenAIException If the call to the GenAI native API fails.
+   */
```
sayanshaw24, line 165: should we document the currently supported options and their values here as well?

baijumeswani: No, I think just the C API is fine. Or else it'll be too much to maintain 😮‍💨

PR #1840, opened 2025-10-23
Change OpenVINO ep device selection priority

chemwolf6922:
## Why is this change made
GenAI added the ep device selection APIs for selecting ep devices according to the given ep name, device type, device vendor id and device id. However, the legacy OpenVINO specific device filtering logic has a higher priority than the new API. Causing the APIs ineffective when the "device_type" is specified in a OpenVINO genai_config.json. 
## What changed
This PR adjusts the two method's priority. So the new API can override what's provided in the config file.

### DIFF
--- a/src/models/model.cpp
+++ b/src/models/model.cpp
@@ -664,9 +664,9 @@ DeviceInterface* SetProviderSessionOptions(OrtSessionOptions& session_options,
       // for OpenVINO, fallback to "device_type" in provider_options if no device filtering is specified
       std::optional<std::string> config_ov_device_type = std::nullopt;
       if (provider_options.name == "OpenVINO" &&
-        !config_device_id.has_value() &&
-        !config_vendor_id.has_value() &&
-        !config_device_type_enum.has_value()) {
+          !config_device_id.has_value() &&
+          !config_vendor_id.has_value() &&
+          !config_device_type_enum.has_value()) {
         for (auto& option : provider_options.options) {
           if (option.first == "device_type") {
             config_ov_device_type = option.second;


src/models/model.cpp
```diff
@@ -661,19 +661,18 @@ DeviceInterface* SetProviderSessionOptions(OrtSessionOptions& session_options,
       std::optional<uint32_t> config_device_id = resolved_device_filtering.hardware_device_id;
       std::optional<uint32_t> config_vendor_id = resolved_device_filtering.hardware_vendor_id;
       std::optional<OrtHardwareDeviceType> config_device_type_enum = resolved_device_filtering.hardware_device_type;
-      // for OpenVINO, use "device_type" in provider_options exclusively if it's provided
+      // for OpenVINO, fallback to "device_type" in provider_options if no device filtering is specified
       std::optional<std::string> config_ov_device_type = std::nullopt;
-      if (provider_options.name == "OpenVINO") {
+      if (provider_options.name == "OpenVINO" &&
```
baijumeswani, line 666: Could you make the change at config parsing time? So, we update the `provider_options.device_filtering_options` after parsing the config. And we use that same option for setting the session options.

chemwolf6922: I agree this would be a cleaner solution. But we need to deprecate the following Intel specific behavior to do so:
As you can see from the code, Intel compares the "device_type" to the "ov_device" in their ep metadata for device selection. This behaves the same if "device_type" is always CPU, GPU or NPU (or any other standard device type in the future). However, Intel also supports setting the ov_device to something like "GPU1" in some versions of their EPs. And use that to select from multiple devices.
IMHO, Intel's old approach is not a very good one since you cannot generate a model with the correct setting if you don't know the CPUx name ahead. And it can be fully replaced by the device filtering APIs. If we can accept deprecation of this behavior, I'll happily remove the OpenVINO spaghetti here and move it forward to config parsing.

PR #1843, opened 2025-10-27
C# binding into STFT pre-processing

nenad1002:


src/ort_genai_c.cpp
```diff
@@ -888,6 +918,68 @@ OgaResult* OGA_API_CALL OgaProcessorProcessImagesAndAudiosAndPrompts(const OgaMu
...
+    OgaTensor* output0) {
+  OGA_TRY
+  if (!input || !sr_tensor || !frame_ms_tensor || !hop_ms_tensor ||
+      !energy_threshold_db_tensor || !output0) {
+    throw std::runtime_error("Null tensor argument passed to OgaSplitSignalSegments");
+  }
+
+  const OrtxTensor* in_tensor = MakeOrtxTensorConst<float>(input);
+  const OrtxTensor* sr_tensor_obj = MakeOrtxTensorConst<int64_t>(sr_tensor);
+  const OrtxTensor* frame_tensor = MakeOrtxTensorConst<int64_t>(frame_ms_tensor);
```
kunal-vaishnavi: Let's move the ORT extensions logic inside a [processor class](https://github.com/microsoft/onnxruntime-genai/blob/fbe74ff99657da87cef52b119fd6b23a1225b0e1/src/models/model.cpp#L1251) within ORT GenAI (e.g. `Processor` or `WhisperProcessor`).

src/ort_genai_c.cpp
```diff
@@ -78,6 +78,36 @@ T* ReturnUnique(std::unique_ptr<U> p) {
...
+  }
+
+  auto* gen = reinterpret_cast<Generators::Tensor*>(src);
+  T* data = const_cast<T*>(gen->GetData<T>());
+  std::vector<int64_t> shape = gen->GetShape();
+
+  auto* ort = new Ort::Custom::Tensor<T>(shape, data);
+  return reinterpret_cast<OrtxTensor*>(ort);
+}
+
```
kunal-vaishnavi: We should move this all inside a processor class. The user should not know what `OrtxTensor` is since it is internal to ORT extensions and is not public-facing in ORT GenAI. The base processor class also contains [utilities](https://github.com/microsoft/onnxruntime-genai/blob/fbe74ff99657da87cef52b119fd6b23a1225b0e1/src/models/processor.h#L51-L55) for dealing with `OrtxTensor` objects.

The C APIs should follow this format.
https://github.com/microsoft/onnxruntime-genai/blob/fbe74ff99657da87cef52b119fd6b23a1225b0e1/src/ort_genai_c.cpp#L873-L894

src/csharp/SignalProcessor.cs
```diff
@@ -0,0 +1,209 @@
...
+                        (ElementType)ET_Float32,
+                        out tensor));
+            }
+            return tensor;
+        }
+
+        /// <summary>
+        /// Create a tensor view over a managed long[] using OgaCreateTensorFromBuffer.
+        /// </summary>
+        public static unsafe IntPtr CreateInt64TensorFromArray(long[] data, long[] shape)
```
kunal-vaishnavi: Moving these conversion utilities into the processor class would allow us to re-use them for other language bindings.

src/ort_genai_c.cpp
nenad1002: Yes, I've seen this, but OgaTensor is defined only here and used only here, hence, the conversion methods will only ever be used in this file as well. The methods are not a part of C API as they're outside the scope.
I can do a workaround by using the Generators::Tensor, and that way use the methods in processor.h, but long term we might have to refactor to pull the types from above into a utils separate file if we want to have a logic separated.

src/ort_genai_c.cpp
nenad1002: I will move it inside the processor class.

src/csharp/SignalProcessor.cs
nenad1002: This makes sense. For now, I’ll keep them local since this PR is focused on getting C# interop working, and I would have to think how to write this properly, but I can follow up with a separate refactor later.

src/ort_genai_c.cpp
nenad1002: Done

src/csharp/SignalProcessor.cs
```diff
@@ -0,0 +1,209 @@
...
+// Licensed under the MIT License.
+
+using System;
+using System.Collections.Generic;
+using System.Runtime.InteropServices;
+
+namespace Microsoft.ML.OnnxRuntimeGenAI
+{
+    public static class SignalProcessor
+    {
```
kunal-vaishnavi, line 12: Can these APIs be added to the `MultiModalProcessor` instead of introducing a new class? All multi-modal models (e.g. Whisper, Phi-4 mm, etc) use that processor for pre-processing.

src/csharp/NativeMethods.cs
```diff
@@ -354,6 +354,21 @@ public static extern UIntPtr OgaSequencesGetSequenceCount(IntPtr /* const OgaSeq
                                                                                                   IntPtr /* const Audios* */ audios,
                                                                                                   out IntPtr /* OgaNamedTensors** */ namedTensors);
 
+        [DllImport(NativeLib.DllName, CallingConvention = CallingConvention.Winapi)]
+        public static extern int /* extError_t */ OgaSplitSignalSegments(
+            IntPtr /* const OgaTensor* */ input,
```
kunal-vaishnavi, line 359: Can we follow the same tab spacings as the other native method APIs listed in this file for the new APIs?

src/csharp/SignalProcessor.cs
kunal-vaishnavi: I have to create a new ORT GenAI 0.11.0 release so these APIs would become public-facing in that release if this PR is merged. If this PR needs to be merged for that release, then these APIs need to be moved inside the processor class within ORT GenAI before merging.

src/models/processor.cpp
```diff
@@ -198,8 +198,93 @@ std::unique_ptr<OrtValue> ProcessTensor<float, int64_t>(OrtxTensor* tensor, Ort:
   return tensor_value;
 }
 
+template <typename T>
+OrtxTensor* MakeOrtxTensor(Generators::Tensor* src) {
```
kunal-vaishnavi, line 202: Since the `Generators` namespace is included at the top, I think you can remove the `Generators::` part before each `Tensor`.

https://github.com/microsoft/onnxruntime-genai/blob/a12c530023b22fb42c1d4f057b46a0a7702d641f/src/models/processor.cpp#L7

src/ort_genai_c.cpp
```diff
@@ -888,6 +889,37 @@ OgaResult* OGA_API_CALL OgaProcessorProcessImagesAndAudiosAndPrompts(const OgaMu
...
 }
 
+OGA_EXPORT OgaResult* OGA_API_CALL OgaSplitSignalSegments(
+    const OgaTensor* input,
+    const OgaTensor* sr_tensor,
+    const OgaTensor* frame_ms_tensor,
+    const OgaTensor* hop_ms_tensor,
+    const OgaTensor* energy_threshold_db_tensor,
+    OgaTensor* output0) {
+  OGA_TRY
```
kunal-vaishnavi, line 899: Can we add unit tests for the new APIs in ORT GenAI's [C API tests](https://github.com/microsoft/onnxruntime-genai/blob/main/test/c_api_tests.cpp)?

src/ort_genai_c.h
```diff
@@ -705,6 +705,19 @@ OGA_EXPORT OgaResult* OGA_API_CALL OgaProcessorProcessImagesAndAudios(const OgaM
  */
 OGA_EXPORT OgaResult* OGA_API_CALL OgaProcessorProcessImagesAndAudiosAndPrompts(const OgaMultiModalProcessor*, const OgaStringArray* prompts, const OgaImages* images, const OgaAudios* audios, OgaNamedTensors** input_tensors);
 
+OGA_EXPORT OgaResult* OGA_API_CALL OgaSplitSignalSegments(
```
kunal-vaishnavi, line 708: Can you add [C# API tests](https://github.com/microsoft/onnxruntime-genai/blob/main/test/csharp/TestOnnxRuntimeGenAIAPI.cs) that call these C APIs?

src/models/processor.cpp
```diff
@@ -198,8 +198,93 @@ std::unique_ptr<OrtValue> ProcessTensor<float, int64_t>(OrtxTensor* tensor, Ort:
...
+    const Generators::Tensor* frame_ms_tensor,
+    const Generators::Tensor* hop_ms_tensor,
+    const Generators::Tensor* energy_threshold_db_tensor,
+    Generators::Tensor* output0) {
+  if (!input || !sr_tensor || !frame_ms_tensor || !hop_ms_tensor ||
+      !energy_threshold_db_tensor || !output0) {
+    throw std::runtime_error("Null tensor argument passed to OgaSplitSignalSegments");
+  }
+
+  const OrtxTensor* in_tensor = Generators::MakeOrtxTensorConst<float>(input);
```
sayanshaw24, line 239: it seems there's a constructor for every cast - given that there aren't tons of calls for it (nor will there be the need), i feel there isn't reason to clutter the API with these constructors, and we can just use `reinterpret_cast<OrtxTensor*>` when needed (which is what the constructors do).

nenad1002: This is doing more than just casting, we need this.

src/ort_genai_c.h
```diff
@@ -705,6 +705,19 @@ OGA_EXPORT OgaResult* OGA_API_CALL OgaProcessorProcessImagesAndAudios(const OgaM
...
+    const OgaTensor* input,
+    const OgaTensor* sr_tensor,
+    const OgaTensor* frame_ms_tensor,
+    const OgaTensor* hop_ms_tensor,
+    const OgaTensor* energy_threshold_db_tensor,
+    OgaTensor* output0);
+
+OGA_EXPORT OgaResult* OGA_API_CALL OgaMergeSignalSegments(
```
kunal-vaishnavi, line 716: Let's add comments above these new APIs to explain their usage and their parameters.

src/ort_genai_c.h
```diff
@@ -705,6 +705,19 @@ OGA_EXPORT OgaResult* OGA_API_CALL OgaProcessorProcessImagesAndAudios(const OgaM
...
+    const OgaTensor* segments_tensor,
+    const OgaTensor* merge_gap_ms_tensor,
```
kunal-vaishnavi, line 718: Given that the type of the object is an `OgaTensor*`, I think we can remove the `_tensor` suffix from the parameter names.

src/ort_genai_c.cpp
```diff
@@ -888,6 +889,37 @@ OgaResult* OGA_API_CALL OgaProcessorProcessImagesAndAudiosAndPrompts(const OgaMu
...
+  return reinterpret_cast<OgaResult*>(Generators::SplitSignalSegments(
+      reinterpret_cast<const Generators::Tensor*>(input),
+      reinterpret_cast<const Generators::Tensor*>(sr_tensor),
+      reinterpret_cast<const Generators::Tensor*>(frame_ms_tensor),
+      reinterpret_cast<const Generators::Tensor*>(hop_ms_tensor),
```
kunal-vaishnavi, line 904: To set the return value to `output0` in a way that does not cause memory corruptions, you may need to use `ReturnShared<OgaTensor>` and change `output0` to `OgaTensor**`.

https://github.com/microsoft/onnxruntime-genai/blob/cac543807ef295d52c27213697a0b9f1fb87b273/src/ort_genai_c.cpp#L632-L638

Let's also rename `output0` to `out` since there's just one output and to be consistent with existing API conventions.

src/ort_genai_c.cpp
```diff
@@ -888,6 +889,37 @@ OgaResult* OGA_API_CALL OgaProcessorProcessImagesAndAudiosAndPrompts(const OgaMu
...
+  OGA_CATCH
+}
+
+OGA_EXPORT OgaResult* OGA_API_CALL OgaMergeSignalSegments(
+    const OgaTensor* segments_tensor,
+    const OgaTensor* merge_gap_ms_tensor,
+    OgaTensor* output0) {
+  OGA_TRY
+  return reinterpret_cast<OgaResult*>(
+      Generators::MergeSignalSegments(
```
kunal-vaishnavi, line 916: We can move `SplitSignalSegments` and `MergeSignalSegments` inside a processor object. The outputs from those methods can then be set to the output tensors from the processor APIs.

```cpp
auto tensor = processor->SplitSignalSegments(...);
*out = ReturnShared<OgaTensor>(tensor);
```

```cpp
auto tensor = processor->MergeSignalSegments(...);
*out = ReturnShared<OgaTensor>(tensor);
```

Then these new APIs could be called similar to how methods such as `OgaTokenizerEncodeBatch` look in the above review comment. This will also help add future support for these new APIs in other language bindings.

src/csharp/SignalProcessor.cs
```diff
@@ -0,0 +1,259 @@
...
+            long[] mergedBacking = new long[MaxSegs * 2];
+
+            IntPtr input = IntPtr.Zero, sr = IntPtr.Zero, frame = IntPtr.Zero, hop = IntPtr.Zero,
+                   thr = IntPtr.Zero, splitOut = IntPtr.Zero, mergeGap = IntPtr.Zero, mergedOut = IntPtr.Zero;
+
+            try
+            {
+                long[] inputShape = new long[] { 1, inputSignal.Length };
+
+                input = CreateFloatTensorFromArray(inputSignal, inputShape);
```
kunal-vaishnavi, line 208: Can you construct a `Tensor` object directly for the input data? Each `Tensor` object has its own `IntPtr` handle that you can provide for the `NativeMethods.CAPI` call.

https://github.com/microsoft/onnxruntime-genai/blob/9d3631dbcf432da00e989c925ef8469b2aac5cc3/src/csharp/Tensor.cs#L28-L45

The `Tensor` class can help with memory management during disposal.

Here is an example with `Tensor` from a unit test.

https://github.com/microsoft/onnxruntime-genai/blob/9d3631dbcf432da00e989c925ef8469b2aac5cc3/test/csharp/TestOnnxRuntimeGenAIAPI.cs#L637-L669

src/models/processor.cpp
```diff
@@ -198,8 +198,93 @@ std::unique_ptr<OrtValue> ProcessTensor<float, int64_t>(OrtxTensor* tensor, Ort:
...
+  const OrtxTensor* sr_tensor_obj = Generators::MakeOrtxTensorConst<int64_t>(sr_tensor);
+  const OrtxTensor* frame_tensor = Generators::MakeOrtxTensorConst<int64_t>(frame_ms_tensor);
+  const OrtxTensor* hop_tensor = Generators::MakeOrtxTensorConst<int64_t>(hop_ms_tensor);
+  const OrtxTensor* thr_tensor = Generators::MakeOrtxTensorConst<float>(energy_threshold_db_tensor);
+  OrtxTensor* out_tensor = Generators::MakeOrtxTensor<int64_t>(output0);
```
kunal-vaishnavi, line 244: I would recommend looking at how the tokenizer and processor APIs are currently implemented (ex: [`Process`](https://github.com/microsoft/onnxruntime-genai/blob/9d3631dbcf432da00e989c925ef8469b2aac5cc3/src/models/whisper_processor.cpp#L18) method in the Whisper processor). That will help with making the implementations of the two methods similar to existing implementations.

1. You can use `CheckResult` in place of the `err` check.

2. Are tensors really needed for most of the parameters passed to `OrtxSplitSignalSegments`? The sampling rate (`sr_tensor`), frame size (`frame_ms`), hop length (`hop_ms`), and energy threshold (`energy_threshold_db`) are singular values. Can we modify `OrtxSplitSignalSegments` to use primitive types so that primitive types can be used in ORT GenAI for the "language binding API --> ORT GenAI C API --> ORT GenAI C++ API" flow? That will also greatly clean up the need for all of the workarounds you have in the C# bindings and in this file.

3. If `OrtxSplitSignalSegments` is updated, you can have the output stored as `OrtxTensorResult** result` inside that extensions API. Then you can create the output tensor inside this method instead of passing it in as a parameter to `SplitSignalSegments`. Here is an example.

https://github.com/microsoft/onnxruntime-genai/blob/9d3631dbcf432da00e989c925ef8469b2aac5cc3/src/models/whisper_processor.cpp#L27-L31

src/ort_genai_c.cpp
nenad1002: It doesn't cause memory corruption anymore, code I have here now is working correctly, so I wouldn't mess up again with changing it.

src/ort_genai_c.cpp
nenad1002: Why would I have to construct a new object which has no connection with my current functionality? I think what we need to do here is to clarify, do we even need a Processor class, and if we do, what exactly should be inside it, because I see many methods not being a part of the class (e.g. ProcessTensor is not a part of the class but is within the same file).

I can do this, but we really need to clarify this.

src/csharp/SignalProcessor.cs
nenad1002: Yes, this approach gave me memory corruption as well, there might be an issue with Tensor class, but I can give it another try.

src/models/processor.cpp
nenad1002: 2) Its more consistant having tensors, but I can do your approach as well, I have no preference, I was just trying to be consistant + I haven't seen built-in types used for Oga API.

3) It is possible, yes, but since I am initializing tensors on C# side, again for consistancy and simplicity (no need for **), I decided to create all on the caller side. But the extension should, in theory, work for both cases.

PR #1844, opened 2025-10-28
Migrate Python bindings from pybind11 to nanobind v2.9.2

snnn:
This pull request migrates the Python bindings for onnxruntime-genai from Pybind11 to Nanobind to leverage
  Nanobind's performance and feature improvements.

  Key Changes:

   * Build System (CMake):
       * The dependency in cmake/deps.txt has been switched from Pybind11 to Nanobind v2.9.2.
       * cmake/external/onnxruntime_external_deps.cmake is updated to fetch Nanobind and its robin-map
         dependency using FetchContent.
       * In src/python/CMakeLists.txt, the module is now built using nanobind_add_module instead of
         pybind11_add_module.

   * Python Bindings (`src/python/python.cpp` and related files):
       * All Pybind11 headers and API calls have been replaced with their Nanobind equivalents.
       * The main module is now initialized with NB_MODULE instead of PYBIND11_MODULE.
       * Class and function bindings have been updated to the Nanobind syntax.
       * A cleanup function has been registered with Python's atexit to ensure proper garbage collection before
         C++ static destructors are called, preventing false memory leak reports.

  Benefits of this migration:

   * Faster Compile Times: Nanobind's header-only and lightweight design significantly reduces compilation time.
   * Smaller Binaries: The resulting Python extension module is smaller.
   * Lower Runtime Overhead: Nanobind has lower overhead for function calls between Python and C++.
   * Modern C++ and Python Features: Provides better support for modern C++ features and improved Python
     integration.

  This migration modernizes the onnxruntime-genai Python bindings, making them more efficient, robust, and
  easier to maintain.

### DIFF
--- a/src/python/CMakeLists.txt
+++ b/src/python/CMakeLists.txt
@@ -37,8 +37,14 @@ target_link_directories(python PRIVATE ${ORT_LIB_DIR})
 target_link_libraries(python PRIVATE 
   onnxruntime-genai-objects
   onnxruntime_extensions
-  ${ONNXRUNTIME_LIB}
 )
+
+# On Windows, we need to link against ONNXRUNTIME_LIB for proper symbol resolution
+# On Linux/macOS/Android, the ORT library is preloaded at runtime via _dll_directory.py
+# before the Python extension loads, so we don't need it at link time
+if(WIN32)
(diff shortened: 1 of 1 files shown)


snnn:
Sorry I didn't have enough time to finish this work.

PR #1846, opened 2025-10-29
Add disable cuda graph when num_beams > 1 and fix set_provider_option bug

anujj:
## Problem

When using beam search (`num_beams > 1`) with cuda graph enable , the following runtime error :

```
RuntimeError: Graph capture is not supported with past_present_share_buffer set to false.
```

### Technical Background

When `num_beams > 1`:
- GenAI automatically sets `past_present_share_buffer` to `false` 
- Beam search is supported with out-of-place KV cache (which requires `past_present_share_buffer=false`)

However, CUDA graph has a conflicting requirement:
- CUDA graph is **only** supported when `past_present_share_buffer=true` 
- CUDA graph requires upfront memory allocations and fixed memory patterns

This creates an incompatibility: beam search needs `past_present_share_buffer=false`, while CUDA graph needs it to be `true`.

### The Solution

In `model-generate.py`, when `num_beams > 1`, we explicitly set `enable_cuda_graph` to `"0"` to disable CUDA graph. This prevents the runtime error by ensuring CUDA graph isn't enabled when using beam search.

### SetProviderOption Bug

`SetProviderOption()` was **appending** new options instead of **replacing** existing ones (if avalable). This created duplicate entries in the provider_options vector:

```cpp
// After calling set_provider_option("NvTensorRtRtx", "enable_cuda_graph", "0")
provider_options = [
  {"name": "NvTensorRtRtx", "options": [("enable_cuda_graph", "1")]},  // Original from genai_config.json
  {"name": "NvTensorRtRtx", "options": [("enable_cuda_graph", "0")]}   // Newly appended
]
```

### The Fix in SetProviderOption (src/config.cpp)

The fix modifies `SetProviderOption()` to:
1. **First remove** any existing option with the same name **if available** 
2. **Then add** the new value

### DIFF
--- a/.github/workflows/linux-cpu-arm64-build.yml
+++ b/.github/workflows/linux-cpu-arm64-build.yml
@@ -28,8 +28,8 @@ jobs:
       - uses: actions/setup-dotnet@v5
         with:
           dotnet-version: |
-          8.0.x
-          9.0.x
+            8.0.x
+            9.0.x
 
       - name: Install Rust with rustup
         run: |
--- a/.pipelines/stages/jobs/steps/capi-win-step.yml
+++ b/.pipelines/stages/jobs/steps/capi-win-step.yml
@@ -153,6 +153,15 @@ steps:
       artifactName: $(artifactName)-capi
       targetPath: '$(Build.Repository.LocalPath)\build\$(ep)\$(os)-$(arch)\package'
 
+  - task: PublishSymbols@2
+    displayName: 'Publish Build Symbols'
+    condition: and(succeeded(), eq(variables['build_config'], 'relwithdebinfo'))
+    inputs:
+      SymbolsFolder: '$(Build.Repository.LocalPath)\build\$(ep)\$(os)-$(arch)\RelWithDebInfo\'
+      SearchPattern: 'onnxruntime-genai*.pdb'
+      SymbolServerType: teamServices
+      SymbolExpirationInDays: 365
+
 - ${{ if eq(parameters.target, 'python') }}:
   - task: BinSkim@4
     displayName: 'Run BinSkim'
--- a/benchmark/python/benchmark_e2e.py
+++ b/benchmark/python/benchmark_e2e.py
@@ -247,6 +247,9 @@ def run_benchmark(args, batch_size, prompt_length, generation_length, max_length
             args.chat_template = '<s>{input}'
         elif model_type.startswith("qwen2"):
             args.chat_template = '<|im_start|>user\n{input}<|im_end|>\n<|im_start|>assistant\n'
+        elif model_type.startswith("gemma"):
+            # Gemma and Gemma2 models use this format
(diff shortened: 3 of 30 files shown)


examples/python/model-generate.py
```diff
@@ -21,24 +21,34 @@ def main(args):
...
+    # Configure execution provider if specified
     if args.execution_provider != "follow_config":
         config.clear_providers()
         if args.execution_provider != "cpu":
             if args.verbose:
                 print(f"Setting model to {args.execution_provider}...")
             config.append_provider(args.execution_provider)
+
+    # Disable CUDA graph if using beam search (num_beams > 1),
+    # num_beams > 1 requires past_present_share_buffer to be false and out of place kv cache is not supported with cuda graph true hance make is false
```
```suggestion
    # num_beams > 1 requires past_present_share_buffer to be false so enable_cuda_graph must be false
```

examples/python/model-generate.py
```diff
@@ -21,24 +21,34 @@ def main(args):
...
+    if args.num_beams > 1:
+        config.set_provider_option(args.execution_provider, "enable_cuda_graph", "0")
```
kunal-vaishnavi, line 41: Does this need to be set by the user? Can it be handled inside ORT GenAI?

anujj: I had the same thought, but the only reason I implemented it this way in the application/example is that enable_cuda_graph is required at the og.Model level and passed through the SetProvider option. In contrast, beam search is a runtime configuration that becomes available later, and the application might also change its value at runtime. At least this approach works for all other EPs.

For the TRT-RTX EP, however, the maximum number of beams also needs to be set at load time because TRT-RTX requires the dynamic shape range at that stage here: https://github.com/microsoft/onnxruntime-genai/blob/cdb7ff3f4f0298ed64e7718c2490bebc086d0ccb/src/models/model.cpp#L501

So, I thought when the application sets the number of beams, it should disable CUDA graph if num_beam > 1.

If you’re okay with handling this in GenAI, I can check for the right place to implement it within the GenAI code.

kunal-vaishnavi: I think it's fine to keep it as-is then.

kunal-vaishnavi (dismissed):

kunal-vaishnavi:
For the Windows CUDA x64 and Windows WinML x64 CI failures, can you try upgrading CUDA from 12.2 to 12.4 in the below locations to see if that resolves the issue?

https://github.com/microsoft/onnxruntime-genai/blob/d4eabac69d919ddf4d499940866dab695058a0f7/.github/workflows/win-cuda-x64-build.yml#L18-L19

https://github.com/microsoft/onnxruntime-genai/blob/d4eabac69d919ddf4d499940866dab695058a0f7/.github/workflows/win-winml-x64-build.yml#L18-L19

kunal-vaishnavi (approved):

PR #1851, opened 2025-11-03
Enable If Node Support for TRT-RTX in Phi-3.5/Phi-4 LongRoPE Models

anujj:
## Problem

Models with LongRoPE (Phi-3.5, Phi-4) use If nodes to conditionally select between small (4K) and large (128K) rotary embedding caches. TRT-RTX execution provider has two issues preventing proper If node execution:

### Issue 1: Mismatched Dimensions
TRT-RTX requires uniform output dimensions in both If node branches. Current implementation has mismatched shapes:
- Then branch (large cache): (131072, 48)
- Else branch (small cache): (4096, 48)

**Previous workaround**: Concatenate both caches into a single tensor, avoiding If nodes entirely.

### Issue 2: trt-rtx Multi-Output Bug
TRT-RTX has a bug where If nodes with multiple outputs cause tensor overwriting. When outputting `[cos_cache, sin_cache]` together, assigns both to the same memory location, corrupting `cos_cache`.

## Solution

Implemented **two workarounds** specifically for TRT-RTX:

### Workaround 1: Padding to Uniform Dimensions

Pad small cache from (4096, 48) to (131072, 48) using mathematically correct values:
- **cos_cache**: Pad with `1.0` (cos(0) = 1)
- **sin_cache**: Pad with `0.0` (sin(0) = 0)

This achieves uniform dimensions while preserving rotary embedding correctness.

### Workaround 2: Split Multi-Output If Nodes

Split one If node with 2 outputs into two separate If nodes with 1 output each:

**Before**:
```
If node → outputs: [cos_cache, sin_cache]  (2 outputs = trt-rtx bug)
```

**After**:
```
If_cos node → output: cos_cache  (1 output)
If_sin node → output: sin_cache  (1 output)
```

Both If nodes use the same condition and select their respective caches independently, avoiding the trt-rtx tensor overwriting bug.

### Modified Files
- `src/python/py/models/builder.py`

### Key Modifications
1. Added `_pad_cache_to_uniform_shape()` helper method for padding
2. Removed TRT-RTX from DML concatenation workaround
3. Apply padding to small caches for TRT-RTX (cos=1, sin=0)
4. Create 2 separate If nodes for TRT-RTX instead of 1 multi-output If node
5. Preserve original single If node behavior for other EPs

**Please note that this is a temporary workaround, final fix will come in the trt-rtx backend**

### DIFF
--- a/.github/workflows/win-winml-x64-build.yml
+++ b/.github/workflows/win-winml-x64-build.yml
@@ -15,8 +15,8 @@ env:
   AZCOPY_AUTO_LOGIN_TYPE: MSI
   AZCOPY_MSI_CLIENT_ID: 63b63039-6328-442f-954b-5a64d124e5b4
   cuda_dir: "${{ github.workspace }}\\cuda_sdk"
-  cuda_version: "12.4"
-  CUDA_PATH: ${{ github.workspace }}\\cuda_sdk\\v12.4
+  cuda_version: "12.2"
+  CUDA_PATH: ${{ github.workspace }}\\cuda_sdk\\v12.2
   binaryDir: 'build/cuda/win-x64'
   ORT_NIGHTLY_REST_API: "https://feeds.dev.azure.com/aiinfra/PublicPackages/_apis/packaging/Feeds/ORT-Nightly/packages?packageNameQuery=Microsoft.ML.OnnxRuntime.Gpu.Windows&api-version=6.0-preview.1"
   ORT_PACKAGE_NAME: "Microsoft.ML.OnnxRuntime.Gpu.Windows"


anujj:
Adding @gaugarg-nv , @baijumeswani , @kunal-vaishnavi for review

src/python/py/models/builder.py
```diff
@@ -1423,6 +1423,27 @@ def make_rotary_embedding_caches(self, **kwargs):
...
+        """Pad small cache to match large cache shape for uniform If node branches.
+
+        This is used for TRT-RTX EP which requires uniform dimensions in both branches of If nodes.
+
+        Args:
+            small_cache: The smaller cache tensor to pad
+            large_cache: The larger cache tensor (defines target shape)
+            pad_value: Value to use for padding (1.0 for cos_cache, 0.0 for sin_cache)
+        """
+        import torch
```
```suggestion
```

src/python/py/models/builder.py
```diff
@@ -1423,6 +1423,27 @@ def make_rotary_embedding_caches(self, **kwargs):
 
         return cos_cache_name, sin_cache_name
 
+    def _pad_cache_to_uniform_shape(self, small_cache, large_cache, pad_value=0.0):
```
```suggestion
    def make_padded_cache(self, small_cache, large_cache, pad_value=0.0):
```

src/python/py/models/builder.py
```diff
@@ -1500,44 +1528,117 @@ def make_rotary_embedding_multi_cache(self, **kwargs):
...
-                inputs=[],
-                outputs=[
-                    cos_cache_large_node.outputs[0],
-                    sin_cache_large_node.outputs[0],
+
+        # For TRT-RTX, split into two separate If nodes to workaround trt-rtx bug
+        # where multiple outputs from a single If node cause tensor overwriting
+        if self.ep == "trt-rtx":
+            # Create separate If node for cos_cache only
+            cos_if_name = f"{basename}/If_cos"
```
```suggestion
            cos_if_name = f"{basename}/cos/If"
```

src/python/py/models/builder.py
```diff
@@ -1500,44 +1528,117 @@ def make_rotary_embedding_multi_cache(self, **kwargs):
...
-            "If", inputs=[f"{greater_name}/output_0"], outputs=[cos_cache_name, sin_cache_name], name=if_name,
-            then_branch=ir.Graph(
-                inputs=[],
-                outputs=[
-                    cos_cache_large_node.outputs[0],
-                    sin_cache_large_node.outputs[0],
+
+        # For TRT-RTX, split into two separate If nodes to workaround trt-rtx bug
+        # where multiple outputs from a single If node cause tensor overwriting
+        if self.ep == "trt-rtx":
```
kunal-vaishnavi: Let's move the splitting logic into a standalone method, call it in the [earlier check](https://github.com/microsoft/onnxruntime-genai/blob/cdb7ff3f4f0298ed64e7718c2490bebc086d0ccb/src/python/py/models/builder.py#L1499-L1501) for the TRT-RTX EP, and then early return. Since it is temporary, it will be easier to just delete that method once the bug is fixed instead of reverting these changes. We should also mark it with a leading underscore so that it is not a public method.

src/python/py/models/builder.py

src/python/py/models/builder.py

src/python/py/models/builder.py

src/python/py/models/builder.py

anujj:
@kunal-vaishnavi : I have addressed all the review comments

src/python/py/models/builder.py
```diff
@@ -1500,44 +1642,50 @@ def make_rotary_embedding_multi_cache(self, **kwargs):
...
             "Constant", [], outputs=[
-                ir.Value(name=sin_cache_small_name, type=ir.TensorType(self.io_dtype), shape=ir.Shape(sin_cache_small.shape))
+                ir.Value(name=sin_cache_small_name, type=ir.TensorType(self.io_dtype), shape=ir.Shape(small_cache_shape))
             ],
             name="/small/sin_cache/Constant", attributes=dict(value=ir.tensor(sin_cache_small)))
+
+        # Create single If node with multiple outputs
         self.make_node(
             "If", inputs=[f"{greater_name}/output_0"], outputs=[cos_cache_name, sin_cache_name], name=if_name,
-            then_branch=ir.Graph(
```
kunal-vaishnavi, line 1544: Let's revert these below changes since `_make_split_if_nodes_for_trt_rtx` now exists and the below logic is not reached for TRT-RTX.

anujj: done

PR #1853, opened 2025-11-03
Mixed precision export support for gptq quantized model

rM-planet:
1> Changes in OGA to support mixed precision export of models quantized with GPTQModel.
2> Changes to decide whether to use packed matmul or not on the basis of q,k,v precisions.

### DIFF
--- a/src/python/py/models/builder.py
+++ b/src/python/py/models/builder.py
@@ -2004,9 +2004,15 @@ class Model:
 
         # Unpack attention weights if needed
         self.make_attention_unpacked(layer_id, attention, root_input, **kwargs)
+        
+        # Get dtype used for MatMul ops
+        q_dtype = getattr(attention.q_proj, "weight", getattr(attention.q_proj, "bits", None))
+        k_dtype = getattr(attention.k_proj, "weight", getattr(attention.k_proj, "bits", None))
+        v_dtype = getattr(attention.v_proj, "weight", getattr(attention.v_proj, "bits", None))
+        qkv_dtype_equal = getattr(q_dtype, "dtype", q_dtype) == getattr(k_dtype, "dtype", k_dtype) == getattr(v_dtype, "dtype", v_dtype)
 
         # Make MatMul nodes
-        if self.attention_attrs["use_packed_matmul"]:
+        if self.attention_attrs["use_packed_matmul"] and qkv_dtype_equal:
             # Combine 3 MatMuls into 1 packed MatMul
             qkv_matmul_basename = f"/model/layers.{layer_id}/attn/qkv_proj/MatMul"
             qkv_matmul_name = self.make_packed_matmul(attention.q_proj, attention.k_proj, attention.v_proj, qkv_matmul_basename, root_input)
@@ -2028,7 +2034,7 @@ class Model:
         v_bias_exists = attention.v_proj.bias is not None and torch.count_nonzero(attention.v_proj.bias) > 0
         any_bias_exists = q_bias_exists or k_bias_exists or v_bias_exists
 
-        if self.attention_attrs["use_packed_matmul"] and any_bias_exists:
+        if self.attention_attrs["use_packed_matmul"] and qkv_dtype_equal and any_bias_exists:
             # Combine 3 Adds into 1 packed Add
             qkv_add_name = f"/model/layers.{layer_id}/attn/qkv_proj/Add"
             self.make_packed_add(attention.q_proj.bias, attention.k_proj.bias, attention.v_proj.bias, qkv_add_name, root_input=self.attention_attrs["q_path"])
--- a/src/python/py/models/quantized_model.py
+++ b/src/python/py/models/quantized_model.py
@@ -863,6 +863,22 @@ class GPTQModel(QuantizedModel):
         self.pack_qzeros(temp_module)
         module.qzeros = temp_module.qzeros
 
+    def _load_quant_config(self, quant_attrs):
+        super()._load_quant_config(quant_attrs)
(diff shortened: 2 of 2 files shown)


src/python/py/models/builder.py
```diff
@@ -373,6 +373,9 @@ def make_attention_init(self):
                 and not self.attention_attrs["q_norm"]
                 and not self.attention_attrs["k_norm"]
             )
+            
+            if "use_packed_matmul" in self.extra_options:
```
kunal-vaishnavi: This is an optimization opportunity that should be auto-detected by the model builder. We should not need to give the responsibility to the user.

src/python/py/models/quantized_model.py
```diff
@@ -863,6 +863,22 @@ def __init__(self, module):
         self.pack_qzeros(temp_module)
         module.qzeros = temp_module.qzeros
 
+    def _load_quant_config(self, quant_attrs):
+        super()._load_quant_config(quant_attrs)
+        self.overrides = quant_attrs["config"]["dynamic"] if "dynamic" in quant_attrs["config"] else {}
```
kunal-vaishnavi:
```suggestion
        self.overrides = quant_attrs["config"].get("dynamic", {})
```

src/python/py/models/builder.py
rM-planet: Hi Kunal, there are some cases where in order to improve accuracy we want to have Q,K,V in different precisions. For those cases, we want to have unpacked q,k,v.

kunal-vaishnavi: This can still be done inside the model builder. For example:

```python
if Q.dtype != K.dtype or K.dtype != V.dtype or Q.dtype != V.dtype:
    make_separate_matmuls(...)
else:
    make_packed_matmul(...)
```

We anyways don't expect to use a packed MatMul where the Q, K, and V tensors contain different precisions. A packed MatMul should contain the same precision for all three tensors.

https://github.com/microsoft/onnxruntime-genai/blob/d4eabac69d919ddf4d499940866dab695058a0f7/src/python/py/models/builder.py#L1066-L1071

rM-planet: @kunal-vaishnavi Thanks for the suggestion. I have addressed the changes accordingly.

src/python/py/models/builder.py
```diff
@@ -2006,7 +2006,7 @@ def make_attention(self, layer_id, attention, root_input, **kwargs):
         self.make_attention_unpacked(layer_id, attention, root_input, **kwargs)
 
         # Make MatMul nodes
-        if self.attention_attrs["use_packed_matmul"]:
+        if self.attention_attrs["use_packed_matmul"] and (self.quant_type is None or (attention.q_proj.bits == attention.k_proj.bits == attention.v_proj.bits)):
```
kunal-vaishnavi: The value of `use_packed_matmul` should already consider whether the bits are equal or not. Can we find a way to use the bits check to set the value of `use_packed_matmul`?

https://github.com/microsoft/onnxruntime-genai/blob/e6ff6972d969d7ee4a4a05f358560955a902dc08/src/python/py/models/builder.py#L368-L375

We should be able to re-use that check in other locations within the model builder as needed (e.g. below).

https://github.com/microsoft/onnxruntime-genai/blob/d4eabac69d919ddf4d499940866dab695058a0f7/src/python/py/models/builder.py#L1066-L1071

rM-planet: Currently, "use_packed_matmul" is like a global check i.e applies to all layers. I think it is good for feature based conditions ('use_lora', "dml ep") applicable for all layers. However, there are cases where we want to have initial sensitive few layers in mixed precision and remaining in same low-bit precision and because of that I want to do this check for each layer instead of setting it globally(more restrictive).

kunal-vaishnavi: Can we leverage a similar style used below for the bits checks?

https://github.com/microsoft/onnxruntime-genai/blob/e6ff6972d969d7ee4a4a05f358560955a902dc08/src/python/py/models/builder.py#L2025-L2031

Something like the following should work.

```py
# Get dtype used for MatMul ops
q_dtype = getattr(attention.q_proj, "bits", attention.q_proj.weight.dtype)
k_dtype = getattr(attention.k_proj, "bits", attention.k_proj.weight.dtype)
v_dtype = getattr(attention.v_proj, "bits", attention.v_proj.weight.dtype)
all_dtype_equal = q_dtype == k_dtype == v_dtype

if self.attention_attrs["use_packed_matmul"] and all_dtype_equal:
```

I would like to keep the boolean expression for the if condition as simple as possible.

rM-planet: I have made the changes accordingly.

src/python/py/models/builder.py
```diff
@@ -2004,9 +2004,15 @@ def make_attention(self, layer_id, attention, root_input, **kwargs):
 
         # Unpack attention weights if needed
         self.make_attention_unpacked(layer_id, attention, root_input, **kwargs)
+        
+        # Get dtype used for MatMul ops
+        q_dtype = getattr(attention.q_proj, "bits", None) or getattr(attention.q_proj.weight, "dtype", None)
+        k_dtype = getattr(attention.k_proj, "bits", None) or getattr(attention.k_proj.weight, "dtype", None)
```
kunal-vaishnavi: Can we change the predicate order to check for `dtype` first and then `bits` second? Most models generated via model builder have not been quantized in advance so the latter predicate is true more often than the latter.

rM-planet: @kunal-vaishnavi made the changes accordingly.

kunal-vaishnavi (approved):

PR #1854, opened 2025-11-03
Shared emb_tokens/lm_head on fp16 & uint4 weights

jixiongdeng:
## Problem

The current model builder doesn't support shared embeddings layers with 4bit qweights and 16bit float weights, which occupies more room in disk (unnecessary for originally tied embeddings models) and hurts compression rate for quantized models. builder.py doesn't provide flexible options to toggle the graph construction and quantization config, like unpacked/packed matmul, rtn, kquant, etc. 

## Solution

Calculated flat_dim in a more generic way on reshape node before `GatherBlockQuantized` (support 4bit and 8bit).
Added CUDA kernel support in ORT [#26484](https://github.com/microsoft/onnxruntime/pull/26484).  
Added more extra_options to enable different quant configs and pack options, and shared embeddings.

Running examples:
**unpacked qkv_projs and shared 4 bit RTN on Llama3.2 1B Instruct**:
```
python src/python/py/models/builder.py -m meta-llama/Llama-3.2-1B-Instruct -p int4 -e cuda -o export_model/llama32_1bi_rtn_4_4_unpacked_tied --extra_options int4_is_symmetric=false unpack_matmul=true int4_algo_config=rtn
```

**shared 4 bit k_quant on Phi-4-Mini Instruct**:
```
python src/python/py/models/builder.py -m microsoft/Phi-4-Mini-Instruct -p int4 -e cuda -o export_model/phi4mini_i_kquant_4_4_tied --extra_options int4_is_symmetric=false unpack_matmul=true int4_algo_config=k_quant
```

**shared 16 bit float emb on Phi-4-Mini Instruct**:
```
python src/python/py/models/builder.py -m microsoft/Phi-4-Mini-Instruct -p fp16 -e cuda -o export_model/phi4mini_i_fp16_tied --extra_options shared_embeddings=true
```

### Modified Files
- `src/python/py/models/builder.py`

### Key Modifications
1. Computed `flat_dim` in a generic manner before feeding in `GatherBlockQuantized`.
2. Explicitly defined gather_axis and quantize_axis for clarity.
3. Added `unpack_matmul` option to separate qvk_proj if needed.
4. Added `shared_embeddings` option to tied embed_tokens/lm_head.
5. Added `rtn_last` like `k_quant_last` as a new mixed precision option
6. Added `k_quant` like `rtn` as a new 4 bit quantizer option

### DIFF
--- a/src/python/py/models/builder.py
+++ b/src/python/py/models/builder.py
@@ -385,8 +385,8 @@ class Model:
             )
 
             # Allow extra_options to override use_packed_matmul
-            if "unpack_matmul" in extra_options:
-                self.attention_attrs["use_packed_matmul"] = not extra_options.get("unpack_matmul", False)
+            if "unpack_matmul" in self.extra_options:
+                self.attention_attrs["use_packed_matmul"] = not self.extra_options.get("unpack_matmul", False)
 
             # Some EPs don't support fusing rotary embeddings inside GQA yet
             self.attention_attrs["use_rope_in_attn"] = self.ep not in ["dml"]


src/python/py/models/builder.py
```diff
@@ -375,6 +376,10 @@ def make_attention_init(self):
                 and not self.attention_attrs["k_norm"]
             )
 
+            # Allow extra_options to override use_packed_matmul
+            if "unpack_matmul" in extra_options:
```
kunal-vaishnavi: This is an optimization opportunity that should be auto-detected by the model builder. We should not need to give the responsibility to the user. You can see the review comments on [this PR](https://github.com/microsoft/onnxruntime-genai/pull/1853) for more details.

src/python/py/models/builder.py
```diff
@@ -484,28 +489,35 @@ def make_int4_algo_config(self, quant_method: str):
...
+        if quant_method in {"rtn", "rtn_last"}:
+            if quant_method == "rtn":
+                int4_algo_config = RTNWeightOnlyQuantConfig()
+            elif quant_method == "rtn_last":
+                customized_weight_config["/lm_head/MatMul"] = {"bits": 8}
+                int4_algo_config = RTNWeightOnlyQuantConfig(customized_weight_config=customized_weight_config)
 
-        elif quant_method in {"k_quant_mixed", "k_quant_last"}:
+        elif quant_method in {"k_quant", "k_quant_mixed", "k_quant_last"}:
             from onnxruntime.quantization.matmul_nbits_quantizer import KQuantWeightOnlyQuantConfig
```
kunal-vaishnavi, line 506: Let's move this import up. It was previously here because it was not part of a stable release.

https://github.com/microsoft/onnxruntime-genai/blob/d4eabac69d919ddf4d499940866dab695058a0f7/src/python/py/models/builder.py#L24-L28

src/python/py/models/builder.py
```diff
@@ -484,28 +489,35 @@ def make_int4_algo_config(self, quant_method: str):
         customized_weight_config = {}
         int4_algo_config = None
 
-        if quant_method == "rtn":
-            int4_algo_config = RTNWeightOnlyQuantConfig()
+        if quant_method in {"rtn", "rtn_last"}:
```
kunal-vaishnavi, line 500: I think this can be simplified to the following.

```python
if quant_method in {"rtn", "rtn_last"}:
    if quant_method == "rtn_last":
        customized_weight_config["/lm_head/MatMul"] = {"bits": 8}
    int4_algo_config = RTNWeightOnlyQuantConfig(customized_weight_config=customized_weight_config)
```

src/python/py/models/builder.py
```diff
@@ -484,28 +489,35 @@ def make_int4_algo_config(self, quant_method: str):
...
+            if quant_method == "rtn":
+                int4_algo_config = RTNWeightOnlyQuantConfig()
+            elif quant_method == "rtn_last":
+                customized_weight_config["/lm_head/MatMul"] = {"bits": 8}
+                int4_algo_config = RTNWeightOnlyQuantConfig(customized_weight_config=customized_weight_config)
 
-        elif quant_method in {"k_quant_mixed", "k_quant_last"}:
+        elif quant_method in {"k_quant", "k_quant_mixed", "k_quant_last"}:
```
kunal-vaishnavi, line 505: I think this can be simplified to the following.

```py
elif quant_method in {"k_quant", "k_quant_mixed", "k_quant_last"}:
    if quant_method != "k_quant":
        customized_weight_config["/lm_head/MatMul"] = {"bits": 8}

    if quant_method == "k_quant_mixed":
        # k_quant_mixed is from llama.cpp.
        # Reference: https://github.com/ggml-org/llama.cpp/blob/36667c8edcded08063ed51c7d57e9e086bbfc903/src/llama-quant.cpp#L136
        # We also consider some MatMuls are more senstive to quantization than other MatMuls.
        layers_to_exclude = [
            i
            for i in range(self.num_layers)
            if i < self.num_layers / 8 or i >= 7 * self.num_layers / 8 or (i - (round)(self.num_layers / 8)) % 3 == 2
        ]
        for i in layers_to_exclude:
            customized_weight_config["/model/layers." + str(i) + "/attn/qkv_proj/MatMul"] = {"bits": 8}
            customized_weight_config["/model/layers." + str(i) + "/attn/v_proj/MatMul"] = {"bits": 8}
            customized_weight_config["/model/layers." + str(i) + "/mlp/down_proj/MatMul"] = {"bits": 8}

    int4_algo_config = KQuantWeightOnlyQuantConfig(customized_weight_config=customized_weight_config)
```

src/python/py/models/builder.py
```diff
@@ -285,9 +285,10 @@ def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
 
         self.int4_tied_embeddings = config.tie_word_embeddings if hasattr(config, "tie_word_embeddings") and config.tie_word_embeddings is not None else False
         self.int4_tied_embeddings = extra_options.get("int4_tied_embeddings", self.int4_tied_embeddings)
-        self.int8_lm_head = extra_options.get("int4_algo_config", "default") in {"k_quant_mixed", "k_quant_last"}
-        if not self.int8_lm_head:
+        self.int8_lm_head = extra_options.get("int4_algo_config", "default") in {"k_quant_mixed", "k_quant_last", "rtn_last"}
+        if not self.int8_lm_head and extra_options.get("int4_algo_config", "default") not in {"rtn", "k_quant"}:
```
kunal-vaishnavi: Can we rewrite the above section and the if condition to just match on the conditions needed for tied embeddings to be true and otherwise set it to false?

Something like this:

```py
self.int8_lm_head = extra_options.get("int4_algo_config", "default") in {"k_quant_mixed", "k_quant_last", "rtn_last"}
self.int4_tied_embeddings = extra_options.get("int4_tied_embeddings", config.tie_word_embeddings if hasattr(config, "tie_word_embeddings") and config.tie_word_embeddings is not None else False)

# matmul_nbits_quantizer.py has a different naming for default quantization, so lm_head.MatMul.weight_Q{}G{} does not match.
# tied_embeddings lm_head.MatMul.weight_Q{}G{} only works with rtn&k_quant on 4bit
self.int4_tied_embeddings = <boolean expression>
```

jixiongdeng:
@kunal-vaishnavi  Thanks for the review! I edited codes and adapted to most of your comments. 
The `unpack_matmul` option relates to my other project which involves weights replacements. 
i.e. Llama3.2 series models originally have q_proj, k_proj, v_proj separately . It would be great, if builder.py can provide an easy option to keep consistent with torch models' projs.

jixiongdeng:
Added an option to create shared emb_tokens/lm_head for float matmul node.
This would benefit small models with smaller size, especially for low-bit quantized models. 
i.e. phi4mini fp16: (8.3G -> 7.2G) ~ 13.2% smaller
      phitmini int4 + fp16 emb_tokens/lm_head: (4.1G -> 2.9G) ~ 29.3 smaller

src/python/py/models/builder.py
jixiongdeng: > I think this can be simplified to the following.
> 
> ```python
> if quant_method in {"rtn", "rtn_last"}:
>     if quant_method == "rtn_last":
>         customized_weight_config["/lm_head/MatMul"] = {"bits": 8}
>     int4_algo_config = RTNWeightOnlyQuantConfig(customized_weight_config=customized_weight_config)
> ```

Done.

src/python/py/models/builder.py
jixiongdeng: > I think this can be simplified to the following.
> 
> ```python
> elif quant_method in {"k_quant", "k_quant_mixed", "k_quant_last"}:
>     if quant_method != "k_quant":
>         customized_weight_config["/lm_head/MatMul"] = {"bits": 8}
> 
>     if quant_method == "k_quant_mixed":
>         # k_quant_mixed is from llama.cpp.
>         # Reference: https://github.com/ggml-org/llama.cpp/blob/36667c8edcded08063ed51c7d57e9e086bbfc903/src/llama-quant.cpp#L136
>         # We also consider some MatMuls are more senstive to quantization than other MatMuls.
>         layers_to_exclude = [
>             i
>             for i in range(self.num_layers)
>             if i < self.num_layers / 8 or i >= 7 * self.num_layers / 8 or (i - (round)(self.num_layers / 8)) % 3 == 2
>         ]
>         for i in layers_to_exclude:
>             customized_weight_config["/model/layers." + str(i) + "/attn/qkv_proj/MatMul"] = {"bits": 8}
>             customized_weight_config["/model/layers." + str(i) + "/attn/v_proj/MatMul"] = {"bits": 8}
>             customized_weight_config["/model/layers." + str(i) + "/mlp/down_proj/MatMul"] = {"bits": 8}
> 
>     int4_algo_config = KQuantWeightOnlyQuantConfig(customized_weight_config=customized_weight_config)
> ```

Done.

src/python/py/models/builder.py
jixiongdeng: > This is an optimization opportunity that should be auto-detected by the model builder. We should not need to give the responsibility to the user. You can see the review comments on [this PR](https://github.com/microsoft/onnxruntime-genai/pull/1853) for more details.

Thanks for the details. We attempt to deliver fine tuned weights on open sources models like llama3.2 and qwen3. These models are not originally with packed `qkv_proj`. We fine tuned these models with unpacked projs which might be better to deliver these weights with consistency of their original forms. The unpack option is False at default until we add it in `extra_options`. It would be great if we have this option for development.

src/python/py/models/builder.py
jixiongdeng: > Can we rewrite the above section and the if condition to just match on the conditions needed for tied embeddings to be true and otherwise set it to false?
> 
> Something like this:
> 
> ```python
> self.int8_lm_head = extra_options.get("int4_algo_config", "default") in {"k_quant_mixed", "k_quant_last", "rtn_last"}
> self.int4_tied_embeddings = extra_options.get("int4_tied_embeddings", config.tie_word_embeddings if hasattr(config, "tie_word_embeddings") and config.tie_word_embeddings is not None else False)
> 
> # matmul_nbits_quantizer.py has a different naming for default quantization, so lm_head.MatMul.weight_Q{}G{} does not match.
> # tied_embeddings lm_head.MatMul.weight_Q{}G{} only works with rtn&k_quant on 4bit
> self.int4_tied_embeddings = <boolean expression>
> ```

Done.

jixiongdeng:
> Can you update the options for `int4_algo_config` here and add their descriptions?
> 
> https://github.com/microsoft/onnxruntime-genai/blob/d4eabac69d919ddf4d499940866dab695058a0f7/src/python/py/models/builder.py#L4685-L4688

Added. Done.

src/python/py/models/builder.py
jixiongdeng: > Let's move this import up. It was previously here because it was not part of a stable release.
> 
> https://github.com/microsoft/onnxruntime-genai/blob/d4eabac69d919ddf4d499940866dab695058a0f7/src/python/py/models/builder.py#L24-L28

The current build checks are still using ort 1.22 which will block this PR if I move the import on the top. Would be better to change it after updating check tests.

src/python/py/models/builder.py
```diff
@@ -283,12 +283,21 @@ def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
...
-        self.int4_tied_embeddings = extra_options.get("int4_tied_embeddings", self.int4_tied_embeddings)
-        self.int8_lm_head = extra_options.get("int4_algo_config", "default") in {"k_quant_mixed", "k_quant_last"}
-        if not self.int8_lm_head:
-            # matmul_nbits_quantizer.py has a different naming for default quantization, so lm_head.MatMul.weight_Q{}G{} does not match.
-            self.int4_tied_embeddings = False
+
+        lm_head_excluded = "/lm_head/MatMul" in self.quant_attrs["int4"]["nodes_to_exclude"]
+
+        self.int4_tied_embeddings = extra_options.get("int4_tied_embeddings", config.tie_word_embeddings if hasattr(config, "tie_word_embeddings") and config.tie_word_embeddings is not None else False)
+        self.int8_lm_head = extra_options.get("int4_algo_config", "default") in {"k_quant_mixed", "k_quant_last", "rtn_last"}
```
tianleiwu, line 290: why int4_algo_config is used to set int8_lm_head here? Our model support different bits for different weights. I thought that it is better to have straight forward setting like weight name to n_bits, or a configuration for lm_head.

src/python/py/models/builder.py
```diff
@@ -4660,11 +4686,20 @@ def get_args():
...
+                    k_quant = k_quant algorithm for int4 quantization.
                     k_quant_mixed = k_quant algorithm with mixed precision (int4 + int8).
                     k_quant_last = k_quant algorithm where only the last MatMul (/lm_head/MatMul) is quantized as int8. Other MatMuls are quantized as int4.
-                int4_tied_embeddings = Enable weight sharing for quantization. Default is false.
-                    Use this option when you want to share the weights in the embedding and unembedding.
+                int4_tied_embeddings = Enable weight sharing for quantized models (INT4/UINT4/INT8/UINT8). Default is false.
+                    Use this option when you want to share the quantized weights between embedding and LM head layers.
+                    Only works with rtn and k_quant quantization algorithms.
+                    Cannot be used if LM head is excluded from quantization (use shared_embeddings instead).
+                shared_embeddings = Enable weight sharing for FP16/FP32/BF16 weights. Default is false.
```
tianleiwu: Why not export model with shared lm_head by default? Is there some specified use case that clone weights help perf/accuracy?

src/python/py/models/builder.py
```diff
@@ -4660,11 +4686,20 @@ def get_args():
...
-                    Currently supported options are: 'default', 'rtn', 'k_quant_mixed', 'k_quant_last'.
+                    Currently supported options are: 'default', 'rtn', 'rtn_last', 'k_quant', 'k_quant_mixed', 'k_quant_last'.
+                    rtn = RTN algorithm for int4 quantization.
+                    rtn_last = RTN algorithm where only the last MatMul (/lm_head/MatMul) is quantized as int8. Other MatMuls are quantized as int4.
+                    k_quant = k_quant algorithm for int4 quantization.
                     k_quant_mixed = k_quant algorithm with mixed precision (int4 + int8).
                     k_quant_last = k_quant algorithm where only the last MatMul (/lm_head/MatMul) is quantized as int8. Other MatMuls are quantized as int4.
-                int4_tied_embeddings = Enable weight sharing for quantization. Default is false.
-                    Use this option when you want to share the weights in the embedding and unembedding.
+                int4_tied_embeddings = Enable weight sharing for quantized models (INT4/UINT4/INT8/UINT8). Default is false.
```
tianleiwu, line 4695: Do we need this option? 

If we shared embeddings (lm_head), that also means the quantized weights shall be shared.

As long as we know quantization method and number of bits, that will be enough.

jixiongdeng:
Closed this PR due to refactor of model builder. 
Avoid massive unnecessary rebase. 
New PR at: https://github.com/microsoft/onnxruntime-genai/pull/1885

PR #1857, opened 2025-11-07
whisper: Support a variant of the whisper pipeline where encoder / decoder are stateful.

RyanMetcalfeInt8:
This will mainly be used to support EPCtx-wrapped OpenVINO IR's -- the same set of models that are compatible with OpenVINO GenAI.

In this case, unlike the default pipeline, the encoder doesn't pull in KV cross projection -- it only outputs encoder_hidden_states.

For the decoder portion of the pipeline, the cross & self KV cache tensors are managed internal to the model (attached to the session).

### DIFF
--- a/src/models/whisper.cpp
+++ b/src/models/whisper.cpp
@@ -346,10 +346,11 @@ DeviceSpan<float> WhisperState::Run(int current_length, DeviceSpan<int32_t>& nex
     // Otherwise the GetOutput(present_key_{self/cross}_{i}) method returns transposed K caches.
     TransposeKCaches(cross_cache_->GetValues());
 
-    // Only transpose caches if kv_cache_ is a DefaultKeyValueCache
-    if (auto default_kv_cache_ptr = dynamic_cast<DefaultKeyValueCache*>(decoder_state_->kv_cache_.get())) {
-      TransposeKCaches(default_kv_cache_ptr->GetPresents());
+    auto default_kv_cache_ptr = dynamic_cast<DefaultKeyValueCache*>(decoder_state_->kv_cache_.get());
+    if (!default_kv_cache_ptr) {
+      throw std::runtime_error("Unable to convert KeyValueCache to DefaultKeyValueCache");
     }
+    TransposeKCaches(default_kv_cache_ptr->GetPresents());
   }
 
   // Update inputs and outputs for decoder


src/models/whisper.cpp
```diff
@@ -412,4 +441,96 @@ OrtValue* WhisperState::GetOutput(const char* name) {
...
+  size_t new_length = next_tokens.size() / batch_size;
+  input_ids_.Update(next_tokens);
+  kv_cache_->Update(beam_indices, current_length);
+  logits_.Update(next_tokens, first_run_ ? current_length : new_length);
+}
+
+WhisperStatefulState::WhisperStatefulState(const WhisperModel& model, const GeneratorParams& params)
+    : State{params, model},
+      model_{model} {
+  encoder_state_ = std::make_unique<AudioEncoderState>(model, params);
```
kunal-vaishnavi: Can we merge these new states in the existing states instead? We can use if conditions to guard against which logic to do when.

For example, we can add `HasCrossCacheOutput` to `AudioEncoderState` to decide whether to initialize `cross_cache_` and `transpose_k_cache_buffer_` or not.

https://github.com/microsoft/onnxruntime-genai/blob/7bf81d50b102487093d85891b698ba5db33b7b6a/src/models/whisper.cpp#L173-L179

We can also add `HasCrossQKOutput` to `WhisperDecoderState` to use as needed.

RyanMetcalfeInt8: > Can we merge these new states in the existing states instead? We can use if conditions to guard against which logic to do when.

This is sort of how I started, but it seemed pretty messy having a single 'mega' state with `if`'s everywhere, rather than a couple of trimmed down, more focused state objects. Still, I can refactor if you'd like.

kunal-vaishnavi: From what I see, the main things to handle for this new pipeline are a) to not initialize any cross caches, b) instead to update the KV caches, and c) to add the encoder hidden states to the decoder. The existing code should otherwise work just fine since there already are `if` guards on the cross QKs, the alignment heads, and the cache indirection.

RyanMetcalfeInt8: Yep that's fine -- let me refactor a bit. Thanks!

RyanMetcalfeInt8: @kunal-vaishnavi -- please take a look at revised changeset.

src/models/whisper.h
```diff
@@ -12,6 +12,8 @@ namespace Generators {
 struct WhisperModel : Model {
   WhisperModel(std::unique_ptr<Config> config, OrtEnv& ort_env);
 
+  bool EncoderHasCrossPresentKVOutputs() const;
```
kunal-vaishnavi: Can we move this inside the `AudioEncoderState` and define it as follows (feel free to add/remove `const` as needed)?

```cpp
bool HasCrossKVCacheOutputs() { return model_.session_info_.HasOutput(ComposeKeyValueName(model_.config_->model.encoder.outputs.cross_present_key_names, 0)); }
```

Then we can use it as follows inside `WhisperState`.

```cpp
if (encoder_state_->HasCrossKVCacheOutputs()) {
   // perform logic as needed
}
```

Here's how similar methods for optional inputs in the text decoder are handled.

https://github.com/microsoft/onnxruntime-genai/blob/fdb7fae86b9a421b6acad600da5a748dcbe62c8c/src/models/whisper.h#L54-L56

src/models/whisper.cpp
```diff
@@ -337,7 +361,11 @@ DeviceSpan<float> WhisperState::Run(int current_length, DeviceSpan<int32_t>& nex
     // Transpose the K caches only when the else branch is run for the first time.
     // Otherwise the GetOutput(present_key_{self/cross}_{i}) method returns transposed K caches.
     TransposeKCaches(cross_cache_->GetValues());
-    TransposeKCaches(decoder_state_->kv_cache_.GetPresents());
+
+    // Only transpose caches if kv_cache_ is a DefaultKeyValueCache
+    if (auto default_kv_cache_ptr = dynamic_cast<DefaultKeyValueCache*>(decoder_state_->kv_cache_.get())) {
```
kunal-vaishnavi: Why is this condition needed? The transposing of the KV caches is guarded with an earlier condition to check that the model uses `DecoderMaskedMHA`, which should already be false for your scenario. Can we do the cast directly?

https://github.com/microsoft/onnxruntime-genai/blob/fdb7fae86b9a421b6acad600da5a748dcbe62c8c/src/models/whisper.cpp#L360-L369

RyanMetcalfeInt8: This is just a downcast and checking that resultant ptr is valid. I suppose it might make sense to check ptr ahead of time and throw exception if the downcast fails (as it really shouldn't), rather than using it as a condition for this if statement. Let me change that.

src/models/whisper.h
RyanMetcalfeInt8: done.

src/models/whisper.cpp
RyanMetcalfeInt8: Okay, please take a look at latest changes.

kunal-vaishnavi (approved):

RyanMetcalfeInt8:
@kunal-vaishnavi -- I rebased against latest main that included MacOS fixes, and pipeline is passing now.

PR #1858, opened 2025-11-09
Enable GENAI with `FetchContent` or `find_package()` for client projects

apwojcik:
The PR changes the onnxruntime-genai CMake files so client projects can use it via FetchContent or find_package() CMake commands.

To accomplish that
* the `project()` needs version configured correctly and up to date,
* the `onnxruntime-genai` target needs a properly configured `INTERFACE` when installing and building,
* correctly place binaries and files needed for development in the designated directories,
* export required information and generate config package scripts (CMakeConfigPackageHelpers),

With those changes, it is enough to use `find_package()` and `target_link_libraries()` commands in a client project to configure it with the onnxruntime-genai. Please take a look at `examples\c\CMakeLists.txt` for details, it has been updated accordingly.

A few changes were required for `--buld_csharp` to accommodate changes from this PR. In addition, the ability to build the `Release` NuGet package for C# has been added.

A minor optimization for MultiConfig generators is that build.py generates a single build directory rather than one per build type.


requirements.txt
```diff
@@ -0,0 +1 @@
+requests
```
kunal-vaishnavi: This file should be removed. Any packages needed for testing purposes can go in `test/python/requirements.txt` instead.

apwojcik: The intention of the file is not to enable tests. It is required to run `build.py` successfully. The `tools/python/util/dependency_resolver.py` depends on `requests`, a Python module not included in the default installation process.
Please see https://github.com/microsoft/onnxruntime-genai/blob/589e09937e8e22e57ced20a65b565c622cc6047f/tools/python/util/dependency_resolver.py#L12

kunal-vaishnavi: The `requests` library is used purely as a convenience library for downloading the necessary assets that will be used in the build process via `build.py`. The library itself is not required during the build process. We could change this to use a built-in Python library instead and remove this `requirements.txt` file.

apwojcik: Understood. I removed the file from the PR.

cmake/ortlib.cmake
```diff
@@ -65,15 +85,45 @@ if(ORT_HOME)
...
     set(ORT_HEADER_DIR ${ORT_HOME}/headers)
     set(ORT_LIB_DIR ${ORT_HOME}/jni/${ANDROID_ABI})
+    set(ORT_BIN_DIR ${ORT_LIB_DIR})
   elseif (IOS OR MAC_CATALYST)
     set(ORT_HEADER_DIR ${ORT_HOME}/Headers)
     set(ORT_LIB_DIR ${ORT_HOME}/)
+    set(ORT_BIN_DIR ${ORT_LIB_DIR})
   elseif (CMAKE_SYSTEM_NAME MATCHES "AIX")
     set(ORT_HEADER_DIR ${ORT_HOME}/include/onnxruntime)
+    set(ORT_BIN_DIR ${ORT_HOME}/bin)
```
baijumeswani, line 95: On AIX, is the ort bin dir inside ${ORT_HOME}/bin?

cmake/ortlib.cmake
```diff
@@ -65,15 +85,45 @@ if(ORT_HOME)
...
+    else()
+      find_file(_onnxruntime_c_api_h onnxruntime_c_api.h REQUIRED
+              HINTS "${ORT_HOME}" PATH_SUFFIXES include include/onnxruntime)
+      if(CMAKE_SYSTEM_NAME STREQUAL "Windows")
+        find_file(_onnxruntime_dll onnxruntime.dll REQUIRED
+                HINTS "${ORT_HOME}" PATH_SUFFIXES bin lib)
+        find_file(_onnxruntime_lib onnxruntime.lib REQUIRED
+                HINTS "${ORT_HOME}" PATH_SUFFIXES lib)
+      else()
+        find_file(_onnxruntime_dll ${CMAKE_SHARED_LIBRARY_PREFIX}onnxruntime${CMAKE_SHARED_LIBRARY_SUFFIX} REQUIRED
```
baijumeswani: Can we rename _onnxruntime_dll to avoid the dll suffix?

cmake/ortlib.cmake
```diff
@@ -147,6 +197,7 @@ else()
       message(FATAL_ERROR "Auto download ONNX Runtime for this platform is not supported.")
     endif()
   endif()
+  set(ORT_BIN_DIR ${ORT_LIB_DIR})
```
baijumeswani, line 200: We've mostly had the lib and bin dir be the same in most scenarios. Why do we need the distinction now?

.gitignore
```diff
@@ -8,13 +8,15 @@
 *obj/
 src/csharp/AssemblyInfo.cs
 /ort
-/build
+build*/
 /build_*
+build*/
```
baijumeswani, line 12: duplicate?

cmake/package.cmake
```diff
@@ -6,25 +7,38 @@ set_target_properties(
...
-  onnxruntime-genai
-  LIBRARY DESTINATION lib
-  RUNTIME DESTINATION lib
-  ARCHIVE DESTINATION lib
-  PUBLIC_HEADER DESTINATION include
+
+install(TARGETS onnxruntime-genai
+  EXPORT onnxruntime-genaiTargets
+  LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
+  RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR}
```
baijumeswani, line 14: Would this put the binary in bin instead of lib? Is that intentional? It breaks existing expectation and will have consequences on our packaging pipelines as well.

CMakeLists.txt
```diff
@@ -158,23 +161,36 @@ find_package(Threads REQUIRED)
...
+  target_compile_definitions(onnxruntime-genai PRIVATE VERSION_PATCH=${PROJECT_VERSION_PATCH})
   target_compile_definitions(onnxruntime-genai PRIVATE VERSION_SUFFIX=${VERSION_SUFFIX})
   target_compile_definitions(onnxruntime-genai PRIVATE FILE_NAME=\"onnxruntime-genai.dll\")
 else()
   add_library(onnxruntime-genai SHARED ${generator_srcs})
+  set_target_properties(onnxruntime-genai PROPERTIES VERSION ${PROJECT_VERSION} SOVERSION ${PROJECT_VERSION_MAJOR})
 endif()
 
+target_include_directories(onnxruntime-genai INTERFACE
+        $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/src> $<INSTALL_INTERFACE:include>)
```
baijumeswani, line 175: Will this leak all internal headers to consumers?

build.py
```diff
@@ -777,13 +788,37 @@ def build_examples(args: argparse.Namespace, env: dict[str, str]):
...
+    else:
+        ortlib_src_dir = args.build_dir / "_deps" / "ortlib-src"
+        ort_include_dir = REPO_ROOT / "ort" / "include"
+        if not ort_include_dir.exists():
+            ort_include_dir = ortlib_src_dir / "build" / "native" / "include"
+        ortlib_runtimes_dir = ortlib_src_dir / "runtimes"
+        ort_lib_dir = REPO_ROOT / "ort" / "lib"
+        if not ort_lib_dir.exists():
+            if util.is_windows():
+                ort_lib_dir = ortlib_runtimes_dir / "win-x64"
```
baijumeswani: Implicit assumption that is_windows() is always win-x64? Haven't looked at is_windows(), does it only return true if it is win-x64?

build.py
```diff
@@ -777,13 +788,37 @@ def build_examples(args: argparse.Namespace, env: dict[str, str]):
...
+            elif util.is_windows_arm() or (util.is_windows() and (args.arm64 or args.arm64ec)):
+                ort_lib_dir = ortlib_runtimes_dir / "win-arm64"
+            elif args.android:
+                ort_lib_dir = ortlib_runtimes_dir / "android"
+            elif util.is_linux():
```
baijumeswani: Same here.

build.py
```diff
@@ -777,13 +788,37 @@ def build_examples(args: argparse.Namespace, env: dict[str, str]):
...
+                ort_lib_dir = ortlib_runtimes_dir / "linux-x64"
+            elif util.is_linux_arm() or (util.is_linux() and (args.arm64 or args.arm64ec)):
+                ort_lib_dir = ortlib_runtimes_dir / "linux-arm64"
+            elif util.is_mac() or args.macos:
+                ort_lib_dir = ortlib_runtimes_dir / "osx-arm64"
+            elif args.ios:
+                ort_lib_dir = ortlib_runtimes_dir / "ios"
+            elif not util.is_aix():
+                raise RuntimeError("Unsupported operating system to build examples for")
```
baijumeswani: Can this error be cleaner (explicitly mention the build platform)? Any reason to not build here?

cmake/ortlib.cmake
```diff
@@ -125,7 +175,7 @@ else()
   else()
     set(ORT_BINARY_PLATFORM "x64")
     if (APPLE)
-      if(CMAKE_OSX_ARCHITECTURES STREQUAL "arm64")
+      if(CMAKE_SYSTEM_PROCESSOR STREQUAL "arm64" OR CMAKE_SYSTEM_PROCESSOR STREQUAL "ARM64")
```
baijumeswani, line 178: Any reason for this change?

baijumeswani (dismissed):

kunal-vaishnavi:
You will need to sync with the main branch to get [this CI fix](https://github.com/microsoft/onnxruntime-genai/pull/2122).

cmake/ortlib.cmake
apwojcik: I am not sure. I don't have access to AIX systems. If there's no bin directory, I will change it to LIB. I must have made that change because CI was failing for AIX (if any)?

cmake/ortlib.cmake
apwojcik: It is required for ONNX Runtime installed with the `cmake --install ...` command. On Windows, the command puts .dlls/.exe in the bin directory. That is expected for this type of installation.

cmake/package.cmake
apwojcik: The command will put the .so files in the lib directory. On Windows, it will put .dll/.exe files in the bin directory, which is expected behavior with `cmake --install ...`.

CMakeLists.txt
apwojcik: No, only the headers marked public and interface.

build.py
apwojcik: The change is needed for CI to pass when building examples, and the ONNX Runtime is extracted from a .nuget package. There's no Windows 32bit variant anymore. In the .nuget package, there are only win-x64 and win-arm64 subdirectories.

But you're right, the if will consume both win-x64 and win-arm64 and hardcode the path for win-x64. I will make win-arm64 the first to be checked.

build.py
apwojcik: The change is to make CI pass with the ONNX Runtime extracted from the .nuget package. When used with `find_package()`, CMake will automatically configure all directories for building examples.

build.py
apwojcik: This part is only when ONNX Runtime has been extracted from the .nuget package. There are no ONNXRuntime libraries for AIX in the .nuget package. The script is not blocking on building AIX here. It only blocks on unknown operating systems. The `find_package()` branch above will build examples for any operating system.

cmake/ortlib.cmake
apwojcik: The variable `CMAKE_OSX_ARCHITECTURE` returns an empty string on macOS Tahoe and CMake 4+. `CMAKE_SYSTEM_PROCESSOR` always returns the requested value for any macOS, OSX, or iOS target and CMake version.

CMakeLists.txt
apwojcik: Those are all the files installed on Windows with the `cmake --install ...` command. I anonymized the actual installation path with `...`. The command installs similar files on other operating systems, except that the .so files are located in the `lib` directory rather than `bin`.
```
❯ cmake --install .\build\Release\ --prefix C:\...\onnxruntime-genai\Release --config Release
-- Installing: C:/.../onnxruntime-genai/Release/lib/onnxruntime-genai.lib
-- Installing: C:/.../onnxruntime-genai/Release/bin/onnxruntime-genai.dll
-- Installing: C:/.../onnxruntime-genai/Release/include/ort_genai_c.h
-- Installing: C:/.../onnxruntime-genai/Release/include/ort_genai.h
-- Installing: C:/.../onnxruntime-genai/Release/lib/cmake/onnxruntime-genai/onnxruntime-genai-targets.cmake
-- Installing: C:/.../onnxruntime-genai/Release/lib/cmake/onnxruntime-genai/onnxruntime-genai-targets-release.cmake
-- Installing: C:/.../onnxruntime-genai/Release/./README.md
-- Installing: C:/.../onnxruntime-genai/Release/./ThirdPartyNotices.txt
-- Installing: C:/.../onnxruntime-genai/Release/./SECURITY.md
-- Installing: C:/.../onnxruntime-genai/Release/./LICENSE
-- Installing: C:/.../onnxruntime-genai/Release/lib/cmake/onnxruntime-genai/onnxruntime-genai-config.cmake
-- Installing: C:/.../onnxruntime-genai/Release/lib/cmake/onnxruntime-genai/onnxruntime-genai-config-version.cmake
```

kunal-vaishnavi:
Closing this PR in favor of https://github.com/microsoft/onnxruntime-genai/pull/2310. Please add any comments or concerns on the new PR.

PR #1861, opened 2025-11-11
Fix gpt-oss model export

apsonawane:
This pull request updates the logic for handling the `block_size` attribute in QMoE (Quantized Mixture of Experts) model building and quantization. The changes ensure that block-wise quantization is only used when explicitly specified, defaulting to tensor-level quantization otherwise. The most important changes are:

**Quantization logic updates:**

* In `make_qmoe_weights`, block-wise quantization is now only used if `int4_block_size` is explicitly present in `extra_options`; otherwise, tensor-level quantization is used by default. The `block_size` attribute in `moe_attrs` is set accordingly.

**Operator construction improvements:**

* In `make_qmoe_op`, the `block_size` attribute is only included in the operator's attributes if it was explicitly set in `moe_attrs`, preventing unnecessary or default values from being passed.
* The direct passing of `block_size` as a parameter to `make_node` is removed; it is now only included via `extra_kwargs` when appropriate.

### DIFF
--- a/src/python/py/models/builders/base.py
+++ b/src/python/py/models/builders/base.py
@@ -359,10 +359,10 @@ class Model:
         int4_algo_config = self.make_int4_algo_config(extra_options.get("int4_algo_config", "default"))
         self.int4_block_size = extra_options.get("int4_block_size", 32)
         
-        # Validate that CUDA EP doesn't use int4_block_size for QMoE
-        if self.ep == "cuda" and "int4_block_size" in extra_options and moe_op_type == "QMoE":
+        # Validate that only CPU and WebGPU EPs support int4_block_size for QMoE
+        if self.ep not in ["cpu", "webgpu"] and "int4_block_size" in extra_options and moe_op_type == "QMoE":
             raise ValueError(
-                "The 'int4_block_size' option is not supported for CUDA execution provider with QMoE. "
+                f"The 'int4_block_size' option is not supported for {self.ep} execution provider with QMoE. "
                 "Block-wise quantization (block_size attribute) is only supported for CPU and WebGPU execution providers."
             )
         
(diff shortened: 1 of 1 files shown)


LorenRd:
Hi @apsonawane @kunal-vaishnavi not sure if it's part of the scope of this PR but after building from this branch, exporting with DML int4 I tried to run it but it fails

```
onnxruntime-directml                     1.23.0
onnxruntime-genai-directml               0.11.0.dev0 C:\Users\ailab\Desktop\onnxruntime-genai\build\Windows\Release\wheel
```

```
RuntimeError: Load model from C:\Users\ailab\Desktop\xllm_lib\artifacts\models\gpt_oss_20b_onnx_dml_int4\model.onnx failed:Type Error: Type parameter (T) of Optype (SkipSimplifiedLayerNormalization) bound to different types (tensor(float16) and tensor(float) in node (/model/layers.1/input_layernorm/SkipLayerNorm).
```

src/python/py/models/builder.py
```diff
@@ -2596,20 +2456,22 @@ def make_qmoe_weights(self, weights):
         dtype = torch.quint4x2 if self.moe_attrs["expert_weight_bits"] == 4 else torch.int8
         qweight, scales = None, None
 
-        # Get block size from quantization attributes
-        block_size = self.quant_attrs["int4"]["block_size"]
+        # For QMoE, only use block-wise quantization when explicitly requested
+        # via int4_block_size and when using the CPU execution provider, since
+        # block_size is only supported for CPU EP in the QMoE operator.
+        use_blockwise_quant = "int4_block_size" in self.extra_options and self.ep == "cpu"
```
tianleiwu: webgpu might also support block_size:
https://github.com/microsoft/onnxruntime/blob/4665804592d1ae341094b93c0717b1c1c6656559/onnxruntime/contrib_ops/webgpu/moe/qmoe.cc#L185

Suggest to throw exception when ep is "cuda" and int4_block_size is specified.

apsonawane: Updated

tianleiwu (approved):

PR #1862, opened 2025-11-11
Model builder refactoring

tianleiwu:
Instead of having all models in one huge py file, this splits them to separated files based on model group like llama.py, phi.py etc;.

### New Directory Structure

The new structure will look like this:
```
|-- builder.py              # (Modified) Main script: arg parsing, create_model factory, main
|-- __init__.py             # (New) 
|-- builders/               # (New) Package for all specific model implementations
|   |-- __init__.py         # (New) Exports all models for easy import
|   |-- base.py             # (New) Contains the large base `Model` class
|   |-- llama.py            # (New) Contains LlamaModel
|   |-- mistral.py          # (New) Contains MistralModel
|   |-- qwen.py             # (New) Contains QwenModel, Qwen3Model
|   |-- phi.py              # (New) Contains all Phi model variants
|   |-- gemma.py            # (New) Contains all Gemma model variants
|   |-- chatglm.py          # (New) Contains ChatGLMModel
|   |-- nemotron.py         # (New) Contains NemotronModel
|   |-- olmo.py             # (New) Contains OLMoModel
|   |-- granite.py          # (New) Contains GraniteModel
|   |-- ernie.py            # (New) Contains ErnieModel
|   |-- smollm.py           # (New) Contains SmolLM3Model
|   |-- gptoss.py           # (New) Contains GPTOSSModel
```


apsonawane:
The command remains the same right?
Also, Should we update the Readme with this directory structure so that everyone is aware of this?

tianleiwu:
> The command remains the same right? Also, Should we update the Readme with this directory structure so that everyone is aware of this?

Yes. Same command can still work. 
The directory structure is straight-forward and align with supported model in readme.

apsonawane (approved):

PR #1864, opened 2025-11-12
Fix continuous decoding bug

apsonawane:
This pull request updates the token appending logic in the `GreedySearch_Cpu` search strategies to always append generated tokens to output sequences, even when the end-of-sequence (EOS) token is encountered. This change ensures that EOS tokens are consistently included in the final output, and relies on the generation loop's `IsDone()` check to halt further processing.

### DIFF
--- a/src/search.cpp
+++ b/src/search.cpp
@@ -155,8 +155,10 @@ void GreedySearch_Cpu::SelectTop() {
     SetNextToken(batch_id, token);
   }
 
-  if (!done_)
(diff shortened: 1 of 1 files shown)


kunal-vaishnavi:
Baiju and I discussed this a while back. The consensus was that the EOS token id should not be part of the sequences for continuous decoding. There is a [PR](https://github.com/microsoft/onnxruntime-genai/pull/1849) to fix all examples that will use this change.

PR #1870, opened 2025-11-13
Fix guidance usage in continuous decoding

kunal-vaishnavi:
### Description

This PR resets the guidance logits processor when generation is completed and fixes the last action that occurred.

### Motivation and Context

This PR fixes a bug with providing grammars in multi-turn conversations where the generator is re-used.


JC1DA (approved):
LGTM

PR #1882, opened 2025-11-19
Add model builder for Qwen2_5_VLTextModel

tianleiwu:
Add model builder for Qwen2_5_VLTextModel.

Snippet of example test output:
```
onnxruntime-genai/test/python/models/qwen_2.5_vl$ bash run.sh fp32

...
--- Comparing Decode Logits ---
Logits: PASS

--- Comparing Decode KV Cache ---
KV Cache (all 72 tensors): PASS

✅ Decode Parity Test Passed!

==============================
🎉 All Parity Tests Passed! 🎉
==============================
--- fp32 run complete ---
```

### DIFF
--- a/src/python/py/models/builders/base.py
+++ b/src/python/py/models/builders/base.py
@@ -498,11 +498,14 @@ class Model:
 
             # Some EPs don't support packed Q/K/V for GQA yet
             # Packed MatMul with LoRA/QLoRA is not currently supported
+            # use_packed_matmul can be overrided by upstream quantization choice
+            # (e.g., when q_proj, k_proj, v_proj have different quantization settings)
             self.attention_attrs["use_packed_matmul"] = (
                 self.ep not in ["dml"]
                 and not self.matmul_attrs["use_lora"]
                 and not self.attention_attrs["q_norm"]
                 and not self.attention_attrs["k_norm"]
+                and not self.extra_options.get("disable_qkv_fusion", False)
             )
 
             # Some EPs don't support fusing rotary embeddings inside GQA yet
--- a/.github/workflows/mac-cpu-arm64-build.yml
+++ b/.github/workflows/mac-cpu-arm64-build.yml
@@ -47,6 +47,27 @@ jobs:
         with:
           gradle-version: '8.6'
 
+      - name: Setup .NET 9.0
+        uses: actions/setup-dotnet@v5
+        with:
+          dotnet-version: '9.0.x'
+          dotnet-quality: 'ga'
+
+      - name: Create global.json to lock .NET SDK version
+        run: |
+          echo '{
+            "sdk": {
+              "version": "9.0.306",
+              "rollForward": "latestFeature"
+            }
+          }' > global.json
+          cat global.json
+
+      - name: Verify .NET version
+        run: |
+          dotnet --version
+          dotnet --list-sdks
+
       - name: Get the Latest OnnxRuntime Nightly Version
         run: |
           ORT_NIGHTLY_VERSION=$(curl -s "${{ env.ORT_NIGHTLY_REST_API }}" | jq -r '.value[0].versions[0].normalizedVersion')
@@ -135,3 +156,4 @@ jobs:
           export ORTGENAI_LOG_ORT_LIB=1
           export DYLD_LIBRARY_PATH=$DYLD_LIBRARY_PATH:$GITHUB_WORKSPACE/build/cpu/osx-arm64
           ./build/cpu/osx-arm64/unit_tests
+ 
--- a/src/models/audio_features.cpp
+++ b/src/models/audio_features.cpp
@@ -12,7 +12,7 @@ AudioFeatures::AudioFeatures(State& state, const std::string& name, const std::v
       name_{name} {
   // Get audio features
   for (const auto& [input_name, value] : extra_inputs) {
-    if (input_name == Config::Defaults::AudioFeaturesName) {
+    if (input_name == name) {
       audio_features_ = model_.ExpandInputs(value->ort_tensor_, state_.params_->search.num_beams);
     }
   }
--- a/src/models/whisper.cpp
+++ b/src/models/whisper.cpp
@@ -54,10 +54,11 @@ DeviceSpan<float> AudioEncoderState::Run(int current_length, DeviceSpan<int32_t>
 WhisperDecoderState::WhisperDecoderState(const WhisperModel& model, const GeneratorParams& params, const int num_frames)
     : State{params, model},
       model_{model},
+      kv_cache_(CreateKeyValueCache(*this)),
       num_frames_{num_frames} {
   input_ids_.Add();
   logits_.Add();
-  kv_cache_.Add();
+  kv_cache_->Add();
 
   // Add past sequence length
   if (HasPastSequenceLengthInput()) {
@@ -117,7 +118,7 @@ void WhisperDecoderState::UpdateInputsOutputs(DeviceSpan<int32_t>& next_tokens, 
   int batch_size = static_cast<int>(input_ids_.GetShape()[0]);
   size_t new_length = next_tokens.size() / batch_size;
   input_ids_.Update(next_tokens);
-  kv_cache_.Update(beam_indices, current_length);
+  kv_cache_->Update(beam_indices, current_length);
   logits_.Update(next_tokens, first_run_ ? current_length : new_length);
 
   // Return early if this method is just initializing the above OrtValue objects and not updating them
@@ -171,16 +172,23 @@ WhisperState::WhisperState(const WhisperModel& model, const GeneratorParams& par
     : State{params, model},
       model_{model} {
   encoder_state_ = std::make_unique<AudioEncoderState>(model, params);
-  cross_cache_ = std::make_unique<CrossCache>(*this, encoder_state_->GetNumFrames() / 2);
-  encoder_state_->AddCrossCache(cross_cache_);
   decoder_state_ = std::make_unique<WhisperDecoderState>(model, params, encoder_state_->GetNumFrames());
-  decoder_state_->AddCrossCache(cross_cache_);
 
-  transpose_k_cache_buffer_ = OrtValue::CreateTensor(model_.p_device_inputs_->GetAllocator(), cross_cache_->GetShape(), cross_cache_->GetType());
+  if (encoder_state_->HasCrossKVCacheOutputs()) {
+    cross_cache_ = std::make_unique<CrossCache>(*this, encoder_state_->GetNumFrames() / 2);
+    encoder_state_->AddCrossCache(cross_cache_);
+    decoder_state_->AddCrossCache(cross_cache_);
+    transpose_k_cache_buffer_ = OrtValue::CreateTensor(model_.p_device_inputs_->GetAllocator(), cross_cache_->GetShape(), cross_cache_->GetType());
+  }
 }
 
 void WhisperState::SetExtraInputs(const std::vector<ExtraInput>& extra_inputs) {
   encoder_state_->SetExtraInputs(extra_inputs);
+
+  if (!encoder_state_->HasCrossKVCacheOutputs()) {
+    decoder_state_->inputs_.push_back(encoder_state_->hidden_states_.get());
+    decoder_state_->input_names_.push_back(model_.config_->model.decoder.inputs.encoder_hidden_states.c_str());
+  }
 
   // Check if alignment heads input exists
   void* alignment_heads_input = nullptr;
@@ -337,7 +345,12 @@ DeviceSpan<float> WhisperState::Run(int current_length, DeviceSpan<int32_t>& nex
     // Transpose the K caches only when the else branch is run for the first time.
     // Otherwise the GetOutput(present_key_{self/cross}_{i}) method returns transposed K caches.
     TransposeKCaches(cross_cache_->GetValues());
-    TransposeKCaches(decoder_state_->kv_cache_.GetPresents());
+
+    auto default_kv_cache_ptr = dynamic_cast<DefaultKeyValueCache*>(decoder_state_->kv_cache_.get());
+    if (!default_kv_cache_ptr) {
+      throw std::runtime_error("Unable to convert KeyValueCache to DefaultKeyValueCache");
+    }
+    TransposeKCaches(default_kv_cache_ptr->GetPresents());
   }
 
   // Update inputs and outputs for decoder
--- a/src/models/whisper.h
+++ b/src/models/whisper.h
@@ -31,6 +31,8 @@ struct AudioEncoderState : State {
 
   int GetNumFrames() { return num_frames_; }
 
+  bool HasCrossKVCacheOutputs() { return model_.session_info_.HasOutput(ComposeKeyValueName(model_.config_->model.encoder.outputs.cross_present_key_names, 0)); }
+
  private:
   friend struct WhisperState;
 
@@ -62,7 +64,7 @@ struct WhisperDecoderState : State {
   const WhisperModel& model_;
 
   DefaultInputIDs input_ids_{*this};                        // Model input
-  DefaultKeyValueCache kv_cache_{*this};                    // Model input and output
+  std::unique_ptr<KeyValueCache> kv_cache_;
   
   // Inputs for beam search attention
   std::unique_ptr<OrtValue> past_sequence_length_;          // Model input
--- a/src/python/py/models/README.md
+++ b/src/python/py/models/README.md
@@ -20,6 +20,7 @@ This folder contains the model builder for quickly creating optimized and quanti
     - [Exclude Language Modeling Head](#exclude-language-modeling-head)
     - [Include Last Hidden States Output](#include-last-hidden-states-output)
     - [Enable Shared Embeddings](#enable-shared-embeddings)
+    - [Disable QKV Projections Fusion](#disable-qkv-projections-fusion)
     - [Enable CUDA Graph](#enable-cuda-graph)
     - [Use 8 Bits Quantization in QMoE](#use-8-bits-quantization-in-qmoe)
     - [Use QDQ Pattern for Quantization](#use-qdq-pattern-for-quantization)
@@ -251,6 +252,18 @@ python3 -m onnxruntime_genai.models.builder -m model_name -o path_to_output_fold
 
 # From source:
 python3 builder.py -m model_name -o path_to_output_folder -p fp16 -e cuda --extra_options shared_embeddings=true
+```
+
+#### Disable QKV Projections Fusion
+
+This scenario is for when you want to keep Q/K/V projections in the attention layer separate instead of fusing them into a single packed MatMul operation. 
+
+```
+# From wheel:
+python3 -m onnxruntime_genai.models.builder -i path_to_local_folder_on_disk -o path_to_output_folder -p precision -e execution_provider -c cache_dir_to_store_temp_files --extra_options disable_qkv_fusion=true
+
+# From source:
+python3 builder.py -i path_to_local_folder_on_disk -o path_to_output_folder -p precision -e execution_provider -c cache_dir_to_store_temp_files --extra_options disable_qkv_fusion=true
 ```
 
 #### Enable CUDA Graph
--- a/src/python/py/models/builder.py
+++ b/src/python/py/models/builder.py
@@ -62,6 +62,7 @@ def check_extra_options(kv_pairs, execution_provider):
         "use_cuda_bf16",
         "shared_embeddings",
         "hf_remote",
+        "disable_qkv_fusion",
     ]
     for key in bools:
         if key in kv_pairs:


src/python/py/models/builders/qwen.py
```diff
@@ -19,3 +26,838 @@ def make_attention_init(self):
...
         self.attention_attrs["k_norm"] = True
         super().make_attention_init()
+
+
+class Qwen25VLTextModel(Model):
+    def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
+        # We must extract the text_config for the text model's parameters
+        text_config_dict = config.text_config.to_dict()
+
+        # Update the main config with text-specific parameters
```
kunal-vaishnavi: Can we simplify this to the following instead?

https://github.com/microsoft/onnxruntime-genai/blob/da58e20df8b33e4ea8638798232720c834d798c1/src/python/py/models/builder.py#L215-L218

We can add it here when the model is created.

https://github.com/microsoft/onnxruntime-genai/blob/7ccc607767bddc2010cc56c9f984635317f3246d/src/python/py/models/builder.py#L306-L311

src/python/py/models/builders/qwen.py
```diff
@@ -19,3 +26,838 @@ def make_attention_init(self):
...
+        # builder to cast bf16 inputs -> fp32, compute LN, and cast fp32
+        # outputs -> bf16, matching the HF model and fixing both errors.
+        #
+        print("Forcing LayerNorm computation to float32 (and enabling all casts) for Qwen2.5-VL parity.")
+        self.layernorm_attrs["cast"]["use_fp32"] = True
+        self.layernorm_attrs["cast"]["root_input"] = True
+        self.layernorm_attrs["cast"]["skip_input"] = True
+        self.layernorm_attrs["cast"]["output_0"] = True
+        self.layernorm_attrs["cast"]["output_3"] = True
+        #
```
kunal-vaishnavi:
```suggestion
```

src/python/py/models/builders/qwen.py
```diff
@@ -19,3 +26,838 @@ def make_attention_init(self):
...
+        # if the io_dtype is bfloat16. This is to improve accuracy in general.
+        #
+        # PROBLEM: The HF model (Qwen2_5_VL) *does not* do this. It computes
+        # the lm_head MatMul in bfloat16 and returns bfloat16 logits.
+        # This causes the parity test (which compares bf16 vs fp32) to fail.
+        #
+        # SOLUTION: We must override the base model's decision and set the
+        # output logits type to match the io_dtype (bfloat16).
+        #
+        self.allow_bf16_logits = os.getenv("ALLOW_BF16_LOGITS") in ["1", "true", "True"]
```
kunal-vaishnavi: Can we simplify this to the following?

```py
def make_outputs_init(self):
    # Override the base model's decision and set the output logits type to match the io_dtype.
    super().make_outputs_init(self)
    self.output_types["logits"] = self.io_dtype
```

src/python/py/models/builders/qwen.py
```diff
@@ -19,3 +26,838 @@ def make_attention_init(self):
...
+        if self.allow_bf16_logits and self.io_dtype == ir.DataType.BFLOAT16:
+            print("Fixing output logits precision. Setting output_types['logits'] to BFLOAT16 to match HF model.")
+            self.output_types["logits"] = ir.DataType.BFLOAT16
+
+        # Manually get the attention_scaling from the rope_config
+        # This replicates the logic from transformers.models.rope_utils._config_to_init_values
+        rope_type = "default"
```
kunal-vaishnavi: If the model uses `mrope` for the [scaling type](https://huggingface.co/Qwen/Qwen2.5-VL-3B-Instruct/blob/main/config.json#L53), why is all of this logic needed?

src/python/py/models/builders/qwen.py
```diff
@@ -19,3 +26,838 @@ def make_attention_init(self):
...
+
+        if sum(self.mrope_splits) != self.head_size:
+            # The sum (128) should now correctly match self.head_size (128)
+            raise ValueError(
+                f"MRoPE splits {self.mrope_splits} sum ({sum(self.mrope_splits)}) does not match head size ({self.head_size})"
+            )
+
+        # Force GroupQueryAttention for fp32 cuda,
+        # as base.py's make_attention_init doesn't include this combo.
+        if self.ep == "cuda" and self.io_dtype == ir.DataType.FLOAT:
```
kunal-vaishnavi: This will create an invalid model. This combination is excluded because there is [no FP32 CUDA kernel](https://github.com/microsoft/onnxruntime/blob/81a04ca45d5c30a4cbe9ebd06c6f37d80d9caf91/onnxruntime/contrib_ops/cuda/bert/group_query_attention.cc#L35-L37) for `GroupQueryAttention` in ONNX Runtime.

src/python/py/models/builders/qwen.py
```diff
@@ -19,3 +26,838 @@ def make_attention_init(self):
...
+            self.attention_attrs["op_type"] = "GroupQueryAttention"
+            print("Forcing GroupQueryAttention (GQA) for FP32 CUDA.")
+
+        if self.attention_attrs["op_type"] != "GroupQueryAttention":
+            raise ValueError(
```
kunal-vaishnavi: Can we remove this? There already is the below logic to compute the `repeat_kv` operation in conjunction with `MultiHeadAttention` when needed.

https://github.com/microsoft/onnxruntime-genai/blob/da58e20df8b33e4ea8638798232720c834d798c1/src/python/py/models/builders/base.py#L2735-L2747

src/python/py/models/builders/qwen.py
```diff
@@ -19,3 +26,838 @@ def make_attention_init(self):
...
+
+        # Call the base Model's make_inputs_and_outputs (skipping MistralModel's)
+        super().make_inputs_and_outputs()
+
+    def make_dynamic_rope_caches(self, layer_id, basename):
+        """
+        Re-implements Qwen2_5_VLRotaryEmbedding.forward using ONNX ops.
+        Takes 3D position_ids and inv_freq and dynamically creates
+        the cos/sin caches.
+        """
```
kunal-vaishnavi: Can you draw what the ONNX subgraph looks like here so that it is easier to visualize?

src/python/py/models/builders/qwen.py
```diff
@@ -19,3 +26,838 @@ def make_attention_init(self):
...
+        pos_ids_name = "position_ids"
+        inv_freq_name = "model.inv_freq"
+        head_dim_half = self.head_size // 2
+
+        # Get Batch Size from position_ids.shape[1]
+        shape_pos_ids_name = f"{basename}/shape_pos_ids"
```
kunal-vaishnavi: Can we follow the namespaces format when defining node names?

https://github.com/microsoft/onnxruntime-genai/blob/da58e20df8b33e4ea8638798232720c834d798c1/src/python/py/models/DESIGN.md?plain=1#L81-L121

src/python/py/models/builders/qwen.py
```diff
@@ -19,3 +26,838 @@ def make_attention_init(self):
...
+        self.make_concat(concat_name, [neg_output, split_output_0], compute_dtype, x_shape, axis=-1)
+
+        return concat_output
+
+    def apply_mrope_rotation(self, layer_id, q_or_k_path, q_or_k_shape, dyn_cos, dyn_sin, num_heads, basename):
+        """
+        Re-implements apply_multimodal_rotary_pos_emb using ONNX ops.
+        Takes Q/K tensor and the dynamically generated 3D caches
+        and applies the rotation.
+        """
```
kunal-vaishnavi: Can you draw what the ONNX subgraph looks like here so that it is easier to visualize?

src/python/py/models/builders/qwen.py
tianleiwu: The model can support yarn if you change the default configuration:
https://huggingface.co/Qwen/Qwen2.5-VL-72B-Instruct#processing-long-texts

src/python/py/models/builders/qwen.py
tianleiwu: It's fine to keep this for parity test (It will fall back to CPU for fp32 GQA). 
Update the code to print some warning for it.
The production model will use fp16/bf16, so it is not an issue right now.

src/python/py/models/builders/qwen.py
tianleiwu: added.

src/python/py/models/builders/qwen.py
tianleiwu: added.

src/python/py/models/builders/qwen.py
kunal-vaishnavi: > However, it should be noted that this method has a significant impact on the performance of temporal and spatial localization tasks, and is therefore not recommended for use.

Since it is not recommended for use, shall we remove it then?

src/python/py/models/builders/qwen.py
```diff
@@ -19,3 +25,895 @@ def make_attention_init(self):
...
+        self.make_value(neg_output, compute_dtype, half_shape)
+
+        # Concat (-x2, x1)
+        concat_name = f"{basename}/rotate_half/Concat"
+        concat_output = f"{concat_name}/output_0"
+        self.make_concat(concat_name, [neg_output, split_output_0], compute_dtype, x_shape, axis=-1)
+
+        return concat_output
+
+    def apply_mrope_rotation(self, layer_id, q_or_k_path, q_or_k_shape, dyn_cos, dyn_sin, num_heads, basename):
```
kunal-vaishnavi, line 405: In the future, can we update the rotary embedding kernels in ORT to do all this so that we don't need the ONNX subgraph?

tianleiwu: It's possible to add a new contrib op for mRope in the future.

src/python/py/models/builders/base.py
```diff
@@ -49,6 +49,7 @@ def parse_hf_token(hf_token):
 
 class Model:
     def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
+        self.config = config
```
kunal-vaishnavi: Why do we need to add the entire config here?

src/python/py/models/builders/base.py
```diff
@@ -805,7 +840,13 @@ def make_inputs_and_outputs(self):
             # Add KV cache to inputs
             key_name = f"past_key_values.{i}.key"
             key_shape = self.make_key_value_cache_shape(i, self.input_shapes["past_key_values.key"])
-            inputs.append(self.make_value(key_name, dtype=self.input_types["past_key_values.key"], shape=key_shape))
+            inputs.append(
```
kunal-vaishnavi: I wanted to add this on the lintrunner PR before it was merged but do we need to lint such examples? I think the one-line version is actually cleaner.

src/python/py/models/builders/qwen.py
```diff
@@ -19,3 +25,750 @@ def make_attention_init(self):
...
+    def make_mrope_flattened_caches(self, layer_id, dyn_cos, dyn_sin):
+        """
+        Converts the 3D MRoPE caches [3, B, S, H] into flattened, interleaved caches [B*S, H/2]
+        suitable for the RotaryEmbedding operator.
+        The logic is:
+        1. Slice dynamic caches to H/2.
+        2. Split into 3 chunks based on mrope_sections (e.g. 16, 24, 24).
+        3. Gather Temporal(0), Height(1), Width(2) specific slices for each chunk.
+        4. Concat back to H/2.
+        5. Flatten to [B*S, H/2].
```
kunal-vaishnavi: Can you draw what the ONNX subgraph looks like here so that it is easier to visualize?

src/python/py/models/builders/qwen.py
```diff
@@ -19,3 +25,750 @@ def make_attention_init(self):
...
+            self.make_cast(cast_cos_node, flat_cos, compute_dtype, ["total_token_count", self.head_size // 2])
+
+            cast_sin_node = f"{basename}/Sin/Cast"
+            rope_sin = f"{cast_sin_node}/output_0"
+            self.make_cast(cast_sin_node, flat_sin, compute_dtype, ["total_token_count", self.head_size // 2])
+
+        # 5. RotaryEmbedding Node
+        rope_node = f"{basename}/RotaryEmbedding"
+        rope_output = f"{rope_node}/output_0"
+        self.make_node(
```
kunal-vaishnavi, line 543: Can we temporarily set the values needed to leverage the below function and then unset them afterwards?

https://github.com/microsoft/onnxruntime-genai/blob/5492721ca3c2691337d3741f69f6d961e77eb7f7/src/python/py/models/builders/base.py#L1848-L1866

That way, any change to `make_rotary_embedding` will be reflected in all models that use the `RotaryEmbedding` op (for example, when a new attribute is added to the op).

src/python/py/models/builders/qwen.py
```diff
@@ -19,3 +25,750 @@ def make_attention_init(self):
...
+        #          +--------+--------+                 |
+        #                   |                          |
+        #           GroupQueryAttention <--------------+
+        #                   |
+        #                O_MatMul
+        #                   |
+        #                 O_Add
+
+        # 1. Unpack QKV if necessary (e.g. qkv_proj)
+        super().make_attention_unpacked(layer_id, attention, root_input, **kwargs)
```
kunal-vaishnavi: We should refactor the base `make_attention` method so that you don't have to copy everything over. How does this sound?

```py
    self.make_attention_input_proj(layer_id, attention, root_input, **kwargs)
    self.make_attention_qk_subgraph(layer_id, attention, root_input, **kwargs)
    # <make any repeat KV nodes>
    # <make any sinks input>
    # <make attention op>
    self.make_attention_output_proj(layer_id, attention, root_input, **kwargs)
```

Each of these new functions would have the following.

1. Input projection
```py
def make_attention_input_proj(self, layer_id, attention, root_input, **kwargs):
    # Implementation is https://github.com/microsoft/onnxruntime-genai/blob/5492721ca3c2691337d3741f69f6d961e77eb7f7/src/python/py/models/builders/base.py#L2685-L2746
```
https://github.com/microsoft/onnxruntime-genai/blob/5492721ca3c2691337d3741f69f6d961e77eb7f7/src/python/py/models/builders/base.py#L2685-L2746

2. QK subgraph
```py
def make_attention_qk_subgraph(self, layer_id, attention, root_input, **kwargs):
    # Implementation is https://github.com/microsoft/onnxruntime-genai/blob/5492721ca3c2691337d3741f69f6d961e77eb7f7/src/python/py/models/builders/base.py#L2748-L2770
```
https://github.com/microsoft/onnxruntime-genai/blob/5492721ca3c2691337d3741f69f6d961e77eb7f7/src/python/py/models/builders/base.py#L2748-L2770

3. Output projection
```py
def make_attention_output_proj(self, layer_id, attention, root_input, **kwargs):
    # Implementation is https://github.com/microsoft/onnxruntime-genai/blob/5492721ca3c2691337d3741f69f6d961e77eb7f7/src/python/py/models/builders/base.py#L2809-L2823
```
https://github.com/microsoft/onnxruntime-genai/blob/5492721ca3c2691337d3741f69f6d961e77eb7f7/src/python/py/models/builders/base.py#L2809-L2823

Then this class just has to override `make_attention_qk_subgraph` for MRoPE.

src/python/py/models/builders/qwen.py
```diff
@@ -19,3 +25,750 @@ def make_attention_init(self):
...
+    def make_model(self, input_path):
+        # Make inputs and outputs to ONNX model
+        self.make_inputs_and_outputs()
+
+        # Make pre-processing nodes
+        self.make_preprocessing_nodes()
+
+        # Load the Hugging Face model
+        print("Loading Qwen2_5_VLForConditionalGeneration model...")
+        hf_model = Qwen2_5_VLForConditionalGeneration.from_pretrained(
```
kunal-vaishnavi: Let's also do some refactoring here so you don't have to copy everything over. How does this sound?

```py
def make_model(self, input_path):
    # Make inputs and outputs to ONNX model
    self.make_inputs_and_outputs()

    # Make pre-processing nodes
    self.make_preprocessing_nodes()

    model = self.load_weights(input_path)

    # <remaining code to loop through each module>
```

The `load_weights` method can then be overridden in this file to use `Qwen2_5_VLForConditionalGeneration.from_pretrained(...).language_model` instead.

```py
def load_weights(self, input_path):
    # Implementation is https://github.com/microsoft/onnxruntime-genai/blob/5492721ca3c2691337d3741f69f6d961e77eb7f7/src/python/py/models/builders/base.py#L3671-L3724
```
https://github.com/microsoft/onnxruntime-genai/blob/5492721ca3c2691337d3741f69f6d961e77eb7f7/src/python/py/models/builders/base.py#L3671-L3724

A tiny `nn.Module` class with the right class attribute setup can be constructed to fix the paths to the modules before returning that class in `load_weights`. That way, the existing base code can be re-used without any modifications.

src/python/py/models/builders/base.py
tianleiwu: Qwen2_5_VLForConditionalGeneration.from_pretrained(
            self.model_name_or_path,
            config=self.config,  # used here in qwen.py

src/python/py/models/builders/base.py
tianleiwu: reverted

src/python/py/models/builders/qwen.py
tianleiwu: done

src/python/py/models/builders/qwen.py
tianleiwu: done

src/python/py/models/builders/qwen.py
tianleiwu: That will make things complicated. For example, we will need code like the following:

        # Temporary overrides to leverage base.make_rotary_embedding
        # 1. Mock cache creation to return our pre-computed MRoPE caches
        original_make_caches = self.make_rotary_embedding_caches
        self.make_rotary_embedding_caches = lambda: (rope_cos, rope_sin)

        # 2. Adjust num_attn_heads for K if necessary.
        #    base.make_rotary_embedding selects num_heads based on check: `if "k_rotary" in name else self.num_attn_heads`.
        #    Since our basename (e.g. `k_mrope`) does not contain `k_rotary`, it would incorrectly select `num_attn_heads` for K.
        #    We temporarily swap `num_attn_heads` with `num_kv_heads` when processing K to force the correct value.
        is_k = (num_heads == self.num_kv_heads) and (self.num_kv_heads != self.num_attn_heads)
        original_num_attn_heads = self.num_attn_heads
        if is_k:
            self.num_attn_heads = self.num_kv_heads

        try:
            # Call base method
            self.make_rotary_embedding(rope_node_name, rope_input, position_ids=pos_ids_out)
        finally:
            # Restore state
            self.make_rotary_embedding_caches = original_make_caches
            if is_k:
                self.num_attn_heads = original_num_attn_heads
 
 Using one make_node will be much simple, and I prefer that.
 
 Also, in the future, we might use a new contrib op for mrope, and RotaryEmbedding is a temp solution.

src/python/py/models/builders/base.py
kunal-vaishnavi: You should be able to remove `config=self.config` from there.

https://github.com/microsoft/onnxruntime-genai/blob/4f372984379efcc68fd6f408b8fcb7638a7c336b/src/python/py/models/builders/base.py#L3712-L3720

src/python/py/models/builders/qwen.py
```diff
@@ -19,3 +25,714 @@ def make_attention_init(self):
...
+        self.attention_attrs["use_rope_in_attn"] = False
+
+        # We need separate Q, K, V tensors to apply MRoPE manually.
+        # Packed MatMul provides a single output which would require splitting.
+        self.attention_attrs["use_packed_matmul"] = False
+
+        if "position_ids" not in self.input_names:
+            print("Re-adding 'position_ids' to self.input_names.")
+            if "attention_mask" in self.input_names:
+                idx = self.input_names.index("attention_mask")
```
kunal-vaishnavi: Can we simplify this to always use the `else` case to append `position_ids` back to the list of input names?

src/python/py/models/builders/qwen.py
```diff
@@ -19,3 +25,714 @@ def make_attention_init(self):
...
+        #       (B, S, H/2)
+        #             |
+        #          Reshape
+        #        (B*S, H/2)
+
+        basename = f"/model/layers.{layer_id}/attn/mrope_flattened_cache"
+
+        def process_cache(input_name, name_suffix):
+            # 1. Slice to H/2: [3, B, S, H] -> [3, B, S, H/2]
+            slice_name = f"{basename}/{name_suffix}/Slice_Half"
```
kunal-vaishnavi: Can we rename this to follow the naming schema?

```suggestion
            slice_name = f"{basename}/{name_suffix}/half/Slice"
```

src/python/py/models/builders/qwen.py
```diff
@@ -19,3 +25,714 @@ def make_attention_init(self):
...
+        hf_model = self.load_weights(input_path)
+
+        # We only want to export the text model
+        model = hf_model.language_model
+        print(f"Isolated language_model ({model.__class__.__name__}) for ONNX export.")
+
+        # Loop through model and map each module to ONNX/ORT ops
+        self.layer_id = 0
+
+        # The base.Model.make_model() loop expects modules from a standard causal LM,
```
kunal-vaishnavi: Can we change `load_weights` to the following to fix this?

```py
def load_weights(self, input_path):
    model = Qwen2_5_VLForConditionalGeneration.from_pretrained(
        self.model_name_or_path,
        cache_dir=self.cache_dir,
        token=self.hf_token,
        trust_remote_code=self.hf_remote,
    )

    class TextModel(torch.nn.Module):
        def __init__(self, model):
            super().__init__()
            self.model = model.language_model
            self.lm_head = model.lm_head

    return TextModel(model)
```

Then we can remove the need to rewrite `make_model`

src/python/py/models/builders/qwen.py
tianleiwu: I think so.

src/python/py/models/builders/qwen.py
tianleiwu: done

src/python/py/models/builders/qwen.py
tianleiwu: updated.

src/python/py/models/builders/base.py
tianleiwu: removed

kunal-vaishnavi (approved):

PR #1886, opened 2025-11-21
Fix bug in Squeeze for getting the value of total_seq_len

Honry:
The `Squeeze` op is used for removing single-dimensional entries from the shape of a tensor. In this node the `axes` input is set to `[0]` which would only eliminate the first axis and lead to the output shape to be `[1]` if the `batch_size` is 1.This would cause ShapeInference error at https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/core/graph/graph.cc#L111-L132 if it is not a strict mode.

This PR fixes the issue by:
-  Changing the ReduceSum output shape to [batch_size] by adding keepdims=0 attribute
-  Using ReduceMax instead of Squeeze to get the value of total_seq_len and make it as a scalar, this would cover scenarios when batch_size > 1

### DIFF
--- a/src/python/py/models/builders/base.py
+++ b/src/python/py/models/builders/base.py
@@ -911,7 +911,7 @@ class Model:
         self.make_node("Expand", inputs=inputs, outputs=[output], name=name)
         self.make_value(output, dtype, shape=shape)
 
-    def make_reduce_sum(self, name, inputs, dtype, shape, keepdims=True):
+    def make_reduce_sum(self, name, inputs, dtype, shape, keepdims=False):
         output = f"{name}/output_0"
         self.make_node("ReduceSum", inputs=inputs, outputs=[output], name=name, keepdims=keepdims)
         self.make_value(output, dtype, shape=shape)
@@ -4187,7 +4187,7 @@ class Model:
         )
         reduce_sum_name = f"{attn_mask_basename}/ReduceSum"
         reduce_sum_inputs = [f"{cast_1_name}/output_0", "/model/constants/INT64/[1]"]
-        self.make_reduce_sum(reduce_sum_name, reduce_sum_inputs, dtype=ir.DataType.INT32, shape=["batch_size"], keepdims=False)
+        self.make_reduce_sum(reduce_sum_name, reduce_sum_inputs, dtype=ir.DataType.INT32, shape=["batch_size"])
 
         # Left branch: Calculate seqlens_k = ReduceSum - 1
         sub_name = f"{attn_mask_basename}/Sub"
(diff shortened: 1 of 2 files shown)


qjia7 (approved):

qjia7 (commented):
> A better way is to let ReduceSum output shape [batch_size] by adding `keepdims=0` attribute. Then use ReduceMax to get the maximum one (assumption is the longest sequence does not have padding), or use Gather to get first item if we assume that batch_size==1 (if we only enable graph capture for one batch scenario)

Great idea! I did assume that batch_size was one when adding those code. Your suggestion looks great. @tianleiwu one more question: why expose a scalar total_seq_len since the batch can be larger than one? 
```
        #          attention_mask
        #               |
        #         Cast to int32
        #               |
        #    ReduceSum (keepdims=0)
        #              /    \
        #             /      \
        #           Sub    ReduceMax
        #            |        |
        #       seqlens_k  total_seq_len
        #         (1D)       (scalar)
```

Honry:
Thanks @tianleiwu @qjia7, I also had the concern about how to handle `batch_size > 1`.

@tianleiwu's suggestion is really a great idea! I've addressed it in the new commit, PTAL again, thanks!

src/python/py/models/builders/base.py
```diff
@@ -4220,12 +4221,12 @@ def make_attention_mask_standard_reformatting_for_gqa(self, attn_mask_basename):
...
         reduce_sum_inputs = ["attention_mask", "/model/constants/INT64/[1]"]
-        self.make_reduce_sum(reduce_sum_name, reduce_sum_inputs, dtype=ir.DataType.INT64, shape=["batch_size", 1])
+        self.make_reduce_sum(reduce_sum_name, reduce_sum_inputs, dtype=ir.DataType.INT64, shape=["batch_size"], keepdims=False)
         sub_name = f"{attn_mask_basename}/Sub"
         sub_inputs = [f"{reduce_sum_name}/output_0", "/model/constants/INT64/[1]"]
-        self.make_sub(sub_name, sub_inputs, dtype=ir.DataType.INT64, shape=["batch_size", 1])
+        self.make_sub(sub_name, sub_inputs, dtype=ir.DataType.INT64, shape=["batch_size"])
         cast_1_name = f"{attn_mask_basename}/Sub/Cast"
-        self.make_cast(cast_1_name, f"{sub_name}/output_0", dtype=ir.DataType.INT32, shape=["batch_size", 1])
+        self.make_cast(cast_1_name, f"{sub_name}/output_0", dtype=ir.DataType.INT32, shape=["batch_size"])
```
kunal-vaishnavi, line 4229: Shall we make the same update for the attention mask subgraph that creates the inputs needed for sparse attention?

https://github.com/microsoft/onnxruntime-genai/blob/2d3838211c6eb01151f84a73b0d51e1cecb81bca/src/python/py/models/builders/base.py#L4254-L4289

Honry: Make sense, `key_total_seq_lens` should be 1D too. Updated it in the new commit.

kunal-vaishnavi (approved):

qjia7 (approved):

PR #1850, opened 2025-10-30
Support pre-registered plug-in cuda execution provider library

baijumeswani:
Up until now, for the cuda execution provider, onnxruntime-genai tried using the built-in cuda ep using the legacy `OrtCUDAProviderOptionsV2` and `AppendExecutionProvider_CUDA_V2` API. These functions are designed to make use of the built-in cuda execution provider and are not compatible with the world of plug-in eps.

The changes in this pull-request now extend support to executing the model with a pre-registered (plugged-in) CUDAExecutionProvider.

In order to use the plug-in capabilities, application layer needs to do the following:

- C++

  ```cpp
  #include "onnxruntime_cxx_api.h"
  #include "ort_genai.h"
  
  auto env = Ort::Env();
  env.RegisterExecutionProviderLibrary("CUDAExecutionProvider", "path\to\onnxruntime_providers_cuda.dll");
  
  auto model = OgaModel("path\to\model\directory");
  ...
  ```

- Python

  ```python
  import onnxruntime_genai as og
  
  og.register_execution_provider_library("CUDAExecutionProvider", "path\to\onnxruntime_providers_cuda.dll")
  
  model = og.Model("path\to\model\directory")
  ...
  ```

- C#

  ```csharp
  using Microsoft.ML.OnnxRuntime;
  using Microsoft.ML.OnnxRuntimeGenAI;
  
  var ortEnv = OrtEnv.Instance();
  ortEnv.RegisterExecutionProviderLibrary("CUDAExecutionProvider", "path\to\onnxruntime_providers_cuda.dll");
  
  using Model model = new Model("path\to\model\directory");
  ...
  ```

PR #1889, opened 2025-11-24
Support pre-registered plug-in NvTensorRtRtx execution provider library

anujj:
This PR basically extend the support for pre-registered plug-in for NvTenosrRtRTX which was earlier done for Cuda in the PR: https://github.com/microsoft/onnxruntime-genai/pull/1850

What We Are Doing

- Enable NvTensorRtRtx pre-registered plugin provider support - Allow users to register custom NvTensorRT execution provider libraries (similar to existing CUDA plugin support)
- Add example code demonstrating plugin provider usage in C++ and Python

Implementation Flow
Pre-Registered Plugin Path (PATH A)
When a custom EP library is registered

```
    app Code:
      └─> register_execution_provider_library("NvTensorRTRTXExecutionProvider", "/path/to/*dll")
           └─> Creates model with provider="NvTensorRtRtx"
    
    Model Creation (model.cpp):
      └─> Check GetOrtEnv().GetEpDevices() for pre-registered "NvTensorRTRTXExecutionProvider"
           └─> IF FOUND (PATH A):
                └─> Use AppendExecutionProvider_V2() with registered provider
```

Built-in/Default Path (PATH B)
When no plugin is registered (standard behavior):

```
User Code:
  └─> Creates model with provider="NvTensorRtRtx" (no pre-registration)

  Model Creation (model.cpp):
    └─> Check GetOrtEnv().GetEpDevices() for pre-registered "NvTensorRTRTXExecutionProvider"
         └─> IF NOT FOUND (PATH B):
              ├─> FALL THROUGH to generic provider registration block
              └─> Generic code uses AppendExecutionProvider() with built-in EP
                   └─> On WinML: includes device filtering (multi-GPU support)
                   └─> On non-WinML: simple provider append
```

### DIFF
--- a/src/models/model.cpp
+++ b/src/models/model.cpp
@@ -584,8 +584,8 @@ DeviceInterface* SetProviderSessionOptions(OrtSessionOptions& session_options,
     if (provider_options.name == "cuda") {
       // Try pre-registered plugin path first
       if (IsProviderRegistered(session_options, provider_options,
-                                            DeviceType::CUDA, "CUDAExecutionProvider",
-                                            is_primary_session_options, p_device)) {
+                               DeviceType::CUDA, "CUDAExecutionProvider",
+                               is_primary_session_options, p_device)) {
         continue;  // Handled via V2 API, skip built-in path
       }
 
@@ -694,8 +694,8 @@ DeviceInterface* SetProviderSessionOptions(OrtSessionOptions& session_options,
 
       // Try pre-registered plugin path first
       if (IsProviderRegistered(session_options, provider_options,
-                                            DeviceType::NvTensorRtRtx, "NvTensorRTRTXExecutionProvider",
-                                            is_primary_session_options, p_device)) {
+                               DeviceType::NvTensorRtRtx, "NvTensorRTRTXExecutionProvider",
+                               is_primary_session_options, p_device)) {
         continue;  // Handled via V2 API
       }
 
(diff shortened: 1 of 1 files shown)


anujj:
Adding @kunal-vaishnavi and @baijumeswani for review

src/models/model.cpp
```diff
@@ -647,174 +649,240 @@ DeviceInterface* SetProviderSessionOptions(OrtSessionOptions& session_options,
...
-      if (provider_options.name == "QNN") {
-        session_options.AddConfigEntry("ep.share_ep_contexts", "1");
-        // TODO set device_type_ in a less hacky way.
-        // now, all QNN EP enable_htp_shared_memory_allocator option values had better be consistent...
-        // on the other hand, not sure if is_primary_session_options is the right thing to check here.
-        if (const auto opt_it = std::find_if(provider_options.options.begin(), provider_options.options.end(),
-                                             [](const auto& pair) { return pair.first == "enable_htp_shared_memory_allocator"; });
-            opt_it != provider_options.options.end() && opt_it->second == "1") {
-          p_device = GetDeviceInterface(DeviceType::QNN);
+    } else if (provider_options.name == "NvTensorRtRtx") {
```
kunal-vaishnavi, line 687: This case looks almost identical to `cuda`. Can we consolidate to reduce code duplication?

https://github.com/microsoft/onnxruntime-genai/blob/271576fcb35d8434945e9301007871348b0b84e4/src/models/model.cpp#L534-L567

src/models/model.cpp
```diff
@@ -647,174 +649,240 @@ DeviceInterface* SetProviderSessionOptions(OrtSessionOptions& session_options,
...
+          if (device_ptrs[i]->Name() == "NvTensorRTRTXExecutionProvider") {
+            nvtrt_already_handled = true;
+            break;
+          }
+        }
+      }
 
-      std::vector<const OrtEpDevice*> ep_devices_ptrs;
-      ep_devices_ptrs.reserve(num_devices);
+      if (!nvtrt_already_handled) {
```
kunal-vaishnavi: Instead of nesting a lot of code inside the `if` condition, can we return early if it is already handled?

src/models/model.cpp
```diff
@@ -647,174 +649,240 @@ DeviceInterface* SetProviderSessionOptions(OrtSessionOptions& session_options,
...
 
-      size_t num_devices = 0;
-      const OrtEpDevice* const* device_ptrs = nullptr;
-      Ort::GetEpDevices(&GetOrtEnv(), &device_ptrs, &num_devices);
+    // Generic provider registration for all providers not handled by specific blocks above
+    // This handles: QNN, WebGPU, OpenVINO, VitisAI, and NvTensorRtRtx (when not pre-registered)
+    if (provider_options.name != "cuda" && provider_options.name != "rocm" && provider_options.name != "DML") {
+      // Check if this is NvTensorRtRtx with a pre-registered provider (already handled above)
+      bool nvtrt_already_handled = false;
+      if (provider_options.name == "NvTensorRtRtx") {
```
kunal-vaishnavi: Can we generalize to something like this?

```cpp
bool already_handled = std::any_of(GetOrtEnv().GetEpDevices().begin(),
                                   GetOrtEnv().GetEpDevices().end(),
                                   [](const auto& device_ptr) { return device_ptr->Name() == "NvTensorRTRTXExecutionProvider"; });
if (already_handled) return;

// For providers that go through the extensible AppendExecutionProvider API
```

anujj: done

baijumeswani:
@anujj is this pull-request still relevant? Could you please update it?

anujj:
yes, @baijumeswani , i will resolve the merge commits and update the PR 

I switched to the other task that's why it got delayed

src/models/model.cpp
```diff
@@ -660,127 +680,206 @@ DeviceInterface* SetProviderSessionOptions(OrtSessionOptions& session_options,
...
-      };
-      std::string ep_name{};
-      if (auto search = s_providerNameToExecutionProvider.find(provider_options.name); search != s_providerNameToExecutionProvider.end()) {
-        ep_name = search->second;
+    // Generic provider registration for all providers not handled by specific blocks above
+    // This handles: QNN, WebGPU, VitisAI, and NvTensorRtRtx (when not pre-registered)
+    // Note: OpenVINO and RyzenAI are handled by their own specific blocks above
+    if (provider_options.name != "cuda" && provider_options.name != "rocm" && provider_options.name != "DML" &&
+        provider_options.name != "OpenVINO" && provider_options.name != "RyzenAI") {
+      // Skip if NvTensorRtRtx was already handled via pre-registered plugin
```
baijumeswani, line 721: Why is this check necessary?

src/models/model.cpp
```diff
@@ -660,127 +680,206 @@ DeviceInterface* SetProviderSessionOptions(OrtSessionOptions& session_options,
...
+          // Append matched EP device
+          if (Ort::api->EpDevice_EpName(device_ptrs[i]) == ep_name &&
+              hardware_device_id_matched &&
+              hardware_vendor_id_matched &&
+              hardware_device_type_matched &&
+              hardware_ov_device_type_matched) {
+            ep_devices_ptrs.push_back(device_ptrs[i]);
+            // WinML Hotfix: DML and WebGPU EP factories currently only support one device at a time
+            // OpenVINO also supports only one device at a time
+            if (provider_options.name == "DML" || provider_options.name == "WebGPU" || provider_options.name == "OpenVINO") {
```
baijumeswani: Why are there openvino changes in this PR?

src/models/model.cpp
```diff
@@ -660,127 +680,206 @@ DeviceInterface* SetProviderSessionOptions(OrtSessionOptions& session_options,
...
+            keys.emplace_back(option.first.c_str());
+            values.emplace_back(option.second.c_str());
+          }
+
+          Ort::api->SessionOptionsAppendExecutionProvider_V2(
+              &session_options,
+              &GetOrtEnv(),
+              ep_devices_ptrs.data(), ep_devices_ptrs.size(),
+              keys.data(), values.data(), keys.size());
+        } else if (provider_options.name == "NvTensorRtRtx") {
```
baijumeswani: Now that we support the NvTensorRtRtx EP both via the plug‑in path and the older API, can we remove the NvTensorRtRtx‑specific code guarded by the WinML preprocessor macro? That code implicitly assumed that, in a WinML context, the EP would always be plugged in. Since we can now explicitly detect whether an EP has been plugged in, this special‑case logic should no longer be necessary.

src/models/model.cpp
```diff
@@ -508,6 +508,48 @@ void ConfigureNvTensorRtRtxProfile(const Config& config, OrtSessionOptions& sess
   }
 }
 
+// Helper to check if a provider is pre-registered and get the matching EP device
+static const OrtEpDevice* FindPreRegisteredEpDevice(const std::string& ep_name) {
```
baijumeswani: Could we move these functions into an anonymous namespace instead of declaring them as static?

src/models/model.cpp
```diff
@@ -508,6 +508,48 @@ void ConfigureNvTensorRtRtxProfile(const Config& config, OrtSessionOptions& sess
...
+  auto it = std::find_if(device_ptrs.begin(), device_ptrs.end(),
+                         [&ep_name](const OrtEpDevice* device) {
+                           return device->Name() == ep_name;
+                         });
+  return (it != device_ptrs.end()) ? *it : nullptr;
+}
+
+// Helper to handle pre-registered plugin provider via V2 API
+// Returns true if the provider was pre-registered and handled, false otherwise
+static bool HandlePreRegisteredPluginProvider(
```
baijumeswani: Could we move this function into an anonymous namespace as well? We could also change the name:

```suggestion
bool IsProviderRegistered(
```

src/models/model.cpp
anujj: This check prevents the generic provider-registration path from running for providers that already have dedicated handlers above (OpenVINO, RyzenAI, etc.). Without it we could double‑register or override their custom setup

src/models/model.cpp
anujj: The OpenVINO changes were an artifact of the rebase — the older OpenVINO handling inside the generic/WinML block got pulled in from an older base. I’ve now removed that dead code so OpenVINO is handled only by OpenVINO_AppendProviderOptions, matching current main

src/models/model.cpp
anujj: done

src/models/model.cpp
anujj: done

src/models/model.cpp
anujj: yeah i guess we can remove it now, @ishwar-raut1 should it be fine if i remove this fall back path now ? 
my understanding it

- The WinML path uses the V2 API (SessionOptionsAppendExecutionProvider_V2) which requires EP devices to be discoverable via GetEpDevices().
- For built‑in NvTensorRtRtx, the EP device list can be empty in WinML builds.
- If ep_devices_ptrs is empty, V2 does nothing → the provider would never be appended.
- So the fallback to legacy AppendExecutionProvider() was added to keep built‑in NvTensorRtRtx working even when no EP devices are found. 

now we have the ebi plugin path so safe the remove this fallback check ?

baijumeswani (dismissed):

kunal-vaishnavi (approved):

PR #1895, opened 2025-11-26
Implement graph models for EPs

qjia7:
This pull request introduces a new graph model building and execution utility using the ONNX Runtime Model Editor API, and integrates it with the WebGPU execution provider. The changes enable dynamic creation and caching of ONNX models (such as Cast and TopK operations) at runtime, allowing for flexible and efficient execution across different hardware backends. The most important changes are grouped below:

**Graph Model Building and Execution Utilities:**

* Added new files `graph_builder.h` and `graph_builder.cpp` implementing the `ModelConfig`, `TensorConfig`, and `AttributeValue` types, and a `GraphBuilder::Build` function to construct ONNX models dynamically using the Model Editor API. [[1]](diffhunk://#diff-ab0bfe9d45e482a78a9dff68d1ad90a00c60ec90190dac3a701d2a9343698079R1-R80) [[2]](diffhunk://#diff-54c386bc82f44fa10113920e8e0613be6a967efa5d0e731e15014540639013f8R1-R226)
* Added new files `graph_executor.h` and `graph_executor.cpp` providing `GraphExecutor::Execute` and a helper `ExecuteCastOp`, enabling execution of dynamically built models with session caching keyed by model configuration and execution provider. [[1]](diffhunk://#diff-09edfaf45433f0a3a49cbe3b1d4bfbddd65f98b006085267d05947143cd50203R1-R70) [[2]](diffhunk://#diff-0fe7b88593c1d2c3f7ca19a20d41f8bfed5c7b0eff302a29760e47584fc16b4aR1-R194)
* Updated `onnxruntime_api.h` to expose the Model Editor API via `GetModelEditorApi()`.
* Added a session cache to `OrtGlobals` for dynamically built graph sessions, with thread safety via a mutex. [[1]](diffhunk://#diff-bc6c7ee0bb3d69f59e0764be50ce24a471e3f415b9110b25a8a5a02a42be4c04R141-R148) [[2]](diffhunk://#diff-bc6c7ee0bb3d69f59e0764be50ce24a471e3f415b9110b25a8a5a02a42be4c04R16)

**WebGPU Integration:**

* Updated `src/webgpu/interface.cpp` to use the new graph execution utilities, implementing `InterfaceImpl::Cast` using `ExecuteCastOp` with WebGPU-specific session configuration. [[1]](diffhunk://#diff-570e78875f2d5b7e18ea9342f8c172e74392a6f584c1588d910c672d3a225fc7R7) [[2]](diffhunk://#diff-570e78875f2d5b7e18ea9342f8c172e74392a6f584c1588d910c672d3a225fc7R180-R209)

These changes significantly improve the flexibility and maintainability of runtime graph operations, making it easier to support new ONNX operators and execution providers.

With this change, phi4 with graph capture becomes 130.4 tps from 116.2 tps on NV 5080.

### DIFF
--- a/src/webgpu/interface.cpp
+++ b/src/webgpu/interface.cpp
@@ -188,7 +188,7 @@ struct InterfaceImpl : DeviceInterface {
     Ort::ThrowOnError(Ort::api->AllocatorGetInfo(ort_allocator_, &webgpu_mem_info));
 
     // WebGPU-specific session configuration
-    static const char* webgpu_config_key = "ep.webgpuexecutionprovider.registerInt64Ops";
+    static const char* webgpu_config_key = "ep.webgpuexecutionprovider.enableInt64";
     static const char* webgpu_config_value = "1";
     std::vector<const char*> session_config_keys = {webgpu_config_key};
     std::vector<const char*> session_config_values = {webgpu_config_value};


src/webgpu/cast_model_builder.cpp
```diff
@@ -0,0 +1,121 @@
...
+}
+
+// Encode embedded message (wire_type = 2)
+static void EncodeMessage(std::vector<uint8_t>& buffer, uint32_t field_number, const std::vector<uint8_t>& message) {
+  EncodeKey(buffer, field_number, 2);
+  EncodeVarint(buffer, message.size());
+  buffer.insert(buffer.end(), message.begin(), message.end());
+}
+
+std::vector<uint8_t> CreateCastModelBytes(ONNXTensorElementDataType input_type, ONNXTensorElementDataType output_type) {
```
kunal-vaishnavi: Can we generalize this process for future 1-op graphs? We have considered implementing this idea to perform computations for some search and sampling algorithms with on-device memory.

qjia7: Could you clarify what you mean by ‘generalize this process’? What specific changes would you like me to make?

qjia7: Done. Please take another look, thanks.

src/models/one_op_model_builder.h
```diff
@@ -0,0 +1,102 @@
...
+  std::string op_type;  // e.g., "Cast", "TopK", "Argmax"
+  std::vector<TensorConfig> inputs;
+  std::vector<TensorConfig> outputs;
+  std::vector<AttributeValue> attributes;
+
+  // ONNX opset version for the generated model.
+  // Default is 17 which is widely supported and has been validated with this infrastructure.
+  // Can be overridden if a specific opset is required, but ensure the ONNX Runtime build
+  // supports it and the operator exists in that opset version.
+  int opset_version{17};
```
kunal-vaishnavi: Can we use opset 21 to match the opset used in the model builder?

src/models/one_op_model_builder.cpp
```diff
@@ -0,0 +1,277 @@
...
+  // ModelProto
+  EncodeInt64(result, 1, 8);               // ir_version = 8
+  EncodeMessage(result, 7, graph_proto);   // graph
+  EncodeMessage(result, 8, opset_import);  // opset_import
+
+  return result;
+}
+
+// Helper to create a Cast model
+std::vector<uint8_t> OneOpModelBuilder::CreateCastModel(
```
baijumeswani: Building ONNX graphs in C/C++ is often complex. You could consider using the [Model Editor API ](https://github.com/microsoft/onnxruntime/blob/e7dfd69dce15ae44e6d64b6566af67cf8ef16ffa/include/onnxruntime/core/session/onnxruntime_c_api.h#L7241) to simplify this process.

Example [here](https://github.com/microsoft/onnxruntime/blob/e7dfd69dce15ae44e6d64b6566af67cf8ef16ffa/onnxruntime/test/shared_lib/test_model_builder_api.cc#L138).

src/models/one_op_model_executor.cpp
```diff
@@ -0,0 +1,260 @@
...
+
+// Global cache for 1-op model sessions
+// Stored in OrtGlobals to ensure proper cleanup before OrtEnv destruction
+struct OneOpSessionCache {
+  std::unordered_map<uint64_t, std::unique_ptr<OrtSession>> sessions_;
+  std::mutex mutex_;
+};
+
+static OneOpSessionCache& GetOneOpSessionCache() {
+  static OneOpSessionCache cache;
```
baijumeswani: Why does this need to be static? Can it not be one per interface created whenever the interface is created?

baijumeswani:
We should consider making this more generic so that each interface can define its own custom model using the ORT Model Editor API. This would allow the model to contain any number of inputs, outputs, and nodes. Once the model is created, we can build a session around it and cache that session within the interface, avoiding a global cache where possible. The responsibility of each interface would then be to build, cache, and execute the model as needed to support ort‑genai's internal requirements.

baijumeswani:
To enable WebGPU support quickly, we could temporarily embed the ONNX model in the source code and have the WebGPU interface build, cache, and execute the corresponding ORT session.

Example here: https://github.com/microsoft/onnxruntime-genai/blob/0c9dc84f69c7b7b24b74742eebbd8c43230d37fa/src/models/model.cpp#L806-L815

If preferred, we can introduce the infrastructure for building the ONNX model on demand at a later time.

src/models/one_op_model_builder.cpp
qjia7: Done.

src/models/one_op_model_executor.cpp
```diff
@@ -0,0 +1,273 @@
...
+    }
+  }
+
+  // Hash session config keys and values
+  for (size_t i = 0; i < session_config_keys.size(); i++) {
+    key ^= hasher(session_config_keys[i]) + 0x9e3779b9 + (key << 6) + (key >> 2);
+    key ^= hasher(session_config_values[i]) + 0x9e3779b9 + (key << 6) + (key >> 2);
+  }
+
+  return key;
```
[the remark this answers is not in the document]
qjia7: We support the dynamic shapes with the same model. So it's not necessary to add shapes into the cache key.

src/models/one_op_model_executor.cpp
qjia7: The static cache is intentional for maximizing session reuse across the entire application. See my reply in https://github.com/microsoft/onnxruntime-genai/pull/1895#issuecomment-3767584345

baijumeswani: Could you move the ownership of the cache to `OrtGlobals`? Then we can get the cache using `GetOrtGlobals().session_cache`. We could also avoid needing the OrtGlobals destructor and the ClearCache functions.

test/CMakeLists.txt
```diff
@@ -46,7 +45,15 @@ set_target_properties(unit_tests PROPERTIES
...
 
 if(NOT (CMAKE_SYSTEM_NAME STREQUAL "Android" OR CMAKE_SYSTEM_NAME STREQUAL "Linux" OR CMAKE_SYSTEM_NAME STREQUAL "Darwin"))
-target_link_libraries(unit_tests PRIVATE ${ONNXRUNTIME_LIB})
+  # Add one-op model infrastructure for standalone tests
+  target_sources(unit_tests PRIVATE
+    ${CMAKE_SOURCE_DIR}/src/models/one_op_model_builder.cpp
+    ${CMAKE_SOURCE_DIR}/src/models/one_op_model_executor.cpp
+  )
+  # Define flag so one_op uses standalone env in test context
+  target_compile_definitions(unit_tests PRIVATE GENAI_STANDALONE_ONE_OP)
```
baijumeswani: Most tests for the library exercise the API through the C interface, which helps ensure the API behaves correctly. Could we avoid introducing special preprocessor macros solely for testing?

src/models/one_op_model_executor.h
```diff
@@ -0,0 +1,94 @@
...
+      : execution_provider_name(ep_name), memory_info(mem_info) {}
+};
+
+// Manages creation, caching, and execution of 1-op ONNX models
+// This is a common utility that all EPs can use
+class OneOpModelExecutor {
+ public:
+  // Execute a 1-op model with the given configuration and parameters
+  // The session is automatically cached based on a hash of the model configuration
+  static void Execute(
```
baijumeswani: I am not sure I understand why all the functions inside this class are marked as static. Could you please explain?

src/models/one_op_model_builder.h
```diff
@@ -0,0 +1,87 @@
...
+  std::string op_type;  // e.g., "Cast", "TopK", "Argmax"
+  std::vector<TensorConfig> inputs;
+  std::vector<TensorConfig> outputs;
+  std::vector<AttributeValue> attributes;
+
+  // ONNX opset version for the generated model.
+  // Default is 21 which is widely supported and matches the opset used elsewhere in the codebase.
+  // Can be overridden if a specific opset is required, but ensure the ONNX Runtime build
+  // supports it and the operator exists in that opset version.
+  int opset_version{21};
```
baijumeswani:
```suggestion
  constexpr int opset_version{21};
```

src/models/one_op_model_builder.h
```diff
@@ -0,0 +1,87 @@
...
+
+  OneOpModelConfig(const std::string& op) : op_type(op) {}
+};
+
+// Builder class for creating 1-op ONNX models using the Model Editor API
+class OneOpModelBuilder {
```
baijumeswani: I you're trying group some functions together under a category, you could consider using `namespace` instead of a class here.

src/models/graph_executor.cpp
```diff
@@ -0,0 +1,286 @@
...
+  OrtModel* model = OneOpModelBuilder::Build(config);
+
+  // Create session from model
+  auto session = CreateSession(model, ep_name, session_config_keys, session_config_values);
+
+  // Release the model - session has its own copy
+  Ort::api->ReleaseModel(model);
+
+  OrtSession* session_ptr = session.get();
+  cache.sessions_[key] = std::move(session);
```
baijumeswani, line 95: If I understand correctly, there is a unique mapping of {EP, input spec, output spec, attribute spec, session options config keys, value} -> ort session. It also seems that once created, this session is not modified until the shutdown when it is destructed.

That would mean that no two interfaces would ever share the same ort session. In that case, I suggest you eliminate the hashing complexity altogether. You could consider creating the session once, and saving it in a global std::vector. Getting the reference to that session and returning it to the interface. The interface can then decide to hold onto the pointer to the session.

src/models/one_op_model_executor.cpp
```diff
@@ -0,0 +1,286 @@
...
+  Ort::api->ReleaseModel(model);
+
+  OrtSession* session_ptr = session.get();
+  cache.sessions_[key] = std::move(session);
+
+  return session_ptr;
+}
+
+// Execute a 1-op model
+void OneOpModelExecutor::Execute(
```
baijumeswani: There shouldn't be anything stopping this function from executing any general purpose ort graph (whether it is 1-op or multiple ops). Any reason for it to be called a OneOpModelExecutor?

src/models/graph_builder.cpp
baijumeswani, line 1: IMO, it would make sense to rename this file to something like model_builder.cpp or graph_builder.cpp instead of one_op_model_builder.cpp. This file could expose reusable utilities for building arbitrary ORT graphs using the model editor API. Each interface could then be responsible for constructing its specific graph, creating a session around it (even if the session is owned by OrtGlobals), and executing that session.

src/models/graph_builder.cpp
```diff
@@ -0,0 +1,226 @@
...
+    const char* domain_name = "";
+    Ort::ThrowOnError(model_editor_api.CreateModel(&domain_name, &config.opset_version, 1, &model));
+
+    // Add graph to model (model takes ownership of graph)
+    Ort::ThrowOnError(model_editor_api.AddGraphToModel(model, graph));
+    graph = nullptr;  // model now owns graph
+
+    return model;
+
+  } catch (...) {
```
baijumeswani, line 209: We can do this in a follow-up pull-request, but it would be good if we can follow the convenience RAII wrapper in onnxruntime_api.h file. That way, we don't need to do this cleanup here.

src/models/onnxruntime_api.h
```diff
@@ -124,6 +124,19 @@ using OrtApiBaseFn = const OrtApiBase* (*)(void);
 /// Before using this C++ wrapper API, you MUST call Ort::InitApi to set the below 'api' variable
 inline const OrtApi* api{};
 
+/// <summary>
+/// This returns a reference to the ORT C Model Editor API. Used if building or augmenting a model at runtime.
+/// </summary>
+/// <returns>ORT C Model Editor API reference</returns>
+inline const OrtModelEditorApi& GetModelEditorApi() {
```
baijumeswani, line 131: As mentioned in another comment, please consider using RAII here as we do for other structs of ort.

src/generators.h
```diff
@@ -137,6 +138,14 @@ struct OrtGlobals {
   };
   Allocator device_allocators_[static_cast<int>(DeviceType::MAX)];
 
+  // Cache for dynamically built graph sessions (e.g., Cast, TopK operations)
+  // Destroyed before env_ to ensure proper cleanup order
+  struct SessionCache {
+    std::unordered_map<uint64_t, std::unique_ptr<OrtSession>> sessions_;
+    std::mutex mutex_;
```
baijumeswani, line 145: Why do we use the mutex here? Execution in ort-genai is single threaded. We can safely assume that it will not be accessed by more than 1 thread at a time.

PR #1898, opened 2025-11-27
expose start/end profiling API in Models class

xiaofeihan1:
GenAI does not support profiling within a specific time range. This PR addresses it.

This PR introduces `start_profiling()` and `end_profiling()` APIs in the Model class, allowing developers to perform profiling only within a selected time span.

A Model owns an ONNX Runtime Session and the relationship is 1:1. Since profiling in ORT operates at the session level, we expose these APIs at the Model layer to make profiling control more accessible.

Note: We don't want to expose these in Generator layer. Because we have a concern for Generator A start profiling and Generator B end profiling.

### DIFF
--- a/src/models/model.cpp
+++ b/src/models/model.cpp
@@ -885,6 +885,7 @@ void EnsureDeviceOrtInit(DeviceInterface& device, const Config& config, std::uni
 }
 
 void SessionInfo::Add(OrtSession& session) {
+  sessions_.push_back(&session);
   auto input_names = session.GetInputNames();
   for (size_t i = 0; i < input_names.size(); i++) {
     auto type_info = session.GetInputTypeInfo(i);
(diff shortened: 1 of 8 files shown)


src/models/model.cpp
```diff
@@ -885,6 +885,7 @@ void EnsureDeviceOrtInit(DeviceInterface& device, const Config& config, std::uni
 }
 
 void SessionInfo::Add(OrtSession& session) {
+  sessions_.push_back(&session);
```
feich-ms, line 888: Do you need to remove the session from sessions_ if the session is deleted or removed? Otherwise the pointer will be dangerous since the session it points has been removed or destroyed.

xiaofeihan1: Thanks for the callout. There is no SessionInfo::Remove(OrtSession& session). UUIC, A model own a `session_info_` and multiple sessions. They are the same lifetime. That means when a session is destoryed, `sessions_ ` is also destoryed.

PR #1900, opened 2025-12-02
OpenVINO: Add support for model caching via 'cache_dir' provider option

RyanMetcalfeInt8:
This PR adds support for model caching, to improve 2nd+ load time.

It allows the `genai_config.json` to specify a 'cache_dir' provider option (e.g. 'openvino_cache'). When set, it will instruct the OpenVINO EP to create a cache directory within the model folder (by default, unless an absolute path is specified). OpenVINO will save compiled blobs to this directory, and then read on successive runs. This drastically reduces model load time for the 2nd+ runs.

This PR moves all OpenVINO-specific provider option handling into a `OpenVINO_AppendProviderOptions` API, provided by `openvino/interface`. The changes in `model.cpp` are mainly to remove all OpenVINO-specific handling that was grouped with the rest of the other EP's, as that's now handled within `OpenVINO_AppendProviderOptions`.

### DIFF
--- a/src/openvino/interface.cpp
+++ b/src/openvino/interface.cpp
@@ -212,8 +212,8 @@ static inline bool EndsWith(const std::string& str, const std::string& suffix) {
 }
 
 static inline std::optional<std::string> AddCacheDirToLoadConfig(const std::string& cache_dir,
-                                                                      std::optional<std::string> load_config_option,
-                                                                      const std::string& ov_device) {
+                                                                 std::optional<std::string> load_config_option,
+                                                                 const std::string& ov_device) {
   // convert raw cache_dir path into OpenVINO key/value pair
   std::string cache_dir_option = "\"CACHE_DIR\":\"" + cache_dir + "\"";
 


src/openvino/interface.cpp
```diff
@@ -66,4 +66,302 @@ bool IsOpenVINOStatefulModel(const Model& model) {
   return false;
 }
 
+static inline std::string GetOVDeviceStringFromOrtDevice(const OrtEpDevice* device_ptr) {
+  const OrtKeyValuePairs* keyvals = Ort::api->EpDevice_EpMetadata(device_ptr);
```
kunal-vaishnavi, line 70: Can we use the [GenAI wrapper APIs](https://github.com/microsoft/onnxruntime-genai/blob/main/src/models/onnxruntime_inline.h) for any calls to `Ort::api->name` or add new ones if the existing ones are insufficient?

RyanMetcalfeInt8: I can add a new GenAI wrapper for this if you want -- btw, it's not a new call that I added. On `main` branch it happens here as well: https://github.com/microsoft/onnxruntime-genai/blob/main/src/models/model.cpp#L748

Just curious, how come ORT GenAI defines a new set of C++ wrappers -- is this more ABI stable than using the C++ APIs which ORT already defines?

kunal-vaishnavi: It's from the original code for ORT GenAI. They are supposed to be lightweight wrappers around the ORT APIs for convenience so we can reduce the amount of code that needs to be written inside GenAI to use the ORT APIs for common calls. They are not meant to introduce new functionality.

The context for this is that I am aiming to clean up the codebase and have consistent usage across files. Some use the GenAI wrapper calls and some don't. This can be done later in another PR but I just wanted to mention it here.

kunal-vaishnavi (approved):

src/openvino/interface.cpp
RyanMetcalfeInt8: Thanks @kunal-vaishnavi -- makes sense. I can do that, let me submit it as a separate PR next week. In addition, there's some duplicate code related to selecting an ORT device given those new device_id, etc. config options -- so I'll try to expose that as a helper function as part of that changeset.

PR #1902, opened 2025-12-03
Integrate FARA-7B model

apsonawane:
This pull request introduces significant improvements to the configuration system and inference pipeline, especially for vision and multimodal models. It adds support for object-style pipeline definitions, enhances flexibility in model configuration, and refines input handling for sliding window models. The changes also include a new Python example for Fara VLM inference using the updated system.

### Configuration and Pipeline Enhancements

* Added support for object-style pipeline definitions for both decoder and vision models, allowing pipelines to be specified as objects with named stages and parameters, increasing flexibility for complex model architectures. (`src/config.cpp` [[1]](diffhunk://#diff-6b2f0a449fdefd8930e23ef0dcd752beec69242e1303d77653f047c5e0766385R588-R592) [[2]](diffhunk://#diff-6b2f0a449fdefd8930e23ef0dcd752beec69242e1303d77653f047c5e0766385R611) [[3]](diffhunk://#diff-6b2f0a449fdefd8930e23ef0dcd752beec69242e1303d77653f047c5e0766385R650-R737) [[4]](diffhunk://#diff-6b2f0a449fdefd8930e23ef0dcd752beec69242e1303d77653f047c5e0766385R770-R786) [[5]](diffhunk://#diff-6b2f0a449fdefd8930e23ef0dcd752beec69242e1303d77653f047c5e0766385R796-R798); `src/config.h` [[6]](diffhunk://#diff-c24f78b3519d763901eb9f67b864f01d802d803df1b24faaf154019cf812bf95R162-R179)
* Introduced new configuration fields for vision pipelines, including `PipelineModel` and `WindowIndexing`, supporting advanced vision processing workflows. (`src/config.h` [src/config.hR162-R179](diffhunk://#diff-c24f78b3519d763901eb9f67b864f01d802d803df1b24faaf154019cf812bf95R162-R179))
* Added support for `image_token_id` in the model configuration to enable multimodal token handling. (`src/config.cpp` [[1]](diffhunk://#diff-6b2f0a449fdefd8930e23ef0dcd752beec69242e1303d77653f047c5e0766385R973-R974); `src/config.h` [[2]](diffhunk://#diff-c24f78b3519d763901eb9f67b864f01d802d803df1b24faaf154019cf812bf95R284-R285)
* Improved handling of alternate naming for image size inputs, mapping `image_grid_thw` to `image_sizes` for greater compatibility. (`src/config.cpp` [src/config.cppR622-R623](diffhunk://#diff-6b2f0a449fdefd8930e23ef0dcd752beec69242e1303d77653f047c5e0766385R622-R623))

### Inference and Example Updates

* Added a comprehensive Python example (`fara_inference.py`) demonstrating multimodal inference with image and text inputs, including detailed prompt formatting and output handling for tool calls. (`examples/python/fara_inference.py` [examples/python/fara_inference.pyR1-R142](diffhunk://#diff-f0dfdb822900fdf9b8218ec5d7ebf08aa4268fbb324d03945da204e23e57569cR1-R142))

### Input Handling Improvements

* Refined padding logic for input IDs in sliding window models, correctly handling left and right alignment during device allocation for improved inference accuracy. (`src/generators.cpp` [src/generators.cppL321-R338](diffhunk://#diff-52e5b706d10acf58568e8717799a49d6ed0fcda1a61b45de8d982d209739fc0cL321-R338))
* Updated decoder-only state initialization to use the new position input creation logic, reflecting changes in the configuration system. (`src/models/decoder_only.cpp` [src/models/decoder_only.cppL19-R21](diffhunk://#diff-9080bf1f9205dfc25250df5bde7458b844511341ee209c6da836fd4ecb105967L19-R21))

### Dependency Update

* Updated the `onnxruntime_extensions` dependency to a newer commit, ensuring compatibility with recent features. (`cmake/deps.txt` [cmake/deps.txtL17-R17](diffhunk://#diff-12c22e06cbb37ea0ed9f9eaf60cbe408dbeef04072df6a9f431c3290822ea835L17-R17))

These changes collectively make the system more flexible for multimodal and vision models, improve configuration clarity and extensibility, and provide a robust example for users to follow.

### DIFF
--- a/src/models/qwen2_5_vl_image_processor.cpp
+++ b/src/models/qwen2_5_vl_image_processor.cpp
@@ -55,11 +55,11 @@ std::unique_ptr<NamedTensors> Qwen2_5VLImageProcessor::Process(const Tokenizer& 
 
   OrtxTensor* grid_thw_tensor = nullptr;
   CheckResult(OrtxTensorResultGetAt(result.get(), 1, &grid_thw_tensor));
-  
+
   if (grid_thw_tensor == nullptr) {
     throw std::runtime_error("grid_thw output not provided");
   }
-  
+
   named_tensors->emplace(image_grid_thw_name_, std::make_shared<Tensor>(ProcessTensor<int64_t>(grid_thw_tensor, allocator)));
 
   return named_tensors;
--- a/src/models/qwen_vl_vision.cpp
+++ b/src/models/qwen_vl_vision.cpp
@@ -46,14 +46,13 @@ QwenVisionPipeline::QwenVisionPipeline(OrtEnv& env,
     auto so = OrtSessionOptions::Create();
 
     so->SetIntraOpNumThreads(2).SetInterOpNumThreads(1);
-    
+
     // QNN provider options
     std::unordered_map<std::string, std::string> qnn_options = {
         {"backend_path", qnn_backend_path_},
         {"htp_performance_mode", "burst"},
         {"htp_graph_finalization_optimization_mode", "3"},
-        {"soc_model", "60"}
-    };
+        {"soc_model", "60"}};
 
     auto providers = Ort::GetAvailableProviders();
     bool has_qnn = std::find(providers.begin(), providers.end(), std::string("QNNExecutionProvider")) != providers.end();


src/config.cpp
```diff
@@ -615,6 +621,8 @@ struct VisionInputs_Element : JSON::Element {
       v_.pixel_values = JSON::Get<std::string_view>(value);
     } else if (name == "image_sizes") {
       v_.image_sizes = JSON::Get<std::string_view>(value);
+    } else if (name == "image_grid_thw") {  // accept alternate naming, map to image_sizes
+      v_.image_sizes = JSON::Get<std::string_view>(value);
```
tianleiwu: Store it to member like v_.image_grid_thw to avoid confusion?

src/config.cpp
```diff
@@ -641,6 +649,94 @@ struct VisionOutputs_Element : JSON::Element {
...
 };
 
+// Vision pipeline support structures
+struct VisionPipelineModel_Element : JSON::Element {
+  explicit VisionPipelineModel_Element(Config::Model::Vision::PipelineModel& v) : v_{v} {}
+
+  void OnValue(std::string_view name, JSON::Value value) override {
+    if (name == "filename") {
+      v_.filename = JSON::Get<std::string_view>(value);
+    } else if (name == "run_on_cpu") {
```
tianleiwu, line 659: Do we want to support choice of different devices (like gpu, npu, cpu) in the future? If so, I suggest to use name "device_type".

src/config.cpp
```diff
@@ -673,6 +769,23 @@ struct Vision_Element : JSON::Element {
     if (name == "outputs") {
       return outputs_;
     }
+    if (name == "window_indexing") {
```
tianleiwu: what is window_indexing for?

src/models/qwen_vl_vision.cpp
```diff
@@ -0,0 +1,229 @@
...
+#include <stdexcept>
+#include <cstring>
+#include <algorithm>
+#include <iostream>
+
+namespace Generators {
+
+// Minimal .npy reader for 1D integer arrays.
+// Only handles C-order, little-endian, shape (N,), for dtypes '<i4' or '<i8'.
+std::vector<int64_t> Load1DNpyIndices(const std::string& file_path) {
```
tianleiwu: why do we use npy format instead of onnx format?

src/config.cpp
apsonawane: Done

src/config.cpp
apsonawane: It is specific to npu for now. We only have few onnx models running on npu. We can keep it this way for now and then add more later

src/config.cpp
apsonawane: Vision encoder uses window-based attention which divides the image in 2x2 patches and window indexing reorders these windows before attention. After attention they are re-ordered back. After this the patches are merged.
This was provided with the NPU model so using npy instead of onnx

src/models/qwen_vl_vision.cpp
apsonawane: Answered above.

examples/python/qwen2_5_vl_inference.py
```diff
@@ -0,0 +1,170 @@
...
+        prompt_text=args.prompt,
+        max_new_tokens=args.max_new_tokens,
+        temperature=args.temperature,
+        top_k=args.top_k,
+        top_p=args.top_p,
+        do_sample=args.do_sample,
+        min_length=args.min_length,
+        repetition_penalty=args.repetition_penalty,
+    )
+
```
tianleiwu: Could you add some parity test (compare with PyTorch result)?

src/models/qwen_vl_vision.cpp
apsonawane: Removed window indexing npy dependency

examples/python/qwen2_5_vl_inference.py
```diff
@@ -0,0 +1,170 @@
+import argparse
+import json
+import sys
+from pathlib import Path
+
+import onnxruntime_genai as og
+import winml
+
+print(winml.register_execution_providers(ort=False, ort_genai=True))
```
kunal-vaishnavi: Can we move this to the [WinML examples](https://learn.microsoft.com/en-us/windows/ai/new-windows-ml/tutorial?tabs=python#full-code-samples) since it's WinML-focused and instead support this model through the existing vision-language model examples ([model_vision.cpp](https://github.com/microsoft/onnxruntime-genai/blob/main/examples/c/src/model_vision.cpp) and [model-vision.py](https://github.com/microsoft/onnxruntime-genai/blob/main/examples/python/model-vision.py))?

src/config.h
```diff
@@ -159,9 +159,22 @@ struct Config {
...
 
+      // Vision pipeline support (patch embed -> vision attn -> patch merger)
+      struct PipelineModel {
+        std::string filename;
+        std::optional<SessionOptions> session_options;
+        std::optional<RunOptions> run_options;
+        std::string model_id;              // Identifier used to link outputs to subsequent stages
+        std::vector<std::string> inputs;   // Graph input names
+        std::vector<std::string> outputs;  // Graph output names
+        bool run_on_cpu{true};             // If true force CPU EP when multiple EPs are configured
```
kunal-vaishnavi: Should this be true by default?

src/config.cpp
```diff
@@ -856,6 +949,8 @@ struct Model_Element : JSON::Element {
       v_.decoder_start_token_id = static_cast<int>(JSON::Get<double>(value));
     } else if (name == "sep_token_id") {
       v_.sep_token_id = static_cast<int>(JSON::Get<double>(value));
+    } else if (name == "image_token_id") {
```
kunal-vaishnavi: Can this token be taken out of the config? For previously-supported VLMs, we have kept the special token ids inside GenAI.

https://github.com/microsoft/onnxruntime-genai/blob/4f372984379efcc68fd6f408b8fcb7638a7c336b/src/models/gemma_image_processor.cpp#L16-L19

https://github.com/microsoft/onnxruntime-genai/blob/4f372984379efcc68fd6f408b8fcb7638a7c336b/src/models/phi_multimodal_processor.cpp#L60-L61

src/models/kv_cache.cpp
```diff
@@ -202,7 +202,10 @@ DefaultKeyValueCache::DefaultKeyValueCache(State& state)
       // Uniform sliding window allocation (backward compatibility)
       shape_[2] = std::min(max_length, sliding_window_size);
     }
-  } else if (past_present_share_buffer_) {
+  } else {
```
kunal-vaishnavi: I think this check should still be needed. A user can choose whether to turn on/off past-present buffer sharing at runtime. When buffer sharing is disabled, the buffers should not be allocated to max length.

src/models/model.cpp
```diff
@@ -234,6 +236,10 @@ State::~State() {
 std::vector<int32_t> PadInputs(std::span<std::span<const int32_t>> sequences, int32_t pad_token_id) {
   bool pad_right_{true};
 
+  if (pad_token_id == 0 && g_log.enabled) {
+    Log("warning", "pad_token_id is 0, which may be uninitialized. Verify genai_config.json contains 'pad_token_id' field.");
```
kunal-vaishnavi: What if a model's pad token id is actually 0?

src/models/model.cpp
```diff
@@ -1193,6 +1199,8 @@ std::shared_ptr<Model> CreateModel(OrtEnv& ort_env, const char* config_path, con
 }
 
 std::shared_ptr<Model> CreateModel(OrtEnv& ort_env, std::unique_ptr<Config> config) {
+  if (config->model.type == "fara" || config->model.type == "qwen2_5_vl")
+    return std::make_shared<Qwen2_5_VL_PipelineModel>(std::move(config), ort_env);
```
kunal-vaishnavi, line 1199: We should consolidate the runtime support for Qwen-2.5 VL and Fara inside [`multi_modal.cpp`](https://github.com/microsoft/onnxruntime-genai/blob/main/src/models/multi_modal.cpp). The `multi_modal.cpp` file works for many VLMs (Phi-3 vision, Phi-3.5 vision, Phi-4 multimodal, Gemma-3 vision, etc). The pre-processing and the processor are fine to keep as separate files/classes.

src/models/qwen_vl_model.cpp
```diff
@@ -0,0 +1,180 @@
...
+  }
+
+  try {
+    image_features_buffer_ = vl_model_.vision_pipeline_->Run(pixel_data, pixel_shape_vec, grid_thw);
+  } catch (const std::exception&) {
+    return;  // Silent failure - pipeline already logs errors
+  }
+
+  auto out_shape = vl_model_.vision_pipeline_->GetLastOutputShape();
+  if (out_shape.size() != 2) return;
```
kunal-vaishnavi: Can we log errors for each of the early returns so the user is aware?

src/models/qwen_vl_vision.cpp
```diff
@@ -0,0 +1,272 @@
...
+      for (size_t i = 0; i < num_devices; ++i) {
+        if (Ort::api->EpDevice_EpName(device_ptrs[i]) == std::string("QNNExecutionProvider")) {
+          ep_devices_ptrs.push_back(device_ptrs[i]);
+        }
+      }
+
+      if (ep_devices_ptrs.empty()) {
+        throw std::runtime_error("QNNExecutionProvider requested for vision attention but not registered.");
+      } else {
+        Ort::api->SessionOptionsAppendExecutionProvider_V2(
```
kunal-vaishnavi: Can we use the [GenAI wrapper APIs](https://github.com/microsoft/onnxruntime-genai/blob/main/src/models/onnxruntime_inline.h) for any calls to `Ort::api->name` or add new ones if the existing ones are insufficient?

test/csharp/Microsoft.ML.OnnxRuntimeGenAI.Tests.csproj
```diff
@@ -75,7 +75,7 @@
         <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
         <Visible>false</Visible>
     </None>
-    <None Condition="Exists('$(OrtLibDir)\libonnxruntime.dylib')" Include="$(OrtLibDir)\libonnxruntime.dylib">
+    <None Condition="Exists('$(NativeBuildOutputDir)\libonnxruntime.dylib')" Include="$(NativeBuildOutputDir)\libonnxruntime.dylib">
```
kunal-vaishnavi: Can we revert these changes and sync with the main branch for the pipeline fixes?

src/config.h
apsonawane: should be false

test/csharp/Microsoft.ML.OnnxRuntimeGenAI.Tests.csproj
apsonawane: Done

src/models/kv_cache.cpp
apsonawane: That is correct, we do not want to allocate max length buffer without any reason. Added past_present_buffer condition in genai_config for this and reverted this back to original

src/config.cpp
apsonawane: Updated

src/models/model.cpp
apsonawane: Removed the warning

src/models/qwen_vl_model.cpp
apsonawane: Done

examples/python/qwen2_5_vl_inference.py
```diff
@@ -0,0 +1,170 @@
...
+    accum_text = ""
+    started_toolcall = False
+    print("\n=== Generating ===")
+
+    while not generator.is_done():
+        generator.generate_next_token()
+        token = generator.get_next_tokens()[0]
+        output_tokens.append(token)
+
+        if eos_ids and token in eos_ids and len(output_tokens) >= min_length:
```
nenad1002: I assume you meant or here? Either eos token is seen or the output token length is higher than min length

src/models/model.cpp
```diff
@@ -1193,6 +1199,8 @@ std::shared_ptr<Model> CreateModel(OrtEnv& ort_env, const char* config_path, con
 }
 
 std::shared_ptr<Model> CreateModel(OrtEnv& ort_env, std::unique_ptr<Config> config) {
+  if (config->model.type == "fara" || config->model.type == "qwen2_5_vl")
```
nenad1002, line 1198: nit missing {}

src/models/qwen2_5_vl_image_processor.cpp
```diff
@@ -0,0 +1,82 @@
...
+  const int64_t* pixel_values_shape{};
+  size_t pixel_values_dims{};
+  CheckResult(OrtxGetTensorData(pixel_values, &pixel_values_data, &pixel_values_shape, &pixel_values_dims));
+
+  if (pixel_values_dims >= 2) {
+    int64_t batch_size = pixel_values_shape[0];
+    int64_t num_patches = pixel_values_shape[1];
+
+    int64_t grid_t = 1;  // Single frame
+    int64_t grid_h = static_cast<int64_t>(std::sqrt(num_patches));
```
nenad1002: Is this correct? This looks like it works only for squares, but if so, then, you don't need the division below

src/models/qwen_vl_model.cpp
```diff
@@ -0,0 +1,180 @@
...
+      pixel_values_val = input.tensor->GetOrtTensor();
+    } else if (input.name == grid_thw_name) {
+      image_grid_thw_val = input.tensor->GetOrtTensor();
+    }
+  }
+  if (!pixel_values_val) return;
+
+  auto pixel_shape = pixel_values_val->GetTensorTypeAndShapeInfo()->GetShape();
+  std::vector<int64_t> pixel_shape_vec(pixel_shape.begin(), pixel_shape.end());
+  const float* pixel_data = pixel_values_val->GetTensorMutableData<float>();
```
nenad1002: What if pixel values are not float? How are we handling it? Or its always float?

apsonawane: Good catch, I added the conversions

src/models/qwen_vl_vision.cpp
```diff
@@ -0,0 +1,272 @@
...
+      env_(env) {
+  // Convert std::string model paths to ORTCHAR_T for cross-platform (char or wchar_t)
+  auto toOrtPath = [](const std::string& s) -> std::basic_string<ORTCHAR_T> {
+    return std::basic_string<ORTCHAR_T>(s.begin(), s.end());
+  };
+  auto pe_path = toOrtPath(patch_embed_model);
+  auto attn_path = toOrtPath(vision_attn_model);
+  auto merger_path = toOrtPath(patch_merger_model);
+
+  // Patch embed and patch merger sessions (CPU for now)
```
nenad1002, line 40: Whats preventing us from checking for CUDA and allocate on GPU? Was the assumption the benefit would not be large as the model is probably small?

apsonawane: This model is specific to NPU, we will be supporting CUDA and CPU after this release

src/models/model.cpp
apsonawane: Qwen uses vision pipeline which has 3 onnx sessions while other have single vision encoder session, also for text model we use are using decoder-only pipeline. Because of these differences from multi_modal we created a separate files and classes.

src/models/qwen2_5_vl_image_processor.cpp
apsonawane: Yes, updated. We are only checking for squares

examples/python/qwen2_5_vl_inference.py
apsonawane: Updated the script to use model-vision. Removed the other inference script

examples/python/qwen2_5_vl_inference.py
apsonawane: removed this file

src/models/model.cpp
apsonawane: Will update linter to add these braces. Will create a separate PR

examples/python/model-vision.py
```diff
@@ -1,6 +1,9 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License
 
+import winml
```
kunal-vaishnavi: Can we try-catch this? Not everyone is using WinML.

examples/python/model-vision.py
```diff
@@ -116,7 +137,17 @@ def run(args: argparse.Namespace):
...
         params = og.GeneratorParams(model)
-        params.set_search_options(max_length=7680)
+        if args.max_length:
+            max_length = args.max_length
+        else:
+            try:
+                config_path = Path(args.model_path) / "genai_config.json"
+                with open(config_path, "r") as f:
+                    config = json.load(f)
+                    max_length = config.get("model", {}).get("context_length", 7680)
```
kunal-vaishnavi: The max length may be too high for some models (e.g. 128K) and cause out-of-memory issues immediately. Can we keep the default of 7680 if `--max_length` is not provided by the user instead of reading from the GenAI config?

src/config.cpp
```diff
@@ -279,10 +279,10 @@ struct DecoderInputs_Element : JSON::Element {
       v_.input_ids = JSON::Get<std::string_view>(value);
     } else if (name == "inputs_embeds") {
       v_.embeddings = JSON::Get<std::string_view>(value);
-    } else if (name == "attention_mask") {
-      v_.attention_mask = JSON::Get<std::string_view>(value);
```
kunal-vaishnavi, line 283: Typically, the input order is `input_ids` or `inputs_embeds`, then `attention_mask`, and then `position_ids` because each subsequent input relies upon the prior input for its calculation. Can we revert this change to preserve that order?

src/models/qwen_vl_model.cpp
```diff
@@ -0,0 +1,244 @@
...
+
+  // Find vision pipeline stage paths
+  auto find_stage = [&](const std::string& id) -> std::string {
+    for (const auto& stage : config_->model.vision.pipeline) {
+      if (stage.model_id == id) return (config_->config_path / fs::path(stage.filename)).string();
+    }
+    return "";
+  };
+
+  auto patch_embed_path = find_stage("patch_embed");
```
kunal-vaishnavi, line 23: Can `std::any_of` be used in this file to simplify the 3 `find_stage` checks and the `use_qnn_attn` check?

src/models/model.cpp
kunal-vaishnavi: We should create a generic `multi_modal_pipeline.cpp` file then. There will likely be other multi-modal pipeline models to support in the future.

src/models/qwen_vl_model.cpp
```diff
@@ -0,0 +1,244 @@
...
+
+  auto pixel_type_info = pixel_values_val->GetTensorTypeAndShapeInfo();
+  auto pixel_shape = pixel_type_info->GetShape();
+  auto pixel_type = pixel_type_info->GetElementType();
+  
+  std::vector<int64_t> pixel_shape_vec(pixel_shape.begin(), pixel_shape.end());
+  const float* pixel_data = nullptr;
+  std::vector<float> converted_data;
+  
+  // Convert pixel values to float32 if needed
```
kunal-vaishnavi: Can we make an overloaded version of [`Cast()`](https://github.com/microsoft/onnxruntime-genai/blob/2e62384b41101cf1aa8346b965d91541f7793784/src/models/model.cpp#L1223) to do this instead?

Example of its current usage:
https://github.com/microsoft/onnxruntime-genai/blob/2e62384b41101cf1aa8346b965d91541f7793784/src/models/logits.cpp#L69-L73

src/models/qwen_vl_vision.cpp
```diff
@@ -0,0 +1,272 @@
...
+  patch_embed_session_ = OrtSession::Create(env_, pe_path.c_str(), nullptr);
+  patch_merger_session_ = OrtSession::Create(env_, merger_path.c_str(), nullptr);
+
+  if (use_qnn_attn_) {
+    // Ensure QNN provider is available
+    auto so = OrtSessionOptions::Create();
+
+    so->SetIntraOpNumThreads(2).SetInterOpNumThreads(1);
```
kunal-vaishnavi, line 48: Why hardcode these instead of reading from the GenAI config?

src/models/qwen_vl_model.cpp
apsonawane: Updated use_qnn_attn, but the find_stage checks remain as individual calls since they need to retrieve specific paths rather than just checking existence

src/config.cpp
apsonawane: Done

examples/python/model-vision.py
apsonawane: Done

src/models/qwen_vl_model.cpp
apsonawane: Updated

src/models/qwen_vl_vision.cpp
```diff
@@ -0,0 +1,272 @@
...
+
+    auto providers = Ort::GetAvailableProviders();
+    bool has_qnn = std::find(providers.begin(), providers.end(), std::string("QNNExecutionProvider")) != providers.end();
+    if (has_qnn) {
+      so->AppendExecutionProvider("QNNExecutionProvider", keys, values, 4);
+    } else {
+      // Use registered QNN EP
+      size_t num_devices = 0;
+      const OrtEpDevice* const* device_ptrs = nullptr;
+      Ort::GetEpDevices(&GetOrtEnv(), &device_ptrs, &num_devices);
```
kunal-vaishnavi: We can re-use some of the GenAI wrapper APIs here.

https://github.com/microsoft/onnxruntime-genai/blob/2e62384b41101cf1aa8346b965d91541f7793784/src/models/onnxruntime_inline.h#L420-L430

https://github.com/microsoft/onnxruntime-genai/blob/2e62384b41101cf1aa8346b965d91541f7793784/src/models/onnxruntime_inline.h#L349-L352

src/models/qwen_vl_vision.cpp
```diff
@@ -0,0 +1,272 @@
...
+    for (size_t i = 0; i < wnd_idx_.size(); ++i) pairs.emplace_back(wnd_idx_[i], i);
+    std::sort(pairs.begin(), pairs.end(), [](auto& a, auto& b) { return a.first < b.first; });
+    for (size_t i = 0; i < pairs.size(); ++i) rev_idx_[i] = static_cast<int64_t>(pairs[i].second);
+  }
+
+  size_t pixel_count = 1;
+  for (auto d : pixel_shape) pixel_count *= static_cast<size_t>(d);
+  auto pixel_tensor = CreateTensor(pixel_data, pixel_count, pixel_shape);
+
+  const char* pe_input_names[] = {"pixel_values"};
```
kunal-vaishnavi: The `const char*` input names should ideally be read from the GenAI config

src/models/qwen2_5_vl_image_processor.cpp
```diff
@@ -0,0 +1,82 @@
...
+  // Run image preprocessing using onnxruntime-extensions
+  // This will execute the full pipeline from processor_config.json:
+  // DecodeImage -> ConvertRGB -> Resize (smart_resize) -> Rescale -> Normalize -> PatchImage
+  ort_extensions::OrtxObjectPtr<OrtxTensorResult> result;
+  CheckResult(OrtxImagePreProcess(processor_.get(), payload.images->images_.get(), result.ToBeAssigned()));
+
+  OrtxTensor* pixel_values = nullptr;
+  CheckResult(OrtxTensorResultGetAt(result.get(), 0, &pixel_values));
+
+  named_tensors->emplace(pixel_values_name_, std::make_shared<Tensor>(ProcessTensor<float>(pixel_values, allocator)));
```
kunal-vaishnavi: Since `ProcessTensor` returns an `OrtValue`, can we query the `OrtValue` for the shape and dims instead of calling `OrtxGetTensorData`?

Something like this:
```cpp
auto pixel_values_ortvalue = ProcessTensor<float>(pixel_values, allocator);
auto shape = pixel_values_ortvalue->GetTensorTypeAndShapeInfo()->GetShape();
named_tensors->emplace(pixel_values_name_, std::make_shared<Tensor>(pixel_values_ortvalue));
```

Then `OrtxGetTensorData` is not needed.

src/models/qwen_vl_vision.cpp
```diff
@@ -0,0 +1,272 @@
...
+
+  std::vector<int64_t> attn_shape{seq_len, hidden_dim};
+  auto attn_in_tensor = CreateTensor(reordered_buf_.data(), reordered_buf_.size(), attn_shape);
+  const char* attn_input_names[] = {"hidden"};
+  OrtValue* attn_inputs[] = {attn_in_tensor.get()};
+
+  attn_out_buf_.resize(seq_len * hidden_dim);
+  auto attn_out_tensor = CreateTensor(attn_out_buf_.data(), attn_out_buf_.size(), attn_shape);
+  auto attn_out_name = vision_attn_session_->GetOutputName(0);
+  const char* attn_output_names[] = {attn_out_name.c_str()};
```
kunal-vaishnavi, line 191: The `const char*` output names should ideally be read from the GenAI config

src/models/model.cpp
apsonawane: Will create another PR to make this generic

src/models/qwen_vl_vision.cpp
apsonawane: Using GenAI wrappers

src/models/qwen2_5_vl_image_processor.cpp
apsonawane: Good suggestion, updated

src/models/qwen_vl_vision.cpp
apsonawane: Updated

src/models/qwen_vl_vision.cpp
apsonawane: Updated

src/models/qwen2_5_vl_image_processor.cpp
```diff
@@ -50,30 +50,47 @@ std::unique_ptr<NamedTensors> Qwen2_5VLImageProcessor::Process(const Tokenizer&
...
+  named_tensors->emplace(pixel_values_name_, std::make_shared<Tensor>(std::move(pixel_values_ortvalue)));
+
+  // Check if processor returns grid_thw as second output
+  OrtxTensor* grid_thw_tensor = nullptr;
+  auto grid_thw_result = OrtxTensorResultGetAt(result.get(), 1, &grid_thw_tensor);
+  
+  if (grid_thw_result == extError_t::kOrtxOK && grid_thw_tensor != nullptr) {
+    named_tensors->emplace(image_grid_thw_name_, std::make_shared<Tensor>(ProcessTensor<int64_t>(grid_thw_tensor, allocator)));
+  } else {
+    // Fallback: calculate grid_thw from pixel_values shape
```
tianleiwu: we shall remove this fallback path since the logic is not right (assumption of square image is most likely wrong: screen is usually close to 16:9).

If we know the raw image height and width, and the fallback path shall compute grid_thw from those information.

For now, we can enforce that that the extension returns it as second output (make sure the extension does return it).

apsonawane: Updated the logic, created a fix in onnxruntime-extensions to return grid_thw. https://github.com/microsoft/onnxruntime-extensions/pull/1014

examples/python/model-vision.py
```diff
@@ -1,17 +1,41 @@
...
+import readline
 import time
 from pathlib import Path
 
 import onnxruntime_genai as og
 
 # og.set_log_options(enabled=True, model_input_values=True, model_output_values=True)
 
+# Tool-calling system prompt for Qwen/Fara models
+TOOL_CALL_SYSTEM_PROMPT = """You are a web agent trying to complete user tasks on websites using function calls.
```
kunal-vaishnavi:
```suggestion
FARA_SYSTEM_PROMPT = """You are a web agent trying to complete user tasks on websites using function calls.
```

src/models/kv_cache.cpp
```diff
@@ -287,8 +290,12 @@ void DefaultKeyValueCache::Update(DeviceSpan<int32_t> beam_indices, int total_le
       state_.outputs_[output_index_ + layer_idx * 2 + 1] = presents_[layer_idx * 2 + 1].get();
     }
   } else {
-    // Uniform shape update (existing behavior)
-    shape_[2] = total_length;
+    // Uniform allocation
+    // With buffer sharing: use full capacity (buffers are reused)
+    // Without buffer sharing: use actual length for memory efficiency
+    if (!past_present_share_buffer_) {
```
kunal-vaishnavi: There already is an early return at the beginning of `DefaultKeyValueCache::Update()` for when buffer sharing is enabled. Everything else is assumed to be without any buffer sharing. I believe this change can be reverted.

https://github.com/microsoft/onnxruntime-genai/blob/4e6babc4ac319b7c9d15c4f852d4f04655570e06/src/models/kv_cache.cpp#L256-L259

src/models/kv_cache.cpp
```diff
@@ -270,13 +270,16 @@ void DefaultKeyValueCache::Update(DeviceSpan<int32_t> beam_indices, int total_le
...
     for (int layer_idx = 0; layer_idx < layer_count_; ++layer_idx) {
-      const int max_cache_length = static_cast<int>(layer_shapes_[layer_idx][2]);
-      const int actual_length = std::min(total_length, max_cache_length);
-
       std::array<int64_t, 4> current_shape = layer_shapes_[layer_idx];
-      current_shape[2] = actual_length;
+
+      // With buffer sharing: use full capacity (buffers are reused)
+      // Without buffer sharing: use actual length for memory efficiency
+      if (!past_present_share_buffer_) {
```
kunal-vaishnavi: Same comment as [this one](https://github.com/microsoft/onnxruntime-genai/pull/1902/files#r2591303498) here

src/models/qwen_vl_vision.cpp
apsonawane: Updated, using GenAI wrapper APIs

examples/python/model-vision.py
apsonawane: Done

src/models/kv_cache.cpp
apsonawane: Done

src/models/kv_cache.cpp
apsonawane: Fixed

src/models/qwen_vl_vision.cpp
apsonawane: Will be updated in following PR

PR #1903, opened 2025-12-03
Quark GPT-OSS support

thpereir:
Add Quark GPT-OSS support
- General support for QMoE zero point/asymmetric quantization
- New layers used by Quark quantized models
- Packing used for gate_up proj and down_proj inside Experts

### DIFF
--- a/src/python/py/models/builders/base.py
+++ b/src/python/py/models/builders/base.py
@@ -3227,7 +3227,11 @@ class Model:
             kwargs.get("weight3", ""),
             kwargs.get("scales3", ""),
             kwargs.get("bias3", ""),
+            kwargs.get("zero_points1", ""),
+            kwargs.get("zero_points2", ""),
+            kwargs.get("zero_points3", ""),
         ]
+
         output = f"{name}/output_0"
 
         extra_kwargs = (
--- a/src/python/py/models/builders/gptoss.py
+++ b/src/python/py/models/builders/gptoss.py
@@ -552,6 +552,8 @@ class GPTOSSModel(Model):
         op_type = self.moe_attrs["op_type"]
         moe_weight_type = f"{'q' if op_type == 'QMoE' else ''}weight"
 
+        has_quark_experts = self.has_quark_experts(mlp.experts)
+
         # Make router nodes
         router_basename = f"{basename}/router/MatMul"
         router_matmul_name = self.make_matmul(mlp.router, router_basename, root_input)
@@ -572,22 +574,25 @@ class GPTOSSModel(Model):
         gate_up_proj_weight = f"model.layers.{layer_id}.moe.experts.gate_up_proj.{moe_weight_type}"
         gate_up_proj_scales = f"model.layers.{layer_id}.moe.experts.gate_up_proj.scales"
         gate_up_proj_bias = f"model.layers.{layer_id}.moe.experts.gate_up_proj.bias"
+        gate_up_proj_zero_points = f"model.layers.{layer_id}.moe.experts.gate_up_proj.zero_points"
         down_proj_weight = f"model.layers.{layer_id}.moe.experts.down_proj.{moe_weight_type}"
         down_proj_scales = f"model.layers.{layer_id}.moe.experts.down_proj.scales"
         down_proj_bias = f"model.layers.{layer_id}.moe.experts.down_proj.bias"
+        down_proj_zero_points = f"model.layers.{layer_id}.moe.experts.down_proj.zero_points"
 
-        # Apply transpose depending on EP/op requirements
+        # Apply transpose depending on EP/op requirements and Quark expert presence
         # For quantized QMoE on CUDA, kernels expect scales along the hidden_size axis,
         # so we keep original orientation (last axis = hidden_size) when quantizing.
         # For non-quantized MoE or non-CUDA EPs, transpose to align MatMul layout.
-        if op_type == "QMoE" and self.ep == "cuda":
-            gate_up_proj_layout = mlp.experts.gate_up_proj
-            down_proj_layout = mlp.experts.down_proj
-        else:
-            gate_up_proj_layout = mlp.experts.gate_up_proj.transpose(-1, -2)
-            down_proj_layout = mlp.experts.down_proj.transpose(-1, -2)
+        if not has_quark_experts:
+            if op_type == "QMoE" and self.ep == "cuda":
+                gate_up_proj_layout = mlp.experts.gate_up_proj
+                down_proj_layout = mlp.experts.down_proj
+            else:
+                gate_up_proj_layout = mlp.experts.gate_up_proj.transpose(-1, -2)
+                down_proj_layout = mlp.experts.down_proj.transpose(-1, -2)
 
-        if op_type == "MoE":
+        if op_type == "MoE" and not has_quark_experts:
             # Save non-quantized MoE weights as initializers
             self.make_initializer(
                 gate_up_proj_layout.view(self.moe_attrs["num_experts"], -1, self.hidden_size),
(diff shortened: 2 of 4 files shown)


src/python/py/models/builders/base.py
```diff
@@ -3227,6 +3227,13 @@ def make_qmoe_op(self, name, **kwargs):
             kwargs.get("scales3", ""),
             kwargs.get("bias3", ""),
         ]
```
kunal-vaishnavi, line 3233: We can simplify this to always add the zero points into the above `inputs` list. The default value is an empty string for each one so the input will be ignored if not provided.

src/python/py/models/builders/base.py
```diff
@@ -3227,6 +3227,13 @@ def make_qmoe_op(self, name, **kwargs):
...
+        if kwargs.get("zp1", None) is not None:
+            inputs.append(kwargs.get("zp1", ""))
```
kunal-vaishnavi: Can we use the full-length name (e.g. zero_points_1, zero_points_2, zero_points_3) instead of the abbreviation (e.g. zp1, zp2, zp3)?

src/python/py/models/builders/gptoss.py
```diff
@@ -575,19 +577,22 @@ def make_moe_fused(self, layer_id, mlp, root_input):
         down_proj_weight = f"model.layers.{layer_id}.moe.experts.down_proj.{moe_weight_type}"
         down_proj_scales = f"model.layers.{layer_id}.moe.experts.down_proj.scales"
         down_proj_bias = f"model.layers.{layer_id}.moe.experts.down_proj.bias"
+        gate_up_proj_zp = f"model.layers.{layer_id}.moe.experts.gate_up_proj.zp" if has_quark_experts else None
```
kunal-vaishnavi: Can we move the condition of setting the string names for the zero points to when `make_moe_op` is called? Then the string names can still be initialized even if `has_quark_experts` is false.

Can we also move the name initialization of the zero points for `gate_up_proj` to where the other name initializations for `gate_up_proj` occur?

src/python/py/models/builders/gptoss.py
```diff
@@ -575,19 +577,22 @@ def make_moe_fused(self, layer_id, mlp, root_input):
...
+        down_proj_zp = f"model.layers.{layer_id}.moe.experts.down_proj.zp" if has_quark_experts else None
```
kunal-vaishnavi: Can we use `.zero_points` instead of `.zp` in the name?

src/python/py/models/builders/gptoss.py
```diff
@@ -649,7 +670,48 @@ def make_moe_fused(self, layer_id, mlp, root_input):
...
+            else:
+                # Get biases from individual projections
+                gate_bias = expert.gate_proj.bias if hasattr(expert.gate_proj, 'bias') and expert.gate_proj.bias is not None else torch.zeros(expert.gate_proj.qweight.shape[0])
+                up_bias = expert.up_proj.bias if hasattr(expert.up_proj, 'bias') and expert.up_proj.bias is not None else torch.zeros(expert.up_proj.qweight.shape[0])
+
+                # Combine gate and up biases (interleaved pattern: even=gate, odd=up)
+                gate_out_dim = gate_bias.shape[0]
+                up_out_dim = up_bias.shape[0]
+
+                combined_bias = torch.zeros(gate_out_dim + up_out_dim, dtype=gate_bias.dtype, device=gate_bias.device)
```
kunal-vaishnavi: Can we keep the tensor on CPU instead of having it on-device since it will be serialized to disk as part of an initializer?

src/python/py/models/builders/gptoss.py
```diff
@@ -649,7 +670,48 @@ def make_moe_fused(self, layer_id, mlp, root_input):
...
             scales2=down_proj_scales,
             bias2=down_proj_bias,
+            zp1=gate_up_proj_zp,
+            zp2=down_proj_zp,
         )
 
         # Assign output 0 of previous MoE as root input to next SkipLayerNorm
         self.layernorm_attrs["skip_input"] = f"{moe_name}/output_0"
+
+    def combine_gate_up_biases_from_experts(self, experts):
```
kunal-vaishnavi: Can we add asserts to ensure the two new combining functions only run for when the model was Quark-quantized or reference Quark in the function names?

src/python/py/models/quantized_model.py
```diff
@@ -70,6 +70,126 @@ def __init__(self):
...
+        self.gate_up_proj = QuantizedTensorModule()
+        self.down_proj = QuantizedTensorModule()
+
+class Experts:
+    """Container for all experts in a MoE layer."""
+    def __init__(self):
+        """Pre-processed experts attributes"""
+        self._experts = {}
+        """QMoE packed attributes"""
+        self.w_fc1_weights = None
```
kunal-vaishnavi: Can we remove the `w_` prefix before the names and expand `zps` to `zero_points`?

src/python/py/models/quantized_model.py
```diff
@@ -70,6 +70,123 @@ def __init__(self):
...
+
+    def values(self):
+        """Get all experts."""
+        return self._experts.values()
+
+    def keys(self):
+        """Get all expert IDs."""
+        return self._experts.keys()
+
+    def release_experts(self):
```
kunal-vaishnavi: Can we remove the methods that aren't being called such as `release_experts`?

src/python/py/models/quantized_model.py
```diff
@@ -70,6 +70,123 @@ def __init__(self):
         self.rotary_emb = TensorModule()
         self.k_norm = TensorModule()
         self.q_norm = TensorModule()
+        self.sinks = None
```
kunal-vaishnavi: Can we initialize this as a `TensorModule`?

src/python/py/models/quantized_model.py
```diff
@@ -70,6 +70,123 @@ def __init__(self):
...
+class QuantizedExpert:
+    """Represents a single expert in MoE with quantized weights."""
+    def __init__(self, expert_id: int):
+        self.expert_id = expert_id
+        self.gate_proj = QuantizedTensorModule()
+        self.up_proj = QuantizedTensorModule()
+        self.gate_up_proj = QuantizedTensorModule()
+        self.down_proj = QuantizedTensorModule()
+
+class Experts:
```
kunal-vaishnavi:
```suggestion

class QuantizedExperts:
```

src/python/py/models/quantized_model.py
```diff
@@ -70,6 +70,123 @@ def __init__(self):
...
+
+    def release_experts(self):
+        self._experts.clear()
+
+    @property
+    def num_experts(self) -> int:
+        """Number of experts."""
+        return len(self._experts)
+
+    def set_weight_data(self, expert_id: int, proj_type: str, param_type: str, tensor, bits = None, group_size = None):
```
kunal-vaishnavi:
```suggestion
    def set_weight_data(self, expert_id: int, proj_type: str, param_type: str, tensor, bits: int, group_size: int):
```

src/python/py/models/quantized_model.py
```diff
@@ -678,6 +806,35 @@ def set_properties(self):
...
+                            expert.gate_proj.in_features = expert.gate_proj.qweight.shape[0]
+                            expert.up_proj.out_features = expert.up_proj.scales.shape[1]
+                            expert.up_proj.in_features = expert.up_proj.qweight.shape[0]
+
+                        expert.down_proj.out_features = expert.down_proj.scales.shape[1]
+                        expert.down_proj.in_features = expert.down_proj.qweight.shape[0]
+
+                        # Set g_idx if not already set
+                        expert.gate_proj.g_idx = expert.gate_proj.g_idx if expert.gate_proj.g_idx is not None else torch.tensor([i // expert.gate_proj.group_size for i in range(expert.gate_proj.in_features)], dtype=torch.int32)
+                        expert.up_proj.g_idx = expert.up_proj.g_idx if expert.up_proj.g_idx is not None else torch.tensor([i // expert.up_proj.group_size for i in range(expert.up_proj.in_features)], dtype=torch.int32)
```
kunal-vaishnavi: Can we create a helper function to initialize `g_idx`? It seems a lot of the initialization code is duplicated in this file.

Function:

```py
def set_g_idx(self, module):
    if module.g_idx is None:
        module.g_idx = torch.tensor([i // module.group_size for i in range(module.in_features)], dtype=torch.int32)
```

Usage:

```py
self.set_g_idx(expert.gate_proj)
self.set_g_idx(expert.up_proj)
self.set_g_idx(expert.down_proj)
```

src/python/py/models/quantized_model.py
```diff
@@ -906,6 +1063,23 @@ def pack_ort_format(self, module, intweight):
         else:
             module.qzeros = intzeros_pt.contiguous()
 
+    def pack_zeros_ort_format(self, module):
+        """
+        Pack `qzeros` to ORT format
```
kunal-vaishnavi, line 997: I presume this was added because you want to only pack `qzeros`. Can we refactor the `pack_ort_format` method so there aren't two methods to maintain for packing `qzeros`?

src/python/py/models/quantized_model.py
```diff
@@ -1092,13 +1266,187 @@ def __init__(self, quant_type, input_path, quant_attrs, q_size, kv_size, interme
                     # Set `g_idx` to None since it's not used in `MatMulNBits`
                     q_tensors.g_idx = None
 
+                if isinstance(q_tensors, Experts) and q_tensors.num_experts > 0:
+                    # Process each expert
+                    self.unpack_repack_experts(q_tensors)
```
kunal-vaishnavi, line 1202: Can we keep unpacking and repacking as separate calls (i.e. `self.unpack(q_tensors)` and `self.repack(q_tensors)` instead)? That way, other quantization providers can choose to write their own implementations for unpacking. It will also be consistent with how the other weights are processed.

src/python/py/models/quantized_model.py
```diff
@@ -1092,13 +1266,187 @@ def __init__(self, quant_type, input_path, quant_attrs, q_size, kv_size, interme
...
             # Set `g_idx` to None since it's not used in `MatMulNBits`
             self.lm_head.g_idx = None
 
+    def unpack_repack_experts(self, experts):
+        """
+        Unpacks weights from pre-quantized Quark experts
+        """
+        for expert in experts.values():
+            # Process gate_proj
+            if hasattr(expert, 'gate_proj') and expert.gate_proj.qweight is not None:
```
kunal-vaishnavi: Does `expert.proj` refer to these tensor modules?

https://github.com/microsoft/onnxruntime-genai/blob/07e50a966ac695874b9ef8f14747cf2eb99950b4/src/python/py/models/quantized_model.py#L76-L83

If so, can the `hasattr` checks be removed since they will always be true?

src/python/py/models/quantized_model.py
```diff
@@ -1092,13 +1266,187 @@ def __init__(self, quant_type, input_path, quant_attrs, q_size, kv_size, interme
...
+        self.repack_qmoe_weights(experts)
+
+    def repack_qmoe_weights(self, experts):
+        """
+        Create quantized MoE weights from pre-quantized Quark experts.
+        For gate_up projection, it interleaves gate_proj and up_proj tensors
+        where even indices are for gate and odd indices are for up.
+        """
+        has_split_gate_up = all(
+            hasattr(expert, 'gate_proj') and expert.gate_proj is not None and expert.gate_proj.qweight is not None and hasattr(expert, 'up_proj') and expert.up_proj is not None and expert.up_proj.qweight is not None for expert in experts.values()
```
kunal-vaishnavi: Same question here about what `expert.proj` refers to

src/python/py/models/quantized_model.py
thpereir: They are separate but as the expert layout adds another layer of indirection, I moved the processing into  unpack_repack_experts()

src/python/py/models/quantized_model.py
thpereir: Yes, they can be removed! Done

src/python/py/models/quantized_model.py
thpereir: Yes, they can be removed! Done

src/python/py/models/quantized_model.py
thpereir: Refactored and re-used the qzero packing

src/python/py/models/quantized_model.py
thpereir: Done

src/python/py/models/quantized_model.py
thpereir: Done

src/python/py/models/quantized_model.py
thpereir: Done

src/python/py/models/quantized_model.py
thpereir: Done

src/python/py/models/quantized_model.py
thpereir: Done

src/python/py/models/quantized_model.py
```diff
@@ -70,6 +70,108 @@ def __init__(self):
...
+    def __getitem__(self, expert_id: int) -> QuantizedExpert:
+        """Get expert by ID."""
+        return self.get_expert(expert_id)
+
+    @property
+    def num_experts(self) -> int:
+        """Number of experts."""
+        return len(self._experts)
+
+    def set_weight_data(self, expert_id: int, proj_type: str, param_type: str, tensor, bits = int, group_size = int):
```
kunal-vaishnavi: I believe type hints have to be denoted with `:` and the `=` is for default values.

Ex:
```suggestion
    def set_weight_data(self, expert_id: int, proj_type: str, param_type: str, tensor, bits: int, group_size: int):
```

or

```suggestion
    def set_weight_data(self, expert_id: int, proj_type: str, param_type: str, tensor, bits: int = -1, group_size: int = -1):
```

src/python/py/models/quantized_model.py
kunal-vaishnavi: I still see them here

```suggestion
            expert.gate_proj is not None and expert.gate_proj.qweight is not None and expert.up_proj is not None and expert.up_proj.qweight is not None for expert in experts.values()
```

thpereir: Fixed! And rebased

src/python/py/models/quantized_model.py
thpereir: You're absolutely right! My oversight, fixed!

thpereir:
Windows arm64 workflow timed out. I force-pushed with a commit message change to re-trigger the workflows (no code change).

thiagocrepaldi:
Great work everybody, and Happy Holidays!

PR #1911, opened 2025-12-09
[VitisAI] EP loader

BoarQing:
Supported for load external provider for VitisAI

see correlated PR: https://github.com/microsoft/onnxruntime/pull/26627

### DIFF
--- a/benchmark/c/main.cpp
+++ b/benchmark/c/main.cpp
@@ -112,7 +112,7 @@ void WriteE2EStats(std::string_view label,
             << "\n";
 }
 
-std::string GeneratePrompt(size_t num_prompt_tokens, const OgaModel& model, const OgaTokenizer& tokenizer, size_t batch_size) {
+std::string GeneratePrompt(const benchmark::Options& opts, size_t num_prompt_tokens, const OgaModel& model, const OgaTokenizer& tokenizer, size_t batch_size) {
   const char* const base_prompt = "A";
   auto base_prompt_sequences = OgaSequences::Create();
   for (size_t i = 0; i < batch_size; ++i) {
@@ -120,12 +120,14 @@ std::string GeneratePrompt(size_t num_prompt_tokens, const OgaModel& model, cons
   }
 
   auto params = OgaGeneratorParams::Create(model);
-  params->SetSearchOption("max_length", static_cast<double>(num_prompt_tokens));
+  if (!opts.no_dynamic_max_length) {
+    params->SetSearchOption("max_length", static_cast<double>(num_prompt_tokens));
+  }
   params->SetSearchOption("min_length", static_cast<double>(num_prompt_tokens));
 
   auto generator = OgaGenerator::Create(model, *params);
   generator->AppendTokenSequences(*base_prompt_sequences);
-  while (!generator->IsDone()) {
+  while (!generator->IsDone() && num_prompt_tokens-- > 0) {
     generator->GenerateNextToken();
   }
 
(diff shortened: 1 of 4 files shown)


src/models/model.cpp
```diff
@@ -812,6 +813,27 @@ DeviceInterface* SetProviderSessionOptions(OrtSessionOptions& session_options,
         values.emplace_back(option.second.c_str());
       }
       session_options.AppendExecutionProvider(provider_options.name.c_str(), keys.data(), values.data(), keys.size());
+      if (provider_options.name == "VitisAI") {
+        if (const auto opt_it = std::find_if(provider_options.options.begin(), provider_options.options.end(),
+                                             [](const auto& pair) { return pair.first == "external_ep_libray"; });
```
kunal-vaishnavi, line 819:
```suggestion
                                             [](const auto& pair) { return pair.first == "external_ep_library"; });
```

src/models/kv_cache.cpp
```diff
@@ -206,6 +206,16 @@ DefaultKeyValueCache::DefaultKeyValueCache(State& state)
     shape_[2] = state_.params_->search.max_length;
   }
 
+  if (past_present_share_buffer_) {  // AMD-specific override
+    for (const auto& [key, val] : state_.params_->config.model.decoder.session_options.config_entries) {
+      if (key == "max_length_for_kv_cache") try {
```
kunal-vaishnavi: The `max_length` in the `search` section of the GenAI config already controls the max length for the KV cache when buffer sharing is enabled.

https://github.com/microsoft/onnxruntime-genai/blob/4227072f3b55cda2a4ede4f40002d2cf0a47dfcb/src/models/kv_cache.cpp#L205-L207

You can update your GenAI config to set `max_length` to the value set in the session options instead of this change.

src/models/model.cpp
```diff
@@ -812,6 +813,27 @@ DeviceInterface* SetProviderSessionOptions(OrtSessionOptions& session_options,
...
+            OrtEpFactory* factory = nullptr;
+            size_t num = 1;
+
+            func(nullptr, OrtGetApiBase(), nullptr, &factory, num, &num);
+          }
+          if (const auto func = (void (*)(OrtSessionOptions*))GetProcAddress(lib, "RyzenAI_SetSessionOptions"))
+            func(&session_options);
+
+          if (const auto api = OrtGetApiBase()->GetApi(ORT_API_VERSION)) {
+            void* unused;
```
kunal-vaishnavi: Can we re-use `Model::CreateSessionOptionsFromConfig` to do this instead?

https://github.com/microsoft/onnxruntime-genai/blob/4227072f3b55cda2a4ede4f40002d2cf0a47dfcb/src/models/model.cpp#L1124-L1126

src/models/model.cpp
BoarQing: Hi, can we accept the typo for now as a proper fix requires additional change to the onnxruntime side?
Or I can make an explicit conversion from option "external_ep_library" at OGA side to "external_ep_libray" at onnxruntime side.

src/models/model.cpp
BoarQing: done

src/models/kv_cache.cpp
akholodnamdcom: Thanks for the suggestion.

I'm afraid few OGA releases back it did not work for us. Basically because `max_length` affects more than just kv-cache size and need to have only kv-cache of a very specific size.

If you think it's not the case anymore, please let me know we can reattempt using `max_length`.

PR #1917, opened 2025-12-11
Add custom model path support for unit tests

mpasumarthi-git:
Add support for custom model paths in unit tests via command-line argument
- Added `--model_path` argument to `unit_tests.exe` to specify custom model locations
- Updated `test/main.cpp` to parse the custom model path argument
- Updated `test/model_tests.cpp`, `test/sampling_benchmark.cpp`, and `test/sampling_tests.cpp` to use custom paths for NvTensorRT tests
- Allows running tests with models in non-default locations

### DIFF
--- a/test/main.cpp
+++ b/test/main.cpp
@@ -14,21 +14,22 @@ std::string g_custom_model_path;
 int main(int argc, char** argv) {
   std::cout << "Generators Utility Library" << std::endl;
 
-  // Parse custom model path argument
-  for (int i = 1; i < argc; ++i) {
-    std::string arg = argv[i];
-    if (arg == "--model_path" && i + 1 < argc) {
-      g_custom_model_path = argv[++i];
-      std::cout << "Using custom model path: " << g_custom_model_path << std::endl;
-      break;
-    }
-  }
-
   std::cout << "Initializing OnnxRuntime... ";
   std::cout.flush();
   try {
     std::cout << "done" << std::endl;
     ::testing::InitGoogleTest(&argc, argv);
+
+    // Parse custom model path argument after InitGoogleTest
+    for (int i = 1; i < argc; ++i) {
+      const std::string arg = argv[i];
+      if (arg == "--model_path" && i + 1 < argc) {
+        g_custom_model_path = argv[++i];
+        std::cout << "Using custom model path: " << g_custom_model_path << std::endl;
+        break;
+      }
+    }
+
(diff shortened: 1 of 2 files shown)


mpasumarthi-git:
@anujj Please help review

kunal-vaishnavi (dismissed):

test/main.cpp
```diff
@@ -2,13 +2,28 @@
...
 
+// Global variable to store custom model base path
+std::string g_custom_model_path;
+
 int main(int argc, char** argv) {
   std::cout << "Generators Utility Library" << std::endl;
+
+  // Parse custom model path argument
+  for (int i = 1; i < argc; ++i) {
+    std::string arg = argv[i];
```
baijumeswani:
```suggestion
    const std::string arg = argv[i];
```

baijumeswani: Might also be a good idea to parse the args after `InitGoogleTest` to avoid accidentally touching gtest args.

baijumeswani (commented):
Is there an external script that is used to call the unit tests that passes in different models to the test? What is the use case of passing in a custom model?

test/sampling_tests.cpp
```diff
@@ -555,16 +558,20 @@ struct NvTensorRtRtxTestSetup {
   static NvTensorRtRtxTestSetup Create(int vocab_size, int batch_size, int max_length = 10) {
     NvTensorRtRtxTestSetup setup;
 
+    // Use custom path if provided, otherwise use default
+    const char* nvtrt_path = MODEL_PATH "hf-internal-testing/phi3-fp16-nvtrt";
```
kunal-vaishnavi: The below line was changed to `std::string` in this PR.

https://github.com/microsoft/onnxruntime-genai/blob/334c1da115bd454d52c31786a26c6099d156c708/test/sampling_benchmark.cpp#L115

Do you want to also change this to be `std::string`?

test/main.cpp
mpasumarthi-git: > Might also be a good idea to parse the args after `InitGoogleTest` to avoid accidentally touching gtest args.

Done

>https://github.com/microsoft/onnxruntime-genai/pull/1917#discussion_r2670458019

Done

test/sampling_tests.cpp
mpasumarthi-git: > Do you want to also change this to be `std::string`?

Done

PR #1918, opened 2025-12-12
[VitisAI] EP loader

akholodnamdcom:
Supported for load external provider for VitisAI

see correlated PR: https://github.com/microsoft/onnxruntime/pull/26627

replacement of https://github.com/microsoft/onnxruntime-genai/pull/1911

### DIFF
--- a/benchmark/python/benchmark_e2e.py
+++ b/benchmark/python/benchmark_e2e.py
@@ -74,7 +74,7 @@ def monitor_cpu_memory():
 
 
 # Use input model to generate prompt
-def generate_prompt(model, tokenizer, prompt_length) -> str:
+def generate_prompt(model, tokenizer, prompt_length, override_max_length) -> str:
     text = "a"
     prompt = f"{args.chat_template.format(input=text)}"
     tokens = tokenizer.encode(prompt)
@@ -82,7 +82,7 @@ def generate_prompt(model, tokenizer, prompt_length) -> str:
     max_length_to_use = prompt_length + len(tokens)
     params.set_search_options(
         min_length=prompt_length,
-        **({ "max_length": max_length_to_use } if not args.no_dynamic_max_length else {})
+        **({ "max_length": max_length_to_use } if override_max_length else {})
     )
 
     generator = og.Generator(model, params)
@@ -285,6 +285,9 @@ def run_benchmark(args, batch_size, prompt_length, generation_length, max_length
             raise ValueError(
                 f"Chat Template for model type {model_type} is not known. Please provide chat template using --chat_template"
             )
+    
+    # When -1 is passed as max_length we should not override that search option
+    override_max_length = max_length != -1
 
     # Generate prompt
     if args.use_random_tokens:
@@ -299,7 +302,7 @@ def run_benchmark(args, batch_size, prompt_length, generation_length, max_length
         prompt = f"{args.chat_template.format(input=text)}"
         tokens = tokenizer.encode(prompt)
     else:
-        text = [generate_prompt(model, tokenizer, prompt_length)] * batch_size
+        text = [generate_prompt(model, tokenizer, prompt_length, override_max_length)] * batch_size
         prompt = f"{args.chat_template.format(input=text)}"
         tokens = tokenizer.encode(prompt)
         prompt_length = len(tokens)
@@ -312,7 +315,7 @@ def run_benchmark(args, batch_size, prompt_length, generation_length, max_length
         top_k=args.top_k,
         top_p=args.top_p,
         temperature=temperature,
-        **({ "max_length": max_length } if not args.no_dynamic_max_length else {}),
+        **({ "max_length": max_length } if override_max_length else {}),
         min_length=max_length,
         batch_size=batch_size,
     )
@@ -357,7 +360,7 @@ def run_benchmark(args, batch_size, prompt_length, generation_length, max_length
             top_k=args.top_k,
             top_p=args.top_p,
             temperature=temperature,
-            **({ "max_length": max_length } if not args.no_dynamic_max_length else {}),
+            **({ "max_length": max_length } if override_max_length else {}),
             min_length=max_length,
             batch_size=batch_size,
         )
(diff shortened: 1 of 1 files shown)


src/models/model.cpp
```diff
@@ -771,7 +772,26 @@ DeviceInterface* SetProviderSessionOptions(OrtSessionOptions& session_options,
...
+          auto lib = LoadLibrary(lib_name.c_str());
+          if (const auto func = (void (*)(void*, const OrtApiBase*, void*, OrtEpFactory**, size_t, size_t*))GetProcAddress(lib, "CreateEpFactories")) {
+            OrtEpFactory* factory = nullptr;
+            size_t num = 1;
+
+            func(nullptr, OrtGetApiBase(), nullptr, &factory, num, &num);
+          }
+          if (const auto func = (void (*)(OrtSessionOptions*))GetProcAddress(lib, "RyzenAI_SetSessionOptions"))
+            func(&session_options);
+          fs::path custom_ops_lib_path(lib_name);
```
baijumeswani, line 788: Is the custom op library the same library where these extra functions are defined?

baijumeswani: If yes, is there a reason to separate out the custom op library registration from RyzenAI_SetSessionOptions? Could that also be moved in there?

benchmark/python/benchmark_e2e.py
```diff
@@ -543,6 +550,9 @@ def str2strlist(value):
         choices=["cpu", "cuda", "dml", "follow_config"],
         help="Execution provider to run the ONNX Runtime session with. Defaults to follow_config that uses the execution provider listed in the genai_config.json instead.",
     )
+    parser.add_argument(
```
baijumeswani: What is dynamic max length? This option may break the existing default behavior. Which means everyone must now include --no_dynamic_max_length in their scripts if they are not using vitis ai?

baijumeswani: It may be that I have misunderstood what dynamic max length is. If you could clarify, that would be helpful. Maybe also put in a comment in the code.

src/models/model.cpp
```diff
@@ -771,7 +772,26 @@ DeviceInterface* SetProviderSessionOptions(OrtSessionOptions& session_options,
...
       session_options.AppendExecutionProvider(provider_options.name.c_str(), keys.data(), values.data(), keys.size());
-
+#if defined(_WIN32)
+      if (provider_options.name == "VitisAI") {
+        if (const auto opt_it = std::find_if(provider_options.options.begin(), provider_options.options.end(),
+                                             [](const auto& pair) { return pair.first == "external_ep_libray"; });
+            opt_it != provider_options.options.end()) {
+          auto lib_name = opt_it->second;
+          auto lib = LoadLibrary(lib_name.c_str());
+          if (const auto func = (void (*)(void*, const OrtApiBase*, void*, OrtEpFactory**, size_t, size_t*))GetProcAddress(lib, "CreateEpFactories")) {
```
baijumeswani, line 782: Do you need an else to gracefully error out?

src/models/model.cpp
akholodnamdcom: There were some strange reasons to do so. But it's not really needed any more. Let me remove call to this function.

src/models/model.cpp
akholodnamdcom: Not an error. 

We might use this strange approach to load addition custom-ops-dlls that are not EPs.

benchmark/python/benchmark_e2e.py
```diff
@@ -543,6 +550,9 @@ def str2strlist(value):
...
+        "--no_dynamic_max_length", action="store_true", help="Disable dynamic max_length"
```
kunal-vaishnavi: The default value should be false like in the C benchmark. Otherwise, the max length is not dynamically set when this option is not provided.

```suggestion
        "--no_dynamic_max_length", default=False, action="store_true", help="Disable dynamic max_length"
```

benchmark/python/benchmark_e2e.py
akholodnamdcom: > What is dynamic max length? This option may break the existing default behavior. Which means everyone must now include --no_dynamic_max_length in their scripts if they are not using vitis ai?

By default benchmarks override `search/max_length` from model's `genai_config.json` with a value equal to number_of_input_tokens+number_of_output_tokens. I called this feature "dynamic max_length". That parameter affects KV-cache size. 

For some of RyzenAI/VitisAI models we need OGA to allocate KV-cache for max length of 4k (usually) whatever prompt size is. 

So, `no_dynamic_max_length` is required to benchmark some of our models only. 

Let me think what to put in argument "help" to make it clear...

src/models/model.cpp
baijumeswani: Will you be removing the call to this function?

akholodnamdcom: Call to `RyzenAI_SetSessionOptions` was removed. 
Call to `CreateEpFactories` must still remain.

akholodnamdcom: Call to RegisterCustomOpsLibrary must remain as well.

akholodnamdcom: Basically, change to this file was done to make it possible to load more custom ops from a file specified in model config file ("external_ep_libray"). 
Since the library could also be an EP, "CreateEpFactories" should be called to let it perform internal initializations.

benchmark/python/benchmark_e2e.py
baijumeswani: Instead of using a new argument to indicate ort-genai to not set the max_length, could we use a marker such as `--max_length -1` to that the code follow whatever max length is set in the genai config?

akholodnamdcom: Sure, it could be done. 

`-1` would mean "do not override" (that's what we need).

Any other value, if set via command line argument, would be passed here as-is:
https://github.com/microsoft/onnxruntime-genai/blob/main/benchmark/c/main.cpp#L123
https://github.com/microsoft/onnxruntime-genai/blob/main/benchmark/c/main.cpp#L182

Does it sound correct?

Any extra validation? Like "should be more than num_prompt_tokens + opts.num_tokens_to_generate"?

baijumeswani: Sounds good. Not sure if the validation already exists. You can add that validation in case you need it.

baijumeswani: Sorry, I should have clarified. Could you also do that for the benchmark_e2e.py file?

akholodnamdcom: Sure, I'm on it right now.

baijumeswani: There already seems to be a max_lengths argument. Could we reuse that?

akholodnamdcom: Just noticed that python benchmark already has "--max_lengths" argument. 
Looks like I could have updated the existing one instead of adding a new one.

PR #1921, opened 2025-12-16
Disable CUDA graph for Phi LongRoPE models with IF nodes on TRT-RTX

anujj:
Disable CUDA graph for Phi LongRoPE models with IF nodes on TRT-RTX

This change disables CUDA graph for the following
models when targeting TRT-RTX:
- Phi3MiniLongRoPEModel
- Phi3SmallLongRoPEModel
- Phi3MoELongRoPEModel

### DIFF
--- a/src/python/py/models/builders/base.py
+++ b/src/python/py/models/builders/base.py
@@ -118,7 +118,6 @@ class Model:
 
         # EP-specific variables
         self.ep = ep
-        self.has_if_nodes = False  # Track whether model contains IF nodes
         self.ep_attrs = {
             "cpu": {},
             "cuda": {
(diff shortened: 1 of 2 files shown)


anujj:
@gaugarg-nv for review

src/python/py/models/builders/base.py
```diff
@@ -2065,6 +2067,7 @@ def make_rotary_embedding_multi_cache(self, **kwargs):
                 name="small_rotemb_caches_graph",
             ),
         )
+        self.has_if_nodes = True
```
kunal-vaishnavi: I think there's a simpler way to support this. We can change this line to the following.

```suggestion
        self.ep_attrs[self.ep]["enable_cuda_graph"] = "0"
```

All of the other changes in this PR can then be removed.

anujj: changes done

PR #1919, opened 2025-12-12
Add support for CUDA and CPU arch for Qwen-2.5-VL and Fara-7B

apsonawane:
Add CUDA and CPU architecture support for Qwen-2.5-VL and Fara-7B model
Validated NPU model is also working with this change

```
{"accuracy": 0.8765493306891423,"task_name": "ScienceQA_Visual"}
{"accuracy": 0.8244818652849741, "task_name": "ai2d_test"}
{"accuracy": 0.8108, "task_name": "chart_qa_test"}
{"accuracy": 0.4825291181364393, "task_name": "intergps_test"}
```

PR #1923, opened 2025-12-17
Remove unused unit tests

apsonawane:
Fixes comments from this PR:
https://github.com/microsoft/onnxruntime-genai/pull/1919

### DIFF
--- a/test/python/test_qwen_fara_models.py
+++ b/test/python/test_qwen_fara_models.py
@@ -22,6 +22,8 @@ import numpy as np
 import onnxruntime_genai as og
 import pytest
 from _test_utils import run_subprocess
+from PIL import Image
+from transformers import AutoProcessor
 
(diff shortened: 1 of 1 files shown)


test/python/test_qwen_fara_models.py
```diff
@@ -213,9 +186,54 @@ def test_qwen_fara_accuracy_comparison(test_data_path, relative_model_path, rela
...
+
+        # Load and process image with PyTorch
+        pil_image = Image.open(image_path)
+        pytorch_inputs = pytorch_processor(text=prompt, images=pil_image, return_tensors="pt")
+
+        if "pixel_values" in pytorch_inputs:
+            pytorch_pixel_values = pytorch_inputs["pixel_values"].numpy()
+
+            # Check if shapes match
+            if pixel_array.shape == pytorch_pixel_values.shape:
```
tianleiwu, line 200: Do we have a test case with shape match so that this branch is hit?

apsonawane:
Not required

PR #1925, opened 2025-12-18
Update handling EOS token id detection

kunal-vaishnavi:
### Description

This PR updates the examples to show how EOS token id detection is handled with ONNX Runtime GenAI when generating tokens. With the addition of the [C# binding for GetNextTokens()](https://github.com/microsoft/onnxruntime-genai/pull/1865), all of the published examples now cover the cases listed below in version 1 of this PR. Previously, the earlier PR mentioned different variations of the generation loop and all of the variations had an issue.

This PR also introduces new APIs for tracking token count and querying the generator params:
- `Generator.TokenCount`
- `Params.GetSearchNumber`
- `Params.GetSearchBool`

Additionally, this PR adds some missing Tokenizer APIs for Objective-C.
- `Tokenizer.GetBosTokenId`
- `Tokenizer.GetEosTokenIds`
- `Tokenizer.GetPadTokenId`

### Motivation and Context

This PR is a follow-up to the issue fixed in [an earlier PR](https://github.com/microsoft/onnxruntime-genai/pull/1849). These APIs can be used by users to distinguish between the cases that `Generator.IsDone()` covers.

For example:

```cpp
bool hit_eos = generator->IsDone() && (generator->TokenCount() < static_cast<int>(params->GetSearchNumber("max_length")));
bool hit_max_length = generator->IsDone() && (generator->TokenCount() == static_cast<int>(params->GetSearchNumber("max_length")));
```

### Description

This PR updates how EOS token id detection is handled with ONNX Runtime GenAI when generating tokens. A new API called `Generator.HitEOS()` is introduced to detect whether an EOS token id has been generated. Another API called `Generator.HitMaxLength()` is also introduced to detect whether the max length has been hit before the generation loop has completed.

### Motivation and Context

This PR is a follow-up to the issue fixed in [an earlier PR](https://github.com/microsoft/onnxruntime-genai/pull/1849). The earlier PR mentions different variations of the generation loop but all of the variations have an issue.

There are two scenarios for terminating the generation loop: 1) hitting the EOS token id and completing the generation loop or 2) hitting the max length before the generation loop has completed. However, none of the variations adequately cover the two scenarios for terminating the generation loop.

#### 1. Original Generation Loop

```
while not IsDone():
    GenerateToken()
    GetLastToken()
    PrintLastToken()
```

Consider scenario 1 with this loop. After `GenerateToken()` produces the EOS token id, `GetLastToken()` will attempt to retrieve that token. However, ORT GenAI does not append the EOS token id to the list of sequences returned to the user (see the earlier PR for why). Instead, the second-to-last token will still be the last token in the list of sequences. Thus, `GetLastToken()` and `PrintLastToken()` will retrieve and again print the last token that the user saw.

#### 2. Return Early Generation Loop

```
while not IsDone():
    GenerateToken()
    if IsDone():
        break
    GetLastToken()
    PrintLastToken()
```

Consider scenario 2 with this loop. After `GenerateToken()` produces a token and the max length has been reached, the generator's state is marked as done. Then `IsDone()` will be true and the newest token won't be retrieved and printed since the loop is exited early.

#### 3. Infinite Generation Loop

```
while True:
    GenerateToken()
    if IsDone():
        break
    GetLastToken()
    PrintLastToken()
```

Consider scenario 2 with this loop. The same issue as the prior loop still applies. `GenerateToken()` will generate all of the tokens but once the max length is hit, `IsDone()` is true and the last token won't be retrieved and printed.

#### Conclusion

The reason that none of these generation loop variants work is because `IsDone()` currently covers both scenarios in one API and does not distinguish between them. One check needs to be in place in the condition of the while loop so that the loop continues, and another check needs to be after token generation to decide whether retrieving the last token should be done or not.

#### Solution

To fix this, a new API called `Generator.HitEOS()` is introduced. It returns `true` when the EOS token id is generated. The generation loop should be modified to the following.

```
while not IsDone():
    GenerateToken()
    if HitEOS():
        break
    GetLastToken()
    PrintLastToken()
```

If scenario 1 occurs in this loop, `HitEOS()` is `true` and the generation loop will exit early. If scenario 2 occurs in this loop, `HitEOS()` is `false` when the max length is reached. The last generated token can still be retrieved and printed. Then because the generator's state is done, `IsDone()` is `true` and the generation loop ends.

Here is a full end-to-end example demonstrating its usage.

```py

model = og.Model("/path/to/model/folder")
tokenizer_stream = tokenizer.create_stream()

params.set_search_options(max_length=25)

tokens = tokenizer.encode("<|system|>You are a helpful AI assistant.<|end|><|user|>What color is the sky?<|end|><|assistant|>")
print(f"Prompt: {len(tokens)}")
generator.append_tokens(tokens)

count = 0
    count += 1
    if generator.hit_eos():
        break

    new_token = generator.get_next_tokens()[0]
    print(tokenizer_stream.decode(new_token), end="", flush=True)

print()
print(f"Generated: {count}")
print(f"Total: {len(tokens) + count}")
```

#### Scenario 1

Before with loop version 1:
```
Prompt: 18
The color of the sky can vary depending on the viewing conditions and the presence of particles and moisture in the atmosphere. On a clear day, the sky appears blue due to Rayleigh scattering, where the atmosphere scatters sunlight in all directions and blue wavelengths are scattered more than other colors because they travel as shorter, smaller waves. This scattering causes the sky to appear blue to an observer on the ground. However, the sky can also appear various shades of blue, gray, or even take on vibrant hues like red or orange just before or just after sunrise or sunset, due to the scattering of sunlight by particles and moisture in the atmosphere..
Generated: 128
Total: 146
```

After with `generator.hit_eos()`:

```
Prompt: 18
The color of the sky can vary depending on the viewing conditions and the presence of particles and moisture in the atmosphere. On a clear day, the sky appears blue due to Rayleigh scattering, where the atmosphere scatters sunlight in all directions and blue wavelengths are scattered more than other colors because they travel as shorter, smaller waves. This scattering causes the sky to appear blue to an observer on the ground. However, the sky can also appear various shades of blue, gray, or even take on vibrant hues like red or orange just before or just after sunrise or sunset, due to the scattering of sunlight by particles and moisture in the atmosphere.
Generated: 128
Total: 146
```

#### Scenario 2

Before with loop version 2:
```
Prompt: 18
The color of the sky can
Generated: 7
Total: 25
```

After with `generator.hit_eos()`:
```
Prompt: 18
The color of the sky can vary
Generated: 7
Total: 25
```

### DIFF
--- a/src/generators.cpp
+++ b/src/generators.cpp
@@ -292,7 +292,7 @@ double GeneratorParams::GetSearchNumber(std::string_view name) const {
   if (name == "batch_size") {
     return static_cast<double>(search.batch_size);
   } else if (name == "chunk_size") {
-    return search.chunk_size.has_value() ? static_cast<double>(search.chunk_size.value()) : 0.0;
+    return static_cast<double>(search.chunk_size.value_or(0));
   } else if (name == "diversity_penalty") {
     return search.diversity_penalty;
   } else if (name == "length_penalty") {


src/cuda/search_cuda.cpp
```diff
@@ -25,14 +25,15 @@ Search_Cuda::Search_Cuda(const GeneratorParams& params)
   sequence_lengths_ = params.p_device->Allocate<int32_t>(batch_beam_size);
 
   eos_seen_buffer_ = CudaMallocArray<bool>(batch_beam_size, &eos_seen_);
-  cudaMemsetAsync(eos_seen_.data(), 0, eos_seen_.size_bytes(), GetStream());
+  done_cpu_ = CudaMallocHostArray<bool>(1);
```
apsonawane, line 28: Can we rename it to done_cuda or something generic since it is inside cuda file?

kunal-vaishnavi: This uses [`cudaMallocHost`](https://developer.download.nvidia.com/compute/DevZone/docs/html/C/doc/html/group__CUDART__HIGHLEVEL_ge439496de696b166ba457dab5dd4f356.html) to get CPU memory so the suffix is meant to clarify that it is on CPU.

https://github.com/microsoft/onnxruntime-genai/blob/f41b3cc910bdf5358a972a52e5038304a7c4230a/src/cuda/cuda_common.h#L78-L88

apsonawane:
In search_cuda.cpp line 173, we are checking for eos do we need to set hit_eos there or no?
https://github.com/microsoft/onnxruntime-genai/blob/4a13b80270d83faf8441713cd2665744dd52d421/src/cuda/search_cuda.cpp#L173

src/cuda/search_cuda.cpp
```diff
@@ -25,14 +25,13 @@ Search_Cuda::Search_Cuda(const GeneratorParams& params)
...
```
baijumeswani, line 28: optional: could move this to .h or to initialization list in constructor.

src/cuda/search_cuda.cpp
```diff
@@ -235,21 +238,21 @@ void GreedySearch_Cuda::AppendTokens(DeviceSpan<int32_t>& next_tokens) {
     return;
   }
 
-  cudaMemsetAsync(eos_seen_.data(), 0, eos_seen_.size_bytes(), GetStream());
-  *done_cpu_ = false;
+  ResetDone();
 }
 
 void BeamSearch_Cuda::AppendTokens(DeviceSpan<int32_t>& next_tokens) {
+  ResetDone();
```
baijumeswani: Why is `ResetDone` needed both at the start and end of this function?

src/cuda/search_cuda.h
```diff
@@ -19,6 +19,7 @@ struct Search_Cuda : Search {
     cudaStreamSynchronize(GetStream());
     return *done_cpu_;
   }  // TODO: Use an event
+  void ResetDone() override;
```
baijumeswani: Should this be private?

src/generators.cpp
```diff
@@ -454,6 +500,10 @@ void Generator::SetRuntimeOption(const char* key, const char* value) {
   state_->SetRunOption(key, value);
 }
 
+int32_t Generator::TokenCount() const {
```
baijumeswani: Should this return size_t? And should the c api also return size_t?

src/objectivec/include/ort_genai_objc.h
```diff
@@ -146,6 +146,27 @@ typedef NS_ENUM(NSInteger, OGAElementType) {
...
+
+/**
+ * Return the int representations of the array of EOS tokens.
+ *
+ * @return The array of EOS token ids
+ */
+- (nullable NSArray<NSNumber*>*)getEosTokenIds:(NSError**)error;
+
+/**
+ * Return the int representation of the PAD token.
```
baijumeswani, line 164: ?
```suggestion
 * Return the int representation of the Pad token.
```

src/ort_genai.h
```diff
@@ -485,6 +499,15 @@ struct OgaGenerator : OgaAbstract {
...
     return {out, out_count};
   }
+#else
+  std::vector<int32_t> GetNextTokens() {
+    std::vector<int32_t> next_tokens;
+    const int32_t* out;
+    size_t out_count;
+    OgaCheckResult(OgaGenerator_GetNextTokens(this, &out, &out_count));
+    next_tokens.assign(out, out + out_count);
+    return next_tokens;
```
baijumeswani:
```suggestion
    const int32_t* out;
    size_t out_count;
    OgaCheckResult(OgaGenerator_GetNextTokens(this, &out, &out_count));
    return std::vector<int32_t>(out, out + out_count);
```

src/search.h
```diff
@@ -19,6 +19,7 @@ struct Search : LeakChecked<Search> {
   virtual DeviceSpan<float> GetLogits() const = 0;
   virtual void SetLogits(DeviceSpan<float> logits) = 0;
   virtual bool IsDone() const = 0;
+  virtual void ResetDone() = 0;
```
baijumeswani: Is this needed in the base class? I don't think anyone outside of search classes calls this function.

src/cuda/search_cuda.cpp
kunal-vaishnavi: It is done in the greedy search version so this is to match that.

https://github.com/microsoft/onnxruntime-genai/blob/90cb011ce0d33e65987f265141d4fe7f694c1b20/src/cuda/search_cuda.cpp#L222-L240

src/cuda/search_cuda.cpp
kunal-vaishnavi: It is already in the `.cpp` file here. This moves it up so that the initialization is done before `ResetDone()` is called.

https://github.com/microsoft/onnxruntime-genai/blob/90cb011ce0d33e65987f265141d4fe7f694c1b20/src/cuda/search_cuda.cpp#L34-L35

src/generators.cpp
kunal-vaishnavi: I was following the logic used when we introduced the tokenizer's BOS, EOS, and PAD token ids and those return `int32_t` types. I can change this to be `size_t` if that is preferred.

src/cuda/search_cuda.h
kunal-vaishnavi: It could but all of the methods in that struct are already public so I left it that way to be consistent.

src/objectivec/include/ort_genai_objc.h
kunal-vaishnavi: These special tokens are typically referenced and written in all capital letters ([here](https://undercodenews.com/understanding-tokenizer-behavior-key-gotchas-for-developers/) is an example).

src/search.h
kunal-vaishnavi: I added it in the base class to be consistent with how the other methods are added there. This is the search class so having it here would make sense.

src/ort_genai.h
kunal-vaishnavi: Fixed

src/cuda/search_cuda.cpp
baijumeswani: Is it needed?

src/generators.cpp
baijumeswani: Usually lengths and sizes are returned at size_t. For token ids we should probably use into because the tokenizer defines it that way.
I'll leave it to you to decide whether or not you want to change it to size_t.

src/search.h
baijumeswani: The functions usually exposed in the interface base class are ones that you want to enforce derived classes to override. In this case ResetDone is only ever used from within the class and is not really an interface function. We could give derived classes more flexibility in whether or not they want to implement it (since nobody outside is invoking it).

src/cuda/search_cuda.cpp
kunal-vaishnavi: Not really since beam search should be handling the EOS state internally. I removed it.

src/generators.cpp
kunal-vaishnavi: Changed it to `size_t`

src/search.h
kunal-vaishnavi: Removed it from here

src/generators.cpp
```diff
@@ -288,6 +288,52 @@ bool GeneratorParams::IsPastPresentShareBufferEnabled(const std::string& model_t
          (search.num_beams == 1 || model_type == "whisper");
 }
 
+double GeneratorParams::GetSearchNumber(std::string_view name) const {
+  if (name == "batch_size") {
+    return static_cast<double>(search.batch_size);
+  } else if (name == "chunk_size") {
+    return search.chunk_size.has_value() ? static_cast<double>(search.chunk_size.value()) : 0.0;
```
baijumeswani: Can get it later:

```suggestion
    return static_cast<double>(search.chunk_size.value_or(0));
```

PR #1926, opened 2025-12-19
Fix QMoE blockwise quantization support for TRT-RTX execution provider

anujj:
Add QMoE and BF16 support for TRT-RTX execution provider

- Enable blockwise quantization for TRT-RTX/NvTensorRtRtx EPs
- Add gpt_oss_swiglu_fusion option for separate gate/up weights
- Add int4_qdq_block_size for MatMul quantization block size
- Add BF16 precision support for TRT-RTX
- Keep padding in QMoE weights for proper alignment

### DIFF
--- a/src/python/py/models/builder.py
+++ b/src/python/py/models/builder.py
@@ -406,7 +406,7 @@ def get_args():
                 int4_block_size = 16/32/64/128/256: Specify the block size for int4 quantization (MatMulNBits).
                     Default value is 32.
                 qmoe_block_size = 16/32/64/128/256: Specify the block size for QMoE expert weights quantization.
-                    Default is 128 for trt-rtx, 32 for others. Supported EPs: cpu, cuda, webgpu, trt-rtx.
+                    Default is 128 for trt-rtx, 32 for others. Supported EPs: cpu, webgpu, trt-rtx.
                 int4_is_symmetric = Quantize the weights symmetrically. Default is true.
                     If true, quantization is done to int4. If false, quantization is done to uint4.
                 int4_op_types_to_quantize = MatMul/Gather: Specify op types to target for int4 quantization.
--- a/src/python/py/models/builders/base.py
+++ b/src/python/py/models/builders/base.py
@@ -359,9 +359,9 @@ class Model:
         int4_algo_config = self.make_int4_algo_config(extra_options.get("int4_algo_config", "default"))
         self.int4_block_size = extra_options.get("int4_block_size", 32)
 
-        # CPU, CUDA, WebGPU, and TRT-RTX support block-wise quantization for QMoE.
+        # CPU, WebGPU, and TRT-RTX support block-wise quantization for QMoE.
         # TRT-RTX defaults to 128; others default to 32 for consistency with MatMulNBits.
-        supported_blockwise_eps = ["cpu", "cuda", "webgpu", "trt-rtx"]
+        supported_blockwise_eps = ["cpu", "webgpu", "trt-rtx"]
         default_qmoe_block_size = 128 if self.ep == "trt-rtx" else 32
         self.qmoe_block_size = int(extra_options.get("qmoe_block_size", default_qmoe_block_size))
 
@@ -3280,7 +3280,7 @@ class Model:
 
         # Use block-wise quantization for supported EPs when qmoe_block_size > 0.
         # TRT-RTX defaults to 128; others default to 32.
-        supported_blockwise_eps = ["cpu", "cuda", "webgpu", "trt-rtx"]
+        supported_blockwise_eps = ["cpu", "webgpu", "trt-rtx"]
         use_blockwise_quant = self.ep in supported_blockwise_eps and self.qmoe_block_size > 0
 
         if use_blockwise_quant:


src/python/py/models/builders/base.py
```diff
@@ -712,7 +720,22 @@ def to_int4(self) -> ir.Model:
             algo_config=self.quant_attrs["int4"]["algo_config"],
         )
         quant.process()
-        return ir.from_proto(quant.model.model)
+        model = ir.from_proto(quant.model.model)
+
+        # Convert float32 scales to bfloat16 if io_dtype is bfloat16.
+        # MatMulNBitsQuantizer doesn't natively support bfloat16, so we saved weights as float32
```
kunal-vaishnavi: In [this PR](https://github.com/microsoft/onnxruntime/pull/26102), I added support for bfloat16 in the `MatMulNBits` quantizer. It will be part of the ORT 1.24 release. You can use a nightly ORT package for now and remove this conversion.

src/python/py/models/builders/base.py
```diff
@@ -3227,11 +3257,15 @@ def make_qmoe_op(self, name, **kwargs):
             kwargs.get("weight3", ""),
             kwargs.get("scales3", ""),
             kwargs.get("bias3", ""),
-            kwargs.get("zero_points1", ""),
-            kwargs.get("zero_points2", ""),
```
kunal-vaishnavi, line 3231: This change can be reverted. The [new op spec](https://github.com/microsoft/onnxruntime/pull/26495) for QMoE includes optional zero points. Those optional inputs are stored as empty strings if unused and empty string inputs should not affect other models that don't support zero points.

src/python/py/models/builders/gptoss.py
```diff
@@ -70,7 +70,7 @@ def make_attention(self, layer_id, attention, root_input, **kwargs):
         self.window_size = original_window_size
 
     def make_moe(self, layer_id, mlp, root_input):
-        if self.ep in {"cpu", "cuda"}:
+        if self.ep in {"cpu", "cuda", "NvTensorRtRtx", "trt-rtx"}:
```
kunal-vaishnavi: We use `trt-rtx` internally and substitute with `NvTensorRtRtx` only when the GenAI config is created.

```suggestion
        if self.ep in {"cpu", "cuda", "trt-rtx"}:
```

src/python/py/models/builder.py
```diff
@@ -403,8 +403,10 @@ def get_args():
                     2 is fp16.
                     1 is fp32.
                     Default is 4 for the CPU EP and 0 for non-CPU EPs.
-                int4_block_size = 16/32/64/128/256: Specify the block size for int4 quantization.
+                int4_block_size = 16/32/64/128/256: Specify the block size for int4 quantization (MatMulNBits).
                     Default value is 32.
+                int4_qmoe_block_size = 16/32/64/128/256: Specify the block size for QMoE expert weights quantization.
```
kunal-vaishnavi: What is the reason for using a different block size in `MatMulNBits` and `QMoE`?

src/python/py/models/builders/base.py
```diff
@@ -3264,21 +3298,19 @@ def make_qmoe_weights(self, weights):
         dtype = torch.quint4x2 if self.moe_attrs["expert_weight_bits"] == 4 else torch.int8
         qweight, scales = None, None
 
-        # For QMoE, only use block-wise quantization when explicitly requested
-        # via int4_block_size and when using CPU or WebGPU execution providers, since
-        # block_size is only supported for these EPs in the QMoE operator.
-        use_blockwise_quant = "int4_block_size" in self.extra_options and self.ep in ["cpu", "webgpu"]
+        # Get block size from quantization attributes
```
kunal-vaishnavi: These changes look to be reverting to before [this PR](https://github.com/microsoft/onnxruntime-genai/pull/1861) was made. There were issues discovered with the old approach that necessitated the linked PR.

src/python/py/models/builder.py
anujj: based on the kernel support there are chances that the qmoe kernel and other can support different block_size, atleast this is the case for the trt-rtx ep

kunal-vaishnavi: Can we rename this option as `qmoe_block_size` then? The op can work for both int4 and int8 precision.

src/python/py/models/builder.py
```diff
@@ -403,8 +403,10 @@ def get_args():
...
+                    Default is 128 for trt-rtx, 0 (tensor-level) for others. Supported EPs: cpu, webgpu, trt-rtx.
```
kunal-vaishnavi: The default should be 32 for the others since the block size used in `MatMulNBits` and `QMoE` for other models has been identical. The list of supported EPs includes CPU, CUDA, WebGPU, and TRT-RTX.

src/python/py/models/builders/base.py
```diff
@@ -359,19 +359,26 @@ def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
         int4_algo_config = self.make_int4_algo_config(extra_options.get("int4_algo_config", "default"))
         self.int4_block_size = extra_options.get("int4_block_size", 32)
 
-        # Validate that only CPU and WebGPU EPs support int4_block_size for QMoE
-        if self.ep not in ["cpu", "webgpu"] and "int4_block_size" in extra_options and moe_op_type == "QMoE":
+        # CPU, WebGPU, and TRT-RTX support block-wise quantization for QMoE.
+        # TRT-RTX defaults to 128; CPU/WebGPU default to 0 (tensor-level) for backward compatibility.
+        supported_blockwise_eps = ["cpu", "webgpu", "trt-rtx"]
```
kunal-vaishnavi: CUDA is also supported

src/python/py/models/builders/base.py
```diff
@@ -359,19 +359,26 @@ def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
...
+                f"Block-wise quantization is only supported for: {', '.join(supported_blockwise_eps)}."
             )
 
         self.quant_attrs = {
             "int4": {
                 "accuracy_level": int(
                     extra_options.get("int4_accuracy_level", 4 if self.ep in ["cpu", "webgpu"] else 0)
                 ),
-                "block_size": int(self.int4_block_size),
+                "block_size": int(self.int4_qmoe_block_size),
```
kunal-vaishnavi:
```suggestion
                "qmoe_block_size": int(self.int4_qmoe_block_size),
```

src/python/py/models/builders/base.py
```diff
@@ -712,7 +720,9 @@ def to_int4(self) -> ir.Model:
             algo_config=self.quant_attrs["int4"]["algo_config"],
         )
         quant.process()
-        return ir.from_proto(quant.model.model)
+        model = ir.from_proto(quant.model.model)
```
kunal-vaishnavi: This can be reverted

src/python/py/models/builders/base.py
```diff
@@ -1056,7 +1066,14 @@ def make_matmul_op(self, matmul, basename, root_input, **kwargs):
 
     def make_matmul_float(self, matmul, name, root_input, **kwargs):
         weight = name[1:].replace("/", ".") + ".weight"
-        self.make_initializer(matmul.weight.T, weight, to=self.io_dtype)
+        # When onnx_dtype is INT4/UINT4, weights will be quantized by MatMulNBitsQuantizer later.
+        # MatMulNBitsQuantizer doesn't properly support BFLOAT16 inputs, so we need to save
```
kunal-vaishnavi: Same point as [mentioned before](https://github.com/microsoft/onnxruntime-genai/pull/1926#discussion_r2670244583)

src/python/py/models/builders/base.py
```diff
@@ -3227,11 +3244,17 @@ def make_qmoe_op(self, name, **kwargs):
             kwargs.get("weight3", ""),
             kwargs.get("scales3", ""),
             kwargs.get("bias3", ""),
-            kwargs.get("zero_points1", ""),
-            kwargs.get("zero_points2", ""),
-            kwargs.get("zero_points3", ""),
         ]
 
+        # TRT-RTX doesn't support zero_points inputs at all
```
kunal-vaishnavi, line 3240: Can the TRT-RTX kernel just ignore the empty inputs? Many kernel implementations do this for unsupported inputs. 

For example, the [MHA CPU kernel](https://github.com/microsoft/onnxruntime/blob/1817f4a81072c49dea2984853edca412674d66fb/onnxruntime/contrib_ops/cpu/bert/multihead_attention.cc#L58-L67) tries to access the `past_sequence_length` and `cache_indirection` inputs. However, those are rarely used and most models have them as empty strings. The variables become null pointers and are ignored if null in the kernel implementation.


I would like to avoid having EP-specific logic as much as possible because it makes it harder to maintain long-term.

src/python/py/models/builder.py
anujj: addressed

src/python/py/models/builders/base.py
anujj: done

src/python/py/models/builders/base.py
anujj: done

src/python/py/models/builders/base.py
anujj: done

src/python/py/models/builders/base.py
anujj: done

src/python/py/models/builders/base.py
anujj: Kunal, I understand your point. Actually, for this release, TRT-backed is already frozen. I can take an AI to handle this in the next cycle and remove this TRT-based condition after that handling.

kunal-vaishnavi: Sounds good. That would be great.

src/python/py/models/builders/base.py
kunal-vaishnavi: Actually, I was mistaken and it looks like [CUDA is still not supported](https://github.com/microsoft/onnxruntime/blob/50a03e4179809c0299bc143e521dcfc7817ace75/onnxruntime/contrib_ops/cuda/quantization/moe_quantization.cc#L40-L41) for block-wise quantization yet. Can you revert this?

anujj: updated

src/python/py/models/builders/base.py
anujj: thanks :)

src/python/py/models/builder.py

src/python/py/models/builders/gptoss.py
anujj: done

src/python/py/models/builders/base.py
anujj: done

anujj:
addressed @kunal-vaishnavi all comments

Issue #1910, opened 2025-12-09
Long context(tokens length >= 4090) encountered a sharp perf regression with beachmark_e2e.py


feich-ms:
When using beachmark_e2e.py to run against phi3 model, we encountered a sharp pref regression with tokens length(prompt length + generation lenght) greater then 4090, e.g., prompt length 3836, generation lenght 254.
Device: Mac with Apple M2 Pro chip
Repro steps:
1. Download Phi3 model
2. Run "python benchmark/python/benchmark_e2e.py -i model_path -l 3836 -g 253"
3. Check the token generation tps(which was 35 on my machine)
4. Run "python benchmark/python/benchmark_e2e.py -i model_path -l 3836 -g 254"
5. Check the token generation tps which should be much smaller than the first run(which was 14 on my machine).

I further dived deep and found the token 4090 took much more time to finish the generation, token 4089 used 0.04ms while token 4090 used 11.34ms, then 4091 recovered to 0.88ms. I also debugged into the ort and print the prameters variable in [ApplyFlashAttention](https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/contrib_ops/webgpu/bert/flash_attention.cc) function when generation token 4089 - 4091 as below. The sequence_length_ 4097 is confused when generating token 4090, the expected value should be 1 as token 4089 and 4091.

I wonder if there are any potential bugs in onnx or genai?

**3836+253=4089:**
(lldb) p parameters
(const onnxruntime::contrib::webgpu::WebgpuAttentionParameters &) 0x000000016b3034b0: {
  is_gqa_ = true
  batch_size_ = 1
  sequence_length_ = 1
  kv_sequence_length_ = 1
  past_sequence_length_ = 4099
  total_sequence_length_ = 4096
  max_sequence_length_ = 0
  input_hidden_size_ = 0
  hidden_size_ = 3072
  head_size_ = 128
  v_hidden_size_ = 1024
  v_head_size_ = 128
  num_heads_ = 24
  rotary_embedding_ = 0
  is_unidirectional_ = true
  past_present_share_buffer_ = true
  do_rotary_ = false
  broadcast_attn_bias_dim_0_ = false
  broadcast_attn_bias_dim_1_ = false
  mask_filter_value_ = -10000
  scale_ = 0.0883883461
  use_tf32_ = false
  seqlen_past_kv_cache_ = 4099
  seqlen_present_kv_cache_ = 4099
  kv_hidden_size_ = 1024
[...]

PR #888, opened 2024-09-12
dml specific change in model builder for phi3.5

zhangxiang1993:

PR #1930, opened 2025-12-22
Make phi model of webgpu ep always use long RoPE to improve tps performance for long context scenario

feich-ms:
We found a perf regression issue for long context which is described in this ticket https://github.com/microsoft/onnxruntime-genai/issues/1910. This PR proposed a fix to avoid the switch from short factor to long factor on webgpu ep of phi model to mitigate above issues caused by recomputation of position IDs and KV cache when switching to long factor. The experimental results from benchmark_e2e.py shows that it can generate the response tokens without perf regression for short(< 1000), middle(> 1000 and < 4097) and long(> 4097) sequence length. The fix includes two pieces:
1. In src/python/py/models/builders/base.py, for webgpu ep, we always use large cos/sin so that the model doesn't need the short-long switch.
`cos_cache = cos_cache_large`
`sin_cache = sin_cache_large`
2. In src/generators.cpp, for webgpu ep, we will skip the workaround fix of recomputation of position IDs and KV cache for the short to long factor switch, and always use the long factor for the model which is converted by above base.py.

Pls note this is not final fix for the issue, just demonstrating a reasonable direction to discuss and move forward.

I have tested against this change with below command:

Convert model: `python3 builder.py -m /Volumes/workspace/ort-web-perf/models/Phi-4-mini-instruct -o /Volumes/workspace/ort-web-perf/models/Phi-4-mini-instruct-onnx-optimized -p int4 -e webgpu --extra_options int4_accuracy_level=4 int4_algo_config=k_quant_last`

Run beachmark like below commands:
- Long context (7936+6000): `benchmark/python/benchmark_e2e.py -i /Volumes/workspace/ort-web-perf/models/Phi-4-mini-instruct-onnx-optimized -l 7936 -g 6000 --use_prompt_set -mo`
- Long context (4096+6000): `benchmark/python/benchmark_e2e.py -i /Volumes/workspace/ort-web-perf/models/Phi-4-mini-instruct-onnx-optimized -l 4096 -g 6000 --use_prompt_set -mo`
- Long context (4096+2000): `benchmark/python/benchmark_e2e.py -i /Volumes/workspace/ort-web-perf/models/Phi-4-mini-instruct-onnx-optimized -l 4096 -g 2000 --use_prompt_set -mo`
- Middle context: `benchmark/python/benchmark_e2e.py -i /Volumes/workspace/ort-web-perf/models/Phi-4-mini-instruct-onnx-optimized -l 2048 -g 1000 --use_prompt_set -mo`
- Short context: `benchmark/python/benchmark_e2e.py -i /Volumes/workspace/ort-web-perf/models/Phi-4-mini-instruct-onnx-optimized -l 256 -g 300 --use_prompt_set -mo`

Perf and Correctness Comparsion(Metrics: tps - average tokens generated per second, Original Solution means original model+generator, Proposed Solution means the updated model+generator by this pr)
| Length  | Original Solution | Proposed Solution |
|-------|-----|-----------|
| Long Context (7936+6000) | 22.62(Correctness: True) | 24.49(Correctness: True) |
| Long Context (4096+6000) | 26.26(Correctness: True) | 29.83(Correctness: True) |
| Long Context (4096+2000) | 26.84(Correctness: True) | 34.1(Correctness: True) |
| Middle Context (2048+1000) | 38.35(Correctness: True) | 38.54(Correctness: True) |
| Short Context (256+300) | 43.74(Correctness: True) | 43.54(Correctness: True) |

### DIFF
--- a/src/generators.cpp
+++ b/src/generators.cpp
@@ -497,8 +497,10 @@ void Generator::GenerateNextToken() {
   // at this stage which is achieved by rewinding to zero and appending the current sequence
   // Scenarios where this solution works: Batch size = 1, Num beams = 1, decoder model, EP is either CPU or CUDA
   // Scenarios where it doesn't work: Batch size > 1 OR Num beams > 1 OR Multimodal model (like phi3 vision) OR EP is DML
+  // Skip this logic for WebGPU EP as it already uses long RoPE scaling graph for all sequence lengths
   if (search_->params_->BatchBeamSize() == 1 && !epUsesSingleRopeFactor) {
-    if (((search_->GetSequenceLength() == 4097) && (model_->config_->model.type == "phi3" || model_->config_->model.type == "phimoe")) || ((search_->GetSequenceLength() == 8193) && (model_->config_->model.type == "phi3small"))) {
+    if (model_->p_device_->GetType() != DeviceType::WEBGPU &&
+      (((search_->GetSequenceLength() == 4097) && (model_->config_->model.type == "phi3" || model_->config_->model.type == "phimoe")) || ((search_->GetSequenceLength() == 8193) && (model_->config_->model.type == "phi3small")))) {
       auto current_seq = cpu_span<int32_t>(GetSequence(0).CopyDeviceToCpu());
       RewindToLength(0);
       AppendTokens(current_seq);
--- a/src/python/py/models/builders/base.py
+++ b/src/python/py/models/builders/base.py
@@ -1921,8 +1921,6 @@ class Model:
 
         # Determine which EPs don't support the If operator
         self.eps_without_if_support = ["dml"]
-        if self.extra_options.get("enable_webgpu_graph", False):
-            self.eps_without_if_support.append("webgpu")
 
         if self.ep in self.eps_without_if_support:
             cos_cache = torch.cat((cos_cache_small, cos_cache_large), dim=0)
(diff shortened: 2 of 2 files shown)


kunal-vaishnavi:
> @kunal-vaishnavi @baijumeswani Could you please advise whether always using cos_cache_large and sin_cache_large would be an acceptable approach? I’m not entirely sure if there would be any quality impact when the total sequence length is < 4096, but adopting this approach would allow us to remove the fixing introduced in #1161 and resolve the performance issue described in #1910. Please let us know if there’s anything we may have overlooked.

The reasoning for why the KV caches need to be re-computed and the impact on quality can be found [here](https://github.com/huggingface/transformers/pull/33129). You should use both the small and large caches for the best output quality. Ideally, the KV cache re-computation should not be avoided here.

This was a similar issue with the DML EP and the [initial fix](https://github.com/microsoft/onnxruntime-genai/pull/661) was to use just the large caches. However, quality issues soon emerged and a [new fix](https://github.com/microsoft/onnxruntime-genai/pull/888) was made. The small and large caches are combined into one tensor and the position ids are updated to index accordingly. The KV cache re-computation is skipped [here](https://github.com/microsoft/onnxruntime-genai/blob/8e0c074005f30e6ae4893ae1afdbb63e53611d35/src/generators.cpp#L491-L500), however.

PR #1931, opened 2025-12-22
Add Support for Toggling Guidance

conleypri:
## Issue
When JSON guidance constraints are enabled, features like Chain of Thought and Tool Calling are disabled, since the model is forced to only output valid JSON at all times. Note: The Generator object is immutable, and guidance constraints (such as JSON output enforcement) must be set during generator initialization.

## Resolution
This PR introduces the ability to toggle guidance constraints on and off. This allows guidance constraints to be set at generator initialization, but only enforced when needed (e.g., at the Final channel). This enables workflows where Chain of Thought and Tool Calling can proceed unconstrained, and strict JSON output is enforced only at the appropriate stage.

- Enables dynamic toggling of guidance constraints.
- Preserves generator immutability and initialization patterns.
- Supports advanced workflows requiring staged constraint enforcement.

Additional Notes:
- Guidance will be enabled by default and only enforced when SetGuidance() is called with constraint and constraint schema passed in.

### DIFF
--- a/src/csharp/Generator.cs
+++ b/src/csharp/Generator.cs
@@ -121,6 +121,25 @@ namespace Microsoft.ML.OnnxRuntimeGenAI
                                                                    StringUtils.ToUtf8(adapterName)));
         }
 
+        /// <summary>
+        /// Toggles the guidance (constrained decoding) on or off.
+        /// Throws on error.
+        /// </summary>
+        /// <param name="enable">true to enable, false to disable</param>
+        public void ToggleGuidance(bool enable)
+        {
+            Result.VerifySuccess(NativeMethods.OgaGeneratorToggleGuidance(_generatorHandle, enable));
+        }
+
+        /// <summary>
+        /// Returns whether the guidance (constrained decoding) is enabled.
+        /// </summary>
+        /// <returns>true if guidance is enabled, false otherwise</returns>
+        public bool IsGuidanceEnabled()
+        {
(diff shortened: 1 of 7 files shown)


kunal-vaishnavi:
Thank you for your contribution! You can still dynamically control when and how guidance is applied through the Lark grammar that you construct.

For example, Phi-4 mini reasoning has a Chain-of-Thought (CoT) process that reasons through before beginning to return output. Here is an example of how you can write the grammar to enable the model to think before returning a tool call via JSON output.

```
start: CoT toolcall
CoT: <think> TEXT </think>
TEXT: /[^{<](.|\\n)*/
toolcall: <|tool_call|> functioncall <|/tool_call|>
functioncall: %json {}
```

This will allow the CoT tokens to be produced before applying a JSON schema to constrain the output.

You can write multiple conditionals to ensure more flexibility in the generated output. Here is another example where the output after CoT can be text or a tool call.

```
start: CoT (TEXT | toolcall)
CoT: <think> TEXT </think>
TEXT: /[^{<](.|\\n)*/
toolcall: <|tool_call|> functioncall <|/tool_call|>
functioncall: %json {}
```

As long as you construct enough options in your grammar, you should be able to cover all of your cases without needing to toggle guidance on or off. Therefore, the changes in this PR should not be necessary.

You can find more information about the available features in a Lark grammar [here](https://github.com/guidance-ai/llguidance/blob/main/docs/syntax.md).

PR #1933, opened 2025-12-23
Add create guards for ort and ort-genai targets

sheldonrobinson:
`slm-engine` is sub-project. Check before doing `add_library` for `ort` and `ort-genai`


sheldonrobinson:

PR #1935, opened 2026-01-07
AMD RyzenAI EP Support

akholodnamdcom:
Enabling initial support of a lightweight AMD RyzenAI EP DLL. 
Unlike VitisAI EP, it's hybrid (runs same model on both NPU and GPU).

### DIFF
--- a/src/models/model.cpp
+++ b/src/models/model.cpp
@@ -918,7 +918,7 @@ Model::Model(std::unique_ptr<Config> config) : config_{std::move(config)} {
   CreateSessionOptions();
   EnsureDeviceOrtInit(*p_device_, *config_, arena_cfg_);
 
-  // Only CUDA, TRT-RTX and DML does every input on the device
+  // Only CUDA, TRT-RTX, RyzenAI and DML does every input on the device
   // For WebGPU, use device memory only if graph capture is enabled, otherwise use CPU
   if (p_device_->GetType() == DeviceType::CUDA || p_device_->GetType() == DeviceType::DML || p_device_->GetType() == DeviceType::NvTensorRtRtx ||
       p_device_->GetType() == DeviceType::RyzenAI ||


src/models/model.cpp
```diff
@@ -829,7 +836,7 @@ void EnsureDeviceOrtInit(DeviceInterface& device, const Config& config, std::uni
   allocator.session_ = OrtSession::Create(GetOrtEnv(), g_trivial_model, sizeof(g_trivial_model), session_options.get());
 
   // Names for the device memory types used by 'OrtMemoryInfo::Create'
-  static const char* device_memory_type_names[] = {"CPU (Not used, see above)", "Cuda", "DML", "WebGPU_Buffer", "QnnHtpShared", "OpenVINO (Not used, see above)", "Cuda"};
+  static const char* device_memory_type_names[] = {"CPU (Not used, see above)", "Cuda", "DML", "WebGPU_Buffer", "QnnHtpShared", "OpenVINO (Not used, see above)", "Cuda", "Cpu"};
```
kunal-vaishnavi, line 839: Why does CPU need to be added twice here?

src/models/model.cpp
```diff
@@ -921,6 +928,9 @@ Model::Model(std::unique_ptr<Config> config) : config_{std::move(config)} {
 
   // The kvcache is always allocated in device memory
   p_device_kvcache_ = p_device_;
+
+  if (p_device_->GetType() == DeviceType::RyzenAI)
```
kunal-vaishnavi: Can you add `RyzenAI` in the `if` case below for `p_device_inputs_ = p_device_` instead?

https://github.com/microsoft/onnxruntime-genai/blob/b4f80f22d11c24fb1f51b40f2715efb43c59a1b6/src/models/model.cpp#L921-L927

The line above will set `p_device_kvcache_ = p_device_` automatically. Then this separate `if` condition is not needed.

src/models/model.cpp
akholodnamdcom: For every `DeviceType` there should be memory type here in this array.

```
enum struct DeviceType {
  CPU,
  CUDA,
  DML,
  WEBGPU,
  QNN,
  OpenVINO,
  NvTensorRtRtx,
  RyzenAI,
  MAX
};
```

Few lines below device type is used as index for this array:
`auto name = device_memory_type_names[static_cast<int>(type)];`

src/ryzenai/interface.cpp
```diff
@@ -0,0 +1,184 @@
...
+
+ private:
+  std::filesystem::path ep_path_;
+};
+
+static std::unique_ptr<Interface> interface_;
+
+}  // namespace RyzenAI
+
+void RyzenAIInterface::Shutdown() {
```
baijumeswani, line 196: Once shutdown is invoked, can the interface not be recreated?

src/models/logits.cpp
```diff
@@ -15,8 +15,8 @@ Logits::Logits(State& state)
 
   input_sequence_lengths.resize(state_.params_->search.batch_size);
 
-  if (IsOpenVINOStatefulModel(state.model_)) {
-    // In the case of OpenVINO stateful models, they are patched in a way so that they only return the
+  if (IsOpenVINOStatefulModel(state.model_) || state_.model_.p_device_->GetType() == DeviceType::RyzenAI) {
+    // In the case of OpenVINO stateful models or RyzenAI models, they are patched in a way so that they only return the
```
baijumeswani, line 19: Is this a model property or is this an EP property? Can the shape of the output logits be determined through the shape of the output in the onnx model?

src/ryzenai/interface.cpp
akholodnamdcom: Yes. 
But that should be okay, considering how OrtGlobals::Shutdown (which calls our Shutdown) is invoked. 
Right?

src/models/model.cpp
baijumeswani: Can you explain why CPU is the device memory type for RyzenAI? Is it because you expect to allocate inputs and outputs on Cpu for this device?

src/ryzenai/interface.cpp
```diff
@@ -0,0 +1,184 @@
...
+      if (const auto func = reinterpret_cast<void (*)()>(GetProcAddress(mod, func_shutdown_)))
+        func();
+#endif  // _WIN32
+  }
+
+  void SetupProvider(OrtSessionOptions& session_options, const ProviderOptions& provider_options) override {
+    std::vector<const OrtEpDevice*> supported;
+
+    {
+      const OrtEpDevice* const* devs = nullptr;
```
baijumeswani: Could you please use `devices` instead of `devs`?

src/models/logits.cpp
akholodnamdcom: As of now there is no need to analyze model dynamically. 
Currently it's safe to say that all the models that we'll deliver for this EP are going to have logits "pruned". 
Later we'll probably extend this check.

src/models/model.cpp
akholodnamdcom: That's an interesting question. 
The code was written quite a long time ago and I can't say for sure why it's CPU. 
Probably I should actually change it to our actual memory type. 
Checking....

src/ryzenai/interface.cpp
```diff
@@ -0,0 +1,184 @@
...
+  void CopyDeviceToCpu() override {}
+  void CopyCpuToDevice() override {}
+
+  void CopyFrom(size_t begin_dest, DeviceBuffer& source, size_t begin_source, size_t size_in_bytes) override {
+    CopyThroughCpu(*this, begin_dest, source, begin_source, size_in_bytes);
+  }
+
+  void Zero() override {
+    memset(p_device_, 0, size_in_bytes_);
+  }
```
baijumeswani, line 55: RyzenAI EP device memory is host accessible?

src/models/model.cpp
```diff
@@ -914,6 +921,7 @@ Model::Model(std::unique_ptr<Config> config) : config_{std::move(config)} {
   // Only CUDA, TRT-RTX and DML does every input on the device
   // For WebGPU, use device memory only if graph capture is enabled, otherwise use CPU
   if (p_device_->GetType() == DeviceType::CUDA || p_device_->GetType() == DeviceType::DML || p_device_->GetType() == DeviceType::NvTensorRtRtx ||
+      p_device_->GetType() == DeviceType::RyzenAI ||
```
baijumeswani, line 924: If RyzenAI memory is host accessible, should p_device_inputs_ = GetDeviceInterface(DeviceType::CPU);?

src/models/model.cpp
akholodnamdcom: Ok. It does not work if I use our EP memory type ("RMM"). 

Problem is that `OrtMemoryInfo::Create` (that you use here) is not aware of memory types/allocators coming from EP Plugin DLLs. See impl: https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/core/framework/allocator.cc#L209-L282

But anyway, that's not a problem for our new EP, coz it's pretty flexible.

akholodnamdcom: Once you switch to OrtApis::CreateMemoryInfo_V2, probably we're gonna be able to specify "RMM" here. 
https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/core/framework/allocator.cc#L284-L324

src/ryzenai/interface.cpp
akholodnamdcom: Yes, it is.

src/models/model.cpp
akholodnamdcom: Even so it's better to have it allocated via our allocator which under hood uses DX API to allocate the shared/unified memory.

src/ryzenai/interface.cpp
akholodnamdcom: Done

src/models/model.cpp
baijumeswani: It is likely because you do p_device_inputs = p_device. You should set p_device_inputs as CPU. See others that are similar like CPU and OpenVino.

baijumeswani: See this line: https://github.com/microsoft/onnxruntime-genai/blob/3be7b360929ab105ab245350d4b0b0e50796cba4/src/models/model.cpp#L800

src/models/model.cpp
baijumeswani: It seems you're not using the RyzenAI allocator. It seems to be using the cpu allocator based on the other thread.

src/models/model.cpp
akholodnamdcom: Im afraid even if "Cpu" is used for memory type, our custom allocator is created here. 

I did an experiment to confirm that:
image
image

akholodnamdcom: > See this line:
> 
> https://github.com/microsoft/onnxruntime-genai/blob/3be7b360929ab105ab245350d4b0b0e50796cba4/src/models/model.cpp#L800

Doing like that does not work (i see a crash somewhere):
image

src/models/model.cpp
akholodnamdcom: As per the experiment in the other thread, our RMM allocator is somehow actually used.

src/models/model.cpp
baijumeswani: It is very interesting that ort creates a ryzen ep allocator when the provided device name is Cpu. Does that happen inside the EP library?

src/ryzenai/interface.cpp
baijumeswani: Yes, that should be fine. Is the shutdown of the interface required?

src/ryzenai/interface.cpp
baijumeswani: Could you please leave a comment in the code to let others know that ryzen ai ep device memory is cpu accessible.

src/models/model.cpp
baijumeswani: Could you update this comment two lines above.

```
// Only CUDA, TRT-RTX and DML does every input on the device
```

to include ryzenai ep as well.

src/ryzenai/interface.cpp
akholodnamdcom: Currently, yes. It actually releases DX resources. If that shutdown is not called, we run into a strange crash lower in stack. Once we figure out how to fix the crash, we can get rid of this method.

src/ryzenai/interface.cpp
akholodnamdcom: Sure. Will update the interface header with suggested note.

src/models/model.cpp
akholodnamdcom: In our implementation of `onnxruntime::IExecutionProvider::CreatePreferredAllocators` we return our custom allocator with type `OrtAllocatorType::OrtDeviceAllocator` and name set to "RMM". 
This instance is stored in `onnxruntime::SessionState::allocators_`.
Later, when OGA calls `Ort::Allocator::Create`, ORT actually just extracts our allocator from the SessionState object. ORT does not use name ("RMM") to match correct allocator. It uses OrtDevice, which in our case is default and consists of zeros. Since OGA requests default device allocator, it just works regardless the name.

baijumeswani: Would that mean changing the name to RMM here instead of CPU should work? If that works, it might be less confusing. If it does not, we can leave it as is.

akholodnamdcom: Unfortunately, it won't work when I change name to "RMM" here.

PR #1938, opened 2026-01-07
Fix sliding window alignment regression in QNN models

apsonawane:
Revert back the alignment logic to the original way it was in 0.11.2 since it works for all the models.
For FARA model we will update the genai_config alignment value to "left"

### DIFF
--- a/src/generators.cpp
+++ b/src/generators.cpp
@@ -318,14 +318,14 @@ DeviceSpan<int32_t> Generator::AllocateInputIdsOnDevice(cpu_span<const int32_t> 
 
   auto input_ids_device = state_->params_->p_device->Allocate<int32_t>(padded_input_ids_size);
   auto cpu_span = input_ids_device.CpuSpan();
-
-  // Handle padding based on alignment setting for sliding window models
-  if (padded_input_ids_size > input_ids.size()) {
-    std::copy(input_ids.begin(), input_ids.end(), cpu_span.begin());
-    std::fill(cpu_span.begin() + input_ids.size(), cpu_span.end(), model_->config_->model.pad_token_id);
-  } else {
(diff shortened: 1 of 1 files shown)


src/generators.cpp
```diff
@@ -321,18 +321,8 @@ DeviceSpan<int32_t> Generator::AllocateInputIdsOnDevice(cpu_span<const int32_t>
...
-      // Left alignment: padding first, then data
-      std::fill_n(cpu_span.begin(), padded_input_ids_size - input_ids.size(), model_->config_->model.pad_token_id);
-      std::copy(input_ids.begin(), input_ids.end(), cpu_span.begin() + (padded_input_ids_size - input_ids.size()));
-    } else {
-      // Right alignment (default): data first, then padding
-      std::copy(input_ids.begin(), input_ids.end(), cpu_span.begin());
-      std::fill(cpu_span.begin() + input_ids.size(), cpu_span.end(), model_->config_->model.pad_token_id);
-    }
+    std::copy(input_ids.begin(), input_ids.end(), cpu_span.begin());
+    std::fill(cpu_span.begin() + input_ids.size(), cpu_span.end(), model_->config_->model.pad_token_id);
```
baijumeswani: This breaks QC AI Hub models. Which need padding first and data next.

apsonawane: What is the alignment value in genai_config for QC AI hub models?

kunal-vaishnavi: This looks different than the original version before Fara was supported.

https://github.com/microsoft/onnxruntime-genai/blob/25962b07e19d91fdae570d776228b7425325e433/src/generators.cpp#L319-L330

Should we revert to this version and then modify it for Fara again?

apsonawane: Reverted back to the original and updated the genai_config of Fara model

PR #1939, opened 2026-01-07
Rewrite all examples using standardization

kunal-vaishnavi:
### Description

This PR rewrites all examples in C/C++, C#, and Python to use standardized approaches for running inference with ONNX Runtime GenAI. It also removes the SLM engine, chat app, and Genny examples since they are now deprecated and not maintained. Users can try out [Foundry Local](https://foundrylocal.ai), which is powered by ONNX Runtime GenAI, if interested in similar examples.

### Motivation and Context

Examples across the languages and across functionalities are often out-of-sync. For example, a Python example may get updated but its equivalent C/C++ and C# examples do not. This PR fixes this by synchronizing feature parity across the various examples.

We can use this standardization to continue maintaining feature parity in the future.

### DIFF
--- a/test/csharp/TestOnnxRuntimeGenAIAPI.cs
+++ b/test/csharp/TestOnnxRuntimeGenAIAPI.cs
@@ -161,9 +161,9 @@ namespace Microsoft.ML.OnnxRuntimeGenAI.Tests
                             generator.AppendTokens(inputIDs);
 
                             Assert.False(generator.IsDone());
-                            Assert.Equal(generatorParams.GetSearchNumber("max_length"), maxLength);
-                            Assert.Equal(generatorParams.GetSearchBool("early_stopping"), true);
-                            Assert.Equal((int)generator.TokenCount(), generator.GetSequence(0).Length);
+                            Assert.Equal(maxLength, generatorParams.GetSearchNumber("max_length"));
+                            Assert.Equal(true, generatorParams.GetSearchBool("early_stopping"));
+                            Assert.Equal(generator.GetSequence(0).Length, (int)generator.TokenCount());
 
                             while (!generator.IsDone())
                             {
@@ -175,7 +175,7 @@ namespace Microsoft.ML.OnnxRuntimeGenAI.Tests
                                 var sequence = generator.GetSequence(i).ToArray();
                                 var expectedSequence = expectedOutput.Skip((int)i * (int)maxLength).Take((int)maxLength);
                                 Assert.Equal(expectedSequence, sequence);
-                                Assert.Equal((int)generator.TokenCount(), generator.GetSequence(i).Length);
+                                Assert.Equal(generator.GetSequence(0).Length, (int)generator.TokenCount());
                             }
                         }
                     }


.github/workflows/linux-gpu-x64-build.yml
```diff
@@ -136,7 +136,23 @@ jobs:
...
+        run: |
+          set -e -x
+          docker run \
+            --gpus all \
+            --rm \
+            --user 0 \
+            --volume $GITHUB_WORKSPACE:/ort_genai_src \
+            -w /ort_genai_src onnxruntimecudabuildx64 bash -c " \
+              alias python3=${{ env.PYTHON_EXECUTABLE }} && \
+              dnf -y update && dnf install -y python3.11-devel && dnf install -y python3-pip python3-setuptools python3-wheel && \
```
baijumeswani, line 149: Should all this happen inside https://github.com/microsoft/onnxruntime-genai/blob/main/tools/ci_build/github/linux/docker/manylinux/Dockerfile.manylinux2_28_cuda_12.2?

baijumeswani: We could perhaps do this later so this PR can go in.

test/python/special_tokens.py
```diff
@@ -0,0 +1,35 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+import argparse
+import os
+
```
baijumeswani, line 8: Could you add a comment here explaining why running this script is required (for context)?

nuget.config
```diff
@@ -1,7 +1,8 @@
-<?xml version="1.0" encoding="utf-8"?>
+﻿<?xml version="1.0" encoding="utf-8"?>
 <configuration>
   <packageSources>
     <clear />
     <add key="ORT-Nightly" value="https://aiinfra.pkgs.visualstudio.com/PublicPackages/_packaging/ORT-Nightly/nuget/v3/index.json" />
+    <add key="NuGet Official" value="https://api.nuget.org/v3/index.json" />
```
baijumeswani: Why is this necessary? It may confuse our CI/packaging system with multiple feeds.

examples/python/phi4-mm.py
```diff
@@ -186,10 +186,10 @@ def run(args: argparse.Namespace):
         "-pr", "--prompt", required=False, help="Input prompts to generate tokens from, mainly for CI usage"
     )
     parser.add_argument(
-        "--non-interactive",
+        "--non_interactive",
         action=argparse.BooleanOptionalAction,
         required=False,
-        help="Non-interactive mode, mainly for CI usage",
+        help="non_interactive mode, mainly for CI usage",
```
baijumeswani: Could the help mention `Non interactive` instead of the option itself?

examples/python/model-chat.py
```diff
@@ -3,182 +3,79 @@
...
-                "additionalProperties": False,
-            },
-        },
-    }
-    if len(param_props) == 0:
-        output_schema["required"] = ["name"]
-    return output_schema
-
+from common import apply_chat_template, get_config, get_generator_params_args, get_guidance, get_guidance_args, get_search_options
+# og.set_log_options(enabled=True, model_input_values=inputs, model_output_values=outputs)
```
baijumeswani: Instead of commenting this, could we add a --debug (or --dev_debug) in the arg parser to enable/disable this?

examples/csharp/Common/Common.cs
```diff
@@ -0,0 +1,852 @@
...
+                        else
+                        {
+                            config.SetProviderOption(ep, k, v);
+                        }
+                    }
+                }
+            }
+
+            /**
+             * TODO: Uncomment the below snippet to use config.Overlay once the C# binding to Config.Overlay
```
baijumeswani, line 110: Now that overlay is available, should be uncomment this?

examples/c/src/whisper.cpp
```diff
@@ -161,8 +163,8 @@ void C_API(const char* model_path, int32_t num_beams) {
       std::cout << str << std::endl;
     }
 
-    for (int i = 0; i < 3; ++i)
-      std::cout << std::endl;
+    std::cout << "\n\n"
+              << std::endl;
```
baijumeswani, line 166:
```suggestion
    std::cout << "\n\n\n";
```

or

```
std::cout << std::endl << std::endl << std::endl;
```

examples/c/README.md
```diff
@@ -67,3 +66,26 @@ For other options to download models, read through [our download options](https:
 2. On Linux and macOS:
    - cd build
    - ./model_qa <path/to/model/from/above> <execution_provider>
+
+## Use constrained decoding
```
baijumeswani: Should we move this into an md of its own instead of duplicating it for each language readme?

examples/c/CMakeLists.txt
```diff
@@ -3,9 +3,27 @@ cmake_minimum_required(VERSION 3.18.1)
...
+  GIT_TAG        v2.6.1      # Or update to latest release
+)
+FetchContent_MakeAvailable(CLI11)
+
 option(USE_CXX "Invoke the C++ example" ON)
 option(MODEL_CHAT "Build the Model Chat example" OFF)
-option(MODEL_QA "Build the Model Q&A example without multi-turn prompting" OFF)
+option(MODEL_QA "Build the Model Q&A example" OFF)
 option(MODEL_VISION "Build the Model Vision example" OFF)
 option(PHI4-MM "Build the Phi-4 mm example" OFF)
```
baijumeswani: Should we also rename phi4-mm to model-mm? Does it have phi4 specific details? Could we factor it out and make it more generic?

.pipelines/stages/jobs/nuget-validation-job.yml
```diff
@@ -177,14 +177,14 @@ jobs:
...
-      WorkingDirectory: '$(Build.Repository.LocalPath)/examples/csharp/HelloPhi4MM'
+      WorkingDirectory: '$(Build.Repository.LocalPath)/examples/csharp/ModelChat4MM'
       HuggingFaceToken: $(HF_TOKEN)
       os: ${{ parameters.os }}
 
   - template: steps/nuget-validation-step.yml
     parameters:
-      CsprojFolder: "examples/csharp/HelloPhi4MM"
-      CsprojName: "HelloPhi4MM"
+      CsprojFolder: "examples/csharp/ModelChat4MM"
```
baijumeswani: Should this one still be Phi4MM?

.github/workflows/linux-gpu-x64-build.yml
kunal-vaishnavi: There was an issue with Python versioning in the Docker container, so I ended up adding it here. It also narrows the scope for adding these additional packages to just the container for the Java build.

test/python/special_tokens.py
kunal-vaishnavi: I already added a comment where this file is called.

https://github.com/microsoft/onnxruntime-genai/blob/9272d1597a9bf35a81a4bef8e7fc45d13dc70518/test/python/test_onnxruntime_genai_e2e.py#L96

I'll add it at the top of this file as well.

nuget.config
kunal-vaishnavi: It was for the `System.CommandLine` package. I'll add it to the ORT nightly feed and remove this feed.

examples/python/phi4-mm.py
kunal-vaishnavi: Typo from search and replace. Fixed it.

examples/csharp/Common/Common.cs
kunal-vaishnavi: The overlay API is not in an official C# NuGet release and the examples rely upon a stable release.

examples/c/src/whisper.cpp
kunal-vaishnavi: Fixed to use the suggested change

examples/c/CMakeLists.txt
kunal-vaishnavi: Refactored all of the multi-modal examples and fixed the chat template issues that needed to be resolved first.

.pipelines/stages/jobs/nuget-validation-job.yml
kunal-vaishnavi: Typo from search and replace. Fixed it.

examples/python/model-chat.py
kunal-vaishnavi: There's a `set_logger` function in `common.py` to enable this. I'll add a flag that calls that and remove the commented-out line from each example.

examples/c/README.md
kunal-vaishnavi: The examples all have duplicate content since the CLI args are now synced up. I think having a section in each is fine for now. Users will likely go through an example's README to see how to run it and then discover that tool calling is supported. There are many requests for showing how to do tool calling across the language bindings in our GitHub issues and discussion posts.

We can add a general-purpose guide to constrained decoding and structured outputs in a `docs` folder.

src/csharp/NativeMethods.cs
```diff
@@ -329,6 +333,10 @@ public static extern UIntPtr OgaSequencesGetSequenceCount(IntPtr /* const OgaSeq
         [DllImport(NativeLib.DllName, CallingConvention = CallingConvention.Winapi)]
         public static extern IntPtr /* OgaResult* */ OgaGetCurrentGpuDeviceId(out IntPtr /* int32_t */ deviceId);
 
+        [DllImport(NativeLib.DllName, CallingConvention = CallingConvention.Winapi)]
+        public static extern IntPtr /* OgaResult* */ OgaRegisterExecutionProviderLibrary(byte[] /* const char* */ name,
+                                                                                         byte[] /* const char* */ path);
```
baijumeswani: This was intentionally left out as it is not needed. Users wanting to register execution provider library must do so using the ort env. Once the library has been registered through ort, it will automatically be available inside ort-genai (since ort env is a singleton).

The C api was added primarily for Python because Python is a special case where users do not get access to the ort env through ort python package (different ort dlls are used).

kunal-vaishnavi: Changed it to follow the examples shown in [your PR](https://github.com/microsoft/onnxruntime-genai/pull/1850)

baijumeswani (approved):

PR #1942, opened 2026-01-08
[RyzenAI] Non-pruned models backward compatibility

akholodnamdcom:
This improvement makes it possible to run previously released RyzenAI Hybrid models with the new RyzenAI EP.

### DIFF
--- a/src/generators.cpp
+++ b/src/generators.cpp
@@ -1,6 +1,6 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
-// 
+//
 // Modifications Copyright(C) 2026 Advanced Micro Devices, Inc. All rights reserved.
 
 #include "generators.h"
--- a/src/models/logits.cpp
+++ b/src/models/logits.cpp
@@ -1,6 +1,6 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
-// 
+//
 // Modifications Copyright(C) 2026 Advanced Micro Devices, Inc. All rights reserved.
 #include "../generators.h"
 #include "model.h"
--- a/src/models/model.h
+++ b/src/models/model.h
@@ -1,6 +1,6 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
-// 
+//
 // Modifications Copyright(C) 2026 Advanced Micro Devices, Inc. All rights reserved.
 #pragma once
 #include "model_type.h"
--- a/src/smartptrs.h
+++ b/src/smartptrs.h
@@ -1,6 +1,6 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
-// 
+//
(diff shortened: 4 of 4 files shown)


src/ryzenai/interface.cpp
```diff
@@ -207,4 +208,18 @@ RyzenAIInterface* GetRyzenAIInterface() {
   return RyzenAI::interface_.get();
 }
 
+bool IsRyzenAIPrunedModel(const Model& model) {
```
baijumeswani, line 213: This function comprises of two checks. Could we move the second check into a function of its own?
- IsRyzenAIModel() - can stay in this function
- IsPrunedModel() - move it into a function of its own

I feel that IsPrunedModel() may be useful for other EPs as well. You can keep this function, but move the IsPrunedModel() part out of this function (into some common file). All this function then does is

```
return model.p_device_->GetType() == DeviceType::RyzenAI && IsPrunedModel();
```

akholodnamdcom: Added `Model::IsPruned()` as requested.

Also, I had to push copyrights that I was supposed to put with prev PR. 

Thanks!

src/models/model.cpp
```diff
@@ -1150,6 +1164,14 @@ std::shared_ptr<MultiModalProcessor> Model::CreateMultiModalProcessor() const {
   return std::make_shared<MultiModalProcessor>(*config_, session_info_);
 }
 
+bool Model::IsPruned() const {
+  const auto& logits_name = config_->model.decoder.outputs.logits;
+  if (!session_info_.HasOutput(logits_name))
+    return false;
+  const auto logits_shape = session_info_.GetOutputShape(logits_name);
```
kunal-vaishnavi, line 1171: What would happen if the RyzenAI model does not have a fixed sequence length of 1 in the ORT session and instead has a dynamic sequence length but the size of that dimension in the output during inference is 1?

akholodnamdcom: Thanks for the question @kunal-vaishnavi.

Looks like you are describing token generation phase, when model produces just one token per run. It works well. 

Never tried to run prefill with just one token (prompt length = 1). Pretty likely it should be okay, like in token generation phase. And probably our internal QA tried this test case earlier.

kunal-vaishnavi: My question is regarding how the output shape is constructed in the RyzenAI ONNX models. If the model output's shape is listed as `('batch_size', 'sequence_length', 'vocab_size')` but the resulting tensor is `(batch_size, 1, vocab_size)` (e.g. prefill with one token), then `IsPruned` is false instead of true. Would that be fine?

akholodnamdcom: Yeah, that could be an interesting corner case. Thanks for heads up! 

I verified how it behaves with this PR and latest EP. 

Used a model with logits shape like that: batch_size,sequence_length,32000.
Used one-token prompt: "?". To be sure, I confirmed that `num_prompt_tokens` in model_benchmark is 1 (see screenshot).

As expected, IsPruned returned false for that test conditions.

Luckly, it worked well. The output was okay (see screenshot). 

image

kunal-vaishnavi: Sounds good. Thanks for confirming that it works.

PR #1944, opened 2026-01-15
Set default top_k and top_p if it is None

xiaoyu-work:
Set default top_k and top_p if it is None.

Exporting Qwen VL model will get 
```
"top_k": null,
"top_p": null
```

### DIFF
--- a/src/python/py/models/builders/base.py
+++ b/src/python/py/models/builders/base.py
@@ -551,7 +551,7 @@ class Model:
             }
             for key, default_val in defaults.items():
                 val = getattr(gen_config, key)
-                if val != default_val:
+                if val is not None and val != default_val:
                     setattr(config, key, getattr(gen_config, key))
         except:
             pass


apsonawane (dismissed):

src/python/py/models/builders/base.py
```diff
@@ -623,8 +623,8 @@ def make_genai_config(self, model_name_or_path, extra_kwargs, out_dir):
                 else self.past_present_share_buffer,
                 "repetition_penalty": config.repetition_penalty if hasattr(config, "repetition_penalty") else 1.0,
                 "temperature": config.temperature if hasattr(config, "temperature") else 1.0,
-                "top_k": config.top_k if hasattr(config, "top_k") else 50,
-                "top_p": config.top_p if hasattr(config, "top_p") else 1.0,
+                "top_k": config.top_k if hasattr(config, "top_k") and config.top_k is not None else 50,
+                "top_p": config.top_p if hasattr(config, "top_p") and config.top_p is not None else 1.0,
```
baijumeswani, line 627: Should we be setting these here or should they be using the defaults as defined in the config when not mentioend? 

https://github.com/microsoft/onnxruntime-genai/blob/c37bc36ba9968f11f78f6707db0109fa73210507/src/config.h#L301-L302

src/python/py/models/builders/base.py
```diff
@@ -623,8 +623,8 @@ def make_genai_config(self, model_name_or_path, extra_kwargs, out_dir):
                 else self.past_present_share_buffer,
                 "repetition_penalty": config.repetition_penalty if hasattr(config, "repetition_penalty") else 1.0,
                 "temperature": config.temperature if hasattr(config, "temperature") else 1.0,
-                "top_k": config.top_k if hasattr(config, "top_k") else 50,
-                "top_p": config.top_p if hasattr(config, "top_p") else 1.0,
+                "top_k": config.top_k if hasattr(config, "top_k") and config.top_k is not None else 50,
```
kunal-vaishnavi, line 626: Can we update this to fix the default values while ensuring [this](https://github.com/microsoft/onnxruntime-genai/pull/1693) is still valid?

https://github.com/microsoft/onnxruntime-genai/blob/0c618767d4154d840b72f537b38dc19245ae4e21/src/python/py/models/builders/base.py#L538-L557

src/python/py/models/builders/base.py
xiaoyu-work: The existing pattern in this codebase is to write all search parameters with their default values to genai_config.json, rather than relying on config.h defaults. The whole logic here should probably be refactored

src/python/py/models/builders/base.py
xiaoyu-work: Added a none check for gen_config

PR #1949, opened 2026-01-19
Add `enable_profiling` in Runtime Options

xiaofeihan1:
# Background
Previously, developers could only enable profiling through genai_config.json, which profiles the entire lifetime of a session—from session creation to session destruction. When the number of runs is large (for example, generating ~3000 tokens during inference), the profiling output can become excessively large, causing profiling to fail due to oversized files.

> 2025-12-17 11:17:36.496 Python[9861:2179219] 2025-12-17 11:17:36.493043 [E:onnxruntime:onnxruntime-genai, profiler.cc:93 EndTimeAndRecordEvent] Maximum number of events reached, could not record profile event.

# Description
This PR addresses this limitation by introducing run-level profiling. We expose `enable_profiling` in `RuntimeOptions`. Developers can now enable profiling for specific runs only.

With the following code, two profiling JSON files will be generated using the `run_profiler_file` prefix:
	•	one containing profiling data for the 100th run
	•	another containing profiling data for the 101th run

```
    if len(new_tokens) == 100 or len(new_tokens) == 101 :
        generator.set_runtime_option("enable_profiling", "run_profiler_file")
    else:
        generator.set_runtime_option("enable_profiling", "0")
    new_token = generator.get_next_tokens()[0]
    new_tokens.append(new_token)
```


.github/workflows/win-cpu-arm64-build.yml
```diff
@@ -94,7 +94,7 @@ jobs:
           # Uninstalling LLVM/Clang as it is no longer required and causes issues with numpy installation
           choco uninstall llvm --yes
           python -m pip install "numpy<2" coloredlogs flatbuffers packaging protobuf sympy pytest
-          python -m pip install onnxruntime-qnn
+          python -m pip install onnxruntime-qnn==1.25.0.dev20260126001 -i https://aiinfra.pkgs.visualstudio.com/PublicPackages/_packaging/ORT-Nightly/pypi/simple/
```
kunal-vaishnavi: Do we need to specify a version here or would not listing a version and using the latest nightly package be sufficient?

.pipelines/nuget-publishing.yml
```diff
@@ -61,7 +61,7 @@ parameters:
   - name: ort_version
     displayName: 'OnnxRuntime version'
     type: string
-    default: '1.23.0'
+    default: '1.25.0-dev-20260125-1205-727db0d3dc'
```
kunal-vaishnavi: These are default values for publishing the official packages for ORT GenAI. I think we should keep the defaults as a stable version of ORT. We can always override what version of ORT to use when the packages are built.

src/models/model.cpp
```diff
@@ -173,6 +173,19 @@ void State::SetRunOption(const char* key, const char* value) {
       throw std::runtime_error(std::string("terminate_session key value unexpected: ") + value);
     }
     return;
+  } else if (strcmp(key, "enable_profiling") == 0) {
+    if (strcmp(value, "0") == 0) {
+      run_options_->DisableProfiling();
+    } else if (strcmp(value, "1") == 0) {
+      run_options_->EnableProfiling(ORT_TSTR("onnxruntime_run_profile"));
+    } else {
+      auto ToProfileString = [](const char* s) -> std::basic_string<ORTCHAR_T> {
```
kunal-vaishnavi: 1. Why is converting a `char*` to `basic_string<ORTCHAR_T>` before going back to a `char*` needed?
2. Can you add a comment here to explain that this `else` condition is for a custom prefix for the log file?
3. Can the `else if` and `else` conditions be merged since they both enable profiling?

.github/workflows/linux-gpu-x64-build.yml
```diff
@@ -20,7 +20,7 @@ env:
 jobs:
   linux-cuda-x64-build:
     env:
-      PYTHON_EXECUTABLE: "/opt/python/cp310-cp310/bin/python3.10"
+      PYTHON_EXECUTABLE: "/opt/python/cp311-cp311/bin/python3.11"
```
baijumeswani: Could we move to 3.12?

src/models/model.cpp
xiaofeihan1: For items 2 and 3, I’ve made the changes.
For item 1, I’m not entirely sure I understand it.

The value passed to State::SetRunOption is a char*:

```
State::SetRunOption(const char* key, const char* value)
```
However, the EnableProfiling API expects an ORTCHAR_T*:
```
inline OrtRunOptions& OrtRunOptions::EnableProfiling(const ORTCHAR_T* profile_file_prefix)
```

That’s why I used a lambda to convert char* to ORTCHAR_T* at this point. Do you have a better suggestion for handling this conversion?

.github/workflows/win-cpu-arm64-build.yml
xiaofeihan1: Good suggestion! I removed the version here.

src/models/model.cpp
kunal-vaishnavi: I checked and `ToProfileString(...).c_str()` returns a `const ORTCHAR_T*` whereas I was thinking it was producing a `const char*`. This looks fine to me.

baijumeswani: Please use 

```
run_options_->EnableProfiling(fs::path(prefix).c_str());
```

src/models/model.cpp
```diff
@@ -173,6 +173,21 @@ void State::SetRunOption(const char* key, const char* value) {
...
+      // Enable run-level profiling. The profiling output file is named: <prefix>_<timestamp>.json
+      // Value "1" uses the default prefix; any other value is treated as a custom prefix.
+      // Convert to basic_string<ORTCHAR_T> for cross-platform compatibility (wchar_t on Windows, char elsewhere).
+      auto ToProfileString = [](const char* s) -> std::basic_string<ORTCHAR_T> {
+        std::string str(s);
+        return std::basic_string<ORTCHAR_T>(str.begin(), str.end());
+      };
+      const char* prefix = (strcmp(value, "1") == 0) ? "onnxruntime_run_profile" : value;
+      run_options_->EnableProfiling(ToProfileString(prefix).c_str());
+    }
```
```suggestion
      constexpr const char* default_profile_prefix = "onnxruntime_run_profile";
      const char* prefix = (strcmp(value, "1") == 0) ? default_profile_prefix : value;
      run_options_->EnableProfiling(fs::path(prefix).c_str());
    }
```

.github/workflows/linux-gpu-x64-build.yml
xiaofeihan1: Sure, done!

src/models/model.cpp
xiaofeihan1: Thanks for the suggestion. Done!

PR #1952, opened 2026-01-22
Support UpdatePositionIds in webgpu's interface

qjia7:
## Summary

- Implement `UpdatePositionIds` override in the WebGPU device interface to
  avoid an unnecessary GPU→CPU→GPU roundtrip on every decode step
- Without this override, the fallback path copies position_ids from GPU to CPU,
  updates values, then copies back. The override generates values directly on
  CPU and uploads once via `CopyTensors`
- For the common single-token decode case, uses a stack variable to avoid
  heap allocation

## Motivation

`UpdatePositionIds` is called on every decode token. The default fallback
performs a GPU→CPU read, CPU update, then CPU→GPU write. Since we are
writing all values (not reading and modifying), the GPU→CPU read is
unnecessary. This eliminates one GPU↔CPU transfer per decode step.


src/webgpu/interface.cpp
```diff
@@ -261,6 +261,49 @@ struct InterfaceImpl : DeviceInterface {
...
   }
+
+  bool UpdatePositionIds(void* position_ids, int batch_beam_size, int total_length, int new_kv_length, ONNXTensorElementDataType type) override {
+    if (batch_beam_size != 1) {
+      return false;
+    }
+
+    int start = total_length - new_kv_length;
+
+    auto upload = [&]<typename T>(T) {
```
kunal-vaishnavi: Can we create a named helper method instead of an anonymous lambda method?

src/webgpu/interface.cpp
```diff
@@ -261,6 +261,49 @@ struct InterfaceImpl : DeviceInterface {
 
     return true;
   }
+
+  bool UpdatePositionIds(void* position_ids, int batch_beam_size, int total_length, int new_kv_length, ONNXTensorElementDataType type) override {
+    if (batch_beam_size != 1) {
+      return false;
+    }
+
+    int start = total_length - new_kv_length;
```
kunal-vaishnavi, line 277: Can we add some error handling to avoid writing invalid position ids in edge cases (for example: `new_kv_length >= 0`, `total_length >= new_kv_length`, etc)?

kunal-vaishnavi (dismissed):

src/webgpu/interface.cpp

src/webgpu/interface.cpp
qjia7: For invalid cases, simply return false and fall back to the CPU path so they can be handled in a single location.

kunal-vaishnavi (approved):

PR #1953, opened 2026-01-26
Add RAII wrappers for ORT Model Editor API types

qjia7:


src/models/graph_builder.cpp
```diff
@@ -109,116 +109,117 @@ namespace GraphBuilder {
...
-      Ort::api->ReleaseOpAttr(attr);
-    }
-    node_attributes.clear();
-    // Create model with opset
-    const char* domain_name = "";
-    Ort::ThrowOnError(model_editor_api.CreateModel(&domain_name, &config.opset_version, 1, &model));
+  // Create node attributes
+  for (const auto& attr : config.attributes) {
+    node_attributes.push_back(CreateOpAttr(attr));
+  }
```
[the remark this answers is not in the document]
qjia7: CreateNode does NOT copy the attributes - it stores references/pointers to them. Therefore:

The [OrtOpAttr*](vscode-file://vscode-app/c:/Users/jiajiaqin/AppData/Local/Programs/Microsoft%20VS%20Code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) pointers must remain valid until AFTER the model is fully built
We cannot use RAII ([std::unique_ptr](vscode-file://vscode-app/c:/Users/jiajiaqin/AppData/Local/Programs/Microsoft%20VS%20Code/resources/app/out/vs/code/electron-browser/workbench/workbench.html)) which would destroy them immediately when the vector goes out of scope
Manual cleanup is required, but only AFTER AddGraphToModel completes

src/generators.h
```diff
@@ -145,11 +144,7 @@ struct OrtGlobals {
 
   // Cache for dynamically built graph sessions (e.g., Cast, TopK operations)
   // Destroyed before env_ to ensure proper cleanup order
-  struct SessionCache {
-    std::unordered_map<uint64_t, std::unique_ptr<OrtSession>> sessions_;
-    std::mutex mutex_;
-  };
-  SessionCache graph_session_cache_;
+  std::unordered_map<uint64_t, std::unique_ptr<OrtSession>> graph_session_cache_;
```
baijumeswani: I think I was mistaken earlier in suggesting we avoid using a mutex here. Even though execution in ort-genai is single‑threaded, these globally scoped caches are still shared across multiple Model instances, so it’s safer to protect them. Sorry for the inconvenience. Could you please revert this change?

src/models/graph_builder.cpp
```diff
@@ -109,116 +109,117 @@ namespace GraphBuilder {
...
-    Ort::ThrowOnError(model_editor_api.SetGraphInputs(graph, graph_inputs.data(), graph_inputs.size()));
-    Ort::ThrowOnError(model_editor_api.SetGraphOutputs(graph, graph_outputs.data(), graph_outputs.size()));
+  // Use RAII wrappers for automatic cleanup
+  std::unique_ptr<OrtGraph> graph;
+  std::unique_ptr<OrtModel> model;
+  std::vector<OrtOpAttr*> node_attributes;  // Manual management - CreateNode stores references, doesn't copy
+
+  // Create graph
+  OrtGraph* graph_ptr = nullptr;
+  Ort::ThrowOnError(model_editor_api.CreateGraph(&graph_ptr));
```
baijumeswani: I suggest moving all the RAII into onnxruntime_api.h to avoid this transferring ownership like we are doing here.

src/generators.h
qjia7: Done

src/models/graph_builder.cpp
qjia7: Please check the latest one to see if it's in the right way.

src/models/graph_builder.cpp
```diff
@@ -109,116 +109,105 @@ namespace GraphBuilder {
...
+    OrtTensorTypeAndShapeInfo* tensor_info = nullptr;
+    Ort::ThrowOnError(Ort::api->CreateTensorTypeAndShapeInfo(&tensor_info));
+    Ort::ThrowOnError(Ort::api->SetTensorElementType(tensor_info, output.elem_type));
+    Ort::ThrowOnError(Ort::api->SetDimensions(tensor_info, output.shape.data(), output.shape.size()));
 
-      OrtValueInfo* value_info = nullptr;
-      Ort::ThrowOnError(model_editor_api.CreateValueInfo(output.name.c_str(), type_info, &value_info));
-      Ort::api->ReleaseTypeInfo(type_info);
+    auto value_info = OrtValueInfo::Create(output.name.c_str(), tensor_info);
+    Ort::api->ReleaseTensorTypeAndShapeInfo(tensor_info);
```
baijumeswani: Could this one also be made to use RAII?

src/models/graph_builder.cpp
```diff
@@ -109,116 +109,105 @@ namespace GraphBuilder {
...
 
-      OrtTypeInfo* type_info = nullptr;
-      Ort::ThrowOnError(model_editor_api.CreateTensorTypeInfo(tensor_info, &type_info));
-      Ort::api->ReleaseTensorTypeAndShapeInfo(tensor_info);
+  // Create input ValueInfos
+  std::vector<std::unique_ptr<OrtValueInfo>> graph_inputs;
+  for (const auto& input : config.inputs) {
+    OrtTensorTypeAndShapeInfo* tensor_info = nullptr;
+    Ort::ThrowOnError(Ort::api->CreateTensorTypeAndShapeInfo(&tensor_info));
+    Ort::ThrowOnError(Ort::api->SetTensorElementType(tensor_info, input.elem_type));
```
kunal-vaishnavi: Can we move the `Ort::api->` calls inside `onnxruntime_inline.h` and call the wrappers?

src/models/graph_builder.cpp
qjia7: Done

src/models/graph_builder.cpp
qjia7: Done

src/models/graph_builder.cpp
```diff
@@ -107,118 +96,104 @@ namespace GraphBuilder {
...
+  output_ptrs.reserve(graph_outputs.size());
+  for (auto& vi : graph_outputs) {
+    output_ptrs.push_back(vi.get());
+  }
 
-      graph_outputs.push_back(value_info);
-    }
+  graph->SetInputs(input_ptrs.data(), input_ptrs.size());
+  // Release ownership since graph took it
+  for (auto& vi : graph_inputs) vi.release();
```
baijumeswani, line 139: Does graph take ownership or does graph make a copy?

src/models/graph_builder.cpp
```diff
@@ -107,118 +96,104 @@ namespace GraphBuilder {
...
-    }
-    if (graph != nullptr) {
-      Ort::api->ReleaseGraph(graph);
-    }
-    if (model != nullptr) {
-      Ort::api->ReleaseModel(model);
-    }
-    throw;
-  }
+  return model.release();  // Return ownership to caller
```
baijumeswani: Why forgo ownership? Can we return a std::unique_ptr here?

src/models/graph_builder.cpp
baijumeswani: My mistake. Looks like we must release ownership here since the graph is taking it.

src/models/graph_builder.cpp
qjia7: Nice catching! Done

baijumeswani (approved):

PR #1956, opened 2026-01-29
Add CI pipeline for WebGPU EP model testing

qjia7:


test/python/_test_utils.py
```diff
@@ -122,20 +123,75 @@ def download_model(model_name, input_path, output_path, precision, device, one_l
...
+                    webgpu_options_entry = provider
+                    break
+
+            if webgpu_options_entry is None:
+                webgpu_options_entry = {"webgpu": {"validationMode": "disabled", "enableGraphCapture": "1"}}
+                provider_options.append(webgpu_options_entry)
+            else:
+                # Update existing entry
+                webgpu_options_entry["webgpu"]["validationMode"] = "disabled"
+                webgpu_options_entry["webgpu"]["enableGraphCapture"] = "1"
```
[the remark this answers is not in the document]
qjia7: The code is case-insensitive for provider names. See https://github.com/microsoft/onnxruntime-genai/blob/main/src/config.cpp#L22. So it's ok to use lowercase.

test/python/webgpu/ort/requirements.txt
```diff
@@ -0,0 +1 @@
+onnxruntime-webgpu==1.24.0.dev20251218001
```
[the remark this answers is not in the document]
qjia7: I think onnxruntime-webgpu only has the dev version no stable version yet.

.github/workflows/win-webgpu-x64-build.yml
```diff
@@ -0,0 +1,108 @@
...
+          cmake-version: '3.31.6'
+          cmake-hash: '0f1584e8666cf4a65ec514bd02afe281caabf1d45d2c963f3151c41484f457386aa03273ab25776a670be02725354ce0b46f3a5121857416da37366342a833a0'
+          add-cmake-to-path: 'true'
+          disable-terrapin: 'false'
+
+      - uses: actions/setup-dotnet@v5
+        with:
+          dotnet-version: '8.0.x'
+
+      - name: Download OnnxRuntime Nightly (CPU package for headers and lib)
```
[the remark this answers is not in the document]
qjia7: WebGPU doesn't have the nuget package. So use the cpu package to make sure the build successfully.

.github/workflows/win-webgpu-x64-build.yml
```diff
@@ -0,0 +1,108 @@
...
+    runs-on: ["self-hosted", "1ES.Pool=onnxruntime-genai-Win2022-GPU-A10"]
+    steps:
+      - name: Checkout OnnxRuntime GenAI repo
+        uses: actions/checkout@v5
+        with:
+          submodules: true
+
+      - uses: actions/setup-python@v6
+        with:
+          python-version: '3.11.x'
```
baijumeswani: Could we use python 3.12?

.github/workflows/win-webgpu-x64-build.yml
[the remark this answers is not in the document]
baijumeswani: In the future, we could consider using the [OnnxRuntime.Foundry](https://aiinfra.visualstudio.com/PublicPackages/_artifacts/feed/ORT-Nightly/NuGet/Microsoft.ML.OnnxRuntime.Foundry/overview/1.23.2.1) package for webgpu. I can look into making this package being uploaded on a nightly cadence.

test/python/test_onnxruntime_genai.py
```diff
@@ -83,6 +83,8 @@ def main():
             output_paths += download_models(os.path.abspath(args.test_models), "int4", "cuda", log)
         if og.is_dml_available():
             output_paths += download_models(os.path.abspath(args.test_models), "int4", "dml", log)
+        if og.is_webgpu_available():
+            output_paths += download_models(os.path.abspath(args.test_models), "int4", "webgpu", log)
```
baijumeswani: webgpu will always be available. See [here](https://github.com/microsoft/onnxruntime-genai/blob/42d1bd2937368d7e86dc161c983f7aad81ebece2/src/python/python.cpp#L635)

Does that mean all the pipelines will build this model?

.github/workflows/win-webgpu-x64-build.yml
```diff
@@ -0,0 +1,108 @@
...
+          Add-Content $env:GITHUB_PATH "$env:USERPROFILE\.cargo\bin"
+
+      - name: Configure CMake
+        run: |
+          cmake --preset windows_x64_cpu_release
+
+      - name: Build with CMake
+        run: |
+          cmake --build --preset windows_x64_cpu_release --parallel
+          cmake --build --preset windows_x64_cpu_release --target PyPackageBuild
```
baijumeswani, line 81: In order to run the webgpu model in the CI pipeline with the unit tests, you will need to define the `PHI2_PATH` preprocessor macro.

https://github.com/microsoft/onnxruntime-genai/blob/42d1bd2937368d7e86dc161c983f7aad81ebece2/test/c_api_tests.cpp#L20-L31

You will also need to include webgpu in the python tests by updating the devices:
https://github.com/microsoft/onnxruntime-genai/blob/42d1bd2937368d7e86dc161c983f7aad81ebece2/test/python/test_onnxruntime_genai_api.py#L23-L35

.github/workflows/win-webgpu-x64-build.yml
```diff
@@ -0,0 +1,108 @@
...
+
+      - name: Build the C# API and Run the C# Tests
+        run: |
+          cd test\csharp
+          dotnet test /p:Configuration=release /p:NativeBuildOutputDir="$env:GITHUB_WORKSPACE\$env:binaryDir\Release" /p:OrtLibDir="$env:GITHUB_WORKSPACE\ort\lib"
+
+      - name: Run C++ Unit Tests
+        run: |-
+          copy $env:GITHUB_WORKSPACE\ort\lib\* .\$env:binaryDir\Release
+          & .\$env:binaryDir\Release\unit_tests.exe
```
baijumeswani: It might also be a good idea to run build and run the examples like @kunal-vaishnavi has done in this PR: https://github.com/microsoft/onnxruntime-genai/pull/1939

baijumeswani: If you would like to break the PR into multiple, that is also fine. This PR can go in first, and then we can update the pipeline to include more tests.

kunal-vaishnavi: Yes, let's make sure we add the examples into this CI as well.

test/python/_test_utils.py
```diff
@@ -122,20 +123,75 @@ def download_model(model_name, input_path, output_path, precision, device, one_l
     extra_options = ["--extra_options", "include_hidden_states=true"]
     if device == "cpu" and precision == "int4":
         extra_options += ["int4_accuracy_level=4"]
+
+    # WebGPU-specific options for model building (int4 optimization, etc.)
+    if device == "webgpu":
+        extra_options += [
+            "shared_embeddings=true",
+            "int4_algo_config=rtn_last",
```
kunal-vaishnavi: I thought WebGPU uses RTN or `k_quant_mixed` for the default algorithm config. Does WebGPU prefer RTN last now?

test/python/_test_utils.py
```diff
@@ -122,20 +123,75 @@ def download_model(model_name, input_path, output_path, precision, device, one_l
...
+        ]
+
     if one_layer:
         extra_options += ["num_hidden_layers=1"]
     if len(extra_options) > 1:
         command += extra_options
 
     run_subprocess(command).check_returncode()
 
+    # For WebGPU, modify genai_config.json to add session EP options
```
kunal-vaishnavi: Can we add these by default into the GenAI config when created in the model builder instead of doing it here?

test/python/_test_utils.py
```diff
@@ -122,20 +123,75 @@ def download_model(model_name, input_path, output_path, precision, device, one_l
...
     log.debug(f"Downloading models to {download_path} with precision {precision} and device {device}")
 
     ci_paths, hf_paths = get_model_paths()
     output_paths = []
 
+    # For WebGPU, only test phi-2 model
+    if device == "webgpu":
+        ci_paths = {k: v for k, v in ci_paths.items() if k == "phi-2"}
+        hf_paths = {k: v for k, v in hf_paths.items() if k == "phi-2"}
+        log.info("WebGPU device: Testing only phi-2 model")
```
kunal-vaishnavi: Why only test Phi-2? The CIs test both Phi-2 and Qwen-2.5 and both should be tested for the WebGPU EP.

test/python/test_onnxruntime_genai.py
qjia7: Add env `TEST_WEBGPU: 'true'` to make sure only webgpu pipeline build this model.

test/python/_test_utils.py
qjia7: Currently, Edge is using rtn_last. But I think all of them can work. The difference is quality and performance. In latest changes, I removed them to use the default config.

test/python/_test_utils.py
qjia7: Yes. They can be moved there. But I haven't tested all models with graph capture. So remove them for now. Will add them to builder when graph capture is mature.

.github/workflows/win-webgpu-x64-build.yml
qjia7: Will add C# related tests in a separate PR. Thanks.

cmake/presets/CMakeWinBuildPresets.json
```diff
@@ -139,6 +139,26 @@
       "configuration": "MinSizeRel",
       "configurePreset": "windows_x64_directml_minsizerel"
     },
+    {
```
baijumeswani: Do we need a special build configuration for webgpu ep? Could we parameterize the test on available models in the test data folder?

.github/workflows/win-webgpu-x64-build.yml
```diff
@@ -0,0 +1,145 @@
...
+          & $exePath -y --default-toolchain=1.86.0
+          Add-Content $env:GITHUB_PATH "$env:USERPROFILE\.cargo\bin"
+
+      - name: Configure CMake
+        run: |
+          cmake --preset windows_x64_webgpu_release -DTEST_PHI2=True
+
+      - name: Install Python dependencies to get WebGPU OnnxRuntime
+        run: |
+          python -m pip install -r test\python\requirements.txt
```
baijumeswani: This seems a lot of effort to get the correct dlls. Perhaps consider using this package: https://aiinfra.visualstudio.com/PublicPackages/_artifacts/feed/ORT-Nightly/NuGet/Microsoft.ML.OnnxRuntime.Foundry/overview/1.24.1.1

test/c_api_tests.cpp
```diff
@@ -25,13 +25,15 @@
...
+#elif USE_WEBGPU
+#define PHI2_PATH MODEL_PATH "phi-2/int4/webgpu"
 #else
 #define PHI2_PATH MODEL_PATH "phi-2/int4/cpu"
 #endif
 #endif
 
 #ifndef ENABLE_ENGINE_TESTS
-#define ENABLE_ENGINE_TESTS TEST_PHI2 && !USE_DML
+#define ENABLE_ENGINE_TESTS TEST_PHI2 && !USE_DML && !USE_WEBGPU
```
baijumeswani: Why disable engine tests?

test/model_tests.cpp
```diff
@@ -24,6 +24,8 @@ extern std::string g_custom_model_path;
 #define PHI2_PATH MODEL_PATH "phi-2/int4/cuda"
 #elif USE_DML
 #define PHI2_PATH MODEL_PATH "phi-2/int4/dml"
+#elif USE_WEBGPU
+#define PHI2_PATH MODEL_PATH "phi-2/int4/webgpu"
```
baijumeswani: Could we perhaps define all the paths. And then parametrize on them and run the tests only if the model is found at that path?

test/python/test_onnxruntime_genai.py
```diff
@@ -81,6 +80,9 @@ def main():
         output_paths += download_models(os.path.abspath(args.test_models), "int4", "cuda", log)
     if og.is_dml_available():
         output_paths += download_models(os.path.abspath(args.test_models), "int4", "dml", log)
+    # Only build WebGPU models if TEST_WEBGPU environment variable is set
+    if og.is_webgpu_available() and os.environ.get("TEST_WEBGPU", "").lower() in ["true", "1", "yes"]:
```
baijumeswani: Could we pass in a list of eps at runtime to this script as an argument?  We can leave the defaults to `og.is_xyzep_avaiable()`. But if ep is explicitly passed in, we should build only those. This way we can build the test models locally as well.

test/test_utils.h
```diff
@@ -0,0 +1,54 @@
...
+      MODEL_PATH "phi-2/int4/webgpu",
+      MODEL_PATH "phi-2/int4/cpu"};
+
+  for (const auto& path : candidate_paths) {
+    std::ifstream test_file(path + std::string("/genai_config.json"));
+    if (test_file.good()) {
+      phi2_path = path;
+      return phi2_path.c_str();
+    }
+  }
```
baijumeswani, line 35: While this is ok for now, we should clean this up so we test all available models and not the first one we find from this array.

We can do this in another pr though.

test/test_utils.h
```diff
@@ -0,0 +1,54 @@
...
+      MODEL_PATH "phi-2/int4/dml",
+      MODEL_PATH "phi-2/int4/webgpu",
+      MODEL_PATH "phi-2/int4/cpu"};
+
+  for (const auto& path : candidate_paths) {
+    std::ifstream test_file(path + std::string("/genai_config.json"));
+    if (test_file.good()) {
+      phi2_path = path;
+      return phi2_path.c_str();
+    }
```
baijumeswani:
```suggestion
    std::filesystem::path model_path(path);
    if (std::filesystem::exists(path / "genai_config.json") {
      phi2_path = model_path;
      return phi2_path;
    }
```

probably safer to use fs path instead of const char*?

test/python/test_onnxruntime_genai.py
```diff
@@ -67,6 +66,12 @@ def parse_arguments():
         help="Whether to run e2e tests. If not specified e2e tests will not run.",
         action="store_true",
     )
+    parser.add_argument(
+        "--eps",
```
kunal-vaishnavi, line 70: Should we add `default = []`?

qjia7: Done

PR #1957, opened 2026-01-29
update WebGPU buffer memory info name

fs-eire:
This is the corresponding change to https://github.com/microsoft/onnxruntime/pull/27207.

Need to work with an updated onnxruntime DLL.


fs-eire:
For 1.24.3, it is fine. See this part: https://github.com/microsoft/onnxruntime/blob/rel-1.24.3/onnxruntime/core/framework/allocator.cc#L239-L252

The only thing we need to take care about is 1.24.0-1.24.2

PR #1967, opened 2026-02-05
Sync packaging changes with ONNX Runtime

baijumeswani:
Changes in this PR:

- Use gcc14 for linux packaging.
- Build Python 3.11 ~ Python 3.14 packages.
- Remove Python 3.10 package.
- Further remove references to macOS x64 packages.
- Remove onnxruntime-genai-cuda.lib from custom nuget packages.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -236,7 +236,7 @@ endif()
 
 if(CMAKE_SYSTEM_NAME STREQUAL "Linux")
   set_target_properties(onnxruntime-genai PROPERTIES POSITION_INDEPENDENT_CODE ON)
-  target_link_libraries(onnxruntime-genai PRIVATE dl)  # For dlopen & co
+  target_link_libraries(onnxruntime-genai PUBLIC dl)  # For dlopen & co
 endif()
 
 if(USE_DML)


.pipelines/stages/jobs/custom-nuget-packaging-job.yml
```diff
@@ -97,7 +89,7 @@ jobs:
 
   - task: UsePythonVersion@0
     inputs:
-      versionSpec: 3.10
+      versionSpec: 3.11
```
kunal-vaishnavi, line 92: Should we use 3.11 or 3.12 as the default to match the above YAML changes?

.pipelines/stages/jobs/nuget-packaging-job.yml
```diff
@@ -311,7 +304,7 @@ jobs:
 
   - task: UsePythonVersion@0
     inputs:
-      versionSpec: 3.10
+      versionSpec: 3.11
```
kunal-vaishnavi, line 307: Same question

.pipelines/stages/jobs/winml-nuget-packaging-job.yml
```diff
@@ -120,7 +120,7 @@ jobs:
 
       - task: UsePythonVersion@0
         inputs:
-          versionSpec: 3.10
+          versionSpec: 3.11
```
kunal-vaishnavi, line 123: Same question

cmake/cuda_configuration.cmake
```diff
@@ -0,0 +1,188 @@
...
+        set(CMAKE_CUDA_COMPILER_VERSION "${version}")
+      else()
+        message(FATAL_ERROR "Failed to determine CUDA version")
+      endif()
+    endif()
+  else()
+    message(FATAL_ERROR "No CUDA compiler found")
+  endif()
+
+  set(CUDA_REQUIRED_VERSION "11.4")
```
kunal-vaishnavi, line 57: Shouldn't this be 12.8 now that we've increased the version we support?

.pipelines/stages/jobs/custom-nuget-packaging-job.yml
baijumeswani: Moving to python 3.12 may require some additional changes to the agent. It should be fine for now as long as we don't use python 3.10.

cmake/cuda_configuration.cmake
baijumeswani: I copied this file directly from ort: https://github.com/microsoft/onnxruntime/blob/main/cmake/external/cuda_configuration.cmake

Should be fine to leave it as is.

kunal-vaishnavi (approved):

PR #1971, opened 2026-02-09
Fix uninitialized tools variable and improve exception debug messages

sheller-ms:
When calling `apply_chat_template `on the `system_prompt`, it was ending up in the Exception case. After printing the exception message, I found it was due to `tools `not being initialized. This PR fixes that issue and has verbose mode print the exception message.

### DIFF
--- a/examples/python/model-mm.py
+++ b/examples/python/model-mm.py
@@ -122,7 +122,7 @@ def main(args):
             prompt = apply_chat_template(model_path=args.model_path, tokenizer=tokenizer, messages=messages, tools=tools, add_generation_prompt=True)
         except Exception as e:
             if args.verbose:
-                print(f"Exception in apply_chat_template for system_prompt: {e}")
+                print(f"Exception in apply_chat_template: {e}")
             prompt = text
         if args.verbose:
             print(f"Prompt: {prompt}")
--- a/examples/python/model-qa.py
+++ b/examples/python/model-qa.py
@@ -105,7 +105,7 @@ def main(args):
             prompt = apply_chat_template(model_path=args.model_path, tokenizer=tokenizer, messages=messages, tools=tools, add_generation_prompt=True)
         except Exception as e:
             if args.verbose:
-                print(f"Exception in apply_chat_template for system_prompt: {e}")
+                print(f"Exception in apply_chat_template: {e}")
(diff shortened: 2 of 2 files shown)


examples/python/model-chat.py
```diff
@@ -40,6 +40,7 @@ def main(args):
     message = [{"role": "system", "content": args.system_prompt}]
 
     # Get and set guidance info if requested
+    tools = "" # optional; initialize to empty string to avoid uninitialized variable error later
```
kunal-vaishnavi: Can you change the variable initializations to the following?

```py
guidance_type, guidance_data, tools = "", "", "" 
```

We want to keep the code consistent across all of the examples.

https://github.com/microsoft/onnxruntime-genai/blob/636b92ba754c55e4b5e8c14201f3b3180f370de1/examples/python/model-qa.py#L50-L62

https://github.com/microsoft/onnxruntime-genai/blob/636b92ba754c55e4b5e8c14201f3b3180f370de1/examples/python/model-mm.py#L58-L70

examples/python/model-chat.py
```diff
@@ -67,7 +68,9 @@ def main(args):
     # Apply chat template
     try:
         system_prompt = apply_chat_template(model_path=args.model_path, tokenizer=tokenizer, messages=json.dumps(message), tools=tools, add_generation_prompt=False)
-    except:
+    except Exception as e:
+        if args.verbose:
+            print("Exception in apply_chat_template for system_prompt.", e)
```
kunal-vaishnavi:
```suggestion
            print(f"Exception in apply_chat_template for system_prompt: {e}")
```

examples/python/model-chat.py
```diff
@@ -100,7 +103,9 @@ def main(args):
         # Apply chat template
         try:
             user_prompt = apply_chat_template(model_path=args.model_path, tokenizer=tokenizer, messages=json.dumps(message), add_generation_prompt=True)
-        except:
+        except Exception as e:
+            if args.verbose:
+                print("Exception in apply_chat_template for user_prompt.", e)
```
kunal-vaishnavi:
```suggestion
                print(f"Exception in apply_chat_template for user_prompt: {e}")
```

PR #1976, opened 2026-02-12
Update cmake cuda architecture and use win-arm64 pool workaround

baijumeswani:
Updates the cuda architectures being built by default in ort-genai so we can support newer sms (like rtx 5090)

### DIFF
--- a/cmake/presets/CMakeLinuxDefaultConfigPresets.json
+++ b/cmake/presets/CMakeLinuxDefaultConfigPresets.json
@@ -28,7 +28,7 @@
       "inherits": "linux_gcc_default",
       "cacheVariables": {
         "USE_CUDA": "ON",
-        "CMAKE_CUDA_ARCHITECTURES": "60-real;70-real;75-real;80-real;86-real;89-real;90-real;100-real;120-real"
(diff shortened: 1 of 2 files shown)


cmake/presets/CMakeLinuxDefaultConfigPresets.json
```diff
@@ -28,7 +28,7 @@
       "inherits": "linux_gcc_default",
       "cacheVariables": {
         "USE_CUDA": "ON",
-        "CMAKE_CUDA_ARCHITECTURES": "60;61;70;75;80;86"
+        "CMAKE_CUDA_ARCHITECTURES": "60-real;70-real;75-real;80-real;86-real;89-real;90-real;100-real;120-real"
```
kunal-vaishnavi: Why `-real` only? Without the suffix, it should [compile](https://cmake.org/cmake/help/latest/prop_tgt/CUDA_ARCHITECTURES.html) for `-real` and `-virtual`.

baijumeswani: Building both means the compiled binary contains both the machine binary as well as some assembly-like code that can be built at runtime to support newer architectures. It would increase our binary size.

PR #1979, opened 2026-02-14
Add support for LFM2 in ORT GenAI

xenova:
This PR adds support for the LFM2 series of LLMs from LiquidAI: https://huggingface.co/models?other=lfm2&sort=trending&search=LiquidAI%2F

For sample inference code, you can refer to the model card I uploaded for the converted models: https://huggingface.co/onnx-community/LFM2-350M-ONNX#onnxruntime

I know there are still a couple of things missing from the PR (this PR only adds model builder support)... so hopefully @kunal-vaishnavi can finalize the last few things on the checklist 😇 (I've enabled edits by maintainers).

cc @ykhrustalev too 🥳


src/python/py/models/builders/lfm2.py
```diff
@@ -0,0 +1,267 @@
...
+        self.rope_attrs["theta"] = config.rope_parameters["rope_theta"]
+
+        # Calculate the dynamic intermediate_size for the MLP.
+        intermediate_size = self.intermediate_size
+        if getattr(config, "block_auto_adjust_ff_dim", False):
+            intermediate_size = int(2 * intermediate_size / 3)
+            if getattr(config, "block_ffn_dim_multiplier", None) is not None:
+                intermediate_size = int(config.block_ffn_dim_multiplier * intermediate_size)
+            multiple_of = getattr(config, "block_multiple_of", 1)
+            intermediate_size = multiple_of * ((intermediate_size + multiple_of - 1) // multiple_of)
```
ykhrustalev: the transformers v5 has a little different logic
https://github.com/huggingface/transformers/blob/main/src/transformers/models/lfm2/modeling_lfm2.py#L138-L146

```suggestion
                multiple_of = getattr(config, "block_multiple_of", 1)
                intermediate_size = multiple_of * ((intermediate_size + multiple_of - 1) // multiple_of)
```

if a future config omits either field or sets block_ffn_dim_multiplier to None, the results would diverge

ykhrustalev (approved):
lgtm

src/python/py/models/builders/lfm2.py
```diff
@@ -0,0 +1,267 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation.  All rights reserved.
+# Licensed under the MIT License.  See License.txt in the project root for
+# license information.
+# --------------------------------------------------------------------------
+import numpy as np
```
kunal-vaishnavi: Can we use torch instead of numpy for `iinfo` to [stay consistent](https://github.com/search?q=repo%3Axenova%2Fonnxruntime-genai%20path%3A%2F%5Esrc%5C%2Fpython%5C%2Fpy%5C%2Fmodels%5C%2Fbuilders%5C%2F%2F%20iinfo&type=code) with other files?

src/python/py/models/builders/lfm2.py
```diff
@@ -0,0 +1,267 @@
...
+import onnx_ir as ir
+
+from .base import Model
+
+
+class Lfm2Model(Model):
```
kunal-vaishnavi:
```suggestion
class LFM2Model(Model):
```

src/python/py/models/builders/lfm2.py
kunal-vaishnavi: Let's add this even though the CI is currently pinned to transformers before v5. There were a few issues discovered with v5 in our CIs, but there should be more stability in transformers v5 now. We can upgrade the CI to v5 and add this change.

src/python/py/models/builders/lfm2.py
```diff
@@ -0,0 +1,267 @@
...
+                # Add KV cache outputs
+                key_out_name = f"present.{i}.key"
+                key_out_shape = self.make_key_value_cache_shape(i, self.output_shapes["present.key"])
+                outputs.append(self.make_value(key_out_name, dtype=self.output_types["present.key"], shape=key_out_shape))
+
+                value_out_name = f"present.{i}.value"
+                value_out_shape = self.make_key_value_cache_shape(i, self.output_shapes["present.value"])
+                outputs.append(self.make_value(value_out_name, dtype=self.output_types["present.value"], shape=value_out_shape))
+
+            elif self.layer_types[i] == "conv":
```
kunal-vaishnavi: From what I read in the sample inference code, these past/present convolution caches work like KV caches. You provide some past convolution caches and produce updated convolution caches. This would require changes within ORT GenAI to construct a new object for managing the convolution caches during runtime.

src/python/py/models/builders/lfm2.py
```diff
@@ -0,0 +1,267 @@
...
+                conv_cache_shape = ["batch_size", self.hidden_size, self.conv_L_cache]
+
+                # Add conv cache input
+                past_conv_name = f"past_conv.{i}"
```
kunal-vaishnavi: Is this the standard naming convention for past/present convolution caches? Could we use a format such as `past.{i}.conv` and `present.{i}.conv` to more closely match the KV cache naming format?

src/python/py/models/builders/lfm2.py
```diff
@@ -0,0 +1,267 @@
...
+        for out_val in [b_out, c_out, x_out]:
+            self.make_value(out_val, self.io_dtype, shape=["batch_size", self.hidden_size, "sequence_length"])
+
+        # Element-wise multiply: bx = b * x
+        mul_1_name = f"{basename}/Mul_1"
+        self.make_mul(mul_1_name, [b_out, x_out], self.io_dtype, shape=["batch_size", self.hidden_size, "sequence_length"])
+
+        # 2. Stateful convolution: concatenate with past conv cache
+        past_conv_name = f"past_conv.{layer_id}"
+        conv_input_name = f"{basename}/Conv_Input"
```
kunal-vaishnavi: Typically, the [node naming convention](https://github.com/microsoft/onnxruntime-genai/blob/main/src/python/py/models/DESIGN.md#namespaces) is to describe the path to the node in the graph and the final term after the `/` is the op type.

Image

Can we rename this node name to end with `/Concat`?

src/python/py/models/builders/lfm2.py
```diff
@@ -0,0 +1,267 @@
...
+        self.make_concat(
+            conv_input_name, [past_conv_name, f"{mul_1_name}/output_0"], self.io_dtype,
+            shape=["batch_size", self.hidden_size, "past_plus_current_seq_len"], axis=2,
```
kunal-vaishnavi:
```suggestion
            shape=["batch_size", self.hidden_size, "total_sequence_length"], axis=2,
```

src/python/py/models/builders/lfm2.py
```diff
@@ -0,0 +1,267 @@
...
+        conv_weight_name = f"model.layers.{layer_id}.conv.conv.weight"
+        self.make_initializer(conv_module.conv.weight, conv_weight_name, to=self.io_dtype)
+        conv_inputs = [conv_input, conv_weight_name]
+        if conv_module.conv.bias is not None:
+            conv_bias_name = f"model.layers.{layer_id}.conv.conv.bias"
+            self.make_initializer(conv_module.conv.bias, conv_bias_name, to=self.io_dtype)
+            conv_inputs.append(conv_bias_name)
+
+        conv_out_full = f"{conv_op_name}/output_0"
+        self.make_node(
```
kunal-vaishnavi: For ops that don't have a `make_op` method, we have to call the `make_value` method as well to add shape inference for the node's output.

src/python/py/models/builders/lfm2.py
```diff
@@ -0,0 +1,267 @@
...
+            "Conv", inputs=conv_inputs, outputs=[conv_out_full], name=conv_op_name,
+            kernel_shape=[self.conv_L_cache], pads=[0, 0], group=self.hidden_size,
+        )
+
+        # Slice the conv output to keep only the last seq_len elements (obtained from root_input shape)
+        shape_name = f"{basename}/Shape"
+        self.make_shape(shape_name, root_input, shape=[3])
+        seq_len_gather_name = f"{basename}/Gather_SeqLen"
```
kunal-vaishnavi: Same comment about node names from here and below as well

src/python/py/models/builders/lfm2.py
```diff
@@ -0,0 +1,267 @@
...
+
+        unsqueeze_starts_name = f"{basename}/Unsqueeze_starts"
+        self.make_unsqueeze(
+            unsqueeze_starts_name, [f"{neg_seq_len_name}/output_0", "/model/constants/INT64/[0]"],
+            ir.DataType.INT64, shape=[1],
+        )
+
+        slice_name = f"{basename}/Slice_Conv_Output"
+        conv_out = f"{slice_name}/output_0"
+        self.make_node(
```
kunal-vaishnavi: Can we use `self.make_slice` instead to simplify?

src/python/py/models/builders/lfm2.py
```diff
@@ -0,0 +1,267 @@
...
+        self.make_value(conv_out, self.io_dtype, shape=["batch_size", self.hidden_size, "sequence_length"])
+
+        # Element-wise multiply: result = c * conv_out
+        mul_2_name = f"{basename}/Mul_2"
+        self.make_mul(mul_2_name, [c_out, conv_out], self.io_dtype, shape=["batch_size", self.hidden_size, "sequence_length"])
+
+        # 3. Cache update: slice the conv input to keep last conv_L_cache elements
+        present_conv_name = f"present_conv.{layer_id}"
+        slice_cache_name = f"{basename}/Slice_Cache"
+        self.make_node(
```
kunal-vaishnavi, line 194: Can we use `self.make_slice` instead to simplify?

src/python/py/models/builders/lfm2.py
xenova: Yep exactly 👍 They can be initialized to all zeros too.

src/python/py/models/builders/lfm2.py
xenova: Yeah, the original choice of name was due to the mismatch between `past_key_values.{i}.key` and `present.{i}.key` (meaning we would name this one `past_conv.{i}.conv` and `present_conv.{i}.conv`... and then leading to `past_conv.{i}` and `present_conv.{i}`).

unfortunately there are many lfm2-based ONNX models already in use which have this convention 😅

xenova: similar naming issue with [mamba](https://netron.app/?url=https://huggingface.co/onnx-community/Falcon-H1-Tiny-90M-Instruct-ONNX/resolve/main/onnx/model.onnx)-based models, which introduce a ssm cache mechanism.
image

If we had a do-over, probably this would work best:
```
past.{i}.key
past.{i}.value
past.{i}.conv
past.{i}.ssm

present.{i}.key
present.{i}.value
present.{i}.conv
present.{i}.ssm
```

kunal-vaishnavi: We can still name new models that are produced by the model builder in this format. The GenAI config will contain a mapping to the input category and the input format.

For example, the current models on Hugging Face can use

```json
"past_conv_names": "past_conv.%d"
```

while newly produced models can use the following

```json
"past_conv_names": "past.%d.conv"
```

In this way, both name formats are supported.

src/models/kv_cache.cpp
```diff
@@ -560,6 +762,22 @@ bool IsCacheNeeded(const Model& model) {
   return false;
 }
 
+bool IsLFM2Model(const Model& model) {
```
kunal-vaishnavi: Can we define a method inside `model_type.h` instead to identify that a model is an LFM2 model and use the model type from the `config_` object?

src/python/py/models/builders/lfm2.py
```diff
@@ -0,0 +1,295 @@
...
+        )
+        for out_val in [b_out, c_out, x_out]:
+            self.make_value(out_val, self.io_dtype, shape=["batch_size", self.hidden_size, "sequence_length"])
+
+        # Element-wise multiply: bx = b * x
+        mul_1_name = f"{basename}/Mul_1"
+        self.make_mul(mul_1_name, [b_out, x_out], self.io_dtype, shape=["batch_size", self.hidden_size, "sequence_length"])
+
+        # 2. Stateful convolution: concatenate with past conv cache
+        past_conv_name = f"past_conv.{layer_id}"
```
kunal-vaishnavi: Let's reference the input/output names via `self.input_names`/`self.output_names`. That way, you only have to know the shorthand syntax and not the full canonical name.

src/python/py/models/builders/lfm2.py
```diff
@@ -0,0 +1,295 @@
...
+        conv_weight_name = f"model.layers.{layer_id}.conv.conv.weight"
+        self.make_initializer(conv_module.conv.weight, conv_weight_name, to=self.io_dtype)
+        conv_inputs = [conv_input, conv_weight_name]
+        if conv_module.conv.bias is not None:
+            conv_bias_name = f"model.layers.{layer_id}.conv.conv.bias"
+            self.make_initializer(conv_module.conv.bias, conv_bias_name, to=self.io_dtype)
+            conv_inputs.append(conv_bias_name)
+
+        conv_out_full = f"{conv_op_name}/output_0"
+        self.make_node(
```
kunal-vaishnavi: Can we use `make_conv` here?

src/python/py/models/builders/lfm2.py
```diff
@@ -0,0 +1,295 @@
...
+        # Hybrid models don't support shared past/present buffer
+        genai_config["search"]["past_present_share_buffer"] = False
+
+        with open(config_path, "w") as f:
+            json.dump(genai_config, f, indent=4)
+
+    def save_processing(self, model_name_or_path, extra_kwargs, out_dir):
+        super().save_processing(model_name_or_path, extra_kwargs, out_dir)
+
+        # Fix tokenizer_class if needed (TokenizersBackend is not recognized by ort-extensions)
```
kunal-vaishnavi: Has this been fixed with the [recent support](https://github.com/microsoft/onnxruntime-genai/pull/2089) of transformers v5 in ORT GenAI?

src/models/kv_cache.cpp
```diff
@@ -572,6 +786,12 @@ std::unique_ptr<KeyValueCache> CreateKeyValueCache(State& state) {
     return std::make_unique<ModelManagedKeyValueCache>(state);
   }
 
+  // LFM2 models interleave attention and conv layers, requiring a cache that handles
+  // both KV cache for attention layers and fixed-size conv state for conv layers.
+  if (ModelType::IsLFM2(state.model_.config_->model.type) && IsLFM2CacheNeeded(state.model_)) {
```
kunal-vaishnavi: If the first predicate is true (i.e. the model type is LFM2), why is the second predicate still needed? Won't the caches already exist or are there LFM2 models that don't use caches?

src/python/py/models/builders/lfm2.py
```diff
@@ -0,0 +1,282 @@
...
+        )
+        for out_val in [b_out, c_out, x_out]:
+            self.make_value(out_val, self.io_dtype, shape=["batch_size", self.hidden_size, "sequence_length"])
+
+        # Element-wise multiply: bx = b * x
+        mul_1_name = f"{basename}/Mul_1"
+        self.make_mul(mul_1_name, [b_out, x_out], self.io_dtype, shape=["batch_size", self.hidden_size, "sequence_length"])
+
+        # 2. Stateful convolution: concatenate with past conv cache
+        past_conv_name = f"past_conv.{layer_id}"
```
kunal-vaishnavi: Let's reference the input/output names via `self.input_names`/`self.output_names`. That way, you only have to know the shorthand syntax and not the full canonical name.

```suggestion
        past_conv_name = self.input_names[f"past_conv.{layer_id}"]
```

src/models/kv_cache.cpp
ykhrustalev: I have no explanation for that, removed it

kunal-vaishnavi (approved):

natke (approved):

kunal-vaishnavi (approved):

xenova:
Thanks @ykhrustalev, @kunal-vaishnavi, and @natke for getting this PR over the line! ❤️

PR #1986, opened 2026-02-23
Fix incorrect batch responses when using multiple prompts

lnigam:
When running with batch_size > 1, only the last prompt's response was correct; other responses decoded as "!!!!". The generator was created with batch_size=1 because search options were taken only from CLI args (which have no --batch_size), so get_search_options() defaulted it to 1 while the model config correctly had batch_size=len(prompts).

Fix: In model-generate.py, set search_options["batch_size"] = len(prompts) before calling params.set_search_options() so the generator's batch size matches the number of prompts.

### DIFF
--- a/examples/python/model-generate.py
+++ b/examples/python/model-generate.py
@@ -30,7 +30,7 @@ def main(args):
             text = input("Input: ")
             prompts = [text]
     setattr(args, "batch_size", len(prompts))
-    search_config = {"batch_size": len(prompts), "chunk_size": args.chunk_size, "num_beams": args.num_beams}
+    search_config = {"batch_size": args.batch_size, "chunk_size": args.chunk_size, "num_beams": args.num_beams}
     config = get_config(args.model_path, args.execution_provider, ep_options={}, search_options=search_config)
 
     model = og.Model(config)


examples/python/model-generate.py
```diff
@@ -55,6 +55,8 @@ def main(args):
 
     params = og.GeneratorParams(model)
     search_options = get_search_options(args)
+    # Generator batch_size must match number of prompts (config overlay already has batch_size for model build)
+    search_options["batch_size"] = len(prompts)
```
kunal-vaishnavi: Can we add batch size to `args` before the following lines via `setattr(args, "batch_size", len(prompts))` instead?

https://github.com/microsoft/onnxruntime-genai/blob/4a5aee1138d9665ef684225c28dd625cc7d810e5/examples/python/model-generate.py#L33-L34

Then, the batch size should get correctly returned from `get_search_options`.

lnigam: done

examples/python/model-generate.py
```diff
@@ -29,7 +29,7 @@ def main(args):
         else:
             text = input("Input: ")
             prompts = [text]
-
+    setattr(args, "batch_size", len(prompts))
     search_config = {"batch_size": len(prompts), "chunk_size": args.chunk_size, "num_beams": args.num_beams}
```
```suggestion
    search_config = {"batch_size": args.batch_size, "chunk_size": args.chunk_size, "num_beams": args.num_beams}
```

PR #1992, opened 2026-02-25
webgpu: Optimize Copyfrom

qjia7:
This pull request refactors the WebGPU device interface implementation to optimize memory info handling and improve buffer copy operations. The main change is caching the WebGPU allocator's memory info to avoid repeated calls, which simplifies tensor creation and improves performance. Additionally, the buffer copy logic now includes a fast path for CPU-to-WebGPU transfers with zero offsets.

These changes reduce overhead and make the codebase more efficient and maintainable.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -235,6 +235,31 @@ if(USE_GUIDANCE)
     # bcrypt is needed for the rust std lib
     target_link_libraries(onnxruntime-genai PRIVATE bcrypt)
   endif()
+  if(MSVC)
+    # The Rust llguidance static library is always compiled against the release MSVC CRT
+    # (Rust has no debug CRT concept). The .lib embeds /DEFAULTLIB directives for the release
+    # CRT and /NODEFAULTLIB directives that suppress the debug CRT (msvcrtd, ucrtd, vcruntimed).
+    # In Debug builds, C++ code (e.g. onnxruntime-extensions) is compiled with /MDd and references
+    # debug-only CRT functions like _CrtDbgReport (in ucrtd.lib). Because the Rust .lib suppresses
+    # ucrtd.lib via its embedded /NODEFAULTLIB, _CrtDbgReport becomes unresolved.
+    #
+    # Fix: explicitly add the debug CRT import libraries in Debug builds. Explicitly specified
+    # libraries are not affected by /NODEFAULTLIB directives (those only suppress /DEFAULTLIB
+    # auto-linking). Also suppress the conflicting release CRT to avoid LNK4098 warnings.
+    target_link_libraries(onnxruntime-genai PRIVATE
+      $<$<CONFIG:Debug>:msvcrtd.lib>
+      $<$<CONFIG:Debug>:ucrtd.lib>
+      $<$<CONFIG:Debug>:vcruntimed.lib>
+    )
+    target_link_options(onnxruntime-genai PRIVATE
+      $<$<CONFIG:Debug>:/NODEFAULTLIB:msvcrt.lib>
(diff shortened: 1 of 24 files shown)


src/webgpu/interface.cpp
```diff
@@ -98,24 +91,33 @@ struct WebGPUMemory final : DeviceBuffer {
...
+      // Full buffer copy using CopyTensors (no offsets)
+      int64_t shape_val = static_cast<int64_t>(size_in_bytes);
+      std::span<const int64_t> shape{&shape_val, 1};
+      auto cpu_mem_info = OrtMemoryInfo::CreateCpu(OrtDeviceAllocator, OrtMemTypeDefault);
+      auto src_tensor = OrtValue::CreateTensor(*cpu_mem_info, source.p_device_, size_in_bytes, shape, ONNX_TENSOR_ELEMENT_DATA_TYPE_UINT8);
+      auto dst_tensor = OrtValue::CreateTensor(*ort_memory_info_, p_device_, size_in_bytes, shape, ONNX_TENSOR_ELEMENT_DATA_TYPE_UINT8);
+
+      // Use ORT C API's CopyTensors for CPU-to-GPU copy
+      OrtValue* src_ptrs[] = {src_tensor.get()};
+      OrtValue* dst_ptrs[] = {dst_tensor.get()};
```
baijumeswani:
```
const std::vector<const OrtValue*> src_ptrs = {src_tensor.get()};
const std::vector<const OrtValue*> dst_ptrs = {dst_tensor.get()};
```

src/webgpu/interface.cpp
```diff
@@ -98,24 +91,33 @@ struct WebGPUMemory final : DeviceBuffer {
...
+      Ort::ThrowOnError(Ort::api->CopyTensors(&GetOrtEnv(), src_ptrs, dst_ptrs, nullptr, 1));
```
baijumeswani: Please use onnxruntime_ai.h definitions where possible

```
GetOrtEnv().CopyTensors(src_ptrs, dst_ptrs, nullptr);
```

src/webgpu/interface.cpp
```diff
@@ -10,6 +10,7 @@ namespace Generators {
 namespace WebGPU {
 
 static Ort::Allocator* ort_allocator_{};
+static const OrtMemoryInfo* ort_memory_info_{};
```
baijumeswani: Why does this need to be static and not a private member variable?

src/webgpu/interface.cpp
```diff
@@ -159,6 +157,8 @@ struct InterfaceImpl : DeviceInterface {
   void InitOrt(const OrtApi& /*api*/, Ort::Allocator& allocator) override {
     assert(!ort_allocator_);
     ort_allocator_ = &allocator;
+    // Cache the memory info to avoid repeated AllocatorGetInfo calls
+    Ort::ThrowOnError(Ort::api->AllocatorGetInfo(ort_allocator_, &ort_memory_info_));
```
baijumeswani:
```
*ort_memory_info_ = ort_allocator_->GetInfo();
```

src/webgpu/interface.cpp
qjia7: Both WebGPUMemory and InterfaceImpl need to use it. We can't use it in one if ort_memory_info_ is another one's private member.

baijumeswani: WebGPUMemory is only ever created from the InterfaceImpl from what I understand. If that is the case, InterfaceImpl can always pass in the pointer to WebGPUMemory as a constructor argument.

PR #1993, opened 2026-02-27
GenAI changes to support EPContext compilation and validation

lnigam:
-Add Support for EPContext compilation by default, If enabled in the config
-Once EPContext is compiled, GenAI loads the context model instead of original model and avoids model recompilation and graph optimization.
-Also added support for API which can validate the existing EPContext. It can also check if the existing EPContext may not run optimal and requires recompilation.


examples/c/src/common.cpp
```diff
@@ -236,18 +236,17 @@ void RegisterEP(const std::string& ep, const std::string& ep_path) {
     return;  // No library path specified, skip registration
   }
 
-  std::cout << "Registering execution provider: " << ep_path << std::endl;
-  auto env = Ort::Env();
+  // Must register on GenAI's OrtEnv (via OgaRegisterExecutionProviderLibrary) so
+  // GetEpDevices() in ValidateCompiledModel sees the plugin; Ort::Env() is a different env.
```
kunal-vaishnavi, line 240: This seems to be an ongoing issue as there are some cases where registering on the `Ort::Env()` is [preferred](https://github.com/microsoft/onnxruntime-genai/pull/1939#discussion_r2756397130) and other cases where registering on GenAI's `OrtEnv` is preferred. We should find a way to consolidate.

examples/c/src/model_compile.cpp
```diff
@@ -0,0 +1,248 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+//
+// Model Compile example: runs the same model under different EP and compile configurations
+// (CPU, CPU+overlay, NvTensorRtRtx no-compile / 4 options / all options). Use -v for verbose,
```
kunal-vaishnavi, line 5: Besides appending some GenAI config options and using the overlay API, it seems to be the same logic as the existing examples.

Instead of creating a new standalone example that needs to be continually maintained, let's integrate the necessary logic into the `common` files. We should update the `model-qa` and/or `model-chat` examples in C/C++, C#, and Python to show this capability. In this way, these changes will be continually tested since the `model-qa` examples are built and tested in the CIs now.

We want to keep the examples consistent across language bindings and reduce maintenance by sharing and re-using logic as much as possible between examples.

src/models/gpt.cpp
```diff
@@ -5,7 +5,8 @@ namespace Generators {
 
 Gpt_Model::Gpt_Model(std::unique_ptr<Config> config, OrtEnv& ort_env)
     : Model{std::move(config)} {
-  session_decoder_ = CreateSession(ort_env, config_->model.decoder.filename, session_options_.get());
+  std::string decoder_model_path = CompileModel(ort_env, config_->model.decoder.filename, session_options_.get(), true, config_->model.decoder.compile_options);
```
kunal-vaishnavi, line 8: Can we simplify this by passing the `config_->model.decoder` object directly? Each object should have a filename and compile options to access. Then, the lines won't be as long and there will be one less parameter.

src/models/model.cpp
```diff
@@ -1213,7 +1214,272 @@ OrtSessionOptions* Model::GetSessionOptions(const std::string& model_id) const {
...
+  return std::unique_ptr<OrtModelCompilationOptions>(p);
+}
+
+bool Model::ValidateCompiledModel(OrtEnv& ort_env, const fs::path& compiled_model_path, bool force_compile_if_needed) {
+  const std::string ep_name = EPContextSupportedProviders(p_device_->GetType());
+  // EP context applicability already checked in CompileModel; ep_name is non-empty here.
+
+  Ort::Allocator& alloc = Ort::Allocator::GetWithDefaultOptions();
+  char* compat_info = nullptr;
+  OrtStatus* st = Ort::api->GetCompatibilityInfoFromModel(
```
kunal-vaishnavi, line 1242: Let's prefer using the [ORT GenAI API wrappers](https://github.com/microsoft/onnxruntime-genai/blob/main/src/models/onnxruntime_inline.h) over calling `Ort::api->MethodName` directly.

src/models/model.cpp
```diff
@@ -1213,7 +1214,272 @@ OrtSessionOptions* Model::GetSessionOptions(const std::string& model_id) const {
...
+    Log("warning", "Found existing EP Context for " + ep_name + " but its performance is sub-optimal in this EP, recommended to recompile EPContext.");
+    return true;
+  }
+
+  return false;  // NOT_APPLICABLE or unknown
+}
+
+bool Model::CheckCompiledModelExists(OrtEnv& ort_env,
+                                    const std::string& model_filename, 
+                                      const Config::CompileOptions& compile_options_config,
```
kunal-vaishnavi, line 1302: Let's fix the formatting of the parameters

src/models/model.cpp
```diff
@@ -1213,7 +1214,272 @@ OrtSessionOptions* Model::GetSessionOptions(const std::string& model_id) const {
...
+                                      fs::path& out_compiled_model_path) {
+  if (compile_options_config.ep_context_file_path.has_value() && !compile_options_config.ep_context_file_path.value().empty()) {
```
kunal-vaishnavi, line 1304: Could we use [`.value_or`](https://en.cppreference.com/w/cpp/utility/optional/value_or) here instead to simplify obtaining the value?

src/models/model.cpp
```diff
@@ -1213,7 +1214,272 @@ OrtSessionOptions* Model::GetSessionOptions(const std::string& model_id) const {
...
+  
+  // Check if compilation is enabled for the specified model
+  if (!compile_options.has_value()) {
+    // No compile options provided, return full path to original model
+    fs::path full_path = config_->config_path / model_filename;
+    return full_path.string();
+  }
+  
+  const auto& comp_opts = compile_options.value();
+  if (!comp_opts.enable_ep_context.has_value() || !comp_opts.enable_ep_context.value()) {
```
kunal-vaishnavi, line 1342: Can these three `if` blocks be merged into one since they all have the same return value?

src/models/model.cpp
```diff
@@ -1213,7 +1214,272 @@ OrtSessionOptions* Model::GetSessionOptions(const std::string& model_id) const {
...
+    
+    // Check if compilation is enabled for this specific model
+    if (!config_compilation_options.has_value()) {
+      // No compile options, return full path to original model
+      fs::path full_path = config_->config_path / model_filename;
+      return full_path.string();
+    }
+    
+    const auto& comp_opts = config_compilation_options.value();
+    if (!comp_opts.enable_ep_context.has_value() || !comp_opts.enable_ep_context.value()) {
```
kunal-vaishnavi, line 1372: Same question here for the several `if` blocks here

src/models/model.cpp
```diff
@@ -1213,7 +1214,272 @@ OrtSessionOptions* Model::GetSessionOptions(const std::string& model_id) const {
...
+      // Compile from file
+      fs::path input_path = config_->config_path / model_filename;
+      compilation_options->SetInputModelPath(input_path.c_str());
+    }
+
+    // Set output model path - use the path from CheckCompiledModelExists
+    // Ensure the output directory exists
+    fs::path output_dir = compiled_model_path.parent_path();
+    if (!fs::exists(output_dir)) {
+      if (!fs::create_directories(output_dir)) {
```
kunal-vaishnavi, line 1405: Can we merge these two `if` conditions into one?

src/models/model.cpp
```diff
@@ -1213,7 +1214,272 @@ OrtSessionOptions* Model::GetSessionOptions(const std::string& model_id) const {
...
+    if (comp_opts.external_initializers_file_path.has_value() && 
+        comp_opts.external_initializers_size_threshold.has_value()) {
+      fs::path external_init_path = config_->config_path / comp_opts.external_initializers_file_path.value();
+      compilation_options->SetOutputModelExternalInitializersFile(
+          external_init_path.c_str(),
+          comp_opts.external_initializers_size_threshold.value());
+    }
+
+    // Compile the model
+    Ort::CompileModel(ort_env, *compilation_options);
```
kunal-vaishnavi, line 1438: I know there's a lot of checks above but what happens if model compilation fails here? Could `compiled_model_path` point to a non-existent model if a failure occurs?

src/models/model.cpp
```diff
@@ -1213,7 +1214,272 @@ OrtSessionOptions* Model::GetSessionOptions(const std::string& model_id) const {
...
+  }
+
+  // EP Context is only applicable for certain EPs (e.g. NvTensorRTRTX); if not applicable, use original model
+  if (EPContextSupportedProviders(p_device_->GetType()).empty()) {
+    fs::path full_path = config_->config_path / model_filename;
+    return full_path.string();
+  }
+
+  // Helper lambda to configure and compile a model
+  auto compile_model_helper = [this, &ort_env](OrtModelCompilationOptions* compilation_options,
```
kunal-vaishnavi, line 1355: Given how big this helper is, I think it should be its own standalone method.

src/models/model.cpp
```diff
@@ -1213,7 +1214,272 @@ OrtSessionOptions* Model::GetSessionOptions(const std::string& model_id) const {
...
+  };
+
+  // Compile the specified model with the provided compile_options
+  auto compilation_options = CreateModelCompilationOptions(ort_env, session_options);
+  std::string main_model_path = compile_model_helper(compilation_options.get(), model_filename, compile_options);
+
+  // Additionally, compile all pipeline models that have compile_options (if primary session option)
+  // Use explicit pipeline session_options when present, otherwise fallback to main session_options_
+  // (consistent with GetSessionOptions() at runtime).
+  if (is_primary_session_option) {
```
kunal-vaishnavi, line 1451: Can you explain more on why this boolean is needed? Why can't we always go through each pipeline model and check its options?

src/models/model.cpp
```diff
@@ -1213,7 +1214,272 @@ OrtSessionOptions* Model::GetSessionOptions(const std::string& model_id) const {
...
+      }
+    }
+  }
+  
+  // Return the main model path (original or compiled)
+  return main_model_path;
+}
+
+std::unique_ptr<OrtSession> Model::CreateSession(OrtEnv& ort_env, const std::string& model_path, OrtSessionOptions* session_options) {
+
```
kunal-vaishnavi, line 1478:
```suggestion
```

src/models/model.h
```diff
@@ -156,10 +156,84 @@ struct Model : std::enable_shared_from_this<Model>, LeakChecked<Model>, External
...
+   * Context is valid only if: (1) compatibility info is present for this EP, and
+   * (2) GetModelCompatibilityForEpDevices returns OPTIMAL or (PREFER_RECOMPILATION when force_compile_if_needed is false).
+   * All other cases return false.
+   *
+   * \param ort_env OrtEnv (for GetEpDevices)
+   * \param compiled_model_path Path to the compiled model file
+   * \param force_compile_if_needed If true, PREFER_RECOMPILATION is treated as invalid (recompile); if false, it is accepted as valid with a warning
+   * \return true if the compiled model is valid for the current EP (or validation not applicable)
+   */
+  bool ValidateCompiledModel(OrtEnv& ort_env, const fs::path& compiled_model_path, bool force_compile_if_needed);
```
kunal-vaishnavi, line 234:
```suggestion
  bool ValidateCompiledModel(OrtEnv& ort_env, const fs::path& compiled_model_path, bool force_compile);
```

src/models/onnxruntime_api.h
```diff
@@ -443,6 +452,16 @@ void RegisterExecutionProviderLibrary(OrtEnv* env, const char* registration_name
...
 
+/** \brief Compiles an input model to generate a model with EPContext nodes
+ *
+ * Wraps OrtCompileApi::CompileModel
+ * Throws an exception on error
+ *
+ * \param env OrtEnv object
+ * \param model_compilation_options Compilation options for the model
+ */
+void CompileModel(OrtEnv& env, const OrtModelCompilationOptions& model_compilation_options);
```
kunal-vaishnavi, line 463:
```suggestion
void CompileModel(OrtEnv& env, const OrtModelCompilationOptions& compilation_options);
```

src/models/onnxruntime_api.h
```diff
@@ -649,6 +668,94 @@ struct OrtSessionOptions {
...
+   *
+   * Wraps OrtCompileApi::ModelCompilationOptions_SetOutputModelPath
+   */
+  OrtModelCompilationOptions& SetOutputModelPath(const ORTCHAR_T* output_model_path);
+
+  /** \brief Sets external initializers file path and size threshold for the compiled model
+   *
+   * Wraps OrtCompileApi::ModelCompilationOptions_SetOutputModelExternalInitializersFile
+   */
+  OrtModelCompilationOptions& SetOutputModelExternalInitializersFile(const ORTCHAR_T* external_initializers_file_path,
```
kunal-vaishnavi, line 704: Looks like there's some formatting issues with some of the newly added parameters in this header file.

src/models/onnxruntime_inline.h
```diff
@@ -759,6 +763,79 @@ inline OrtSessionOptions& OrtSessionOptions::AppendExecutionProvider_V2(OrtEnv&
...
+  return std::unique_ptr<OrtModelCompilationOptions>(p);
+}
+
+inline OrtModelCompilationOptions& OrtModelCompilationOptions::SetInputModelPath(const ORTCHAR_T* input_model_path) {
+  Ort::ThrowOnError(Ort::GetCompileApi().ModelCompilationOptions_SetInputModelPath(this, input_model_path));
+  return *this;
+}
+
+inline OrtModelCompilationOptions& OrtModelCompilationOptions::SetInputModelFromBuffer(const void* input_model_data, 
+                                                                                         size_t input_model_data_size) {
```
kunal-vaishnavi, line 779: Looks like there's some formatting issues with some of the newly added parameters in this file.

src/config.cpp
```diff
@@ -227,6 +227,39 @@ struct RunOptions_Element : JSON::Element {
...
+      auto level = static_cast<int>(JSON::Get<double>(value));
+      if (level < ORT_DISABLE_ALL || level > ORT_ENABLE_ALL) {
+        throw std::runtime_error("Invalid graph_optimization_level value: " + std::to_string(level));
+      }
+      v_.graph_optimization_level = static_cast<GraphOptimizationLevel>(level);
+    } else if (name == "ep_context_file_path") {
+      v_.ep_context_file_path = JSON::Get<std::string_view>(value);
+    } else if (name == "ep_context_embed_mode") {
+      v_.ep_context_embed_mode = JSON::Get<bool>(value);
+    } else if (name == "flags") {
```
kunal-vaishnavi, line 248: What is the `flags` option? The name doesn't provide much information.

src/models/model.cpp
lnigam: Pipeline compile options live under model.decoder.pipeline only. is_primary_session_option marks the decoder’s CompileModel invocation so we don’t re-compile pipeline stages when the same Model also compiles encoder/vision/embedding (e.g. whisper, multimodal). For decoder-only models with an empty pipeline vector the loop is a no-op.

examples/c/src/common.cpp
lnigam: Will take this up after this PR separately

src/models/marian.cpp
```diff
@@ -8,10 +8,13 @@ namespace Generators {
 MarianModel::MarianModel(std::unique_ptr<Config> config, OrtEnv& ort_env)
     : Model{std::move(config)} {
   encoder_session_options_ = OrtSessionOptions::Create();
-  CreateSessionOptionsFromConfig(config_->model.encoder.session_options.has_value() ? config_->model.encoder.session_options.value() : config_->model.decoder.session_options, *encoder_session_options_, true, false);
+  CreateSessionOptionsFromConfig(config_->model.encoder.session_options.has_value() ? config_->model.encoder.session_options.value() : config_->model.decoder.session_options, *encoder_session_options_, false, false);
```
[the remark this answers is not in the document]
lnigam: this is correct.

baijumeswani:
Closing this PR. Please re-open if this work is sill relevant.

PR #1995, opened 2026-02-28
allow pruned models for prefill

fs-eire:
### Motivation

During prefill (prompt processing), the LM head computes logits for **all** input tokens (`[B, S, vocab_size]`), but only the **last** token's logits are used for sampling. For long prompts, this is a significant waste — the LM head MatMul is one of the most expensive operations due to the large vocabulary dimension. For example, with a 512-token prompt and `vocab_size=200K`, this eliminates ~99.8% of unnecessary computation in the LM head.

### Performance Comparison

This optimization increased Phi-4-mini-instruct prefill tps from 1851.37 to 2757.54 (1.48x) on a nVidia GTX 4070Ti Windows11/x64.

### Approach

Insert a `Gather(axis=1, idx=-1)` + `Unsqueeze(axis=1)` pair before the LM head to extract only the last token's hidden state, changing the MatMul input from `[B, S, H]` to `[B, 1, H]` and the logits output from `[B, S, vocab_size]` to `[B, 1, vocab_size]`.

```
Before:                          After:
  [B, S, H]                       [B, S, H]
      |                               |
  LM Head MatMul                  Gather(axis=1, idx=-1)
      |                               |
  [B, S, vocab]                   [B, H]
                                       |
                                  Unsqueeze(axis=1)
                                       |
                                   [B, 1, H]
                                       |
                                  LM Head MatMul
                                       |
                                  [B, 1, vocab]
```

#### C++ Runtime (`src/models/`)

- **`model.h` / `model.cpp`**: Added `Model::IsPruned()` method that inspects the ONNX output metadata — returns `true` when the logits output has a fixed sequence dimension of 1, indicating the model graph has been pruned.
- **`logits.cpp`**: Extended the `trimmed_prefill_logits_` condition in the `Logits` constructor to also check `state.model_.IsPruned()`. This reuses the existing mechanism (originally for OpenVINO stateful models and RyzenAI) that allocates the output tensor as `[B, 1, vocab_size]` instead of `[B, S, vocab_size]`. **No behavioral change for unmodified models** since `IsPruned()` returns `false` when the sequence dimension is dynamic.

#### Python Model Builder (`src/python/py/models/`)

- **`builder.py`**: Registered `prune_lm_head` as a boolean `extra_option`.
- **`builders/base.py`**:
  - `make_outputs_init()`: Parses `prune_lm_head` from extra options. Warns and disables if combined with `exclude_lm_head`.
  - `make_lm_head()`: When enabled, inserts `Gather` + `Unsqueeze` nodes before the MatMul and sets `seq_dim=1`. Propagates `seq_dim` through all downstream nodes (bias Add, scale Mul, logits mask Where, softcap Div/Tanh/Mul, final Cast).
  - `make_matmul_float()`, `make_matmul_int4()`, `make_matmul_int4_qdq()`: Accept `seq_dim` kwarg for shape annotations (defaults to `"sequence_length"` for backward compatibility).
  - `make_add_bias()`: Same `seq_dim` kwarg support.
  - `make_matmul_lora()`: Uses `seq_dim` from kwargs in the LoRA Add output shape.

### Usage

```bash
# Build a new model with pruned LM head
python builder.py -m my-model -e prune_lm_head=true -o output_dir
```

### Compatibility

- **Backward compatible**: `IsPruned()` returns `false` for all existing models (their logits sequence dimension is dynamic, not fixed to 1), so no behavior change for unmodified models.
- **`exclude_lm_head`**: `prune_lm_head` is silently disabled with a warning if `exclude_lm_head` is also set (no LM head to prune).

### DIFF
--- a/src/cuda/beam_search_scorer_cuda.cpp
+++ b/src/cuda/beam_search_scorer_cuda.cpp
@@ -4,6 +4,7 @@
 #include "generators.h"
 #include "search.h"
 #include "search_cuda.h"
+#include "cuda_common.h"
 #include "beam_search_scorer_cuda.cuh"
 #include "beam_search_scorer_cuda.h"
 #include "interface.h"
@@ -22,7 +23,7 @@ BeamSearchScorer_Cuda::BeamSearchScorer_Cuda(const GeneratorParams& parameters, 
   state_cpu_->not_done_count_ = parameters.search.batch_size;
   state_cpu_->hypothesis_buffer_used_ = 0;
   state_gpu_ = CudaMallocArray<cuda::BeamScorerState>(1);
-  cudaMemcpyAsync(state_gpu_.get(), state_cpu_.get(), sizeof(cuda::BeamScorerState), ::cudaMemcpyHostToDevice, stream_);
+  CUDA_CHECK(cudaMemcpyAsync(state_gpu_.get(), state_cpu_.get(), sizeof(cuda::BeamScorerState), ::cudaMemcpyHostToDevice, stream_));
 
   size_t batch_beam_size = state_cpu_->batch_size_ * state_cpu_->num_beams_;
 
@@ -63,7 +64,7 @@ void BeamSearchScorer_Cuda::Process(Sequences& sequences,
                                        next_tokens,
                                        next_indices,
                                        stream_);
-  cudaEventRecord(event_process_complete_, stream_);
+  CUDA_CHECK(cudaEventRecord(event_process_complete_, stream_));
 
   cuda::LaunchBeamSearchScorer_AppendNextTokenToSequences(*state_cpu_,
                                                           *state_gpu_,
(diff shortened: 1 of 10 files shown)


src/models/logits.cpp
```diff
@@ -18,8 +18,9 @@ Logits::Logits(State& state)
 
   input_sequence_lengths.resize(state_.params_->search.batch_size);
 
-  if (IsOpenVINOStatefulModel(state.model_) || IsRyzenAIPrunedModel(state_.model_)) {
-    // In the case of OpenVINO stateful models or RyzenAI pruned models, they are patched in a way so that they only return the
+  if (IsOpenVINOStatefulModel(state.model_) || IsRyzenAIPrunedModel(state_.model_) || state.model_.IsPruned()) {
```
baijumeswani: IsRyzenPrunedModel has a condition to check for the device type in addition to model_.IsPruned().

https://github.com/microsoft/onnxruntime-genai/blob/3c47932e9d7afa0d44db0b3918e479bbdd4c5353/src/ryzenai/interface.cpp#L212

Could we remove the function `IsRyzenAIPrunedModel` and the references to it in this PR?

PR #1996, opened 2026-02-28
Handle VLM weight name prefixes in QuantizedModel loader

uday610:
This change makes the quantized model weight loader work with VLM checkpoints where weight names include extra prefixes like `model.language_model.*` and vision tower weights like `model.visual.*`.

- Skip weights under `model.visual.*`

- Normalize `model.language_model.* → model.*` before existing pattern matching and layer parsing

No functional change for pure LLM checkpoints.


baijumeswani:
Closing this PR. Please re-open if the work here is still relevant.

PR #1997, opened 2026-03-02
Nemotron ASR Support for Streaming

nenad1002:
image

image

### DIFF
--- a/examples/python/nemotron_speech.py
+++ b/examples/python/nemotron_speech.py
@@ -52,7 +52,6 @@ def simulate_microphone(model_path, audio_path, execution_provider):
     sample_rate, chunk_samples = load_config(model_path)
     audio = load_audio(audio_path, sample_rate)
     duration = len(audio) / sample_rate
-    chunk_duration = chunk_samples / sample_rate
 
     config = get_config(model_path, execution_provider)
     model = og.Model(config)
--- a/examples/csharp/NemotronSpeech/README.md
+++ b/examples/csharp/NemotronSpeech/README.md
@@ -1,0 +1,60 @@
+# Nemotron Speech Streaming ASR — C# Example
+
+This example demonstrates real-time streaming speech recognition using the
+NVIDIA Nemotron Speech Streaming model with the ONNX Runtime GenAI C# API.
+
+Audio is streamed through the model in chunks (simulating a microphone feed),
+and transcribed text is printed incrementally as it becomes available.
+
+## Prerequisites
+
+- .NET 8.0 SDK or later
+- ONNX Runtime GenAI C# package ([installation instructions](https://onnxruntime.ai/docs/genai/howto/install))
+- A Nemotron Speech Streaming ONNX model (e.g., `nvidia/nemotron-speech-streaming-en-0.6b`)
+
+## Build
+
+```bash
+cd examples/csharp/
+dotnet build NemotronSpeech -c Release
+```
+
+## Run
+
+```bash
+cd ./NemotronSpeech/bin/Release/net8.0/
+./NemotronSpeech <model_path> <audio_file.wav> [execution_provider]
+```
+
+### Arguments
+
+| Argument | Description |
+|---|---|
+| `model_path` | Path to the Nemotron ONNX model directory |
+| `audio_file.wav` | Path to a WAV audio file (any sample rate — resampled automatically) |
+| `execution_provider` | *(Optional)* Execution provider: `cpu`, `cuda`, `dml`, or `follow_config` (default: `follow_config`) |
+
+### Example
+
+```bash
+# CPU inference
+./NemotronSpeech /path/to/nemotron-cpu-int4 /path/to/audio.wav
+
+# CUDA inference
+./NemotronSpeech /path/to/nemotron-cuda-int4 /path/to/audio.wav cuda
+```
+
+### Output
+
+The example prints transcribed text incrementally as each audio chunk is processed,
+followed by a summary with the full transcript, audio duration, wall-clock time,
+and real-time factor (RTFx).
+
+```
+------------------------------------------------------------
+ This is an example of streaming speech recognition...
+============================================================
+  This is an example of streaming speech recognition using Nemotron.
+============================================================
+  Audio: 10.50s | Wall: 2.13s | RTFx: 4.93x
+```


src/ort_genai_c.cpp
```diff
@@ -1084,4 +1086,44 @@ void OGA_API_CALL OgaUnregisterExecutionProviderLibrary(const char* registration
...
+  *out = ReturnUnique<OgaStreamingASR>(std::move(asr));
+  return nullptr;
+  OGA_CATCH
+}
+
+OgaResult* OGA_API_CALL OgaStreamingASRTranscribeChunk(OgaStreamingASR* asr, const float* audio_data, size_t num_samples, const char** text) {
+  OGA_TRY
+  std::string result = asr->TranscribeChunk(audio_data, num_samples);
+  *text = AllocOgaString(result);
+  return nullptr;
```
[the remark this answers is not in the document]
nenad1002: This seems to be a pre-existing bug, and I think Copilot is correct, char[] alloc should always be paired with delete[] not delete. Fixing it.

src/models/nemo_streaming_asr.h
```diff
@@ -0,0 +1,57 @@
...
+
+  std::string TranscribeChunk(const float* audio_data, size_t num_samples) override;
+  std::string Flush() override;
+  const std::string& GetTranscript() const override { return full_transcript_; }
+  void Reset() override;
+
+ private:
+  Model& model_;
+  NemotronCacheConfig cache_config_;
+
```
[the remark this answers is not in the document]
nenad1002: Pre-existing design pattern, Copilot is correct, but it's the same throughout codebase.

examples/python/nemotron_streaming.py
```diff
@@ -0,0 +1,99 @@
...
+    if sr != SAMPLE_RATE:
+        import scipy.signal
+        num_samples = int(len(audio) * SAMPLE_RATE / sr)
+        audio = scipy.signal.resample(audio, num_samples).astype(np.float32)
+    return audio
+
+
+def load_tokenizer(model_path):
+    import sentencepiece as spm
+    path = os.path.join(model_path, "tokenizer.model")
```
kunal-vaishnavi: We should obtain the `tokenizer.json` file for this since ORT extensions can support `SentencePiece` tokenizers. This was [previously done](https://huggingface.co/zai-org/chatglm3-6b/discussions/12) for ChatGLM-3 when [AMD integrated it](https://github.com/microsoft/onnxruntime-genai/pull/921).

examples/python/nemotron_streaming.py
```diff
@@ -0,0 +1,99 @@
...
+def simulate_microphone(model_path, audio_path):
+    audio = load_audio(audio_path)
+    duration = len(audio) / SAMPLE_RATE
+    num_chunks = (len(audio) + CHUNK_SAMPLES - 1) // CHUNK_SAMPLES
+    print(f"Audio: {duration:.1f}s | {num_chunks} chunks × {CHUNK_DURATION*1000:.0f}ms")
+
+    config = og.Config(model_path)
+    model = og.Model(config)
+    sp = load_tokenizer(model_path)
+    asr = og.StreamingASR(model)
```
kunal-vaishnavi: I get the reasoning behind having a different object instead of a generator for ASR scenarios. There are multiple ASR pipelines from Conformer-Transducer to Sequence-to-Sequence to CTC. However, each of these pipelines can have its own process and the `StreamingASR` name is not specific enough to clarify which pipeline it could encapsulate. Can we rename this to be clearer?

examples/python/nemotron_streaming.py
```diff
@@ -0,0 +1,99 @@
...
+
+    print("-" * 60)
+    stream_start = time.time()
+
+    for i in range(0, len(audio), CHUNK_SAMPLES):
+        chunk = audio[i:i + CHUNK_SAMPLES]
+        if len(chunk) < CHUNK_SAMPLES:
+            chunk = np.pad(chunk, (0, CHUNK_SAMPLES - len(chunk)))
+        chunk = chunk.astype(np.float32)
+        raw_text = asr.transcribe_chunk(chunk)
```
kunal-vaishnavi: Can this be named as `transcribe` since the chunk is already passed in as a parameter?

examples/python/nemotron_streaming.py
```diff
@@ -0,0 +1,99 @@
...
+
+    for _ in range(4):
+        silence = np.zeros(CHUNK_SAMPLES, dtype=np.float32)
+        raw_text = asr.transcribe_chunk(silence)
+        if raw_text:
+            print(raw_text, end="", flush=True)
+
+    total_wall = time.time() - stream_start
+
+    full_raw = asr.get_transcript()
```
kunal-vaishnavi: If we already accumulate the raw text from each streamed chunk, what is the purpose of this then?

examples/python/nemotron_streaming.py
```diff
@@ -0,0 +1,99 @@
...
+    path = os.path.join(model_path, "tokenizer.model")
+    if not os.path.exists(path):
+        return None
+    sp = spm.SentencePieceProcessor()
+    sp.Load(path)
+    return sp
+
+
+def parse_token_ids(raw_text):
+    return [int(m.group(1)) for m in re.finditer(r'<(\d+)>', raw_text)]
```
kunal-vaishnavi: Can the regex matching and extraction be done inside `Tokenizer.Decode` within ORT extensions instead?

src/models/nemo_streaming_asr.cpp
```diff
@@ -0,0 +1,324 @@
...
+
+#include "../generators.h"
+#include "nemo_streaming_asr.h"
+
+namespace Generators {
+
+void NemoStreamingASR::LoadVocab() {
+  if (vocab_loaded_) return;
+
+  auto tokenizer = model_.CreateTokenizer();
```
kunal-vaishnavi: I think generating a `tokenizer.json` file that can be used with ORT extensions will remove the need to load the vocab here.

src/models/nemo_streaming_asr.cpp
```diff
@@ -0,0 +1,324 @@
...
+}
+
+std::string NemoStreamingASR::TranscribeChunk(const float* audio_data, size_t num_samples) {
+  LoadVocab();
+
+  // Append incoming audio to accumulation buffer
+  audio_buffer_.insert(audio_buffer_.end(), audio_data, audio_data + num_samples);
+
+  std::string result;
+  const size_t chunk_sz = static_cast<size_t>(cache_config_.chunk_samples);
```
kunal-vaishnavi: Let's use full names instead of abbreviations to be clear (e.g. I believe this is `chunk_size`).

src/models/nemo_streaming_asr.cpp
```diff
@@ -0,0 +1,324 @@
...
+  encoder_cache_.Reset(cache_config_, allocator);
+  decoder_state_.Reset(cache_config_, allocator);
+  full_transcript_.clear();
+  mel_extractor_.Reset();
+  mel_pre_encode_cache_.assign(
+      static_cast<size_t>(cache_config_.num_mels) * cache_config_.pre_encode_cache_size, 0.0f);
+  audio_buffer_.clear();
+}
+
+std::string NemoStreamingASR::TranscribeChunk(const float* audio_data, size_t num_samples) {
```
kunal-vaishnavi: What if the audio data is float16 instead of float32?

src/models/nemo_streaming_asr.cpp
```diff
@@ -0,0 +1,324 @@
...
+  LoadVocab();
+
+  // Append incoming audio to accumulation buffer
+  audio_buffer_.insert(audio_buffer_.end(), audio_data, audio_data + num_samples);
+
+  std::string result;
+  const size_t chunk_sz = static_cast<size_t>(cache_config_.chunk_samples);
+
+  // Process chunks as soon as you get it full chunk size.
```
kunal-vaishnavi:
```suggestion
  // Process chunks as soon as you get the full chunk size.
```

src/models/nemo_streaming_asr.cpp
```diff
@@ -0,0 +1,324 @@
...
+std::string NemoStreamingASR::TranscribeMelChunk(const std::vector<float>& mel_data, int num_frames) {
+  auto& allocator = model_.allocator_cpu_;
+  int cache_size = cache_config_.pre_encode_cache_size;
+  const int num_mels = cache_config_.num_mels;
+
+  int total_mel_frames = cache_size + num_frames;
+
+  // Create processed_signal: [1, num_mels, total_mel_frames]
+  auto signal_shape = std::array<int64_t, 3>{1, num_mels, total_mel_frames};
+  auto processed_signal = OrtValue::CreateTensor(allocator, signal_shape, ONNX_TENSOR_ELEMENT_DATA_TYPE_FLOAT);
```
kunal-vaishnavi: Can we get the dtype from the model instead of assuming float32?

src/models/nemo_streaming_asr.cpp
```diff
@@ -0,0 +1,324 @@
...
+  float* signal_data = processed_signal->GetTensorMutableData<float>();
+
+  // Fill row by row: mel is [num_mels, time] in row-major layout
+  for (int m = 0; m < num_mels; ++m) {
```
kunal-vaishnavi: For these memory copies, can we construct `DeviceSpan` objects and use [`CopyFrom`](https://github.com/microsoft/onnxruntime-genai/blob/f02bebe8a20fdb91560c98682017095d56c966db/src/smartptrs.h#L74) to leverage each EP's implementation of memory copy? [Here](https://github.com/microsoft/onnxruntime-genai/blob/f02bebe8a20fdb91560c98682017095d56c966db/src/models/kv_cache.cpp#L105-L109) is an example.

src/models/nemo_streaming_asr.cpp
```diff
@@ -0,0 +1,324 @@
...
+    for (int m = 0; m < num_mels; ++m) {
+      std::memmove(mel_pre_encode_cache_.data() + m * cache_size,
+                   mel_pre_encode_cache_.data() + m * cache_size + num_frames,
+                   keep * sizeof(float));
+      std::memcpy(mel_pre_encode_cache_.data() + m * cache_size + keep,
+                  mel_data.data() + m * num_frames,
+                  num_frames * sizeof(float));
+    }
+  }
+
```
kunal-vaishnavi: The above sections to invoke several `memcpy` and `memmove` calls appear to be optimization opportunities. One opportunity I can think of is using offset and indexing calculations to avoid some `memmove` and `memcpy` calls altogether.

Here is what Copilot generated as suggestions for improving performance.

---

You can simplify it a lot, and in a way that’s *usually* faster (or at least no worse) by changing the representation of your “pre-encode cache” from **per-mel rows** (shape `[num_mels, cache_size]`) to a **ring buffer over time** (shape `[cache_size, num_mels]`, i.e. time-major), then doing:

- **0 memmove/memcpy per mel row** to maintain the cache
- **1 pass** to write new frames into the ring
- **1 pass** to materialize `processed_signal` contiguous as `[num_mels, total_mel_frames]` when you actually need to feed the ONNX input

Right now you’re paying:
- `num_mels` memcpy’s to prepend cache into `processed_signal`
- `num_mels` memcpy’s to append new mel into `processed_signal`
- and then **either** `num_mels` memcpy’s (long chunk) **or** `num_mels` memmove+memcpy (short chunk) to update the cache

That’s a lot of small copies.

## Option A (best simplification): make the cache a ring buffer (time-major)

### 1) Store cache as time-major + head index
Instead of `mel_pre_encode_cache_[m * cache_size + t]`, use:

- `mel_cache_time_major_[t * num_mels + m]`
- `cache_write_pos_` in `[0, cache_size)`

When you get `mel_data` (which in your code is mel-major: `[m * num_frames + t]`), write frames into the ring:

```cpp
// members:
std::vector<float> mel_cache_;   // size = cache_size * num_mels, time-major
int cache_pos_ = 0;              // next time index to write (ring)

// on reset/init: mel_cache_.assign(cache_size * num_mels, 0.f); cache_pos_ = 0;

for (int t = 0; t < num_frames; ++t) {
  float* dst = mel_cache_.data() + (cache_pos_ * num_mels);
  for (int m = 0; m < num_mels; ++m) {
    dst[m] = mel_data[m * num_frames + t];
  }
  cache_pos_ = (cache_pos_ + 1) % cache_size;
}
```

No memmove, no shifting, no special-casing short chunks.

### 2) Materialize `processed_signal` from the ring + current chunk
You still need a contiguous `[1, num_mels, cache_size + num_frames]` tensor. Build it in two phases:

- write cache frames in chronological order from the ring
- write the current chunk frames

Because your ONNX input expects mel-major `[m][time]`, you’ll do a nested loop anyway — but it’s *one straightforward loop*, no “copy this row, then memmove cache”, etc.

Pseudo:

```cpp
// signal_data is mel-major: signal_data[m * total + time]

auto write_cache_in_order = [&](int time_offset) {
  // oldest frame is cache_pos_ (the next write position)
  for (int t = 0; t < cache_size; ++t) {
    int idx = (cache_pos_ + t) % cache_size;
    const float* src = mel_cache_.data() + idx * num_mels; // [num_mels]
    for (int m = 0; m < num_mels; ++m) {
      signal_data[m * total_mel_frames + (time_offset + t)] = src[m];
    }
  }
};

auto write_chunk = [&](int time_offset) {
  for (int t = 0; t < num_frames; ++t) {
    for (int m = 0; m < num_mels; ++m) {
      signal_data[m * total_mel_frames + (time_offset + t)] = mel_data[m * num_frames + t];
    }
  }
};

write_cache_in_order(0);
write_chunk(cache_size);
```

That’s it: no `memcpy`/`memmove` cache maintenance at all, and no row-wise copies.

**Why it’s not slower:** you’re already touching basically all elements to build `processed_signal`; the ring buffer approach makes cache update O(num_frames * num_mels) with good locality and eliminates the extra passes (especially the short-chunk memmove).

---

## Option B (minimal change): use `std::copy` / `std::rotate` and make it clearer (but not necessarily faster)

If you want to keep your current layout, you can improve readability but you won’t remove the fundamental work.

For the “short chunk” case you can replace per-row `memmove` with a per-row `std::move`/`std::copy` (still moves the same bytes), or do a `std::rotate` per row:

```cpp
auto* row = mel_pre_encode_cache_.data() + m * cache_size;
std::rotate(row, row + num_frames, row + cache_size);
std::copy(mel_data.data() + m * num_frames,
          mel_data.data() + m * num_frames + num_frames,
          row + (cache_size - num_frames));
```

This is “simpler”, but it’s still `num_mels` operations and still shifting memory.

---

## Option C (bigger win): avoid building `processed_signal` at all (only if the encoder model can accept 2 inputs)
The real “copy killer” is: you’re concatenating cache + chunk into a new tensor every chunk. If you can modify the encoder graph (or it already supports it), you could pass cache and new mel as separate inputs and do concatenation inside the model (or eliminate it entirely by changing the model). That’s the most efficient, but it’s a model/graph contract change.

---

## What I’d recommend here
- If you can change only C++: **Option A (ring buffer)** gives the best tradeoff: simpler logic *and* fewer total memory passes.
- If you want a tiny refactor only: Option B cleans up code, but don’t expect perf gains.

---

src/models/nemo_streaming_asr.cpp
```diff
@@ -0,0 +1,324 @@
...
+  // Create processed_signal_length: [1]
+  auto len_shape = std::array<int64_t, 1>{1};
+  auto signal_length = OrtValue::CreateTensor(allocator, len_shape, ONNX_TENSOR_ELEMENT_DATA_TYPE_INT64);
```
kunal-vaishnavi: Same comment about getting the dtype from the inference session

src/models/nemo_streaming_asr.cpp
```diff
@@ -0,0 +1,324 @@
...
+      encoder_cache_.cache_last_channel_len.get()};
+
+  const char* enc_output_names[] = {
+      cache_config_.enc_out_encoded.c_str(), cache_config_.enc_out_length.c_str(),
+      cache_config_.enc_out_cache_channel.c_str(), cache_config_.enc_out_cache_time.c_str(),
+      cache_config_.enc_out_cache_channel_len.c_str()};
+
+  // Run encoder
+  auto run_options = OrtRunOptions::Create();
+  auto enc_outputs = encoder_session_->Run(
```
kunal-vaishnavi: We should route session running through ORT GenAI rather than using direct ORT APIs. To do this, we can create a `Model` subclass and a `State` subclass per inference session and then have a parent `Model` and parent `State` to orchestrate the logic. [Here](https://github.com/microsoft/onnxruntime-genai/blob/165037b1f77503dd57bcb77c8eaa4115e6cf3f3c/src/models/multi_modal.cpp#L63) is an example of the parent `Model` and [here](https://github.com/microsoft/onnxruntime-genai/blob/165037b1f77503dd57bcb77c8eaa4115e6cf3f3c/src/models/multi_modal.cpp#L218) is an example of the parent `State`.

This will also pass in the run options and session options from the GenAI config directly to the inference sessions. [Here](https://github.com/microsoft/onnxruntime-genai/blob/165037b1f77503dd57bcb77c8eaa4115e6cf3f3c/src/models/multi_modal.cpp#L121-L141) is an example of using them from the config.

src/models/nemo_streaming_asr.cpp
```diff
@@ -0,0 +1,324 @@
...
+
+  auto enc_info = encoder_output->GetTensorTypeAndShapeInfo();
+  auto enc_shape = enc_info->GetShape();
+  int64_t hidden_dim = enc_shape[1];
+
+  // Decode ALL encoder output frames, pre-encode cache is already removed by the ONNX graph.
+  int64_t time_steps = std::min(enc_shape[2], encoded_len);
+  const float* enc_data = encoder_output->GetTensorData<float>();
+
+  auto run_options = OrtRunOptions::Create();
```
kunal-vaishnavi: Same comments about dtypes, run options, session running, etc.

src/models/nemo_streaming_asr.h
```diff
@@ -0,0 +1,56 @@
...
+  OrtSession* encoder_session_{};
```
kunal-vaishnavi: These should be `std::unique_ptr<OrtSession>`  and in the parent `Model` class to [be consistent with other files](https://github.com/microsoft/onnxruntime-genai/blob/165037b1f77503dd57bcb77c8eaa4115e6cf3f3c/src/models/multi_modal.h#L23-L26).

src/models/nemotron_speech.cpp
```diff
@@ -0,0 +1,167 @@
...
+#include "../generators.h"
+#include "nemo_mel_spectrogram.h"
+#include "nemotron_speech.h"
+
+namespace Generators {
+
+void NemotronCacheConfig::PopulateFromConfig(const Config& config) {
+  const auto& enc = config.model.encoder;
+  const auto& dec = config.model.decoder;
+  const auto& sp = config.model.speech;
```
kunal-vaishnavi: If there are three ONNX models (encoder, decoder, joiner), why are there four sections (encoder, decoder, speech, joiner) used in the GenAI config? We can put these attributes in the `model` section directly since they are properties of the full model. [Here](https://huggingface.co/microsoft/Fara-7B-onnx/blob/main/npu/qnn-int4/genai_config.json) is an example.

src/models/nemotron_speech.cpp
```diff
@@ -0,0 +1,167 @@
...
+  dec_out_prednet_lengths = dec.outputs.prednet_lengths;
+  dec_out_states_1 = dec.outputs.states_1;
+  dec_out_states_2 = dec.outputs.states_2;
+}
+
+void NemotronEncoderCache::Initialize(const NemotronCacheConfig& cfg, OrtAllocator& allocator) {
+  // cache_last_channel: [batch, num_layers, left_context, hidden_dim]
+  auto ch_shape = std::array<int64_t, 4>{1, cfg.num_encoder_layers, cfg.left_context, cfg.hidden_dim};
+  cache_last_channel = OrtValue::CreateTensor(allocator, ch_shape, ONNX_TENSOR_ELEMENT_DATA_TYPE_FLOAT);
+  std::memset(cache_last_channel->GetTensorMutableRawData(), 0,
```
kunal-vaishnavi: Once you convert these to `DeviceSpan` objects, you can device-memset a tensor with the [`Zero`](https://github.com/microsoft/onnxruntime-genai/blob/f02bebe8a20fdb91560c98682017095d56c966db/src/smartptrs.h#L71-L72) API like [this](https://github.com/microsoft/onnxruntime-genai/blob/f02bebe8a20fdb91560c98682017095d56c966db/src/models/whisper.cpp#L85).

src/models/nemotron_speech.h
```diff
@@ -0,0 +1,116 @@
...
+
+  void Initialize(const NemotronCacheConfig& cfg, OrtAllocator& allocator);
+  void Reset(const NemotronCacheConfig& cfg, OrtAllocator& allocator);
+};
+
+/// Holds the RNNT decoder LSTM hidden states between decoding steps.
+struct NemotronDecoderState {
+  std::unique_ptr<OrtValue> state_1;
+  std::unique_ptr<OrtValue> state_2;
+  int last_token{0};  // Last emitted non-blank token (for autoregressive feedback)
```
kunal-vaishnavi, line 89: What do you mean by autoregressive feedback if inference with Nemotron is not autoregressive?

src/python/python.cpp
```diff
@@ -476,7 +476,8 @@ PYBIND11_MODULE(onnxruntime_genai, m) {
       .def_property_readonly("type", [](const OgaModel& model) -> std::string { return model.GetType().p_; })
       .def_property_readonly(
           "device_type", [](const OgaModel& model) -> std::string { return model.GetDeviceType().p_; }, "The device type the model is running on")
-      .def("create_multimodal_processor", [](const OgaModel& model) { return OgaMultiModalProcessor::Create(model); });
+      .def("create_multimodal_processor", [](const OgaModel& model) { return OgaMultiModalProcessor::Create(model); })
+      .def("create_streaming_asr", [](OgaModel& model) { return OgaStreamingASR::Create(model); }, "Create a StreamingASR instance for real-time streaming speech recognition.");
```
kunal-vaishnavi: Why is `OgaModel` not `const`?

src/config.cpp
```diff
@@ -315,6 +315,14 @@ struct DecoderInputs_Element : JSON::Element {
       v_.past_sequence_lengths = JSON::Get<std::string_view>(value);
     } else if (name == "block_table") {
       v_.block_table = JSON::Get<std::string_view>(value);
+    } else if (name == "targets") {
+      v_.targets = JSON::Get<std::string_view>(value);
+    } else if (name == "target_length") {
+      v_.target_length = JSON::Get<std::string_view>(value);
+    } else if (name == "states_1") {
```
kunal-vaishnavi: Can we use more descriptive names for `states_1` and `states_2`?

src/config.h
```diff
@@ -202,6 +202,32 @@ struct Config {
...
+      int left_context{};
+      int conv_context{};
+      int pre_encode_cache_size{};
+      int sample_rate{};
+      int chunk_samples{};
+      int blank_id{};
+      int max_symbols_per_step{};
+
+      // Cache-aware streaming encoder I/O names
+      std::string enc_in_length{"length"};
```
kunal-vaishnavi: Can we move these default strings into the `Defaults::VarName` scope?

src/generators.cpp
```diff
@@ -4,6 +4,7 @@
 // Modifications Copyright(C) 2026 Advanced Micro Devices, Inc. All rights reserved.
 
 #include "generators.h"
+#include "streaming_asr.h"
```
kunal-vaishnavi: Why is this import needed if generators are not being used?

src/models/nemo_streaming_asr.cpp
nenad1002: tokenizer.json file exists, maybe from the samples it was not clear as there was some leftover code, I'll fix samples. But we follow this structure.

src/models/nemotron_speech.cpp
nenad1002: I am cleaning up speech as we speek as some should be under encoder/decoder/joint, I can put others to`model`.

src/models/nemotron_speech.h
nenad1002: It's not the standard autoregressive as with LLMs, but technically it's still autoregressive in nature due to LSTM. 

RNN-Ts are considered autoregressive decoders, but usually when people say `autoregressive` they mean LLM-like decoders. (seq2seq decoders like Whisper).

src/config.cpp
nenad1002: Yes, waiting on finalized ONNX model to complete this, then I'm going to change this.

src/models/nemo_streaming_asr.cpp
nenad1002: mel spectogram is too sensitive to feed in float16, haven't seen such a case, if it happens, can always come back and change it.

src/generators.cpp
nenad1002: It is used by memory leak checker (line 90).

src/models/nemo_streaming_asr.cpp
nenad1002: I've seen this and was thinking about it, we should probably use what you say. The problem was the Streaming ASR doesn't quite fit in here again (due to GeneratorParams), I'll try to see how to best support Model/State pattern if that's what you prefer. It might not be 1:1 match with your proposal.

examples/python/nemotron_streaming.py
nenad1002: https://huggingface.co/nenad1002/nemotron-onnx-quantized/tree/main

This is the model structure (genai_config will still change a bit), but tokenizer.json is used. This is the old code I haven't deleted, I'll remove this, no tokenizer.model is used.

examples/python/nemotron_streaming.py
nenad1002: My hope was to have a single abstraction for StreamingASR, even though they work differently, transcribe_chunk will always return a chunk regardless of the ASR model. This abstraction should work for any streaming ASR, do you have any in mind where it would not work?

examples/python/nemotron_streaming.py
nenad1002: It can, but it's convention to use _chunk for streaming, some other major platforms use it. Whisper has process_audio(audio) even though it's obvious from argument its the audio. Same for get_transcript() is often kept for convinience even though not needed as with Azure Speech. 
But it is useful if your intermediate chunks adding up does not equal to full final transcription (true for models where you have hypothesis and final transcription, sometimes hypothesis will be wrong, but the model corrects itself, Nemotron is not like it, but some models are).
Keep in mind streaming models are not only used for live transcription, but can be used only for final transcription quality, and you might have custom postprocessing, again not used with Nemotron, but some other model might use it.

examples/python/nemotron_streaming.py
nenad1002: Removing this, dead code, and yes, it should work the way you say

src/python/python.cpp
nenad1002: Cause it maintains/modifies internal state

src/models/nemo_streaming_asr.cpp
nenad1002: Mel spectogram is always calculated on CPU, so for this line here, CopyFrom migt not make sense, but it does make sense for some other lines that can execute on any EP, I'll try to change it there

src/models/nemo_streaming_asr.cpp
nenad1002: A) gives gain of maybe 1%, but could be stastical error, so not much gain, but RTFX was on avg 1% higher, so will go with A)

src/models/nemo_streaming_asr.cpp
```diff
@@ -0,0 +1,334 @@
...
+  auto tgt_len_type = model_.session_info_.GetInputDataType(cache_config_.dec_in_target_length);
+  auto target_length = OrtValue::CreateTensor(allocator, tgt_len_shape, tgt_len_type);
+  *target_length->GetTensorMutableData<int64_t>() = 1;
+
+  const int max_sym = cache_config_.max_symbols_per_step;
+
+  for (int64_t t = 0; t < time_steps; ++t) {
+    // Fill encoder frame data (reusing pre-allocated tensor)
+    for (int64_t d = 0; d < hidden_dim; ++d) {
+      frame_data[d] = enc_data[d * enc_shape[2] + t];
```
nenad1002: This is strided copy and non-efficient, will wait for transpose on the model, and then impl this as DeviceSpan.

src/models/nemo_streaming_asr.cpp
nenad1002: I can implement this, but again, it's not really suitable for this model. State's Run method does not make much sense for us here, so would feed it with dummy data, if you are fine with that, I can go with it. 
Creating a new abstraction could also be a solution, and maybe have State for Generators and other models, let me know if you would prefer that.

src/models/nemotron_speech.cpp
```diff
@@ -0,0 +1,364 @@
...
+  encoder_frame_ = OrtValue::CreateTensor(allocator, frame_shape, enc_out_type);
+
+  auto targets_type = model_.session_info_.GetInputDataType(cache_config_.dec_in_targets);
+  auto targets_shape = std::array<int64_t, 2>{1, 1};
+  targets_ = OrtValue::CreateTensor(allocator, targets_shape, targets_type);
+
+  auto tgt_len_type = model_.session_info_.GetInputDataType(cache_config_.dec_in_target_length);
+  auto tgt_len_shape = std::array<int64_t, 1>{1};
+  target_length_ = OrtValue::CreateTensor(allocator, tgt_len_shape, tgt_len_type);
+  *target_length_->GetTensorMutableData<int64_t>() = 1;
```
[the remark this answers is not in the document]
nenad1002: Not relevant

src/models/nemotron_speech.cpp
```diff
@@ -0,0 +1,364 @@
...
+    throw std::runtime_error("No mel input set. Call generator.set_model_input(\"audio_features\", mel) first.");
+
+  OrtValue* mel_tensor = current_mel_->ort_tensor_.get();
+  int64_t total_mel_frames = mel_tensor->GetTensorTypeAndShapeInfo()->GetShape()[1];
+
+  auto& allocator = model_.allocator_cpu_;
+  auto len_shape = std::array<int64_t, 1>{1};
+  auto len_type = model_.session_info_.GetInputDataType(cache_config_.enc_in_length);
+  auto signal_length = OrtValue::CreateTensor(allocator, len_shape, len_type);
+  *signal_length->GetTensorMutableData<int64_t>() = total_mel_frames;
```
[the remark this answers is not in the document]
nenad1002: Not relevant

src/models/nemotron_speech.cpp
```diff
@@ -0,0 +1,364 @@
...
+  auto& allocator = model_.allocator_cpu_;
+
+  // Run decoder
+  while (time_step_ < time_steps) {
+    // Copy current encoder frame
+    auto src_frame = enc_span.subspan(static_cast<size_t>(time_step_) * frame_bytes, frame_bytes);
+    frame_span.CopyFrom(src_frame);
+
+    // Run prediction network
+    *targets_->GetTensorMutableData<int64_t>() = decoder_state_.last_token;
```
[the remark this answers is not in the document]
nenad1002: Not relevant

src/models/nemotron_streaming_processor.cpp
```diff
@@ -0,0 +1,134 @@
...
+  audio_buffer_.insert(audio_buffer_.end(), audio_data, audio_data + num_samples);
+
+  const size_t chunk_size = static_cast<size_t>(cache_config_.chunk_samples);
+
+  // Process the first complete chunk available
+  if (audio_buffer_.size() >= chunk_size) {
+    auto mel = BuildMelTensor(audio_buffer_.data(), chunk_size);
+    audio_buffer_.erase(audio_buffer_.begin(),
+                        audio_buffer_.begin() + static_cast<ptrdiff_t>(chunk_size));
+    auto result = std::make_unique<NamedTensors>();
```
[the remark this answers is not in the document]
nenad1002: I would have to significantly change code for little benefit, this is not bottleneck, this is moving few Kbs of data at its worst, in most cases negligble. ~500 cycles at most is spent on this, while mel spectogram can consume 500k cycles, hence thousand times more computation is spent in mel spectogram for each move. Negligble.

examples/csharp/StreamingASR/Program.cs
```diff
@@ -0,0 +1,92 @@
...
+
+Console.WriteLine($"\n{new string('=', 60)}");
+Console.WriteLine($"  {fullTranscript.Trim()}");
+Console.WriteLine(new string('=', 60));
+
+static float[] LoadWavAudio(string path) {
+  using var reader = new BinaryReader(File.OpenRead(path));
+
+  // Skip RIFF header (44 bytes for standard WAV)
+  reader.ReadBytes(44);
```
kunal-vaishnavi: If your example is assuming that only a WAV file is given, can we have an assert check above to ensure that the audio file provided is a WAV file?

Separately, can we support other popular audio file extensions (since not every audio file is a WAV)?

examples/csharp/NemotronSpeech/NemotronSpeech.csproj
```diff
@@ -0,0 +1,16 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
```
kunal-vaishnavi, line 2: Can we name this project folder and `.csproj` file as `NemotronSpeech` instead? If we have future streaming ASR models that can use this identical setup, then we can deal with renaming to `StreamingASR`. Currently, there are still some Nemotron speech specific values such as the chunk size.

examples/python/nemotron_speech.py
```diff
@@ -0,0 +1,99 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
```
kunal-vaishnavi, line 3: Can we rename this file to `nemotron_speech.py` so that the C# and Python filenames match?

examples/python/nemotron_streaming.py
```diff
@@ -0,0 +1,99 @@
...
+
+import argparse
+import os
+import sys
+import time
+import numpy as np
+import onnxruntime_genai as og
+
+SAMPLE_RATE = 16000
+CHUNK_SAMPLES = 8960
```
kunal-vaishnavi: Are these values that we can pull from the GenAI config or calculate from other values in the GenAI config rather than hardcoding? Same question for the C# example as well.

examples/python/nemotron_speech.py
```diff
@@ -0,0 +1,99 @@
...
+def load_audio(audio_path):
+    import soundfile as sf
+    audio, sr = sf.read(audio_path, dtype="float32")
+    if len(audio.shape) > 1:
+        audio = audio.mean(axis=1)
+    if sr != SAMPLE_RATE:
+        import scipy.signal
+        num_samples = int(len(audio) * SAMPLE_RATE / sr)
+        audio = scipy.signal.resample(audio, num_samples).astype(np.float32)
```
kunal-vaishnavi, line 32: If we show resampling in the Python example, then we should show resampling in the C# example as well. It's fine if additional libraries have to be imported to enable this.

We want to keep feature parity across our examples in different languages. That way, users don't feel they have to pick a particular language binding to gain access to additional features.

examples/python/nemotron_speech.py
```diff
@@ -0,0 +1,99 @@
...
+
+def simulate_microphone(model_path, audio_path):
+    """Stream audio through Generator + StreamingProcessor API."""
+    audio = load_audio(audio_path)
+    duration = len(audio) / SAMPLE_RATE
+    num_chunks = (len(audio) + CHUNK_SAMPLES - 1) // CHUNK_SAMPLES
+    print(f"Audio: {duration:.1f}s | {num_chunks} chunks × {CHUNK_DURATION*1000:.0f}ms")
+
+    config = og.Config(model_path)
+    model = og.Model(config)
```
kunal-vaishnavi, line 57: If you are not modifying the config object in any way, you can simply do `model = og.Model(model_path)`. The `og.Config` construct exists for when you want to modify the config object in code.

Separately, can we re-use methods defined in the `common` utilities in both C# and Python? These common utilities are defined to deal with EP-specific cases across our examples. For example, we can use `GetConfig` / `get_config` for obtaining the config and model while maintaining EP-specific config settings.

examples/python/nemotron_speech.py
```diff
@@ -0,0 +1,99 @@
...
+    tokenizer_stream = tokenizer.create_stream()
+    params = og.GeneratorParams(model)
+    generator = og.Generator(model, params)
+
+    print("-" * 60)
+    stream_start = time.time()
+    full_transcript = ""
+
+    def decode_chunk():
+        nonlocal full_transcript
```
kunal-vaishnavi: Why use `nonlocal` on `full_transcript` if you are already referring to other variables above it like `generator` and `tokenizer_stream`? It would be cleaner to move `decode_chunk` to its own function outside of `simulate_microphone` that takes in some parameters and returns the full transcript.

examples/python/nemotron_streaming.py
```diff
@@ -0,0 +1,99 @@
...
+
+    # Flush remaining audio
+    inputs = processor.flush()
+    if inputs is not None:
+        generator.set_inputs(inputs)
+        decode_chunk()
+
+    # Feed silence chunks for right context
+    for _ in range(4):
+        silence = np.zeros(CHUNK_SAMPLES, dtype=np.float32)
```
kunal-vaishnavi: Can you explain more on the goal here? Why does feeding in silence chunks at the end after flushing out the remaining audio help?

src/csharp/NativeMethods.cs
```diff
@@ -440,5 +440,35 @@ public static extern UIntPtr OgaSequencesGetSequenceCount(IntPtr /* const OgaSeq
...
+
+        // NamedTensors creation API
+        [DllImport(NativeLib.DllName, CallingConvention = CallingConvention.Winapi)]
+        public static extern IntPtr /* OgaResult* */ OgaCreateNamedTensors(out IntPtr /* OgaNamedTensors** */ namedTensors);
+
+        [DllImport(NativeLib.DllName, CallingConvention = CallingConvention.Winapi)]
+        public static extern IntPtr /* OgaResult* */ OgaNamedTensorsSet(IntPtr /* OgaNamedTensors* */ namedTensors,
+                                                                         byte[] /* const char* */ name,
+                                                                         IntPtr /* OgaTensor* */ tensor);
+
```
kunal-vaishnavi: I see the following native APIs are being used.

- `OgaCreateStreamingProcessor`
- `OgaStreamingProcessorProcess`
- `OgaStreamingProcessorFlush`
- `OgaStreamingProcessorReset`
- `OgaDestroyStreamingProcessor`

The above two `NamedTensors` native APIs are not used and can be removed.

src/models/nemotron_speech.cpp
```diff
@@ -0,0 +1,364 @@
...
+  left_context = config.model.left_context;
+  conv_context = config.model.conv_context;
+  pre_encode_cache_size = config.model.pre_encode_cache_size;
+  sample_rate = config.model.sample_rate;
+  chunk_samples = config.model.chunk_samples;
+  blank_id = config.model.blank_id;
+  max_symbols_per_step = config.model.max_symbols_per_step;
+
+  // Vocab size from top-level config (includes blank)
+  vocab_size = config.model.vocab_size - 1;
```
kunal-vaishnavi: Why do we need to subtract 1?

src/models/nemotron_speech.cpp
```diff
@@ -0,0 +1,364 @@
...
+  dec_in_targets = dec.inputs.targets;
+  dec_in_target_length = dec.inputs.target_length;
+  dec_in_lstm_hidden = dec.inputs.lstm_hidden_state;
+  dec_in_lstm_cell = dec.inputs.lstm_cell_state;
+  dec_out_outputs = dec.outputs.outputs;
+  dec_out_prednet_lengths = dec.outputs.prednet_lengths;
+  dec_out_lstm_hidden = dec.outputs.lstm_hidden_state;
+  dec_out_lstm_cell = dec.outputs.lstm_cell_state;
+}
+
```
kunal-vaishnavi, line 75: Can we read the dtypes for the created tensors from the ORT inference sessions? That way, we don't need to modify the core code in the future if we want to test out other dtype combinations.

src/models/nemotron_speech.cpp
```diff
@@ -0,0 +1,364 @@
...
+      nemotron_model_{model} {
+  cache_config_ = model.cache_config_;
+
+  auto& allocator = model_.allocator_cpu_;
+  auto& device = *model_.p_device_;
+  encoder_cache_.Initialize(cache_config_, allocator, device);
+  decoder_state_.Initialize(cache_config_, allocator, device);
+
+  // Pre-allocate reusable decoder tensors
+  run_options_ = OrtRunOptions::Create();
```
kunal-vaishnavi: In order to set the run options from the GenAI config to this object, we need to call `State::SetRunOptions` like the following.

https://github.com/microsoft/onnxruntime-genai/blob/bc815abe944d3d2ae8babf35e385c968f01337f2/src/models/whisper.cpp#L46-L52

src/models/nemotron_speech.cpp
```diff
@@ -0,0 +1,364 @@
...
+                                           DeviceSpan<int32_t>& /*next_tokens*/,
+                                           DeviceSpan<int32_t> /*next_indices*/) {
+  throw std::runtime_error(
+      "NemotronSpeechState::Run() is not used directly. "
+      "Use Generator::GenerateNextToken() with set_inputs.");
+}
+
+void NemotronSpeechState::SetExtraInputs(const std::vector<ExtraInput>& extra_inputs) {
+  for (const auto& input : extra_inputs) {
+    if (input.name == "audio_features" || input.name == cache_config_.enc_in_audio) {
```
kunal-vaishnavi:
```suggestion
    if (input.name == Config::Defaults::AudioFeaturesName || input.name == cache_config_.enc_in_audio) {
```

src/models/nemotron_speech.cpp
```diff
@@ -0,0 +1,364 @@
...
+      mel_tensor, signal_length.get(),
+      encoder_cache_.cache_last_channel.get(),
+      encoder_cache_.cache_last_time.get(),
+      encoder_cache_.cache_last_channel_len.get()};
+  const char* enc_output_names[] = {
+      cache_config_.enc_out_encoded.c_str(), cache_config_.enc_out_length.c_str(),
+      cache_config_.enc_out_cache_channel.c_str(), cache_config_.enc_out_cache_time.c_str(),
+      cache_config_.enc_out_cache_channel_len.c_str()};
+
+  auto enc_outputs = nemotron_model_.session_encoder_->Run(
```
kunal-vaishnavi: Why do we have to call the run method of the inference session directly? We should use the built-in run method for a state.

https://github.com/microsoft/onnxruntime-genai/blob/bc815abe944d3d2ae8babf35e385c968f01337f2/src/models/model.cpp#L113
https://github.com/microsoft/onnxruntime-genai/blob/bc815abe944d3d2ae8babf35e385c968f01337f2/src/models/multi_modal.cpp#L337-L343

If we add the right input and output names to the state, then the state class can manage this run call as well as any run options for us. The `Add()` methods for the class members within a state's constructor do this already.

https://github.com/microsoft/onnxruntime-genai/blob/bc815abe944d3d2ae8babf35e385c968f01337f2/src/models/whisper.cpp#L59-L61

- Input ids: https://github.com/microsoft/onnxruntime-genai/blob/bc815abe944d3d2ae8babf35e385c968f01337f2/src/models/input_ids.cpp#L36
- KV cache: https://github.com/microsoft/onnxruntime-genai/blob/bc815abe944d3d2ae8babf35e385c968f01337f2/src/models/kv_cache.cpp#L237
- Logits: https://github.com/microsoft/onnxruntime-genai/blob/bc815abe944d3d2ae8babf35e385c968f01337f2/src/models/logits.cpp#L114

src/models/nemotron_speech.cpp
```diff
@@ -0,0 +1,364 @@
...
+        cache_config_.dec_in_targets.c_str(), cache_config_.dec_in_target_length.c_str(),
+        cache_config_.dec_in_lstm_hidden.c_str(), cache_config_.dec_in_lstm_cell.c_str()};
+    OrtValue* dec_inputs[] = {
+        targets_.get(), target_length_.get(),
+        decoder_state_.lstm_hidden_state.get(), decoder_state_.lstm_cell_state.get()};
+    const char* dec_output_names[] = {
+        cache_config_.dec_out_outputs.c_str(), cache_config_.dec_out_prednet_lengths.c_str(),
+        cache_config_.dec_out_lstm_hidden.c_str(), cache_config_.dec_out_lstm_cell.c_str()};
+
+    auto dec_outputs = nemotron_model_.session_decoder_->Run(
```
kunal-vaishnavi: Same question here for the decoder and joiner models. We could have separate state classes for them and have a parent state class that does the management and orchestration of all three states in the pipeline execution order that you need. Then, this file can operate similar to the `multi_modal.cpp` file. There already is a `NemotronDecoderState` class above.

src/models/nemotron_streaming_processor.cpp
```diff
@@ -0,0 +1,139 @@
...
+    result->emplace("audio_features", std::make_shared<Tensor>(std::move(mel)));
```
kunal-vaishnavi:
```suggestion
    result->emplace(Config::Defaults::AudioFeaturesName, std::make_shared<Tensor>(std::move(mel)));
```

src/models/nemotron_streaming_processor.cpp
```diff
@@ -0,0 +1,139 @@
...
+    return nullptr;
+  }
+
+  const size_t chunk_size = static_cast<size_t>(cache_config_.chunk_samples);
+  audio_buffer_.resize(chunk_size, 0.0f);  // Pad with silence
+
+  auto mel = BuildMelTensor(audio_buffer_.data(), chunk_size);
+  audio_buffer_.clear();
+  auto result = std::make_unique<NamedTensors>();
+  result->emplace("audio_features", std::make_shared<Tensor>(std::move(mel)));
```
kunal-vaishnavi:
```suggestion
  result->emplace(Config::Defaults::AudioFeaturesName, std::make_shared<Tensor>(std::move(mel)));
```

src/models/nemotron_streaming_processor.cpp
```diff
@@ -0,0 +1,139 @@
...
+      mel_time_major[t * num_mels + m] = mel_data[m * num_frames + t];
+    }
+  }
+
+  // Create output tensor: [1, total_mel_frames, num_mels]
+  auto signal_type = model_.session_info_.GetInputDataType(cache_config_.enc_in_audio);
+
+  // TODO: Optimize for GPU/CUDA later, CPU always expects float32.
+  if (signal_type != ONNX_TENSOR_ELEMENT_DATA_TYPE_FLOAT) {
+    throw std::runtime_error("NemotronStreamingProcessor only supports float32 encoder input. Got type: " + std::to_string(signal_type));
```
kunal-vaishnavi, line 91: Do we have a check for batch size as well? If not, perhaps we should add one and throw a runtime error early on to avoid any issues later on.

src/models/nemotron_streaming_processor.cpp
```diff
@@ -0,0 +1,139 @@
...
+  auto processed_signal = OrtValue::CreateTensor(allocator, signal_shape, signal_type);
+  float* signal_data = processed_signal->GetTensorMutableData<float>();
+
+  // Materialize cache frames from ring buffer (oldest-first starting at cache_pos_)
+  for (int t = 0; t < cache_size; ++t) {
+    int ring_idx = (cache_pos_ + t) % cache_size;
+    const float* src = mel_pre_encode_cache_.data() + ring_idx * num_mels;
+    std::memcpy(signal_data + t * num_mels, src, num_mels * sizeof(float));
+  }
+
```
kunal-vaishnavi, line 108: Can we remove the need for `mel_time_major` and return a transposed tensor from `Process()` directly? I also asked Copilot for its thoughts, and it thinks these `std::memcpy` calls can still be optimized further.

---

Yes—there are a few straightforward optimizations you can make around the `memcpy` usage here. The biggest wins are (1) eliminating the extra “transpose buffer” and (2) reducing the number of small `memcpy`s in the ring-buffer update.

Below are practical options, ordered from “low risk / small refactor” to “bigger change”.

---

## 1) Remove `mel_time_major` entirely (write directly into the output tensor)

Right now you do:

1. Build `mel_time_major` (a full extra buffer) by transposing `[num_mels, num_frames]` → `[num_frames, num_mels]`
2. `memcpy` that buffer into `signal_data` after the cache
3. Copy each frame from `mel_time_major` into the ring buffer

You can instead transpose *directly* into the final destination (`signal_data + cache_size*num_mels`) and then update the ring buffer from that same region.

That removes:
- One full allocation (`std::vector<float> mel_time_major`)
- One big `memcpy` of the whole chunk
- And it improves locality because you write final layout once

Sketch:

```cpp
float* chunk_dst = signal_data + cache_size * num_mels;

// transpose directly into output tensor
for (int t = 0; t < num_frames; ++t) {
  float* dst_row = chunk_dst + t * num_mels;
  for (int m = 0; m < num_mels; ++m) {
    dst_row[m] = mel_data[m * num_frames + t];
  }
}

// later: update ring buffer from chunk_dst (already time-major)
```

Even without SIMD, this typically saves a noticeable amount of memory traffic.

---

## 2) Update the ring buffer with 1–2 large `memcpy`s instead of per-frame copies

This section is currently the hottest “memcpy overhead” area:

```cpp
for (int t = 0; t < num_frames; ++t) {
  float* destination = mel_pre_encode_cache_.data() + cache_pos_ * num_mels;
  std::memcpy(destination, mel_time_major.data() + t * num_mels, num_mels * sizeof(float));
  cache_pos_ = (cache_pos_ + 1) % cache_size;
}
```

You can rewrite it as “copy a contiguous run to the end of the ring, then (if wrapped) copy the remainder to the beginning”. That reduces many small memcpys to at most **two**.

Assuming you have `chunk_src_time_major` pointing at your time-major frames (either `mel_time_major.data()` today, or `chunk_dst` if you implement #1):

```cpp
int write0 = std::min(num_frames, cache_size - cache_pos_);
std::memcpy(mel_pre_encode_cache_.data() + cache_pos_ * num_mels,
            chunk_src_time_major,
            static_cast<size_t>(write0) * num_mels * sizeof(float));

int remaining = num_frames - write0;
if (remaining > 0) {
  std::memcpy(mel_pre_encode_cache_.data(),
              chunk_src_time_major + write0 * num_mels,
              static_cast<size_t>(remaining) * num_mels * sizeof(float));
}

cache_pos_ = (cache_pos_ + num_frames) % cache_size;
```

This is a classic ring-buffer optimization and usually helps a lot.

---

## 3) Cache materialization: already “good”, but can be reduced to 1–2 `memcpy`s too

This part:

```cpp
for (int t = 0; t < cache_size; ++t) {
  int ring_idx = (cache_pos_ + t) % cache_size;
  const float* src = mel_pre_encode_cache_.data() + ring_idx * num_mels;
  std::memcpy(signal_data + t * num_mels, src, num_mels * sizeof(float));
}
```

…copies one frame at a time. You can also do it in at most two copies:

- First copy from `cache_pos_` to end of ring
- Second copy from start of ring to `cache_pos_-1`

```cpp
int tail = cache_size - cache_pos_; // frames from cache_pos_ .. cache_size-1

std::memcpy(signal_data,
            mel_pre_encode_cache_.data() + cache_pos_ * num_mels,
            static_cast<size_t>(tail) * num_mels * sizeof(float));

if (cache_pos_ > 0) {
  std::memcpy(signal_data + tail * num_mels,
              mel_pre_encode_cache_.data(),
              static_cast<size_t>(cache_pos_) * num_mels * sizeof(float));
}
```

This reduces loop overhead and tends to help with large `cache_size`.

---

## 4) Consider storing the cache in the same layout you need most

Right now your cache is effectively time-major in memory already (`ring_idx * num_mels`), which is good for the output tensor. The main pain is that the extractor returns mel-major (`[num_mels, num_frames]`), forcing a transpose.

If you can change `nemo_mel::NemoStreamingMelExtractor::Process` (or add an overload) to output time-major directly, you could eliminate the transpose work entirely (not just the extra buffer). That’s a bigger change but it’s the “real” compute win.

---

## 5) Micro-optimizations (lower impact than the above)

- **Avoid repeated multiplications** inside loops by using pointers (`dst_row`, `src_col`).
- **Use `std::span` or restrict-like hints** only if your toolchain benefits; in portable C++ it’s limited.
- **Threading/SIMD**: transposition can be vectorized, but usually the memory-traffic reductions above beat SIMD complexity.

---

### Recommendation
If you want the best bang-for-buck with minimal risk:

1) Implement **2 (ring update in 1–2 memcpys)**  
2) Implement **3 (cache materialization in 1–2 memcpys)**  
3) Then implement **1 (remove `mel_time_major`)** as a follow-up

---

src/python/python.cpp
```diff
@@ -246,6 +246,10 @@ struct PyGenerator {
     generator_->SetModelInput(name.c_str(), *ToOgaTensor(value, false));
   }
 
+  void SetModelInputTensor(const std::string& name, OgaTensor& tensor) {
+    generator_->SetModelInput(name.c_str(), tensor);
```
kunal-vaishnavi: If we are using `SetInputs` to set any class-defined tensors, why is this needed?

src/generators.cpp
```diff
@@ -504,10 +515,18 @@ void Generator::SetRuntimeOption(const char* key, const char* value) {
 }
 
 size_t Generator::TokenCount() const {
+  if (is_rnnt_) return 0;
```
kunal-vaishnavi: Why do we have to return 0? Why can't the generator still keep track of tokens returned to the user?

src/generators.cpp
```diff
@@ -504,10 +515,18 @@ void Generator::SetRuntimeOption(const char* key, const char* value) {
...
   return static_cast<size_t>(search_->GetSequenceLength());
 }
 
 bool Generator::IsDone() {
+  if (is_rnnt_) {
```
kunal-vaishnavi: Should this `if` block be after the session terminated check?

test/python/test_onnxruntime_genai_api.py
```diff
@@ -961,3 +961,201 @@ def test_audio_preprocessing_multiple_audios(test_data_path, relative_model_path
...
+    # Generate 440Hz sine wave
+    t = np.arange(chunk_samples, dtype=np.float32) / sample_rate
+    audio = (0.5 * np.sin(2.0 * np.pi * frequency * t)).astype(np.float32)
+
+    transcript = ""
+    # Feed multiple chunks
+    for _ in range(4):
+        mel = processor.process(audio)
+        if mel is not None:
+            generator.set_inputs(mel)
```
kunal-vaishnavi: It seems this pattern of `if mel is not None, run loop, else fail test` is quite common across your test functions. Can we refactor this to define a common method shared across your test functions for this pattern?

examples/python/nemotron_streaming.py
nenad1002: I'm doing this in Flush now, leftover, this was needed before I added functionality to Flush to do this automatically, removing it

src/models/nemotron_speech.cpp
nenad1002: To exclude blank token as we're skipping it with >= 1024 condition later, but looking at other models now, they're not doing it, I'll match other tokens for vocab_size to match all tokens, including blanks/bos/eos.

src/models/nemotron_speech.cpp
nenad1002: Ah, this is a regression, I had this previously, thanks for catching this.

src/models/nemotron_speech.cpp
nenad1002: > I already responded to this in my last comment, this is a hack and it doesn't fit here. I am not sure how would that look like as API does not fit. And code looks more messy. What is the benefit of this? Observability? 

I'll go with it for now to avoid delays, we might refactor it later.

src/generators.cpp
nenad1002: We can

src/models/nemotron_streaming_processor.cpp
nenad1002: Batch size is always 1, do you think model could accidentally be more than 1 on batch? We dont need a check for this IMO

examples/csharp/StreamingASR/Program.cs
nenad1002: Supports more now

src/models/nemotron_speech.cpp
```diff
@@ -0,0 +1,534 @@
...
+    symbol_step_ = 0;
+  }
+
+  last_tokens_.clear();
+
+  auto enc_shape = encoded_output_->GetTensorTypeAndShapeInfo()->GetShape();
+  int64_t time_steps = std::min(enc_shape[1], encoded_len_);
+  int64_t hidden_dim = enc_shape[2];
+  size_t frame_bytes = static_cast<size_t>(hidden_dim) * sizeof(float);
+
```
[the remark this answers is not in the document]
nenad1002: Not relevant for now

src/models/nemotron_streaming_processor.cpp
```diff
@@ -0,0 +1,146 @@
...
+  audio_buffer_.insert(audio_buffer_.end(), audio_data, audio_data + num_samples);
+
+  const size_t chunk_size = static_cast<size_t>(cache_config_.chunk_samples);
+
+  // Process the first complete chunk available
+  if (audio_buffer_.size() >= chunk_size) {
+    auto mel = BuildMelTensor(audio_buffer_.data(), chunk_size);
+    audio_buffer_.erase(audio_buffer_.begin(),
+                        audio_buffer_.begin() + static_cast<ptrdiff_t>(chunk_size));
+    auto result = std::make_unique<NamedTensors>();
```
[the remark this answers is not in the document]
nenad1002: Went fixing this, no RTFx gain with more code complexity. leaving this

src/models/nemotron_speech.cpp
```diff
@@ -0,0 +1,534 @@
...
+  // cache_last_channel: [batch, num_layers, left_context, hidden_dim]
+  auto ch_shape = std::array<int64_t, 4>{1, cfg.num_encoder_layers, cfg.left_context, cfg.hidden_dim};
+  cache_last_channel = OrtValue::CreateTensor(allocator, ch_shape, cache_channel_type);
+  ByteWrapTensor(device, *cache_last_channel).Zero();
+
+  // cache_last_time: [batch, num_layers, hidden_dim, conv_context]
+  auto tm_shape = std::array<int64_t, 4>{1, cfg.num_encoder_layers, cfg.hidden_dim, cfg.conv_context};
+  cache_last_time = OrtValue::CreateTensor(allocator, tm_shape, cache_time_type);
+  ByteWrapTensor(device, *cache_last_time).Zero();
```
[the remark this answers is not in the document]
nenad1002: This is not a bug

src/models/nemotron_streaming_processor.cpp
nenad1002: Implemented 1) _(Implement 2 (ring update in 1–2 memcpys_)  and did in-place transpose

test/c_api_tests.cpp
```diff
@@ -1362,3 +1364,162 @@ TEST(CAPITests, SetGuidance) {
...
+  auto processor = OgaStreamingProcessor::Create(*model);
+  auto params = OgaGeneratorParams::Create(*model);
+  auto generator = OgaGenerator::Create(*model, *params);
+
+  constexpr size_t chunk_samples = 8960;
+  std::vector<float> silence(chunk_samples, 0.0f);
+
+  auto mel = processor->Process(silence.data(), silence.size());
+  DecodeInputs(*generator, mel.get());
+  SUCCEED();
```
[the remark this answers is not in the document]
nenad1002: added asserts

src/python/python.cpp
```diff
@@ -631,6 +632,34 @@ PYBIND11_MODULE(onnxruntime_genai, m) {
...
+           "The model must be of type 'nemotron_speech'.")
+      .def(
+          "process",
+          [](OgaStreamingProcessor& proc, pybind11::array_t<float> audio_chunk) -> pybind11::object {
+            auto buf = audio_chunk.request();
+            auto result = proc.Process(static_cast<const float*>(buf.ptr), static_cast<size_t>(buf.size));
+            if (result) {
+              return pybind11::cast(std::move(result));
+            }
+            return pybind11::none();
```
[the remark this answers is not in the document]
nenad1002: fixed

examples/python/nemotron_speech.py
```diff
@@ -0,0 +1,105 @@
...
+                text += token_text
+    return text
+
+
+def simulate_microphone(model_path, audio_path, execution_provider):
+    """Stream audio through Generator + StreamingProcessor API."""
+    sample_rate, chunk_samples = load_config(model_path)
+    audio = load_audio(audio_path, sample_rate)
+    duration = len(audio) / sample_rate
+    chunk_duration = chunk_samples / sample_rate
```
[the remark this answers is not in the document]
nenad1002: Fixed

rui-ren (dismissed):

src/ort_genai_c.h
```diff
@@ -1146,6 +1147,46 @@ OGA_EXPORT void OGA_API_CALL OgaRegisterExecutionProviderLibrary(const char* reg
...
+ * \return OgaResult on error, nullptr on success.
+ */
+OGA_EXPORT OgaResult* OGA_API_CALL OgaStreamingProcessorFlush(OgaStreamingProcessor* processor, OgaNamedTensors** out);
+
+/**
+ * \brief Reset audio processor state for a new utterance.
+ * \param[in] processor The StreamingProcessor instance.
+ * \return OgaResult on error, nullptr on success.
+ */
+OGA_EXPORT OgaResult* OGA_API_CALL OgaStreamingProcessorReset(OgaStreamingProcessor* processor);
```
baijumeswani: Can application layer re-create the StreamingProcessor instead of resetting it? Is there any advantage in resetting vs scoping it for use with 1 full audio sample at a time?

nuget.config
```diff
@@ -3,5 +3,6 @@
   <packageSources>
     <clear />
     <add key="ORT-Nightly" value="https://aiinfra.pkgs.visualstudio.com/PublicPackages/_packaging/ORT-Nightly/nuget/v3/index.json" />
+    <add key="nuget.org" value="https://api.nuget.org/v3/index.json" />
```
baijumeswani: This will likely fail packaging since the pipeline only allows 1 nuget source. Can we undo this change?

CMakeLists.txt
```diff
@@ -134,6 +134,15 @@ if(ENABLE_TESTS)
     add_compile_definitions(TEST_PHI2=0)
   endif()
 
+  if (TEST_STREAMING_ASR)
+    add_compile_definitions(TEST_STREAMING_ASR=1)
```
baijumeswani: Instead of defining this, can we check at runtime if the path exists and skip the test if it does not? Similar to how we do in python.

src/generators.cpp
```diff
@@ -504,11 +515,20 @@ void Generator::SetRuntimeOption(const char* key, const char* value) {
...
 bool Generator::IsDone() {
   ThrowErrorIfSessionTerminated(state_->session_terminated_);
+
+  if (is_rnnt_) {
+    // Pending mel input means we haven't started processing this chunk yet
+    if (!extra_inputs_.empty()) return false;
+    auto* speech_state = static_cast<NemotronSpeechState*>(state_.get());
+    return speech_state->IsChunkDone();
+  }
+
```
baijumeswani, line 530: Can use dynamic_pointer_cast here. You could also move the extra_inputs_ check to be inside IsChunkDone()?

```suggestion
  if (auto nemotron_speech_state = std::dynamic_pointer_cast<NemotronSpeechState>(state_)) {
    // Pending mel input means we haven't started processing this chunk yet
    if (!extra_inputs_.empty()) return false;
    return nemotron_speech_state->IsChunkDone();
  }
```

src/generators.cpp
```diff
@@ -504,11 +515,20 @@ void Generator::SetRuntimeOption(const char* key, const char* value) {
 }
 
 size_t Generator::TokenCount() const {
+  if (is_rnnt_) return rnnt_token_count_;
```
baijumeswani: Can this token count be managed inside the NemotronSpeechState?

src/config.cpp
```diff
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.
 // Modifications Copyright(C) 2024-2025 Advanced Micro Devices, Inc. All rights reserved.
 #include "generators.h"
+#include "models/model_type.h"
```
baijumeswani, line 5: Is this include required in this file?

nenad1002: It is, line 1526

src/ort_genai_c.h
nenad1002: Just a convenience method

nenad1002:
> > @kunal-vaishnavi can we sign this off?
> 
> There are a few issues with the ONNX models that may change the input/output structure.
> 
> ### Decoder
> image
> From inspecting the graph, `target_length_orig` and `target_length` are not used at all. This is an identity operation to go from `target_length_orig` to `target_length`. The input and the output can be removed from the graph.
> 
> ### Joiner
> image image
> This is the end of the encoder and the decoder. The decoder contains a `Squeeze` op with `axes = [1]` before the final `Transpose` op. The `Transpose` op has `perm = [1, 2, 0]` so this means that its input (i.e. the output from the `Squeeze` op) must be 3D then. Because the axes are specified on the `Squeeze` op, the operation is only applied on those axes. This implies that the input to the `Squeeze` op (i.e. the output from the `LSTM` op) is 4D. Therefore, the `Squeeze` op is removing a dimension of size 1 to go from `[A, 1, B, C]` to `[A, B, C]`.
> 
> image
> Here is the joiner model. The joiner is receiving 3D inputs from both the encoder and decoder and returns 4D outputs. The decoder input is 3D and the `Unsqueeze` op in the decoder path has `axes = [1]`. This means that the `Unsqueeze` op is going from `[A, B, C]` to `[A, 1, B, C]`.
> 
> That shape of `[A, 1, B, C]` is already obtained in the decoder. If the `Squeeze` op is removed and the `Transpose` op is changed to handle a 4D output, then the output of the decoder can be 4D. The decoder input to the joiner model can be 4D and then the `Unsqueeze` op in the decoder path can be removed. These changes can eliminate two ops across two ONNX models.
> 
> If we want to keep the input dimensions for the joiner model as the same, then we can move the `Unsqueeze` op in the encoder path to the output of the encoder. Then, the joiner model will take in two 4D inputs and produce one 4D output.

Right, this is not related to this PR. I assume this PR will deal with ONNX model finalization and if there is any model signature change:
https://github.com/microsoft/onnxruntime-genai/pull/2010

I can wait for the team's changes though if you prefer.

src/ort_genai_c.h
baijumeswani: I’m fine with this approach, but I have a minor concern regarding the lifecycle of these components. While `StreamingProcessor` has a Reset method, the `Generator` doesn't support `RewindTo(0)`. This means we’re still forced to recreate the `Generator` every time. Since that likely carries more overhead than recreating the processor itself, I’m not sure how much value the Reset functionality actually adds here.
I’ll leave it to your discretion whether we should add this to the public API or not.

kunal-vaishnavi (approved):

PR #2001, opened 2026-03-04
Add support for CUDA 13

baijumeswani:
This pull request adds support for CUDA 13.0 across the build and packaging pipelines, replacing the previous CUDA 11.8 support.

### DIFF
--- a/.pipelines/stages/py-packaging-stage.yml
+++ b/.pipelines/stages/py-packaging-stage.yml
@@ -151,15 +151,6 @@ stages:
         build_config: ${{ parameters.build_config }}
         ort_version: ${{ parameters.ort_version }}
 
-  - ${{ if eq(parameters.enable_macos_cpu, true) }}:
-    - template: jobs/py-packaging-job.yml
-      parameters:
-        arch: 'x64'
-        ep: 'cpu'
-        os: 'osx'
-        build_config: ${{ parameters.build_config }}
-        ort_version: ${{ parameters.ort_version }}
-
     - template: jobs/py-packaging-job.yml
       parameters:
         arch: 'arm64'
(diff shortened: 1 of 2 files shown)


.pipelines/stages/jobs/steps/capi-linux-step.yml
```diff
@@ -71,7 +71,12 @@ steps:
     set -e -x
     cmake_extra_args=""
     if [ "$(ep)" = "cuda" ]; then
-      cmake_extra_args="-DCMAKE_CUDA_COMPILER=/usr/local/cuda/bin/nvcc"
+      if [ "$(cuda_version)" = "13.0" ]; then
```
kunal-vaishnavi: Can we make the CUDA version checks as 13.x? Both 13.0 and 13.1 have been released.

CMakeLists.txt
```diff
@@ -25,11 +26,6 @@ if(CMAKE_VERSION VERSION_GREATER_EQUAL "3.30.0")
 endif()
 
 if(MSVC)
-  # DLL initialization errors due to old conda msvcp140.dll dll are a result of the new MSVC compiler
-  # See https://developercommunity.visualstudio.com/t/Access-violation-with-std::mutex::lock-a/10664660#T-N10668856
-  # Remove this definition once the conda msvcp140.dll dll is updated.
```
kunal-vaishnavi, line 30: Has the DLL been updated?

baijumeswani: This should be resolved. If I remember correctly, this problem was specific to the anaconda env. And I am assuming that the msvcp.dll is now updated in conda distributions.
IF we see crashes, we can re-add this.

PR #2002, opened 2026-03-04
Reuse a single generator via RewindTo(0) in benchmark instead of creating multiple generators

qjia7:
Using multiple generators is not friendly for graph capture mode:

1. Each generator maps to its own captured graph. When the generator's lifecycle ends, there is no way to notify ORT to release the captured graph, so GPU memory is not freed until the ORT model session is released.
2. It also fails if the underlying EP does not support multiple captured graphs.

This PR changes the benchmark to create a single generator and reuse it across prompt generation, warmup, and benchmark iterations by calling RewindTo(0) before each iteration. This ensures the KV cache is consistently sized for the full workload and avoids accumulating stale captured graphs.

### DIFF
--- a/benchmark/python/benchmark_e2e.py
+++ b/benchmark/python/benchmark_e2e.py
@@ -316,7 +316,10 @@ def run_benchmark(args, batch_size, prompt_length, generation_length, max_length
         text_seed = "a"
         seed_prompt = f"{args.chat_template.format(input=text_seed)}"
         seed_tokens = tokenizer.encode(seed_prompt)
-        gen.append_tokens(seed_tokens)
+        # Tile seed tokens for batch_size > 1 so append_tokens gets
+        # batch_size * seq_len tokens in the expected layout.
+        batched_seed_tokens = np.tile(seed_tokens, batch_size)
+        gen.append_tokens(batched_seed_tokens)
         while not gen.is_done() and gen.token_count() < prompt_length:
             gen.generate_next_token()
         generated_text = tokenizer.decode(gen.get_sequence(0))


benchmark/c/main.cpp
```diff
@@ -166,31 +147,61 @@ void RunBenchmark(const benchmark::Options& opts) {
...
+  std::string prompt;
+
+  if (const size_t* npt = std::get_if<size_t>(&opts.prompt_num_tokens_or_content)) {
+    num_prompt_tokens = *npt;
+    need_generate_prompt = true;
+  } else {
+    prompt = std::get<std::string>(opts.prompt_num_tokens_or_content);
+    auto temp_sequences = OgaSequences::Create();
+    tokenizer->Encode(prompt.c_str(), *temp_sequences);
+    num_prompt_tokens = temp_sequences->SequenceCount(0);
```
kunal-vaishnavi, line 168: Can we use the new `Generator.TokenCount` API to keep track of the number of tokens after appending the tokens to the generator and for any subsequent token counting after the generation loop completes?

benchmark/python/benchmark_e2e.py
```diff
@@ -320,19 +300,39 @@ def run_benchmark(args, batch_size, prompt_length, generation_length, max_length
         batch_size=batch_size,
     )
 
+    # Create a single generator and reuse it for prompt generation, warmup, and
+    # benchmark iterations via rewind_to(0).  This guarantees the KV cache is
+    # always sized for the full workload.
```
kunal-vaishnavi: Does this comment assume that buffer sharing is enabled?

qjia7: Yes. But it works no matter past_present_share_buffer is true or false.

benchmark/c/main.cpp
qjia7: Done

kunal-vaishnavi (approved):

benchmark/c/main.cpp
```diff
@@ -166,41 +147,70 @@ void RunBenchmark(const benchmark::Options& opts) {
...
   }
 
-  const size_t num_prompt_tokens = prompt_sequences->SequenceCount(0);
   const size_t num_tokens = num_prompt_tokens + opts.num_tokens_to_generate;
   const auto generator_params = MakeGeneratorParams(opts, *model, num_tokens);
 
+  // Create a single generator and reuse it for prompt generation, warmup, and
+  // benchmark iterations via RewindTo(0).  This guarantees the KV cache is
+  // always sized for the full workload.
```
baijumeswani: Not all models support RewindTo functionality. Historically, we've required applications to create new Generators for every prompt.

Since ORT might eventually get a native way to clear captured graph memory, do you think this implementation is just a temporary fix? I'm curious if we'll end up reverting these changes once that's available.

Could we parameterize this in the benchmark? Something like a switch to choose between a fresh generator vs. RewindTo would be a nice way for testing both paths.

Also, `RewindTo` removes KV cache allocation from the loop, which changes the benchmark profile significantly. So, the results might look a lot better than they actually are compared to creating a new Generator.

benchmark/python/benchmark_e2e.py
```diff
@@ -543,6 +553,14 @@ def str2strlist(value):
...
     parser.add_argument("-v", "--verbose", action="store_true", help="Print extra information")
     parser.add_argument("-mo", "--print_model_output", action="store_true", help="Print model output")
+    parser.add_argument(
+        "--reuse_generator", action="store_true", default=True,
+        help="Reuse a single generator via rewind_to(0) instead of creating a new one per iteration (default)"
+    )
+    parser.add_argument(
+        "--no_reuse_generator", dest="reuse_generator", action="store_false",
+        help="Create a new generator for each iteration instead of reusing one"
+    )
```
baijumeswani, line 574: Can the default be false and can we just have a single argument?

benchmark/c/options.cpp
```diff
@@ -135,6 +138,10 @@ Options ParseOptionsFromCommandLine(int argc, const char* const* argv) {
         opts.num_warmup_iterations = ParseNumber<size_t>(next_arg(i));
       } else if (arg == "-ml" || arg == "--max_length") {
         opts.max_length = ParseNumber<int64_t>(next_arg(i));
+      } else if (arg == "--reuse_generator") {
+        opts.reuse_generator = true;
+      } else if (arg == "--no_reuse_generator") {
+        opts.reuse_generator = false;
```
baijumeswani: Same for C++. Can the default be false and can we have a single argument?

benchmark/c/main.cpp
```diff
@@ -166,42 +147,88 @@ void RunBenchmark(const benchmark::Options& opts) {
...
+    while (!gen->IsDone() && gen->TokenCount() < target_token_count) {
+      gen->GenerateNextToken();
     }
 
     if (opts.verbose && i == 0) {
       // show prompt and output on first iteration
       std::cout << "[PROMPT BEGIN]" << prompt << "[PROMPT END]\n";
-      const auto output_sequence_length = generator->GetSequenceCount(0);
-      const auto* output_sequence_data = generator->GetSequenceData(0);
+      const auto output_sequence_length = gen->TokenCount();
```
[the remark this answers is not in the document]
qjia7: @kunal-vaishnavi We don't support batch > 1 case for this benchmark. So it's safe to use gen->TokenCount(), right?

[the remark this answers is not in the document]
baijumeswani: Yes, it is safe to use TokenCount()

benchmark/python/benchmark_e2e.py
```diff
@@ -302,37 +283,67 @@ def run_benchmark(args, batch_size, prompt_length, generation_length, max_length
...
+        seed_prompt = f"{args.chat_template.format(input=text_seed)}"
+        seed_tokens = tokenizer.encode(seed_prompt)
+        gen.append_tokens(seed_tokens)
+        while not gen.is_done() and gen.token_count() < prompt_length:
+            gen.generate_next_token()
+        generated_text = tokenizer.decode(gen.get_sequence(0))
+        if not args.reuse_generator:
+            del gen
+        text = [generated_text] * batch_size
+        prompt = f"{args.chat_template.format(input=text)}"
```
kunal-vaishnavi, line 329: Can we use `Tokenizer.ApplyChatTemplate` instead of trying to apply a user-provided format? I understand that the purpose of using a custom chat template is to not have extra tokens appended in the prompt. However, it can be hard to get a true performance evaluation without using the chat template that users would typically use during inference.

benchmark/c/options.cpp
```diff
@@ -50,6 +50,9 @@ namespace {
     << "    -ml,--max_length <number>\n"
     << "      Max sequence length (prompt + output). Overrides genai_config.json.\n"
     << "      Default: prompt_length + generation_length. Pass -1 to use config file value.\n"
+    << "    --reuse_generator\n"
```
kunal-vaishnavi, line 53: We should upgrade the options parsing to a modern library for more flexibility, stability, and maintainability. In the [E2E examples](https://github.com/microsoft/onnxruntime-genai/blob/6e200ea0515de7efa4c60f66cb3069f0723f5a42/examples/c/src/common.cpp#L179), I used the [CLI11 library](https://github.com/CLIUtils/CLI11) to achieve this.

benchmark/python/benchmark_e2e.py
qjia7: @kunal-vaishnavi Do you mind if I use a separate PR to address your new comments, including adopting Tokenizer.ApplyChatTemplate and the CLI11 library for option parsing?

kunal-vaishnavi: Sure, that's fine

benchmark/c/main.cpp
```diff
@@ -166,42 +147,88 @@ void RunBenchmark(const benchmark::Options& opts) {
...
+
+    const char* const base_prompt = "A";
+    auto base_prompt_sequences = OgaSequences::Create();
+    for (size_t i = 0; i < opts.batch_size; ++i) {
+      tokenizer->Encode(base_prompt, *base_prompt_sequences);
+    }
+    gen->AppendTokenSequences(*base_prompt_sequences);
+    while (!gen->IsDone() && gen->TokenCount() < num_prompt_tokens) {
+      gen->GenerateNextToken();
+    }
```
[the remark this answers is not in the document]
qjia7: The current behavior is intentional. --prompt_length is meant to specify the total number of tokens in the generated prompt (as the help text says: "Number of tokens in the generated prompt"), not the number of additional tokens to generate on top of the seed. The seed "A" (1 token) is just a bootstrap to start autoregressive generation — the loop runs until the total sequence length reaches [num_prompt_tokens](vscode-file://vscode-app/c:/Users/jiajiaqin/AppData/Local/Programs/Microsoft%20VS%20Code/072586267e/resources/app/out/vs/code/electron-browser/workbench/workbench.html), producing a prompt of exactly the requested length. Since [TokenCount()](vscode-file://vscode-app/c:/Users/jiajiaqin/AppData/Local/Programs/Microsoft%20VS%20Code/072586267e/resources/app/out/vs/code/electron-browser/workbench/workbench.html) returns the total sequence length (prompt + generated), the comparison [gen->TokenCount() < num_prompt_tokens](vscode-file://vscode-app/c:/Users/jiajiaqin/AppData/Local/Programs/Microsoft%20VS%20Code/072586267e/resources/app/out/vs/code/electron-browser/workbench/workbench.html) correctly stops when we have a prompt with [num_prompt_tokens](vscode-file://vscode-app/c:/Users/jiajiaqin/AppData/Local/Programs/Microsoft%20VS%20Code/072586267e/resources/app/out/vs/code/electron-browser/workbench/workbench.html) total tokens, which is what we want.

The edge case where the seed tokenizes to >= [num_prompt_tokens](vscode-file://vscode-app/c:/Users/jiajiaqin/AppData/Local/Programs/Microsoft%20VS%20Code/072586267e/resources/app/out/vs/code/electron-browser/workbench/workbench.html) is valid in theory but not a practical concern — "A" is 1 token, and the default prompt_length is 16.

benchmark/python/benchmark_e2e.py
qjia7: Thanks. Will do it in separate PR.

benchmark/c/options.cpp
qjia7: Will do it in separate PR.

PR #2012, opened 2026-03-06
Add support for QNN stateful models

qti-ashimaj:
This PR enhances support for QNN stateful models that execute entirely on Qualcomm's Neural Processing Unit (NPU), including attention operations and internal KV cache management. This significant improvement eliminates previous limitations where certain operations would fall back to CPU execution, resulting in performance bottlenecks.

The implementation identifies QNN stateful models through the `genie_model=True` provider option. For these models:
- The ONNX Runtime GenAI framework delegates KV cache management entirely to the QNN backend
- The model architecture is optimized to return logits only for the last token in the sequence

### DIFF
--- a/src/qnn/interface.cpp
+++ b/src/qnn/interface.cpp
@@ -80,29 +80,21 @@ DeviceInterface* GetQNNInterface() {
 }
 
 bool IsQNNStatefulModel(const Model& model) {
-  // Check for both QNN and CPU device types
-  // When using QNN EP with genai_model=True, the model is stateful regardless of device type (QNN/CPU)
-  // For QNN models with enable_htp_shared_memory_allocator=1, p_device_ will be QNN type
-  // For QNN models without shared memory allocator, p_device_ will be CPU type
-  // Both cases need to be handled the same way for stateful models where KV cache is managed internally
-  if (model.p_device_->GetType() == DeviceType::QNN || model.p_device_->GetType() == DeviceType::CPU) {
-    const auto& provider_options = model.config_->model.decoder.session_options.provider_options;
-    for (const auto& po : provider_options) {
-      if (po.name == "QNN") {
-        for (const auto& option : po.options) {
-          // For QNN, if session option 'genie_model' is set to true, the session will encapsulate
-          // a stateful model, so KVCache will be managed internally.
-          if (option.first == "genie_model") {
-            std::string lower_value(option.second);
-            std::transform(lower_value.begin(), lower_value.end(), lower_value.begin(),
-                           [](unsigned char c) { return static_cast<unsigned char>(std::tolower(c)); });
-            return lower_value == "true";
-          }
+  const auto& provider_options = model.config_->model.decoder.session_options.provider_options;
+  for (const auto& po : provider_options) {
+    if (po.name == "QNN") {
+      for (const auto& option : po.options) {
+        // For QNN, if session option 'genie_model' is set to true, the session will encapsulate
+        // a stateful model, so KVCache will be managed internally.
+        if (option.first == "genie_model") {
+          std::string lower_value(option.second);
+          std::transform(lower_value.begin(), lower_value.end(), lower_value.begin(),
+                         [](unsigned char c) { return static_cast<unsigned char>(std::tolower(c)); });
+          return lower_value == "true";
         }
       }
     }
   }
-
   return false;
 }
 


src/models/logits.cpp
```diff
@@ -17,8 +18,8 @@ Logits::Logits(State& state)
 
   input_sequence_lengths.resize(state_.params_->search.batch_size);
 
-  if (IsOpenVINOStatefulModel(state.model_) || state.model_.IsPruned()) {
-    // In the case of OpenVINO stateful models, or any model whose ONNX graph
+  if (IsOpenVINOStatefulModel(state.model_) || IsQNNStatefulModel(state.model_) || state.model_.IsPruned()) {
```
baijumeswani: Is the check for `state.model_.IsPruned()` not enough?

src/qnn/interface.cpp
```diff
@@ -78,4 +79,24 @@ DeviceInterface* GetQNNInterface() {
   return g_device.get();
 }
 
+bool IsQNNStatefulModel(const Model& model) {
+  if (model.p_device_->GetType() == DeviceType::QNN || model.p_device_->GetType() == DeviceType::CPU) {
```
baijumeswani: Why would p_device be CPU for QNN models that are stateful?

src/qnn/interface.cpp
```diff
@@ -78,4 +79,24 @@ DeviceInterface* GetQNNInterface() {
...
+    const auto& provider_options = model.config_->model.decoder.session_options.provider_options;
+    for (auto& po : provider_options) {
```
baijumeswani:
```suggestion
    for (const auto& po : provider_options) {
```

src/qnn/interface.cpp
```diff
@@ -78,4 +79,24 @@ DeviceInterface* GetQNNInterface() {
...
+      if (po.name == "QNN") {
```
baijumeswani: Does this need to handle other names? Like QNNExecutionProvider or qnn? Is this the standard naming for the qnn provider?

src/qnn/interface.cpp
```diff
@@ -78,4 +79,24 @@ DeviceInterface* GetQNNInterface() {
...
+        const auto& qnn_options = po.options;
+        for (auto& option : qnn_options) {
```
baijumeswani:
```suggestion
        for (const auto& option : po.options) {
```

src/qnn/interface.cpp
baijumeswani: Is it because npu memory is host accessible and that the memory allocations still happen on cpu?

baijumeswani: If so, please add a comment.

src/models/logits.cpp
qti-ashimaj: Yes, we can use that API, removed the extra check

src/qnn/interface.cpp
qti-ashimaj: Device type is [QNN ](https://github.com/microsoft/onnxruntime-genai/blob/main/src/models/model.cpp#L751)only when we use the provider option `enable_htp_shared_memory_allocator=1`, Since we can run the stateful model regardless of the shared memory allocator, hence the checks for both device types.

src/qnn/interface.cpp
qti-ashimaj: Done

src/qnn/interface.cpp
qti-ashimaj: The normalized provider name is [QNN](https://github.com/microsoft/onnxruntime-genai/blob/main/src/config.cpp#L22) so we do not need to handle other strings.

src/qnn/interface.cpp
```diff
@@ -78,4 +79,28 @@ DeviceInterface* GetQNNInterface() {
...
+        for (const auto& option : po.options) {
+          // For QNN, if session option 'genai_model' is set, the session will encapsulate
+          // a stateful model, so KVCache will be managed internally.
+          if (option.first == "genai_model" && option.second == "True") {
+            return true;
+          }
+        }
+      }
+    }
+  }
```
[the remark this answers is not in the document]
qti-ashimaj: This check is only relevant for QNN models, hence it should rightly return false for non-CPU/non-QNN device type. The check is valid.

src/qnn/interface.cpp
```diff
@@ -78,4 +79,28 @@ DeviceInterface* GetQNNInterface() {
...
+  // For QNN models without shared memory allocator, p_device_ will be CPU type
+  // Both cases need to be handled the same way for stateful models where KV cache is managed internally
+  if (model.p_device_->GetType() == DeviceType::QNN || model.p_device_->GetType() == DeviceType::CPU) {
+    const auto& provider_options = model.config_->model.decoder.session_options.provider_options;
+    for (const auto& po : provider_options) {
+      if (po.name == "QNN") {
+        for (const auto& option : po.options) {
+          // For QNN, if session option 'genai_model' is set, the session will encapsulate
+          // a stateful model, so KVCache will be managed internally.
+          if (option.first == "genai_model" && option.second == "True") {
```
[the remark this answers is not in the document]
qti-ashimaj: Changed it to first convert the value to lower case and then compare with `\"true\"`

src/qnn/interface.h
```diff
@@ -5,4 +5,7 @@ namespace Generators {
 
 DeviceInterface* GetQNNInterface();
 
+struct Model;
```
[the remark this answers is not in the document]
qti-ashimaj: It is declared as [struct Model](https://github.com/microsoft/onnxruntime-genai/blob/main/src/models/model.h#L145) in `model.h`

src/qnn/interface.cpp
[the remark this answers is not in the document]
baijumeswani:
```cpp
for (const auto& po : provider_options) {
      if (po.name == "QNN") {
        for (const auto& option : po.options) {
          // For QNN, if session option 'genai_model' is set, the session will encapsulate
          // a stateful model, so KVCache will be managed internally.
          if (option.first == "genai_model" && option.second == "True") {
            return true;
          }
        }
      }
    }
```

Already checks for the provider option. Do we need a check at the top as well? Because that check will still fail if the p_device = CPU (which is the case for multiple providers (such as QNN, VitisAI...).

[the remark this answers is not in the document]
qti-ashimaj: Okay, I am removing the checks for device type. It will check for only QNN as provider followed by "genie_model"="true" as provider option. This will run regardless of device types.

qti-ashimaj:
Hi @baijumeswani, could you please help in reviewing and getting the PR merged.
Thanks!

PR #2015, opened 2026-03-10
[WebGPU] Add compact attention mask support and fix prefill regression for graph capture

qjia7:
This PR includes below changes:

**1. Fix the prefill regression when graph capture is ON**

**Problem**: WebGPU Prefill Performance Regression with Graph Capture
When WebGPU graph capture is enabled, the upstream code sets p_device_inputs_ to GPU (WebGPU device) for model inputs. However, GeneratorParams::p_device was initialized from p_device_inputs_, which meant it also pointed to the GPU device. This caused a problem because:

WebGPU uses GreedySearch_Cpu — the search/scoring logic runs entirely on CPU. But GeneratorParams::p_device (GPU) was used to allocate the sequences buffer and token arrays.

AllocateInputIdsOnDevice allocated the initial input IDs tensor on the GPU device (via p_device), then immediately needed the data on CPU for the CPU-based search, causing unnecessary GPU↔CPU round-trips.

GreedySearch_Cpu::AppendTokens similarly suffered from device mismatch — tokens were allocated on GPU but consumed on CPU, creating synchronization overhead on every token append during both prefill and generation.

The net effect was that prefill throughput was significantly degraded (roughly ~5,000 tps vs ~5,900 tps expected) due to these unnecessary cross-device data transfers happening in the hot path.

**Solution**: Introduce p_device_scoring_
We added a dedicated p_device_scoring_ member to the Model class that represents the correct device for search/scoring operations:

For WebGPU: p_device_scoring_ = CPU (since GreedySearch_Cpu is used)
For CUDA: p_device_scoring_ = same as p_device_inputs_ (since CUDA has its own GPU-based greedy search)
GeneratorParams::p_device now initializes from model.p_device_scoring_ instead of model.p_device_inputs_. This ensures input ID allocation, sequence buffers, and token operations all stay on CPU for WebGPU — eliminating the GPU↔CPU round-trips entirely.

**Result**: Prefill throughput improved from ~5,000 tps back to ~5,900 tps at 1024 prompt length.

**2.  **Feature**: Compact Attention Mask**

Added compact_attention_mask support (config, position_inputs, CPU/WebGPU device interfaces, Python model builder) to enable a more memory-efficient attention mask representation for models that support it.

**3. **Instrumentation**: Tracing**
Added [DurationTrace](vscode-file://vscode-app/c:/Users/jiajiaqin/AppData/Local/Programs/Microsoft%20VS%20Code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) instrumentation to key functions (AllocateInputIdsOnDevice, ComputeLogits, DecoderOnly_State::Run, UpdateInputsOutputs, GreedySearch::AppendTokens) for performance profiling

### DIFF
--- a/benchmark/c/main.cpp
+++ b/benchmark/c/main.cpp
@@ -124,6 +124,25 @@ static std::unique_ptr<OgaGeneratorParams> MakeGeneratorParams(const benchmark::
   return params;
 }
 
+std::string GeneratePrompt(const benchmark::Options& opts, size_t num_prompt_tokens, const OgaModel& model, const OgaTokenizer& tokenizer, size_t batch_size) {
+  const char* const base_prompt = "A";
+  auto base_prompt_sequences = OgaSequences::Create();
+  for (size_t i = 0; i < batch_size; ++i) {
+    tokenizer.Encode(base_prompt, *base_prompt_sequences);
+  }
+
+  auto params = MakeGeneratorParams(opts, model, num_prompt_tokens);
+  auto generator = OgaGenerator::Create(model, *params);
+  generator->AppendTokenSequences(*base_prompt_sequences);
+  while (!generator->IsDone() && num_prompt_tokens-- > 0) {
+    generator->GenerateNextToken();
+  }
+
+  const auto output_sequence_length = generator->GetSequenceCount(0);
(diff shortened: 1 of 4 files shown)


qjia7:
Have split this one into #2022 and #2021. Close it. Thanks.

PR #2021, opened 2026-03-13
[WebGPU] Fix the prefill regression when graph capture is ON

qjia7:
**Problem: WebGPU Prefill Performance Regression with Graph Capture**
When WebGPU graph capture is enabled, the upstream code sets p_device_inputs_ to GPU (WebGPU device) for model inputs. However, GeneratorParams::p_device was initialized from p_device_inputs_, which meant it also pointed to the GPU device. This caused a problem because:

WebGPU uses GreedySearch_Cpu — the search/scoring logic runs entirely on CPU. But GeneratorParams::p_device (GPU) was used to allocate the sequences buffer and token arrays.

AllocateInputIdsOnDevice allocated the initial input IDs tensor on the GPU device (via p_device), then immediately needed the data on CPU for the CPU-based search, causing unnecessary GPU↔CPU round-trips.

GreedySearch_Cpu::AppendTokens similarly suffered from device mismatch — tokens were allocated on GPU but consumed on CPU, creating synchronization overhead on every token append during both prefill and generation.

The net effect was that prefill throughput was significantly degraded (roughly ~5,000 tps vs ~5,900 tps expected) due to these unnecessary cross-device data transfers happening in the hot path.

**Solution: Introduce p_device_scoring_**
We added a dedicated p_device_scoring_ member to the Model class that represents the correct device for search/scoring operations:

For WebGPU: p_device_scoring_ = CPU (since GreedySearch_Cpu is used)
For CUDA: p_device_scoring_ = same as p_device_inputs_ (since CUDA has its own GPU-based greedy search)
GeneratorParams::p_device now initializes from model.p_device_scoring_ instead of model.p_device_inputs_. This ensures input ID allocation, sequence buffers, and token operations all stay on CPU for WebGPU — eliminating the GPU↔CPU round-trips entirely.

**Result**
Prefill throughput improved from ~5,000 tps back to ~5,900 tps at 1024 prompt length.

### DIFF
--- a/src/models/model.h
+++ b/src/models/model.h
@@ -166,7 +166,7 @@ struct Model : std::enable_shared_from_this<Model>, LeakChecked<Model>, External
 
   DeviceInterface* p_device_{};          // The device we're running on (matches device_type_) used for things that work the same on all devices
   DeviceInterface* p_device_inputs_{};   // For some model inputs, the device might be the CPU device (all but KV cache currently for WebGPU and DML)
-  DeviceInterface* p_device_scoring_{};  // Device for search/scoring (sequences, token allocation). CPU for WebGPU since it uses GreedySearch_Cpu.
+  DeviceInterface* p_device_scoring_{};  // Device for search/scoring (sequences, token allocation).
   DeviceInterface* p_device_kvcache_{};  // The kvcache is always allocated in device memory  (TODO: Remove in favor of just p_device_?)
 
   Ort::Allocator& allocator_cpu_{GetDeviceInterface(DeviceType::CPU)->GetAllocator()};


src/models/model.cpp
```diff
@@ -1030,6 +1030,15 @@ Model::Model(std::unique_ptr<Config> config) : config_{std::move(config)} {
   else
     p_device_inputs_ = GetDeviceInterface(DeviceType::CPU);
 
+  // Scoring device for search/sequences. WebGPU uses GreedySearch_Cpu, so scoring always stays on CPU.
+  // For CUDA, scoring runs on GPU (with a CUDA-specific GreedySearch implementation).
+  // TODO: DML also delegates GreedySearch to CPU (Dml::InterfaceImpl::CreateGreedy). Consider expanding
+  // this check to include DML (and other CPU-search providers) to avoid CPU<->device transfer overhead.
+  if (p_device_->GetType() == DeviceType::WEBGPU)
```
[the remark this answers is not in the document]
qjia7: I prefer to keep the current changes. Theoretically, they should help DML and RyzenAI as well, but this hasn’t been verified. We can mark this as a TODO for now.

src/models/model.cpp
```diff
@@ -1030,6 +1030,15 @@ Model::Model(std::unique_ptr<Config> config) : config_{std::move(config)} {
   else
     p_device_inputs_ = GetDeviceInterface(DeviceType::CPU);
 
+  // Scoring device for search/sequences. WebGPU uses GreedySearch_Cpu, so scoring always stays on CPU.
+  // For CUDA, scoring runs on GPU (with a CUDA-specific GreedySearch implementation).
+  // TODO: DML also delegates GreedySearch to CPU (Dml::InterfaceImpl::CreateGreedy). Consider expanding
```
kunal-vaishnavi: Can we add the check for DML?

qjia7: Done

src/models/model.cpp
```diff
@@ -1030,6 +1030,15 @@ Model::Model(std::unique_ptr<Config> config) : config_{std::move(config)} {
...
 
+  // Scoring device for search/sequences. WebGPU, DML, and RyzenAI all use GreedySearch_Cpu,
+  // so scoring always stays on CPU. For CUDA, scoring runs on GPU (with a CUDA-specific GreedySearch implementation).
+  if (p_device_->GetType() == DeviceType::WEBGPU ||
+      p_device_->GetType() == DeviceType::DML ||
+      p_device_->GetType() == DeviceType::RyzenAI)
+    p_device_scoring_ = GetDeviceInterface(DeviceType::CPU);
+  else
+    p_device_scoring_ = p_device_inputs_;
+
```
baijumeswani: For scoring, we only do either on CPU or using CUDA.

CPU scoring is used for all EPs except where CUDA (CUDA EP and NvTensorRtRtx EP) can be used.

Checking for cuda or nvtrtrtx is probably simpler than checking the others.


```suggestion
  // Search and sampling are performed on the CPU for all device types,
  // except for CUDA and NvTensorRtRtx, where this is performed on the device.
  if (p_device_->GetType() == DeviceType::CUDA ||
      p_device_->GetType() == DeviceType::DeviceType::NvTensorRtRtx)
    p_device_scoring_ = p_device;
  else
    p_device_scoring_ = GetDeviceInterface(DeviceType::CPU);
```

src/models/model.h
```diff
@@ -166,6 +166,7 @@ struct Model : std::enable_shared_from_this<Model>, LeakChecked<Model>, External
 
   DeviceInterface* p_device_{};          // The device we're running on (matches device_type_) used for things that work the same on all devices
   DeviceInterface* p_device_inputs_{};   // For some model inputs, the device might be the CPU device (all but KV cache currently for WebGPU and DML)
+  DeviceInterface* p_device_scoring_{};  // Device for search/scoring (sequences, token allocation). CPU for WebGPU since it uses GreedySearch_Cpu.
```
kunal-vaishnavi:
```suggestion
  DeviceInterface* p_device_scoring_{};  // Device for search/scoring (sequences, token allocation).
```

PR #2022, opened 2026-03-13
[WebGPU] Add compact attention mask support

qjia7:
### Problem

The standard attention mask [batch_size, total_sequence_length] has below disadvantages for WebGPU graph capture:

1. Large pre-allocation — Graph capture requires all tensor shapes to be constant across runs. So the mask must be pre-allocated as [batch_size, max_length] (e.g., 4096) even when the actual sequence is short (e.g., 100 tokens). This wastes GPU memory.

2. Full buffer copy every step — ORT WebGPU's CopyTensors supports creating a smaller tensor view on an existing buffer, but only starting from offset 0 — you can't copy to an arbitrary offset within the buffer. For the attention mask, each decode step appends a 1 at the end (position = total_seq_len - 1). Since we can't copy starting from that offset, we must copy all previous values from the start — i.e., copy total_seq_len elements every step instead of just the 1 new element. As the sequence grows, this copy grows linearly (100 elements at step 100, 1000 at step 1000, etc.).

3. Additionally, the ONNX graph contains ReduceSum/ReduceMax subgraphs to derive seqlens_k/total_seq_len from the full binary mask, even though GenAI already knows these values on the CPU.

### Solution
Compact attention mask [batch_size, 1] — a single scalar per batch containing the total sequence length. This eliminates the large buffer, replaces the full-buffer copy with a 4-byte scalar copy, and simplifies the ONNX graph.

### Result
Graph capture works with continuous decoding. Simpler ONNX graph. No large static mask allocation needed.

### DIFF
--- a/src/python/py/models/builder.py
+++ b/src/python/py/models/builder.py
@@ -103,6 +103,10 @@ def check_extra_options(kv_pairs, execution_provider):
         )
         kv_pairs["enable_webgpu_graph"] = False
 
+    # When enable_webgpu_graph is on, also enable compact_attention_mask (unless explicitly set)
+    if kv_pairs.get("enable_webgpu_graph", False) and "compact_attention_mask" not in kv_pairs:
+        kv_pairs["compact_attention_mask"] = True
+
     if kv_pairs.get("compact_attention_mask", False) and execution_provider != "webgpu":
         print(
             "WARNING: compact_attention_mask is currently only supported with WebGPU execution provider. Disabling compact_attention_mask."


src/webgpu/interface.cpp
```diff
@@ -215,6 +215,53 @@ struct InterfaceImpl : DeviceInterface {
...
+
+      int64_t shape_val = static_cast<int64_t>(byte_count);
+      std::span<const int64_t> shape{&shape_val, 1};
+      auto cpu_mem_info = OrtMemoryInfo::CreateCpu(OrtDeviceAllocator, OrtMemTypeDefault);
+      auto src_tensor = OrtValue::CreateTensor(*cpu_mem_info, cpu_buffer.data(), byte_count, shape, ONNX_TENSOR_ELEMENT_DATA_TYPE_UINT8);
+      auto dst_tensor = OrtValue::CreateTensor(*ort_memory_info_, mask_data, byte_count, shape, ONNX_TENSOR_ELEMENT_DATA_TYPE_UINT8);
+
+      const std::vector<const OrtValue*> src_ptrs = {src_tensor.get()};
+      const std::vector<OrtValue*> dst_ptrs = {dst_tensor.get()};
+      GetOrtEnv().CopyTensors(src_ptrs, dst_ptrs, nullptr);
```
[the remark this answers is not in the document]
qjia7: Simplified [UpdateCompactAttentionMask](vscode-file://vscode-app/c:/Users/jiajiaqin/AppData/Local/Programs/Microsoft%20VS%20Code/ce099c1ed2/resources/app/out/vs/code/electron-browser/workbench/workbench.html) to only support [batch_beam_size == 1](vscode-file://vscode-app/c:/Users/jiajiaqin/AppData/Local/Programs/Microsoft%20VS%20Code/ce099c1ed2/resources/app/out/vs/code/electron-browser/workbench/workbench.html) (returns false for larger sizes, falling back to CPU path). Uses a single stack scalar instead of [std::vector](vscode-file://vscode-app/c:/Users/jiajiaqin/AppData/Local/Programs/Microsoft%20VS%20Code/ce099c1ed2/resources/app/out/vs/code/electron-browser/workbench/workbench.html), eliminating all per-token heap allocation. This is sufficient because graph capture enforces [batch_size == 1](vscode-file://vscode-app/c:/Users/jiajiaqin/AppData/Local/Programs/Microsoft%20VS%20Code/ce099c1ed2/resources/app/out/vs/code/electron-browser/workbench/workbench.html), and the compact mask fast path is gated on [batch_size == 1](vscode-file://vscode-app/c:/Users/jiajiaqin/AppData/Local/Programs/Microsoft%20VS%20Code/ce099c1ed2/resources/app/out/vs/code/electron-browser/workbench/workbench.html).

qjia7:
@kunal-vaishnavi @baijumeswani Please take a look, thanks.

src/models/position_inputs.cpp
```diff
@@ -372,6 +418,59 @@ void DefaultPositionInputs::CreateAndInitializeAttentionMask(DeviceSpan<int32_t>
...
+  std::array<int64_t, 2> compact_shape{attention_mask_shape_[0], 1};
+  auto attention_mask = OrtValue::CreateTensor(model_.allocator_cpu_, compact_shape, type_);
+  auto* mask_data = attention_mask->GetTensorMutableData<T>();
+
+  if (compact_shape[0] == 1) {
+    // Batch size == 1: no padding, total seq len = prompt_length
+    mask_data[0] = static_cast<T>(prompt_length);
+  } else {
+    // Count non-pad tokens per batch to get the effective sequence length
+    const auto* word_id = const_cast<DeviceSpan<int32_t>&>(next_tokens).CpuSpan().data();
```
kunal-vaishnavi, line 435: This feels like it is doing the same thing as the ONNX model already performs within itself: count the number of tokens to get the sequence length. Why not let the ONNX model do this work with a boolean mask of size `[batch_size, total_sequence_length]`? If the issue is about dtypes, we can always convert the dtype internally while keeping the existing input/output (I/O) structure.

In other words, what is the benefit of doing this? We intentionally picked a boolean mask of shape `[batch_size, total_sequence_length]` to standardize the I/O structure across models, to support a wide range of attention ops across EPs, and to allow the produced ONNX models to match and run with other solutions such as Optimum and Transformers.js for direct comparisons.

Does graph capture not work with the existing attention mask? If yes, then why do we have the following subgraph?

https://github.com/microsoft/onnxruntime-genai/blob/41c4ce18fec1240c2f848725e31fbe8854010188/src/python/py/models/builders/base.py#L4394-L4412

src/config.h
```diff
@@ -357,6 +357,9 @@ struct Config {
 
       std::vector<PipelineModel> pipeline;
 
+      bool compact_attention_mask{false};  // When true, attention_mask has shape [batch_size, 1] with the total sequence length value
```
kunal-vaishnavi, line 376: Instead of having an attribute and then adding a boolean here, couldn't we check the shape of the attention mask tensor to know whether it is compact or not? I don't think a user should need to specify this. It should be inferrable from the ONNX model.

src/models/position_inputs.cpp
qjia7: >Why not let the ONNX model do this work with a boolean mask of size [batch_size, total_sequence_length]?

In current genai logic, it seems that we have known the total_sequence_length in cpu, but still pass a boolean mask of size [batch_size, total_sequence_length] to the model. And let the model to get the total_sequence_length again. It seems that the purpose of attention mask is let model get the total sequence length. Why we can't directly pass the total_sequence_length to the model? In this case, the model itself doesn't need to calculate the total_sequence_length/seqlens_k from a boolean mask again. The benefit is that 
1) we don't need to allocate a big [batch_size, max_length] as the static tensor to support graph capture and the model itself can also be simplified. 
2) The interface's `UpdateCompactAttentionMask` method can be easily implemented by CopyTensors API. However, for the boolean attention mask, I need a CopyTensors with offset support to update a partial part of attention mask. (We haven't gotten an agreement how to support CopyTensors with offset in ORT)

That's why I want to support the compact attention which can perfectly resolve above two issues.

Ideally, I hope compact attention mask is only used when batch_size = 1. But we can't forbid users to use it when batch_size > 1. That's why I also add the else branch when batch_size > 1. How do you think?

src/config.h
qjia7: The config flag is needed because a [batch_size, 1] attention mask shape is ambiguous at runtime — it could be a compact mask (fixed shape, value = total_seq_len) or a standard binary mask where total_sequence_length happens to be 1.

qjia7:
**Performance Comparison: 2D Attention Opt (PR #2083) vs Compact Attention Mask (PR #2022)**

Benchmarked both approaches on phi4-graph-prune (WebGPU EP, RTX 5080) with `--reuse_generator -l 50 -g 500 -w 3 -r 3` across max lengths [4096, 8192, 12288, 16384]:

| Max Length | 2D Attention Opt | Compact Attention | Delta |
|---|---|---|---|
| 4096 | 160.47 tps | 161.89 tps | +0.9% |
| 8192 | 160.19 tps | 162.16 tps | +1.2% |
| 12288 | 161.69 tps | 161.88 tps | +0.1% |
| 16384 | 161.57 tps | 159.98 tps | -1.0% |

**Conclusion: The two approaches perform equivalently**  all differences are within 1.2% noise margin.

The benefit of compact attention mask is purely on memory: for max_length=8192, the 2D mask allocates 8192  8 bytes = 64 KB, while compact attention uses only 8 bytes. However, 64 KB is negligible relative to model weights and KV cache (hundreds of MB), so the complexity of compact attention mask doesn't justify the marginal memory savings.

**Decision:** Proceeding with the 2D attention optimization approach (PR #2083) as it's simpler and achieves equivalent performance. Closing this PR in favor of #2083.

qjia7:
**Correction:** The two approaches were tested on their respective models:
- **2D Attention Opt (PR #2083):** phi4-graph-prune (standard 2D attention mask)
- **Compact Attention Mask (PR #2022):** phi4-graph-prune-compact (compact attention mask enabled)

Both models are derived from the same phi4 base model  the only difference is the attention mask configuration. The performance comparison results above remain valid.

qjia7:
**Correction:** The two approaches were tested on their respective models:
- **2D Attention Opt (PR #2083):** phi4-graph-prune (standard 2D attention mask)
- **Compact Attention Mask (PR #2022):** phi4-graph-prune-compact (compact attention mask enabled)

Both models are derived from the same phi4 base model - the only difference is the attention mask configuration. The performance comparison results above remain valid.

PR #2027, opened 2026-03-13
[VitisAI] external_ep_library typo fix

akholodnamdcom:
Depends on https://github.com/microsoft/onnxruntime/pull/27647


kunal-vaishnavi (dismissed):

src/models/model.cpp
```diff
@@ -854,7 +854,7 @@ DeviceInterface* SetProviderSessionOptions(OrtSessionOptions& session_options,
 #if defined(_WIN32)
       if (provider_options.name == "VitisAI") {
         if (const auto opt_it = std::find_if(provider_options.options.begin(), provider_options.options.end(),
-                                             [](const auto& pair) { return pair.first == "external_ep_libray"; });
+                                             [](const auto& pair) { return pair.first == "external_ep_library"; });
```
baijumeswani: Please add backwards compatibility support. Since we do not know the version of the runtime library, we should support this failing and then falling back to the old name with the typo.

src/models/model.cpp
```diff
@@ -854,7 +854,7 @@ DeviceInterface* SetProviderSessionOptions(OrtSessionOptions& session_options,
...
             opt_it != provider_options.options.end()) {
           auto lib_name = opt_it->second;
           auto lib = LoadLibrary(lib_name.c_str());
```
baijumeswani: This loadlibrary call seems buggy. It loads the library, does not check for errors when loading the library, and does not close the library (leaking the resource).

Please use proper RAII. See [existing support here](https://github.com/microsoft/onnxruntime-genai/blob/8a31d029c53ca661b71a1b84f548eeeb7ae0d142/src/generators.cpp#L145-L193) for loading libraries.

Also, let's not guard this on windows if that is possible.

baijumeswani: On a separate note, why does the library need to be loaded inside ort-genai? It seems the library will be loaded inside ort anyway since we are registering it as a custom opes library. What value does it add to load it here? 

Will `CreateEpFactories` not be invoked when the plug-in library is loaded inside ort?

src/models/model.cpp
akholodnamdcom: This thing appears in configs of models, not runtime library. Not sure we should care about backward compatibility. Will double check.

src/models/model.cpp
akholodnamdcom: Let's ask original author (@BoarQing) of that thing to take a look. I do not have an env to try this out.

akholodnamdcom: @BoarQing could you please answer the question.

src/models/model.cpp
akholodnamdcom: No models with "external_ep_libray" were released. Backward compatibility is not required here.

src/models/model.cpp
akholodnamdcom: Addressed in c6162a3b:

  - **RAII**: each loaded library is now wrapped in a `std::unique_ptr` with a platform-aware deleter (`FreeLibrary` on Windows, `dlclose` elsewhere). Per the existing comment, the library has to stay loaded for the process lifetime (it provides the EP factory + custom ops the
  ORT session uses), so each handle is parked in a function-local `static std::vector<DllHandle>` — static destruction at process exit closes them in reverse load order. No more implicit leak.
  - **Error checking**: the load failure now throws with the path and the underlying `GetLastError()` / `dlerror()` message.
  - **Cross-platform**: dropped the `#if defined(_WIN32)` guard around the block. The new helpers route to `LoadLibrary`/`GetProcAddress` on Windows and `dlopen`/`dlsym` on POSIX, so the same code path covers both.

  Note on "use existing support": I checked `LibraryHandle` in `src/generators.cpp:163`, but it's a file-local struct *and* it prepends `CurrentModulePath()` to the filename (right for genai's internal `onnxruntime-genai-cuda.dll` lookup, wrong for an arbitrary user-provided EP
  library path). Rather than lifting it into a shared header and refactoring its path-resolution policy in the same PR, I inlined a minimal RAII wrapper here. Happy to do the consolidation as a follow-up if you'd prefer it lives in one place.

akholodnamdcom: Addressed in c6162a3b:

  - **RAII**: each loaded library is now wrapped in a `std::unique_ptr` with a platform-aware deleter (`FreeLibrary` on Windows, `dlclose` elsewhere). Per the existing comment, the library has to stay loaded for the process lifetime (it provides the EP factory + custom ops the
  ORT session uses), so each handle is parked in a function-local `static std::vector<DllHandle>` — static destruction at process exit closes them in reverse load order. No more implicit leak.
  - **Error checking**: the load failure now throws with the path and the underlying `GetLastError()` / `dlerror()` message.
  - **Cross-platform**: dropped the `#if defined(_WIN32)` guard around the block. The new helpers route to `LoadLibrary`/`GetProcAddress` on Windows and `dlopen`/`dlsym` on POSIX, so the same code path covers both.

  Note on "use existing support": I checked `LibraryHandle` in `src/generators.cpp:163`, but it's a file-local struct *and* it prepends `CurrentModulePath()` to the filename (right for genai's internal `onnxruntime-genai-cuda.dll` lookup, wrong for an arbitrary user-provided EP
  library path). Rather than lifting it into a shared header and refactoring its path-resolution policy in the same PR, I inlined a minimal RAII wrapper here. Happy to do the consolidation as a follow-up if you'd prefer it lives in one place.

akholodnamdcom: Restored the `#if defined(_WIN32)` guard in 218ee346, but kept the RAII + error-checking improvements (HModulePtr `unique_ptr` with `FreeLibrary` deleter, function-local static vector for process-lifetime ownership, `GetLastError`-aware error message). Going cross-platform here
   would need OGA to expose an `Ort::api_base` global (populated in `InitApiWithDynamicFn`) or route `CreateEpFactories` through an ORT-loaded entry point — happy to do that as a separate PR if it'd be worthwhile.

baijumeswani: >Going cross-platform here would need OGA to expose an Ort::api_base global (populated in InitApiWithDynamicFn) or route CreateEpFactories through an ORT-loaded entry point

Could you elaborate on this? I am not sure I follow

baijumeswani: @akholodnamdcom could you help me understand why the library loading is needed in ort-genai?

src/vitisai/session_options.cpp
```diff
@@ -25,28 +54,34 @@ DeviceInterface* AppendExecutionProvider(OrtSessionOptions& session_options,
...
 
 #if defined(_WIN32)
+  // Windows-only because the CreateEpFactories handshake below needs the
+  // OrtApiBase pointer that OrtGetApiBase returns, and only the Windows build
+  // of libonnxruntime-genai links against onnxruntime to resolve that symbol;
+  // on Linux ORT is loaded dynamically and OrtGetApiBase is not available at
+  // link time.
   if (const auto opt_it = std::find_if(provider_options.options.begin(), provider_options.options.end(),
-                                       [](const auto& pair) { return pair.first == "external_ep_libray"; });
+                                       [](const auto& pair) { return pair.first == "external_ep_library"; });
```
kunal-vaishnavi, line 63: Can we separate this PR into two (one for the typo fix and one for all of these other changes)? It should be cleaner that way since we should not condition the typo fix on these other changes.

baijumeswani:
@akholodnamdcom apologies for the ping. I this PR still relevant? I'll close this for now. If it is still relevant, please re-open and update.

PR #2035, opened 2026-03-17
Add Silero VAD Support to Nemotron Streaming ASR

sayanshaw24:
### Summary

Integrates [Silero VAD](https://github.com/snakers4/silero-vad) into the `StreamingProcessor` base class so that audio chunks without speech are automatically skipped before reaching the encoder/decoder pipeline. VAD is **disabled by default** and activated at runtime via `SetOption("use_vad", "true")` (requires a `"vad"` section in `genai_config.json`).

### Motivation

In streaming audio pipelines, many chunks contain silence or background noise. Without VAD, every chunk goes through the full RNNT encoder→decoder→joiner pipeline. By gating on speech detection, we skip silent chunks — saving compute and reducing end-to-end latency.

### Design

**Config-driven**: VAD configuration lives in the `"vad"` section of `genai_config.json`. The `SileroVad` class uses GenAI's `Model::CreateSession()` and `Model::CreateSessionOptionsFromConfig()` to inherit provider options, session options, and run options from the config.

**State infrastructure**: `SileroVadState` inherits from `State`, so VAD inference uses `State::Run()` — getting EP support, run options, `session_terminate`, and graph capture for free.

**Consecutive silence logic**: Instead of dropping every silent chunk immediately (which breaks natural pauses between words/sentences), the processor tracks consecutive silent chunks and only starts dropping after `silence_duration_ms` worth of silence elapses. `prefix_padding_ms` sets a minimum floor of silence to always keep for context. Speech resets the counter.

**Generic key-value API**: Following the same pattern as tokenizer options, all VAD settings are controlled via `SetOption`/`GetOption` rather than VAD-specific methods. This avoids API surface bloat and is extensible.

**Reusable base class**: All VAD logic lives in `StreamingProcessor` (base class), so any future streaming processor (Whisper, etc.) gets VAD support by calling `InitVadFromConfig()` and `ShouldDropChunk()`.

### Config

Add to `genai_config.json`:

    {
      "model": {
        "vad": {
          "filename": "silero_vad.onnx",
          "threshold": 0.3,
          "silence_duration_ms": 3360,
          "prefix_padding_ms": 560,
          "session_options": {},
          "run_options": {}
        }
      }
    }

VAD is off by default. Use `SetOption("use_vad", "true")` or `--use_vad true` in examples to enable. The `"vad"` section must exist in the config for VAD to be available.

### API Surface

All languages use the same generic `SetOption`/`GetOption` pattern on `StreamingProcessor`:

| Key | Values | Description |
|---|---|---|
| `use_vad` | `"true"` / `"false"` | Enable/disable VAD at runtime |
| `vad_threshold` | float as string | Speech probability threshold (default: `"0.5"`) |
| `silence_duration_ms` | int as string | Silence duration before dropping chunks (default: `"500"`) |
| `prefix_padding_ms` | int as string | Minimum silence to keep for context (default: `"300"`) |

C API: `OgaStreamingProcessorSetOption` / `OgaStreamingProcessorGetOption` (2 functions)

#### Core
- `src/config.h` / `src/config.cpp` — Added `VAD` struct to `Config::Model` with `filename`, `threshold`, `silence_duration_ms`, `prefix_padding_ms`, `session_options`, `run_options`
- `src/models/streaming_processor.h` / `src/models/streaming_processor.cpp` — Base class with `SetOption()`/`GetOption()`, `ShouldDropChunk()`, `InitVadFromConfig()`, ms→chunks conversion
- `src/models/nemotron_streaming_processor.h/.cpp` — Calls `InitVadFromConfig()` and `ShouldDropChunk()` from base class
- `src/models/silero_vad.h/.cpp` — `SileroVad` class using `Model::CreateSession()` + `SileroVadState` inheriting from `State` for proper EP/run-options support
- `src/models/model.h` — `CreateSessionOptionsFromConfig` made public (for VAD session creation)

#### C API
- `src/ort_genai_c.h/.cpp` — `OgaStreamingProcessorSetOption`, `OgaStreamingProcessorGetOption`

#### C++ Wrapper
- `src/ort_genai.h` — `SetOption()`/`GetOption()` on `OgaStreamingProcessor`

#### Python
- `src/python/python.cpp` — `.set_option()`, `.get_option()` on `StreamingProcessor`

#### C#
- `src/csharp/StreamingProcessor.cs` — `SetOption()`, `GetOption()` using `StringUtils`
- `src/csharp/NativeMethods.cs` — P/Invoke declarations

#### Other
- `src/leakcheck.h` — Added `SileroVad` to leak-tracked types

### Examples Updated

- **Python** (`examples/python/nemotron_speech.py`) — `--use_vad true/false` flag, VAD metrics output
- **C++** (`examples/c/src/nemotron_speech.cpp`) — `--use_vad true/false` flag, VAD metrics output
- **C#** (`examples/csharp/NemotronSpeech/Program.cs`) — `--use_vad true/false` flag, VAD metrics output

All examples follow the same structure: VAD off by default, `--use_vad true` enables if config has `"vad"` section, `--use_vad false` forces off.

### Tests Added

- **C++** (`test/c_api_tests.cpp`) — `StreamingASRVadSetGetOption`, `StreamingASRVadConsecutiveSilence`
- **Python** (`test/python/test_onnxruntime_genai_api.py`) — `test_streaming_asr_vad_set_get_option`, `test_streaming_asr_vad_consecutive_silence`

### Usage

Python:

    processor = og.StreamingProcessor(model)

    # Enable VAD (requires "vad" section in genai_config.json)
    processor.set_option("use_vad", "true")
    processor.set_option("vad_threshold", "0.3")
    processor.set_option("silence_duration_ms", "3360")

    # Check status
    print(processor.get_option("use_vad"))  # "true"

C#:

    processor.SetOption("use_vad", "true");
    processor.SetOption("vad_threshold", "0.5");
    Console.WriteLine(processor.GetOption("use_vad"));  // "true"

C++:

    processor->SetOption("use_vad", "true");
    processor->SetOption("silence_duration_ms", "3360");
    auto enabled = std::string(processor->GetOption("use_vad"));  // "true"

### Setup

Download `silero_vad.onnx`:

    huggingface-cli download onnx-community/silero-vad onnx/model.onnx --local-dir silero-vad
    cp silero-vad/onnx/model.onnx <nemotron_model_dir>/silero_vad.onnx

### DIFF
--- a/src/ort_genai.h
+++ b/src/ort_genai.h
@@ -910,12 +910,10 @@ struct OgaStreamingProcessor : OgaAbstract {
     OgaCheckResult(OgaStreamingProcessorSetOption(this, key, value));
   }
 
-  std::string GetOption(const char* key) const {
+  OgaString GetOption(const char* key) const {
     const char* value;
     OgaCheckResult(OgaStreamingProcessorGetOption(this, key, &value));
-    std::string result(value);
-    OgaDestroyString(value);
-    return result;
+    return value;
   }
 
   static void operator delete(void* p) { OgaDestroyStreamingProcessor(reinterpret_cast<OgaStreamingProcessor*>(p)); }


nenad1002:
Do you have WER results? If not, we first need to get those before we can discuss which approach to take.

sayanshaw24:
> Do you have WER results? If not, we first need to get those before we can discuss which approach to take.

Yes, here are WER results for 4 datasets (skipping larger ones which I'll run over the weekend but this gives us a good idea):
 
| Dataset | Baseline WER (no VAD) % | WER with VAD % | Delta |
|---|---|---|---|
| LibriSpeech test-clean | 2.39 | 2.98 | +0.59 |
| LibriSpeech test-other | 5.05 | 7.40 | +2.35 |
| Earnings22 | 13.52 | 18.12 | +4.60 |
| AMI | 12.64 | 22.14 | +9.50 |
 
Copilot suggests trying with threshold of 0.3, I think long term we'll need more experimentation to determine a good value. Regardless, I think consensus is to have VAD disabled by default.

examples/c/src/nemotron_speech.cpp
```diff
@@ -166,6 +166,17 @@ void StreamingTranscribe(const std::string& model_path, const std::string& audio
   auto config = OgaConfig::Create(model_path.c_str());
   auto model = OgaModel::Create(*config);
   auto processor = OgaStreamingProcessor::Create(*model);
+
+  // VAD is disabled by default. Enable via genai_config.json or --enable_vad.
+  if (enable_vad && std::string(processor->GetOption("vad_enabled")) != "true") {
+    processor->SetOption("vad_enabled", "true");
+  }
+  if (std::string(processor->GetOption("vad_enabled")) == "true") {
+    std::cout << "  VAD: enabled" << std::endl;
```
nenad1002: Is this standard to print like this from genai? Would this show up to FL users? I would assume we want FL SDK logger to handle on that side.

src/models/nemotron_streaming_processor.cpp
```diff
@@ -47,7 +53,28 @@ std::unique_ptr<NamedTensors> NemotronStreamingProcessor::Process(const float* a
 
   // Process the first complete chunk available
   if (audio_buffer_.size() >= chunk_size) {
-    auto mel = BuildMelTensor(audio_buffer_.data(), chunk_size);
+    const float* chunk_data = audio_buffer_.data();
+
+    // VAD check with consecutive silence tracking:
+    // Only start dropping chunks after min_silence_chunks_ consecutive silent chunks.
+    // This preserves natural pauses between words/sentences.
+    if (vad_) {
```
nenad1002: I'm wondering should this be a part of a separate processor that can get called by nemotron (or any other model?)
I know we can't have full logic there, but at least logic to decide whether to erase the buffer

src/models/nemotron_streaming_processor.cpp
```diff
@@ -135,4 +162,60 @@ std::unique_ptr<StreamingProcessor> CreateStreamingProcessor(Model& model) {
...
+
+void NemotronStreamingProcessor::EnableVadFromConfig() {
+  vad_ = CreateSileroVad(model_);
+  consecutive_silence_chunks_ = 0;
+}
+
+void NemotronStreamingProcessor::SetOption(const char* key, const char* value) {
+  std::string_view k{key};
+
+  if (k == "vad_enabled") {
```
nenad1002: Same for all of this, should probably be reusable in a separate processor / class

src/models/silero_vad.cpp
```diff
@@ -0,0 +1,175 @@
...
+#include "silero_vad.h"
+
+namespace Generators {
+
+// Shared initialization for both constructors
+void SileroVad::Initialize(int sample_rate, float threshold) {
+  sample_rate_ = sample_rate;
+  threshold_ = threshold;
+
+  if (sample_rate != 16000 && sample_rate != 8000) {
```
nenad1002, line 64: Nemotron only supports 16k, but we still don't hardcode like this, you never know when silero-like model could come, and could easily get reused without changing code. 

I don't have a strong stance about this, but I would consider either 1) adding silero genai_config SR or even better 2) reusing SR by the backend model if possible (which in this case is nemotron, and both have to use the same SR).

src/models/silero_vad.cpp
```diff
@@ -0,0 +1,175 @@
...
+    throw std::runtime_error("SileroVad only supports sample rates 16000 and 8000. Got: " +
+                             std::to_string(sample_rate));
+  }
+
+  window_size_ = (sample_rate == 16000) ? 512 : 256;
```
nenad1002: const int factor = sample_rate / 8000;

window_size_  = 256 * factor;
context_size_ = 32  * factor;

Is this correct?

src/models/silero_vad.cpp
```diff
@@ -0,0 +1,175 @@
...
+  auto& vad_config = model.config_->model.vad;
+
+  // Use the model's existing session options (inherits provider options, thread settings, etc.)
+  // VAD is a lightweight model that doesn't need its own provider configuration.
+  OrtSessionOptions* session_opts = model.session_options_.get();
+
+  // If VAD-specific session options are provided in config, create dedicated options
+  if (vad_config.session_options.has_value()) {
+    session_options_ = OrtSessionOptions::Create();
+    session_options_->SetIntraOpNumThreads(1);
```
nenad1002: Document why, I believe 1 is sufficient for such a small model, but I would be wondering why not try to utilize all cores even if gain is marginal.

src/models/silero_vad.cpp
```diff
@@ -0,0 +1,175 @@
...
+
+  if (sample_rate != 16000 && sample_rate != 8000) {
+    throw std::runtime_error("SileroVad only supports sample rates 16000 and 8000. Got: " +
+                             std::to_string(sample_rate));
+  }
+
+  window_size_ = (sample_rate == 16000) ? 512 : 256;
+  context_size_ = (sample_rate == 16000) ? 64 : 32;
+
+  memory_info_ = OrtMemoryInfo::CreateCpu(OrtDeviceAllocator, OrtMemTypeCPU);
```
nenad1002: No plan for this to run on GPU for future models?

src/models/silero_vad.cpp
```diff
@@ -0,0 +1,175 @@
...
+  std::memcpy(context_.data(), samples + window_size_ - context_size_,
+              context_size_ * sizeof(float));
+
+  return speech_prob;
+}
+
+bool SileroVad::ContainsSpeech(const float* samples, size_t num_samples) {
+  const size_t window = static_cast<size_t>(window_size_);
+
+  for (size_t offset = 0; offset + window <= num_samples; offset += window) {
```
nenad1002, line 157: What does this do? Document/comment

examples/c/src/nemotron_speech.cpp
sayanshaw24: I believe so, we do this elsewhere too - do not think this will get picked up in the FL output since it just takes the model generation output; + @kunal-vaishnavi to confirm.

src/models/nemotron_streaming_processor.cpp
sayanshaw24: Good point, will refactor so it can be used by any model.

src/models/silero_vad.cpp
sayanshaw24: This was given the nemotron model from David is for CPU int4 - thoughts @jiafatom?

src/models/silero_vad.cpp
sayanshaw24: sounds good, will update it to use the model config and keep a default for fallback instead.

src/models/silero_vad.cpp
sayanshaw24: Will do

src/models/silero_vad.cpp
sayanshaw24: yep!

src/models/silero_vad.cpp
nenad1002: The current one is CPU only, but we will almost certainly extend to GPU support, although not a priority atm

examples/c/src/nemotron_speech.cpp
kunal-vaishnavi: This is an example to a user so it should be fine to print with `cout`. Inside GenAI, we can use the logger and set the logging level.

src/csharp/StreamingProcessor.cs
```diff
@@ -46,6 +46,33 @@ public StreamingProcessor(Model model)
...
 
+        /// <summary>
+        /// Set a processor option as a key-value pair.
+        /// Supported keys: "vad_enabled", "vad_threshold", "vad_min_silence_chunks", "vad_model_path".
+        /// </summary>
+        public void SetOption(string key, string value)
+        {
+            Result.VerifySuccess(NativeMethods.OgaStreamingProcessorSetOption(
+                _processorHandle,
+                System.Text.Encoding.UTF8.GetBytes(key + '\0'),
```
kunal-vaishnavi: Can we use `StringUtils.ToUtf8` instead? Ex:

https://github.com/microsoft/onnxruntime-genai/blob/2ea630fc884a86b365eca65b4a28f6aa80a48f0b/src/csharp/Model.cs#L14-L17

src/csharp/StreamingProcessor.cs
```diff
@@ -46,6 +46,33 @@ public StreamingProcessor(Model model)
...
+
+        /// <summary>
+        /// Get a processor option value by key.
+        /// </summary>
+        public string GetOption(string key)
+        {
+            IntPtr valuePtr;
+            Result.VerifySuccess(NativeMethods.OgaStreamingProcessorGetOption(
+                _processorHandle,
+                System.Text.Encoding.UTF8.GetBytes(key + '\0'),
```
kunal-vaishnavi: Can we use `StringUtils.FromUtf8` instead? Ex:

https://github.com/microsoft/onnxruntime-genai/blob/2ea630fc884a86b365eca65b4a28f6aa80a48f0b/src/csharp/Model.cs#L26-L38

src/models/silero_vad.cpp
kunal-vaishnavi: The runtime infra for Nemotron speech is also using the same approach.

https://github.com/microsoft/onnxruntime-genai/blob/80dd558361931a967b8e818a3592dc00857e9241/src/models/nemotron_speech.cpp#L460

Inside ORT GenAI, there are multiple ways for getting an allocator.

https://github.com/microsoft/onnxruntime-genai/blob/80dd558361931a967b8e818a3592dc00857e9241/src/models/whisper.cpp#L41
https://github.com/microsoft/onnxruntime-genai/blob/80dd558361931a967b8e818a3592dc00857e9241/src/models/whisper.cpp#L67

Instead of storing `memory_info_`, we can use the first approach to get the allocator from the device inputs to know which allocator to use. This will allow us to support multiple devices. If this model will always require the input to be on CPU first, then we can use the second approach.

src/models/silero_vad.cpp
```diff
@@ -0,0 +1,189 @@
...
+  // VAD is a lightweight model that doesn't need its own provider configuration.
+  OrtSessionOptions* session_opts = model.session_options_.get();
+
+  // If VAD-specific session options are provided in config, create dedicated options
+  if (vad_config.session_options.has_value()) {
+    session_options_ = OrtSessionOptions::Create();
+    // Silero VAD is a ~2MB model with minimal compute (<1ms per window).
+    // Single-threaded execution avoids thread pool overhead which would exceed
+    // the inference time itself, and prevents contention with the main ASR model.
+    session_options_->SetIntraOpNumThreads(1);
```
kunal-vaishnavi: Can we set these values as defaults in the GenAI config as well and read their values from that instead? Someone may want to experiment with those values on their machine. If the values are not set, then we can use these default values.

src/models/silero_vad.cpp
```diff
@@ -0,0 +1,189 @@
...
+    session_options_->SetInterOpNumThreads(1);
+    session_options_->SetGraphOptimizationLevel(GraphOptimizationLevel::ORT_ENABLE_ALL);
+    session_opts = session_options_.get();
+  }
+
+  // Load session through Model::CreateSession (handles model data spans, path resolution)
+  std::string filename = vad_config.filename;
+  if (filename.empty()) filename = "silero_vad.onnx";
+  session_ = model.CreateSession(GetOrtEnv(), filename, session_opts);
```
kunal-vaishnavi: Can we use the GenAI infra of a `State` to run the model instead? There are extra benefits to using the GenAI infra instead of using direct ORT APIs.

src/models/silero_vad.cpp
```diff
@@ -0,0 +1,189 @@
...
+
+  // If VAD-specific session options are provided in config, create dedicated options
+  if (vad_config.session_options.has_value()) {
+    session_options_ = OrtSessionOptions::Create();
+    // Silero VAD is a ~2MB model with minimal compute (<1ms per window).
+    // Single-threaded execution avoids thread pool overhead which would exceed
+    // the inference time itself, and prevents contention with the main ASR model.
+    session_options_->SetIntraOpNumThreads(1);
+    session_options_->SetInterOpNumThreads(1);
+    session_options_->SetGraphOptimizationLevel(GraphOptimizationLevel::ORT_ENABLE_ALL);
```
kunal-vaishnavi: This can be removed. Graph optimization is set to enable all by [default](https://onnxruntime.ai/docs/performance/model-optimizations/graph-optimizations.html#graph-optimization-levels).

src/models/silero_vad.cpp
```diff
@@ -0,0 +1,189 @@
...
+  }
+
+  // Load session through Model::CreateSession (handles model data spans, path resolution)
+  std::string filename = vad_config.filename;
+  if (filename.empty()) filename = "silero_vad.onnx";
+  session_ = model.CreateSession(GetOrtEnv(), filename, session_opts);
+
+  // Create run options from config if specified
+  if (vad_config.run_options.has_value()) {
+    run_options_ = OrtRunOptions::Create();
```
kunal-vaishnavi: This is another reason for using the GenAI infra of a `State` because run options are [already supported](https://github.com/microsoft/onnxruntime-genai/blob/80dd558361931a967b8e818a3592dc00857e9241/src/models/model.cpp#L200) there.

src/models/silero_vad.cpp
```diff
@@ -0,0 +1,189 @@
...
+  }
+
+  // Use VAD-specific sample_rate if configured, otherwise fall back to model's sample_rate
+  int sr = vad_config.sample_rate > 0 ? vad_config.sample_rate : model.config_->model.sample_rate;
+  Initialize(sr, vad_config.threshold);
+}
+
+// Constructor with explicit path — standalone/programmatic usage
+SileroVad::SileroVad(const char* model_path, int sample_rate, float threshold) {
+  session_options_ = OrtSessionOptions::Create();
```
kunal-vaishnavi: These constructors still seem duplicated. Only one should be necessary for creating the `Model` and `State` classes. Ex:

https://github.com/microsoft/onnxruntime-genai/blob/80dd558361931a967b8e818a3592dc00857e9241/src/models/decoder_only.cpp#L5-L28

src/models/silero_vad.cpp
```diff
@@ -0,0 +1,189 @@
...
+  session_options_->SetInterOpNumThreads(1);
+  session_options_->SetGraphOptimizationLevel(GraphOptimizationLevel::ORT_ENABLE_ALL);
+
+  auto& ort_env = GetOrtEnv();
+#ifdef _WIN32
+  std::wstring wide_path;
+  int len = MultiByteToWideChar(CP_UTF8, 0, model_path, -1, nullptr, 0);
+  wide_path.resize(len - 1);
+  MultiByteToWideChar(CP_UTF8, 0, model_path, -1, wide_path.data(), len);
+  session_ = OrtSession::Create(ort_env, wide_path.c_str(), session_options_.get());
```
kunal-vaishnavi: There is an existing `CreateSession` method that handles all of this. Let's use that and clean this up.

https://github.com/microsoft/onnxruntime-genai/blob/80dd558361931a967b8e818a3592dc00857e9241/src/models/decoder_only.cpp#L7

src/models/silero_vad.cpp
```diff
@@ -0,0 +1,189 @@
...
+  int64_t sr_shape[] = {1};
+  auto sr_tensor = OrtValue::CreateTensor<int64_t>(
+      *memory_info_, std::span<int64_t>(&sr_value_, 1), std::span<const int64_t>(sr_shape, 1));
+
+  const char* input_names[] = {"input", "state", "sr"};
+  const char* output_names[] = {"output", "stateN"};
+  OrtValue* inputs[] = {input_tensor.get(), state_tensor.get(), sr_tensor.get()};
+
+  // Run inference through the session with proper run options
+  auto ort_outputs = session_->Run(
```
kunal-vaishnavi: Same comments as before. We should use the GenAI infra. Then, we don't need to manually specify all of this: input and output names, run options, etc.

src/python/python.cpp
```diff
@@ -660,7 +660,72 @@ PYBIND11_MODULE(onnxruntime_genai, m) {
...
+      .def(
+          "get_option",
+          [](OgaStreamingProcessor& proc, const std::string& key) {
+            return std::string(proc.GetOption(key.c_str()));
+          },
+          pybind11::arg("key"),
+          "Get a processor option value by key.");
+
+  pybind11::class_<OgaSileroVad>(m, "SileroVad")
+      .def(pybind11::init([](const std::string& model_path, int sample_rate, float threshold) {
```
kunal-vaishnavi: We should remove this constructor and only construct the VAD internally.

src/python/python.cpp
```diff
@@ -660,7 +660,72 @@ PYBIND11_MODULE(onnxruntime_genai, m) {
...
+          "Set a processor option. Keys: 'vad_enabled', 'vad_threshold', 'vad_min_silence_chunks', 'vad_model_path'.")
+      .def(
+          "get_option",
+          [](OgaStreamingProcessor& proc, const std::string& key) {
+            return std::string(proc.GetOption(key.c_str()));
+          },
+          pybind11::arg("key"),
+          "Get a processor option value by key.");
+
+  pybind11::class_<OgaSileroVad>(m, "SileroVad")
```
kunal-vaishnavi: Why do we need to expose `SileroVad` at the language binding level? The `SetOption`/`GetOption` APIs on the processor as well as any GenAI config parsing should be sufficient.

src/config.cpp
```diff
@@ -334,6 +334,8 @@ struct DecoderInputs_Element : JSON::Element {
       v_.block_table = JSON::Get<std::string_view>(value);
     } else if (name == "targets") {
       v_.targets = JSON::Get<std::string_view>(value);
+    } else if (name == "target_length") {
```
kunal-vaishnavi: We removed the need for these extra inputs and outputs for length. I believe this should go away once you are synced with the main branch.

src/config.cpp
```diff
@@ -962,6 +966,45 @@ struct Joiner_Element : JSON::Element {
   JoinerOutputs_Element outputs_{v_.outputs};
 };
 
+struct Vad_Element : JSON::Element {
```
kunal-vaishnavi:
```suggestion
struct VAD_Element : JSON::Element {
```

src/config.h
```diff
@@ -275,6 +275,16 @@ struct Config {
       } outputs;
     } joiner;
 
+    struct Vad {
```
```suggestion
    struct VAD {
```

src/config.h
```diff
@@ -275,6 +275,16 @@ struct Config {
...
+      bool enabled{false};
+      std::string filename{"silero_vad.onnx"};
```
kunal-vaishnavi: Filename should always be required in a GenAI config section. Let's not set a default filename.

src/config.h
```diff
@@ -319,6 +329,7 @@ struct Config {
 
         // RNNT decoder inputs
         std::string targets;
+        std::string target_length;
```
kunal-vaishnavi: Same comment as [here](https://github.com/microsoft/onnxruntime-genai/pull/2035/changes#r2977095088)

src/config.h
```diff
@@ -333,6 +344,7 @@ struct Config {
 
         // RNNT decoder outputs
         std::string outputs;
+        std::string prednet_lengths;
```
kunal-vaishnavi: Same comment as [here](https://github.com/microsoft/onnxruntime-genai/pull/2035/changes#r2977095088)

src/ort_genai.h
```diff
@@ -906,5 +906,45 @@ struct OgaStreamingProcessor : OgaAbstract {
...
+  OgaString GetOption(const char* key) {
+    const char* value;
+    OgaCheckResult(OgaStreamingProcessorGetOption(this, key, &value));
+    return OgaString(value);
+  }
+
   static void operator delete(void* p) { OgaDestroyStreamingProcessor(reinterpret_cast<OgaStreamingProcessor*>(p)); }
 };
+
+struct OgaSileroVad : OgaAbstract {
```
kunal-vaishnavi: Same question as [here](https://github.com/microsoft/onnxruntime-genai/pull/2035/changes#r2977089348) but for the C++ language binding

src/ort_genai_c.cpp
```diff
@@ -1128,4 +1130,57 @@ OgaResult* OGA_API_CALL OgaStreamingProcessorFlush(OgaStreamingProcessor* proces
...
+
+OgaResult* OGA_API_CALL OgaStreamingProcessorGetOption(OgaStreamingProcessor* processor, const char* key, const char** value) {
+  OGA_TRY
+  auto result = processor->GetOption(key);
+  *value = AllocOgaString(result);
+  return nullptr;
+  OGA_CATCH
+}
+
+OgaResult* OGA_API_CALL OgaCreateSileroVad(const char* model_path, int32_t sample_rate, float threshold, OgaSileroVad** out) {
```
kunal-vaishnavi: Same question as [here](https://github.com/microsoft/onnxruntime-genai/pull/2035/changes#r2977089348) but for the below C APIs

test/c_api_tests.cpp
```diff
@@ -1510,3 +1510,63 @@ TEST(CAPITests, StreamingASRRawCAPI) {
   OgaDestroyStreamingProcessor(processor);
   OgaDestroyModel(model);
 }
+
+// Test VAD set_option/get_option on StreamingProcessor
+TEST(CAPITests, StreamingASRVadSetGetOption) {
+  if (!std::filesystem::exists(STREAMING_ASR_PATH))
```
kunal-vaishnavi, line 1520: Where is `STREAMING_ASR_PATH` defined?

test/c_api_tests.cpp
```diff
@@ -1510,3 +1510,63 @@ TEST(CAPITests, StreamingASRRawCAPI) {
...
+
+  // Default: VAD disabled
+  ASSERT_EQ(std::string(processor->GetOption("vad_enabled")), "false");
+
+  // Set and get threshold
+  processor->SetOption("vad_min_silence_chunks", "10");
+  ASSERT_EQ(std::string(processor->GetOption("vad_min_silence_chunks")), "10");
+
+  // Enable VAD if silero_vad.onnx is available
+  std::string vad_path = std::string(STREAMING_ASR_PATH) + "/silero_vad.onnx";
```
kunal-vaishnavi: To use a cleaner, more cross-platform approach:

```suggestion
  std::filesystem::path vad_path = std::filesystem::path(STREAMING_ASR_PATH) / "silero_vad.onnx";
```

Then, you can get the type you need afterwards.

```cpp
std::string vad_path_str = vad_path.string();          // or .u8string() if you want UTF-8
const char* vad_path_cstr = vad_path_str.c_str();
```

test/c_api_tests.cpp
```diff
@@ -1510,3 +1510,63 @@ TEST(CAPITests, StreamingASRRawCAPI) {
...
+  }
+  SUCCEED();
+}
+
+// Test consecutive silence logic: VAD should not drop chunks until min_silence_chunks exceeded
+TEST(CAPITests, StreamingASRVadConsecutiveSilence) {
+  if (!std::filesystem::exists(STREAMING_ASR_PATH))
+    GTEST_SKIP() << "Streaming ASR model not found at " << STREAMING_ASR_PATH;
+
+  std::string vad_path = std::string(STREAMING_ASR_PATH) + "/silero_vad.onnx";
```
kunal-vaishnavi: Same comment as [here](https://github.com/microsoft/onnxruntime-genai/pull/2035/changes#r2977154849)

src/models/silero_vad.cpp
sayanshaw24: Sounds good, will go with the first approach with CPU fallback, so we can use this even after GPU support is added in the future with no issues.

src/models/silero_vad.cpp
sayanshaw24: Updated to use registered I/O vectors and a Run() method matching the State::Run pattern. Considered full State inheritance but that requires GeneratorParams which is heavyweight for a ~2MB VAD model. The current approach mirrors the State pattern (registered inputs/outputs, run options from config) without the GeneratorParams dependency.

test/c_api_tests.cpp
sayanshaw24: Here: https://github.com/microsoft/onnxruntime-genai/blob/main/test/c_api_tests.cpp#L1370 (from Nenad's ASR PR)

src/models/silero_vad.cpp
sayanshaw24: Done. Removed the standalone SileroVad(const char* model_path, ...) constructor and its factory. Only the SileroVad(Model&) constructor remains, using Model::CreateSession().

src/python/python.cpp
sayanshaw24: Good point, will remove it!

src/python/python.cpp
```diff
@@ -660,7 +660,71 @@ PYBIND11_MODULE(onnxruntime_genai, m) {
...
+      .def("reset", [](OgaSileroVad& vad) { vad.Reset(); },
+           "Reset VAD state for a new utterance.")
+      .def("set_threshold", [](OgaSileroVad& vad, float threshold) { vad.SetThreshold(threshold); },
+           pybind11::arg("threshold"),
+           "Set the speech probability threshold.")
+      .def_property_readonly("window_size", [](OgaSileroVad&) -> int {
+             // These are fixed by the model; expose for convenience.
+             // We can't call GetWindowSize through the C API wrapper, so return the known values.
+             // 512 for 16kHz, 256 for 8kHz (default 16kHz).
+             return 512;  // TODO: expose via C API if needed
```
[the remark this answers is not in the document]
sayanshaw24: Resolved by removing the standalone SileroVad Python binding entirely. VAD window size is now internal.

src/models/silero_vad.h
```diff
@@ -0,0 +1,83 @@
...
+  /// Process arbitrary-length audio.
+  /// Internally splits into windows and runs VAD on each.
+  /// Returns true if any window's speech probability exceeds the threshold.
+  bool ContainsSpeech(const float* samples, size_t num_samples);
+
+  /// Reset hidden state and context for a new utterance.
+  void Reset();
+
+  float GetThreshold() const { return threshold_; }
+  void SetThreshold(float threshold) { threshold_ = threshold; }
```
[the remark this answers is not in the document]
sayanshaw24: Done. SetThreshold() now clamps to [0.0, 1.0].

src/models/silero_vad.cpp
```diff
@@ -0,0 +1,140 @@
...
+  session_options_->SetGraphOptimizationLevel(GraphOptimizationLevel::ORT_ENABLE_ALL);
+
+  // Load the ONNX model
+  auto& ort_env = GetOrtEnv();
+#ifdef _WIN32
+  // Convert to wide string on Windows
+  std::wstring wide_path;
+  int len = MultiByteToWideChar(CP_UTF8, 0, model_path, -1, nullptr, 0);
+  wide_path.resize(len - 1);
+  MultiByteToWideChar(CP_UTF8, 0, model_path, -1, wide_path.data(), len);
```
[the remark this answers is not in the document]
sayanshaw24: removed constructor

test/c_api_tests.cpp
```diff
@@ -1510,3 +1510,53 @@ TEST(CAPITests, StreamingASRRawCAPI) {
...
+  if (!std::filesystem::exists(vad_path))
+    GTEST_SKIP() << "silero_vad.onnx not found in model dir";
+
+  auto model = OgaModel::Create(STREAMING_ASR_PATH);
+  auto processor = OgaStreamingProcessor::Create(*model);
+  processor->EnableVad(vad_path.c_str(), 0.5f);
+
+  constexpr size_t chunk_samples = 8960;
+  std::vector<float> silence(chunk_samples, 0.0f);
```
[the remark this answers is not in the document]
sayanshaw24: updated!

test/python/test_onnxruntime_genai_api.py
```diff
@@ -1066,6 +1066,50 @@ def test_streaming_asr_config_model_type(nemotron_speech_model_path):
     assert model.type == "nemotron_speech"
 
 
+def test_streaming_asr_vad_enable_disable(nemotron_speech_model_path):
+    """Test that VAD can be enabled and disabled on StreamingProcessor."""
+    model = og.Model(nemotron_speech_model_path)
+    processor = og.StreamingProcessor(model)
+
+    # VAD may or may not be auto-enabled depending on silero_vad.onnx presence
+    initially_enabled = processor.is_vad_enabled
```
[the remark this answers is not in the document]
sayanshaw24: outdated

examples/python/nemotron_speech.py
```diff
@@ -56,6 +56,31 @@ def simulate_microphone(model_path, audio_path, execution_provider):
     config = get_config(model_path, execution_provider)
     model = og.Model(config)
     processor = og.StreamingProcessor(model)
+
+    # Track memory before/after VAD init
+    try:
+        import psutil
```
kunal-vaishnavi: Can we move this import to the top of the file instead of try-except? Most users should be able to install it before getting this example to run.

src/models/silero_vad.cpp
```diff
@@ -0,0 +1,185 @@
...
+  inputs_ = {input_tensor_.get(), state_tensor_.get(), sr_tensor_.get()};
+
+  // Register output names and pointers (ORT allocates outputs)
+  output_names_ = {"output", "stateN"};
+  outputs_ = {nullptr, nullptr};
+}
+
+void SileroVad::Run() {
+  // Run inference through registered I/O, matching the State::Run pattern
+  session_->Run(
```
kunal-vaishnavi: This matches `State::Run` but is still calling the ORT API directly. If we construct a `State` object, we can invoke its `Run` method. Since the VAD model will exist alongside other models such as the encoder, decoder, and joiner, the generator params object needed for the VAD state will already exist or can otherwise be created. It is fine to re-use from the decoder section when something is not specified as that is the default/fallback section.

src/models/silero_vad.h
```diff
@@ -0,0 +1,100 @@
...
+
+ private:
+  void Initialize(int sample_rate, float threshold);
+  void RegisterInputsOutputs();
+  void Run();
+
+  std::unique_ptr<OrtSession> session_;
+  std::unique_ptr<OrtSessionOptions> session_options_;
+  std::unique_ptr<OrtRunOptions> run_options_;
+  Ort::Allocator* allocator_{nullptr};
```
kunal-vaishnavi: Let's use `std::unique_ptr` here instead of a raw pointer

src/models/silero_vad.cpp
```diff
@@ -0,0 +1,185 @@
...
+// Constructor from Model — uses GenAI's session creation infrastructure
+SileroVad::SileroVad(Model& model) {
+  auto& vad_config = model.config_->model.vad;
+
+  // Use the model's device input allocator — supports GPU if configured
+  allocator_ = &model.p_device_inputs_->GetAllocator();
+
+  // Use the model's existing session options (inherits provider options, thread settings, etc.)
+  // VAD is a lightweight model that doesn't need its own provider configuration.
+  OrtSessionOptions* session_opts = model.session_options_.get();
```
kunal-vaishnavi: If `session_options_` exists, why does this need to be created? A cleaner approach would be to re-use the `CreateSessionOptionsFromConfig` method.

https://github.com/microsoft/onnxruntime-genai/blob/07d21ba199eb4e375d588e76a7b8ba427cf5e3db/src/models/multi_modal.cpp#L103-L105

src/models/silero_vad.cpp
```diff
@@ -0,0 +1,185 @@
...
+    session_options_->SetInterOpNumThreads(so.inter_op_num_threads.value_or(1));
+    session_opts = session_options_.get();
+  }
+
+  // Load session through Model::CreateSession (handles model data spans, path resolution)
+  std::string filename = vad_config.filename;
+  if (filename.empty()) {
+    throw std::runtime_error("VAD filename must be specified in genai_config.json when vad.enabled is true.");
+  }
+  session_ = model.CreateSession(GetOrtEnv(), filename, session_opts);
```
kunal-vaishnavi: We can convert this class to `SileroVadModel` and then use the `CreateSession` method directly. That would follow the setup that other models have (even ones that aren't generative and are only called once).

src/models/silero_vad.cpp
```diff
@@ -0,0 +1,185 @@
...
+
+  // Create run options from config if specified
+  if (vad_config.run_options.has_value()) {
+    run_options_ = OrtRunOptions::Create();
```
kunal-vaishnavi: This section should ideally be done in `SileroVadState::Run`. There are other run options that could get called that don't go through `AddConfigEntry`. Ex:

https://github.com/microsoft/onnxruntime-genai/blob/07d21ba199eb4e375d588e76a7b8ba427cf5e3db/src/models/multi_modal.cpp#L340-L346

src/models/streaming_processor.h
```diff
@@ -4,21 +4,58 @@
...
+  /// Returns true if VAD is enabled and the chunk should be skipped (prolonged silence).
+  /// Handles consecutive silence tracking internally.
+  bool ShouldDropChunk(const float* chunk_data, size_t chunk_size);
+
+  /// Initialize VAD from the model's genai_config.json vad section.
+  /// Called by derived constructors if config has vad.enabled = true.
+  void InitVadFromConfig(Model& model);
+
+ private:
+  Model* model_{nullptr};  // Stored for deferred VAD creation via SetOption
```
kunal-vaishnavi: Can we use a `std::unique_ptr` instead of a raw pointer?

src/config.h
```diff
@@ -275,6 +275,16 @@ struct Config {
       } outputs;
     } joiner;
 
+    struct VAD {
+      bool enabled{false};
+      std::string filename;
+      float threshold{0.5f};
```
kunal-vaishnavi, line 280: Should the threshold determine whether VAD is enabled or disabled rather than a standalone boolean?

src/ort_genai.h
```diff
@@ -906,5 +906,15 @@ struct OgaStreamingProcessor : OgaAbstract {
...
   }
 
+  void SetOption(const char* key, const char* value) {
+    OgaCheckResult(OgaStreamingProcessorSetOption(this, key, value));
+  }
+
+  OgaString GetOption(const char* key) {
+    const char* value;
+    OgaCheckResult(OgaStreamingProcessorGetOption(this, key, &value));
+    return OgaString(value);
```
kunal-vaishnavi: To match existing APIs:

https://github.com/microsoft/onnxruntime-genai/blob/fd445627f2b23d886bc24387c12562479fcd2221/src/ort_genai.h#L244-L254

```suggestion
    return value;
```

src/ort_genai.h
```diff
@@ -906,5 +906,15 @@ struct OgaStreamingProcessor : OgaAbstract {
     return std::unique_ptr<OgaNamedTensors>(out);
   }
 
+  void SetOption(const char* key, const char* value) {
+    OgaCheckResult(OgaStreamingProcessorSetOption(this, key, value));
+  }
+
+  OgaString GetOption(const char* key) {
```
kunal-vaishnavi:
```suggestion
  OgaString GetOption(const char* key) const {
```

src/models/silero_vad.cpp
sayanshaw24: The fundamental challenge is that VAD is initialized in the StreamingProcessor constructor before any GeneratorParams exists, and State requires GeneratorParams. For the full State refactor, we'd need to either create lightweight GeneratorParams for VAD or defer VAD state creation until Generator is created.

src/models/streaming_processor.h
sayanshaw24: updated to shared pointer (non-owning shared reference, since Model's lifetime is managed elsewhere).

src/models/silero_vad.h
sayanshaw24: This is a non-owning pointer borrowed from `Model::p_device_inputs_->GetAllocator()`, which returns a reference. Using `unique_ptr` would imply ownership and cause a double-free on destruction. Can't use a reference either since it's set in the constructor body (not initializer list). Added a comment clarifying ownership.

src/config.h
sayanshaw24: The threshold is only for when VAD is enabled - it determines the sensitivity of VAD. If speech probability of an audio chunk is higher than the set threshold, we assume it has speech.

src/models/silero_vad.cpp
```diff
@@ -0,0 +1,179 @@
...
+  }
+
+  // Use VAD-specific sample_rate if configured, otherwise fall back to model's sample_rate
+  int sr = vad_config.sample_rate > 0 ? vad_config.sample_rate : model.config_->model.sample_rate;
+  Initialize(sr, vad_config.threshold);
+}
+
+SileroVad::~SileroVad() = default;
+
+void SileroVad::Reset() {
```
nenad1002: Is this needed or a new SileroVad can just be created?

src/models/streaming_processor.cpp
```diff
@@ -0,0 +1,90 @@
...
+  }
+
+  bool has_speech = vad_->ContainsSpeech(chunk_data, chunk_size);
+  if (has_speech) {
+    consecutive_silence_chunks_ = 0;
+    return false;
+  }
+
+  consecutive_silence_chunks_++;
+  if (consecutive_silence_chunks_ > min_silence_chunks_) {
```
nenad1002: Shouldn't it be >=?

src/models/silero_vad.cpp
sayanshaw24: good point, will remove!

src/models/streaming_processor.cpp
sayanshaw24: fixed!

src/models/silero_vad.h
sayanshaw24: removed need for this.

src/models/silero_vad.cpp
sayanshaw24: updated to use `State::Run` with dummy `GeneratorParams`.

examples/c/src/nemotron_speech.cpp
```diff
@@ -166,6 +166,17 @@ void StreamingTranscribe(const std::string& model_path, const std::string& audio
   auto config = OgaConfig::Create(model_path.c_str());
   auto model = OgaModel::Create(*config);
   auto processor = OgaStreamingProcessor::Create(*model);
+
+  // VAD is disabled by default. Enable via genai_config.json or --enable_vad.
```
kunal-vaishnavi: Since the presence of the VAD section in the GenAI config implies that it will be used, can we change this?

```suggestion
  // VAD is disabled by default. Enable via --enable_vad.
```

examples/c/src/nemotron_speech.cpp
```diff
@@ -166,6 +166,17 @@ void StreamingTranscribe(const std::string& model_path, const std::string& audio
...
+  if (enable_vad && std::string(processor->GetOption("vad_enabled")) != "true") {
```
kunal-vaishnavi: To match the above comment:

```suggestion
  if (!enable_vad) {
      processor->SetOption("vad_enabled", "false");
  }
```

examples/c/src/nemotron_speech.cpp
```diff
@@ -166,6 +166,17 @@ void StreamingTranscribe(const std::string& model_path, const std::string& audio
...
+    processor->SetOption("vad_enabled", "true");
+  }
+  if (std::string(processor->GetOption("vad_enabled")) == "true") {
```
kunal-vaishnavi: Can we remove the `if-else` conditionals and simplify to something like the following?

```suggestion
  auto use_vad = std::string(processor->GetOption("vad_enabled"));
  std::cout << "Use VAD: " << use_vad << std::endl;
```

examples/c/src/nemotron_speech.cpp
```diff
@@ -166,6 +166,17 @@ void StreamingTranscribe(const std::string& model_path, const std::string& audio
   auto config = OgaConfig::Create(model_path.c_str());
   auto model = OgaModel::Create(*config);
   auto processor = OgaStreamingProcessor::Create(*model);
+
+  // VAD is disabled by default. Enable via genai_config.json or --enable_vad.
+  if (enable_vad && std::string(processor->GetOption("vad_enabled")) != "true") {
+    processor->SetOption("vad_enabled", "true");
```
kunal-vaishnavi: Should we name it `"use_vad"` instead of `"vad_enabled"`?

examples/csharp/NemotronSpeech/Program.cs
```diff
@@ -29,6 +29,18 @@
...
 using var model = new Model(config);
 using var processor = new StreamingProcessor(model);
+
+// VAD is disabled by default. Enable via genai_config.json ("vad": {"enabled": true})
+// or programmatically:
+// processor.SetOption("vad_enabled", "true");
+// processor.SetOption("vad_threshold", "0.3");
+// processor.SetOption("silence_duration_ms", "500");
+// processor.SetOption("prefix_padding_ms", "300");
+if (processor.GetOption("vad_enabled") == "true")
```
kunal-vaishnavi: Same comment as with the C++ example for simplifying the `if-else` logic.

```csharp
var use_vad = processor.GetOption("vad_enabled");
Console.WriteLine("Use VAD: " + use_vad);
```

examples/csharp/NemotronSpeech/Program.cs
```diff
@@ -29,6 +29,18 @@
 using var config = Common.GetConfig(path: modelPath, ep: executionProvider, null, new GeneratorParamsArgs());
 using var model = new Model(config);
 using var processor = new StreamingProcessor(model);
+
+// VAD is disabled by default. Enable via genai_config.json ("vad": {"enabled": true})
```
kunal-vaishnavi: Same comment as with the C++ example for the comment.

```suggestion
// VAD is disabled by default. Enable programmatically via --enable_vad.
```

We should also enable it in the C# example via `--enable_vad` to match the C++ example.

examples/python/nemotron_speech.py
```diff
@@ -56,6 +57,24 @@ def simulate_microphone(model_path, audio_path, execution_provider):
...
     model = og.Model(config)
     processor = og.StreamingProcessor(model)
+
+    process = psutil.Process()
+
+    # VAD is controlled via genai_config.json (disabled by default).
+    # Override programmatically:
+    if enable_vad and processor.get_option("vad_enabled") != "true":
+        processor.set_option("vad_enabled", "true")
+    if vad_threshold != 0.5 and processor.get_option("vad_enabled") == "true":
```
kunal-vaishnavi: There seems to be some mismatch between the C++, C#, and Python examples in terms of showing which options to get and set on the streaming processor. Can we get them to be aligned? For example, we only show how to set the `vad_threshold` in the Python example based on a certain value.

examples/python/nemotron_speech.py
```diff
@@ -56,6 +57,24 @@ def simulate_microphone(model_path, audio_path, execution_provider):
...
+        processor.set_option("vad_threshold", str(vad_threshold))
+    if processor.get_option("vad_enabled") == "true":
+        actual_threshold = processor.get_option("vad_threshold")
+        print(f"  VAD: enabled (threshold={actual_threshold})")
```
kunal-vaishnavi: Let's also match the conditions used to print the logs and the printed logs themselves across the examples.

examples/python/nemotron_speech.py
```diff
@@ -56,6 +57,24 @@ def simulate_microphone(model_path, audio_path, execution_provider):
...
+    else:
+        print("  VAD: disabled")
+
+    # Measure memory after first chunk (VAD session is created lazily)
+    mem_before_vad = process.memory_info().rss
```
kunal-vaishnavi: Since we measure memory in the Python example, can we also do something similar in the C++ and C# examples? We want to keep feature parity across the examples.

If it is not simple to do, then let's remove it from here for now and move it to the benchmark tools, which can be used to measure memory usage.

examples/python/nemotron_speech.py
```diff
@@ -64,13 +83,29 @@ def simulate_microphone(model_path, audio_path, execution_provider):
...
+    vad_enabled = processor.get_option("vad_enabled") == "true"
+    chunks_total = 0
+    chunks_processed = 0
+    chunks_skipped = 0
+    vad_time_total = 0.0
 
     for i in range(0, len(audio), chunk_samples):
         chunk = audio[i:i + chunk_samples].astype(np.float32)
+        chunks_total += 1
+        chunk_start = time.time()
```
kunal-vaishnavi: Can we use `time.perf_counter()` to be more precise?

examples/python/nemotron_speech.py
```diff
@@ -84,20 +119,34 @@ def simulate_microphone(model_path, audio_path, execution_provider):
     print(f"  {full_transcript.strip()}")
     print(f"{'=' * 60}")
     print(f"  Audio: {duration:.2f}s | Wall: {total_wall:.2f}s | RTF: {duration/total_wall:.2f}x")
+    if vad_enabled:
+        avg_vad_ms = (vad_time_total / max(chunks_total, 1)) * 1000
+        chunk_duration_ms = (chunk_samples / sample_rate) * 1000
```
kunal-vaishnavi: Can we print these metrics in all of our examples to maintain feature parity?

src/models/nemotron_streaming_processor.cpp
```diff
@@ -35,6 +35,9 @@ NemotronStreamingProcessor::NemotronStreamingProcessor(Model& model)
   mel_pre_encode_cache_.assign(
       static_cast<size_t>(cache_config_.pre_encode_cache_size) * cache_config_.num_mels, 0.0f);
   cache_pos_ = 0;
+
+  // Initialize VAD from config (disabled by default)
```
kunal-vaishnavi: Since disabling is handled in the inference examples by setting the option:

```suggestion
  // Initialize VAD from config
```

src/models/silero_vad.cpp
```diff
@@ -0,0 +1,190 @@
...
+#include <cstring>
+#include <stdexcept>
+#include <vector>
+
+#include "../generators.h"
+#include "silero_vad.h"
+
+namespace Generators {
+
+// --- SileroVadState: proper State subclass for VAD inference ---
```
kunal-vaishnavi:
```suggestion
```

src/models/silero_vad.cpp
```diff
@@ -0,0 +1,190 @@
...
+
+  // Register outputs (ORT allocates)
+  output_names_.push_back("output");
+  outputs_.push_back(nullptr);
+  output_names_.push_back("stateN");
+  outputs_.push_back(nullptr);
+
+  // Apply run options from config if specified
+  if (model.config_->model.vad.run_options.has_value()) {
+    SetRunOptionsFromConfig(model.config_->model.vad.run_options.value());
```
kunal-vaishnavi: We can call the following line directly here instead of creating another wrapper function.

https://github.com/microsoft/onnxruntime-genai/blob/df72b3f108886909ed77add1dbcdc1552a7c582a/src/models/silero_vad.cpp#L63

src/models/silero_vad.cpp
```diff
@@ -0,0 +1,190 @@
...
+  // Use State::Run which handles run options, session terminate, EP options, graph capture
+  State::Run(vad_session_);
+  return {};
+}
+
+void SileroVadState::SetRunOptionsFromConfig(const Config::RunOptions& config_run_options) {
+  State::SetRunOptions(config_run_options);
+}
+
+// --- SileroVad: main wrapper ---
```
kunal-vaishnavi:
```suggestion
```

src/models/silero_vad.cpp
```diff
@@ -0,0 +1,190 @@
...
+
+void SileroVad::Initialize(int sample_rate, float threshold) {
+  sample_rate_ = sample_rate;
+  threshold_ = threshold;
+
+  if (sample_rate != 16000 && sample_rate != 8000) {
+    throw std::runtime_error("SileroVad only supports sample rates 16000 and 8000. Got: " +
```
kunal-vaishnavi:
```suggestion
    throw std::runtime_error("SileroVad only supports a sample rate of 8,000 Hz or 16,000 Hz. Got: " +
```

src/models/silero_vad.cpp
```diff
@@ -0,0 +1,190 @@
...
+                             std::to_string(sample_rate));
+  }
+
+  const int factor = sample_rate / 8000;
+  window_size_ = 256 * factor;
```
kunal-vaishnavi, line 72: For these three lines, can you explain what the constants in the expressions represent and how they are calculated/determined? The reasons are not immediately clear to the reader at first glance.

src/models/silero_vad.h
```diff
@@ -0,0 +1,93 @@
...
+///            stateN [2, 1, 128]                      float32 (updated hidden state)
+struct SileroVad {
+  SileroVad(Model& model);
+  ~SileroVad();
+
+  float ProcessWindow(const float* samples, size_t num_samples);
+  bool ContainsSpeech(const float* samples, size_t num_samples);
+
+  float GetThreshold() const { return threshold_; }
+  void SetThreshold(float threshold) {
```
kunal-vaishnavi, line 39: There's a simpler way to clamp without needing `if` branches.

```suggestion
  void SetThreshold(float threshold) {
    threshold_ = std::max(0.0f, std::min(1.0f, threshold));  // clamp threshold between [0.0, 1.0]
  }
```

src/models/silero_vad.h
```diff
@@ -0,0 +1,93 @@
...
+ private:
+  void Initialize(int sample_rate, float threshold);
+  void EnsureState();  // Lazily creates the State on first use
+
+  Model& model_;
+  std::unique_ptr<OrtSession> session_;
+  std::unique_ptr<OrtSessionOptions> session_options_;
+
+  // State is created lazily on first ProcessWindow call with internal GeneratorParams
+  std::shared_ptr<GeneratorParams> params_;
```
kunal-vaishnavi, line 56: Baiju and I discussed this. We should refactor `State` to have a base implementation that does not require inputs and then construct a `GeneratorState` for a state that does. Then, we won't need a dummy object for the generator params here.

src/models/silero_vad.cpp
```diff
@@ -0,0 +1,190 @@
...
+  state_data_.assign(kStateSize, 0.0f);
+  context_.assign(static_cast<size_t>(context_size_), 0.0f);
+
+  const int effective_size = context_size_ + window_size_;
+  input_data_.resize(static_cast<size_t>(effective_size), 0.0f);
+  sr_value_ = static_cast<int64_t>(sample_rate_);
+}
+
+SileroVad::SileroVad(Model& model)
+    : model_{model} {
```
kunal-vaishnavi, line 83: Can we convert this into `SileroVadModel` where `SileroVadModel` inherits from `Model`? Here is some pseudocode to mimic how the other files create a `Model` and a `State`.

```cpp
SileroVadModel::SileroVadModel(std::unique_ptr<Config> config, OrtEnv& ort_env)
    : Model{std::move(config)} {
  session_options_ = OrtSessionOptions::Create();
  CreateSessionOptionsFromConfig(config_->model.vad.session_options.has_value() ? config_->model.vad.session_options.value() : config_->model.decoder.session_options, *session_options_, true);

  session_ = CreateSession(ort_env, config_->model.vad.filename, session_options_.get());
  session_info_.Add(*session_);
}

std::unique_ptr<State> SileroVadModel::CreateState(DeviceSpan<int32_t> sequence_lengths) const {
  return std::make_unique<SileroVadState>(*this, sequence_lengths);
}
```

This would require a change in the base `Model` class to make a `CreateState` that takes in the desired parameters.

https://github.com/microsoft/onnxruntime-genai/blob/935d4267453cfdd32b6eae6df10a632066ab0132/src/models/model.h#L145-L153

src/models/silero_vad.cpp
```diff
@@ -0,0 +1,190 @@
...
+  model.CreateSessionOptionsFromConfig(
+      vad_config.session_options.has_value()
+          ? vad_config.session_options.value()
+          : model.config_->model.decoder.session_options,
+      *session_options_, false, true);
+
+  // Load session through Model::CreateSession
+  std::string filename = vad_config.filename;
+  if (filename.empty()) {
+    throw std::runtime_error("VAD filename must be specified in genai_config.json when vad.enabled is true.");
```
kunal-vaishnavi:
```suggestion
    throw std::runtime_error("VAD filename must be specified in genai_config.json");
```

src/models/silero_vad.cpp
```diff
@@ -0,0 +1,190 @@
...
+  }
+  session_ = model.CreateSession(GetOrtEnv(), filename, session_options_.get());
+
+  int sr = vad_config.sample_rate > 0 ? vad_config.sample_rate : model.config_->model.sample_rate;
+  Initialize(sr, vad_config.threshold);
```
kunal-vaishnavi: Can we always use the sampling rate from the `model` section since they have to align? Then, it doesn't have to be specified twice.

```suggestion
  Initialize(model.config_->model.sample_rate, vad_config.threshold);
```

src/models/silero_vad.h
```diff
@@ -0,0 +1,93 @@
...
+  std::unique_ptr<SileroVadState> state_;
+
+  float threshold_;
+  int sample_rate_;
```
kunal-vaishnavi: Why are there two values for sample rate: `sample_rate_` and `sr_value_`?

src/models/silero_vad.h
```diff
@@ -0,0 +1,93 @@
...
+                 float* input_data, int input_size,
+                 float* state_data, int64_t* sr_value);
+
+  DeviceSpan<float> Run(int total_length, DeviceSpan<int32_t>& next_tokens,
+                        DeviceSpan<int32_t> next_indices = {}) override;
+
+  void SetRunOptionsFromConfig(const Config::RunOptions& run_options);
+
+ private:
+  OrtSession& vad_session_;
```
kunal-vaishnavi, line 80: If we create the `SileroVadModel` object, the `Model` object can then own the session instead of having to define the session here.

src/models/streaming_processor.cpp
```diff
@@ -0,0 +1,127 @@
...
+    if (vad_) {
+      vad_->SetThreshold(threshold);
+    }
+  } else if (k == "silence_duration_ms") {
+    int ms = std::stoi(value);
+    if (model_) {
+      float chunk_duration_ms = (static_cast<float>(model_->config_->model.chunk_samples) /
+                                 model_->config_->model.sample_rate) * 1000.0f;
+      silence_duration_chunks_ = std::max(1, static_cast<int>(ms / chunk_duration_ms));
+    }
```
kunal-vaishnavi: The `silence_duration_ms` and `prefix_padding_ms` cases are identical except for the final destination where the value is saved. Can the two cases be combined as follows?

```cpp
else if (k == "silence_duration_ms" || k == "prefix_padding_ms") {
  int ms = std::stoi(value);
  if (model_) {
    float chunk_duration_ms = (static_cast<float>(model_->config_->model.chunk_samples) /
                               model_->config_->model.sample_rate) * 1000.0f;
    int num_chunks = std::max(1, static_cast<int>(ms / chunk_duration_ms));
    if (k == "silence_duration_ms") {
      silence_duration_chunks_ = num_chunks;
    } else {
      prefix_padding_chunks_ = num_chunks;
    }
  }
}
```

src/models/streaming_processor.cpp
```diff
@@ -0,0 +1,127 @@
...
+  } else if (k == "silence_duration_ms") {
+    if (model_) {
+      float chunk_duration_ms = (static_cast<float>(model_->config_->model.chunk_samples) /
+                                 model_->config_->model.sample_rate) * 1000.0f;
+      return std::to_string(static_cast<int>(silence_duration_chunks_ * chunk_duration_ms));
+    }
+    return "500";
+  } else if (k == "prefix_padding_ms") {
+    if (model_) {
+      float chunk_duration_ms = (static_cast<float>(model_->config_->model.chunk_samples) /
```
kunal-vaishnavi: Same comment here about merging the two `else if` conditions

src/models/streaming_processor.cpp
```diff
@@ -0,0 +1,127 @@
...
+  }
+}
+
+std::string StreamingProcessor::GetOption(const char* key) const {
+  std::string_view k{key};
+
+  if (k == "vad_enabled") {
+    return vad_ ? "true" : "false";
+  } else if (k == "vad_threshold") {
+    return std::to_string(vad_ ? vad_->GetThreshold() : 0.5f);
```
kunal-vaishnavi: Instead of storing a default value here, can we read from the config's default value (or somehow read it from `vad_` even if the VAD is not enabled)? The config already has a default value of 0.5 in it.

https://github.com/microsoft/onnxruntime-genai/blob/df72b3f108886909ed77add1dbcdc1552a7c582a/src/config.h#L280

src/models/silero_vad.h
```diff
@@ -0,0 +1,93 @@
...
+  std::unique_ptr<OrtSessionOptions> session_options_;
+
+  // State is created lazily on first ProcessWindow call with internal GeneratorParams
+  std::shared_ptr<GeneratorParams> params_;
+  std::unique_ptr<SileroVadState> state_;
+
+  float threshold_;
+  int sample_rate_;
+  int window_size_;
+  int context_size_;
```
kunal-vaishnavi: Can these variables be initialized with curly braces? For example:

```suggestion
  int context_size_{};
```

examples/python/nemotron_speech.py
sayanshaw24: Removed from Python (not easy in C++/C#)

src/models/silero_vad.h
sayanshaw24: Deferring to separate PR as discussed.

src/models/silero_vad.cpp
sayanshaw24: Deferred (requires Model base class changes)

src/models/silero_vad.h
sayanshaw24: Added a comment: `sr_value_` is int64 copy needed by ORT tensor

src/models/silero_vad.h
sayanshaw24: Deferred as per previous related comments.

PR #2036, opened 2026-03-18
Support 4 inputs for nemotron model

jiafatom:
We want to make the encoder input `length` optional, because this can help constant folding a lot of operators at the beginning. The value of `length` is equal to the 2nd dim value of `audio_signal`.
name: audio_signal
tensor: float32[1,65,128]

Manually tested in `test/python/test_onnxruntime_genai_e2e.py` using a 4-input nemotron model, and it works well.

We also remove `target_length_orig` and `target_length` from decoder graph, and here we support it in genai.

### DIFF
--- a/src/models/nemotron_speech.h
+++ b/src/models/nemotron_speech.h
@@ -56,11 +56,9 @@ struct NemotronCacheConfig {
 
   // Decoder (prediction network) I/O names
   std::string dec_in_targets;
-  std::string dec_in_target_length;
   std::string dec_in_lstm_hidden;
   std::string dec_in_lstm_cell;
   std::string dec_out_outputs;
-  std::string dec_out_prednet_lengths;
   std::string dec_out_lstm_hidden;
   std::string dec_out_lstm_cell;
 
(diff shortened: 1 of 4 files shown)


jiafatom:
> Why optional? Shouldn't we just remove it from config?

We want to support both 4 inputs (static shape input) and 5 inputs case (dynamic shape input).

src/models/nemotron_speech.h
```diff
@@ -154,10 +155,8 @@ struct NemotronPredictionSubState : State {
   const NemotronSpeechModel& model_;
   NemotronDecoderState lstm_state_;
   std::unique_ptr<OrtValue> targets_;
-  std::unique_ptr<OrtValue> target_length_;
```
nenad1002, line 157: is has_length_input is true, this isn't used?

src/models/nemotron_speech.h
```diff
@@ -130,6 +128,9 @@ struct NemotronEncoderSubState : State {
   NemotronEncoderCache cache_;
   std::unique_ptr<OrtValue> signal_length_;
 
+  // Whether the encoder model has a "length" input
+  bool has_length_input_{};
```
nenad1002, line 132: is this set in genai_config?

nenad1002 (approved):

src/models/nemotron_speech.h
kunal-vaishnavi: We don't need to set it. If the input is in the `inputs` section of the GenAI config, then it exists and is used.

src/models/nemotron_speech.h
jiafatom: This is for decoder output `target_length`. `has_length_input` is for encoder input `length`

PR #2038, opened 2026-03-19
Decouple plugin execution providers (EPs) from the USE_WINML pre-processor macro

baijumeswani:
Previously, plugin execution providers were restricted to the WinML package because they relied on a build-time macro. This PR replaces that static check with dynamic discovery. The system now verifies if a requested provider name exists within the ONNX Runtime EP device list:

- If found: It utilizes the V2 API for appending the execution provider.
- If not found: It falls back to the provider-bridge logic using the legacy API.

| Execution Provider | Status |
| :--- | :--- |
| **CPU** | 🟢 Pass |
| **CUDA** | 🟢 Pass |
| **WebGPU** | 🟢 Pass |
| **NvTensorRtRtx** | 🟢 Pass |
| **QNN** | 🟢 Pass |
| **OpenVINO** | 🟢 Pass |
| **VitisAI** | 🟢 Pass |

### DIFF
--- a/src/java/build.gradle
+++ b/src/java/build.gradle
@@ -39,6 +39,9 @@ def mavenUser = System.properties['mavenUser']
 def mavenPwd = System.properties['mavenPwd']
 def mavenArtifactId = useCUDA == null ? project.name : project.name + "_gpu"
 
+def onnxruntimeArtifactId = useCUDA == null ? "onnxruntime" : "onnxruntime_gpu"
+def onnxruntimeVersion = System.properties["ONNXRUNTIME_VERSION"] ?: "1.23.0"
+
 def defaultDescription = 'ONNX Runtime GenAI is <TODO>'
 
 logger.lifecycle("cmakeBuildDir:${cmakeBuildDir}")
@@ -160,6 +163,7 @@ if (cmakeNativeLibDir != null) {
 }
 
 dependencies {
+	implementation "com.microsoft.onnxruntime:${onnxruntimeArtifactId}:${onnxruntimeVersion}"
 	testImplementation 'org.junit.jupiter:junit-jupiter-api:5.9.2'
 	testImplementation 'org.junit.platform:junit-platform-launcher:1.10.1'
 	testRuntimeOnly 'org.junit.jupiter:junit-jupiter-engine:5.9.2'


src/models/model.cpp
```diff
@@ -709,168 +762,95 @@ DeviceInterface* SetProviderSessionOptions(OrtSessionOptions& session_options,
...
-          }
-          fs::path custom_ops_lib_path(lib_name);
-          session_options.RegisterCustomOpsLibrary(custom_ops_lib_path.c_str());
+      if (const auto opt_it = std::find_if(provider_options.options.begin(), provider_options.options.end(),
+                                           [](const auto& pair) { return pair.first == "external_ep_libray"; });
+          opt_it != provider_options.options.end()) {
+        auto lib_name = opt_it->second;
+        auto lib = LoadLibrary(lib_name.c_str());
+        if (const auto func = (void (*)(void*, const OrtApiBase*, void*, OrtEpFactory**, size_t, size_t*))GetProcAddress(lib, "CreateEpFactories")) {
+          OrtEpFactory* factory = nullptr;
```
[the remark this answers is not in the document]
baijumeswani: I will ask original contributors to address this. Seems like a bug.

src/models/model.cpp
```diff
@@ -709,168 +762,95 @@ DeviceInterface* SetProviderSessionOptions(OrtSessionOptions& session_options,
...
-            OrtEpFactory* factory = nullptr;
-            size_t num = 1;
-
-            func(nullptr, OrtGetApiBase(), nullptr, &factory, num, &num);
-          }
-          fs::path custom_ops_lib_path(lib_name);
-          session_options.RegisterCustomOpsLibrary(custom_ops_lib_path.c_str());
+      if (const auto opt_it = std::find_if(provider_options.options.begin(), provider_options.options.end(),
+                                           [](const auto& pair) { return pair.first == "external_ep_libray"; });
+          opt_it != provider_options.options.end()) {
```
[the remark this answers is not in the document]
baijumeswani: I will ask original contributors to address this. Seems like a bug.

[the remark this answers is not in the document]
kunal-vaishnavi: This is actually a typo that was introduced inside ORT and had to be propagated inside ORT GenAI for WinML scenarios. There are open PRs in both repos to fix this in a future release.

src/models/model.cpp
```diff
@@ -392,161 +393,92 @@ int32_t Tokenizer::TokenToTokenId(const char* token) const {
...
-
-      shapes << "," << key_name << ":" << batch_size << "x" << num_kv_heads << "x" << seq_len << "x" << head_dim;
-      shapes << "," << value_name << ":" << batch_size << "x" << num_kv_heads << "x" << seq_len << "x" << head_dim;
+namespace {
+
+std::vector<const OrtEpDevice*> ApplyDeviceFiltering(const Config::ProviderOptions& provider_options,
+                                                     const std::vector<const OrtEpDevice*>& devices) {
+  if (!provider_options.device_filtering_options) {
+    return devices;
+  } else if (!provider_options.device_filtering_options->hardware_device_id &&
```
kunal-vaishnavi: Can we combine these conditions since they have the same return value?

src/models/model.cpp
```diff
@@ -555,17 +487,34 @@ bool IsProviderRegistered(
 
   if (is_primary_session_options) {
```
kunal-vaishnavi: Why do we need to check for `is_primary_session_options`?

src/models/model.cpp
```diff
@@ -555,17 +487,34 @@ bool IsProviderRegistered(
...
+  }
 
-  std::vector<const OrtEpDevice*> ep_devices_ptrs = {ep_device};
-  session_options.AppendExecutionProvider_V2(GetOrtEnv(), ep_devices_ptrs, options);
-  return true;  // Handled
+  session_options.AppendExecutionProvider_V2(GetOrtEnv(), filtered_ep_device_ptrs, options);
+  return true;
+}
+
+void AppendGenericExecutionProvider(
```
kunal-vaishnavi: Can we rename this to something more specific than `generic`? The method above makes it clear that the type of EP being appended is a plug-in EP.

src/models/model.cpp
```diff
@@ -599,52 +548,15 @@ DeviceInterface* SetProviderSessionOptions(OrtSessionOptions& session_options,
...
+                                        DeviceType::CUDA, "CUDAExecutionProvider",
+                                        is_primary_session_options, p_device)) {
+        continue;
+      } else {
+        // Register the CUDA execution provider as a provider-bridge provider.
+        Generators::CUDAExecutionProvider::AppendExecutionProvider(session_options, provider_options,
+                                                                   is_primary_session_options, p_device,
+                                                                   arena_cfg);
       }
     } else if (provider_options.name == "rocm") {
```
kunal-vaishnavi: Can we remove any references to `ROCm` since ORT has deprecated it?

src/models/model.cpp
```diff
@@ -703,174 +615,105 @@ DeviceInterface* SetProviderSessionOptions(OrtSessionOptions& session_options,
     } else if (provider_options.name == "NvTensorRtRtx") {
```
kunal-vaishnavi: Can we clean up the DML section that comes before this case to look like the subsequent OpenVINO case?

https://github.com/microsoft/onnxruntime-genai/blob/d53bcd00943bbae244f249ca4a1c39364ab11025/src/models/model.cpp#L572-L609

src/models/model.cpp
```diff
@@ -703,174 +615,105 @@ DeviceInterface* SetProviderSessionOptions(OrtSessionOptions& session_options,
...
       // Configure NvTensorRT-specific settings (needed for both pre-registered and built-in paths)
       bool is_multi_profile_enabled = IsMultiProfileEnabled(config.model.decoder.session_options);
-      ConfigureNvTensorRtRtxProfile(config, session_options, is_multi_profile_enabled);
+      Generators::NvTensorRtRtxExecutionProvider::ConfigureProfile(
+          config, session_options, is_multi_profile_enabled);
       if (IsGraphCaptureEnabled(config.model.decoder.session_options)) {
         session_options.AddConfigEntry("ep.nvtensorrtrtxexecutionprovider.enable_cuda_graph", "1");
```
kunal-vaishnavi: Should this be happening inside `session_options.cpp`?

src/models/model.cpp
baijumeswani: Because, we need to return nullptr if these settings are not for the primary session. The device returned here should only be for the primary session.

src/models/model.cpp
```diff
@@ -703,174 +615,105 @@ DeviceInterface* SetProviderSessionOptions(OrtSessionOptions& session_options,
...
+        AppendGenericExecutionProvider(session_options, provider_options);
+      }
+    } else if (provider_options.name == "WebGPU") {
+      if (AppendPluginExecutionProvider(session_options, provider_options,
+                                        DeviceType::WEBGPU, "WebGpuExecutionProvider",
+                                        is_primary_session_options, p_device)) {
+        continue;
+      } else {
+        if (is_primary_session_options) {
+          p_device = GetDeviceInterface(DeviceType::WEBGPU);
```
kunal-vaishnavi: This case to assign a value to `p_device` if `is_primary_session_options == true` seems to keep occurring for multiple EPs (DML, TRT-RTX, WebGPU). Is there any additional shared refactoring we can do?

src/models/model.cpp
```diff
@@ -703,174 +615,105 @@ DeviceInterface* SetProviderSessionOptions(OrtSessionOptions& session_options,
...
-          // WinML Hotfix: DML and WebGPU EP factories currently only support one device at a time
-          if (provider_options.name == "DML" || provider_options.name == "WebGPU") {
-            break;
+      } else {
+        // Built-in path: Configure stream via config entry (generic AppendExecutionProvider will be called below)
+        if (is_primary_session_options) {
+          p_device = GetDeviceInterface(DeviceType::NvTensorRtRtx);
+          if (p_device) {
+            void* stream_ptr = p_device->GetCudaStream();
+            std::stringstream stream_value;
```
kunal-vaishnavi: This section looks to be very similar as the below. Can we refactor these?

https://github.com/microsoft/onnxruntime-genai/blob/d53bcd00943bbae244f249ca4a1c39364ab11025/src/models/model.cpp#L488-L497

src/models/model.cpp
baijumeswani: I can rename to `AppendProviderBridgeExecutionProvider`. But the naming is not ideal. Even the Plugin naming is misleading because it may not always be the case that the provider was plugged in to use with v2 api. But it was the way I thought of distinguishing the two functions.
Will rename to something more comprehensible.

src/models/model.cpp
baijumeswani: There might be some interest in re-introducing rocm in the future from amd's end. I will leave it here for now since this PR does not target removing existing functionality. Moreover, cleaning up rocm might be a slightly bigger change (since it might have some package related settings as well).

src/models/model.cpp
baijumeswani: IsGraphCaptureEnabled is defined in this file. I think it is ok for these settings to be here. I need to do a bigger change to clean up the entire session options creation so that this file can focus only on model related code. Will do this in a subsequent pull-request.

src/models/model.cpp
```diff
@@ -703,174 +615,105 @@ DeviceInterface* SetProviderSessionOptions(OrtSessionOptions& session_options,
...
+        auto lib_name = opt_it->second;
+        auto lib = LoadLibrary(lib_name.c_str());
+        if (const auto func = (void (*)(void*, const OrtApiBase*, void*, OrtEpFactory**, size_t, size_t*))GetProcAddress(lib, "CreateEpFactories")) {
+          OrtEpFactory* factory = nullptr;
+          size_t num = 1;
+
+          func(nullptr, OrtGetApiBase(), nullptr, &factory, num, &num);
         }
+        fs::path custom_ops_lib_path(lib_name);
+        session_options.RegisterCustomOpsLibrary(custom_ops_lib_path.c_str());
```
kunal-vaishnavi: Should we move all of the content in the `if defined` for VitisAI into `session_options.cpp` and use something like `Generators::VitisAI::RegisterCustomOps` to call it and mimic how a TRT-RTX profile is configured?

src/models/model.cpp
baijumeswani: yes, we can remove this check if the calling function does not accept the p_device returned. But I would like to avoid needing to even `GetDeviceInterface` if it is not needed. Which is why this check may be better off here.

Not all providers require this. For example, providers that allocate directly on cpu (qnn, cpu, vitisai) do not need to set the p_device. Since p_device is assumed to be cpu for these eps.

Can try to clean this up.

src/models/model.cpp
baijumeswani: Yes, can do this.

Overall, this library registration code is not written well. It leaks the library since it is never closed. There is no error handling either. I need to understand from AMD why it is needed here and what purpose it solves.

src/models/model.cpp
```diff
@@ -932,7 +450,26 @@ void EnsureDeviceOrtInit(DeviceInterface& device, const Config& config, std::uni
   // Get the allocator from the OrtSession for the DeviceType (it's called 'AllocatorCreate' but it's really 'AllocatorGet')
   auto name = device_memory_type_names[static_cast<int>(type)];
   auto memory_info = OrtMemoryInfo::Create(name, OrtAllocatorType::OrtDeviceAllocator, 0, OrtMemType::OrtMemTypeDefault);
-  allocator.allocator_ = Ort::Allocator::Create(*allocator.session_, *memory_info);
+  try {
+    allocator.allocator_ = Ort::Allocator::Create(*allocator.session_, *memory_info);
+  } catch (const Ort::Exception& e) {
+    // WebGPU memory type name changed from "WebGPU_Buffer" to "WebGPU_Buf" in ORT 1.24.3.
```
kunal-vaishnavi, line 457: Should we add a deprecation notice for supporting the old name and then we can remove support for the old name in a future release?

kunal-vaishnavi (dismissed):

src/models/model.cpp
baijumeswani: We don't need to add a deprecation notice since this is internal to ort-genai. This change went in here: https://github.com/microsoft/onnxruntime-genai/pull/1957

But the change is not compatible with ort < 1.24.3. So, FL winml (which ships ort 1.23.2) still needs the old name. Which is what this change does.

Issue #2030, opened 2026-03-16
[Java] Add explicit Maven dependency on `onnxruntime` to prevent code duplication in native library loading


EPNW-Eric:
Following a recent discussion in microsoft/onnxruntime#27656 a question came up regarding the Java module: why doesn't `onnxruntime-genai` declare an explicit Maven dependency on the core `onnxruntime` artifact?

In the current implementation, [`GenAI.java`](https://github.com/microsoft/onnxruntime-genai/blob/639dcc06fbffff14c06612b64d3f7773a7b98a2f/src/java/src/main/java/ai/onnxruntime/genai/GenAI.java) attempts to load the native libraries for both `onnxruntime-genai` and core `onnxruntime` directly from the classpath. However, without an explicit Maven dependency, there is no guarantee that the core `onnxruntime` libraries will actually be present on the classpath.

Furthermore, by having [`GenAI.java`](https://github.com/microsoft/onnxruntime-genai/blob/639dcc06fbffff14c06612b64d3f7773a7b98a2f/src/java/src/main/java/ai/onnxruntime/genai/GenAI.java) handle the native library loading for both, we are effectively duplicating the complex loading logic already established in `onnxruntime` (specifically in [`OrtRuntime.java`](https://github.com/microsoft/onnxruntime/blob/98a5495b854d80c12f8c146aff5b1b72a158addf/java/src/main/java/ai/onnxruntime/OnnxRuntime.java)). This creates an ongoing maintenance burden, as any improvements or bug fixes to the core library loader must be manually synced over to `onnxruntime-genai`. For example, we are currently out of sync, since e.g. `onnxruntime_providers_shared` [isn't extracted](https://github.com/microsoft/onnxruntime-genai/blob/639dcc06fbffff14c06612b64d3f7773a7b98a2f/src/java/src/main/java/ai/onnxruntime/genai/GenAI.java#L35). Adding an explicit dependency would ensure a predictable classpath and open the door to deduplicating this loading logic.

What are your thoughts on this?

PR #2040, opened 2026-03-20
[Java] Add a dependency on onnxruntime (#2030)

EPNW-Eric:
This PR addresses the concern raised in #2030 (and before in microsoft/onnxruntime#27656) regarding the missing dependency between `onnxruntime-genai` and the core `onnxruntime` Java artifact.

Currently, `GenAI.java` assumes that the native ONNX Runtime libraries are available on the classpath and attempts to load them directly. However, without an explicit Maven dependency, this is not guaranteed and can lead to runtime issues.

Additionally, this setup duplicates native library loading logic that already exists in `OrtRuntime.java`, creating unnecessary maintenance overhead and inconsistencies (e.g., missing extraction of `onnxruntime_providers_shared`).

Other ecosystem packages already declare this dependency:

* The NuGet package includes a [dependency on ONNX Runtime](https://www.nuget.org/packages/Microsoft.ML.OnnxRuntimeGenAI)
* The PyPI package does as well (see the `METADATA` file in a [onnxruntime-genai wheel](https://pypi.org/project/onnxruntime-genai))

Aligning the Java package with this behavior ensures consistency across platforms and provides a more predictable and reliable setup for users.

Adding this dependency makes the expectation explicit and lays the groundwork for potentially reusing the existing native loading logic in the future.

### DIFF
--- a/src/java/build.gradle
+++ b/src/java/build.gradle
@@ -30,7 +30,7 @@ def mavenPwd = System.properties['mavenPwd']
 def mavenArtifactId = useCUDA == null ? project.name : project.name + "_gpu"
 
 def onnxruntimeArtifactId = useCUDA == null ? "onnxruntime" : "onnxruntime_gpu"
-def onnxruntimeVersion = System.properties["ONNXRUNTIME_VERSION"] ?: "1.24.3"
+def onnxruntimeVersion = System.properties["ONNXRUNTIME_VERSION"] ?: "1.23.0"
 
 def defaultDescription = 'ONNX Runtime GenAI is <TODO>'
 


src/java/build.gradle
```diff
@@ -29,6 +29,9 @@ def mavenUser = System.properties['mavenUser']
 def mavenPwd = System.properties['mavenPwd']
 def mavenArtifactId = useCUDA == null ? project.name : project.name + "_gpu"
 
+def onnxruntimeArtifactId = useCUDA == null ? "onnxruntime" : "onnxruntime_gpu"
+def onnxruntimeVersion = "1.24.3"
```
baijumeswani: Is it possible to override this from outside? I would like to build ort-genai packaging that can pass in the onnxruntime dependency version so we don't need to keep changing this field from time to time.

EPNW-Eric: Good idea! Let’s make it [analogous to Python](https://github.com/microsoft/onnxruntime-genai/blob/80dd558361931a967b8e818a3592dc00857e9241/src/python/setup.py.in#L42).

src/java/build.gradle
```diff
@@ -29,6 +29,9 @@ def mavenUser = System.properties['mavenUser']
 def mavenPwd = System.properties['mavenPwd']
 def mavenArtifactId = useCUDA == null ? project.name : project.name + "_gpu"
 
+def onnxruntimeArtifactId = useCUDA == null ? "onnxruntime" : "onnxruntime_gpu"
+def onnxruntimeVersion = System.properties["ONNXRUNTIME_VERSION"] ?: "1.24.3"
```
kunal-vaishnavi: Can we maintain parity for the default ORT versions used when building from source across languages? This uses 1.24.3 but Python uses 1.23.0 instead.

PR #2042, opened 2026-03-22
Rename NemotronCacheConfig to NemotronConfig and add blank penalty to the decoder

nenad1002:


src/models/nemotron_speech.cpp
```diff
@@ -39,6 +39,7 @@ void NemotronCacheConfig::PopulateFromConfig(const Config& config) {
   chunk_samples = config.model.chunk_samples;
   blank_id = config.model.blank_id;
   max_symbols_per_step = config.model.max_symbols_per_step;
+  blank_penalty = config.model.blank_penalty;
```
kunal-vaishnavi: Blank penalty isn't a model-specific attribute. It would be better to add it to the `search` section of the GenAI config. There can be a default value in the config so that it can be optionally included in the `search` section (so that existing GenAI configs do not break).

kunal-vaishnavi (approved):

PR #2043, opened 2026-03-24
Add Qwen3.5 hybrid decoder export support (GatedDeltaNet + Attention)

apsonawane:
This PR adds support for exporting Qwen3.5 hybrid decoder models to ONNX via the
onnxruntime-genai model builder. Qwen3.5 uses a novel hybrid architecture that
alternates between standard attention layers (with KV cache) and GatedDeltaNet
linear attention layers (with conv_state + recurrent_state).

## Architecture

Qwen3.5 has two layer types controlled by `full_attention_interval`:
- **Full attention** (every 4th layer): Doubled Q projection (Q + output gate),
  per-head QK OffsetRMSNorm, partial interleaved mRoPE, opset 23 Attention op
- **Linear attention** (remaining layers): GatedDeltaNet with depthwise causal
  conv1d, L2-normalized Q/K, Scan-based recurrence, gated RMSNorm

## Key design decisions

- **Opset 23** standard-domain ops (`RMSNormalization`, `RotaryEmbedding`,
  `Attention`) instead of `com.microsoft` contrib ops for portability
- **CumSum-based causal mask** with `[B, 1, S, total_S]` shape — correctly
  handles both prefill and single-token decode modes
- **Pre-computed cos/sin cache tables** with Where-mask interleaving for
  efficient mRoPE at runtime
- **Hybrid cache I/O**: Per-layer KV cache for attention layers,
  conv_state + recurrent_state for GatedDeltaNet layers
- **SkipSimplifiedLayerNormalization** with pre-baked `weight + 1` offset
  (reuses base class)

### `src/python/py/models/builders/qwen.py`
- Added `Qwen35Model` class with:
  - `_make_full_attention_layer`: Q/gate split, QK norm, mRoPE, Attention, gating
  - `_make_linear_attention_layer`: Conv1d, L2 norm, Scan recurrence, gated norm
  - `_make_causal_mask`: CumSum-based mask shared across attention layers
  - `_make_rotary_caches`: Pre-computed cos/sin + h/w interleaving masks
  - `_build_scan_body`: GatedDeltaNet delta rule in ONNX Scan body graph
  - `make_genai_config`: Text-only config (reuses base class via `super()`)

### `src/python/py/models/builders/__init__.py`
- Export `Qwen35Model`

### `src/python/py/models/builder.py`
- Dispatch `Qwen3_5ForConditionalGeneration` → `Qwen35Model`

## Export command

```bash
cd src/python/py/models
python builder.py -m "Qwen/Qwen3.5-0.8B" -o <output_dir> -p fp32 -e cpu
python builder.py -m "Qwen/Qwen3.5-0.8B" -o <output_dir> -p int4 -e cpu

### DIFF
--- a/src/python/py/models/builders/base.py
+++ b/src/python/py/models/builders/base.py
@@ -654,10 +654,7 @@ class Model:
 
         if self.ep != "cpu":
             ep_name = self.ep.replace("trt-rtx", "NvTensorRtRtx")
-            ep_settings = dict(self.ep_attrs[self.ep])
-            if self.ep == "dml":
-                ep_settings["enable_graph_capture"] = "0"
-            ep_options = {ep_name: ep_settings}
+            ep_options = {ep_name: self.ep_attrs[self.ep]}
             genai_config["model"]["decoder"]["session_options"]["provider_options"].append(ep_options)
 
         print(f"Saving GenAI config in {out_dir}")

### CODE REVIEW

File: src/python/py/models/builders/qwen.py [hunk 1]
```diff
@@ -701,7 +710,7 @@
         # Fix model_type: HF architecture "Qwen3VLForConditionalGeneration" would produce "qwen3vl"
         # but the C++ runtime expects "qwen3_vl" (with underscore).
         # Intentional override of the superclass attribute (used in genai_config.json).
-        self.model_type = "Qwen3_VLForConditionalGeneration"  # noqa: overrides Model.model_type on purpose
+        self.model_type = "Qwen3_VLForConditionalGeneration"
```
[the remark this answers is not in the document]
apsonawane: This is intentional by design. Renaming to [self.qwen3_vl_model_type](vscode-file://vscode-app/c:/Users/asonawane/AppData/Local/Programs/Microsoft%20VS%20Code/ce099c1ed2/resources/app/out/vs/code/electron-browser/workbench/workbench.html) would break [make_genai_config()](vscode-file://vscode-app/c:/Users/asonawane/AppData/Local/Programs/Microsoft%20VS%20Code/ce099c1ed2/resources/app/out/vs/code/electron-browser/workbench/workbench.html) because the base class [Model](vscode-file://vscode-app/c:/Users/asonawane/AppData/Local/Programs/Microsoft%20VS%20Code/ce099c1ed2/resources/app/out/vs/code/electron-browser/workbench/workbench.html) reads [self.model_type](vscode-file://vscode-app/c:/Users/asonawane/AppData/Local/Programs/Microsoft%20VS%20Code/ce099c1ed2/resources/app/out/vs/code/electron-browser/workbench/workbench.html) to write the ["type"](vscode-file://vscode-app/c:/Users/asonawane/AppData/Local/Programs/Microsoft%20VS%20Code/ce099c1ed2/resources/app/out/vs/code/electron-browser/workbench/workbench.html) field in [genai_config.json](vscode-file://vscode-app/c:/Users/asonawane/AppData/Local/Programs/Microsoft%20VS%20Code/ce099c1ed2/resources/app/out/vs/code/electron-browser/workbench/workbench.html). The C++ runtime uses that field to dispatch to the correct model handler.
- Base class sets [self.model_type = config.architectures[0]](vscode-file://vscode-app/c:/Users/asonawane/AppData/Local/Programs/Microsoft%20VS%20Code/ce099c1ed2/resources/app/out/vs/code/electron-browser/workbench/workbench.html) → ["Qwen3VLForConditionalGeneration"](vscode-file://vscode-app/c:/Users/asonawane/AppData/Local/Programs/Microsoft%20VS%20Code/ce099c1ed2/resources/app/out/vs/code/electron-browser/workbench/workbench.html) → normalized to "qwen3vl"
- C++ runtime expects ["qwen3_vl"](vscode-file://vscode-app/c:/Users/asonawane/AppData/Local/Programs/Microsoft%20VS%20Code/ce099c1ed2/resources/app/out/vs/code/electron-browser/workbench/workbench.html) (with underscore)
- So the subclass must override [self.model_type](vscode-file://vscode-app/c:/Users/asonawane/AppData/Local/Programs/Microsoft%20VS%20Code/ce099c1ed2/resources/app/out/vs/code/electron-browser/workbench/workbench.html) to produce the correct config

src/python/py/models/builders/qwen.py
```diff
@@ -911,3 +926,1688 @@ def load_weights(self, input_path):
...
+        # SkipSimplifiedLayerNormalization can be used directly.
+        self.layernorm_attrs["add_offset"] = 1
+
+        # 3D position_ids for mRoPE: [3, batch_size, sequence_length]
+        self.input_shapes["position_ids"] = [3, "batch_size", "sequence_length"]
+        if "position_ids" not in self.input_names:
+            self.input_names["position_ids"] = "position_ids"
+
+        # mRoPE config
+        self.mrope_sections = self.rope_attrs.get("mrope", {}).get("sections", [])
```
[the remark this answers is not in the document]
apsonawane: Fixed

src/python/py/models/builders/qwen.py
```diff
@@ -5,9 +5,18 @@
 # --------------------------------------------------------------------------
 
 
+import json
+import os
```
kunal-vaishnavi: We shouldn't need to import `json` and `os` here to patch issues with the produced files. We can fix their contents before they are serialized to disk.

src/python/py/models/builders/qwen.py
```diff
@@ -911,3 +926,1709 @@ def load_weights(self, input_path):
...
+        if hasattr(config, "rope_scaling") and config.rope_scaling is not None:
+            if "rope_theta" in config.rope_scaling:
+                config.rope_theta = config.rope_scaling["rope_theta"]
+            if "partial_rotary_factor" in config.rope_scaling:
+                config.partial_rotary_factor = config.rope_scaling["partial_rotary_factor"]
+
+        super().__init__(config, io_dtype, onnx_dtype, ep, cache_dir, extra_options)
+
+        # Use opset 23 for newer ops (RMSNormalization, etc.)
+        self.model.opset_imports[""] = 23
```
kunal-vaishnavi: If the opset needs to be increased, then let's increase it for all models and not just this one. Then we don't need this line.

src/python/py/models/builders/qwen.py
```diff
@@ -911,3 +926,1709 @@ def load_weights(self, input_path):
...
+        # architecture uses per-layer cache I/O (KV for attention, conv/recurrent
+        # for linear attention) instead of a single shared template.
+        for suffix in ("key", "value"):
+            for store in (self.input_names, self.input_types, self.input_shapes):
+                del store[f"past_key_values.{suffix}"]
+            for store in (self.output_names, self.output_types, self.output_shapes):
+                del store[f"present.{suffix}"]
+
+        # Build per-layer cache I/O
+        for i in range(self.num_layers):
```
kunal-vaishnavi: Why are we using new naming formats for the full attention layers? We can re-use the existing formats and remove the per-layer entries that aren't needed.

src/python/py/models/builders/qwen.py
```diff
@@ -911,3 +926,1709 @@ def load_weights(self, input_path):
...
+            return super().load_weights(input_path)
+        return AutoModelForCausalLM.from_pretrained(
+            self.model_name_or_path,
+            cache_dir=self.cache_dir,
+            token=self.hf_token,
+            trust_remote_code=self.hf_remote,
+            torch_dtype=torch.float32,
+        )
+
+    def is_layer(self, module):
```
kunal-vaishnavi: Why is this needed? The default already checks for `DecoderLayer` as the name's suffix.

https://github.com/microsoft/onnxruntime-genai/blob/e8c80b6c03f0112b6c5abceab3efe7e159a8da0b/src/python/py/models/builders/base.py#L3951-L3953

src/python/py/models/builders/qwen.py
```diff
@@ -911,3 +926,1709 @@ def load_weights(self, input_path):
...
+        return module.__class__.__name__ == "Qwen3_5DecoderLayer"
+
+    def has_final_norm(self, module, orig_model):
+        model = orig_model
+        if model.__class__.__name__.startswith("Peft"):
```
kunal-vaishnavi: Does Qwen-3.5 have LoRA weights by default? If not, then why is this logic needed?

src/python/py/models/builders/qwen.py
```diff
@@ -911,3 +926,1709 @@ def load_weights(self, input_path):
...
+            model = model.base_model.model
+        return hasattr(model, "model") and hasattr(model.model, "norm") and module == model.model.norm
+
+    def make_layer(self, layer_id, layer):
+        """Build one decoder layer. Dispatches to full attention or
+        GatedDeltaNet linear attention based on self.layer_types."""
+
+        if self.layer_types[layer_id] == "linear_attention":
```
kunal-vaishnavi, line 1112: We should ideally handle the alternating attention layer types by using the same logic that is done for other models: override the `make_attention` method and dispatch to the right `make` method based on the layer id.

src/python/py/models/builders/qwen.py
```diff
@@ -911,3 +926,1709 @@ def load_weights(self, input_path):
...
+            [q_norm_weight_raw, offset_const],
+            [q_norm_weight],
+            name=f"/model/layers.{layer_id}/attn/q_norm/Add_offset",
+        )
+        self.make_value(q_norm_weight, self.io_dtype, list(attention.q_norm.weight.shape))
+
+        q_norm_name = f"/model/layers.{layer_id}/attn/q_norm/RMSNormalization"
+        q_norm_output = f"{q_norm_name}/output_0"
+        self.make_node(
+            "RMSNormalization",
```
kunal-vaishnavi: Have we verified that the ONNX `RMSNormalization` op has the same performance as the ORT `SimplifiedLayerNormalization` op? Can we also check that other IHVs have implemented this op for their EPs? Otherwise, they will likely modify this code to use `SimplifiedLayerNormalization` instead. I would like to avoid EP-specific logic as much as possible.

src/python/py/models/builders/qwen.py
```diff
@@ -911,3 +926,1709 @@ def load_weights(self, input_path):
...
+        self.make_reshape(
+            k_reshape_2,
+            [k_norm_output, f"/model/constants/INT64/[0, -1, {kv_size}]"],
+            self.io_dtype,
+            ["batch_size", "sequence_length", kv_size],
+        )
+        self.attention_attrs["k_path"] = f"{k_reshape_2}/output_0"
+
+    def _make_causal_mask(self):
+        """Build causal attention mask [B, 1, S, total_S] for Attention op.
```
kunal-vaishnavi: Why is the existing causal mask that gets built insufficient?

https://github.com/microsoft/onnxruntime-genai/blob/e8c80b6c03f0112b6c5abceab3efe7e159a8da0b/src/python/py/models/builders/base.py#L4037-L4039

src/python/py/models/builders/qwen.py
```diff
@@ -911,3 +926,1709 @@ def load_weights(self, input_path):
...
+        """
+        mask_output = "/model/causal_mask/output"
+        if mask_output in self.node_names:
+            # Already built — return cached output
+            return "/model/causal_mask/Unsqueeze/output_0"
+
+        basename = "/model/causal_mask"
+        attn_mask = self.input_names["attention_mask"]  # [B, total_S]
+
+        # Constants
```
kunal-vaishnavi: Constant nodes have a special naming format. The `make_constant` method assists with this. Can we use that instead?

https://github.com/microsoft/onnxruntime-genai/blob/e8c80b6c03f0112b6c5abceab3efe7e159a8da0b/src/python/py/models/builders/base.py#L883-L895

src/python/py/models/builders/qwen.py
```diff
@@ -911,3 +926,1709 @@ def load_weights(self, input_path):
...
+        kv_indices_out = f"{kv_indices_name}/output_0"
+        self.make_unsqueeze(
+            kv_indices_name,
+            [cumsum_out, "/model/constants/INT64/[1]"],
+            ir.DataType.INT64,
+            ["batch_size", 1, "total_sequence_length"],
+        )
+
+        # Step 2: Get query length S from inputs_embeds shape
+        embeds_input = self.input_names.get("inputs_embeds", "inputs_embeds")
```
kunal-vaishnavi: Rather than having default names if the input name does not exist in the dictionary, let's access `self.input_names["inputs_embeds"]` directly. If an error occurs, that is an indication that something should have been changed earlier.

src/python/py/models/builders/qwen.py
```diff
@@ -911,3 +926,1709 @@ def load_weights(self, input_path):
...
+        shape_embeds_name = f"{basename}/Shape_embeds"
```
kunal-vaishnavi: Let's make sure that node names [end with the op name itself](https://github.com/microsoft/onnxruntime-genai/blob/main/src/python/py/models/DESIGN.md#namespaces) and move the extra metadata from that end name into the path to the name.

```suggestion
        shape_embeds_name = f"{basename}/embeds/Shape"
```

src/python/py/models/builders/qwen.py
```diff
@@ -911,3 +926,1709 @@ def load_weights(self, input_path):
...
+        self.make_gather(
+            seq_len_name,
+            [f"{shape_embeds_name}/output_0", "/model/constants/INT64/[1]"],
+            ir.DataType.INT64,
+            [1],
+            0,
+        )
+
+        # Get total_S as 1D [1] tensor from attention_mask shape
+        shape_mask_name = f"{basename}/Shape_mask"
```
kunal-vaishnavi:
```suggestion
        shape_mask_name = f"{basename}/mask/Shape"
```

src/python/py/models/builders/qwen.py
```diff
@@ -911,3 +926,1709 @@ def load_weights(self, input_path):
...
+        self.make_value(and_out, ir.DataType.BOOL, ["batch_size", "sequence_length", "total_sequence_length"])
+
+        # Step 6: Where(mask, 0.0, -inf) -> [B, S, total_S]
+        where_name = f"{basename}/Where"
+        where_out = f"{where_name}/output_0"
+        self.make_node("Where", [and_out, zero_f_name, neg_inf_name], [where_out], name=where_name)
+        self.make_value(where_out, ir.DataType.FLOAT, ["batch_size", "sequence_length", "total_sequence_length"])
+
+        # Cast to io_dtype
+        cast_where_name = f"{basename}/Cast_where"
```
kunal-vaishnavi:
```suggestion
        cast_where_name = f"{basename}/where/Cast"
```

src/python/py/models/builders/qwen.py
```diff
@@ -911,3 +926,1709 @@ def load_weights(self, input_path):
...
+
+        # Step 4: Causal mask: q_indices >= kv_indices -> [B, S, total_S]
+        ge_name = f"{basename}/GreaterOrEqual"
+        ge_out = f"{ge_name}/output_0"
+        self.make_node("GreaterOrEqual", [q_indices_out, kv_indices_out], [ge_out], name=ge_name)
+        self.make_value(ge_out, ir.DataType.BOOL, ["batch_size", "sequence_length", "total_sequence_length"])
+
+        # Step 5: Combine with attention_mask padding
+        # attn_mask_bool: [B, 1, total_S]
+        attn_mask_bool_name = f"{basename}/Cast_mask"
```
kunal-vaishnavi:
```suggestion
        attn_mask_bool_name = f"{basename}/mask/Cast"
```

src/python/py/models/builders/qwen.py
```diff
@@ -911,3 +926,1709 @@ def load_weights(self, input_path):
...
+        w_mask = torch.tensor([d == 2 for d in dim_assignments], dtype=torch.bool)
+
+        self.make_initializer(h_mask, "model.rotary_emb.h_mask", to=ir.DataType.BOOL)
+        self.make_initializer(w_mask, "model.rotary_emb.w_mask", to=ir.DataType.BOOL)
+        print(f"Created rotary caches [{max_len}, {rdim_half}] + h/w masks [{rdim_half}].")
+
+    def _get_shared_q_scale(self, head_dim):
+        """Return the name of a shared 1/sqrt(head_dim) constant (created once)."""
+        name = "/model/constants/q_scale"
+        if name not in self.node_names:
```
kunal-vaishnavi: There's no need to do these checks. These node names are automatically checked for duplicates when inserted.

https://github.com/microsoft/onnxruntime-genai/blob/e8c80b6c03f0112b6c5abceab3efe7e159a8da0b/src/python/py/models/builders/base.py#L806-L816

src/python/py/models/builders/qwen.py
```diff
@@ -911,3 +926,1709 @@ def load_weights(self, input_path):
...
+                name,
+                to=self.io_dtype,
+            )
+            self.node_names.add(name)
+        return name
+
+    def _get_shared_l2_eps(self):
+        """Return the name of a shared L2 epsilon constant (created once)."""
+        name = "/model/constants/l2_eps"
+        if name not in self.node_names:
```
kunal-vaishnavi: Same comment here

src/python/py/models/builders/qwen.py
```diff
@@ -911,3 +926,1709 @@ def load_weights(self, input_path):
...
+            g_out = f"{g_name}/output_0"
+            self.make_node("Gather", [pos_ids, f"/model/constants/INT64/[{dim_idx}]"], [g_out], name=g_name, axis=0)
+            sq_name = f"{basename}/{suffix}/dim{dim_idx}/Squeeze"
+            sq_out = f"{sq_name}/output_0"
+            self.make_squeeze(
+                sq_name, [g_out, "/model/constants/INT64/[0]"], ir.DataType.INT64, ["batch_size", "sequence_length"]
+            )
+            gc_name = f"{basename}/{suffix}/dim{dim_idx}/Gather_cache"
+            gc_out = f"{gc_name}/output_0"
+            self.make_node("Gather", [cache_name, sq_out], [gc_out], name=gc_name, axis=0)
```
kunal-vaishnavi: Can we use `make_gather` here?

src/python/py/models/builders/qwen.py
```diff
@@ -911,3 +926,1709 @@ def load_weights(self, input_path):
...
+        def gather_dim(dim_idx, cache_name, suffix):
+            g_name = f"{basename}/{suffix}/dim{dim_idx}/Gather_pos"
+            g_out = f"{g_name}/output_0"
+            self.make_node("Gather", [pos_ids, f"/model/constants/INT64/[{dim_idx}]"], [g_out], name=g_name, axis=0)
+            sq_name = f"{basename}/{suffix}/dim{dim_idx}/Squeeze"
+            sq_out = f"{sq_name}/output_0"
+            self.make_squeeze(
+                sq_name, [g_out, "/model/constants/INT64/[0]"], ir.DataType.INT64, ["batch_size", "sequence_length"]
+            )
+            gc_name = f"{basename}/{suffix}/dim{dim_idx}/Gather_cache"
```
kunal-vaishnavi:
```suggestion
            gc_name = f"{basename}/{suffix}/dim{dim_idx}/cache/Gather"
```

src/python/py/models/builders/qwen.py
```diff
@@ -911,3 +926,1709 @@ def load_weights(self, input_path):
...
+        """
+        pos_ids = self.input_names["position_ids"]
+        cos_cache = "model.rotary_emb.cos_cache"
+        sin_cache = "model.rotary_emb.sin_cache"
+        h_mask = "model.rotary_emb.h_mask"
+        w_mask = "model.rotary_emb.w_mask"
+        rdim_half = self.mrope_rotary_dim // 2
+
+        def gather_dim(dim_idx, cache_name, suffix):
+            g_name = f"{basename}/{suffix}/dim{dim_idx}/Gather_pos"
```
kunal-vaishnavi:
```suggestion
            g_name = f"{basename}/{suffix}/dim{dim_idx}/pos/Gather"
```

src/python/py/models/builders/qwen.py
```diff
@@ -911,3 +926,1709 @@ def load_weights(self, input_path):
...
+        b_name = f"{basename}/in_proj_b/MatMul"
+        self.make_matmul(linear_attn.in_proj_b, b_name, root_input)
+
+        # a (decay/alpha): [B, S, hidden] -> [B, S, n_kv]
+        a_name = f"{basename}/in_proj_a/MatMul"
+        self.make_matmul(linear_attn.in_proj_a, a_name, root_input)
+
+        # 3. Depthwise causal conv1d
+        # Transpose QKV: [B, S, D] -> [B, D, S]
+        qkv_t_name = f"{basename}/qkv_transpose"
```
kunal-vaishnavi:
```suggestion
        qkv_t_name = f"{basename}/qkv_transpose/Transpose"
```

src/python/py/models/builders/qwen.py
```diff
@@ -911,3 +926,1709 @@ def load_weights(self, input_path):
...
+        )
+        self.make_value(conv_output, self.io_dtype, ["batch_size", conv_dim, "sequence_length"])
+
+        # SiLU activation on conv output
+        silu_name = f"{basename}/conv/Sigmoid"
+        silu_sig_output = f"{silu_name}/output_0"
+        self.make_node("Sigmoid", [conv_output], [silu_sig_output], name=silu_name)
+        self.make_value(silu_sig_output, self.io_dtype, ["batch_size", conv_dim, "sequence_length"])
+
+        silu_mul_name = f"{basename}/conv/SiLU_Mul"
```
kunal-vaishnavi:
```suggestion
        silu_mul_name = f"{basename}/conv/SiLU/Mul"
```

src/python/py/models/builders/qwen.py
```diff
@@ -911,3 +926,1709 @@ def load_weights(self, input_path):
...
+
+        # Save new conv state: last (kernel_size-1) timesteps of the padded input
+        present_conv = f"present.{layer_id}.conv_state"
+        conv_state_slice_name = f"{basename}/conv_state/Slice"
+        self.make_slice(
+            conv_state_slice_name,
+            [
+                conv_cat_output,
+                f"/model/constants/INT64/[{-(kernel_size - 1)}]",
+                f"/model/constants/INT64/[{2**62}]",  # large value = end
```
kunal-vaishnavi: Rather than using a specific value, we can use the largest value available in that precision dtype. Ex:

https://github.com/microsoft/onnxruntime-genai/blob/e8c80b6c03f0112b6c5abceab3efe7e159a8da0b/src/python/py/models/builders/base.py#L4184

src/python/py/models/builders/qwen.py
```diff
@@ -911,3 +926,1709 @@ def load_weights(self, input_path):
...
+                "/model/constants/INT64/[-1]",
+            ],
+            self.io_dtype,
+            ["batch_size", conv_dim, kernel_size - 1],
+        )
+        self.make_node(
+            "Identity",
```
kunal-vaishnavi: Why do we need an Identity node? That is a big indicator that something can be optimized.

src/python/py/models/builders/qwen.py
```diff
@@ -911,3 +926,1709 @@ def load_weights(self, input_path):
...
+            [f"{conv_state_slice_name}/output_0"],
+            [present_conv],
+            name=f"{basename}/conv_state/Identity",
+        )
+        self.make_value(present_conv, self.io_dtype, ["batch_size", conv_dim, kernel_size - 1])
+
+        # Transpose conv output back: [B, D, S] -> [B, S, D]
+        conv_out_t_name = f"{basename}/conv_out_transpose"
```
kunal-vaishnavi:
```suggestion
        conv_out_t_name = f"{basename}/conv_out/Transpose"
```

src/python/py/models/builders/qwen.py
```diff
@@ -911,3 +926,1709 @@ def load_weights(self, input_path):
...
+        self.make_node("Mul", [neg_exp_a_name, softplus_output], [g_output], name=g_name)
+        self.make_value(g_output, self.io_dtype, ["batch_size", "sequence_length", n_kv])
+
+        # 7. Reshape Q, K, V to per-head: [B, S, D] -> [B, S, N, H] -> [B, N, S, H]
+        q_4d = self._reshape_to_bhsd(f"{basename}/q", q_scaled_output, self.linear_num_key_heads, hk)
+        k_4d = self._reshape_to_bhsd(f"{basename}/k", k_norm_out, self.linear_num_key_heads, hk)
+        v_4d = self._reshape_to_bhsd(f"{basename}/v", v_out, n_kv, hv)
+
+        # Transpose gates: [B, S, N] -> [B, N, S]
+        beta_t_name = f"{basename}/beta_transpose"
```
kunal-vaishnavi:
```suggestion
        beta_t_name = f"{basename}/beta/Transpose"
```

src/python/py/models/builders/qwen.py
```diff
@@ -911,3 +926,1709 @@ def load_weights(self, input_path):
...
+        beta_t_output = f"{beta_t_name}/output_0"
+        self.make_transpose(beta_t_name, beta_output, self.io_dtype, ["batch_size", n_kv, "sequence_length"], [0, 2, 1])
+
+        g_t_name = f"{basename}/g_transpose"
```
kunal-vaishnavi:
```suggestion
        g_t_name = f"{basename}/g/Transpose"
```

src/python/py/models/builders/qwen.py
```diff
@@ -911,3 +926,1709 @@ def load_weights(self, input_path):
...
+            g_t_output,
+            beta_t_output,
+            n_kv,
+            hk,
+            hv,
+            present_state_name=present_recurrent,
+        )
+
+        # 9. Transpose back: [B, N, S, H] -> [B, S, N*H]
+        la_t_name = f"{basename}/la_transpose"
```
kunal-vaishnavi:
```suggestion
        la_t_name = f"{basename}/la/Transpose"
```

src/python/py/models/builders/qwen.py
```diff
@@ -911,3 +926,1709 @@ def load_weights(self, input_path):
...
+        la_t_output = f"{la_t_name}/output_0"
+        self.make_transpose(
+            la_t_name,
+            la_output,
+            self.io_dtype,
+            ["batch_size", "sequence_length", n_kv, hv],
+            [0, 2, 1, 3],
+        )
+
+        la_flat_name = f"{basename}/la_reshape"
```
kunal-vaishnavi:
```suggestion
        la_flat_name = f"{basename}/la/Reshape"
```

src/python/py/models/builders/qwen.py
```diff
@@ -911,3 +926,1709 @@ def load_weights(self, input_path):
...
+        # The GatedDeltaNet recurrence for each head is:
+        #   S_t = exp(g_t) * S_{t-1} + beta_t * outer(k_t, v_t)
+        #   o_t = q_t @ S_t
+        #
+        # We use the ONNX Scan op to iterate over the sequence dimension.
+        # First, merge batch and heads: [B, N, S, D] -> [B*N, S, D]
+        past_recurrent = f"past_key_values.{layer_id}.recurrent_state"
+        present_recurrent = f"present.{layer_id}.recurrent_state"
+
+        la_output, la_state_output = self._make_scan_recurrence(
```
kunal-vaishnavi: Can this not be replaced by other ops instead? Another indicator that a more optimized approach could be used is if a Scan op is needed.

src/python/py/models/builders/qwen.py
```diff
@@ -911,3 +926,1709 @@ def load_weights(self, input_path):
...
+            [rs_output],
+            name=rs_name,
+            keepdims=1,
+        )
+        self.make_value(rs_output, self.io_dtype, reduced_shape)
+
+        # Add shared epsilon constant
+        eps_name = self._get_shared_l2_eps()
+
+        add_eps_name = f"{basename}/AddEps"
```
kunal-vaishnavi:
```suggestion
        add_eps_name = f"{basename}/epsilon/Add"
```

src/python/py/models/builders/qwen.py
```diff
@@ -911,3 +926,1709 @@ def load_weights(self, input_path):
...
+        unflat_out = f"{unflat_name}/output_0"
+        self.make_reshape(
+            unflat_name,
+            [norm_out, f"{in_shape_name}/output_0"],
+            self.io_dtype,
+            ["batch_size", "sequence_length", total_dim],
+        )
+        return unflat_out
+
+    def _make_l2_normalize(self, basename, input_name, last_dim, leading_dims=None):
```
kunal-vaishnavi, line 1843: Can we use the [`ReduceL2`](https://onnx.ai/onnx/operators/onnx__ReduceL2.html) op instead of these primitive ops?

src/python/py/models/builders/qwen.py
```diff
@@ -911,3 +926,1709 @@ def load_weights(self, input_path):
...
+        out_rs_output = f"{out_rs_name}/output_0"
+        self.make_reshape(
+            out_rs_name,
+            [scan_output_seq, out_target_out],
+            self.io_dtype,
+            ["batch_size", n_heads, "sequence_length", hv],
+        )
+
+        # Unmerge state: [B*N, hk, hv] -> [B, N, hk, hv]
+        state_out_rs_name = f"{scan_basename}/state_out_unmerge"
```
kunal-vaishnavi:
```suggestion
        state_out_rs_name = f"{scan_basename}/state_out_unmerge/Reshape"
```

src/python/py/models/builders/qwen.py
```diff
@@ -911,3 +926,1709 @@ def load_weights(self, input_path):
...
+        self.make_reshape(
+            state_out_rs_name,
+            [scan_output, f"/model/constants/INT64/[-1, {n_heads}, {hk}, {hv}]"],
+            self.io_dtype,
+            ["batch_size", n_heads, hk, hv],
+        )
+        # Rename reshape output to target present state name
+        if present_state_name:
+            self.make_node(
+                "Identity",
```
kunal-vaishnavi: Same question here

src/python/py/models/builder.py
```diff
@@ -141,7 +139,9 @@ def parse_hf_token(hf_token):
 def set_io_dtype(precision, execution_provider, extra_options) -> ir.DataType:
     int4_cpu = precision == "int4" and execution_provider == "cpu"
     fp32_webgpu = execution_provider == "webgpu" and extra_options.get("use_webgpu_fp32", False)
-    bf16_cuda = precision == "int4" and execution_provider in {"cuda", "trt-rtx"} and extra_options.get("use_cuda_bf16", False)
```
kunal-vaishnavi, line 144: Can we keep the spacings in this file as is?

src/python/py/models/builders/qwen.py
apsonawane: Removed

src/python/py/models/builders/qwen.py
apsonawane: removed

src/python/py/models/builders/qwen.py
apsonawane: Qwen3.5 uses standard domain Attention Op which expects float mask. So created a new function for that.

src/python/py/models/builders/qwen.py
apsonawane: Fixed

src/python/py/models/builders/qwen.py
apsonawane: Fixed

src/python/py/models/builders/qwen.py
apsonawane: Fixed

src/python/py/models/builders/qwen.py
apsonawane: Fixed

src/python/py/models/builders/qwen.py
apsonawane: Fixed

src/python/py/models/builders/qwen.py
apsonawane: Fixed

src/python/py/models/builders/qwen.py
apsonawane: Fixed

src/python/py/models/builders/qwen.py
apsonawane: Fixed

src/python/py/models/builders/qwen.py
apsonawane: Fixed

src/python/py/models/builders/qwen.py
apsonawane: Fixed

src/python/py/models/builders/qwen.py
apsonawane: Fixed

src/python/py/models/builders/qwen.py
apsonawane: Fixed

src/python/py/models/builders/qwen.py
apsonawane: Fixed

src/python/py/models/builders/qwen.py
apsonawane: Fixed

src/python/py/models/builders/qwen.py
apsonawane: Fixed

src/python/py/models/builders/qwen.py
apsonawane: We are not calling make_node which checks for duplicates. But make_initializer calls make_value which uses setdefault which returns the exisiting one so the dedupe is not necessary removing the condition

src/python/py/models/builders/qwen.py
apsonawane: We are not calling make_node which checks for duplicates. But make_initializer calls make_value which uses setdefault which returns the exisiting one so the dedupe is not necessary removing the condition

src/python/py/models/builders/qwen.py
apsonawane: Updated to use make_gather

src/python/py/models/builders/qwen.py
apsonawane: Removed identity node

src/python/py/models/builders/qwen.py
apsonawane: Increased the opset version to 23

src/python/py/models/builders/qwen.py
apsonawane: Fixed. The full-attention layers now reuse the base class's existing KV cache naming format ```[past_key_values.{i}.key] / [present.{i}.key]```— we just filter the template lists to exclude linear-attention layer indices instead of deleting and recreating them. Also removed the redundant ```is_layer``` and ```has_final_norm``` overrides (base class handles both), and restructured ```make_layer/make_attention``` to follow the standard dispatch pattern used by other models.

src/python/py/models/builders/qwen.py
apsonawane: Done. Added a ```[make_attention]``` override that dispatches to ```[_make_full_attention]``` or ```[_make_linear_attention]``` based on ```[self.layer_types[layer_id]]```. The [make_layer] override only exists because linear-attention layers use ```[layer.linear_attn]``` instead of ```[layer.self_attn]```— it picks the right module and delegates everything else (layernorm, MLP) to the base class pattern. The attention methods only handle the attention-specific subgraph and end by setting ```[self.layernorm_attrs["skip_input"]]```.

src/python/py/models/builders/qwen.py
apsonawane: Updated

src/python/py/models/builders/qwen.py
apsonawane: Done

src/python/py/models/builders/qwen.py
apsonawane: Updated

src/python/py/models/builders/qwen.py
apsonawane: Removed Identity

src/python/py/models/builders/qwen.py
apsonawane: Agreed — the Scan op is a temporary implementation. We're planning to replace it with a dedicated ```LinearAttention``` op that fuses the entire GatedDeltaNet recurrence (decay, delta rule update, query readout) into a single kernel. Once LinearAttention is ready we will replace Scan with that

src/python/py/models/builders/qwen.py
apsonawane: Updated, using ReduceL2

src/python/py/models/builders/qwen.py
apsonawane: Yes true, ```RMSNormalization``` is a standard op. Replaced it with ```SimplifiedLayerNormalization```

src/python/py/models/builders/qwen.py
apsonawane: Done. ```make_genai_config``` now temporarily adjusts self attributes (```num_layers```, ```model_type```, ```past_present_share_buffer```, KV cache template keys) so the base class produces the correct config directly — no post-hoc patching.

src/python/py/models/builder.py
apsonawane: I just ran lintrunner, that adjusted this

apsonawane: Fixed the spacings

src/python/py/models/builders/base.py
```diff
@@ -108,7 +108,7 @@ def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
             inputs=(),
             outputs=(),
             nodes=(),
-            opset_imports={"": 21, "com.microsoft": 1},
+            opset_imports={"": 23, "com.microsoft": 1},
```
kunal-vaishnavi: Can we check which ops change their signature from 21 to 23 so that we don't break other models?

src/python/py/models/builders/qwen.py
```diff
@@ -701,7 +708,7 @@ def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
         # Fix model_type: HF architecture "Qwen3VLForConditionalGeneration" would produce "qwen3vl"
         # but the C++ runtime expects "qwen3_vl" (with underscore).
         # Intentional override of the superclass attribute (used in genai_config.json).
-        self.model_type = "Qwen3_VLForConditionalGeneration"  # noqa: overrides Model.model_type on purpose
+        self.model_type = "Qwen3_VLForConditionalGeneration"
```
kunal-vaishnavi: Should we revert this because of the CodeQL warning?

src/python/py/models/builders/qwen.py
```diff
@@ -911,3 +924,1458 @@ def load_weights(self, input_path):
...
+        super().__init__(config, io_dtype, onnx_dtype, ep, cache_dir, extra_options)
+
+        # OffsetRMSNorm: Qwen3.5 uses (1 + weight) * RMSNorm(x).
+        # Pre-bake the +1 into the weight initializer so the base class's
+        # SkipSimplifiedLayerNormalization can be used directly.
+        self.layernorm_attrs["add_offset"] = 1
+
+        # 3D position_ids for mRoPE: [3, batch_size, sequence_length]
+        self.input_shapes["position_ids"] = [3, "batch_size", "sequence_length"]
+        if "position_ids" not in self.input_names:
```
kunal-vaishnavi: I believe `position_ids` is always in the input names dictionary by default now. Can you check this and remove this `if` block if it always is?

src/python/py/models/builders/qwen.py
```diff
@@ -911,3 +924,1458 @@ def load_weights(self, input_path):
...
+        if not self.mrope_sections:
+            raise ValueError("MRoPE sections not found in config.text_config.rope_scaling.mrope_section")
+        if len(self.mrope_sections) != 3:
+            raise ValueError(
+                f"Expected 3 MRoPE sections [T, H, W], got {len(self.mrope_sections)}: {self.mrope_sections}"
+            )
+        self.mrope_rotary_dim = int(self.rope_attrs["partial_rotary_factor"] * self.head_size)
+
+        # Force RoPE computation in float32 for numerical stability
+        if "rope_cast" not in self.attention_attrs:
```
kunal-vaishnavi: Given `self.rope_attrs` exists, can the casting setting be moved to that dictionary instead?

src/python/py/models/builders/qwen.py
```diff
@@ -911,3 +924,1458 @@ def load_weights(self, input_path):
...
+            self.attention_attrs["rope_cast"] = {}
+        self.attention_attrs["rope_cast"]["use_fp32"] = True
+
+        # Pre-compute cos/sin cache tables and interleaving masks for mRoPE
+        self._make_rotary_caches()
```
kunal-vaishnavi, line 985: There is a method called `make_rotary_embedding_caches_from_scratch` which already does this and should be overwritten. However, given how much AI-generated code has been added into the model builder lately, all of it will need a refactoring. This is fine for now as I will clean this up later.

src/python/py/models/builders/qwen.py
```diff
@@ -911,3 +924,1458 @@ def load_weights(self, input_path):
...
+        # Derived dimensions for GatedDeltaNet
+        self.linear_key_dim = self.linear_num_key_heads * self.linear_key_head_dim
+        self.linear_value_dim = self.linear_num_value_heads * self.linear_value_head_dim
+        self.linear_conv_dim = self.linear_key_dim * 2 + self.linear_value_dim
+
+        # Full attention uses QK norm and output gating
+        self.attention_attrs["q_norm"] = True
+        self.attention_attrs["k_norm"] = True
+        # Disable packed matmul since Q projection is doubled (4096 vs normal 2048)
+        self.attention_attrs["use_packed_matmul"] = False
```
kunal-vaishnavi: Why does packed MatMul have to be disabled for this reason? Even if the hidden sizes for Q, K, V differ, the packed MatMul should still work. The use of QK norm is a better reason for why this should be disabled.

src/python/py/models/builders/qwen.py
```diff
@@ -911,3 +924,1458 @@ def load_weights(self, input_path):
...
+
+    def _setup_hybrid_cache_io(self):
+        """Set up hybrid cache I/O: KV cache for attention layers,
+        conv_state + recurrent_state for linear attention layers."""
+
+        # Filter the base class's KV cache template lists to only include
+        # full-attention layer indices.  The base class creates entries for
+        # all num_layers — we remove linear-attention layers that use
+        # conv/recurrent state instead of KV cache.
+        attn_indices = {i for i, lt in enumerate(self.layer_types) if lt == "full_attention"}
```
kunal-vaishnavi: If `self.layer_types` exists, why do we need to calculate `attn_indices` separately? I think it would be better to loop through `self.layer_types` and merge both of the following for-loops together.

src/python/py/models/builders/qwen.py
```diff
@@ -911,3 +924,1458 @@ def load_weights(self, input_path):
...
+            self.input_names[f"past_state.{i}.recurrent"] = f"past_key_values.{i}.recurrent_state"
+            self.input_types[f"past_state.{i}.recurrent"] = self.io_dtype
+            self.input_shapes[f"past_state.{i}.recurrent"] = [
+                "batch_size",
+                self.linear_num_value_heads,
+                self.linear_key_head_dim,
+                self.linear_value_head_dim,
+            ]
+
+            self.output_names[f"present_state.{i}.conv"] = f"present.{i}.conv_state"
```
kunal-vaishnavi: Let's make sure that the past/present naming schemas match the same schemas that will be used in the [LFM2 models](https://github.com/microsoft/onnxruntime-genai/pull/1979) so that we don't have to duplicate work across both PRs.

src/python/py/models/builders/qwen.py
```diff
@@ -911,3 +924,1458 @@ def load_weights(self, input_path):
...
+            self.output_types[f"present_state.{i}.recurrent"] = self.io_dtype
+            self.output_shapes[f"present_state.{i}.recurrent"] = [
+                "batch_size",
+                self.linear_num_value_heads,
+                self.linear_key_head_dim,
+                self.linear_value_head_dim,
+            ]
+
+    def load_weights(self, input_path):
+        # Load in float32 to get full precision weights (HF default is bfloat16
```
kunal-vaishnavi: Why do we need full precision weights when loaded into memory? This will use more memory than initially needed. We can cast a weight tensor to float32 as needed before serializing to disk.

Additionally, the `AutoModelForCausalLM` setup is already used in the base `load_weights` method. Why is this method overridden?

src/python/py/models/builders/qwen.py
```diff
@@ -911,3 +924,1458 @@ def load_weights(self, input_path):
...
+        present_k = f"present.{layer_id}.key"
+        present_v = f"present.{layer_id}.value"
+
+        # Build causal attention mask [B, 1, S, total_S]
+        mask_name = self._make_causal_mask()
+
+        attn_name = f"/model/layers.{layer_id}/attn/Attention"
+        attn_output = f"{attn_name}/output_0"
+        self.make_node(
+            "Attention",
```
kunal-vaishnavi: The ONNX Attention op does not support past-present buffer sharing. Can we use `GroupQueryAttention` until it does? Otherwise, users cannot run the model inside ORT GenAI with buffer sharing.

src/python/py/models/builders/qwen.py
```diff
@@ -911,3 +924,1458 @@ def load_weights(self, input_path):
...
+            name=attn_name,
+            q_num_heads=self.num_attn_heads,
+            kv_num_heads=self.num_kv_heads,
+            scale=float(1.0 / np.sqrt(self.head_size)),
+        )
+        self.make_value(
+            attn_output, self.io_dtype, ["batch_size", "sequence_length", self.num_attn_heads * self.head_size]
+        )
+        self.make_value(
+            present_k, self.io_dtype, ["batch_size", self.num_kv_heads, "total_sequence_length", self.head_size]
```
kunal-vaishnavi: The `ValueInfoProto` for each `present_k` and `present_v` output already exists because they are all created when you call `make_inputs_and_outputs`.

src/python/py/models/builders/qwen.py
```diff
@@ -911,3 +924,1458 @@ def load_weights(self, input_path):
...
+        )
+        self.make_value(
+            present_v, self.io_dtype, ["batch_size", self.num_kv_heads, "total_sequence_length", self.head_size]
+        )
+
+        # 7. Output gating: attn_output * sigmoid(gate)
+        sigmoid_name = f"/model/layers.{layer_id}/attn/gate/Sigmoid"
+        sigmoid_output = f"{sigmoid_name}/output_0"
+        self.make_node("Sigmoid", [gate_output], [sigmoid_output], name=sigmoid_name)
```
kunal-vaishnavi: Can we use existing make methods (e.g. `make_sigmoid` or `make_mul`) or add them if they don't already exist?

src/python/py/models/builders/qwen.py
```diff
@@ -911,3 +924,1458 @@ def load_weights(self, input_path):
...
+        self.make_value(sigmoid_output, self.io_dtype, ["batch_size", "sequence_length", q_size])
+
+        gated_name = f"/model/layers.{layer_id}/attn/gate/Mul"
+        gated_output = f"{gated_name}/output_0"
+        self.make_node("Mul", [attn_output, sigmoid_output], [gated_output], name=gated_name)
+        self.make_value(gated_output, self.io_dtype, ["batch_size", "sequence_length", q_size])
+
+        # 8. Output projection
```
kunal-vaishnavi, line 1283: I think we can re-use `make_attention_output_proj` here.

https://github.com/microsoft/onnxruntime-genai/blob/07d21ba199eb4e375d588e76a7b8ba427cf5e3db/src/python/py/models/builders/base.py#L2944-L2966

src/python/py/models/builders/qwen.py
```diff
@@ -911,3 +924,1458 @@ def load_weights(self, input_path):
...
+            [attn_mask, "/model/constants/INT64/[1]"],
+            ir.DataType.INT64,
+            ["batch_size", 1, "total_sequence_length"],
+        )
+        self.make_cast(attn_mask_bool_name, unsq_mask_out, ir.DataType.BOOL, ["batch_size", 1, "total_sequence_length"])
+
+        # And(causal, padding) -> [B, S, total_S]
+        and_name = f"{basename}/And"
+        and_out = f"{and_name}/output_0"
+        self.make_node("And", [ge_out, attn_mask_bool_out], [and_out], name=and_name)
```
kunal-vaishnavi: Same comment with `make_and`, `make_where`, etc throughout this file

src/python/py/models/builders/qwen.py
```diff
@@ -911,3 +924,1458 @@ def load_weights(self, input_path):
...
+        total_s_out = f"{total_s_name}/output_0"
+        self.make_gather(
+            total_s_name, [f"{shape_mask_name}/output_0", "/model/constants/INT64/[1]"], ir.DataType.INT64, [1], 0
+        )
+
+        # Step 3: Slice q_indices from cumsum — last S elements
+        # start = total_S - S (both are 1D [1] tensors)
+        start_name = f"{basename}/q_start/Sub"
+        start_out = f"{start_name}/output_0"
+        self.make_node("Sub", [total_s_out, seq_len_out], [start_out], name=start_name)
```
kunal-vaishnavi: Same comment with `make_sub`, `make_greater_or_equal`, etc throughout this file

src/python/py/models/builders/qwen.py
```diff
@@ -911,3 +924,1458 @@ def load_weights(self, input_path):
...
+        norm_name = f"{basename}/Normalize/Div"
+        norm_output = f"{norm_name}/output_0"
+        self.make_node("Div", [input_name, clip_output], [norm_output], name=norm_name)
+        self.make_value(norm_output, self.io_dtype, full_shape)
+
+        return norm_output
+
+    def _reshape_to_bhsd(self, basename, input_name, num_heads, head_dim):
+        """Reshape [B, S, N*H] -> [B, S, N, H] -> Transpose to [B, N, S, H]"""
+        rs_name = f"{basename}/reshape_4d"
```
kunal-vaishnavi:
```suggestion
        rs_name = f"{basename}/reshape_4d/Reshape"
```

src/python/py/models/builders/qwen.py
```diff
@@ -911,3 +924,1458 @@ def load_weights(self, input_path):
...
+        rs_output = f"{rs_name}/output_0"
+        self.make_reshape(
+            rs_name,
+            [input_name, f"/model/constants/INT64/[0, 0, {num_heads}, {head_dim}]"],
+            self.io_dtype,
+            ["batch_size", "sequence_length", num_heads, head_dim],
+        )
+
+        tp_name = f"{basename}/transpose_bhsd"
```
kunal-vaishnavi:
```suggestion
        tp_name = f"{basename}/transpose_bhsd/Transpose"
```

src/python/py/models/builders/qwen.py
```diff
@@ -911,3 +924,1458 @@ def load_weights(self, input_path):
...
+        )
+
+        # Norm weight (NO offset — Qwen3_5RMSNormGated uses raw weight, not 1+w)
+        norm_weight = f"model.layers.{layer_id}.linear_attn.norm.weight"
+        self.make_initializer(norm_module.weight, norm_weight, to=self.io_dtype)
+
+        # SimplifiedLayerNormalization (com.microsoft, no offset for gated norm)
+        norm_name = f"{basename}/SimplifiedLayerNormalization"
+        norm_output = f"{norm_name}/output_0"
+        self.make_node(
```
kunal-vaishnavi, line 1906: Let's use `make_layernorm` so that the right domain for the op gets registered.

src/python/py/models/builders/qwen.py
apsonawane: Updated

src/python/py/models/builders/qwen.py
apsonawane: Updated

src/python/py/models/builders/qwen.py
apsonawane: Done

src/python/py/models/builders/qwen.py
apsonawane: Yeah this is not required, The base class already disables packed_matmul when q_norm is true

src/python/py/models/builders/qwen.py
apsonawane: Merged into a single loop over ```self.layer_types```. Each iteration either keeps the KV cache entry (full_attention) or adds conv/recurrent state entries (linear_attention). Eliminated the separate attn_indices set and the second loop.

src/python/py/models/builders/qwen.py
apsonawane: Replaced with existing make methods

src/python/py/models/builders/qwen.py
apsonawane: Updated, removed the redundant make_value

src/python/py/models/builders/qwen.py
apsonawane: Done

src/python/py/models/builders/qwen.py
apsonawane: Done

src/python/py/models/builders/qwen.py
apsonawane: Done

src/python/py/models/builders/qwen.py
apsonawane: Done

src/python/py/models/builders/qwen.py
apsonawane: ```make_layernorm``` can't be used here because it's tightly coupled to the decoder layer pipeline (expects ```[B, S, hidden_size]``` shapes). The gated RMSNorm operates on ```[B*S*N, hv]``` tensors with no skip connection. To use ```make_layernorm``` we need a major refactor instead this is simpler. We can add domain here so that it is registered

src/python/py/models/builders/qwen.py
apsonawane: Yes, updated to use GQA

src/python/py/models/builders/qwen.py
apsonawane: Yes this is not required. Removed it

src/python/py/models/builders/qwen.py
```diff
@@ -911,3 +918,1542 @@ def load_weights(self, input_path):
...
+            self.make_cast(cos_cast_name, dyn_cos, compute_dtype, cos_sin_shape)
+            rope_cos = f"{cos_cast_name}/output_0"
+
+            sin_cast_name = f"{basename}/sin/Cast"
+            self.make_cast(sin_cast_name, dyn_sin, compute_dtype, cos_sin_shape)
+            rope_sin = f"{sin_cast_name}/output_0"
+
+        rope_name = f"{basename}/RotaryEmbedding"
+        rope_out = f"{rope_name}/output_0"
+        self.make_node(
```
kunal-vaishnavi, line 1521: Can we use `make_rotary_embedding`? The `make_rotary_embedding_caches` method can be overridden to return the names you want.

src/python/py/models/builders/qwen.py
```diff
@@ -911,3 +918,1542 @@ def load_weights(self, input_path):
...
+
+        silu_mul_name = f"{basename}/conv/SiLU/Mul"
+        self.make_mul(
+            silu_mul_name, [conv_output, silu_sig_output], self.io_dtype, ["batch_size", conv_dim, "sequence_length"]
+        )
+        silu_output = f"{silu_mul_name}/output_0"
+
+        present_conv = f"present.{layer_id}.conv_state"
+        conv_state_slice_name = f"{basename}/conv_state/Slice"
+        self.make_node(
```
kunal-vaishnavi: Can we use `make_slice` here?

src/python/py/models/builders/qwen.py
```diff
@@ -911,3 +918,1542 @@ def load_weights(self, input_path):
...
+        )
+
+        # --- Fused recurrence: LinearAttention (com.microsoft) ---
+        past_recurrent = f"past_key_values.{layer_id}.recurrent_state"
+        present_recurrent = f"present.{layer_id}.recurrent_state"
+
+        la_op_name = f"{basename}/LinearAttention"
+        la_output = f"{la_op_name}/output_0"
+        self.make_node(
+            "LinearAttention",
```
kunal-vaishnavi: We should move this into `make_attention_op` for other models to leverage.

src/python/py/models/builders/qwen.py
```diff
@@ -911,3 +918,1542 @@ def load_weights(self, input_path):
...
+            unsq_name,
+            [input_4d, "/model/constants/INT64/[2]"],
+            self.io_dtype,
+            ["batch_size", n_heads, 1, "sequence_length", head_dim],
+        )
+
+        # Expand: [B, N, 1, S, H] -> [B, N, n_rep, S, H]
+        expand_name = f"{basename}/Expand"
+        expand_out = f"{expand_name}/output_0"
+        self.make_node(
```
kunal-vaishnavi: Can we use `make_expand`?

src/python/py/models/builders/qwen.py
```diff
@@ -911,3 +918,1542 @@ def load_weights(self, input_path):
...
+            name=l2_name,
+            keepdims=1,
+        )
+        self.make_value(l2_output, self.io_dtype, reduced_shape)
+
+        # Clamp to avoid division by zero
+        eps_name = self._get_shared_l2_eps()
+        clip_name = f"{basename}/Clip"
+        clip_output = f"{clip_name}/output_0"
+        self.make_node("Clip", [l2_output, eps_name], [clip_output], name=clip_name)
```
kunal-vaishnavi: There is a lot of the `self.make_node` + `self.make_value` pattern in this file. Can you go through this file and switch them to use `self.make_op_name` as a wrapper? It should simplify the process and reduce the number of lines in this file to add. Many ops already have a `self.make_op_name` wrapper (e.g. `self.make_expand`).

src/models/model.cpp
```diff
@@ -680,7 +680,24 @@ DeviceInterface* SetProviderSessionOptions(OrtSessionOptions& session_options,
...
+                           [](const Config::NamedString& option) {
+                             return option.first == "enable_graph_capture" && option.second == "0";
+                           }) != dml_provider_options.options.end()) {
+            dml_graph_capture_disabled = true;
+            break;
+          }
+        }
+      }
+
+      if (!dml_graph_capture_disabled) {
```
baijumeswani: Why is this change required?

apsonawane: Merged with main, removed this part

src/config.cpp
```diff
@@ -1329,6 +1329,22 @@ void SetProviderOption(Config& config, std::string_view provider_name, std::stri
...
+        for (const auto& value : provider_options.options) {
+          if (value.first == "enable_graph_capture" && value.second == "0") {
+            return false;
+          }
+        }
+      }
+
+      return true;
+    }
+
```
baijumeswani: This as well. Is this needed?

apsonawane: Removed it

baijumeswani (approved):

src/python/py/models/builders/qwen.py
```diff
@@ -911,3 +908,1094 @@ def load_weights(self, input_path):
...
+        z_name, b_name, a_name, qkv_t_output, conv_weight_name = self._make_linear_attention_projections(
+            layer_id, linear_attn, root_input
+        )
+
+        # --- Fused conv: CausalConvWithState (com.microsoft) ---
+        conv_bias_name = f"model.layers.{layer_id}.linear_attn.conv1d.bias"
+        self.make_initializer(torch.zeros(conv_dim, dtype=torch.float32), conv_bias_name, to=self.io_dtype)
+
+        past_conv = f"past_key_values.{layer_id}.conv_state"
+        present_conv = f"present.{layer_id}.conv_state"
```
xenova, line 1587: The naming here is a bit unfortunate, as past_key_values (referring to attention) isn't a very good prefix for conv state. @kunal-vaishnavi and I have had discussions about this in the past, and the naming we currently use on https://huggingface.co/onnx-community/Qwen3.5-0.8B-ONNX looks like this:

Image

as in previous discussions, better names could look something like:
```
past.0.key
past.0.value
past.0.conv
```
and
```
present.0.key
present.0.value
present.0.conv
```

PR #2070, opened 2026-04-07
Enable CUDA graph capture for CUDA EP to improve decode throughput

apsonawane:
### Summary

Enable CUDA graph capture for the CUDA execution provider, which was previously blocked with a hard exception. This eliminates per-token kernel launch overhead during the decode phase, improving throughput by **20-45%** across Qwen3.5 model sizes (0.8B through 9B).

#### 1. `src/config.cpp`

Remove the exception that blocked `enable_cuda_graph=1` for CUDA EP. Instead of throwing, return `true` to enable graph capture — matching the behavior already implemented for DML and NvTensorRtRtx providers.

#### 2. `src/cuda/session_options.cpp`

Make CUDA's `AppendExecutionProvider` respect the `disable_graph_capture` parameter (previously ignored). When `disable_graph_capture=true` (used for vision and embedding sessions), force `enable_cuda_graph` to `"0"` in the provider options. This ensures CUDA graph is only enabled on the decoder session, preventing crashes from dynamic-shape vision/embedding models.

#### 3. `src/models/recurrent_state.cpp`

Fix `RecurrentState::Update()` for CUDA graph compatibility. The original code used `std::swap` to alternate past/present buffers each decode step, which changes memory addresses and invalidates the captured CUDA graph. When graph capture is enabled, copy present→past in-place instead, preserving the stable pointers that the CUDA graph expects.

### Performance Results (A100, Qwen3.5 INT4, 256 tokens, greedy)

| Model | Before | After | Improvement |
|-------|--------|-------|-------------|
| 0.8B  | 166 t/s | 244 t/s | **+47%** |
| 2B    | 141 t/s | 195 t/s | **+38%** |
| 4B    | 94 t/s  | 120 t/s | **+28%** |
| 9B    | 76 t/s  | 93 t/s  | **+22%** |

### Testing

- **Text-only generation**: all 4 model sizes pass
- **VLM with images**: tested 10 different image sizes (224×224 to 800×656) on 0.8B — all pass without CUDA errors
- **Vision and embedding sessions** correctly run without graph capture
- **RecurrentState** (DeltaNet layers) pointer stability verified across multiple generations


src/models/recurrent_state.cpp
```diff
@@ -125,10 +125,22 @@ void RecurrentState::Update() {
...
-    state_.outputs_[output_index_ + i] = presents_[i].get();
+
+  if (state_.params_->use_graph_capture) {
+    // When CUDA graph capture is enabled, we must not swap pointers because
+    // the graph has captured the original memory addresses. Instead, copy
+    // present→past in-place so the pointers remain stable.
+    auto& device = *model_.p_device_kvcache_;
+    for (int i = 0; i < num_layers * 2; ++i) {
+      ByteWrapTensor(device, *pasts_[i]).CopyFrom(ByteWrapTensor(device, *presents_[i]));
+    }
```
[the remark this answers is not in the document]
apsonawane: Fixed

src/models/recurrent_state.cpp
```diff
@@ -119,16 +119,41 @@ void RecurrentState::Add() {
...
-    state_.inputs_[input_index_ + i] = pasts_[i].get();
-    state_.outputs_[output_index_ + i] = presents_[i].get();
+
+  if (state_.params_->use_graph_capture) {
+    // When CUDA graph capture is enabled, we must not swap pointers because
+    // the graph has captured the original memory addresses. Instead, copy
+    // present→past in-place so the pointers remain stable.
+    // Uses cached byte spans to avoid recomputing tensor metadata each step.
+    for (int i = 0; i < num_layers * 2; ++i) {
+      past_byte_spans_[i].CopyFrom(present_byte_spans_[i]);
```
kunal-vaishnavi, line 148: Have we looked at past-present buffer sharing for these? That would eliminate the need to incur any copies after each step of the decoding loop.

src/config.cpp
```diff
@@ -1377,12 +1377,12 @@ bool IsGraphCaptureEnabled(const Config::SessionOptions& session_options) {
                                                });
     if (provider_options != session_options.provider_options.end()) {
       if (provider_options->name == "cuda") {
-        // Graph Capture is currently broken for CUDA
         for (const auto& value : provider_options->options) {
           if (value.first == "enable_cuda_graph" && value.second == "1") {
-            throw std::runtime_error("Graph Capture is currently unsupported for CUDA");
+            return true;
```
kunal-vaishnavi, line 1401: Can we add unit tests for the CUDA CI pipelines to show that graph capture is working / tested?

src/cuda/session_options.cpp
```diff
@@ -81,15 +81,27 @@ void AddCudaStreamConfig(OrtSessionOptions& session_options, DeviceInterface* de
 DeviceInterface* AppendExecutionProvider(OrtSessionOptions& session_options,
                                          const Config::ProviderOptions& provider_options,
                                          const Config& /*config*/,
-                                         bool /*disable_graph_capture*/) {
+                                         bool disable_graph_capture) {
   auto device = GetDeviceInterface(DeviceType::CUDA);
   AddCudaStreamConfig(session_options, device);
+
+  // For non-decoder sessions, disable CUDA graph capture since they can have
+  // dynamic shapes incompatible with graph capture.
```
baijumeswani: This comment suggests that non-decoder sessions will always disable CUDA graph capture. However, the argument passed in could be `false` even for non-decoder sessions.

src/config.cpp
baijumeswani: I remember there was a reason we turned off cuda graph capture. But I do not recollect what it was. @apsonawane it might be a good idea to understand what was the reason we were throwing here to being with (if all the infratructure existed already).

src/models/recurrent_state.cpp
```diff
@@ -119,16 +119,41 @@ void RecurrentState::Add() {
...
   if (layer_indices_.empty()) return;
 
   const int num_layers = static_cast<int>(layer_indices_.size());
-  for (int i = 0; i < num_layers * 2; ++i) {
-    std::swap(pasts_[i], presents_[i]);
-    state_.inputs_[input_index_ + i] = pasts_[i].get();
-    state_.outputs_[output_index_ + i] = presents_[i].get();
+
+  if (state_.params_->use_graph_capture) {
+    // When CUDA graph capture is enabled, we must not swap pointers because
```
baijumeswani: This makes an assumption that the code is being run on CUDA. Is this statement also true for other EPs?

src/config.cpp
kunal-vaishnavi: This was the [original PR](https://github.com/microsoft/onnxruntime-genai/pull/1305) to disable graph capture for CUDA. It was done because there was a performance regression that Nvidia reported. We found that graph input updates were not happening on the GPU.

---

From Copilot:

CUDA graph capture was disabled because:

1. Execution model mismatch

- GenAI decoding is dynamic (KV cache + sequence growth)
- Graph capture requires static execution

2. Fragility

- Requires:
  - All ops on EP
  - No CPU interaction
  - Stable shapes
- Real models violate these

3. Observed issues (same class as PR 1305)

- Incorrect execution paths
- Performance regressions
- Input/device residency problems

4. Not broadly validated

- Only works for some models/configs
- Unsafe to expose globally

baijumeswani: From what I recollect, graph capture did not work with continuous decoding at the time for cuda. I think this was when we started supporting continuous decoding (probably before our engagement with NV team).

Having said that, if enabling cuda graph capture shows ~ 20-45% improvement without degrading quality, we should remove this throw.

src/cuda/session_options.cpp
baijumeswani: Thinking more about this, would the session options have cuda graph capture enabled for non-decoder sessions? We should be able to set the session options on all sessions from the config now. So we shouldn't overwrite what the config has here?

apsonawane: So we need to keep this overwrite because there is a fallback in multimodal.cpp where if session options is not defined then it overwrites with decoder options.

```
  if (vision) {
    vision_session_options_ = OrtSessionOptions::Create();
    CreateSessionOptionsFromConfig(config_->model.vision.session_options.has_value() ? config_->model.vision.session_options.value() : config_->model.decoder.session_options, *vision_session_options_, true, /*disable_graph_capture=*/true);
    vision_session_ = CreateSession(ort_env, config_->model.vision.filename, vision_session_options_.get());
  }
```

src/models/recurrent_state.cpp
apsonawane: It is true for other EPs. Made the wording EP agnostic

src/models/recurrent_state.cpp
apsonawane: We cannot enable past_present_buffer sharing here because for recurrent states the model overwrites the entire tensor in each step. When input and output alias the same buffer, the model reads from memory thats being simultaneously written causing illegal memory access.

src/cuda/session_options.cpp
baijumeswani: We should update the model in that case. Because we intentionally kept the configuration so that we honor fallback to the decoder session options. With this, we are overwriting the expected session options.

baijumeswani (dismissed):

kunal-vaishnavi (approved):

src/cuda/session_options.cpp
```diff
@@ -84,6 +84,11 @@ DeviceInterface* AppendExecutionProvider(OrtSessionOptions& session_options,
                                          bool /*disable_graph_capture*/) {
   auto device = GetDeviceInterface(DeviceType::CUDA);
   AddCudaStreamConfig(session_options, device);
+
+  // CUDA EP honors enable_cuda_graph from the config as-is. Non-decoder
+  // sessions (vision, embedding, speech) should set enable_cuda_graph=0
+  // in their own session_options in genai_config.json.
```
justinchuby, line 90: @rui-ren maybe we should add this to the mobius generator as well?

baijumeswani (approved):

PR #2077, opened 2026-04-08
[Mistral3] Add VLM support with multi-image inference

titaiwangms:
## Summary

Adds Mistral3/Pixtral VLM support to onnxruntime-genai with multi-image inference. Includes C++ image processor, PixtralVisionState for per-image vision processing, Python export support, and comprehensive tests.

### C++ Runtime
- **Mistral3 image processor** — `[IMG]`/`[IMG_BREAK]`/`[IMG_END]` token expansion based on image resolution and patch geometry, multi-image support
- **PixtralVisionState** — per-image vision processing loop with bounds checks and overflow guard; slices from padded batch tensor using `image_sizes` metadata from ort-extensions `PixtralImageSizes` op
- **Virtual `SetExtraInputs`** — proper polymorphic dispatch for vision state subclasses
- **`IsPixtralFamily()` model type detection** — enables Pixtral-specific codepath
- **`processor_config.json`** — with `PixtralImageSizes` preprocessing step
- **`context_length` / `max_length` separation** — `context_length` controls KV cache allocation while `max_length` controls generation stopping, preventing premature EOS with large image token counts
- **INT32 `input_ids`** — token IDs above 32767 (Pixtral `[IMG]`=128011) require int32

### Python Export Support
- Mistral3 model classes (`Mistral3Config`, `Mistral3ForConditionalGeneration`)
- FP8 dtype promotion for checkpoint loading
- `get_user_content()` handler for Mistral3 prompt formatting

## Multi-Image Architecture

Pixtral uses dynamic image sizes (28×28 to 1540×1540) so images can't be batched in the vision encoder. `PixtralVisionState` processes each image individually by:
1. Reading `image_sizes` tensor from ort-extensions `PixtralImageSizes` op (provides per-image H×W)
2. Slicing the padded `[N, C, max_H, max_W]` batch tensor to extract each image's actual pixels
3. Running vision encoder on each image separately
4. Concatenating vision embeddings for the decoder

## Dependencies

- **onnxruntime-extensions PR #1050** — `PixtralImageSizes` custom op for image size metadata
- **PR #2076** — YaRN RoPE parity fixes (merged) ✅

## Testing

- 5 multi-image token expansion tests (various image sizes and counts)
- Virtual dispatch verification tests for `SetExtraInputs`
- YaRN RoPE parity tests (from merged #2076)
- E2E verified: multi-image (fish.jpg + challenge.jpg) correctly describes both images

### DIFF
--- a/examples/python/common.py
+++ b/examples/python/common.py
@@ -4,11 +4,10 @@
 import argparse
 import json
 import os
-from dataclasses import asdict, dataclass
-from typing import Any
-
 import onnxruntime_genai as og
 
+from dataclasses import dataclass, asdict
+from typing import Any
 
 def set_logger(inputs: bool = True, outputs: bool = True) -> None:
     """
@@ -21,7 +20,6 @@ def set_logger(inputs: bool = True, outputs: bool = True) -> None:
         None
     """
     og.set_log_options(enabled=True, model_input_values=inputs, model_output_values=outputs)
-
 
 def register_ep(ep: str, ep_path: str, use_winml: bool) -> None:
     """
@@ -44,7 +42,6 @@ def register_ep(ep: str, ep_path: str, use_winml: bool) -> None:
         # Modified from here: https://learn.microsoft.com/en-us/windows/ai/new-windows-ml/tutorial?tabs=python#acquiring-the-model-and-preprocessing
         try:
             import winml
-
             print(winml.register_execution_providers(ort=False, ort_genai=True))
         except ImportError:
             print("WinML not available, using default execution providers")
@@ -56,13 +53,10 @@ def register_ep(ep: str, ep_path: str, use_winml: bool) -> None:
         og.register_execution_provider_library("NvTensorRTRTXExecutionProvider", ep_path)
     else:
         print(f"Warning: EP registration not supported for {ep}")
-        print(
-            "Only 'cuda' and 'NvTensorRtRtx' support plug-in libraries. Use Windows ML via '--use_winml' to register EPs."
-        )
+        print("Only 'cuda' and 'NvTensorRtRtx' support plug-in libraries. Use Windows ML via '--use_winml' to register EPs.")
         return
 
     print(f"Registered {ep} successfully!")
-
 
 def get_config(path: str, ep: str, ep_options: dict[str, str] = {}, search_options: dict[str, int] = {}) -> og.Config:
     """
@@ -104,7 +98,6 @@ def get_config(path: str, ep: str, ep_options: dict[str, str] = {}, search_optio
     config.overlay(json.dumps({"search": search_options}))
     return config
 
-
 def get_search_options(args: argparse.Namespace):
     """
     Get search options for a generator's params during decoding
@@ -135,10 +128,7 @@ def get_search_options(args: argparse.Namespace):
     search_options["batch_size"] = search_options.get("batch_size", 1)
     return search_options
 
-
-def apply_chat_template(
-    model_path: str, tokenizer: og.Tokenizer, messages: str, add_generation_prompt: bool, tools: str = ""
-) -> str:
+def apply_chat_template(model_path: str, tokenizer: og.Tokenizer, messages: str, add_generation_prompt: bool, tools: str = "") -> str:
     """
     Apply the chat template with various fallback options
 
@@ -160,7 +150,6 @@ def apply_chat_template(
         messages=messages, tools=tools, add_generation_prompt=add_generation_prompt, template_str=template_str
     )
     return prompt
-
 
 def get_user_prompt(prompt: str, non_interactive: bool) -> str:
     """
@@ -190,7 +179,6 @@ def get_user_prompt(prompt: str, non_interactive: bool) -> str:
 
     return text
 
-
 def get_user_media_paths(media_paths: list[str], non_interactive: bool, media_type: str) -> list[str]:
     """
     Get paths to media for user
@@ -214,9 +202,7 @@ def get_user_media_paths(media_paths: list[str], non_interactive: bool, media_ty
         # If interactive mode is on
         paths = [
             path.strip()
-            for path in input(
-                f"{media_type.capitalize()} Path (comma separated; leave empty if no {media_type}): "
-            ).split(",")
+            for path in input(f"{media_type.capitalize()} Path (comma separated; leave empty if no {media_type}): ").split(",")
         ]
 
     paths = [path for path in paths if path]
@@ -226,7 +212,6 @@ def get_user_media_paths(media_paths: list[str], non_interactive: bool, media_ty
         print(f"Using {media_type}: {path}")
 
     return paths
-
 
 def get_user_images(image_paths: list[str], non_interactive: bool) -> tuple[og.Images, int]:
     """
@@ -247,7 +232,6 @@ def get_user_images(image_paths: list[str], non_interactive: bool) -> tuple[og.I
     images = og.Images.open(*paths)
     return images, len(paths)
 
-
 def get_user_audios(audio_paths: list[str], non_interactive: bool) -> tuple[og.Audios, int]:
     """
     Get audios for user
@@ -266,7 +250,6 @@ def get_user_audios(audio_paths: list[str], non_interactive: bool) -> tuple[og.A
 
     audios = og.Audios.open(*paths)
     return audios, len(paths)
-
 
 def get_user_content(model_type: str, num_images: int, num_audios: int, prompt: str) -> str | list[dict[str, str]]:
     """
@@ -306,52 +289,43 @@ def get_user_content(model_type: str, num_images: int, num_audios: int, prompt: 
         content = image_tags + [{"type": "text", "text": prompt}]
     return content
 
-
 @dataclass
 class ToolSchema:
     """
     A class for defining a tool in a JSON schema compatible way
     """
-
     description: str
     type: str
     properties: dict[str, Any]
     required: list[str]
     additionalProperties: bool
 
-
 @dataclass
 class JsonSchema:
     """
     A class for defining a JSON schema for guidance
     """
-
     x_guidance: dict[str, Any]
     type: str
     items: dict[str, list[ToolSchema]]
     minItems: int
-
 
 @dataclass
 class FunctionDefinition:
     """
     A class for defining a function in an OpenAI-compatible way
     """
-
     name: str
     description: str
     parameters: dict[str, Any]
-
 
 @dataclass
 class Tool:
     """
     A class for defining a tool in an OpenAI-compatible way
     """
-
     type: str
     function: FunctionDefinition
-
 
 def tools_to_schemas(tools: list[Tool]) -> list[ToolSchema]:
     """
@@ -386,7 +360,6 @@ def tools_to_schemas(tools: list[Tool]) -> list[ToolSchema]:
         tool_schemas.append(tool_schema)
     return tool_schemas
 
-
 def get_json_schema(tools: list[Tool], tool_output: bool) -> str:
     """
     Create a JSON schema from a list of tools
@@ -402,7 +375,6 @@ def get_json_schema(tools: list[Tool], tool_output: bool) -> str:
     json_schema = JsonSchema(x_guidance=x_guidance, type="array", items={"anyOf": schemas}, minItems=int(tool_output))
     d = {k.replace("x_guidance", "x-guidance"): v for k, v in asdict(json_schema).items()}
     return json.dumps(d)
-
 
 def get_lark_grammar(
     tools: list[Tool],
@@ -451,7 +423,6 @@ def get_lark_grammar(
 
     return "\n".join(rows)
 
-
 def to_tool(tool_defs: list[dict[str, Any]]) -> list[Tool]:
     """
     Convert a JSON-deserialized object of tools to a list of Tool objects
@@ -471,7 +442,6 @@ def to_tool(tool_defs: list[dict[str, Any]]) -> list[Tool]:
         tool = Tool(type="function", function=func)
         tools.append(tool)
     return tools
-
 
 def get_guidance(
     response_format: str = "",
@@ -504,7 +474,7 @@ def get_guidance(
     if tool_output:
         if os.path.exists(filepath):
             # If tools are provided as a file
-            with open(filepath) as f:
+            with open(filepath, 'r') as f:
                 tool_defs = json.load(f)
                 tools = to_tool(tool_defs)
         elif tools_str != "":
@@ -518,18 +488,14 @@ def get_guidance(
             if type(tools[0]) != Tool:
                 tools = to_tool(tools)
         else:
-            raise ValueError(
-                "Please provide the list of tools through a file, JSON-serialized string, or a list of tools"
-            )
+            raise ValueError("Please provide the list of tools through a file, JSON-serialized string, or a list of tools")
 
         assert len(tools) > 0, "Could not obtain a list of tools in memory"
 
     # Create guidance based on user-provided response format
     if response_format in {"text", "lark_grammar"}:
         if response_format == "text":
-            assert text_output and not tool_output, (
-                "A response format of 'text' requires text_output = True and tool_output = False"
-            )
+            assert text_output and not tool_output, "A response format of 'text' requires text_output = True and tool_output = False"
 
         guidance_type = "lark_grammar"
         guidance_data = get_lark_grammar(
@@ -540,9 +506,7 @@ def get_guidance(
             tool_call_end=tool_call_end,
         )
     elif response_format in {"json_schema", "json_object"}:
-        assert tool_output and not text_output, (
-            "A response format of 'json_schema' or 'json_object' requires text_output = False and tool_output = True"
-        )
+        assert tool_output and not text_output, "A response format of 'json_schema' or 'json_object' requires text_output = False and tool_output = True"
 
         guidance_type = "json_schema"
         guidance_data = get_json_schema(tools=tools, tool_output=tool_output)
@@ -550,7 +514,6 @@ def get_guidance(
         raise ValueError("Invalid response format provided")
 
     return guidance_type, guidance_data, json.dumps([asdict(tool) for tool in tools])
-
 
 def get_generator_params_args(parser: argparse.ArgumentParser) -> None:
     """
@@ -562,34 +525,16 @@ def get_generator_params_args(parser: argparse.ArgumentParser) -> None:
         None
     """
     generator_params = parser.add_argument_group("Generator Params")
-    generator_params.add_argument(
-        "-c",
-        "--chunk_size",
-        type=int,
-        default=0,
-        help="Chunk size for prefill chunking during context processing (default: 0 = disabled, >0 = enabled)",
-    )
-    generator_params.add_argument(
-        "-s",
-        "--do_sample",
-        action="store_true",
-        help="Do random sampling. When false, greedy or beam search are used to generate the output. Defaults to false",
-    )
-    generator_params.add_argument(
-        "-i", "--min_length", type=int, help="Min number of tokens to generate including the prompt"
-    )
-    generator_params.add_argument(
-        "-l", "--max_length", type=int, help="Max number of tokens to generate including the prompt"
-    )
-    generator_params.add_argument("-b", "--num_beams", type=int, default=1, help="Number of beams to create")
-    generator_params.add_argument(
-        "-rs", "--num_return_sequences", type=int, default=1, help="Number of return sequences to produce"
-    )
-    generator_params.add_argument("-r", "--repetition_penalty", type=float, help="Repetition penalty to sample with")
-    generator_params.add_argument("-t", "--temperature", type=float, help="Temperature to sample with")
-    generator_params.add_argument("-k", "--top_k", type=int, help="Top k tokens to sample from")
-    generator_params.add_argument("-p", "--top_p", type=float, help="Top p probability to sample with")
-
+    generator_params.add_argument('-c', '--chunk_size', type=int, default=0, help="Chunk size for prefill chunking during context processing (default: 0 = disabled, >0 = enabled)")
+    generator_params.add_argument('-s', '--do_sample', action='store_true', help='Do random sampling. When false, greedy or beam search are used to generate the output. Defaults to false')
+    generator_params.add_argument('-i', '--min_length', type=int, help='Min number of tokens to generate including the prompt')
+    generator_params.add_argument('-l', '--max_length', type=int, help='Max number of tokens to generate including the prompt')
+    generator_params.add_argument('-b', '--num_beams', type=int, default=1, help='Number of beams to create')
+    generator_params.add_argument('-rs', '--num_return_sequences', type=int, default=1, help='Number of return sequences to produce')
+    generator_params.add_argument('-r', '--repetition_penalty', type=float, help='Repetition penalty to sample with')
+    generator_params.add_argument('-t', '--temperature', type=float, help='Temperature to sample with')
+    generator_params.add_argument('-k', '--top_k', type=int, help='Top k tokens to sample from')
+    generator_params.add_argument('-p', '--top_p', type=float, help='Top p probability to sample with')
 
 def get_guidance_args(parser: argparse.ArgumentParser) -> None:
     """
@@ -601,38 +546,9 @@ def get_guidance_args(parser: argparse.ArgumentParser) -> None:
         None
     """
     guidance = parser.add_argument_group("Guidance Arguments")
-    guidance.add_argument(
-        "-rf",
-        "--response_format",
-        type=str,
-        default="",
-        choices=["", "text", "json_object", "json_schema", "lark_grammar"],
-        help="Provide response format for the model",
-    )
-    guidance.add_argument(
-        "-tf",
-        "--tools_file",
-        type=str,
-        default="",
-        help="Path to file containing list of OpenAI-compatible tool definitions. Ex: test/test_models/tool-definitions/weather.json",
-    )
-    guidance.add_argument(
-        "-text", "--text_output", action="store_true", default=False, help="Produce a text response in the output"
-    )
-    guidance.add_argument(
-        "-tool", "--tool_output", action="store_true", default=False, help="Produce a tool call in the output"
-    )
-    guidance.add_argument(
-        "-tcs",
-        "--tool_call_start",
-        type=str,
-        default="",
-        help="String representation of tool call start (ex: <|tool_call|>). Needs to be marked as special in tokenizer.json for guidance to work.",
-    )
-    guidance.add_argument(
-        "-tce",
-        "--tool_call_end",
-        type=str,
-        default="",
-        help="String representation of tool call end (ex: <|/tool_call|>). Needs to be marked as special in tokenizer.json for guidance to work.",
-    )
+    guidance.add_argument('-rf', '--response_format', type=str, default="", choices=["", "text", "json_object", "json_schema", "lark_grammar"], help='Provide response format for the model')
+    guidance.add_argument('-tf', '--tools_file', type=str, default="", help='Path to file containing list of OpenAI-compatible tool definitions. Ex: test/test_models/tool-definitions/weather.json')
+    guidance.add_argument('-text', '--text_output', action='store_true', default=False, help='Produce a text response in the output')
+    guidance.add_argument('-tool', '--tool_output', action='store_true', default=False, help='Produce a tool call in the output')
+    guidance.add_argument('-tcs', '--tool_call_start', type=str, default="", help='String representation of tool call start (ex: <|tool_call|>). Needs to be marked as special in tokenizer.json for guidance to work.')
+    guidance.add_argument('-tce', '--tool_call_end', type=str, default="", help='String representation of tool call end (ex: <|/tool_call|>). Needs to be marked as special in tokenizer.json for guidance to work.')
--- a/examples/python/common.py
+++ b/examples/python/common.py
@@ -4,10 +4,11 @@
 import argparse
 import json
 import os
+from dataclasses import asdict, dataclass
+from typing import Any
+
 import onnxruntime_genai as og
 
-from dataclasses import dataclass, asdict
-from typing import Any
 
 def set_logger(inputs: bool = True, outputs: bool = True) -> None:
     """
@@ -20,6 +21,7 @@ def set_logger(inputs: bool = True, outputs: bool = True) -> None:
         None
     """
     og.set_log_options(enabled=True, model_input_values=inputs, model_output_values=outputs)
+
 
 def register_ep(ep: str, ep_path: str, use_winml: bool) -> None:
     """
@@ -42,6 +44,7 @@ def register_ep(ep: str, ep_path: str, use_winml: bool) -> None:
         # Modified from here: https://learn.microsoft.com/en-us/windows/ai/new-windows-ml/tutorial?tabs=python#acquiring-the-model-and-preprocessing
         try:
             import winml
+
             print(winml.register_execution_providers(ort=False, ort_genai=True))
         except ImportError:
             print("WinML not available, using default execution providers")
@@ -53,10 +56,13 @@ def register_ep(ep: str, ep_path: str, use_winml: bool) -> None:
         og.register_execution_provider_library("NvTensorRTRTXExecutionProvider", ep_path)
     else:
         print(f"Warning: EP registration not supported for {ep}")
-        print("Only 'cuda' and 'NvTensorRtRtx' support plug-in libraries. Use Windows ML via '--use_winml' to register EPs.")
+        print(
+            "Only 'cuda' and 'NvTensorRtRtx' support plug-in libraries. Use Windows ML via '--use_winml' to register EPs."
+        )
         return
 
     print(f"Registered {ep} successfully!")
+
 
 def get_config(path: str, ep: str, ep_options: dict[str, str] = {}, search_options: dict[str, int] = {}) -> og.Config:
     """
@@ -98,6 +104,7 @@ def get_config(path: str, ep: str, ep_options: dict[str, str] = {}, search_optio
     config.overlay(json.dumps({"search": search_options}))
     return config
 
+
 def get_search_options(args: argparse.Namespace):
     """
     Get search options for a generator's params during decoding
@@ -128,7 +135,10 @@ def get_search_options(args: argparse.Namespace):
     search_options["batch_size"] = search_options.get("batch_size", 1)
     return search_options
 
-def apply_chat_template(model_path: str, tokenizer: og.Tokenizer, messages: str, add_generation_prompt: bool, tools: str = "") -> str:
+
+def apply_chat_template(
+    model_path: str, tokenizer: og.Tokenizer, messages: str, add_generation_prompt: bool, tools: str = ""
+) -> str:
     """
     Apply the chat template with various fallback options
 
@@ -150,6 +160,7 @@ def apply_chat_template(model_path: str, tokenizer: og.Tokenizer, messages: str,
         messages=messages, tools=tools, add_generation_prompt=add_generation_prompt, template_str=template_str
     )
     return prompt
+
 
 def get_user_prompt(prompt: str, non_interactive: bool) -> str:
     """
@@ -179,6 +190,7 @@ def get_user_prompt(prompt: str, non_interactive: bool) -> str:
 
     return text
 
+
 def get_user_media_paths(media_paths: list[str], non_interactive: bool, media_type: str) -> list[str]:
     """
     Get paths to media for user
@@ -202,7 +214,9 @@ def get_user_media_paths(media_paths: list[str], non_interactive: bool, media_ty
         # If interactive mode is on
         paths = [
             path.strip()
-            for path in input(f"{media_type.capitalize()} Path (comma separated; leave empty if no {media_type}): ").split(",")
+            for path in input(
+                f"{media_type.capitalize()} Path (comma separated; leave empty if no {media_type}): "
+            ).split(",")
         ]
 
     paths = [path for path in paths if path]
@@ -212,6 +226,7 @@ def get_user_media_paths(media_paths: list[str], non_interactive: bool, media_ty
         print(f"Using {media_type}: {path}")
 
     return paths
+
 
 def get_user_images(image_paths: list[str], non_interactive: bool) -> tuple[og.Images, int]:
     """
@@ -232,6 +247,7 @@ def get_user_images(image_paths: list[str], non_interactive: bool) -> tuple[og.I
     images = og.Images.open(*paths)
     return images, len(paths)
 
+
 def get_user_audios(audio_paths: list[str], non_interactive: bool) -> tuple[og.Audios, int]:
     """
     Get audios for user
@@ -250,6 +266,7 @@ def get_user_audios(audio_paths: list[str], non_interactive: bool) -> tuple[og.A
 
     audios = og.Audios.open(*paths)
     return audios, len(paths)
+
 
 def get_user_content(model_type: str, num_images: int, num_audios: int, prompt: str) -> str | list[dict[str, str]]:
     """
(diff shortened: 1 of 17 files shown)


titaiwangms:
> ⚠️ **Merge Dependency:** This PR is stacked on top of PR #2076 (`pr/fix-yarn-config-resolution`). **PR #2076 must be merged first**, then this PR can be rebased onto `main` and merged.

src/models/mistral3_image_processor.cpp
```diff
@@ -0,0 +1,295 @@
...
+
+  auto [input_ids, num_img_tokens] =
+      ProcessPixtralPrompt(tokenizer, prompt, pixel_values, patch_size_,
+                           spatial_merge_size_, allocator);
+
+  named_tensors->emplace(std::string(Config::Defaults::InputIdsName),
+                         std::make_shared<Tensor>(std::move(input_ids)));
+
+  // Process pixel_values and transpose from NHWC [N,H,W,C] to NCHW [N,C,H,W]
+  // since ort-extensions outputs NHWC but the vision model expects NCHW.
```
kunal-vaishnavi: Can we fix this inside ORT extensions instead of doing the transpose inside ORT GenAI?

src/python/py/models/builder.py
```diff
@@ -246,6 +247,24 @@ def create_model(
...
+        # theoretical YaRN limit; cap to a practical context length so the
+        # generated genai_config.json is usable without OOM.
+        # Users can override via: --extra_options context_length=<value>
+        default_context_length = int(extra_options.get("context_length", 32768))
+        if hasattr(text_config, "max_position_embeddings") and text_config.max_position_embeddings > default_context_length:
+            config.max_position_embeddings = default_context_length
+        if hasattr(config, "quantization_config"):
+            delattr(config, "quantization_config")
+        if "exclude_embeds" not in extra_options:
+            print("WARNING: This is only generating the text component of the model. Setting `--extra_options exclude_embeds=true` by default.")
```
kunal-vaishnavi: Can you always turn this on and remove the `if` guard? The `exclude_embeds` flag will always need to be true.

src/python/py/models/builder.py
```diff
@@ -246,6 +247,24 @@ def create_model(
...
         onnx_model = MistralModel(config, io_dtype, onnx_dtype, execution_provider, cache_dir, extra_options)
+    elif config.architectures[0] == "Mistral3ForConditionalGeneration":
+        text_config = config.text_config
+        for key in text_config:
+            if not hasattr(config, key):
+                setattr(config, key, getattr(text_config, key))
+        # The text_config max_position_embeddings (262144) reflects the
+        # theoretical YaRN limit; cap to a practical context length so the
+        # generated genai_config.json is usable without OOM.
+        # Users can override via: --extra_options context_length=<value>
```
kunal-vaishnavi: The `context_length` field in the GenAI config's model section is used inside ORT GenAI to determine the theoretical max. The `max_length` field in the GenAI config's search section is used to cap to a practical length. 

Rather than having `context_length=<value>`, can we instead set the `context_length` to its original value and set the `max_length` to `self.max_length`? We can add `self.max_length = self.context_length` in the base constructor and then override `self.max_length` when creating the GenAI config.

titaiwangms: Agree this separation is architecturally cleaner — `context_length` for theoretical max, `max_length` for practical cap.

One concern: `context_length` is used in `base.py` to set the cos/sin cache length (`rope_attrs["cache_length"]`). If we set `context_length=262144` (Ministral-3-3B's full YaRN limit), that allocates a 262K-entry RoPE cache:

- FP16, head_dim=128: `262144 × 128 × 2 (cos+sin) × 2 bytes = ~128 MB`
- Compared to current 32K cap: `32768 × 128 × 2 × 2 = ~16 MB`

Two options for the cache:

**A.** Keep cache at `context_length` (262K) — correct for full RoPE extrapolation but 128MB extra memory

**B.** Cap cache at `max_length` while keeping `context_length` at 262K — saves memory but limits RoPE to the practical generation length

For Ministral-3-3B, option B is fine since YaRN dynamically extends beyond the cache via NTK scaling. But for Phi-3/LLaMA-3 with SU/longrope, the cache needs the full context_length for the long/short factor split.

Happy to implement either way — which tradeoff do you prefer? Or should we add `self.max_length` to the base constructor with a default of `self.context_length` and let Mistral3TextModel override it?

titaiwangms:
Addressed review feedback from @kunal-vaishnavi:

**#3 (NHWC transpose)**: ✅ Implemented — moved NHWC→NCHW transpose from C++ to ort-extensions via `Permute3D` in `processor_config.json`, matching the Gemma3 pattern. Removed ~40 lines of C++ transpose code.

**#4 (exclude_embeds)**: ✅ Implemented — always set `exclude_embeds=True` unconditionally for VLM text decoder. Removed the conditional check and warning print.

**#5 (context_length)**: Awaiting preference — current default is 32768 (configurable via `--extra_options context_length=<value>`). Happy to change to `original_max_position_embeddings` (16384) if preferred for memory efficiency.

src/python/py/models/builder.py
kunal-vaishnavi: The cos/sin caches are always pre-computed to the context length that the model can support. This is because re-computation is not done within the RoPE ops or the fused attention ops that compute RoPE. Without pre-computing to the theoretical max, the generated ONNX models won't run at a sequence length larger than the size of the first dimension of the caches.

titaiwangms (commented):
Implemented context_length/max_length separation as suggested. context_length=262144 for full RoPE cache pre-computation (cos/sin caches cover the theoretical YaRN range), max_length=32768 as practical generation cap. Added `self.max_length` to base Model class (defaults to `self.context_length` for backward compatibility with all existing models). Mistral3TextModel overrides via `extra_options["max_length"]`.

src/python/py/models/builder.py
titaiwangms: Done

src/models/mistral3_image_processor.cpp
```diff
@@ -0,0 +1,248 @@
...
+  if (num_images > 0 && image_h > 0 && image_w > 0) {
+    patch_rows = static_cast<int>(image_h) / patch_size / spatial_merge_size;
+    patch_cols = static_cast<int>(image_w) / patch_size / spatial_merge_size;
+    image_token_sequence = BuildImageTokenSequence(patch_rows, patch_cols,
+                                                   img_token_id, img_break_id, img_end_id);
+    num_img_tokens = static_cast<int64_t>(std::count(image_token_sequence.begin(),
+                                                     image_token_sequence.end(),
+                                                     img_token_id));
+  }
+
```
tianleiwu, line 94: **Suggestion**: The patch grid division truncates `int64_t` to `int` before dividing:
```cpp
patch_rows = static_cast<int>(image_h) / patch_size / spatial_merge_size;
```
For safety, consider performing the division in `int64_t` space first, then narrowing:
```cpp
patch_rows = static_cast<int>(image_h / patch_size / spatial_merge_size);
```
Also, `tokens.reserve(patch_rows * patch_cols + patch_rows)` uses `int` arithmetic — while practically safe for typical image sizes, `size_t` would be more defensive.

src/models/mistral3_image_processor.cpp
```diff
@@ -0,0 +1,248 @@
...
+  // Tokenize the text prompt
+  std::vector<int32_t> input_ids;
+  if (!prompt.empty()) {
+    input_ids = tokenizer.Encode(prompt.c_str());
```
tianleiwu, line 98: **Nitpick**: `num_img_tokens` only counts `[IMG]` tokens (not `[IMG_BREAK]`/`[IMG_END]`), which is correct for the embedding model's vision feature count. A brief comment here would help future maintainers understand why `count(... == img_token_id)` is used rather than `image_token_sequence.size()`.

tianleiwu (commented):
## Review Summary

Well-structured addition of Mistral3/Pixtral VLM support with clean separation of concerns across C++ runtime and Python model builder. The YaRN RoPE bug fixes (dict-based `rope_scaling`, `rope_theta` fallback, `mscale` override, `inv_freq` double-inversion) are important correctness improvements, well-validated by 42 tests with guard assertions.

**Positives:**
- Clean `BuildImageTokenSequence`/`ProcessPixtralPrompt` decomposition in the C++ processor
- Three-level fallback for image token insertion handles diverse prompt formats gracefully
- `isinstance(Mapping)` is the correct fix for dict-based `rope_scaling` inspection
- `self.max_length` defaults to `self.context_length` — zero behavior change for existing models
- FP8 dequantization with proper `ValueError` for missing `scale_inv`
- Excellent test coverage with guard assertions preventing vacuously-passing tests

A few minor suggestions below — no blocking issues found.

src/python/py/models/builders/mistral.py
```diff
@@ -3,9 +3,90 @@
...
+        # the runtime can locate image tokens in the input sequence.
+        config_path = os.path.join(out_dir, "genai_config.json")
+        with open(config_path) as f:
+            genai_config = json.load(f)
+
+        from transformers import AutoConfig
+
+        hf_config = AutoConfig.from_pretrained(
+            model_name_or_path,
+            token=self.hf_token,
```
tianleiwu: **Suggestion**: `make_genai_config` calls `AutoConfig.from_pretrained` to fetch `image_token_id`, but the HF config is already loaded earlier in the pipeline (by `builder.py`). Consider reusing the config object (e.g., storing it in `self`) to avoid a redundant network call to HuggingFace.

test/python/test_mistral3_preprocessor.py
```diff
@@ -0,0 +1,254 @@
...
+        hf_result = hf_processor(images=img, return_tensors="np")
+        hf_pv = self._unwrap_hf_pixels(hf_result)
+
+        our_h, our_w = smart_resize(720, 1280)
+        assert hf_pv.shape[1] == our_h, f"Height mismatch: HF={hf_pv.shape[1]}, ours={our_h}"
+        assert hf_pv.shape[2] == our_w, f"Width mismatch: HF={hf_pv.shape[2]}, ours={our_w}"
+
+    def test_pixel_values_match(self, hf_processor):
+        """Our preprocessing produces same values as HF (within FP tolerance)."""
+        img = _create_test_image(1280, 720, color=(100, 150, 200))
```
tianleiwu, line 217: **Nitpick**: Hardcoded developer-local path `"/home/titaiwang/fish.jpg"`. The test correctly skips when the file is absent, but consider using an environment variable (e.g., `MISTRAL3_TEST_IMAGE`) for consistency with the pattern in `test_mistral3_tokens.py`.

titaiwangms:
Addressed all 4 review comments from @tianleiwu in da3379d6:

1. **int64_t division order** (line 92): Division now happens in int64_t space before narrowing to int: `static_cast<int>(image_h / patch_size / spatial_merge_size)`
2. **num_img_tokens comment** (line 96): Added clarifying comment explaining why we count only [IMG] tokens (vision feature count for embedding model)
3. **Redundant AutoConfig call** (mistral.py): Cached `image_token_id` from HF config during `__init__`, removed redundant `AutoConfig.from_pretrained` network call from `make_genai_config`
4. **Hardcoded test image path** (preprocessor test): Now uses `os.environ.get("MISTRAL3_TEST_IMAGE", ...)` matching the pattern in test_mistral3_tokens.py

All 41 tests pass.

titaiwangms:
Updated d3325a78: Improved the test image path fix — removed the hardcoded fallback entirely. Now uses only the env var with no developer-local path:
```python
image_path = os.environ.get("MISTRAL3_TEST_IMAGE")
if not image_path or not os.path.exists(image_path):
    pytest.skip("MISTRAL3_TEST_IMAGE not set or file not found")
```
Matches the existing pattern in test_mistral3_tokens.py. CI will skip cleanly.

src/python/py/models/builders/mistral.py
```diff
@@ -3,9 +3,84 @@
...
+            genai_config["model"]["image_token_id"] = self.image_token_id
+
+            with open(config_path, "w") as f:
+                json.dump(genai_config, f, indent=4)
+
+    def load_weights(self, input_path):
+        if self.quant_type is not None or input_path.endswith(".gguf"):
+            return super().load_weights(input_path)
+
+        import torch
```
kunal-vaishnavi: Can we make these global imports instead?

src/python/py/models/builders/mistral.py
```diff
@@ -3,9 +3,84 @@
...
+        print("Loading Mistral3ForConditionalGeneration model...")
+        model = Mistral3ForConditionalGeneration.from_pretrained(
+            self.model_name_or_path,
+            cache_dir=self.cache_dir,
+            token=self.hf_token,
+            trust_remote_code=self.hf_remote,
+            **extra_kwargs,
+        )
+
+        # Dequantize FP8 weights in-place: dequantized = fp8_value * scale_inv
```
kunal-vaishnavi, line 63: Is dequantization not already handled by Hugging Face's `from_pretrained` method if you specify `torch_dtype`?

src/python/py/models/builder.py
```diff
@@ -246,32 +257,69 @@ def create_model(
...
-    elif config.architectures[0] == "Phi3ForCausalLM" and config.max_position_embeddings != config.original_max_position_embeddings:
+    elif (
+        config.architectures[0] == "Phi3ForCausalLM"
+        and config.max_position_embeddings != config.original_max_position_embeddings
+    ):
         onnx_model = Phi3MiniLongRoPEModel(config, io_dtype, onnx_dtype, execution_provider, cache_dir, extra_options)
-    elif config.architectures[0] == "PhiMoEForCausalLM" and config.max_position_embeddings != config.original_max_position_embeddings:
-        print("WARNING: This model only works for CUDA currently because `MoE` is only supported for CUDA in ONNX Runtime. Setting `--execution_provider cuda` by default.")
-        print("WARNING: This model currently only supports the quantized version. Setting `--precision int4` by default.")
+    elif (
```
kunal-vaishnavi: Can we revert all of these spacing changes as well? The only changes in this file should be related to the new model.

src/python/py/models/builders/mistral.py
titaiwangms: Done in 0b4a99b2. Moved `import torch` and `from transformers import Mistral3ForConditionalGeneration` to top-level imports.

src/python/py/models/builders/mistral.py
titaiwangms: Good question. HF's `from_pretrained` with `torch_dtype=` does handle dtype conversion, but the Ministral-3-3B checkpoint stores weights as `float8_e4m3fn` with per-tensor `weight_scale_inv`. HF's FP8 integration (`transformers.integrations.finegrained_fp8`) requires CUDA for the `w8a8_fp8_matmul` kernel — it fails on CPU with `ValueError: Expected a cuda device`. The manual dequant (`weight.to(bf16) * scale_inv`) allows CPU export. However, users can avoid this by using the `-BF16` HF variant (pre-dequantized weights, no FP8 handling needed). No code change needed.

src/python/py/models/builder.py
titaiwangms: Done in 0b4a99b2. Reset builder.py to origin/main and re-applied only the functional changes: Mistral3TextModel import + Mistral3ForConditionalGeneration dispatch (13 lines added, 0 cosmetic changes).

examples/python/common.py
```diff
@@ -469,7 +504,7 @@ def get_guidance(
     if tool_output:
         if os.path.exists(filepath):
             # If tools are provided as a file
-            with open(filepath, 'r') as f:
+            with open(filepath) as f:
```
kunal-vaishnavi: Why remove `'r'`?

src/python/py/models/builders/mistral.py
kunal-vaishnavi: If a BF16 variant already exists on Hugging Face, can we point users to convert that model to ONNX instead and remove any dequantization logic?

examples/python/common.py
titaiwangms: It looks like there is still some formatting issue I have not reverted. I will do it. Also, we don't have a linter in genai codebase?

src/python/py/models/builders/mistral.py
titaiwangms: The Ministral-3-3B-Instruct model ships with **native FP8 (fp8_e4m3) weights only** — there is no BF16 variant available on HuggingFace. The FP8 implementation uses Triton kernels that cannot be exported to ONNX, making dequantization to BF16 the only viable approach for ONNX export.

Benchmark validation confirms the pipeline is healthy:
- PyTorch CUDA BF16: 73.00% (AI2D, 200 samples)
- ONNX CUDA FP16: 71.65% (200 samples) — 1.35pp gap, within threshold
- ONNX is also 1.5× faster (0.13s vs 0.20s per sample)

The dequantization introduces no additional quality loss beyond what Mistral AI's original FP8 quantization introduced. The wider BF16 format simply stores the already-rounded values.

test/python/test_mistral3_preprocessor.py
```diff
@@ -0,0 +1,252 @@
...
+This module does not directly execute the C++ Mistral3ImageProcessor or the
+ort-extensions pipeline.
+
+Run with:
+    python -m pytest test/python/test_mistral3_preprocessor.py -v
+"""
+
+import numpy as np
+import pytest
+from pathlib import Path
```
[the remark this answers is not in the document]
titaiwangms: Fixed in b9bd58f5: removed unused `Path` import.

test/python/test_mistral3_preprocessor.py
```diff
@@ -0,0 +1,283 @@
...
+    3. Normalize: (pixel - CLIP_mean) / CLIP_std
+    4. Channel order: RGB, NCHW layout
+
+Run with:
+    python -m pytest test/python/test_mistral3_preprocessor.py -v
+"""
+
+import numpy as np
+import pytest
+from PIL import Image
```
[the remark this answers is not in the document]
titaiwangms: Already handled: the file uses `pytest.importorskip("PIL")` at module level (line 31), which skips the entire module if Pillow is unavailable.

test/python/test_mistral3_tokens.py
```diff
@@ -0,0 +1,296 @@
...
+        # Verify break count = rows - 1
+        num_img = int(np.sum(seq == IMG_TOKEN_ID))
+        num_break = int(np.sum(seq == IMG_BREAK_TOKEN_ID))
+        num_end = int(np.sum(seq == IMG_END_TOKEN_ID))
+        total_rows = num_break + num_end  # breaks + 1 end = rows
+        assert num_img == total_rows * (num_img // total_rows), (
+            f"[IMG] count {num_img} not divisible by grid rows {total_rows}"
+        )
+
+        del generator
```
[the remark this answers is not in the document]
titaiwangms: Fixed in b9bd58f5: kept the `del generator` with an explanatory comment — it ensures deterministic C++ generator cleanup before model teardown, which matters for GPU memory management.

test/python/test_mistral3_preprocessor.py
```diff
@@ -0,0 +1,283 @@
...
+
+Verifies that the preprocessing pipeline used by the C++ Mistral3ImageProcessor
+(via ort-extensions processor_config.json) produces pixel values that match
+the HuggingFace PixtralImageProcessor reference implementation:
+
+    1. Smart resize: snap to multiples of patch_size × merge_size (28)
+    2. Rescale: pixel / 255.0
+    3. Normalize: (pixel - CLIP_mean) / CLIP_std
+    4. Channel order: RGB, NCHW layout
+
```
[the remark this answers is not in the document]
titaiwangms: Fixed in b9bd58f5: updated the docstring to clarify that these tests validate the Python reference preprocessing logic that feeds into the C++ pipeline, and optionally compare against HuggingFace.

test/python/test_mistral3_preprocessor.py
```diff
@@ -0,0 +1,252 @@
...
+        assert arr[0, 0, 0] > 0, "Channel 0 should be R (positive for red image)"
+        assert arr[1, 0, 0] < 0, "Channel 1 should be G (negative for red image)"
+        assert arr[2, 0, 0] < 0, "Channel 2 should be B (negative for red image)"
+
+
+class TestMatchHuggingFace:
+    """Compare our preprocessing against HuggingFace PixtralImageProcessor."""
+
+    @pytest.fixture
+    def hf_processor(self):
```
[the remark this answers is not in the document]
titaiwangms: Fixed in b9bd58f5: added explicit `return None` after `pytest.skip()` to satisfy CodeQL mixed-returns check.

test/python/test_mistral3_tokens.py
```diff
@@ -0,0 +1,211 @@
...
+
+        model_path = "/datadisks/disk1/titaiwang/ministral3-vlm-dynamic-final"
+        model = og.Model(model_path)
+        tokenizer = og.Tokenizer(model)
+        processor = model.create_multimodal_processor()
+
+        images = og.Images.open("/home/titaiwang/fish.jpg")
+        prompt = "<s>[INST][IMG]Describe this image.[/INST]"
+        inputs = processor(prompt, images=images)
+
```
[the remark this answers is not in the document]
titaiwangms: The integration test uses the `test_data_path` fixture from conftest.py (configured via `--test_models` CLI option), not hardcoded paths. The test also uses `pytest.skip()` if required files are missing, so it gracefully skips in environments without the model.

test/python/test_mistral3_tokens.py
```diff
@@ -0,0 +1,211 @@
...
+Token IDs: [IMG]=10, [IMG_BREAK]=12, [IMG_END]=13
+
+Run with:
+    python -m pytest test/python/test_mistral3_tokens.py -v
+"""
+
+import pytest
+
+# Pixtral special token IDs from the Ministral-3-3B tokenizer.
+# These are validated at runtime in test_genai_processor_token_counts.
```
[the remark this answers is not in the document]
titaiwangms: Fixed in b9bd58f5: updated the comment to clarify that the integration test validates C++ processor token sequence output (correct counts of [IMG], [IMG_BREAK], [IMG_END]), not the tokenizer-to-ID mapping itself.

src/python/py/models/builders/mistral.py
```diff
@@ -3,9 +3,83 @@
...
+        fp8_count = 0
+        for name, module in model.named_modules():
+            if isinstance(module, torch.nn.Linear) and module.weight.dtype == torch.float8_e4m3fn:
+                scale_inv = getattr(module, "weight_scale_inv", None)
+                if scale_inv is not None:
+                    module.weight = torch.nn.Parameter(
+                        module.weight.to(torch.bfloat16) * scale_inv,
```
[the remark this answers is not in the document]
titaiwangms: Fixed in b9bd58f5: both `module.weight` and `scale_inv` are now explicitly cast to `torch.bfloat16` before multiplication, preventing unintended promotion to float32. Also added `.reshape(-1, 1)` for correct per-channel broadcasting.

src/models/mistral3_image_processor.cpp
```diff
@@ -0,0 +1,272 @@
...
+    num_images = shape[0];
+    if (num_dims == 4 && shape[3] <= 4 && shape[1] > 4) {
+      // NHWC format (last dim is channels: 3 or 4)
+      image_h = shape[1];
+      image_w = shape[2];
+    } else {
+      // NCHW format
+      image_h = shape[2];
+      image_w = shape[3];
+    }
```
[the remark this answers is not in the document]
titaiwangms: Already addressed: `num_dims == 4` validation was added at line 80 — throws a clear error if pixel_values is not 4D [N,C,H,W].

src/models/mistral3_image_processor.cpp
```diff
@@ -0,0 +1,279 @@
...
+    int64_t padded_h = shape[2];
+    int64_t padded_w = shape[3];
+
+    // Read per-image sizes if available, otherwise use padded dimensions
+    const int64_t* sizes_data = nullptr;
+    if (image_sizes_tensor) {
+      const void* raw{};
+      const int64_t* sizes_shape{};
+      size_t sizes_dims{};
+      CheckResult(OrtxGetTensorData(image_sizes_tensor, &raw, &sizes_shape, &sizes_dims));
```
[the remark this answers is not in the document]
titaiwangms: Already addressed: validation for `sizes_dims == 2`, `sizes_shape[1] == 2`, and batch dimension match was added at lines 85-98.

src/models/mistral3_image_processor.cpp
```diff
@@ -0,0 +1,272 @@
...
+        for (int64_t h = 0; h < H; ++h) {
+          for (int64_t w = 0; w < W; ++w) {
+            for (int64_t c = 0; c < C; ++c) {
+              auto dst_idx = (n * C * H * W + c * H * W + h * W + w) * element_size;
+              auto src_idx = (n * H * W * C + h * W * C + w * C + c) * element_size;
+              std::memcpy(dst + dst_idx, src + src_idx, element_size);
+            }
+          }
+        }
+      }
```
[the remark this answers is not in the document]
titaiwangms: Acknowledged. The transpose runs once during preprocessing (not per-token), so the cost is amortized. Per kunal's comment below, the ideal long-term fix is moving the transpose into ort-extensions so the pipeline outputs NCHW directly. Keeping the current approach for now as it's correct and the ort-extensions change is a separate PR.

src/models/multi_modal.cpp
```diff
@@ -359,6 +360,177 @@ DeviceSpan<float> QwenVisionState::Run(int current_length, DeviceSpan<int32_t>&
...
+  if (model_.config_->model.vision.run_options.has_value()) {
+    State::SetRunOptions(model_.config_->model.vision.run_options.value());
+  }
+
+  // Single image or no per-image size metadata: run directly.
+  if (num_images_ <= 1 || image_heights_.empty()) {
+    State::Run(*model_.vision_session_);
+    return {};
+  }
+
```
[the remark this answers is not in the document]
titaiwangms: Already addressed: the early return now only fires for `num_images_ <= 1` (single image). For `num_images_ > 1` without `image_sizes` metadata, we throw an explicit error: "multi-image inputs require image_sizes metadata" (line ~471).

src/models/mistral3_image_processor.cpp
```diff
@@ -0,0 +1,258 @@
...
+
+  if (num_images > 0 && image_h > 0 && image_w > 0) {
+    patch_rows = static_cast<int>(image_h) / patch_size / spatial_merge_size;
+    patch_cols = static_cast<int>(image_w) / patch_size / spatial_merge_size;
+    image_token_sequence = BuildImageTokenSequence(patch_rows, patch_cols,
+                                                   img_token_id, img_break_id, img_end_id);
+    num_img_tokens = static_cast<int64_t>(std::count(image_token_sequence.begin(),
+                                                     image_token_sequence.end(),
+                                                     img_token_id));
+  }
```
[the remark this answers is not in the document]
titaiwangms: This is a false positive. The function iterates per-image in the `for (int64_t i = 0; i < num_images; ++i)` loop (line 103), computing per-image `patch_rows`/`patch_cols` from individual image dimensions. Each image gets its own `PerImageInfo` with correct token count. The `total_img_tokens` is the sum across all images.

src/models/multi_modal.cpp
```diff
@@ -359,6 +360,177 @@ DeviceSpan<float> QwenVisionState::Run(int current_length, DeviceSpan<int32_t>&
...
+          ") exceeds pre-allocated feature buffer (" + std::to_string(total_feats) + ")");
+
+    // Create output sub-tensor view into the pre-allocated feature buffer.
+    std::vector<int64_t> sub_feat_shape = {num_feats, hidden_size};
+    auto sub_feat = OrtValue::CreateTensor(
+        *cpu_mem,
+        feat_raw + static_cast<size_t>(feat_offset * hidden_size) * feat_elem_size,
+        static_cast<size_t>(num_feats * hidden_size) * feat_elem_size,
+        std::span<const int64_t>(sub_feat_shape), feat_type);
+
```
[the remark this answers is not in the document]
titaiwangms: Already addressed: the sub-tensor view now uses `feat_full->GetTensorMemoryInfo()` to match the underlying buffer's allocation (CPU or GPU). This was fixed in the previous commit iteration.

test/virtual_dispatch_test.cpp
```diff
@@ -0,0 +1,64 @@
...
+namespace Generators::test {
+
+// Verify VisionState is polymorphic (has at least one virtual function).
+static_assert(std::is_polymorphic_v<VisionState>,
+              "VisionState must be polymorphic — SetExtraInputs should be virtual");
+
+// Verify PixtralVisionState is polymorphic (inherits from VisionState).
+static_assert(std::is_polymorphic_v<PixtralVisionState>,
+              "PixtralVisionState must be polymorphic");
+
```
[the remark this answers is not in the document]
titaiwangms: Acknowledged. The `is_polymorphic_v` check tests for any virtuals, not specifically `SetExtraInputs`. The real compile-time guard is the `override` keyword on `PixtralVisionState::SetExtraInputs` — if the base declaration is no longer virtual, the override fails to compile. The static_asserts document the type hierarchy assumption.

src/python/py/models/builders/mistral.py
```diff
@@ -3,9 +3,81 @@
...
+    is architecturally identical to MistralModel. This builder loads
+    the full VLM and dequantizes FP8 weights if present.
+    """
+
+    def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
+        super().__init__(config, io_dtype, onnx_dtype, ep, cache_dir, extra_options)
+        # Cache image_token_id from the HF config to avoid a redundant network
+        # call in make_genai_config (the config is already loaded by builder.py).
+        self.image_token_id = getattr(config, "image_token_id", None)
+
```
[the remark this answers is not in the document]
titaiwangms: Addressed: the `max_length` override was reverted. `max_length` is now controlled at runtime via `search_options` in eval.py/model-mm.py, not hard-coded in the model config.

src/models/mistral3_image_processor.cpp
titaiwangms: Agreed — the ideal fix is adding NCHW output support in ort-extensions so the transpose is unnecessary. The current in-GenAI transpose is a correctness shim until that ort-extensions change lands. Happy to file a tracking issue for the ort-extensions work.

src/models/mistral3_image_processor.cpp
titaiwangms: Fixed in b9bd58f5: the division is now done entirely in `int64_t` space using `effective_patch` (= `patch_size * spatial_merge_size`), then narrowed to `int` only for the result: `static_cast<int>(h / effective_patch)`. This avoids any intermediate truncation.

src/python/py/models/builder.py
titaiwangms: `exclude_embeds` is set unconditionally for `Mistral3ForConditionalGeneration` in the `create_model` dispatch (builder.py line 256). It's always True for this architecture since the embedding model is exported separately via the VLM pipeline. Other architectures that go through `create_model` don't set it, so the `if` guard in the base class is still needed for backward compatibility.

src/models/mistral3_image_processor.cpp
titaiwangms: Fixed in b9bd58f5: added a comment explaining that `num_img_tokens` counts only `[IMG]` tokens (= vision feature count = patch_rows × patch_cols), excluding structural `[IMG_BREAK]`/`[IMG_END]` tokens.

src/python/py/models/builders/mistral.py
titaiwangms: Already addressed: `Mistral3TextModel.__init__` now caches `image_token_id` from the config object (`getattr(config, "image_token_id", None)`) which is already loaded by builder.py. The `make_genai_config` method uses `self.image_token_id` directly — no redundant `AutoConfig.from_pretrained` call.

test/python/test_mistral3_preprocessor.py
titaiwangms: No hardcoded developer-local paths exist in this file. All test images are generated programmatically via `_create_test_image()`. The HF comparison tests use `pytest.skip` when HuggingFace/model is unavailable.

src/python/py/models/builders/mistral.py
```diff
@@ -3,9 +3,91 @@
...
+        # Mistral3 models can advertise extremely large context windows
+        # (e.g. 262144 for Ministral-3-3B). Using that as default max_length
+        # in genai_config.json causes oversized buffer allocations and OOMs.
+        # Honor an explicit override first, otherwise clamp to a practical ceiling.
+        configured_max_length = self.extra_options.get("max_length")
+        if configured_max_length is not None:
+            self.max_length = int(configured_max_length)
```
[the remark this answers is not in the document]
titaiwangms: Resolved: the `max_length` override was reverted in b9bd58f5. No attribute overwriting remains.

src/python/py/models/builders/mistral.py
```diff
@@ -3,9 +3,91 @@
...
+        elif self.max_length is None or self.max_length > 32768:
+            self.max_length = 32768
```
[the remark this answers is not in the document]
titaiwangms: Resolved: the `max_length` override was reverted in b9bd58f5. No attribute overwriting remains.

examples/python/common.py
titaiwangms: Reverted: builder.py was reset to origin/main with only functional Mistral3 changes (import + dispatch). No formatting changes remain.

src/models/multi_modal.cpp
```diff
@@ -359,6 +360,183 @@ DeviceSpan<float> QwenVisionState::Run(int current_length, DeviceSpan<int32_t>&
...
+  }
+
+  if (static_cast<int64_t>(image_heights_.size()) < num_images_)
+    throw std::runtime_error(
+        "PixtralVisionState: image_heights_ has " + std::to_string(image_heights_.size()) +
+        " entries but num_images_ is " + std::to_string(num_images_));
+
+  // Multi-image: pixel_values is [N, C, H_max, W_max] with zero-padding.
+  // image_heights_/image_widths_ hold the actual per-image dimensions.
+  // Run vision.onnx once per image with [1, C, H_i, W_i].
```
kunal-vaishnavi, line 424: This is running the vision model on each image. Why can't we make the vision ONNX model support batched inputs? This is quite inefficient otherwise.

src/python/py/models/builders/mistral.py
```diff
@@ -3,9 +3,81 @@
...
+        super().__init__(config, io_dtype, onnx_dtype, ep, cache_dir, extra_options)
+        # Cache image_token_id from the HF config to avoid a redundant network
+        # call in make_genai_config (the config is already loaded by builder.py).
+        self.image_token_id = getattr(config, "image_token_id", None)
+
+    def make_genai_config(self, model_name_or_path, extra_kwargs, out_dir):
+        super().make_genai_config(model_name_or_path, extra_kwargs, out_dir)
+
+        # Patch the generated genai_config.json with image_token_id so that
+        # the runtime can locate image tokens in the input sequence.
```
kunal-vaishnavi: Is this not something we can set in the GenAI config when the parent method is called? That would be cleaner than reading the file, setting an attribute, and then saving the file again.

src/models/multi_modal.cpp
titaiwangms: Good question! The per-image execution is intentional due to Pixtral's variable-resolution architecture. Unlike CLIP-based vision encoders that resize all images to a fixed resolution (e.g. 224x224), Pixtral preserves native aspect ratios — each image is tiled into patches at its original resolution, producing tensors with different H and W dimensions per image.

The vision encoder uses 2D RoPE positional embeddings computed from actual spatial dimensions, so feeding a padded batch tensor would produce incorrect positional encodings and wasted compute on padding tokens.

Batching would require either (a) padding all images to H_max x W_max with attention masking (the vision transformer doesn't currently support attention masks), or (b) restructuring the ONNX model to accept flattened/concatenated patch sequences with per-image position IDs — a larger architectural change.

This is consistent with the HuggingFace reference implementation. Happy to file a tracking issue for batched execution as a follow-up optimization.

src/python/py/models/builders/mistral.py
titaiwangms: Agreed this isn't the cleanest pattern. The base class's `make_genai_config` (in base.py) constructs the genai_config dict as a local variable, writes it to disk, and doesn't return it or provide a hook for subclasses to inject additional fields before the write.

`image_token_id` is a VLM-specific field that the base text-decoder builder has no awareness of, so there's no existing slot for it.

To do this more cleanly, we could refactor the base class to add an overridable `_extra_model_config()` method that subclasses can use to inject additional keys. That said, this refactor touches the base builder used by all model types, so I kept the scope narrow for this PR. Happy to do the refactor here or as a follow-up — what do you prefer?

src/models/mistral3_image_processor.cpp
```diff
@@ -0,0 +1,301 @@
...
+      if (input_ids[i] == img_token_id && next_image < image_infos.size()) {
+        // Replace this [IMG] (and consecutive [IMG] tokens) with the image's token sequence
+        expanded_ids.insert(expanded_ids.end(),
+                            image_infos[next_image].token_sequence.begin(),
+                            image_infos[next_image].token_sequence.end());
+        ++next_image;
+        // Skip consecutive [IMG] tokens from the original prompt
+        while (i + 1 < input_ids.size() && input_ids[i + 1] == img_token_id) {
+          ++i;
+        }
```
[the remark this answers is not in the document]
titaiwangms: This is a false positive — the reviewer misunderstands the Pixtral tokenization scheme. In the input prompt (pre-expansion), each image is represented by a single `[IMG]` placeholder token. The long sequences of `[IMG]` tokens (with `[IMG_BREAK]` between rows and `[IMG_END]` at the end) are the OUTPUT of `BuildImageTokenSequence()` — they are what each single `[IMG]` placeholder gets EXPANDED into.

The while loop at line 162 is a defensive guard for the case where the tokenizer emits multiple `[IMG]` token IDs for a single `[IMG]` text string. It operates on the pre-expansion input_ids, where consecutive `[IMG]` tokens would never represent separate images. Separate images are always separated by other tokens in the prompt.

src/models/multi_modal.cpp
titaiwangms: To add more context with concrete examples of how other VLMs handle this:

1. **Fixed-size (e.g. Phi-4-multimodal)** — resizes all images to the same fixed dimensions, so batching is straightforward since all tensors have identical shapes.
2. **Grid-based (e.g. Qwen2-VL)** — tiles images into fixed-size grid cells. The grid cells can be batched together even though original images vary in size.
3. **Variable-resolution (Pixtral/Ministral)** — preserves native aspect ratios with variable H x W per image. There's no fixed grid or resize step, so tensors have genuinely different spatial dimensions and can't be stacked without padding + attention masking.

Pixtral falls into category 3, which is why per-image execution is the current approach. Batching would require adding padding to max dimensions and attention mask support in the vision transformer — doable as a follow-up optimization.

src/models/multi_modal.cpp
```diff
@@ -359,6 +360,183 @@ DeviceSpan<float> QwenVisionState::Run(int current_length, DeviceSpan<int32_t>&
...
+  if (image_heights_.empty() || image_widths_.empty()) {
+    throw std::runtime_error(
+        "PixtralVisionState: multi-image inputs require image_sizes metadata");
+  }
+
+  if (static_cast<int64_t>(image_heights_.size()) < num_images_)
+    throw std::runtime_error(
+        "PixtralVisionState: image_heights_ has " + std::to_string(image_heights_.size()) +
+        " entries but num_images_ is " + std::to_string(num_images_));
+
```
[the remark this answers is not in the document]
titaiwangms: Fixed in 085bee33 — added `image_widths_.size() < num_images_` validation mirroring the existing heights check.

src/python/py/models/builders/mistral.py
titaiwangms: Addressed in d89140c3. Added an `update_genai_config()` hook to the base builder class (base.py) that subclasses can override to inject model-specific config fields before the file is written. Mistral3 now uses a clean 3-line override instead of the read-modify-write pattern.

src/models/mistral3_image_processor.cpp
```diff
@@ -0,0 +1,301 @@
...
+  CheckResult(OrtxImagePreProcess(processor_.get(), images->images_.get(),
+                                  result.ToBeAssigned()));
+
+  OrtxTensor* pixel_values = nullptr;
+  CheckResult(OrtxTensorResultGetAt(result.get(), 0, &pixel_values));
+
+  // Tensor 1: image_sizes[N, 2] from PixtralImageSizes (post-resize, pre-padding).
+  // Only present when processor_config.json includes the PixtralImageSizes step.
+  OrtxTensor* image_sizes = nullptr;
+  if (OrtxTensorResultGetAt(result.get(), 1, &image_sizes) != kOrtxOK) {
```
kunal-vaishnavi: Would it be simpler to do the following?

```suggestion
CheckResult(OrtxTensorResultGetAt(result.get(), 1, &image_sizes));
```

src/models/multi_modal.cpp
kunal-vaishnavi: Let's add a todo as a follow-up optimization.

src/python/py/models/builders/base.py
```diff
@@ -667,10 +667,15 @@ def make_genai_config(self, model_name_or_path, extra_kwargs, out_dir):
...
             genai_config["model"]["decoder"]["session_options"]["provider_options"].append(ep_options)
 
+        self.update_genai_config(genai_config)
+
         print(f"Saving GenAI config in {out_dir}")
         with open(os.path.join(out_dir, "genai_config.json"), "w") as f:
             json.dump(genai_config, f, indent=4)
 
+    def update_genai_config(self, genai_config):
+        """Override in subclasses to modify genai_config before it is written to disk."""
```
kunal-vaishnavi, line 677:
```suggestion
        """Override in subclasses to modify genai_config before it is written to disk."""
        pass
```

kunal-vaishnavi (dismissed):

kunal-vaishnavi (approved):

PR #2083, opened 2026-04-13
[WebGPU] Support continuous decoding (RewindTo) with graph capture

qjia7:
This pull request introduces improvements to the handling of attention masks in both the CUDA and WebGPU backends, focusing on more efficient and correct updates of mask buffers during decoding. The main changes are the implementation of a CPU-side update for static attention masks in CUDA and the addition of a reusable staging buffer for efficient mask updates in WebGPU, with logic to avoid redundant work for single-beam cases.

**CUDA backend improvements:**

* Replaced the previous (commented-out and incorrect) CUDA memory set logic in `DefaultPositionInputs::RewindMask` with a CPU-side update that correctly sets attended and non-attended positions in the attention mask for each batch/beam, followed by a copy back to the device. This ensures the mask is set with 1s for attended tokens and 0s for future tokens, supporting both `int32_t` and `int64_t` types.

**WebGPU backend improvements:**

* Added a reusable CPU staging buffer (`mask_staging_buffer_`) to the `InterfaceImpl` struct for efficient attention mask updates, avoiding repeated allocations and redundant writes.
* Implemented the `UpdateAttentionMask` method to efficiently update the mask for single-beam cases by only filling new positions with 1s and copying the relevant portion to the device, falling back to CPU for multi-beam cases. This method handles static update path and supports both `int32_t` and `int64_t` mask types.


src/models/position_inputs.cpp
```diff
@@ -381,19 +381,30 @@ void DefaultPositionInputs::InitializeSequenceLengths(std::array<int64_t, 2> sha
 
 void DefaultPositionInputs::RewindMask(size_t index) {
```
kunal-vaishnavi, line 382: From Copilot:

---

`RewindMask()` is **incomplete and currently only does anything when `use_graph_capture` is enabled**. In the non-graph-capture path it silently returns without rewinding the mask, which makes `DefaultPositionInputs::RewindTo()` effectively a no-op for the attention mask in the common “dynamic mask” mode.

Here’s a focused review of the implementation in your file:

```c++ name=src/models/position_inputs.cpp url=https://github.com/microsoft/onnxruntime-genai/blob/3db347e66a450c80d20deffa2ff608967d26422b/src/models/position_inputs.cpp#L382-L409
void DefaultPositionInputs::RewindMask(size_t index) {
  if (state_.params_->use_graph_capture) {
    // Static mask layout: [batch_beam_size, max_length]
    // Rewind to index: write 1s for [0, index), 0s for [index, max_length)
    ...
    byte_span.CopyCpuToDevice();
    return;
  }
}
```

## What’s good
- **Correct intent for static mask**: for graph capture they keep a fixed `[batch_beam_size, max_length]` tensor and overwrite it to represent the rewound length (`[0, index)` = 1, rest 0). That matches how a “static mask” should be updated.
- **Bounds check**: throws if `index > max_length`.
- **Handles both int32/int64**: consistent with the earlier `type_` validation.

## Issues / risks
1. **No behavior when `use_graph_capture == false`**
   - The function just falls off the end with no action.
   - But `DefaultPositionInputs::RewindTo()` calls `RewindMask(index)` when `has_mask_input_` and `batch_size == 1` (see below), so callers will think rewind succeeded when it didn’t.
   - This is especially problematic because the *dynamic* mask path (`ShouldUseStaticMaskHandling() == false`) stores a mask whose second dimension tracks the current length, and rewinding should either:
     - shrink/recreate the tensor to `[batch_beam, index]`, or
     - at least ensure future updates rebuild from the rewound length.

2. **Mismatch between caller condition and mask mode**
   - `RewindTo()` decides to call `RewindMask()` only based on `use_graph_capture`? No—it checks `attention_mask_shape_[0] == 1`, then calls `RewindMask(index)` regardless of whether the mask handling is static/dynamic.
   - But `RewindMask()` only updates for `use_graph_capture`.
   - Net: `RewindTo(index>0)` for batch=1 in non-graph-capture mode does nothing.

3. **Potentially confusing memory pathway**
   - It uses `attention_mask_->GetByteSpan()` then `CpuSpan()` and `reinterpret_cast` to `int32_t*`/`int64_t*`.
   - This can be OK if `CpuSpan()` is guaranteed to be host-accessible backing storage for that tensor (e.g., unified/pinned staging). If `CpuSpan()` is just a view into an internal CPU buffer that requires an explicit `CopyDeviceToCpu()` first, then this would be wrong. (In other parts of the file, the dynamic update path explicitly does `CopyDeviceToCpu()` before mutation.)

4. **`max_len` type mixing**
   - `max_len` is `size_t`, but later arithmetic uses `int` for the outer loop and `std::fill_n` takes a `size_t` count. This is mostly fine, but keep it consistent to avoid signed/unsigned pitfalls.

## What I’d change (minimum fix)
- Make `RewindMask()` handle **both** static and dynamic mask modes:
  - If `ShouldUseStaticMaskHandling()` (not just `use_graph_capture`), do the current “overwrite [0,index)=1 else 0” behavior with max_length.
  - Else (dynamic), either:
    1) **Recreate** `attention_mask_` to shape `[batch_beam_size, index]` and fill with ones (since batch_size==1 implies no padding), then update `state_.inputs_[mask_input_index_]`, *or*
    2) Set `is_first_update_ = true` and adjust `attention_mask_shape_` so the next `Update()` will rebuild properly (but that may not be sufficient if downstream expects the mask tensor immediately after rewind).

qjia7:
Addressed the review comments in 80be21ca:

**1. `RewindMask()` now handles both static and dynamic mask paths**

- **Static path** (`ShouldUseStaticMaskHandling()` — graph capture or NvTensorRtRtx with shared buffers): overwrites the fixed `[batch_beam_size, max_length]` tensor with 1s for `[0, index)` and 0s for `[index, max_length)`, then copies to device. Same as before but now gated on `ShouldUseStaticMaskHandling()` instead of just `use_graph_capture`, consistent with `UpdateAttentionMask()`.

- **Dynamic path** (everything else): sets `attention_mask_shape_[1] = index`. The next `Update()` call will:
  1. `CreateNextAttentionMaskTensor(total_length)` — creates a correctly sized tensor using the adjusted shape
  2. CPU `UpdateAttentionMask` for `batch_beam_size == 1` — fills the entire new mask with 1s (ignores old data)
  3. Replaces `attention_mask_` with `attention_mask_next_`

  No tensor recreation or `is_first_update_` reset needed — `RewindTo` already restricts to `batch_size == 1` (no padding), so a shape-only adjustment is sufficient.

**2. Fixed `max_len` type mixing** — `batch_beam_size`, `max_len`, and loop indices now all use `size_t` consistently.

**Tested on WebGPU (NVIDIA RTX):**
- Static mask (phi4-graph-prune, `enableGraphCapture=1`): :white_check_mark:
- Dynamic mask (phi4-prune, `past_present_share_buffer=false`): :white_check_mark:

Both tests: generate 10 tokens -> `rewind_to(prompt_len + 5)` -> generate 5 more tokens. Output is consistent and coherent after rewind.

kunal-vaishnavi (dismissed):

src/models/position_inputs.cpp
```diff
@@ -380,21 +380,37 @@ void DefaultPositionInputs::InitializeSequenceLengths(std::array<int64_t, 2> sha
...
-                    (type_ == Ort::TypeToTensorType<int32_t> ? sizeof(int32_t) : sizeof(int64_t)) * max_length,
-                    model_.cuda_stream_);
-    cudaMemsetAsync(attention_mask_->GetTensorMutableRawData(),
-                    1,
-                    (type_ == Ort::TypeToTensorType<int32_t> ? sizeof(int32_t) : sizeof(int64_t)) * past_length,
-                    model_.cuda_stream_);
-#endif
+  if (ShouldUseStaticMaskHandling()) {
+    // Static mask layout: [batch_beam_size, max_length]
+    // Rewind to index: write 1s for [0, index), 0s for [index, max_length)
```
baijumeswani, line 385: Does this change work for cuda as well?

src/webgpu/interface.cpp
```diff
@@ -171,6 +171,11 @@ struct InterfaceImpl : DeviceInterface {
  private:
   Ort::Allocator* ort_allocator_{};
   const OrtMemoryInfo* ort_memory_info_{};
+  // Reusable CPU staging buffers for UpdateAttentionMask, pre-filled with 1s.
+  // Content is always all 1s so sharing across generators is safe; only upload_bytes
+  // worth of data is copied each call, regardless of buffer capacity.
+  std::vector<int32_t> mask_staging_buffer_i32_;
+  std::vector<int64_t> mask_staging_buffer_i64_;
```
baijumeswani, line 178: Why do these buffers need to be shared across generators?

src/models/position_inputs.cpp
qjia7: Currently, cuda doesn't satisfy ShouldUseStaticMaskHandling(). But once the graph capture works on cuda. I think it should also work on CUDA.

src/webgpu/interface.cpp
qjia7: Currently, webgpu doesn't support multi-generators when graph capture on. But even it supports, it doesn't impact the correctness. `mask_staging_buffer_i32_` is just to avoid repeated to allocated and assign 1s. I think it may be better to put it here instead of UpdateAttentionMask

src/models/position_inputs.cpp
qjia7: Comment inserted above DefaultPositionInputs::ShouldUseStaticMaskHandling() to explain what kind of situations will trigger the static path.

src/webgpu/interface.cpp
```diff
@@ -190,6 +195,44 @@ struct InterfaceImpl : DeviceInterface {
...
+      }
+      staging_data = mask_staging_buffer_i64_.data();
+      upload_bytes = num_elements * sizeof(int64_t);
+    }
+
+    int64_t shape_val = static_cast<int64_t>(upload_bytes);
+    std::span<const int64_t> shape{&shape_val, 1};
+    auto cpu_mem_info = OrtMemoryInfo::CreateCpu(OrtDeviceAllocator, OrtMemTypeDefault);
+    auto src_tensor = OrtValue::CreateTensor(*cpu_mem_info, staging_data, upload_bytes, shape, ONNX_TENSOR_ELEMENT_DATA_TYPE_UINT8);
+    auto dst_tensor = OrtValue::CreateTensor(*ort_memory_info_, mask_data, upload_bytes, shape, ONNX_TENSOR_ELEMENT_DATA_TYPE_UINT8);
```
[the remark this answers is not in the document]
qjia7: Add (void) casts for unused params (next_mask_data, new_kv_length, max_length) and validate type is int32 or int64. Since the data is in gpu not cpu, we can't only update new_kv_length without reading data back. And webgpu doesn't support CopyTensors with offset > 0.

test/test_models/webgpu/tiny-graph-capture-gqa/create_model.py
```diff
@@ -0,0 +1,221 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation.  All rights reserved.
+# Licensed under the MIT License.  See License.txt in the project root for
+# license information.
+# --------------------------------------------------------------------------
+"""
+Create a minimal 1-layer GQA decoder model for testing GenAI graph capture + RewindTo.
```
kunal-vaishnavi: Instead of having a new script to create a model, can we just use `--extra_options num_hidden_layers=1` inside the model builder and call the model builder directly to create the assets needed to test inside the CIs?

src/webgpu/interface.cpp
```diff
@@ -190,6 +195,50 @@ struct InterfaceImpl : DeviceInterface {
...
+    const std::vector<const OrtValue*> src_ptrs = {src_tensor.get()};
+    const std::vector<OrtValue*> dst_ptrs = {dst_tensor.get()};
```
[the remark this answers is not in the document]
qjia7: Cache cpu_mem_info is done. std::array is not supported in genai's CopyTensors interface. Can do it in a separate PR.

test/test_models/webgpu/tiny-graph-capture-gqa/create_model.py
qjia7: Removed the webgpu test and related models in this PR. Will support this in a separate PR to enable the graph capture tests for webgpu. Thanks.

src/webgpu/interface.cpp
```diff
@@ -190,6 +195,50 @@ struct InterfaceImpl : DeviceInterface {
 
   void Synchronize() override {}  // Nothing to do?
 
+  bool UpdateAttentionMask(void* next_mask_data, void* mask_data, int batch_beam_size, int new_kv_length, int total_length, int max_length, bool update_only, ONNXTensorElementDataType type) override {
```
kunal-vaishnavi: Can we use `[[maybe_unused]]` instead to follow the [existing syntax structure](https://github.com/search?q=repo%3Amicrosoft%2Fonnxruntime-genai%20maybe_unused&type=code)?

src/models/position_inputs.cpp
```diff
@@ -380,23 +380,50 @@ void DefaultPositionInputs::InitializeSequenceLengths(std::array<int64_t, 2> sha
...
+    size_t max_len = static_cast<size_t>(state_.params_->search.max_length);
+    if (index > max_len) {
+      throw std::runtime_error("RewindMask: index exceeds max_length");
+    }
+    size_t batch_beam_size = static_cast<size_t>(attention_mask_shape_[0]);
+    auto byte_span = attention_mask_->GetByteSpan();
+    auto cpu_data = byte_span.CpuSpan();
+    if (type_ == Ort::TypeToTensorType<int32_t>) {
+      auto* data = reinterpret_cast<int32_t*>(cpu_data.data());
```
baijumeswani, line 394: Do you need a reinterpret cast here? Would static cast work?

src/models/position_inputs.cpp
```diff
@@ -380,23 +380,50 @@ void DefaultPositionInputs::InitializeSequenceLengths(std::array<int64_t, 2> sha
...
+      for (size_t i = 0; i < batch_beam_size; i++) {
+        std::fill_n(data + i * max_len, index, static_cast<int32_t>(1));
+        std::fill_n(data + i * max_len + index, max_len - index, static_cast<int32_t>(0));
+      }
+    } else {
+      auto* data = reinterpret_cast<int64_t*>(cpu_data.data());
```
baijumeswani, line 400: Same here.

src/webgpu/interface.cpp
```diff
@@ -190,6 +195,47 @@ struct InterfaceImpl : DeviceInterface {
 
   void Synchronize() override {}  // Nothing to do?
 
+  bool UpdateAttentionMask([[maybe_unused]] void* next_mask_data, void* mask_data, int batch_beam_size, [[maybe_unused]] int new_kv_length, int total_length, [[maybe_unused]] int max_length, bool update_only, ONNXTensorElementDataType type) override {
```
baijumeswani, line 198: Are these maybe unused or definitely unused?

baijumeswani (approved):

src/models/position_inputs.cpp
qjia7: [cpu_data.data()](vscode-file://vscode-app/c:/Users/jiajiaqin/AppData/Local/Programs/Microsoft%20VS%20Code/ce099c1ed2/resources/app/out/vs/code/electron-browser/workbench/workbench.html) returns uint8_t* (raw byte span). Converting uint8_t* to int32_t*/int64_t* requires [reinterpret_cast](vscode-file://vscode-app/c:/Users/jiajiaqin/AppData/Local/Programs/Microsoft%20VS%20Code/ce099c1ed2/resources/app/out/vs/code/electron-browser/workbench/workbench.html) — [static_cast](vscode-file://vscode-app/c:/Users/jiajiaqin/AppData/Local/Programs/Microsoft%20VS%20Code/ce099c1ed2/resources/app/out/vs/code/electron-browser/workbench/workbench.html) cannot convert between unrelated pointer types.

src/webgpu/interface.cpp
qjia7: They are definitely unused in the WebGPU implementation. However, since this is an override of the virtual DeviceInterface::UpdateAttentionMask, we can't remove the parameters from the signature. And they are used in cuda's implementation. So it's ok to use [[maybe_unused]] here.

baijumeswani: If they are definitely unused, you can remove the argument names and just keep type info. But let's not block this pr on this right now.

kunal-vaishnavi (approved):

PR #2084, opened 2026-04-14
TurboQuant for ORT WebGPU along with model_bechmark updates to measure GPU memory

sushraja-msft:
Add support for kv cache quantization of existing models. 

To enable this for webgpu ep provider, make the following change to genai_config.json.

           "session_options": {
                "provider_options": [
                    {
                        "webgpu": {
                            "kvCacheQuantizationBits": "4"
                        }
                    }
                ]
            },
            

ORT side changes https://github.com/microsoft/onnxruntime/pull/28059

This change also updates the model_benchmark tool to print GPU memory usage numbers and allow profiling.


src/models/kv_cache.cpp
```diff
@@ -150,6 +150,19 @@ void CombinedKeyValueCache::PickPastState(DeviceSpan<int32_t> beam_indices, int
 
 namespace {
 
+// Compute the compressed KV cache head dimension for TurboQuant.
+// TurboQuant packs each head into (1 + head_size/8) u32 words: one fp32 scale followed by
+// head_size values quantized to 4 bits. The tensor dimension depends on the element type.
+// Returns the original head_size unchanged when turbo_quant_bits is 0 (disabled).
+int64_t ComputeTurboQuantHeadSize(int head_size, int turbo_quant_bits, ONNXTensorElementDataType type) {
+  if (turbo_quant_bits == 4) {
+    const int compressed_u32_words = head_size / 8 + 1;
```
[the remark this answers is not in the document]
sushraja-msft: // TurboQuant requires head_size > 4 and power of 2 (Hadamard transform needs power-of-2,
  // and packing 8 indices per u32 needs head_size >= 8). Silently disable if not met.
  
@copilot Allocate the regular kv cache size instead of the above condition is not met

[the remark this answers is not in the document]
sushraja-msft: @copilot  when the head size is incompatible for turbo quant can we throw an error instead of silently ignoring turbo quant ?

src/config.h
```diff
@@ -467,6 +467,8 @@ void SetProviderOption(Config& config, std::string_view provider_name, std::stri
 void OverlayConfig(Config& config, std::string_view json);
 int SafeDoubleToInt(double x, std::string_view name);
 
+int GetKvCacheQuantizationBits(const Config::SessionOptions& session_options,
```
baijumeswani: Could we move this utility into a new config_utils.h/cpp (others as well can go there but if you want to limit the scope of this PR to only GetKvCacheQuantizationBits, that should be fine)?

src/models/kv_cache.cpp
```diff
@@ -279,6 +309,19 @@ DefaultKeyValueCache::DefaultKeyValueCache(State& state)
 
   // Derive the KV data type from the first KV input
   type_ = model_.session_info_.GetInputDataType(input_name_strings_[0]);
+
+  // Detect KV cache quantization configuration from provider options.
+  const int kv_cache_quantization_bits = GetKvCacheQuantizationBits(model_.config_->model.decoder.session_options);
```
baijumeswani: Could we explicitly pass in the provider type and avoid the default value of webgpuep?

benchmark/c/CMakeLists.txt
```diff
@@ -28,6 +28,10 @@ target_include_directories(model_benchmark PRIVATE
 
 target_link_libraries(model_benchmark PRIVATE onnxruntime-genai ${ONNXRUNTIME_LIB})
 
+if(WIN32)
+  target_link_libraries(model_benchmark PRIVATE dxgi.lib)
```
baijumeswani: Where does one get dxgi.lib from? Is that available by default with vs build generator?

src/python/py/models/builders/base.py
```diff
@@ -170,8 +170,8 @@ def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
         self.output_shapes = {
             "hidden_states": ["batch_size", "sequence_length", self.hidden_size],                                # For standard models where you want to remove the language modeling head from the model (note that `hidden_states` is written this way to match Hugging Face format)
             "logits": ["batch_size", "sequence_length", self.vocab_size],                                        # For standard models
-            "present.key": ["batch_size", self.num_kv_heads, "total_sequence_length", self.head_size],           # For standard models (note that `present.key` is written this way to match Hugging Face format)
```
kunal-vaishnavi, line 173: Let's align the spacing on the comments next to these output shapes.

src/python/py/models/builders/base.py
```diff
@@ -149,8 +149,8 @@ def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
             "attention_mask": ["batch_size", "total_sequence_length"],                                           # For standard models
             "position_ids": ["batch_size", "sequence_length"],                                                   # For standard models
             "inputs_embeds": ["batch_size", "sequence_length", self.hidden_size],                                # For standard models where you want to remove the embedding layer from the model (note that `inputs_embeds` is written this way to match Hugging Face format)
-            "past_key_values.key": ["batch_size", self.num_kv_heads, "past_sequence_length", self.head_size],    # For standard models (note that `past_key_values.key` is written this way to match Hugging Face format)
```
kunal-vaishnavi, line 152: Same comment here

src/python/py/models/builders/base.py
```diff
@@ -170,8 +170,8 @@ def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
         self.output_shapes = {
             "hidden_states": ["batch_size", "sequence_length", self.hidden_size],                                # For standard models where you want to remove the language modeling head from the model (note that `hidden_states` is written this way to match Hugging Face format)
             "logits": ["batch_size", "sequence_length", self.vocab_size],                                        # For standard models
-            "present.key": ["batch_size", self.num_kv_heads, "total_sequence_length", self.head_size],           # For standard models (note that `present.key` is written this way to match Hugging Face format)
-            "present.value": ["batch_size", self.num_kv_heads, "total_sequence_length", self.head_size],         # For standard models (note that `present.value` is written this way to match Hugging Face format)
+            "present.key": ["batch_size", self.num_kv_heads, "total_sequence_length", "kv_cache_dim"],          # For standard models (note that `present.key` is written this way to match Hugging Face format). Last dim is symbolic so a single export serves both non-quantized (head_size) and quantized (compressed) KV caches.
```
kunal-vaishnavi: 1. Why does this need to be symbolic even when quantized? Can its value not be pre-computed?
2. How does this change work for pre-allocating KV cache tensors? Without a constant in that position, I believe ORT GenAI will have trouble reading the head size from the model's input/output shapes unless it is now reading from the GenAI config.
3. Can we still pre-allocate with this change or is buffer sharing disabled now?

sushraja-msft: 1. Symbolic allows the same model to flip between quantized and unquantized kv cache with just a config change.
2/3. I have been testing with the symbolic value in the head size and see no issues so far. We do use shared past and present buffers with my change. Is there a breakpoint I can set to verify your concern about not taking a particular optimized path in gen ai ?

sushraja-msft: 2. Where does GenAI read the head size for pre-allocation — graph shapes or config?

From the GenAI config, and that's unchanged. The allocator sizes the cache from model_.config_->model.decoder.head_size, then compresses it:


It reads the graph's input shapes only to detect per-layer variation, and there it explicitly ignores any non-concrete dim (if (input_shape[kHeadDimAxis] > 0)), falling back to config. So a symbolic head dim in the graph doesn't break sizing — GenAI never relied on a graph constant for the compressed cache; the compressed value can only come from config + the packing formula anyway.

3. Can we still pre-allocate / is buffer sharing disabled?

Still works fully — nothing is disabled. Every allocation path sizes off shape_ (which already holds the compressed dim): empty_past_, the past_present_share_buffer_ branch (sets only shape_[2], the seq dim), and per-layer layer_shapes_. The symbolic graph dim actually enables sharing: because ORT no longer enforces a static 128, the single shared past/present buffer (allocated once at compressed size) binds cleanly every step. The WebGPU kernel still validates buffer size at runtime and fails with INVALID_ARGUMENT on any drift, so the contract is enforced — just at bind/runtime instead of via a graph constant.

benchmark/c/CMakeLists.txt
sushraja-msft: Part of windows sdk.

kunal-vaishnavi:
From Copilot:

This makes the KV-cache last dimension symbolic for all standard exports, but the allocator support in GenAI is only partially updated.

`DefaultKeyValueCache` now derives the concrete KV width from config and looks consistent with this change, but other KV allocators in `src/models/kv_cache.cpp` still hardcode `decoder.head_size` (for example `CombinedKeyValueCache` and `LFM2Cache`). If this symbolic `kv_cache_dim` can ever flow through those paths, they would still allocate the uncompressed width and diverge from the new contract.

Can we either:
1. confirm this export change is only reachable for models that use `DefaultKeyValueCache`, or
2. update the other KV allocators to use the same quantization-aware head-dimension logic for consistency?

src/python/py/models/builders/base.py
kunal-vaishnavi: Do you plan to support using the same ONNX model for various quantization modes from TurboQuant or using one ONNX model per mode? If it's the former, then I understand the dynamic dim. If it's the latter, then we should know the dim size ahead of time. Technically, we could make all dims symbolic. However, in practice we try to use static dims when they are known for better compatibility across all IHV EPs.

sushraja-msft: it is the former. Same model, enable / disable turboquant - hopefully at run time.

baijumeswani (approved):

sushraja-msft:
> From Copilot:
> 
> This makes the KV-cache last dimension symbolic for all standard exports, but the allocator support in GenAI is only partially updated.
> 
> `DefaultKeyValueCache` now derives the concrete KV width from config and looks consistent with this change, but other KV allocators in `src/models/kv_cache.cpp` still hardcode `decoder.head_size` (for example `CombinedKeyValueCache` and `LFM2Cache`). If this symbolic `kv_cache_dim` can ever flow through those paths, they would still allocate the uncompressed width and diverge from the new contract.
> 
> Can we either:
> 
> 1. confirm this export change is only reachable for models that use `DefaultKeyValueCache`, or
> 2. update the other KV allocators to use the same quantization-aware head-dimension logic for consistency?

For this at the moment GPT and LFM2 models using the other allocator path would indeed allocate a larger KV cache than necessary. Ill fix in a follow up change.

PR #2085, opened 2026-04-14
Reduce CPU-side per-token overhead in GenerateNextToken and SampleTopP

hanbitmyths:
This PR is to reduce CPU-side per-token costs in two areas:

**1. Cache per-token decisions at construction** (generators.cpp, generators.h)
- Replace 3 per-token `dynamic_cast<NemotronSpeechState*>` with a one-time check at construction (`is_nemotron_speech_model_`)
- Pre-compute Phi3 ROPE rewind threshold and EP single-rope-factor flag, eliminating per-token string comparisons and device type checks
- Pre-determine sampling method (`SamplingMethod` enum + `switch`), replacing per-token conditional chain and moving parameter validation to construction

**2. Eliminate O(V) exp() pass in SampleTopP** (search.cpp)
- Use a tail-bound in `FindNucleus`: after partial-sorting K elements, bound the unsorted tail as `(V-i-1) * exp(score_i)` and terminate early when `prefix_sum >= p * (prefix_sum + tail_bound)`
- Start with K₀=16 (adopted from OpenVINO GenAI) with 4× geometric growth — minimizes wasted sort work when the top few tokens dominate, while avoiding O(V log V) full-sort fallback for flat distributions
- Exact-cutoff refinement for small vocabularies (V ≤ K₀ × growth_factor) where the tail bound is loose
- Saves ~V calls to `std::exp()` per token (128K-200K for production models)

**Benchmarks** (Batch=1, CPU, Release build)

SampleTopP (peaked distribution, p=0.95):
| Vocab | Before (μs) | After (μs) | Speedup |
|---|---|---|---|
| 32K | 321 | 62 | **81%** |
| 128K | 1119 | 272 | **76%** |
| 200K | 1637 | 544 | **67%** |

Consistent 65-81% reduction across distribution shapes and top_p values (0.80-0.99).

Orchestration overhead (V=1000, greedy — isolates dispatch path):
| | Before (μs) | After (μs) | Speedup |
|---|---|---|---|
| Greedy | 0.41 | 0.30 | **27%** |

Adds two `DISABLED_` benchmark tests for reproducibility.

### DIFF
--- a/src/generators.h
+++ b/src/generators.h
@@ -131,9 +131,8 @@ struct Generator : LeakChecked<Generator> {
                 rewound };  // Set after RewindToLength
   Action last_action_{standard};
 
-  // Pre-computed per-token decisions (P1 optimization: avoid repeated checks each token)
+  // Pre-computed per-token decisions: avoid repeated checks each token
   bool is_nemotron_speech_model_{};
-  bool ep_uses_single_rope_factor_{};
   int phi3_rope_threshold_{};  // 0 means no ROPE rewind needed
   enum class SamplingMethod { kGreedy, kTopK, kTopP, kTopKTopP };
   SamplingMethod sampling_method_{SamplingMethod::kGreedy};
--- a/src/generators.h
+++ b/src/generators.h
@@ -139,7 +139,8 @@ struct Generator : LeakChecked<Generator> {
                               kTopP,
                               kTopKTopP };
   SamplingMethod sampling_method_{SamplingMethod::kGreedy};
-  void InitPerTokenCache(const GeneratorParams& params);
+  void InitializeSamplingMethod(const GeneratorParams& params);
+  void InitializePhi3RopeThreshold(const GeneratorParams& params);
 };
 
 struct OrtGlobals {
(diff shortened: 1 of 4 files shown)


src/generators.h
```diff
@@ -130,6 +130,16 @@ struct Generator : LeakChecked<Generator> {
...
+
+  // Pre-computed per-token decisions: avoid repeated checks each token
+  bool is_nemotron_speech_model_{};
+  int phi3_rope_threshold_{};  // 0 means no ROPE rewind needed
+  enum class SamplingMethod { kGreedy,
+                              kTopK,
+                              kTopP,
+                              kTopKTopP };
+  SamplingMethod sampling_method_{SamplingMethod::kGreedy};
+  void InitPerTokenCache(const GeneratorParams& params);
```
baijumeswani: Can we break this function down into more meaningful functions like InitializeSamplingMethond() and InitializePhi3RopeThreshold()?

baijumeswani (dismissed):

baijumeswani (approved):

Issue #2062, opened 2026-04-03
Feature Request: Support for Google Gemma 4 model family (PLE architecture, variable head dims, KV cache sharing)


elbruno:
### Summary

Google released the [Gemma 4 model family](https://blog.google/innovation-and-ai/technology/developers-tools/gemma-4/) (April 2, 2026) with four variants (E2B, E4B, 26B-A4B MoE, 31B). All are Apache 2.0 licensed and use a novel architecture that introduces three features not currently supported by `onnxruntime-genai` v0.12.2.

I'm the author of [ElBruno.LocalLLMs](https://github.com/elbruno/ElBruno.LocalLLMs), a C# library that uses ONNX Runtime GenAI for local LLM inference via `IChatClient`. I've spent significant time attempting to convert Gemma 4 models and have detailed technical findings on what's blocking support.

### What I Tried

1. **Patched `builder.py`** to route `Gemma4ForConditionalGeneration` through the Gemma 3 pipeline — conversion produced a 1.6GB ONNX file (INT4), but the runtime failed at load time:
   ```
   ShapeInferenceError: Incompatible dimensions for matrix multiplication
   at /model/layers.4/attn/o_proj/MatMul_Q4
   ```
   Layer 4 is a full-attention layer with `global_head_dim=512`, but the Gemma 3 model builder created the graph assuming uniform `head_dim=256`.

2. **Changed `genai_config.json` model type** from `gemma4` to `gemma3_text` — same shape error.

3. **Examined `onnx-community/gemma-4-E2B-it-ONNX`** — correct ONNX graph structure, but uses a different I/O contract (separate `embed_tokens.onnx` + `decoder_model_merged.onnx` with `per_layer_inputs` tensor) that's incompatible with GenAI's KV cache management.

4. **Attempted `Gemma4ForCausalLM` loading** — weights stored under `model.language_model.*` prefix (multimodal wrapper), causing weight mismatch.

#### 1. Per-Layer Embeddings (PLE)

Gemma 4's `embed_tokens` produces TWO outputs:
- `inputs_embeds`: `[batch, seq, hidden_size]` (standard)
- `per_layer_inputs`: `[batch, seq, num_hidden_layers, hidden_size_per_layer_input]` — e.g., `[batch, seq, 35, 256]` for E2B

Each transformer layer receives its own slice from `per_layer_inputs`. The GenAI runtime currently expects a single embedding tensor flowing into the decoder stack.

**Config reference (E2B):**
[...]

PR #2078, opened 2026-04-10
extend modelbuilder to build Olmo3, SmolLM3 and other models

xadupre:
- it adds unit test to check discrepancies on all supported architectures
- it fixes a few things in existing architectures detected through the unit tests
- it supports transformers>=5 by modifying the final genai config to make it lookl like a transformers<5
- it adds a job to run the created models on cpu with the released version of onnxruntime-genai
- the tests runs in 3 or 4 minutes on two versions of transformers (4.57, 5.5)
- it uploads artifact showing the test results (below an example)

|     | model_id                               | experiment        | precision   | provider   | input_type   |   max_abs_err |   avg_abs_discrepancy |   dnan | next_token   |   next_token_id_tch |   next_token_id_ort | test                                            |   first_diff |   total_diff | kind   | step    |       %>0.1 |      %>0.01 |
|----:|:---------------------------------------|:------------------|:------------|:-----------|:-------------|--------------:|----------------------:|-------:|:-------------|--------------------:|--------------------:|:------------------------------------------------|-------------:|-------------:|:-------|:--------|------------:|------------:|
|   0 | arnir0/Tiny-LLM                        | attention_decode  | fp32        | cpu        | text         |   1.3411e-07  |           3.00374e-08 |      0 | OK           |                 324 |                 324 | test_llama_attention_decode_discrepancies       |              |              |        |         | 0           | 0           |
|   1 | arnir0/Tiny-LLM                        | attention_prefill | fp32        | cpu        | text         |   3.27826e-07 |           4.81524e-08 |      0 | OK           |                 155 |                 155 | test_llama_attention_prefill_discrepancies      |              |              |        |         | 0           | 0           |
|   2 | THUDM/chatglm3-6b                      | generate          | fp16        | cpu        | text         |               |                       |        |              |                     |                     | test_generation_chatglm_fp16_cpu                |              |            0 | fast   |         |             |             |
|   3 | THUDM/chatglm3-6b                      | generate          | fp32        | cpu        | text         |               |                       |        |              |                     |                     | test_generation_chatglm_fp32_cpu                |              |            0 | fast   |         |             |             |
|   4 | THUDM/chatglm3-6b                      | forward           | fp16        | cpu        | text         |   0.000906795 |           0.000173806 |      0 | OK           |                 905 |                 905 | test_discrepancies_chatglm_fp16_cpu             |              |              | fast   | prefill | 0           | 0           |
[...]

PR #2088, opened 2026-04-16
Gemma 4 builder scaffolding (Part 1 of N, #2062)

jaburges:
## Description

Scaffolds Gemma 4 support in the Python model builder, addressing part of #2062. Intentionally opened as a **Draft** because this is **Part 1 of N** — the builder-side work only. See "Scope" below for exactly what is and isn't covered, and why a functional end-to-end path requires paired C++ runtime changes.

Credit to @elbruno for the original architectural write-up in #2062; this PR follows that decomposition.

## Motivation

Google released the [Gemma 4 family](https://blog.google/technology/developers/gemma-4/) (E2B / E4B / 26B-A4B / 31B) in April 2026. All variants use a new architecture with three features that don't fit GenAI's current assumptions:

1. **Per-Layer Embeddings (PLE)** — `embed_tokens` emits two tensors; each decoder layer consumes its own slice.
2. **Variable attention head dimension** — sliding layers use `head_dim=256`, full-attention layers use `global_head_dim=512`.
3. **KV cache sharing** — E2B has 35 decoder layers but only 15 own unique KV caches (`num_kv_shared_layers=20`).

Existing workarounds (patching `builder.py` to route through Gemma 3, or using the `onnx-community/gemma-4-E2B-it-ONNX` export) fail with shape errors or I/O-contract mismatches. See #2062 for the full analysis.

### ✅ Covered by this PR

- `builders/gemma4.py` (new): `Gemma4TextModel(Gemma3Model)` overriding:
  - `is_local(layer_id)` driven by `config.layer_types` (Gemma 3 used a hard-coded every-6th-layer rule; Gemma 4 exposes the full list).
  - `make_key_value_cache_shape(layer_id, shape)` using the existing per-layer extension point to emit `head_dim=256` on sliding layers and `global_head_dim=512` on full-attention layers.
  - `make_rotary_embedding_multi_cache()` building the two RoPE caches Gemma 4 needs (sliding: `theta=10_000`, `partial_rotary_factor=1.0`; full: `theta=1_000_000`, `partial_rotary_factor=0.25`, `rope_type=proportional`) from `config.rope_parameters`.
- `builder.py`: dispatch for `Gemma4ForConditionalGeneration` and `Gemma4ForCausalLM`, with `text_config` flattening and `exclude_embeds=true` defaulting, matching the existing Gemma 3 pattern.
- `builders/__init__.py`: exports `Gemma4TextModel`.
- `test/python/test_gemma4_builder.py` (new): 10 unit tests for the Gemma 4-specific overrides. Runs in a clean Python env — no torch / transformers / onnxruntime needed — by stubbing the `Gemma3Model` parent.

### ❌ Not covered — requires C++ runtime changes (Parts 2/3/4)

These are explicitly out of scope. The builder raises `NotImplementedError` listing exactly which ones were tripped, so users get an actionable error rather than a broken graph:

1. **Per-Layer Embeddings routing.** Needs a new `per_layer_inputs` side-channel input both in the ONNX graph (new `make_inputs_and_outputs` branch) and in the C++ model loader. Touches `src/models/model.cc` and friends.
2. **KV cache sharing.** `src/models/kv_cache.cpp` currently allocates one past/present pair per layer. Gemma 4 needs layer aliasing (20 of 35 layers reuse an earlier layer's KV).
3. **Per-layer `head_size` in the runtime.** Even with this builder emitting correct per-layer past_kv shapes, `kv_cache.cpp:17` reads a single `model_.config_->model.decoder.head_size` and applies it uniformly. The `genai_config.json` schema and the KV cache manager need per-layer awareness, or a `head_size` + `global_head_size` split driven by the same `layer_types` pattern the builder already understands.

Until those land, this builder supports `--extra_options config_only=true` only. That's still useful for validating the dispatcher and for downstream tooling that only needs `genai_config.json` / processing files. The gate makes that explicit rather than silently emitting a broken graph.

## Verification

```
cd test/python && python test_gemma4_builder.py -v
```

```
test_gate_quiet_in_config_only_mode ... ok
test_gate_raises_on_kv_sharing ... ok
test_gate_raises_on_ple ... ok
test_gate_raises_on_variable_head_dim ... ok
test_is_local_falls_back_when_layer_types_missing ... ok
test_is_local_uses_layer_types ... ok
test_kv_cache_shape_preserves_non_head_dim_axes ... ok
test_kv_cache_shape_uses_full_head_dim_on_full_layers ... ok
test_kv_cache_shape_uses_sliding_head_dim_on_sliding_layers ... ok
test_rope_caches_are_emitted_for_both_attention_types ... ok

Ran 10 tests in 0.009s
OK
```

The tests pin down:
- `is_local` matches the expected E2B pattern — full layers at exactly 4, 9, 14, 19, 24, 29, 34.
- `make_key_value_cache_shape` emits 512 on those seven layers and 256 everywhere else.
- The unsupported-feature gate raises individually for each of the three runtime blockers.
- `config_only=true` mode bypasses the gate cleanly.
- Both RoPE caches are emitted with the correct theta and partial-rotary factor.

## Request for maintainer review

Three specific things I'd welcome feedback on before taking Parts 2/3 further:

1. **PR decomposition.** Happy to split Parts 2/3/4 into separate PRs (runtime PLE, runtime KV sharing, runtime per-layer head_size) or keep them together — whichever matches your preference.
2. **Config schema.** For per-layer `head_size`, do you prefer (a) a `head_size` list in `genai_config.json`, (b) a `head_size` + `global_head_size` pair plus a `layer_types`-like pattern, or (c) keeping both values in the model metadata only and teaching `kv_cache.cpp` to read them from the ONNX graph? I have mild preference for (b) for symmetry with how `is_local` already works.
3. **RoPE caches at different head dims.** Currently the builder emits both RoPE caches sized to the sliding `head_dim`. Once per-layer head_size support lands in the runtime, the builder should emit the full-attention RoPE cache at `global_head_dim` — open to suggestions on the cleanest way to thread that through `make_rotary_embedding_caches` without duplicating logic.

## References

- Issue: #2062 (with @elbruno's analysis)
- Config: [`google/gemma-4-E2B-it/config.json`](https://huggingface.co/google/gemma-4-E2B-it/blob/main/config.json)
- Community ONNX export (architectural reference only; uses a GenAI-incompatible I/O contract): [`onnx-community/gemma-4-E2B-it-ONNX`](https://huggingface.co/onnx-community/gemma-4-E2B-it-ONNX)

### DIFF
--- a/src/python/py/models/builder.py
+++ b/src/python/py/models/builder.py
@@ -21,6 +21,7 @@ from builders import (
     ErnieModel,
     Gemma2Model,
     Gemma3Model,
+    Gemma4TextModel,
     GemmaModel,
     GPTOSSModel,
     GraniteModel,
@@ -233,6 +234,27 @@ def create_model(
         print("WARNING: This is only generating the text component of the model. Setting `--extra_options exclude_embeds=true` by default.")
         extra_options["exclude_embeds"] = True
         onnx_model = Gemma3Model(config, io_dtype, onnx_dtype, execution_provider, cache_dir, extra_options)
+    elif config.architectures[0] in {"Gemma4ForConditionalGeneration", "Gemma4ForCausalLM"}:
+        # Gemma 4 ships multimodal-first. The text decoder config lives under
+        # `text_config` on the conditional-generation wrapper; weights are
+        # stored under the `model.language_model.*` prefix. We flatten the
+        # text_config onto the top-level config the way the Gemma 3 branch
+        # does, then hand off to the scaffolded Gemma 4 builder.
+        # NOTE (issue #2062): Gemma 4 support is incomplete. This branch
+        # currently only supports `--extra_options config_only=true`. See
+        # `builders/gemma4.py` for the full scope statement and the list
+        # of runtime changes required for end-to-end support.
+        if hasattr(config, "text_config"):
+            text_config = config.text_config
+            for key in text_config:
+                if not hasattr(config, key):
+                    setattr(config, key, getattr(text_config, key))
+        print("WARNING: Gemma 4 support is scaffolded but not yet end-to-end functional. See issue #2062.")
+        print("WARNING: This model loses accuracy with float16 precision. It is recommended to set `--precision bf16` or `--precision int4 --extra_options use_cuda_bf16=true` by default.")
+        print("WARNING: This is only generating the text component of the model. Setting `--extra_options exclude_embeds=true` by default.")
+        extra_options["exclude_embeds"] = True
+        onnx_model = Gemma4TextModel(config, io_dtype, onnx_dtype, execution_provider, cache_dir, extra_options)
+        onnx_model.model_type = "gemma4_text"
     elif config.architectures[0] == "GptOssForCausalLM":
         print("WARNING: This model only supports symmetric quantization for `QMoE`.")
         if hasattr(config, "quantization_config") and config.quantization_config.get("quant_method") != "quark":
--- a/src/python/py/models/builders/__init__.py
+++ b/src/python/py/models/builders/__init__.py
@@ -10,6 +10,7 @@ from .base import Model
 from .chatglm import ChatGLMModel
 from .ernie import ErnieModel
 from .gemma import Gemma2Model, Gemma3Model, GemmaModel
+from .gemma4 import Gemma4TextModel
 from .gptoss import GPTOSSModel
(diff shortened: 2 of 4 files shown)


jaburges:

xadupre:
#2078 is adding many unittest on modelbuilder, that would be great if you add for your changes once this PR is merged.

PR #2093, opened 2026-04-20
Add AMDGPU execution provider support

aditya-dl:
Add AMD GPU (MIGraphX) execution provider support to ONNX Runtime GenAI. The provider is exposed as "amdgpu" to users and maps to the MIGraphX EP in ONNX Runtime internally.

Changes:
- Create src/amdgpu/session_options.{h,cpp} with AppendExecutionProvider that tries V2 plugin path then falls back to V1 legacy API
- Add provider name normalization ("amdgpu" -> "AMDGPU") and register in the dispatch table
- Enable graph capture for AMDGPU to allow compiled graph reuse during token generation
- Add static input shape padding (prompt_gen_ flag) so the EP avoids recompilation on varying prompt lengths. Gated behind NeedsStaticInputShapes() to only activate for AMDGPU, not other EPs
- Fix input_ids padding to handle both int32 and int64 element types, and correct per-row copy for batch_size > 1
- Fix position_ids padding to prevent out-of-bounds read on next_tokens when tensor shape is padded to max_length for batch_size > 1

Configuration:
  "provider_options": [{ "amdgpu": {} }] or: config.append_provider("amdgpu")

Known limitations:
- Beam search not supported (requires past_present_share_buffer=true which requires num_beams=1)
- Inputs allocated on CPU; the MIGraphX EP handles CPU<->GPU transfers internally

### DIFF
--- a/src/amdgpu/session_options.cpp
+++ b/src/amdgpu/session_options.cpp
@@ -1,0 +1,34 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+//
+// Modifications Copyright(C) 2026 Advanced Micro Devices, Inc. All rights reserved.
+
+#include "session_options.h"
+
+#include "../models/session_options.h"
+
+namespace Generators::AMDGPUExecutionProvider {
+
+DeviceInterface* AppendExecutionProvider(OrtSessionOptions& session_options,
+                                         const Config::ProviderOptions& provider_options,
+                                         const Config& config,
+                                         bool /*disable_graph_capture*/) {
+  // AMDGPU does not have a device type specific allocator in OGA yet,
+  // so we use CPU as the device type. The AMDGPU EP handles CPU<->GPU
+  // transfers internally.
+  // V2 path: plugin EP registers as "AMDGPUExecutionProvider"
+  if (!AppendExecutionProviderV2(session_options, provider_options,
+                                 DeviceType::CPU, "AMDGPUExecutionProvider")) {
+    // V1 fallback: legacy ORT recognizes "MIGraphX"
+    std::vector<const char*> keys, values;
+    for (auto& option : provider_options.options) {
+      keys.emplace_back(option.first.c_str());
+      values.emplace_back(option.second.c_str());
+    }
+    session_options.AppendExecutionProvider("MIGraphX", keys.data(), values.data(), keys.size());
+  }
+
+  return nullptr;
+}
+
+}  // namespace Generators::AMDGPUExecutionProvider
--- a/src/models/logits.cpp
+++ b/src/models/logits.cpp
@@ -107,6 +107,9 @@ void Logits::Update(const DeviceSpan<int32_t>& next_tokens, size_t new_kv_length
   }
 
   shape_[1] = new_kv_length;
+  if (state_.prompt_gen_) {
+    shape_[1] = state_.params_->search.max_length;
+  }
   output_raw_->CreateTensor(shape_, state_.params_->use_graph_capture && shape_[1] == 1);
   state_.outputs_[output_index_] = output_raw_->GetOrtTensor();
 }
--- a/src/models/model.cpp
+++ b/src/models/model.cpp
@@ -154,8 +154,51 @@ void State::Run(OrtSession& session, bool graph_capture_this_run) {
     run_options_->AddConfigEntry("disable_synchronize_execution_providers", "1");
   }
 
+  std::unique_ptr<OrtValue> new_input_ids;
+
+  if (prompt_gen_) {
+    for (int i = 0; i < inputs_.size(); i++) {
+      std::string input_name = input_names_[i];
+
+      if (input_name == "input_ids") {
+        int64_t batch_size = params_->search.batch_size;
+        int64_t padded_seq_len = static_cast<int64_t>(params_->search.max_length);
+        std::vector<int64_t> new_shape{batch_size, padded_seq_len};
+
+        OrtValue* value = inputs_[i];
+        auto info = value->GetTensorTypeAndShapeInfo();
+        ONNXTensorElementDataType elem_type = info->GetElementType();
+        std::vector<int64_t> dims = info->GetShape();
+        int64_t orig_seq_len = std::min(dims[1], padded_seq_len);
+
+        new_input_ids = OrtValue::CreateTensor(model_.allocator_cpu_, new_shape, elem_type);
+
+        if (elem_type == ONNX_TENSOR_ELEMENT_DATA_TYPE_INT64) {
+          auto* src = value->GetTensorMutableData<int64_t>();
+          auto* dst = new_input_ids->GetTensorMutableData<int64_t>();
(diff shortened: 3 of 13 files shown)


src/amdgpu/session_options.cpp
```diff
@@ -0,0 +1,34 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+//
+// Modifications Copyright(C) 2026 Advanced Micro Devices, Inc. All rights reserved.
+
+#include "session_options.h"
+
+#include "../models/session_options.h"
+
+namespace Generators::AMDGPUExecutionProvider {
```
kunal-vaishnavi, line 10: Why name the provider as `AMDGPU` instead of `MIGraphX`?

src/models/model.cpp
```diff
@@ -154,9 +154,52 @@ void State::Run(OrtSession& session, bool graph_capture_this_run) {
     run_options_->AddConfigEntry("disable_synchronize_execution_providers", "1");
   }
 
+  std::unique_ptr<OrtValue> new_input_ids;
+
+  if (prompt_gen_) {
+    for (int i = 0; i < inputs_.size(); i++) {
+      std::string input_name = input_names_[i];
+
+      if (input_name == "input_ids") {
```
kunal-vaishnavi, line 163: A boolean called `is_prompt_` already exists for detecting the prefill stage when updating the input ids. 

https://github.com/microsoft/onnxruntime-genai/blob/f1855375637a41ff159f41fe31523d760e7e0ec6/src/models/input_ids.cpp#L84-L94

Can we set the correct input ids when constructed rather than doing it here right before the model runs?

src/models/logits.cpp
```diff
@@ -107,6 +107,10 @@ void Logits::Update(const DeviceSpan<int32_t>& next_tokens, size_t new_kv_length
   }
 
   shape_[1] = new_kv_length;
+  if (state_.prompt_gen_)
+  {
+    shape_[1] = state_.params_->search.max_length;
```
kunal-vaishnavi, line 111: Typically, we have used enabling or disabling graph capture to decide on static vs dynamic shapes. Can we use that rather than introducing new boolean feature flags such as `prompt_gen_`?

aditya-dl: We considered using `use_graph_capture` but it's also `true` for DML and WebGPU. Gating the static input padding behind `use_graph_capture` would cause DML/WebGPU to also pad inputs to `max_length` during prompt processing, which they don't need. They handle static shapes through their own mechanisms (DML has ep.dml.enable_graph_capture session config, WebGPU checks enableGraphCapture provider option). The `prompt_gen_` flag, gated behind `NeedsStaticInputShapes()`, ensures the padding only activates for AMDGPU without affecting other EPs. 
Let me know if my understanding is right.

src/models/model.cpp
aditya-dl: The current approach pads in `State::Run()` because the padding needs to be coordinated across input_ids, position_ids, and logits - all three must be padded to the same `max_length` for the shapes to be consistent. Moving input_ids padding to `DefaultInputIDs::Update()` would require scattering the padding logic across three separate classes (input_ids, position_inputs, logits) rather than keeping it centralized. 
Also, `is_prompt_` is currently used for beam search token duplicating - overloading it for static shape padding could be confusing. 
That said, if you feel the architectural separation is important, we can refactor to move each padding into its respective Update() function. Let me know your thoughts.

src/amdgpu/session_options.cpp
aditya-dl: Moving forward, we're renaming the EP to `AMDGPU Execution Provider`. Internally, OGA maps "AMDGPU" to "MIGraphX" when communicating with ORT, so there are no ORT-side changes needed. Users configure `"amdgpu": {}` in their `genai_config.json` or call `config.append_provider("amdgpu")`.

baijumeswani: Why not rename it inside ORT as well? Having multiple names is a problem because there are multiple components that are execution provider aware (including foundry local and foundry local catalog). It would be problematic if on the surface it appeared that the name is AMDExecutionProvider but when we query OrtGetEpDevices, it returns MIGraphXExecutionProvider?
If the name needs to change, it should change in ort as well.

src/amdgpu/session_options.cpp
```diff
@@ -0,0 +1,34 @@
...
+  // Note: ORT recognizes "MIGraphXExecutionProvider" — do not change these strings.
+  if (!AppendExecutionProviderV2(session_options, provider_options,
+                                 DeviceType::CPU, "MIGraphXExecutionProvider")) {
+    // V1 fallback: ORT recognizes "MIGraphX", not "AMDGPU"
+    std::vector<const char*> keys, values;
+    for (auto& option : provider_options.options) {
+      keys.emplace_back(option.first.c_str());
+      values.emplace_back(option.second.c_str());
+    }
+    session_options.AppendExecutionProvider("MIGraphX", keys.data(), values.data(), keys.size());
```
baijumeswani, line 28: How come the execution provider has a different name when using append execution provider v1 vs v2?

src/ort_genai_c.cpp
```diff
@@ -1088,11 +1088,20 @@ void OGA_API_CALL OgaDestroyEngine(OgaEngine* p) { p->ExternalRelease(); }
 void OGA_API_CALL OgaDestroyRequest(OgaRequest* p) { p->ExternalRelease(); }
 
 void OGA_API_CALL OgaRegisterExecutionProviderLibrary(const char* registration_name, const char* library_path) {
-  Ort::RegisterExecutionProviderLibrary(&(Generators::GetOrtEnv()), registration_name, fs::path(library_path).c_str());
+  // Map AMDGPU to MIGraphX for ORT compatibility
+  const char* ort_name = registration_name;
+  if (std::string_view(registration_name) == "AMDGPUExecutionProvider") {
+    ort_name = "MIGraphXExecutionProvider";
+  }
+  Ort::RegisterExecutionProviderLibrary(&(Generators::GetOrtEnv()), ort_name, fs::path(library_path).c_str());
```
baijumeswani: As I mentioned above, this has consequences on other layers in the stack (such as foundry local and foundry local catalog). Please unify the naming.

aditya-dl:
@baijumeswani Thank you for the feedback on naming consistency. We understand the concern.
 
A larger AMDGPU execution provider effort is underway. We are open-sourcing the plugin EP that will use the AMDGPU name consistently across the stack. This PR is the initial OGA-side integration to support that direction.
 
The MIGraphX name will be retained in legacy ORT for backward compatibility. Once the plugin EP is available, the naming will be unified end-to-end.

aditya-dl:
Superseded by #2165. Per AMD direction we have reverted the EP naming from AMDGPU back to MIGraphX in OGA. The new PR carries the same functional changes (EP registration, graph capture enable, static input shape padding, batch_size>1 fixes for input_ids and position_ids) plus a  catalog-name alias for Windows ML EP discovery compatibility. Please direct further review there.

PR #2094, opened 2026-04-22
[Qwen3.5] Make linear attention INT8 quantization opt-in via extra_options

daijh:
Change Qwen3.5 mixed-precision quantization for linear attention layers from always-on to opt-in. Users can enable INT8 promotion of linear attention nodes by passing linear_attention_int8=true in extra_options.

### DIFF
--- a/src/python/py/models/builders/qwen.py
+++ b/src/python/py/models/builders/qwen.py
@@ -1013,29 +1013,28 @@ class Qwen35TextModel(Model):
         # Disable fused RoPE in attention op - we apply mRoPE manually
         self.attention_attrs["use_rope_in_attn"] = False
 
-        # Mixed-precision quantization for linear attention layers.
-        # Baseline: whole model INT4. Override linear attention layer nodes
-        # to INT8 for better accuracy with modest size increase.
+        # Optional mixed-precision quantization for linear attention layers.
+        # When enabled via extra_options linear_attention_int8=true, promotes
+        # linear attention layer nodes from INT4 to INT8 for better accuracy.
         #
         # Linear attention recurrence accumulates errors across the full sequence,
         # unlike softmax attention which normalizes per-step.
-        int8_nodes = {}
-        for i, lt in enumerate(self.layer_types):
-            if lt == "linear_attention":
-                # All linear attention projections: INT8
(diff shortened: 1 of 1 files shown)


daijh:
Fix #2095

daijh:
@guschmue PTAL

src/python/py/models/builders/qwen.py
```diff
@@ -1013,29 +1013,37 @@ def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
...
-                    int8_nodes[f"/model/layers.{i}/mlp/{proj}/MatMul"] = {"bits": 8}
-
-        if int8_nodes:
-            algo_config = self.quant_attrs["int4"].get("algo_config")
-            if algo_config is not None and hasattr(algo_config, "customized_weight_config"):
-                algo_config.customized_weight_config.update(int8_nodes)
-            else:
-                algo_config = RTNWeightOnlyQuantConfig(customized_weight_config=int8_nodes)
-                self.quant_attrs["int4"]["algo_config"] = algo_config
+        linear_attention_int8 = extra_options.get("linear_attention_int8", False)
```
apsonawane, line 1022: Add ```linear_attention_int8``` to the bools list in ```check_extra_options``` in builder.py, then just simplify this to 
```if extra_options.get("linear_attention_int8", False)```

kunal-vaishnavi:
Since this is just affecting the MatMuls, can we instead introduce a new int4 algo config or use an existing one? In the future, there could be another op instead of linear attention and we wouldn't want to change the extra options for that new op then. The algo config change would be a more generic solution.

https://github.com/microsoft/onnxruntime-genai/blob/2a2ef8cda8aea7b88658b6cc871d9857e7b568eb/src/python/py/models/builder.py#L403-L410

daijh:
> Since this is just affecting the MatMuls, can we instead introduce a new int4 algo config or use an existing one? In the future, there could be another op instead of linear attention and we wouldn't want to change the extra options for that new op then. The algo config change would be a more generic solution.
> 
> https://github.com/microsoft/onnxruntime-genai/blob/2a2ef8cda8aea7b88658b6cc871d9857e7b568eb/src/python/py/models/builder.py#L403-L410

Thank you.
Using a new INT4 algo config is definitely the right move here.

I’m going to close this PR for now, as PR #2100 has been created as a draft for this purpose.

PR #2099, opened 2026-04-24
Enable graph capture for WebGPU models and DML continuous decoding tests

qjia7:
- Add enable_graph_capture flag to model download config in _test_utils.py
- When enable_graph_capture=True and device=webgpu, pass enable_webgpu_graph=true to the model builder so generated models have enableGraphCapture=1
- Enable graph capture for qwen-2.5-0.5b model (used by guidance tests)
- Remove #if !USE_DML guard from multi-turn guidance tests since PR #2083 fixed RewindTo for DML (DML always uses graph capture/static mask)


test/python/_test_utils.py
```diff
@@ -94,7 +96,7 @@ def get_model_paths():
     return ci_paths, hf_paths
 
 
-def download_model(model_name, input_path, output_path, precision, device, one_layer):
+def download_model(model_name, input_path, output_path, precision, device, one_layer, enable_graph_capture=False):
```
kunal-vaishnavi: Can we remove the preset default value in the parameter here? The caller should know what value to set it to.

```suggestion
def download_model(model_name, input_path, output_path, precision, device, one_layer, enable_graph_capture):
```

test/python/_test_utils.py
```diff
@@ -126,6 +128,8 @@ def download_model(model_name, input_path, output_path, precision, device, one_l
         extra_options += ["int4_accuracy_level=4"]
     if one_layer:
         extra_options += ["num_hidden_layers=1"]
+    if enable_graph_capture and device == "webgpu":
```
kunal-vaishnavi, line 131: There is [another PR](https://github.com/microsoft/onnxruntime-genai/pull/2070) to re-enable graph capture for the CUDA EP as well. Can we move the comments about graph capture being specific to WebGPU to here? The `enable_graph_capture` flag will generically add support for using graph capture, and we can later on enable graph capture for CUDA.

test/python/_test_utils.py
```diff
@@ -141,19 +145,20 @@ def download_models(download_path, precision, device, log):
...
     # python -m onnxruntime_genai.models.builder -i <input_path> -o <output_path> -p <precision> -e <device>
-    for model_name, (input_path, one_layer) in ci_paths.items():
+    for model_name, (input_path, one_layer, graph_capture) in ci_paths.items():
         try:
             output_path = os.path.join(download_path, model_name, precision, device)
             log.debug(f"Downloading {model_name} from {input_path} to {output_path}")
             if not os.path.exists(output_path):
-                download_model(None, input_path, output_path, precision, device, one_layer)
+                download_model(None, input_path, output_path, precision, device, one_layer,
+                               enable_graph_capture=graph_capture)
```
kunal-vaishnavi:
```suggestion
                               graph_capture)
```

test/python/_test_utils.py
```diff
@@ -169,7 +174,8 @@ def download_models(download_path, precision, device, log):
         log.debug(f"Downloading {model_name} from {hf_name} to {output_path}")
 
         if not os.path.exists(output_path):
-            download_model(hf_name, "", output_path, precision, device, one_layer)
+            download_model(hf_name, "", output_path, precision, device, one_layer,
+                           enable_graph_capture=graph_capture)
```
kunal-vaishnavi:
```suggestion
                           graph_capture)
```

test/c_api_tests.cpp
```diff
@@ -1384,6 +1384,62 @@ TEST(CAPITests, RewindGraphCaptureNvTensorRtRtxCAPI) {
   EXPECT_TRUE(0 == std::memcmp(first_output.data(), generator->GetSequenceData(0), seq_len * sizeof(int32_t)));
 }
 
+// Test RewindTo with the qwen-2.5 model. Exercises the static mask rewind path if
+// the EP supports it (DML by default, WebGPU with enableGraphCapture=1 set by
+// _test_utils.py), otherwise falls back to the dynamic mask path.
+// Skipped when qwen-2.5 model is not available.
+// TODO: This test fails on CUDA EP. Need to investigate the root cause.
+#if TEST_QWEN_2_5 && !USE_CUDA
```
kunal-vaishnavi, line 1440: Do you have a stack trace for why this test fails on the CUDA EP? This should be passing and we should have this test merged in without `!USE_CUDA`.

test/guidance_tests.cpp
```diff
@@ -111,7 +111,6 @@ TEST_F(GuidanceTests, UseJsonSchemaSingleTurn) {
   EXPECT_EQ(output, expected_output);
 }
 
-#if !USE_DML  // DML doesn't support continuous decoding
```
kunal-vaishnavi, line 114: Why remove this?

test/test_utils.h
```diff
@@ -14,11 +15,13 @@
 
 namespace test_utils {
 
-// Helper function to get the appropriate model path based on available models
+// Helper function to get the appropriate model path based on available models.
+// Caches results per model_type so different models resolve independently.
```
kunal-vaishnavi, line 21: If each ONNX model per EP is stored in a different directory path, why is there still some resolution needed?

test/c_api_tests.cpp
qjia7: Kept `!USE_CUDA` because CUDA CI currently fails this test on a pre-existing rewind issue that this PR exposed. Added an inline note in `test/c_api_tests.cpp`; we will fix CUDA rewind behavior in a separate follow-up PR and then remove this guard.

test/guidance_tests.cpp
qjia7: Good point. I will keep this untouched for now since the entire block is still disabled under `#if 0`, so this does not affect active test execution. We can revisit/restructure this when those tests are re-enabled.

test/test_utils.h
qjia7: Resolution is still needed because different CI/local runs may only have one EP artifact directory present. `GetModelPath` probes available EP directories and selects the first valid one containing `genai_config.json`; rationale is documented in `test/test_utils.h`.

qjia7:
@kunal-vaishnavi Please take another look. The windows CUDA CI seems not stable.

qjia7:
@kunal-vaishnavi @baijumeswani It seems that the CUDA CIs still fail.

test/test_utils.h
baijumeswani: Tests should run across all available eps. Not limited to the first non-empty ep model directory it finds (if possible).

test/python/_test_utils.py
```diff
@@ -126,12 +127,22 @@ def download_model(model_name, input_path, output_path, precision, device, one_l
...
+        extra_options += ["enable_cuda_graph=true"]
     if len(extra_options) > 1:
         command += extra_options
 
     run_subprocess(command).check_returncode()
 
 
+# Devices that support graph capture. Models with enable_graph_capture=True
+# are only built for these devices.
+_GRAPH_CAPTURE_DEVICES = {"webgpu", "dml", "nvtensorrtrtx"}
```
baijumeswani, line 151: Does any CI test nvtensorrtrtx?

baijumeswani: Can you also add a comment detailing that the cuda ep is explicitly turned off here and the reasons for it?

test/test_utils.h
qjia7: Agreed this is a good improvement. The current single-EP resolution is consistent with the existing test pattern used by `PHI2_PATH` and `IsEngineTestsEnabled()`, and is sufficient for this PR's scope. Refactoring to multi-EP parameterized tests would touch multiple test files and infrastructure -- will address in a follow-up PR.

test/python/_test_utils.py
qjia7: Addressed both questions in commit 20242cef:

1. **nvtensorrtrtx CI**: No, there is currently no CI pipeline that provisions the `phi3-fp16-nvtrt` model artifacts. The NvTensorRtRtx tests in `model_tests.cpp` and `c_api_tests.cpp` are all guarded with `GTEST_SKIP` when the model is not present, so they are effectively skipped in CI. Added a comment noting this.

2. **CUDA exclusion comment**: Added a comment above `_GRAPH_CAPTURE_DEVICES` explaining that CUDA is excluded because the Windows CUDA CI consistently fails to download this new model from Hugging Face. Also added a detailed comment in `c_api_tests.cpp` above the `!USE_CUDA` guard on `RewindQwen25CAPI` explaining the CUDA partial rewind bug (verified locally: `RewindTo(seq_len - 3)` produces incorrect output on CUDA with both static-mask and dynamic-mask code paths, across multiple models).

qjia7: Addressed both questions in commit 20242cef:

1. **nvtensorrtrtx CI**: No, there is currently no CI pipeline that provisions the `phi3-fp16-nvtrt` model artifacts. The NvTensorRtRtx tests in `model_tests.cpp` and `c_api_tests.cpp` are all guarded with `GTEST_SKIP` when the model is not present, so they are effectively skipped in CI. Added a comment noting this.

2. **CUDA exclusion comment**: Added a comment above `_GRAPH_CAPTURE_DEVICES` explaining that CUDA is excluded because the Windows CUDA CI consistently fails to download this new model from HF. Also added a detailed comment in `c_api_tests.cpp` above the `!USE_CUDA` guard explaining the CUDA partial rewind bug (verified locally with both static-mask and dynamic-mask paths, across multiple models).

baijumeswani (approved):

kunal-vaishnavi (approved):

PR #2100, opened 2026-04-24
Add k_quant_linear mixed-precision quantization for hybrid attention …

apsonawane:
Adds `k_quant_linear` as a new `int4_algo_config` option that promotes linear attention projections and their MLPs from INT4 to INT8. This improves accuracy for hybrid models like Qwen3.5 where linear attention recurrence accumulates quantization errors across the full sequence (unlike softmax attention which normalizes per-step).

### onnxruntime-genai
- **`qwen.py`**:
  - Added `make_int4_algo_config()` override with `k_quant_linear` support
  - Moved `layer_types` parsing before `super().__init__()` so `make_int4_algo_config` can access it during base class initialization
  - Removed the always-on INT8 quantization for linear attention layers (now opt-in via `int4_algo_config`)

### DIFF
--- a/src/python/py/models/builder.py
+++ b/src/python/py/models/builder.py
@@ -401,13 +401,14 @@ def get_args():
                     Use this option when you want to exclude certain nodes from being quantized.
                     Separate the node names with a ',' when passing them here (e.g. int4_nodes_to_exclude=/lm_head/MatMul,/model/embed_tokens/Gather)
                 int4_algo_config = Method for int4 quantization. Default is 'default'.
-                    Currently supported options are: 'default', 'rtn', 'rtn_last', 'k_quant', 'k_quant_mixed', 'k_quant_last'.
+                    Currently supported options are: 'default', 'rtn', 'rtn_last', 'k_quant', 'k_quant_mixed', 'k_quant_last', 'k_quant_linear'.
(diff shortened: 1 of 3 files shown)


kunal-vaishnavi:
Can we add `k_quant_linear` to the list of supported options when the help message is displayed?

https://github.com/microsoft/onnxruntime-genai/blob/19c2f7bbf349b1dbd083e6a9d57edf17fb99534c/src/python/py/models/builder.py#L413-L420

kunal-vaishnavi (approved):

xiaofeihan1:
Hi @apsonawane I tested the latest main branch with Qwen 3.5 0.8B using the following build command on NV5080, but encountered an accuracy regression (the model outputs gibberish/repetitive symbols). Have you encountered this?
`python src/python/py/models/builder.py -m Qwen/Qwen3.5-0.8B -o D:/models/Qwen3.5-0.8B-pr2100-simple  -e webgpu -p int4 --extra_options int4_accuracy_level=4 exclude_embeds=False` 

> (.venv) PS D:\ort-web-perf\ort-llm> python ort-genai.py --root D:/models --model qwen35-0.8b-test-2 --task generation --provider webgpu --use-iob --max_tokens=50
D:/models\Qwen3.5-0.8B-pr2100-simple
Process ID: 34148
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!`

PR #2105, opened 2026-04-27
Enable zero-copy for QNN GPU

qti-mattsinc:
* Use the new GPU shared memory allocator in QNN EP to
  allocate the KV cache on CPU-accessible GPU memory.
  This provides a large speedup by eliminating unnecessary
  copy overhead.
* Refactor QNN-specific checks out of the common
  `EnsureDeviceOrtInit` by adding `GetMemoryInfo` and
  `GetProviderOptionsForAllocatorSession` to `DeviceInterface`.


src/models/onnxruntime_api.h
qti-mattsinc, line 1: I had to modify this file and `onnxruntime_inline.h` to be able to access the `CreateMemoryInfo_V2` API. It looks like these files are manually checked-in?

Edit: Latest commit doesn't require changing these files

src/models/model.cpp
```diff
@@ -433,10 +432,29 @@ void EnsureDeviceOrtInit(DeviceInterface& device, const Config& config) {
...
+            use_dx12_shared_memory = true;
+            break;
+          }
+        }
+        break;
+      }
+    }
+    if (use_dx12_shared_memory) {
+      provider_options_list.back().options.emplace_back("enable_dx12_shared_memory_allocator", "1");
+      provider_options_list.back().device_filtering_options = Config::DeviceFilteringOptions { OrtHardwareDeviceType_GPU };
```
qti-mattsinc: note to self: would rather not touch device_filtering_options here if possible. Needs to be cleaned up in a way that still selects the correct allocator

Edit: The allocator session needs to select the GPU backend, so either device_filtering_options or the "backend_type" provider option is needed.

johnpaultaken (changes requested):
This change does not look ideal to me,  we need to discuss why it deviates from the norm. 
Ideally what I like to see is no change to genai at all. 
When OrtMemoryInfo is of type QnnShared can we just return the QnnGpuAllocator based on the device selection made by the user ie GPU ?
Everything else should work transparently, just like how it works for the other EPs cuda, openvino etc.
I don't see a need for enable_dx12_shared_memory_allocator option, why not enable it always in the EP ?
Also we cannot have variable names like use_dx12_shared_memory etc which then turns out to be only a Qnn specific option. Ideally, I would want to avoid any code in genai that are EP specific.

src/qnn/session_options.cpp
```diff
@@ -1,22 +1,126 @@
...
 
+static bool IsQNNGPUSharedAllocatorAvailable() {
+  // GPU shared memory allocator availability depends on both:
+  //   a. The graphics drivers installed on the system
+  //   b. The QNN EP package loaded by onnxruntime-genai
+  // If either dependency is out-of-date, the allocator will not be available.
+  // To check for support, we try to execute a single-node model through QNN
+  // with shared memory allocations and cache the result.
+  static std::optional<bool> cached_result = std::nullopt;
+  if (!cached_result) {
```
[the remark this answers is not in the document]
qti-mattsinc: Done

src/qnn/interface.cpp
```diff
@@ -49,6 +49,30 @@ struct InterfaceImpl : DeviceInterface {
...
+    auto provider_options = Config::ProviderOptions{"QNN", {}};
+
+    // QNN has separate allocators for NPU and GPU, and chooses which one to use based on the selected backend.
+    // So, we must use device_filtering_options to ensure the allocator session is created with the correct device.
+    // Otherwise, a conflict could arise between the allocator session and the main inference session (e.g. HTP
+    // allocations provided to a GPU session).
+    if (Generators::IsQNNGPUBackend(config)) {
+      provider_options.device_filtering_options = Config::DeviceFilteringOptions { OrtHardwareDeviceType_GPU };
+    } else {
+      // For the HTP backend, thea allocator for QnnHtpShared is only made available when the provider option
```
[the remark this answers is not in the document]
qti-mattsinc: Done

src/webgpu/interface.cpp
```diff
@@ -160,6 +160,22 @@ struct InterfaceImpl : DeviceInterface {
...
+      try {
+        return OrtMemoryInfo::Create("WebGPU_Buffer", OrtAllocatorType::OrtDeviceAllocator, 0, OrtMemType::OrtMemTypeDefault);
+      } catch (const Ort::Exception& fallback_e) {
+        throw std::runtime_error(
+            "Failed to create allocator for WebGPU. "
+            "Primary name 'WebGPU_Buf' error: " + std::string(e.what()) +
+            "; fallback 'WebGPU_Buffer' error: " + std::string(fallback_e.what()));
+      }
+    }
+  }
```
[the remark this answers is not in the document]
qti-mattsinc: > Since constructing an OrtMemoryInfo succeeds for any name string, the catch block is effectively unreachable,

No, the point at which the the exception is thrown for an invalid memory info name is in `OrtMemoryInfo::Create(...)`, not `Ort::Allocator::Create`, see [OrtApis::CreateMemoryInfo](https://github.com/microsoft/onnxruntime/blob/d2836a8f9a0814fd882c79cc91667c390537dad6/onnxruntime/core/framework/allocator.cc#L204). So, this change should be correct for WebGPU.

src/qnn/session_options.cpp
```diff
@@ -1,22 +1,77 @@
...
+
+  const auto session = OrtSession::Create(GetOrtEnv(),
+                                          Generators::GetTrivialModel().data(),
+                                          Generators::GetTrivialModel().size(),
+                                          session_options.get());
+
+  try {
+    const auto memory_info = GetQNNInterface()->GetMemoryInfo(config);
+    const auto allocator = Ort::Allocator::Create(*session, *memory_info);
+    return allocator != nullptr;
```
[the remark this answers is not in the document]
qti-mattsinc: `allocator` is `std::unique_ptr<Ort::Allocator>`. Comparing it against `nullptr` is reasonable.

src/smartptrs.h
```diff
@@ -98,12 +99,46 @@ enum struct DeviceType {
...
   virtual Ort::Allocator& GetAllocator() = 0;
+  virtual std::unique_ptr<OrtMemoryInfo> GetMemoryInfo(const Config& /* config */) const {
+    // Names for the device memory types used by 'OrtMemoryInfo::Create'
+    static const char* device_memory_type_names[] = {"CPU (Not used, see above)", "Cuda", "DML", "WebGPU_Buf", "QnnHtpShared", "OpenVINO (Not used, see above)", "Cuda", "Cpu"};
+    static_assert(std::size(device_memory_type_names) == static_cast<size_t>(DeviceType::MAX));
+
+    // Get the allocator from the OrtSession for the DeviceType (it's called 'AllocatorCreate' but it's really 'AllocatorGet')
+    auto name = device_memory_type_names[static_cast<int>(GetType())];
+    return OrtMemoryInfo::Create(name, OrtAllocatorType::OrtDeviceAllocator, 0, OrtMemType::OrtMemTypeDefault);
+  }
```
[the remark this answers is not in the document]
qti-mattsinc: Can be done as a follow up by those familiar with OpenVINO / CPU paths. The point of the refactor is just to get QNN specific branches out of the common code.

src/qnn/session_options.cpp
```diff
@@ -1,22 +1,77 @@
...
+  } catch (const Ort::Exception&) {
+    return false;
+  }
```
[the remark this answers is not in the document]
qti-mattsinc: `Ort::Exception` is the correct exception to catch here. Catching other exception types would potentially hide real issues that this function is not intended to handle.

src/qnn/session_options.cpp
```diff
@@ -1,22 +1,77 @@
...
           provider_options.options.begin(), provider_options.options.end(),
-          [](const auto& pair) { return pair.first == "enable_htp_shared_memory_allocator"; });
+          [](const auto& pair) { return pair.first == "enable_htp_shared_memory_allocator"
+                                        || pair.first == "enable_dx12_shared_memory_allocator"; });
       opt_it != provider_options.options.end() && opt_it->second == "1") {
-    device = GetDeviceInterface(DeviceType::QNN);
+    allocator_requested = true;
+  }
+
+  static const bool allocator_available = IsAllocatorAvailable(config);
```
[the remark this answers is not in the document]
qti-mattsinc: As it stands, `IsAllocatorAvailable` is really only used to check if QNN EP is exposing the Dx12SharedMemoryAllocator, which depends on a.) the version of QNN EP and b.) the gfx driver version  on the system. These won't change mid-process, so it's ok to leave this as-is for now.

src/smartptrs.h
```diff
@@ -7,6 +7,7 @@
 #include <atomic>
 #include <memory>
```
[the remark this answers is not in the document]
qti-mattsinc: Done

src/webgpu/interface.cpp
```diff
@@ -160,6 +160,22 @@ struct InterfaceImpl : DeviceInterface {
...
+    try {
+      return OrtMemoryInfo::Create("WebGPU_Buf", OrtAllocatorType::OrtDeviceAllocator, 0, OrtMemType::OrtMemTypeDefault);
+    } catch (const Ort::Exception& e) {
+      // WebGPU memory type name changed from "WebGPU_Buffer" to "WebGPU_Buf" in ORT 1.24.3.
+      // Try the old name before giving up.
+      try {
+        return OrtMemoryInfo::Create("WebGPU_Buffer", OrtAllocatorType::OrtDeviceAllocator, 0, OrtMemType::OrtMemTypeDefault);
+      } catch (const Ort::Exception& fallback_e) {
+        throw std::runtime_error(
+            "Failed to create allocator for WebGPU. "
```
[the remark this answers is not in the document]
qti-mattsinc: Done

src/models/model.cpp
```diff
@@ -424,56 +426,29 @@ void EnsureDeviceOrtInit(DeviceInterface& device, const Config& config) {
...
-  allocator.session_ = OrtSession::Create(GetOrtEnv(), g_trivial_model, sizeof(g_trivial_model), session_options.get());
-
-  // Names for the device memory types used by 'OrtMemoryInfo::Create'
-  static const char* device_memory_type_names[] = {"CPU (Not used, see above)", "Cuda", "DML", "WebGPU_Buf", "QnnHtpShared", "OpenVINO (Not used, see above)", "Cuda", "Cpu"};
-  static_assert(std::size(device_memory_type_names) == static_cast<size_t>(DeviceType::MAX));
+  // Use the trivial model to create an OrtSession that lets us get a device Ort::Allocator for each device type.
+  // This is necessary because the Ort::Allocator needs to persist and is valid for the lifetime of this OrtSession.
+  allocator.session_ = OrtSession::Create(GetOrtEnv(),
+                                          GetTrivialModel().data(),
+                                          GetTrivialModel().size(),
```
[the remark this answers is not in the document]
qti-mattsinc: Done

johnpaultaken (approved):
Looks cleaner than existing code.
Qnn specific checks are moved to qnn specifie interface now.
And so is web gpu specific code.

johnpaultaken:
@kunal-vaishnavi @baijumeswani   we need this change for IO binding to work on Qualcomm GPUs. This is required for the EP certification coming up soon. I mentioned it to Devang in a meeting about 2 weeks back. Is it possible to review and merge this change asap ?

kunal-vaishnavi:
> @kunal-vaishnavi @baijumeswani we need this change for IO binding to work on Qualcomm GPUs. This is required for the EP certification coming up soon. I mentioned it to Devang in a meeting about 2 weeks back. Is it possible to review and merge this change asap ?

There are a lot of CI failures in this PR. Please resolve the errors that are occurring.

src/smartptrs.h
```diff
@@ -98,12 +100,46 @@ enum struct DeviceType {
   MAX
 };
 
+static inline std::string DeviceTypeToString(DeviceType type) {
```
kunal-vaishnavi: There already exists a mapping. Can you re-use the existing one?

https://github.com/microsoft/onnxruntime-genai/blob/b8a784854b5bb08e26d1678092a91ef206e3e049/src/generators.cpp#L239-L260

src/models/model.cpp
```diff
@@ -395,16 +394,17 @@ int32_t Tokenizer::TokenToTokenId(const char* token) const {
...
+std::span<const uint8_t> GetTrivialModel() {
+  static const auto trivial_model = std::array<uint8_t, 96> {
+      0x08, 0x0a, 0x12, 0x01, 0x61, 0x3a, 0x53, 0x0a, 0x38, 0x12, 0x06, 0x76, 0x61, 0x6c, 0x75, 0x65,
+      0x73, 0x22, 0x08, 0x43, 0x6f, 0x6e, 0x73, 0x74, 0x61, 0x6e, 0x74, 0x2a, 0x24, 0x0a, 0x05, 0x76,
+      0x61, 0x6c, 0x75, 0x65, 0x2a, 0x18, 0x08, 0x01, 0x10, 0x01, 0x42, 0x0c, 0x63, 0x6f, 0x6e, 0x73,
+      0x74, 0x5f, 0x74, 0x65, 0x6e, 0x73, 0x6f, 0x72, 0x4a, 0x04, 0x00, 0x00, 0x00, 0x00, 0xa0, 0x01,
+      0x04, 0x12, 0x01, 0x62, 0x62, 0x14, 0x0a, 0x06, 0x76, 0x61, 0x6c, 0x75, 0x65, 0x73, 0x12, 0x0a,
+      0x0a, 0x08, 0x08, 0x01, 0x12, 0x04, 0x0a, 0x02, 0x08, 0x01, 0x42, 0x04, 0x0a, 0x00, 0x10, 0x15};
+
+  return std::span{trivial_model};
```
```suggestion
auto GetTrivialModel() {
  static constexpr std::array<uint8_t, 96> trivial_model{
      0x08, 0x0a, 0x12, 0x01, 0x61, 0x3a, 0x53, 0x0a, 0x38, 0x12, 0x06, 0x76, 0x61, 0x6c, 0x75, 0x65,
      0x73, 0x22, 0x08, 0x43, 0x6f, 0x6e, 0x73, 0x74, 0x61, 0x6e, 0x74, 0x2a, 0x24, 0x0a, 0x05, 0x76,
      0x61, 0x6c, 0x75, 0x65, 0x2a, 0x18, 0x08, 0x01, 0x10, 0x01, 0x42, 0x0c, 0x63, 0x6f, 0x6e, 0x73,
      0x74, 0x5f, 0x74, 0x65, 0x6e, 0x73, 0x6f, 0x72, 0x4a, 0x04, 0x00, 0x00, 0x00, 0x00, 0xa0, 0x01,
      0x04, 0x12, 0x01, 0x62, 0x62, 0x14, 0x0a, 0x06, 0x76, 0x61, 0x6c, 0x75, 0x65, 0x73, 0x12, 0x0a,
      0x0a, 0x08, 0x08, 0x01, 0x12, 0x04, 0x0a, 0x02, 0x08, 0x01, 0x42, 0x04, 0x0a, 0x00, 0x10, 0x15};

  return std::span{trivial_model};
```

src/models/model.cpp
```diff
@@ -426,56 +426,30 @@ void EnsureDeviceOrtInit(DeviceInterface& device, const Config& config) {
...
-  }
-  const std::vector<std::string> providers{device_type_names[static_cast<int>(type)]};
+  std::vector<Config::ProviderOptions> provider_options_list = {device.GetProviderOptionsForAllocatorSession(config)};
+  const std::vector<std::string> providers{provider_options_list.back().name};
   SetProviderSessionOptions(*session_options, providers, provider_options_list, true, config);
   session_options->SetLogSeverityLevel(ORT_LOGGING_LEVEL_ERROR);  // Errors only here, as warnings are not useful to the user
 
-  allocator.session_ = OrtSession::Create(GetOrtEnv(), g_trivial_model, sizeof(g_trivial_model), session_options.get());
+  // Use the trivial model to create an OrtSession that lets us get a device Ort::Allocator for each device type.
+  // This is necessary because the Ort::Allocator needs to persist and is valid for the lifetime of this OrtSession.
```
baijumeswani: I guess this comment is not capturing the essence of why we need to process wide scoped allocator. I think it misses the link between allocator for this session and the session being globally scoped.

The allocator needs to be a process wide scope because:

1. Some allocators have a hard shutdown process. For example with webgpu, it was that once you destroy the allocator, you cannot create one again.
2. Any allocator specific memory arena should persist beyond the lifetime of a specific model. So when multiple models are created one after another, we don't incur a heavy re-allocation cost.

Not that this comment is incorrect. But I think I can maybe add some more details in another separate PR.

src/models/model.cpp
baijumeswani: You can also inline it in the header so there is only 1 signature definition.

src/models/model.cpp
```diff
@@ -426,56 +426,30 @@ void EnsureDeviceOrtInit(DeviceInterface& device, const Config& config) {
...
   auto session_options = OrtSessionOptions::Create();
-  std::vector<Config::ProviderOptions> provider_options_list;
-  provider_options_list.emplace_back(Config::ProviderOptions{device_type_names[static_cast<int>(type)], {}});
-  // QnnHtpShared is a special case. This allocator is only made available when the provider option
-  // 'enable_htp_shared_memory_allocator' is set to 1.
-  if (type == DeviceType::QNN) {
-    provider_options_list.back().options.emplace_back("enable_htp_shared_memory_allocator", "1");
-  }
-  const std::vector<std::string> providers{device_type_names[static_cast<int>(type)]};
+  std::vector<Config::ProviderOptions> provider_options_list = {device.GetProviderOptionsForAllocatorSession(config)};
```
baijumeswani: This makes the qnn allocator sticky to the first model that got loaded? Is that expected? Any subsequent models being loaded will use the same allocator settings provided in the session options of the first model being used.

src/smartptrs.h
```diff
@@ -98,12 +100,46 @@ enum struct DeviceType {
...
+  }
+}
+
 struct DeviceInterface {
   virtual ~DeviceInterface() {}
 
   virtual DeviceType GetType() const = 0;
   virtual void InitOrt(const OrtApi& api, Ort::Allocator& allocator) = 0;
   virtual Ort::Allocator& GetAllocator() = 0;
+  virtual std::unique_ptr<OrtMemoryInfo> GetMemoryInfo(const Config& /* config */) const {
```
baijumeswani: `config` seems to be unused in all `GetMemoryInfo` implementations.

src/qnn/session_options.cpp
```diff
@@ -1,22 +1,78 @@
...
+  // If either dependency is out-of-date, the allocator will not be available.
+  auto session_options = OrtSessionOptions::Create();
+  Config::ProviderOptions provider_options = GetQNNInterface()->GetProviderOptionsForAllocatorSession(config);
+
+  if (!AppendExecutionProviderV2(*session_options, provider_options,
+                                 DeviceType::QNN, "QNNExecutionProvider")) {
+    AppendExecutionProviderV1(*session_options, provider_options);
+  }
+
+  session_options->SetLogSeverityLevel(ORT_LOGGING_LEVEL_ERROR);
```
baijumeswani, line 40: This overrides other log settings. We should avoid setting log level here.

src/qnn/session_options.cpp
```diff
@@ -1,22 +1,78 @@
...
+                                          session_options.get());
+
+  try {
+    const auto memory_info = GetQNNInterface()->GetMemoryInfo(config);
+    const auto allocator = Ort::Allocator::Create(*session, *memory_info);
+    return allocator != nullptr;
+  } catch (const Ort::Exception&) {
+    return false;
+  }
+}
```
baijumeswani, line 56: I don't think i understand what this is doing.

A globally scoped ort session and ort allocator was already created here: https://github.com/microsoft/onnxruntime-genai/pull/2105/changes#diff-3da792c72182817b1bdb08f4ec7eb7687fa19d9219aea8ccfdb310a1061a5c7bR439-R447

Why do we need to recreate a scoped session and allocator, and soon after dispose it?

src/qnn/session_options.cpp
```diff
@@ -1,22 +1,78 @@
...
+
+  if (Generators::IsQNNGPUBackend(config)) {
+    if (allocator_available) {
+      device = GetDeviceInterface(DeviceType::QNN);
+    } else if (allocator_requested) {
+      // Only warn if allocator requested but not available
+      Log("warning",
+          "Shared memory allocator for QNN GPU is not available!"
+          " Falling back to CPU allocations for the KV cache. This will reduce performance."
+          " To avoid this, try updating the QNN EP package and the graphics drivers on your system.");
```
baijumeswani, line 110: This needs to be behind if(g_log.enabled).

src/qnn/session_options.cpp
```diff
@@ -1,22 +1,78 @@
...
+    }
+  } else {
+    if (allocator_available) {
+      device = GetDeviceInterface(DeviceType::QNN);
+    }
```
baijumeswani:
```suggestion
  if (allocator_available) {
      device = GetDeviceInterface(DeviceType::QNN);
  } else if (Generators::IsQNNGPUBackend(config) && allocator_requested && g_log.enabled) {
    // Only warn if allocator requested but not available
    Log("warning",
        "Shared memory allocator for QNN GPU is not available!"
        " Falling back to CPU allocations for the KV cache. This will reduce performance."
        " To avoid this, try updating the QNN EP package and the graphics drivers on your system.");
  }
```

src/models/model.cpp
qti-mattsinc: `EnsureDeviceOrtInit` is called in the constructor of `Model::Model`. So if a client loads a new model with `OgaCreateModel`, `DeviceInterface::GetProviderOptionsForAllocatorSession` gets called as part of `EnsureDeviceOrtInit`. I'm not sure I follow why it would be sticky to the first model. What am I missing?

src/models/model.cpp
qti-mattsinc: Acknowledged. Note that this comment is just copied from above `g_trivial_model` which was moved here.

src/qnn/session_options.cpp
qti-mattsinc: `IsAllocatorAvailable` creates its own session in order to test for allocator support. If the allocator isn't available, the EP logs an error message about updating drivers (with some more detailed info than oga log added at https://github.com/microsoft/onnxruntime-genai/pull/2105/changes#diff-b93e01fc3193f016ad1ee8000ce7547346c421c799473199a67ea65168130201R65). I'd prefer to keep this trivial session at ORT_LOGGING_LEVEL_ERROR, so that it can surface this EP log without polluting the rest of the logs for the main session. 

Also note that for the global allocator session, oga is already choosing to set ORT_LOGGING_LEVEL_ERROR (https://github.com/microsoft/onnxruntime-genai/pull/2105/changes#diff-3da792c72182817b1bdb08f4ec7eb7687fa19d9219aea8ccfdb310a1061a5c7bR422), so doing the same thing here doesn't seem unprecedented.

src/qnn/session_options.cpp
qti-mattsinc: That globally scoped allocator isn't yet created at this point.

This code executes first, as part of the decision whether to have that global allocator that oga is going to use be either a.) the QNN shared memory allocator or b.) the CPU allocator (as a fallback path when lacking driver support). That check is happening at https://github.com/microsoft/onnxruntime-genai/pull/2105/changes#diff-b93e01fc3193f016ad1ee8000ce7547346c421c799473199a67ea65168130201R37.

Now, I do think that it would be better if, instead of having to create another session to check allocator support ahead of time, oga would have an (opt-in?) EP-agnostic fallback path that could defer to using the CPU allocations if creating the session for the global allocator fails (or at least have some mechanism to optionally communicate the allocator session upwards to `EnsureDeviceOrtInit`). But I'd consider that out of scope of this change

src/qnn/session_options.cpp
qti-mattsinc: Done

src/qnn/session_options.cpp
qti-mattsinc: Done

src/smartptrs.h
qti-mattsinc: Removed it with the latest commit. When first adding it, the config was necessary as QNN GPU vs NPU would determine some parameters of the memory info. But later we changed it on the EP side to use the same memory info for GPU and NPU.

src/models/model.cpp
qti-mattsinc: Done

src/smartptrs.h
qti-mattsinc: Good catch, I didn't notice the existing util. Done

qti-mattsinc:
> There are a lot of CI failures in this PR. Please resolve the errors that are occurring.

Resolved two valid errors in CI (lint and C++17 build issue). But some of the CI failures seem unrelated:

e.g.

( windows-directml-x64-build, windows-webgpu-x64-build)
> The self-hosted runner lost communication with the server. Verify the machine is running and has a healthy network connection. Anything in your workflow that terminates the runner process, starves it for CPU/Memory, or blocks its network access can cause this error.

and

(windows-cuda-x64-build)
> LINK : fatal error LNK1104: cannot open file 'onnxruntime-genai.lib' [D:\a\_work\onnxruntime-genai\onnxruntime-genai\examples\c\build\whisper.vcxproj]

@kunal-vaishnavi, are these jobs flaky?

src/models/model.cpp
baijumeswani: https://github.com/microsoft/onnxruntime-genai/blob/b8a784854b5bb08e26d1678092a91ef206e3e049/src/models/model.cpp#L389-L395

If the global allocator is already created for that device, it is not recreated.

qti-mattsinc: Oh I see. Yes that would make the QNN allocator sticky. This looks like it would apply to all EPs, not just QNN, though (but I suppose QNN is just the only one with different allocators to use depending on the config?). The latest commit (https://github.com/microsoft/onnxruntime-genai/pull/2105/changes/6c6515886102f0e413cbd256ca1a976ed45f6375) will let oga use the correct allocator in the case of an HTP model being loaded/unloaded followed by a GPU model (or vice versa).

kunal-vaishnavi:
> > There are a lot of CI failures in this PR. Please resolve the errors that are occurring.
> 
> Resolved two valid errors in CI (lint and C++17 build issue). But some of the CI failures seem unrelated:
> 
> e.g.
> 
> ( windows-directml-x64-build, windows-webgpu-x64-build)
> 
> > The self-hosted runner lost communication with the server. Verify the machine is running and has a healthy network connection. Anything in your workflow that terminates the runner process, starves it for CPU/Memory, or blocks its network access can cause this error.
> 
> and
> 
> (windows-cuda-x64-build)
> 
> > LINK : fatal error LNK1104: cannot open file 'onnxruntime-genai.lib' [D:\a_work\onnxruntime-genai\onnxruntime-genai\examples\c\build\whisper.vcxproj]
> 
> @kunal-vaishnavi, are these jobs flaky?

Yes, some of the CI failures are flaky. We are working on improving them but in the meantime I can re-run them for you.

src/models/model.cpp
baijumeswani: This PR modifies a significant amount of global state and moves it behind device interfaces. I suggest we avoid this approach for now, as it's unclear what impact it may have on real-world applications.

Instead of hiding the allocator logic internally, it may be better to introduce two DeviceInterface implementations for QNN: one for GPU and one for HTP.

This would allow the PR to focus only on QNN-specific code paths, rather than affecting surrounding components, which increases the overall integration risk.

qti-mattsinc: I agree that seems like a better solution. I've split things up into two device interfaces now and removed the change that would allow each interface to manage its own allocator session state.

> This would allow the PR to focus only on QNN-specific code paths, rather than affecting surrounding components, which increases the overall integration risk.

Do you also want to get rid of the added `GetMemoryInfo` and `GetProviderOptionsForAllocatorSession` members of the `DeviceInterface`? I think those help to simplify `EnsureDeviceOrtInit` by removing the EP-specific branches. But it's not necessary.

baijumeswani: We could decouple the refactoring from this PR since the goal for this PR is not that. We can keep this PR focused on enabling the qnn gpu allocator.
We can address the refactoring in another pull-request.

qti-mattsinc: Ok, I've removed the refactoring from this PR and can raise it separately afterwards.

src/models/model.cpp
```diff
@@ -439,19 +429,32 @@ void EnsureDeviceOrtInit(DeviceInterface& device, const Config& config) {
...
-  // 'enable_htp_shared_memory_allocator' is set to 1.
-  if (type == DeviceType::QNN) {
-    provider_options_list.back().options.emplace_back("enable_htp_shared_memory_allocator", "1");
+
+  if (type == DeviceType::QnnHtp || type == DeviceType::QnnGpu) {
+    const auto& config_providers = config.model.decoder.session_options.providers;
+    const auto& config_provider_options = config.model.decoder.session_options.provider_options;
+
+    // Copy the config QNN provider options to the allocator session.
+    // Certain options (e.g. "enable_htp_shared_memory_allocator") are required for QNN EP to expose an allocator.
```
baijumeswani: These are again sticky. Are there some defaults that can be used for qnn htp and qnn gpu global ort sessions so the session is not tied to the first model that is loaded?

qti-mattsinc: The behavior on mainline is also sticky, but it's just that there is only one allocator recognized (the HTP shared memory allocator). Now, we have two allocators (HTP shared and DX12 shared), but we've split `DeviceType::QNN` into `DeviceType::QnnHtp` and `DeviceType::QnnGpu` so HTP and GPU allocator sessions do not step on each other.

The fact that the allocator for `DeviceType::QnnHtp` is sticky is ok because there is only one allocator that can be used for that device (i.e. the HTP shared memory allocator), and this is the same behavior as mainline currently. Likewise, for `DeviceType::QnnGpu`, only one allocator can be used (the DX12 allocator).

Also note: only the "enable_htp_shared_memory_allocator" and "enable_dx12_shared_memory_allocator" provider options (as well as the "backend_type"/"backend_path" options which chooses HTP vs GPU) are relevant to the allocator session. Maybe it would make it more obvious if we only copied over those options over when creating the allocation session.

qti-mattsinc: bump @baijumeswani.

src/models/model.cpp
```diff
@@ -439,19 +429,32 @@ void EnsureDeviceOrtInit(DeviceInterface& device, const Config& config) {
   }
 
   provider_options_list.emplace_back(std::move(init_session_provider_options));
-  // QnnHtpShared is a special case. This allocator is only made available when the provider option
-  // 'enable_htp_shared_memory_allocator' is set to 1.
-  if (type == DeviceType::QNN) {
-    provider_options_list.back().options.emplace_back("enable_htp_shared_memory_allocator", "1");
+
+  if (type == DeviceType::QnnHtp || type == DeviceType::QnnGpu) {
```
kunal-vaishnavi: Let's move such device-specific logic into the device-specific interface. We should find a way to design this so that we don't need to introduce more `if A, then B, else C` logic for the provider options list.

src/models/model.cpp
```diff
@@ -402,7 +392,7 @@ void EnsureDeviceOrtInit(DeviceInterface& device, const Config& config) {
   // This ensures memory allocated on-device for model inputs/outputs is valid for the lifetime of GenAI.
 
   // Names for the device types used by 'SetProviderSessionOptions'
-  static const char* device_type_names[] = {"CPU (Not used, see above)", "cuda", "DML", "WebGPU", "QNN", "OpenVINO (Not used, see above)", "NvTensorRtRtx", "RyzenAI"};
+  static const char* device_type_names[] = {"CPU (Not used, see above)", "cuda", "DML", "WebGPU", "QNN", "QNN", "OpenVINO (Not used, see above)", "NvTensorRtRtx", "RyzenAI"};
```
kunal-vaishnavi, line 428: We should find a way to simplify this for the future. It is odd at first glance that `QNN` has to be written twice here. Instead, we could move from parallel string arrays to perhaps `GetProviderName()` and `GetMemoryInfoName()` APIs on `DeviceInterface`.

src/models/model.cpp
qti-mattsinc: @kunal-vaishnavi, I had moved this into the device specific interface originally. However, following the advice from @baijumeswani  in https://github.com/microsoft/onnxruntime-genai/pull/2105#discussion_r3425756628, I reverted it in this PR and plan to raise the refactoring in a follow-up.

src/models/model.cpp
qti-mattsinc: Agreed, but same response as to your other comment. We can leave the refactoring for a follow-up.

src/models/model.cpp
qti-mattsinc: https://github.com/microsoft/onnxruntime-genai/pull/2232 introduced `ShapeInitSessionProviderOptions`. I've used that now after rebasing.

src/qnn/session_options.cpp
```diff
@@ -1,28 +1,107 @@
...
+  bool dx12_shared_allocator_requested = false;
+  for (const auto& opt : provider_options.options) {
+    if (opt.first == "enable_htp_shared_memory_allocator" && opt.second == "1") {
+      htp_shared_allocator_requested = true;
+    } else if (opt.first == "enable_dx12_shared_memory_allocator" && opt.second == "1") {
+      dx12_shared_allocator_requested = true;
+    }
   }
+
+  if (Generators::IsQNNGPUBackend(config)) {
```
baijumeswani: IsQNNGPUBackend us only used from this file. Can it be defined in an unnamed namespace here?

baijumeswani (commented):
Added one comment. Otherwise, changes look good.

src/qnn/session_options.cpp
qti-mattsinc: Done in latest commit

baijumeswani (approved):

kunal-vaishnavi (approved):

qti-mattsinc:
Are these CI failures flaky / do they need to be re-run? I don't have access to Azure pipelines to view the logs for the failing integration tests, but I can see that the windows-cuda-x64-build failed because the test runner lost connection.

kunal-vaishnavi:
> Are these CI failures flaky / do they need to be re-run? I don't have access to Azure pipelines to view the logs for the failing integration tests, but I can see that the windows-cuda-x64-build failed because the test runner lost connection.

Some of these CI failures are flaky. I am re-running them now and hopefully they will pass.

qti-mattsinc:
> > Are these CI failures flaky / do they need to be re-run? I don't have access to Azure pipelines to view the logs for the failing integration tests, but I can see that the windows-cuda-x64-build failed because the test runner lost connection.
> 
> Some of these CI failures are flaky. I am re-running them now and hopefully they will pass.

@kunal-vaishnavi, looks like they've failed again. They are all failing after 1m so I assume it's still a test runner issue, but I can't view the [Azure link](https://aiinfra.visualstudio.com/6a833879-cd9b-44a4-a9de-adc2d818f13c/_build/results?buildId=1287831&view=logs&jobId=75da2411-aabb-5d40-b2f2-bf9bcf2c965e) that github actions references, so it's somewhat of a black-box for me.

src/qnn/interface.cpp
```diff
@@ -70,20 +65,62 @@ struct InterfaceImpl : DeviceInterface {
...
   }
 };
 
 }  // namespace QNN
 
-DeviceInterface* GetQNNInterface() {
-  static std::unique_ptr<DeviceInterface> g_device = std::make_unique<QNN::InterfaceImpl>();
-  return g_device.get();
+DeviceInterface* GetQNNInterface(DeviceType device_type) {
+  assert(type == DeviceType::QnnHtp || type == DeviceType::QnnGpu);
```
qti-mattsinc, line 112: I've just noticed there's a typo here that will cause a Debug build to fail. I was building for Release and didn't notice it when testing. It should be:

```suggestion
  assert(device_type == DeviceType::QnnHtp || device_type == DeviceType::QnnGpu);
```

I've put the fix in #2259, along with the MemoryInfo refactoring follow-up changes that Kunal requested.

FYI @baijumeswani / @kunal-vaishnavi

PR #2106, opened 2026-04-29
Release captured graph resources when generator is destroyed

qjia7:
Call ReleaseGraph on the ORT session in State::~State() to free per-graph buffer resources in GPU. This pairs with the per-graph buffer manager changes in ORT https://github.com/microsoft/onnxruntime/pull/28260.

Changes:
- Add OrtSession::ReleaseGraph wrapper calling the new C API
- Store graph_capture_session_ pointer in State during graph capture runs
- Call ReleaseGraph in State destructor with the graph annotation ID


src/models/model.cpp
```diff
@@ -222,6 +224,21 @@ void State::SetActiveAdapter(Adapters* adapters, const std::string& adapter_name
...
+  // Note: the EP contract is to no-op when the id was never captured (e.g., when
+  // graph_capture_this_run was always false for this State). The call is wrapped
+  // in try/catch because destructors must not throw -- a throw during unwinding
+  // would call std::terminate.
+#if ORT_API_VERSION >= 27
+  if (graph_capture_session_ && graph_id_value_ > 0) {
+    try {
+      graph_capture_session_->ReleaseCapturedGraph(graph_id_value_);
+    } catch (...) {
+      // Best-effort cleanup; swallow to keep the destructor non-throwing.
```
baijumeswani, line 238: We can add a log statement here if that will help with debugging.

src/models/model.cpp
```diff
@@ -63,7 +63,8 @@ State::State(const GeneratorParams& params, const Model& model)
     std::random_device rd;
     std::mt19937 gen(rd());
     std::uniform_int_distribution<> dis(1, INT_MAX);
-    graph_id_ = std::to_string(dis(gen));
+    graph_id_value_ = dis(gen);
+    graph_id_ = std::to_string(graph_id_value_);
```
baijumeswani, line 68: Why does this need to be a random value?

qjia7: Good question - this isn't something this PR introduced. The random `graph_id_value_` has been here since PR #1305 (`37cddd40`, 2025-03-19) when DML graph capture was first added. This PR is just adding the matching `ReleaseCapturedGraph` cleanup on the genai side; the id-generation strategy is unchanged.

I agree it's worth a closer look (a per-session monotonic counter would be sufficient since the EP scopes ids per session), but I'd prefer to address that in a follow-up PR so we keep this one focused on the per-graph buffer manager wiring.

src/models/model.cpp
qjia7: Done

qjia7:
It seems that the webgpu CI is not stable. The failure should not be related with this PR's changes.

src/models/onnxruntime_api.h
```diff
@@ -786,6 +786,10 @@ struct OrtSession {
 
   void SetEpDynamicOptions(_In_opt_ const char* const* keys, const char* const* values, size_t kv_len);
 
+#if ORT_API_VERSION >= 27
```
kunal-vaishnavi, line 812: Is the version guard needed since we ideally want to use ORT 1.27?

kunal-vaishnavi (dismissed):

src/models/onnxruntime_api.h
qjia7: This can be removed once CUDA upgrades to 1.27 as well. Currently, CUDA is still pinned to 1.26 because version 1.27 requires CUDA 13. I’d prefer to make this change in a separate PR.

cmake/ortlib.cmake
```diff
@@ -80,7 +80,7 @@ if(ORT_HOME)
   endif()
 else()
   # If ORT_HOME is not specified, download the onnxruntime headers and libraries from the nightly feed
-  set(ORT_VERSION "1.26.0")
+  set(ORT_VERSION "1.27.0")
```
baijumeswani: Please keep this at 1.26.0 for now. If 1.27 is needed for webgpu, we can externally supply ort_home. But let's not use 1.27.0 as default build version of ort for now.

qjia7: Done

baijumeswani (approved):

kunal-vaishnavi (approved):

PR #2109, opened 2026-04-29
Removes QNN packaging from onnxruntime-genai pipelines

baijumeswani:


.github/workflows/win-cpu-arm64-build.yml
```diff
@@ -106,8 +106,7 @@ jobs:
           python -m pip install -r test\python\requirements.txt --user
           python -m pip install -r test\python\cpu\torch\requirements.txt --user
           python -m pip install -r test\python\cpu\ort\requirements.txt --user
-          python -m pip install onnxruntime-qnn -i https://aiinfra.pkgs.visualstudio.com/PublicPackages/_packaging/ORT-Nightly/pypi/simple/
-          python -m pip install (Get-ChildItem ("$env:binaryDir\wheel\*.whl")) --no-deps
+          python -m pip install (Get-ChildItem ("$env:binaryDir\wheel\*.whl"))
```
qti-kromero, line 109: small comment here cause I noticed this was installed by default - so user would then be expected to install onnxruntime-qnn separately after this change is made?

further, are there plans to move towards using extras to track these dependencies or should we look to add onnxruntime-genai-qnn to upstream repository as it appears others have done?

baijumeswani: The idea is that ep specific packages should no longer be needed as the default `onnxruntime-genai` should work with the default `onnxruntime` package. If an EP needs to be plugged in, then they should invoke `og.register_execution_provider_library` on onnxruntime-genai and that would by default enable the specific plug-in ep.

qti-kromero: got it for first question but still unclear regarding second question - would the other IHV specific OGA packages be removed in the future such that users need to install their plugin EP separately? or would those still be maintained as a shortcut of installing OGA + ORT + IHV plugin EP

baijumeswani: We do not publish any other IHV specific packages. `onnxruntime-genai-qnn` was the only one since onnxruntime was publishing `onnxruntime-qnn`.

kunal-vaishnavi (approved):

PR #2110, opened 2026-04-30
Fix heap overflow issue

apsonawane:
This pull request enhances the safety and robustness of the `QwenImageProcessor` by adding checks to prevent integer overflows when processing image dimensions and calculating buffer sizes. These changes help avoid potential runtime errors and security issues when handling large or malformed images.

**Image dimension and overflow validation:**

* Added validation to ensure `height_patches`, `width_patches`, and `channels` are positive before proceeding with patch calculations.
* Introduced explicit checks to prevent integer overflow when computing `total_patches`, `patch_dim`, and the total buffer size for patched image data.
* Included `<limits>` header to support overflow checks using `std::numeric_limits`.


src/models/qwen2_5_vl_image_processor.cpp
```diff
@@ -231,8 +232,40 @@ std::unique_ptr<NamedTensors> QwenImageProcessor::Process(const Tokenizer& token
...
+    patch_dim *= kPatchSize;
+    if (patch_dim > std::numeric_limits<int64_t>::max() / kPatchSize) {
+      throw std::runtime_error("Integer overflow computing patch_dim");
+    }
+    patch_dim *= kPatchSize;
+
+    // Check overflow for total buffer size = total_patches * patch_dim
+    if (total_patches > std::numeric_limits<int64_t>::max() / patch_dim) {
+      throw std::runtime_error("Integer overflow computing patched buffer size");
+    }
```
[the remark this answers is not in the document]
apsonawane: Not needed keep it int64_t

src/models/qwen2_5_vl_image_processor.cpp
```diff
@@ -231,57 +232,103 @@ std::unique_ptr<NamedTensors> QwenImageProcessor::Process(const Tokenizer& token
...
+      throw std::runtime_error("Invalid channel count for image patching: channels=" +
+                               std::to_string(channels));
+    }
+
+    // Image too small to form any patches — skip patching entirely.
+    // The downstream code will use the original (unpatched) pixel_values.
+    if (height_patches <= 0 || width_patches <= 0) {
+      // Fall through: patched_pixel_values remains null, so the else branch
+      // (non-patched pixel_values path) handles it.
+    } else {
```
kunal-vaishnavi: Rather than having a very large `else` branch, can we have an early return for the `if` branch? That would also better clarify what changed in this PR.

apsonawane: We cannot return or continue, just flipped the condition

kunal-vaishnavi (approved):

PR #2111, opened 2026-04-30
Fix AppendNextTokensToSequences heap overflow

apsonawane:
### Summary

Adds a fast-fail guard at the entry of `Generator::GenerateNextToken()` that throws a clear `std::runtime_error` when the sequence has already reached `max_length`. This prevents a heap-buffer-overflow write into the preallocated sequences buffer that is otherwise reachable via the C API.

### Reported bug (AddressSanitizer)

A libFuzzer harness driving the `OgaGenerator_*` C API triggered the following crash:

```
==ERROR: AddressSanitizer: heap-buffer-overflow ... WRITE of size 4
  #0 Generators::GreedySearch_Cpu::AppendNextTokensToSequences()  src/search.cpp:323
  #1 Generators::GreedySearch_Cpu::SampleTopK(int, float)         src/search.cpp:192
  #2 Generators::Generator::GenerateNextToken()                   src/generators.cpp:602
  #3 OgaGenerator_GenerateNextToken                               src/ort_genai_c.cpp:473
  #4 <fuzzer harness>

0 bytes after 64-byte region  ← writes at index max_length
allocated by Generators::Sequences::Sequences(...) at construction
```

### Root cause

Inside `GreedySearch_Cpu::AppendTokens`, `ResetDone()` is called unconditionally at the end. If the previous call filled the sequences buffer up to exactly `max_length`, `done_` is cleared. A subsequent `GenerateNextToken()` then:

1. Enters `SampleTopK()`.
2. The `!done_` check falsely passes (buffer is actually full).
3. `AppendNextTokensToSequences()` writes at index `max_length` → OOB write into the redzone immediately following the 64-byte sequences buffer.

### Fix

Add an early runtime error at the top of `Generator::GenerateNextToken()`:

```cpp
if (search_->GetSequenceLength() >= state_->params_->search.max_length)
  throw std::runtime_error(
      "GenerateNextToken called with sequence length already at max_length (" +
      std::to_string(state_->params_->search.max_length) + ")");
```

This sits directly above frame #2 of the crash stack, so the OOB path in frames #0 and #1 is unreachable. The exception propagates through `OgaGenerator_GenerateNextToken` and is surfaced to the caller as a normal API error instead of memory corruption.


kunal-vaishnavi:
The changes in this PR seem to indicate a larger issue inside ORT GenAI. The `done` state should reflect the true state, and we should not have to insert max length checks in so many locations. Do we know why this issue is happening in the first place?

src/cuda/search_cuda.cpp
```diff
@@ -166,12 +166,12 @@ void GreedySearch_Cuda::SampleTopKTopP(int k, float p, float temperature) {
...
   // Append tokens
   CUDA_CHECK(cudaStreamSynchronize(GetStream()));
-  if (!*done_cpu_) {
+  if (!*done_cpu_ && sequences_.GetSequenceLength() < static_cast<size_t>(params_->search.max_length)) {
     cuda::Launch_AppendNextTokensToSequences(next_tokens_buffer_.Span(), sequences_.GetSequences().Span(), params_->BatchBeamSize(), sequences_.GetSequenceLength(), sequences_.max_length_, GetStream());
     sequences_.AfterAppendNextTokens(next_tokens_buffer_, params_->BatchBeamSize());
   }
 
-  if (sequences_.GetSequenceLength() == params_->search.max_length) {
+  if (sequences_.GetSequenceLength() >= params_->search.max_length) {
```
baijumeswani: GetSequenceLength starts at 0 and increases as token count increases. max_length is expected to be non zero. We have checks in place in generators.cpp to ensure that the tokens appended do not exceed max_length.

I am not sure I see where the heap overflow may occur. Perhaps we need to document where the checks are happening so we have a good understanding. But I do not think there is a need for this change.

Was there a heap-overflow encountered at runtime?

apsonawane: heap-overflow is caused by this
1. `AppendTokens()` fills buffer to `max_length`, then `ResetDone()` clears done_
2. Next `GenerateNextToken()` -> `SampleTopk()` -> `!done_` passes -> `AppendNextTokensToSequences()` writes at index  `max_length` -> OOB write

baijumeswani (approved):

PR #2125, opened 2026-05-06
Add JSON DOM and RFC 7386 merge_patch to src/json

xiaoyu-work:
Adds a lightweight in-memory JSON DOM (Document / Object / Array) on top of the existing streaming parser, plus serializer and RFC 7386 JSON Merge Patch. This is the foundation for v4 model-package per-variant overlays, which carry a JSON Merge Patch in consumer_metadata.genai_config_overlay and need to be merged into the package-shipped base genai_config.json before the existing streaming-parser-driven Config loader sees it.

The new APIs are additive — every existing JSON::Element-based config loader is unaffected. The DOM is intentionally minimal (no fancy number representations, no object-key insertion order, no schema validation); its only consumers are Config-time overlay merging and the unit tests.

Also fixes a pre-existing off-by-one in the streaming parser's Skip(literal) bounds check that silently dropped a top-level true / false / null literal when it sat at the very end of the input buffer. The bug never surfaced in real genai_config.json files because every keyword there is followed by a comma, brace or bracket, but the new DOM round-trip tests exercise it directly.

Tests: test/json_merge_patch_test.cpp covers the full RFC 7386 example table plus GenAI-relevant shapes (overlay overrides context_length and I/O names; pipeline arrays replace wholesale; null deletes optional fields) and a small set of edge cases beyond the RFC.

### DIFF
--- a/src/json.cpp
+++ b/src/json.cpp
@@ -3,7 +3,11 @@
 
 #include <cmath>
 #include <charconv>
+#include <cstdio>
+#include <limits>
+#include <list>
 #include <sstream>
+#include <type_traits>
 
 namespace JSON {
 static constexpr const char* value_names[] = {"string", "number", "bool", "null"};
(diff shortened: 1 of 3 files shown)


xiaoyu-work:
Folded into the umbrella draft PR #2136. Closing to keep a single review surface for the v4 package integration.

PR #2126, opened 2026-05-06
Add Config::shared_assets_path for shared tokenizer/processor assets

xiaoyu-work:
Introduce a new fs::path field next to Config::config_path to address the directory holding tokenizer files, processor JSON, chat template, and other assets that are common across components (decoder/vision/speech/encoder/...).

In flat-dir mode this field is initialized to config_path in Config::Config(path, overlay), so behavior is identical. In a future ORT v4 package mode the package loader will set it to <pkg>/configs/.

Rewrite all call sites that load shared assets to use shared_assets_path:
- Tokenizer (model.cpp:308) and constrained_logits_processor (tokenizer.json).
- Vision/multimodal processor configs in gemma/gemma4/mistral3/phi/phi-multimodal/qwen2_5_vl.
- Whisper speech processor config.

Per-component model assets (ONNX, LoRA, custom_ops, external data) and EP-private paths (OpenVINO cache_dir, RyzenAI/VitisAI model_root) are intentionally NOT migrated — they are addressed per-component variant folder and require deeper schema work in a follow-up workstream.

This is preparatory plumbing for the ORT v4 model-package format integration.

### DIFF
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -1567,7 +1567,7 @@ void OverlayConfig(Config& config, std::string_view json) {
   JSON::Parse(element, json);
 }
 
-Config::Config(const fs::path& path, std::string_view json_overlay) : config_path{path} {
+Config::Config(const fs::path& path, std::string_view json_overlay) : config_path{path}, shared_assets_path{path} {
   ParseConfig(path / "genai_config.json", json_overlay, *this);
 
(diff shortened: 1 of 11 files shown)


xiaoyu-work:
Folded into the umbrella draft PR #2136 (via cherry-pick of the same patch). Closing to keep a single review surface.

Issue #1699, opened 2025-08-23
Support dedicated session and provider options for each model in VLM


uday610:
The latest VLMs generally consist of multiple models such as Vision, Embedding, and LLM. For efficient acceleration, it typically requires to configure these models with different session options and provider options. This gives flexibility when multiple hardware resources are available (for example, on an AI PC with CPU, GPU, and NPU), so that each model can have its own configuration and potentially run on different hardware.

This doesn’t seem possible today. In the code [multi_modal.cpp#L78](https://github.com/microsoft/onnxruntime-genai/blob/v0.9.0/src/models/multi_modal.cpp#L78)
, the embedding model’s session options are strictly tied to the text model’s session options. The same applies to the vision model. If we try to set session options for either the embedding or vision model, the flow errors out.

Fundamentally, ONNX Runtime supports multiple models through different sessions, with each session having its own session options and provider options. Since ONNX Runtime GenAI is built on top of ONNX Runtime, it should allow developers to do the same for VLMs, which are typically composed of multiple sub-models.

Thank you

kunal-vaishnavi:
Support for this has been added in [this PR](https://github.com/microsoft/onnxruntime-genai/pull/1795). Using different execution providers between the ONNX models is not in the scope of that PR. That scenario would require more work to support.

PR #2129, opened 2026-05-06
Add CUDA multimodal debugging skill

justinchuby:
Skill-only PR (no code changes). Documents 11 CUDA issues found during Gemma4 multimodal enablement with debugging methodology.


kunal-vaishnavi:
This skill file is more applicable to ONNX Runtime and should be added there. Regarding the GenAI-specific notes in the file, the TopK issue is a bug that we can open and fix for all models and the GenAI config [currently requires that the same EP is used in all ONNX models](https://github.com/microsoft/onnxruntime-genai/issues/1699).

PR #2130, opened 2026-05-06
[v4-pkg W9] Pin layer-2 config-overlay contract with regression tests

xiaoyu-work:
Part of the ORT v4 model-package integration. **No production code changes** — pure documentation + tests. Independent of all other v4 work.

## What

Adds `test/runtime_settings_test.cpp` covering the two layer-2 channels that override `genai_config.json` values at runtime:

1. **`RuntimeSettings::GenerateConfigOverlay()`** — handle-driven JSON emitted at `Model::Create`. Today only `dawnProcTable` is recognised and projected to `model.decoder.session_options.provider_options[*].WebGPU.dawnProcTable`.
2. **`OverlayConfig()` / `OgaConfigOverlay()`** — caller-driven JSON exposed through C#, Java, Python, and ObjC bindings.

Both feed through the same JSON parser registered for `Config`, so any field writable from `genai_config.json` is also writable via these runtime channels. The new tests pin that contract so future refactors of the overlay-merge logic don't silently break it.

## Tests

* Empty handles → empty overlay (unconditional feed-through contract)
* Unrelated handle keys are silently ignored
* `dawnProcTable` overlay JSON path is asserted explicitly: `model` / `decoder` / `session_options` / `provider_options` / `WebGPU` / `dawnProcTable` keywords + decimal-encoded handle value (regression guard for every WebGPU consumer on Windows)
* `OverlayConfig` applies a `search.top_k` override and re-applies cleanly without state leaking
* End-to-end: generate the WebGPU overlay → parse via `OverlayConfig` → assert WebGPU `provider_options` entry materializes with `dawnProcTable` value preserved

## Doc comments

Inline doc comments added on `RuntimeSettings::GenerateConfigOverlay` and `OverlayConfig` describing the layer-2 contract and pointing at the regression suite.

## Stack

Independent of W1 (#2125), W4-prep (#2126), W5c-prep. All 4 v4-prep branches are siblings on `main`.

### DIFF
--- a/src/config.h
+++ b/src/config.h
@@ -430,6 +430,25 @@ void SetSearchNumber(Config::Search& search, std::string_view name, double value
 void SetSearchBool(Config::Search& search, std::string_view name, bool value);
 void ClearProviders(Config& config);
 void SetProviderOption(Config& config, std::string_view provider_name, std::string_view option_name, std::string_view option_value);
+// Layer-2 override channel (caller-driven).
+//
+// Applies a JSON document to an existing Config in place, using the same
+// JSON parser registered for genai_config.json. Any field that can be set via
+// genai_config.json can therefore also be overridden here. The JSON parser
+// merges objects key-wise (ProviderOptionsObject_Element matches by `name`),
+// so e.g. an overlay that adds a single WebGPU provider_option does NOT
+// replace pre-existing provider_options entries for other providers.
+//
+// Used by:
+//   * OgaConfigOverlay() — exposed to C# (Config.Overlay), Java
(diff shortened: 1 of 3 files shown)


xiaoyu-work:
Closing as standalone — the value is mostly as a safety net for W3/W5/W6 (v4 model-package work that touches the same overlay merge logic). Will be folded into a later PR that actually exercises the layer-2 channel changes, so the regression context is co-located with the change. Branch `feature/v4-package-w9-runtime-settings-tests` retained locally.

PR #2131, opened 2026-05-06
[v4-pkg W5c-prep] Extract EffectiveSessionOptions helper for non-decoder components

xiaoyu-work:
Part of the ORT v4 model-package integration. **Behavior change is zero** on every flat-directory model. Independent of W1, W4-prep, W9.

## What

Five non-decoder session-creation paths inline the same fallback ternary:

\\\cpp
config_->model.<component>.session_options.has_value()
    ? config_->model.<component>.session_options.value()
    : config_->model.decoder.session_options
\\\

— at `marian.cpp` (encoder), `whisper.cpp` (encoder), and `multi_modal.cpp` × 3 (vision / speech / embedding).

Replaced with a single free helper `EffectiveSessionOptions(config, component_session_options)` declared in `config.h` next to `OverlayConfig` and the other layer-2 free helpers.

## Why

Today the helper is purely cosmetic. In the v4 model-package world the same call sites need a single seam where the **layer-1 baseline** (`variant.json files[].session_options`) gets merged in front of the today-only **layer-2 view** (`genai_config`). Centralising the lookup here means the v4 work touches one function body, not five ternaries scattered across three files.

## Tests

`test/config_helpers_test.cpp` covers the three contract bullets that matter for the call sites:

1. Fallback identity to decoder when component is `nullopt` (must alias decoder storage, not copy)
2. Identity to the component's stored value when set (must alias the optional's storage)
3. Live-aliasing of the decoder fallback (mutations after the call are visible through the returned reference, so future refactors don't accidentally start copying `SessionOptions`)

## Stack

Independent of all other v4-prep branches. Reviewable / mergeable in any order relative to W1 (#2125), W4-prep (#2126), W9 (#2130).

### DIFF
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -1567,6 +1567,13 @@ void OverlayConfig(Config& config, std::string_view json) {
   JSON::Parse(element, json);
 }
 
+const Config::SessionOptions& EffectiveSessionOptions(
+    const Config& config,
+    const std::optional<Config::SessionOptions>& component_session_options) {
+  return component_session_options.has_value() ? component_session_options.value()
+                                               : config.model.decoder.session_options;
+}
+
 Config::Config(const fs::path& path, std::string_view json_overlay) : config_path{path} {
   ParseConfig(path / "genai_config.json", json_overlay, *this);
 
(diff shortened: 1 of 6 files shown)


xiaoyu-work:
Folded into the umbrella draft PR #2136 (via cherry-pick of the same patch). Closing to keep a single review surface.

PR #2132, opened 2026-05-06
[v4-pkg W2-stub] Add ModelPackageContext stub directory walker

xiaoyu-work:
Stacks on #2125 (W1 JSON DOM + merge_patch). Review/merge after #2125 lands.

## What

Adds an internal abstraction over the ORT v4 model-package layout that will be wired into `Config::Config` in a follow-up PR. The single public surface is the abstract `ModelPackageContext` / `ComponentInstance` pair declared in `src/models/model_package.h`.

Today the only implementation is a stub directory walker that reads the package's `manifest.json` + per-variant `variant.json` files using the JSON DOM from #2125. When ORT v4 lands, an alternate implementation backed by `OrtModelPackageContext` will slot in alongside the stub; callers see only the abstract surface.

## Public surface

- `Open(path)` — returns `nullptr` if no `manifest.json` (caller falls back to flat-dir mode), throws on a malformed manifest. The contract is: nullptr means "not a v4 package", throw means "broken v4 package".
- `NumComponents` / `ComponentName(cix)` — manifest-order traversal.
- `EpsCompatibleWith(cix)` — union of EP names per-component-variants, consumed by the EP-defaulting algorithm (W5b) which intersects across components.
- `SelectComponent(cix, ep_name)` — returns a `ComponentInstance` for a matching variant, or `nullptr`.
- `SharedAssetsPath` / `VariantFolderPath` / `GenAIConfigOverlay` — paths consumed by tokenizer/processor loaders (W4-real) and by Config's per-variant overlay merge (W3).

## Notes

- Variant directory iteration is sorted before traversal so test and EP-defaulting behaviour is stable across platforms (filesystem iterators expose implementation-defined order).
- The implementation bridges between the codebase's custom `fs::path` (from `src/filesystem.h`, the public surface) and `std::filesystem::path` (needed for directory iteration, which `fs::path` does not support).
- `files[]` and other ORT-internal metadata are deliberately not surfaced here — they are pipeline-runner / W6 territory.

## Dead code

This change is dead code from the rest of the codebase's perspective — `Config` does not yet consult `ModelPackageContext`. Wiring follows in W3 (config detection + base load + overlay merge). Tests verify the abstract contract directly via 12 GTest cases (detection / shared assets / EP compatibility / variant selection / overlay extraction).

This is part of an incremental v4 package integration; see also #2125 (W1) and #2126 (W4-prep) and #2131 (W5c-prep).

### DIFF
--- a/src/models/model_package.cpp
+++ b/src/models/model_package.cpp
@@ -1,0 +1,715 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#include "../json.h"
+#include "model_package.h"
+
+#include <algorithm>
+#include <cstddef>
+#include <filesystem>
+#include <fstream>
+#include <sstream>
+#include <system_error>
+#include <unordered_set>
+
+namespace Generators {
+
+namespace {
+
+// ---- fs::path <-> std::filesystem::path bridge -------------------------------
+//
+// The codebase's `fs::path` (src/filesystem.h) is the public surface; it
+// treats narrow strings as UTF-8 and uses `MultiByteToWideChar(CP_UTF8, ...)`
+// on Windows. `std::filesystem::path`'s narrow ctor on Windows interprets
+// its input as the *active code page*. To keep encoding correct for non-ASCII
+// paths we must round-trip through the wide form on Windows; on Unix the
+// native narrow encoding IS UTF-8 and the simple ctor is correct.
+
+std::filesystem::path ToStd(const fs::path& p) {
+#ifdef _WIN32
+  return std::filesystem::path(p.c_str());
+#else
+  return std::filesystem::path(p.string());
+#endif
+}
+
+fs::path FromStd(const std::filesystem::path& p) {
+#ifdef _WIN32
+  const std::wstring w = p.wstring();
+  if (w.empty()) return fs::path{};
+  const int needed = WideCharToMultiByte(
+      CP_UTF8, 0, w.data(), static_cast<int>(w.size()),
+      nullptr, 0, nullptr, nullptr);
+  std::string utf8(static_cast<std::size_t>(needed), '\0');
+  WideCharToMultiByte(CP_UTF8, 0, w.data(), static_cast<int>(w.size()),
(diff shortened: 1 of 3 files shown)


xiaoyu-work:
Realigned the stub to the updated v4 model-package spec ([authoritative version saved locally](https://github.com/microsoft/onnxruntime-genai/pull/2132)). Schema deltas vs the prior version of this PR:

- `manifest.json` `components` is now a string array (was object array). Whole file is optional; absent file means scan subdirs.
- Per-component selection metadata moved to a separate `<component>/metadata.json` (with shape `variants.<name>.ep_compatibility[]`); `variant.json` now carries `files[]` (with per-file `session_options`, `provider_options`, `shared_files`) and `consumer_metadata`.
- New `<component>/shared_weights/<checksum>/<blob>` directory layout with `ComponentInstance::ResolveSharedWeight(checksum)`.
- EP-compatibility entries gain optional `device` (OpenVINO `GPU`/`NPU`) and `compatibility` strings.
- `SelectComponent` takes a `ModelPackageSelectionOptions{vector<EpSelection>}` (ordered EP-priority list with optional device) instead of a single EP-name string. Implements the spec's captured-EP-priority + metadata-declaration-order tie-break.
- `ConsumerMetadata()` returns the raw blob verbatim (was: pre-extracted `genai_config_overlay`). GenAI parses it itself in W3.

Per critique from rubber-duck review:
- `Files()` removed from the abstract surface — consumers parse `variant.json` directly via the standalone `ParseVariantManifest` helper, matching the v4 C API which does not expose per-file accessors via `cix`.
- Path-traversal validation applied to every package-controlled fragment (component name, variant name, file filename, shared-weight checksum).
- Detection requires `manifest.json` OR a direct child with `metadata.json`; the bare presence of `configs/` is no longer a positive v4 signal (would over-classify some flat-dir layouts).
- Variant declaration order is sourced from `metadata.json` (preserved via streaming `JSON::Element` parser, since `JSON::Object` = `std::map` doesn't preserve insertion order). Filesystem order is never authoritative.

xiaoyu-work:
Folded into the umbrella draft PR #2136. Closing to keep a single review surface.

PR #2133, opened 2026-05-06
Cohere Transcribe Support

nenad1002:
This PR introduces the Cohere Transcribe model alongside support for long-form transcription on both CUDA and CPU backends.

Although the model was trained with long-form transcription in mind, the supported sequence length remains limited. Additionally, the current export uses a fixed 30-second input window, similar to Whisper.

For longer audio, we utilize Silero VAD to split speech into smaller segments based on silence boundaries before transcription. The current model export also does not support word-level timestamps, although this is not required for the intended Foundry Local integration scenarios, as Silero VAD already provides sufficiently accurate segmentation for long-form transcription. Future work will focus on improving VAD robustness and segmentation quality in noisy acoustic environments.

While this results in exceptional WER quality accross standard benchmarks, punctuation and capitalization can sometimes become inconsistent across chunk boundaries due to limited cross-chunk context in the current model. We are looking to work with Cohere to improve this behavior.


src/models/cohere_processor.cpp
```diff
@@ -0,0 +1,346 @@
...
+
+  const size_t window = static_cast<size_t>(vad_->GetWindowSize());
+  const float threshold = vad_->GetThreshold();
+
+  std::vector<uint8_t> is_speech;
+  is_speech.reserve(num_samples / window);
+  for (size_t off = 0; off + window <= num_samples; off += window) {
+    float p = vad_->ProcessWindow(samples + off, window);
+    is_speech.push_back(p >= threshold ? 1u : 0u);
+  }
```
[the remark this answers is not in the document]
nenad1002: low value comment, at most 32ms can be skipped, no real world issue.

examples/python/cohere_transcribe.py
```diff
@@ -0,0 +1,137 @@
+"""
+Audio chunking is handled internally by CohereProcessor + Generator.
```
kunal-vaishnavi, line 2: Rather than creating new example files for each speech model, let's find a way to support all of them in a more generic and scalable way. Otherwise, we will need to keep creating, maintaining, and validating each model's example in each language.

src/models/cohere_model.cpp
```diff
@@ -0,0 +1,329 @@
...
+                                            shape_info->GetElementType());
+  }
+}
+
+void CohereEncoderState::SetExtraInputs(const std::vector<ExtraInput>& extra_inputs) {
+  const std::string& audio_name = model_.config_->model.encoder.inputs.audio_features;
+  for (const auto& [name, value] : extra_inputs) {
+    if (name == audio_name) {
+      SetOrReplaceInput(audio_name.c_str(), value->ort_tensor_, audio_features_);
+    } else if (name == "mel_length") {
```
kunal-vaishnavi, line 75: If this is an extra named input that comes from the preprocessor, can we compare its name to the one set in the GenAI config?

src/models/cohere_model.cpp
```diff
@@ -0,0 +1,329 @@
...
+}
+
+void CohereState::SetExtraInputs(const std::vector<ExtraInput>& extra_inputs) {
+  // Extract and store chunk data before passing to encoder
+  std::vector<ExtraInput> encoder_inputs;
+  std::map<int, std::shared_ptr<Tensor>> indexed_mels;
+  std::map<int, std::shared_ptr<Tensor>> indexed_mel_lengths;
+
+  for (const auto& [name, tensor] : extra_inputs) {
+    if (name == "cohere_chunk_count") {
```
kunal-vaishnavi, line 138: We should remove the `cohere` prefixes in these names. Same comment as [here](https://github.com/microsoft/onnxruntime-genai/pull/2133/changes#r3258021330) in addition.

src/models/cohere_model.cpp
```diff
@@ -0,0 +1,329 @@
...
+namespace {
+
+// Tokens with id < this are byte-fallback for C0 control bytes; strip them
+// at chunk seams before deciding whether to add a connecting space.
+constexpr int32_t kSpecialTokenIdMax = 16;
+
+// Returns true if the rendered text of `token_id` does not begin with
+// whitespace, i.e. a connecting space must be injected before it at a
+// chunk seam to avoid "scamsIn" / "thisInto" artifacts.
+bool TokenLacksLeadingSpace(int32_t token_id, const Tokenizer& tokenizer) {
```
kunal-vaishnavi, line 240: We could get/set a logits mask during the generation loop using the `get_logits`/`set_logits` APIs so that we can customize this logic as needed for future models (e.g. fine-tuned variants that don't have this issue). It also has the extra benefit that we don't have to hardcode this logic beforehand.

src/models/silero_vad.h
```diff
@@ -67,22 +67,6 @@ struct SileroVad {
   std::vector<float> input_data_;
 };
 
-/// Internal State for SileroVad — uses State::Run for proper EP/run-options support.
```
kunal-vaishnavi, line 70: Why move from header file to implementation file?

src/config.cpp
```diff
@@ -1134,6 +1138,14 @@ struct Model_Element : JSON::Element {
       v_.blank_id = static_cast<int>(JSON::Get<double>(value));
     } else if (name == "max_symbols_per_step") {
       v_.max_symbols_per_step = static_cast<int>(JSON::Get<double>(value));
+    } else if (name == "cohere_vad_min_silence_ms") {
```
kunal-vaishnavi, line 1141: Can we remove the `cohere_vad` prefix for these names and re-use any existing names where possible?

src/generators.cpp
```diff
@@ -369,6 +370,8 @@ Generator::Generator(const Model& model, const GeneratorParams& params) : model_
   state_ = model.CreateState(search_->GetSequenceLengths(), params);    // Search sequence lengths set when creating state
   guidance_logits_processor_ = CreateGuidanceLogitsProcessor(*state_);  // Could be nullptr if use_guidance (constrained decoding) is not used
 
+  is_cohere_model_ = model.config_->model.type == "cohere_transcribe";
```
kunal-vaishnavi, line 373: There is a lot of "if cohere, do X, else do Y" logic that is being added in this file. This pattern was already done with Nemotron speech, but ideally this shouldn't be happening. This file is meant to be generic for all model types and not have model-specific logic as much as possible. Let's find another way to add these changes or introduce new higher-level internal methods for better abstraction.

PR #2134, opened 2026-05-06
[v4-pkg W9] Pin layer-2 config-overlay contract with regression tests

xiaoyu-work:
Reopens the W9 work that was parked when #2130 was closed; reopened as a standalone draft because it can land independently of the v4 wiring (it pins existing behavior).

## What

Adds 6 GTest cases under `test/runtime_settings_test.cpp` covering the Layer-2 config-overlay channel: `RuntimeSettings::GenerateConfigOverlay()` + `OverlayConfig` over `model.decoder.session_options.provider_options[*]`.

The tests pin the contract that the v4 work (W3 / W5c-real) will rely on:
- `GenerateConfigOverlay()` projects `dawnProcTable` to the canonical JSON path under `model.decoder.session_options.provider_options[*].WebGPU.dawnProcTable`.
- `OverlayConfig` merges by EP **name** when `provider_options` already contains an entry for the same EP, instead of replacing the whole array — important so a partial overlay doesn't wipe other `provider_options`.
- `ProviderOptionsArray_Element::OnComplete` normalizes EP names (`qnn` → `QNN`, `webgpu` → `WebGPU`) **after** the overlay merge — tests look for the normalized form.

## Why standalone

These tests don't introduce any production code changes; they exercise existing public APIs (`OgaConfigOverlay` etc.) and lock current behavior. Keeping them on a sibling PR avoids forcing the W3 reviewer to look at unrelated test infra.

Part of the v4 model-package integration; see #2125 (W1) / #2126 (W4-prep) / #2131 (W5c-prep) / #2132 (W2-stub).

### DIFF
--- a/src/config.h
+++ b/src/config.h
@@ -430,6 +430,25 @@ void SetSearchNumber(Config::Search& search, std::string_view name, double value
 void SetSearchBool(Config::Search& search, std::string_view name, bool value);
 void ClearProviders(Config& config);
 void SetProviderOption(Config& config, std::string_view provider_name, std::string_view option_name, std::string_view option_value);
+// Layer-2 override channel (caller-driven).
+//
+// Applies a JSON document to an existing Config in place, using the same
+// JSON parser registered for genai_config.json. Any field that can be set via
+// genai_config.json can therefore also be overridden here. The JSON parser
+// merges objects key-wise (ProviderOptionsObject_Element matches by `name`),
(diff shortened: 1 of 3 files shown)


xiaoyu-work:
Folding into the upcoming v4-package integration draft PR — comments-and-tests-only, doesn't warrant a standalone PR.

PR #2135, opened 2026-05-06
Fix Gemma4 pixel_values trimming

apsonawane:
### Summary

Fix NaN in vision encoder caused by incorrect dtype handling in the pixel_values trimming path of `Gemma4MultimodalProcessor`.

### Problem

The image processor (`Gemma4ImageTransform`) outputs `pixel_values` as **float32**, but the trimming code used `pixel_values_type_` (float16) for `elem_size` and stride calculations. This caused `std::memcpy` to copy 2 bytes per element from a 4-byte float32 buffer — reinterpreting float32 bit patterns as float16, producing NaN/garbage.

**NaN propagation chain:**
```
Image processor (float32) → raw memcpy with fp16 strides → garbage fp16 pixel_values
→ vision encoder outputs NaN → NaN embeddings → NaN through all decoder layers
→ NaN logits → CUDA crash in topk
```

### Fix

| Change | Description |
|--------|-------------|
| Trimming path | Use `sizeof(float)` for strides; create `OrtValue::CreateTensor<float>` instead of `OrtValue::CreateTensor(allocator, shape, pixel_values_type_)` |
| Non-trimming path | Pass `ONNX_TENSOR_ELEMENT_DATA_TYPE_FLOAT` to `EmplaceProcessedTensor` instead of `pixel_values_type_` |

The vision encoder ONNX model now accepts float32 input and casts to model dtype (fp16) internally via a mobius-side change. This eliminates all host-side dtype conversion for pixel_values.

### Testing

- Multimodal image+text generation (CPU): clean output ✅
- Multimodal image+text generation (CUDA): clean output, no NaN, no crashes ✅
- Text-only generation: unaffected ✅

### DIFF
--- a/src/models/gemma4_multimodal_processor.cpp
+++ b/src/models/gemma4_multimodal_processor.cpp
@@ -302,11 +302,26 @@ std::unique_ptr<NamedTensors> Gemma4MultiModalProcessor::Process(const Tokenizer
     constexpr int64_t kPoolingKernelSize = 3;
     const int64_t actual_patches = static_cast<int64_t>(actual_soft_tokens) * kPoolingKernelSize * kPoolingKernelSize;
 
-    // Get padded pixel_values shape
+    // Get padded pixel_values shape.
+    // The image processor is expected to produce float32 data; the vision encoder
+    // ONNX model accepts float32 and casts to model dtype internally.
     const float* pv_data{};
     const int64_t* pv_shape{};
     size_t pv_dims;
     CheckResult(OrtxGetTensorData(pixel_values, reinterpret_cast<const void**>(&pv_data), &pv_shape, &pv_dims));
+
+    // Verify the image processor produced float32. The trimming and passthrough
+    // paths below assume sizeof(float) strides. Fail fast if the assumption breaks.
+    {
+      extDataType_t pv_elem_type{};
+      CheckResult(OrtxGetTensorType(pixel_values, &pv_elem_type));
+      if (pv_elem_type != kOrtxFloat) {
+        throw std::runtime_error(
+            "pixel_values from the image processor must be float32 (got element type " +
+            std::to_string(static_cast<int>(pv_elem_type)) +
+            "). The vision encoder ONNX model handles dtype conversion internally.");
+      }
+    }
 
     // Determine the patches dimension and patch_dim based on tensor rank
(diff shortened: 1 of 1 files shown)


justinchuby (approved):

src/models/gemma4_multimodal_processor.cpp
```diff
@@ -320,23 +335,22 @@ std::unique_ptr<NamedTensors> Gemma4MultiModalProcessor::Process(const Tokenizer
...
-        std::memcpy(dst + b * dst_stride, src + b * src_stride, dst_stride);
+        std::memcpy(dst + b * dst_row, src + b * src_row, dst_row * sizeof(float));
       }
 
       named_tensors->emplace(std::string(Config::Defaults::PixelValuesName),
                              std::make_shared<Tensor>(std::move(trimmed_pv)));
     } else {
-      EmplaceProcessedTensor(*named_tensors, Config::Defaults::PixelValuesName, pixel_values, pixel_values_type_, allocator);
+      // Image processor returns float32; the vision encoder ONNX model
+      // accepts float32 and casts to model dtype internally.
```
kunal-vaishnavi, line 352: Why are we casting within the ONNX model? We should return the right dtype from all of this preprocessing logic.

justinchuby: I think it helps to pin down the contract for this case and make it consistent across models. Whichever is preferred.

kunal-vaishnavi: The returned dtypes from pre-processing should match the expected dtypes for the model's inputs.

justinchuby: @apsonawane could you confirm this is consistent with all architectures we support in genai? With this assumption I will document the contract and update all vision and audio components to take specified dtypes. Though - what is the expected input type for quantized models? float16? bfloat16? float32?

justinchuby:
Any updates to this PR? @apsonawane Either way since genai will detect and cast inputs appropriately for the models, changes to existing building logic is not blocking

src/models/gemma4_multimodal_processor.cpp
justinchuby: https://github.com/onnxruntime/mobius/pull/319

apsonawane:
Closing this PR since we opened this: https://github.com/microsoft/onnxruntime-genai/pull/2188

PR #2137, opened 2026-05-07
[Qwen3] Allow packed QKV MatMul under QK-Norm via post-MatMul Split

xiaofeihan1:
Previously `use_packed_matmul` was disabled whenever q_norm or k_norm was set, so Qwen3 (and any other QK-Norm architecture) emitted three separate q_proj / k_proj / v_proj MatMulNBits nodes per layer.

This PR allows packed QKV MatMul for QK-Norm models by inserting a single ONNX `Split` after the packed projection so the existing per-head Q/K SimplifiedLayerNormalization path is unchanged. Math is equivalent; quantization is unchanged.

### Graph change (per attention layer)

Before:
```
root -> q_proj/MatMul -> Q
root -> k_proj/MatMul -> K
root -> v_proj/MatMul -> V
```
After (Split placed *after* the optional packed bias-Add so packed bias fusion is preserved):
```
root -> qkv_proj/MatMul -> [qkv_proj/Add] -> Split -> Q, K, V
```

Net per layer: **−2 MatMulNBits, +1 Split**. For Qwen3-4B (36 layers): −72 MatMulNBits, +36 Split, total ONNX nodes 700 → 665.

### Why Split (not Slice)

Single read of the packed tensor, single dispatch with 3 outputs — avoids re-reading the same packed Q/K/V tensor three times per decode step.

### Verification — Qwen3-4B int4 (RelWithDebInfo build, accuracy_level=4)

Compared two builds of the same source model:
- **base** = pre-PR builder (reset to `664a61b1`, QK-Norm forces 3 independent MatMuls)
- **curr** = this PR

Output token sequences are byte-identical on both GPUs (no accuracy drift).

#### WebGPU GPU profile (aggregate: 1 prefill + 50 decode × 2 phases)

| OpType | base ms | curr ms | Δ ms |
|---|---:|---:|---:|
| MatMulNBits | 932.6 | 876.9 | **−55.7** |
| Split       | 0     | 10.8  | +10.8 |
| GroupQueryAttention | 440.2 | 442.5 | +2.3 |
| **TOTAL GPU** | **1509.0** | **1466.4** | **−42.6 (−2.8%)** |

- Decode path (M=1): A 3× independent GEMV = 96.8 ms → B packed GEMV + Split = 64.5 ms → **decode net −32 ms**.
- Prefill (M=1000): A 3× DP4A = 71.3 ms → B packed DP4A = 62.6 ms → prefill net −8.7 ms.

#### End-to-end perf — Qwen3-4B prefill-1000, 5 runs

| GPU | metric | base | curr | Δ |
|---|---|---:|---:|---:|
| NVIDIA RTX 5080 (steady P0) | gen tps | 105.4 | 108.5 | **+2.9%** |
| Intel(R) Graphics iGPU | gen tps | 12.44 | 13.06 | **+5.0%** |

(NV result matches the −2.8% GPU profile measurement; iGPU receives a similar relative gain. Prompt TPS and TTFT are unchanged on both vendors.)

### Compatibility

- Models without QK-Norm: behavior unchanged (the new Split block is gated on `q_norm && k_norm`).
- LoRA / packed-Attention path (`use_matmul_in_attn`): unchanged.
- Packed bias fusion still applies for QK-Norm models when bias exists, because Split is placed after the packed Add.

### DIFF
--- a/src/python/py/models/builders/base.py
+++ b/src/python/py/models/builders/base.py
@@ -518,6 +518,9 @@ class Model:
         return (self.ep, self.io_dtype) in valid_packed_attn_configurations
 
     def make_attention_init(self):
+        self.q_size = self.num_attn_heads * self.head_size
+        self.kv_size = self.num_kv_heads * self.head_size
+
         if self.is_gqa_supported():
             # Change model settings for GroupQueryAttention
             self.attention_attrs["op_type"] = "GroupQueryAttention"
@@ -1059,12 +1062,10 @@ class Model:
         self.make_node("Slice", inputs=inputs, outputs=[output], name=name)
         self.make_value(output, dtype, shape=shape)
 
-    def make_split(self, name, inputs, axis, output_shapes, dtype):
-        outputs = [f"{name}/output_{i}" for i in range(len(output_shapes))]
+    def make_split(self, name, inputs, outputs, dtypes, shapes, axis=-1):
         self.make_node("Split", inputs=inputs, outputs=outputs, name=name, axis=axis)
-        for out, shape in zip(outputs, output_shapes):
-            self.make_value(out, dtype, shape=shape)
-        return outputs
+        for out, dt, shape in zip(outputs, dtypes, shapes):
+            self.make_value(out, dt, shape=shape)
 
     def make_mul(self, name, inputs, dtype, shape):
         output = f"{name}/output_0"
@@ -2373,13 +2374,13 @@ class Model:
         q_reshape_2_name = f"/model/layers.{layer_id}/attn/q_norm/Reshape_2"
         q_reshape_2_inputs = [
             q_layernorm_output,
-            f"/model/constants/INT64/[0, -1, {self.num_attn_heads * self.head_size}]",
+            f"/model/constants/INT64/[0, -1, {self.q_size}]",
         ]
         self.make_reshape(
             q_reshape_2_name,
             q_reshape_2_inputs,
             dtype=self.io_dtype,
-            shape=["batch_size", "sequence_length", self.num_attn_heads * self.head_size],
+            shape=["batch_size", "sequence_length", self.q_size],
         )
 
         # Reshape K MatMul from BxSxD to Bx(SxN)xH before LayerNorm
@@ -2426,13 +2427,13 @@ class Model:
         k_reshape_2_name = f"/model/layers.{layer_id}/attn/k_norm/Reshape_2"
         k_reshape_2_inputs = [
             k_layernorm_output,
-            f"/model/constants/INT64/[0, -1, {self.num_kv_heads * self.head_size}]",
+            f"/model/constants/INT64/[0, -1, {self.kv_size}]",
         ]
         self.make_reshape(
             k_reshape_2_name,
             k_reshape_2_inputs,
             dtype=self.io_dtype,
-            shape=["batch_size", "sequence_length", self.num_kv_heads * self.head_size],
+            shape=["batch_size", "sequence_length", self.kv_size],
         )
 
         # Update q_path and k_path now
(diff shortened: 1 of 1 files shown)


src/python/py/models/builders/base.py
```diff
@@ -1061,6 +1059,13 @@ def make_slice(self, name, inputs, dtype, shape):
         self.make_node("Slice", inputs=inputs, outputs=[output], name=name)
         self.make_value(output, dtype, shape=shape)
 
+    def make_split(self, name, inputs, axis, output_shapes, dtype):
```
kunal-vaishnavi: To be consistent with the other `make_op_name` methods, can we order the parameters as follows (with the understanding that there are multiple outputs, dtypes, shapes, etc unlike the other `make_op_name` methods)?

```suggestion
    def make_split(self, name, inputs, outputs, dtypes, shapes, axis=default_value_here):
```

src/python/py/models/builders/base.py
```diff
@@ -2952,6 +2957,30 @@ def make_attention_input_proj(self, layer_id, attention, root_input, **kwargs):
                     attention.q_proj, attention.k_proj, attention.v_proj, qkv_matmul_basename, root_input
                 )
                 self.attention_attrs["q_path"] = f"{qkv_matmul_name}/output_0"
+
+                # When q_norm/k_norm are present, the packed-QKV path inside GQA cannot be used
+                # (norm runs per-head before attention). Split here so downstream sees Q/K/V separately.
+                if self.attention_attrs["q_norm"] and self.attention_attrs["k_norm"]:
+                    q_size = self.num_attn_heads * self.head_size
```
kunal-vaishnavi: I have noticed there are a lot of times we calculate `q_size` and `kv_size` in this file. Can we refactor its calculation on a per-layer basis in some way (e.g. calculate `q_size` and `kv_size` at the beginning of each attention layer's creation)?

src/python/py/models/builders/base.py
```diff
@@ -2980,7 +3009,12 @@ def make_attention_input_proj(self, layer_id, attention, root_input, **kwargs):
 
         else:
             # Make Add nodes (if bias exists)
-            if self.attention_attrs["use_packed_matmul"] and qkv_dtype_equal and any_bias_exists:
+            if (
+                self.attention_attrs["use_packed_matmul"]
```
kunal-vaishnavi: If the condition for `use_packed_matmul` has changed, then why do we need the latter two predicates for `q_norm` and `k_norm`?

src/python/py/models/builders/base.py
xiaofeihan1: Done — signature updated to `make_split(self, name, inputs, outputs, dtypes, shapes, axis=-1)` to match the other `make_op_name` helpers, with `axis` defaulted last. The call site now passes `outputs`, `dtypes`, and `shapes` lists explicitly.

src/python/py/models/builders/base.py
xiaofeihan1: Refactored — `self.q_size` and `self.kv_size` are now initialized once in `make_attention_init` and reused everywhere downstream. Replaced raw `self.num_attn_heads * self.head_size` / `self.num_kv_heads * self.head_size` computations in `make_qk_norm`, `make_repeat_kv`, `make_attention_unpacked_lora`, and `make_attention_unpacked_regular`. The only remaining raw computation is in `load_weights`, which runs before `make_attention_init`; I left it as-is to avoid introducing an initialization-order coupling.

src/python/py/models/builders/base.py
xiaofeihan1: Good catch — the extra `q_norm` / `k_norm` predicates were a symptom of placing the Split right after the packed MatMul, which forced this branch to skip the packed bias-Add for QK-Norm models. I moved the Split to *after* the (optional) packed Add at the end of `make_attention_input_proj`. This preserves the packed bias-Add fusion unchanged for QK-Norm models, so the fusion condition reverts to its original form. The extra predicates are gone.

kunal-vaishnavi (approved):

PR #2139, opened 2026-05-07
Enable Qwen3.5 TRT-RTX EP path with CUDA graph

yen-shi:
This PR enables Qwen3.5 text-only INT4 QDQ export and TRT-RTX EP inference with CUDA graph/shared past-present buffers.

## Structure

The branch is rebased on latest `main` and intentionally split into two commits:

1. `Add Qwen3.5 text-only export support`
   - Mirrors the overlapping Qwen3.5 text-only builder/model-type work from PR #2157.
   - Keeps Qwen3.5 text-only ONNX input `position_ids` as `[B, S]` and expands inside the graph to `[3, B, S]` for mRoPE.
2. `Enable Qwen3.5 TRT-RTX shared-buffer inference`
   - Shares Qwen3.5 recurrent/conv state buffers when `past_present_share_buffer` is enabled, preserving stable input/output addresses for TRT-RTX graph replay.
   - Fixes QDQ SkipLayerNorm `output_3` producer wiring.
   - Keeps mixed-precision quantization logic in the shared base `k_quant_linear` path rather than Qwen-local code.
   - Adds the canonical `NvTensorRtRtx` name to the example EP choices.

## PR #2157 compatibility

This branch was compared against microsoft/onnxruntime-genai#2157 using `git merge-tree`. The same Qwen files are touched, but Git auto-merges them cleanly and the simulation produced no conflict markers.

If #2157 merges first, the first commit in this branch is the overlap and can be dropped/rebased away; the second commit contains the TRT-RTX-specific delta.

## Validation

- Rebased onto latest upstream `main` (`bf6cf3fe`).
- Built CUDA Release wheel with CUDA 13.2:
  `python build.py --use_cuda --cuda_home="C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v13.2" --config Release --update --build --parallel --skip_tests --skip_examples`
- Installed the rebuilt wheel in the minimal TRT-RTX package environment.
- Exported and ran Qwen3.5 0.8B and 9B text-only INT4 QDQ models with TRT-RTX EP + CUDA graph enabled.
  - 0.8B: TTFT 1.18s, decode 64.44 tok/s, answer starts: "The history of artificial intelligence..."
  - 9B: TTFT 1.47s, decode 48.00 tok/s, answer is a coherent reasoning-style response.


yen-shi:
@microsoft-github-policy-service agree company="NVIDIA Corporation"

src/models/position_inputs.cpp
```diff
@@ -595,6 +595,10 @@ Qwen2VLPositionInputs::Qwen2VLPositionInputs(const Model& model, State& state, D
 
   if (has_posid_input_) {
     ONNXTensorElementDataType posid_type = model_.session_info_.GetInputDataType(model_.config_->model.decoder.inputs.position_ids);
+    if (has_mask_input_ && posid_type != type_) {
+      throw std::runtime_error("position_ids & attention_mask must have the same data type");
+    }
+    type_ = posid_type;
```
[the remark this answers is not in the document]
yen-shi: Done in 1a7742cd: split into independent `mask_type` / `posid_type` queries, validate, and assign `type_` exactly once.

src/models/position_inputs.cpp
```diff
@@ -857,30 +863,122 @@ void Qwen2VLPositionInputs::CreateAndInitializeAttentionMask(DeviceSpan<int32_t>
...
+  auto num_beams = state_.params_->search.num_beams;
+  auto max_length = attention_mask_shape_[1];
+
+  for (int i = 0; i < batch_size; i++) {
+    for (int j = 0; j < num_beams; j++) {
+      auto output_subspan = output_span.subspan((i * num_beams + j) * max_length, prompt_length);
+      auto input_subspan = input_span.subspan(i * prompt_length, prompt_length);
+      output_subspan.CopyFrom(input_subspan);
+    }
+  }
```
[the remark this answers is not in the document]
yen-shi: Done in 1a7742cd: replaced the per-(batch, beam) `CopyFrom` loop with one CPU-mirror fill + a single `CopyCpuToDevice`.

src/models/position_inputs.cpp
```diff
@@ -921,6 +1038,31 @@ void Qwen2VLPositionInputs::RewindTo(size_t index) {
...
+          std::fill_n(data + i * max_len, index, static_cast<int32_t>(1));
+          std::fill_n(data + i * max_len + index, max_len - index, static_cast<int32_t>(0));
+        }
+      } else {
+        auto* data = reinterpret_cast<int64_t*>(cpu_data.data());
+        for (size_t i = 0; i < batch_beam_size; i++) {
+          std::fill_n(data + i * max_len, index, static_cast<int64_t>(1));
+          std::fill_n(data + i * max_len + index, max_len - index, static_cast<int64_t>(0));
+        }
+      }
```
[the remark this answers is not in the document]
yen-shi: Done in 1a7742cd: switched to `Tensor::GetDeviceSpan<T>().CpuSpan()` so the writes go through ORT's properly-aligned `GetTensorMutableData<T>` pointer; no more `reinterpret_cast` from a byte span.

src/models/position_inputs.cpp
```diff
@@ -889,6 +987,23 @@ void Qwen2VLPositionInputs::UpdateAttentionMask() {
...
+          model_.p_device_->GetType() == DeviceType::NvTensorRtRtx);
+}
+
+bool Qwen2VLPositionInputs::ShouldUseStaticPositionIDHandling() const {
+  // Keep this predicate separate from mask handling in case a future model
+  // supports static masks but still needs dynamic mRoPE position IDs, or vice versa.
+  return state_.params_->use_graph_capture ||
+         (state_.params_->IsPastPresentShareBufferEnabled(model_.config_->model.type) &&
+          model_.p_device_->GetType() == DeviceType::NvTensorRtRtx);
+}
```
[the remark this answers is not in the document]
yen-shi: Done in 1a7742cd: factored the shared logic into a new private `ShouldUseStaticInputsForGraphReplay()` helper. Both entry points are kept as semantic call-site distinctions and delegate to it.

src/models/position_inputs.cpp
```diff
@@ -857,30 +863,122 @@ void Qwen2VLPositionInputs::CreateAndInitializeAttentionMask(DeviceSpan<int32_t>
...
+
+template <typename T>
+void Qwen2VLPositionInputs::Update3DPositionIDsInPlace(int base_pos) {
+  int64_t batch_size = position_ids_shape_[1];  // This is already expanded (batch*beams)
+  int64_t seq_len = position_ids_shape_[2];     // This will be 1 for generation
+
+  auto position_ids = OrtValue::CreateTensor(model_.allocator_cpu_, position_ids_shape_, type_);
+  UpdatePositionIdsFunctor{position_ids.get(), base_pos, batch_size, seq_len, rope_deltas_}.template operator()<T>();
+
+  auto src_span = WrapTensor<T>(*GetDeviceInterface(DeviceType::CPU), *position_ids);
```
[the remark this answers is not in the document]
yen-shi: Done in 1a7742cd: removed the per-decode CPU `OrtValue` allocation. The mRoPE position IDs are now written straight into the persistent device tensor's CPU mirror and pushed with a single `CopyCpuToDevice` per step.

anskumar01:
@kunal-vaishnavi , @baijumeswani , can you please help review.
Cc @anujj

src/models/model.cpp
```diff
@@ -76,6 +76,19 @@ class DirGuard {
   }
 };
 
+bool IsTextOnlyQwen3_5(const Config& config) {
```
kunal-vaishnavi: There is a [similar PR](https://github.com/microsoft/onnxruntime-genai/pull/2157) for enabling a text-only Qwen-3.5 model. Let's align the two PRs so they are on the same page.

src/models/multi_modal.cpp
```diff
@@ -650,8 +651,12 @@ DecoderState::DecoderState(const MultiModalLanguageModel& model, DeviceSpan<int3
       recurrent_state_{CreateRecurrentState(*this)} {
   inputs_embeds_.Add();
 
-  // Some multimodal decoders (e.g., Gemma4) require input_ids alongside inputs_embeds
-  if (model_.session_info_.HasInput(model_.config_->model.decoder.inputs.input_ids)) {
+  // Some multimodal decoders (e.g., Gemma4) require input_ids alongside inputs_embeds.
```
kunal-vaishnavi, line 653: There was a [PR fix](https://github.com/microsoft/onnxruntime-genai/pull/2148) for a regression introduced by these changes. Let's align the two PRs so we can agree on the same fix.

src/models/position_inputs.cpp
```diff
@@ -626,13 +626,23 @@ Qwen2VLPositionInputs::Qwen2VLPositionInputs(const Model& model, State& state, D
...
+  // Both inputs must agree when both are present; we query each one and
+  // assign type_ exactly once after validating they match. Default is int64.
+  ONNXTensorElementDataType mask_type = Ort::TypeToTensorType<int64_t>;
   if (has_mask_input_) {
-    type_ = model_.session_info_.GetInputDataType(model_.config_->model.decoder.inputs.attention_mask);
+    mask_type = model_.session_info_.GetInputDataType(model_.config_->model.decoder.inputs.attention_mask);
   }
+  ONNXTensorElementDataType posid_type = mask_type;
+  if (has_posid_input_) {
+    posid_type = model_.session_info_.GetInputDataType(model_.config_->model.decoder.inputs.position_ids);
```
kunal-vaishnavi: Why are these changes for the position ids type needed?

src/python/py/models/builders/qwen.py
```diff
@@ -1007,6 +1008,32 @@ def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
         # Disable fused RoPE in attention op - we apply mRoPE manually
         self.attention_attrs["use_rope_in_attn"] = False
 
+        # Mixed-precision quantization for linear attention layers.
+        # Baseline: whole model INT4. Override linear attention layer nodes
+        # to INT8 for better accuracy with modest size increase.
+        #
+        # Linear attention recurrence accumulates errors across the full sequence,
+        # unlike softmax attention which normalizes per-step.
+        int8_nodes = {}
```
kunal-vaishnavi: There is an algorithm called `k_quant_linear` that was added for mixed-precision quantization. This logic was moved in [this PR](https://github.com/microsoft/onnxruntime-genai/pull/2100) to the base class for re-usability. Can you sync with the main branch and use that instead?

src/models/position_inputs.cpp
```diff
@@ -641,6 +651,7 @@ Qwen2VLPositionInputs::Qwen2VLPositionInputs(const Model& model, State& state, D
     position_ids_shape_[2] = 0;  // Will be set during first update
 
     position_ids_ = std::make_unique<Tensor>(model_.p_device_inputs_, posid_type);
+    position_ids_next_ = std::make_unique<Tensor>(model_.p_device_inputs_, posid_type);
```
kunal-vaishnavi: We shouldn't have both `position_ids_` and `position_ids_next_` for representing the position ids during different steps of the decoding loop. We should find a way to re-use `position_ids_`.

src/models/position_inputs.cpp
```diff
@@ -895,30 +907,140 @@ void Qwen2VLPositionInputs::CreateAndInitializeAttentionMask(DeviceSpan<int32_t>
...
+    return;
+  }
+
   attention_mask_->ort_tensor_ = model_.ExpandInputs(attention_mask, state_.params_->search.num_beams);
   attention_mask_shape_[0] *= state_.params_->search.num_beams;
   state_.inputs_[mask_input_index_] = attention_mask_->GetOrtTensor();
 }
 
+template <typename T>
+void Qwen2VLPositionInputs::InitializeStaticMask(OrtValue& cpu_attention_mask) {
```
kunal-vaishnavi: Can we move all of the logic for Qwen-specific position inputs inside the Qwen-specific files? This file should ideally just have logic that is shared across all types of models.

src/models/position_inputs.cpp
```diff
@@ -927,6 +1049,28 @@ void Qwen2VLPositionInputs::UpdateAttentionMask() {
   state_.inputs_[mask_input_index_] = attention_mask_->GetOrtTensor();
 }
 
+bool Qwen2VLPositionInputs::ShouldUseStaticInputsForGraphReplay() const {
+  // TRT-RTX graph reuse has the same fixed-address/fixed-shape requirement as
+  // CUDA graph capture, even when graph capture is requested through shared
+  // past/present buffers rather than the generic CUDA EP path.
+  return state_.params_->use_graph_capture ||
+         (state_.params_->IsPastPresentShareBufferEnabled(model_.config_->model.type) &&
+          model_.p_device_->GetType() == DeviceType::NvTensorRtRtx);
```
kunal-vaishnavi: Why do we need to specify the device is TRT-RTX? Multiple EPs support graph capture and will require static inputs.

src/models/position_inputs.cpp
```diff
@@ -959,6 +1105,34 @@ void Qwen2VLPositionInputs::RewindTo(size_t index) {
...
+        using T = decltype(type_tag);
+        auto typed_span = attention_mask_->GetDeviceSpan<T>();
+        auto cpu = typed_span.CpuSpan();
+        for (size_t i = 0; i < batch_beam_size; i++) {
+          std::fill_n(cpu.data() + i * max_len, index, static_cast<T>(1));
+          std::fill_n(cpu.data() + i * max_len + index, max_len - index, static_cast<T>(0));
+        }
+        typed_span.CopyCpuToDevice();
+      };
+      if (type_ == Ort::TypeToTensorType<int32_t>) {
```
apsonawane: Can we use DispatchOnType here, to keep it consistent with other code

yen-shi:
Thanks for the reviews. I rewrote/rebased the branch and addressed the outstanding threads in a smaller two-commit form:

- The first commit is the Qwen3.5 text-only export overlap with #2157.
- The second commit is only the TRT-RTX-specific delta needed for Qwen3.5 shared-buffer/CUDA-graph inference.
- The previous Qwen-specific runtime edits in `position_inputs.cpp` were removed; the current branch no longer touches that file.
- The previous Qwen-local mixed-precision quantization block was removed; the branch now uses the shared base `k_quant_linear` path from latest `main`.
- I checked compatibility with #2157 via `git merge-tree`; the same Qwen files are touched, but Git auto-merges them with no conflict markers.
- Rebuilt and verified Qwen3.5 0.8B and 9B export + TRT-RTX inference with the rebuilt wheel.

src/models/position_inputs.cpp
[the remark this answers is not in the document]
yen-shi: Addressed by the rewrite. The previous Qwen-specific runtime changes in `position_inputs.cpp` were removed, so the current branch no longer touches this file; Qwen3.5 text-only position handling is kept in the builder/export path instead of shared runtime position-input code.

src/models/position_inputs.cpp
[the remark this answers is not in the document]
yen-shi: Addressed by the rewrite. The previous Qwen-specific runtime changes in `position_inputs.cpp` were removed, so the current branch no longer touches this file; Qwen3.5 text-only position handling is kept in the builder/export path instead of shared runtime position-input code.

src/models/position_inputs.cpp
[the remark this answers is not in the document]
yen-shi: Addressed by the rewrite. The previous Qwen-specific runtime changes in `position_inputs.cpp` were removed, so the current branch no longer touches this file; Qwen3.5 text-only position handling is kept in the builder/export path instead of shared runtime position-input code.

src/models/position_inputs.cpp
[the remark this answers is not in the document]
yen-shi: Addressed by the rewrite. The previous Qwen-specific runtime changes in `position_inputs.cpp` were removed, so the current branch no longer touches this file; Qwen3.5 text-only position handling is kept in the builder/export path instead of shared runtime position-input code.

src/models/position_inputs.cpp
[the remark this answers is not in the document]
yen-shi: Addressed by the rewrite. The previous Qwen-specific runtime changes in `position_inputs.cpp` were removed, so the current branch no longer touches this file; Qwen3.5 text-only position handling is kept in the builder/export path instead of shared runtime position-input code.

src/models/model.cpp
yen-shi: Addressed in the two-commit rewrite and latest-main rebase. The overlap with #2157 is isolated in the first commit, while the second commit contains only the TRT-RTX-specific Qwen3.5 delta. I also checked this branch against #2157 with `git merge-tree`; the touched files auto-merge with no conflict markers.

src/models/multi_modal.cpp
yen-shi: Addressed in the two-commit rewrite and latest-main rebase. The overlap with #2157 is isolated in the first commit, while the second commit contains only the TRT-RTX-specific Qwen3.5 delta. I also checked this branch against #2157 with `git merge-tree`; the touched files auto-merge with no conflict markers.

src/models/position_inputs.cpp
yen-shi: Addressed by the rewrite. The previous Qwen-specific runtime changes in `position_inputs.cpp` were removed, so the current branch no longer touches this file; Qwen3.5 text-only position handling is kept in the builder/export path instead of shared runtime position-input code.

src/python/py/models/builders/qwen.py
yen-shi: Addressed in the rewritten branch. I removed the Qwen-local mixed-precision quantization block and now rely on the shared base `k_quant_linear` path from latest `main`; the remaining Qwen builder delta is limited to text-only/TRT-RTX export plumbing.

src/models/position_inputs.cpp
yen-shi: Addressed by the rewrite. The previous Qwen-specific runtime changes in `position_inputs.cpp` were removed, so the current branch no longer touches this file; Qwen3.5 text-only position handling is kept in the builder/export path instead of shared runtime position-input code.

src/models/position_inputs.cpp
yen-shi: Addressed by the rewrite. The previous Qwen-specific runtime changes in `position_inputs.cpp` were removed, so the current branch no longer touches this file; Qwen3.5 text-only position handling is kept in the builder/export path instead of shared runtime position-input code.

src/models/position_inputs.cpp
yen-shi: Addressed by the rewrite. The previous Qwen-specific runtime changes in `position_inputs.cpp` were removed, so the current branch no longer touches this file; Qwen3.5 text-only position handling is kept in the builder/export path instead of shared runtime position-input code.

src/models/position_inputs.cpp
yen-shi: Addressed by the rewrite. The previous Qwen-specific runtime changes in `position_inputs.cpp` were removed, so the current branch no longer touches this file; Qwen3.5 text-only position handling is kept in the builder/export path instead of shared runtime position-input code.

yen-shi:
Hi @apsonawane and @kunal-vaishnavi,
I have updated my changes based on PR #2100 , PR #2148, and PR #2157. Trim down changes to only 2 commits, one is changes borrowed from https://github.com/microsoft/onnxruntime-genai/pull/2157, and the other commit enables the shared past/present buffers to TRT-RTX EP to do optimized inference with CUDA graph enabled.

Can you review again please? Thanks!

src/python/py/models/builders/base.py
```diff
@@ -1619,6 +1619,11 @@ def _make_layernorm_op(self, layer_id, layernorm, skip, simple, location):
             root_input = inputs[0]
             skip_input = inputs[1] if skip else None
 
+        # Cast insertion can redirect SkipLayerNorm's fourth output to a casted
+        # value. Pass that redirected name to the primitive so QDQ export does
```
kunal-vaishnavi: Can you explain this further? Why is this an issue exactly? QDQ should impact the MatMul ops used in the model and not Cast ops. Some screenshots of the invalid ONNX model and valid ONNX model in Netron would be helpful to visualize this comparison.

examples/python/common.py
```diff
@@ -557,6 +557,7 @@ def get_ep_args(parser: argparse.ArgumentParser) -> None:
         "cpu",                             # CPU EP
         "cuda",                            # GenAI canonical name for CUDA EP
         "CUDAExecutionProvider",           # CUDA EP
+        "NvTensorRtRtx",                   # GenAI canonical name for Nvidia IHV EP
```
baijumeswani: Why is this change needed? Can you use the default `-e ` argument instead (i.e. `follow_config`)?

src/models/recurrent_state.cpp
```diff
@@ -88,22 +88,42 @@ RecurrentState::RecurrentState(State& state)
...
+      pasts_[i * 2] = OrtValue::CreateTensor(allocator, conv_shape_, conv_type_);
+      pasts_[i * 2 + 1] = OrtValue::CreateTensor(allocator, recurrent_shape_, recurrent_type_);
+
+      presents_.push_back(OrtValue::CreateTensor(allocator, conv_shape_, conv_type_));
+      presents_.push_back(OrtValue::CreateTensor(allocator, recurrent_shape_, recurrent_type_));
+    }
+
+    // Zero-initialize past and present states
+    ZeroStates(pasts_);
+    ZeroStates(presents_);
```
baijumeswani: @apsonawane any reason to continue supporting `past_present_share_buffer_ ` = false?

Should we make past_present_share_buffer_ = true and have only 1 code path?

apsonawane: I tried this change but was not able to run it still getting CUDA illegal memory error. Shared buffer mode still does not work I think. @yen-shi can you confirm it works well for CUDA as well.
I would suggest not to enable past_present_buffer until we are sure that it works with CUDA and TRT.
If the code is specific to TRT then we can add a condition here then.

apsonawane:
@yen-shi can you integrate this change as well: https://github.com/microsoft/onnxruntime-genai/pull/2127
I feel this will break cuda graph capture, because LpNormalization op does have a cuda kernel so it fallbacks to CPU. Please confirm your change works after getting PR in your branch

src/python/py/models/builders/base.py
yen-shi: This was only needed when I was using Qwen2.5/VL path to run for Qwen3.5 model. No longer needed, have removed, thanks for pointing out!

examples/python/common.py
yen-shi: So I did a further look into this issue. There's a naming mismatch from genai_config.json vs. TRT-RTX EP registration.

genai_config.json uses "NvTensorRtRtx", and we need the longer name to register EP:
```
og.register_execution_provider_library(
    "NvTensorRTRTXExecutionProvider",
    ep_path,
)
```

My local test command will not pass if there's no mapping between two names:

```
python onnxruntime-genai\examples\python\model-qa.py `
  -m $MODEL `
  -e NvTensorRtRtx `
  --ep_path $PKG\onnxruntime_providers_nv_tensorrt_rtx.dll `
  -up "Write me a poem about AI." `
  -l 4096 `
```

The issue is that if going through `use_winml` path, then there'll not be an issue because TRT-RTX EP has been registered. However, the default `elif ep_path` will then hit an error.

```
    if use_winml:
        # If use_winml is true, all winml execution providers will be queried and registered.
        # Please refer to https://pypi.org/project/windowsml/
        try:
            import winml
            for ep_name in winml.register_execution_providers():
                print(f"Registered WinML execution provider: {ep_name}")
        except Exception as e:
            print(f"Failed to register WinML execution providers: {e}")
    elif ep_path:
        og.register_execution_provider_library(ep, ep_path)

        print(f"Registered {ep} from {ep_path} successfully!")
```

I need to add a further check to map "NvTensorRtRtx" to "NvTensorRTRTXExecutionProvider" before registering the EP.

baijumeswani: Can you use the ort name for testing? It avoids the need for mapping. And is also probably cleaner.

```sh
python onnxruntime-genai\examples\python\model-qa.py `
  -m $MODEL `
  -e NvTensorRTRTXExecutionProvider `
  --ep_path $PKG\onnxruntime_providers_nv_tensorrt_rtx.dll `
  -up "Write me a poem about AI." `
  -l 4096 `
```

yen-shi: Hi @baijumeswani,

So the conflict still exists either ways:

1. If using `-e NvTensorRTRTXExecutionProvider`, then the get_config() from here https://github.com/microsoft/onnxruntime-genai/blob/main/examples/python/common.py#L73 will do `config.append_provider("NvTensorRTRTXExecutionProvider")` and will have a conflict with the name in session_options.cpp - https://github.com/microsoft/onnxruntime-genai/blob/21192daaf005caa55d1c1ddd58efcf8dad5b6886/src/models/session_options.cpp#L153

2. If using follow_config to read the `NvTensorRtRtx` from genai_config.json, the name itself is correct, but `NvTensorRTRTXExecutionProvider` DLL is still not registered. (using default path not use_winml path) - my reading is that this path may work when `use_winml` is True, because winml package will handle the registration of NvTensorRTRTXExecutionProvider DLL.
```
"provider_options": [
  {
    "NvTensorRtRtx": {
      "enable_cuda_graph": "1"
    }
  }
]
```

With the explicit model-qa.py testing command, it'll still hit an error.
Because in this Qwen3.5 TRT-RTX EP test, I have the `enable_cuda_graph` option enabled for `provider_options`, which is checked in https://github.com/microsoft/onnxruntime-genai/blob/main/src/config.cpp#L1400

```
            "session_options": {
                "provider_options": [
                    {
                        "NvTensorRtRtx": {
                            "enable_cuda_graph": "1"
                        }
                    }
                ]
            },
```

baijumeswani: I see. That makes sense. In that case, could you please update this block of code:

https://github.com/microsoft/onnxruntime-genai/blob/21192daaf005caa55d1c1ddd58efcf8dad5b6886/examples/python/common.py#L69-L73

to not clear the providers when --ep_path is provided?

```python
if not ep_path and ep != "follow_config":
    config.clear_providers()
    if ep != "cpu":
        print(f"Setting model to {ep}")
        config.append_provider(ep)
```

You would need to pass in the ep_path to the function.

That would mean that if `ep_path` is passed in, do not clear the providers. Then it should work.

src/models/recurrent_state.cpp
yen-shi: There're various places currently in the codebase checking `model_.p_device_->GetType() == DeviceType::NvTensorRtRtx`, so the shared buffer will not work with CUDA EP for now.

Do you want me to guard this `if (past_present_share_buffer_)` branch with `model_.p_device_->GetType() == DeviceType::NvTensorRtRtx` as well?
I feel like this is model specific file and should not include checks for EP.

examples/python/common.py
yen-shi: Sure, updated as requested.

yen-shi:
> @yen-shi can you integrate this change as well: #2127 I feel this will break cuda graph capture, because LpNormalization op does have a cuda kernel so it fallbacks to CPU. Please confirm your change works after getting PR in your branch

My changes have already been applied on top of #2127, and I've tested TRT-RTX EP on Qwen3.5 models without issues.

src/models/recurrent_state.cpp
baijumeswani: No, let's not add the device type check. @apsonawane can you confirm that the cuda ep models currently being built will work with this change? Do cuda ep models have `past_present_share_buffer_` = false in the genai_config.json? Do we need to separate the default kv cache past_present_share_buffer from the recurrent state past_present_share_buffer or have another way to determine whether or not the past_present_share_buffer should be enabled?

apsonawane: @yen-shi I tried the CUDA models, when we export the model with olive recipe past_present_share_buffer is true, so the else condition is never hit. Can you remove the else statement and keep the shared buffer path only

baijumeswani:
@yen-shi could you please resolve the merge-conflicts as well?

src/models/recurrent_state.cpp
yen-shi: Ok,so past_present_share_buffer_ is required to be true then, I update the else branch to throw an exception..

src/models/recurrent_state.cpp
```diff
@@ -114,7 +124,10 @@ void RecurrentState::Add() {
 
   const int num_layers = static_cast<int>(layer_indices_.size());
   for (int i = 0; i < num_layers * 2; ++i) {
-    state_.inputs_.push_back(pasts_[i].get());
+    // In shared-buffer mode the same OrtValue is intentionally registered as
+    // input and output. Non-shared mode keeps the older ping-pong buffers.
+    auto* past = past_present_share_buffer_ ? presents_[i].get() : pasts_[i].get();
```
baijumeswani: It seems like you will never reach this state if past_present_share_buffer_ = false. Can you remove the conditional checks?

yen-shi: Sure will do another cleanup

src/models/recurrent_state.h
```diff
@@ -47,6 +47,8 @@ struct RecurrentState {
 
   std::vector<int64_t> conv_shape_;
   std::vector<int64_t> recurrent_shape_;
+
+  bool past_present_share_buffer_{};
```
baijumeswani:
```suggestion
```

src/models/recurrent_state.cpp
```diff
@@ -88,22 +88,32 @@ RecurrentState::RecurrentState(State& state)
 
   const int num_layers = static_cast<int>(layer_indices_.size());
 
-  pasts_.resize(num_layers * 2);
+  past_present_share_buffer_ = state_.params_->IsPastPresentShareBufferEnabled(model_.config_->model.type);
+  if (g_log.enabled && past_present_share_buffer_) {
+    Log("info", "RecurrentState: using shared past/present buffers");
+  }
```
baijumeswani:
```suggestion
  if (!sPastPresentShareBufferEnabled(model_.config_->model.type)) {
    throw std::runtime_error(
        "RecurrentState requires past_present_share_buffer=true. "
        "Set past_present_share_buffer to true in genai_config.json.");
  }
```

src/models/recurrent_state.cpp
```diff
@@ -88,22 +88,32 @@ RecurrentState::RecurrentState(State& state)
...
+    }
 
-    presents_.push_back(OrtValue::CreateTensor(allocator, conv_shape_, conv_type_));
-    presents_.push_back(OrtValue::CreateTensor(allocator, recurrent_shape_, recurrent_type_));
+    ZeroStates(presents_);
+  } else {
+    throw std::runtime_error(
+        "RecurrentState requires past_present_share_buffer=true. "
+        "Set past_present_share_buffer to true in genai_config.json.");
   }
```
baijumeswani:
```suggestion
  // Qwen3.5 linear-attention state is a compressed recurrent state, not a
  // token-indexed KV cache. For graph replay, bind each state tensor as both
  // past input and present output so ORT/TRT-RTX sees stable addresses.
  // The EP/plugin kernels must read the previous contents before writing the
  // updated state back to the same buffer.
  for (int i = 0; i < num_layers; ++i) {
    presents_.push_back(OrtValue::CreateTensor(allocator, conv_shape_, conv_type_));
    presents_.push_back(OrtValue::CreateTensor(allocator, recurrent_shape_, recurrent_type_));
  }

  ZeroStates(presents_);
```

src/models/recurrent_state.cpp
baijumeswani:
```suggestion
    auto* past = presents_[i].get();
```

src/models/recurrent_state.cpp
```diff
@@ -128,14 +141,18 @@ void RecurrentState::Add() {
...
     for (int i = 0; i < num_layers * 2; ++i) {
-      past_byte_spans_.push_back(ByteWrapTensor(device, *pasts_[i]));
+      auto& past = past_present_share_buffer_ ? presents_[i] : pasts_[i];
+      past_byte_spans_.push_back(ByteWrapTensor(device, *past));
       present_byte_spans_.push_back(ByteWrapTensor(device, *presents_[i]));
     }
   }
 }
 
 void RecurrentState::Update() {
```
baijumeswani, line 128: Entire function should be a no-op.

src/models/recurrent_state.cpp
```diff
@@ -169,14 +186,20 @@ void RecurrentState::RewindTo(size_t index) {
     return;
   }
 
-  const int num_layers = static_cast<int>(layer_indices_.size());
+  if (past_present_share_buffer_) {
+    // Shared recurrent states keep stable input/output pointers for graph replay.
+    // Reset the state contents in place without rebinding.
+    ZeroStates(presents_);
+    return;
+  }
```
baijumeswani:
```suggestion
  // Shared recurrent states keep stable input/output pointers for graph replay.
  // Reset the state contents in place without rewinding.
  ZeroStates(presents_);
  return;
```

And remove anything after this from the function.

yen-shi:
@baijumeswani I've done final clean-up, can you review again please? Thanks!

examples/python/common.py
```diff
@@ -59,14 +65,16 @@ def get_config(path: str, ep: str, ep_options: dict[str, str] = {}, search_optio
         ep (str): Name of execution provider to set
         ep_options (dict[str, str]): Map of EP-specific option names and their values
         search_options (dict[str, int]): Map of search-specific option names and their values
+        ep_path (str): Path to an external execution provider library. If set, the
```
kunal-vaishnavi: Can we reorder the args so that `ep_path` is after `ep` and does not have a default value? We are always passing it in so it should be grouped with the required args.

baijumeswani (dismissed):

examples/python/common.py
baijumeswani: Will submit a follow-up PR for this

baijumeswani:
@yen-shi I hope you don't mind if I push on top of your branch. There seems to be a test failing. And I am also addressing @kunal-vaishnavi comment.

I want to include this PR for the ort-genai release 0.14.0

kunal-vaishnavi (approved):

apsonawane (approved):

PR #2141, opened 2026-05-08
fix: enable Generator.rewind_to(0) for multimodal models

### LINKED ISSUE #2140: Generator.rewind_to(0) fails for multimodal models (gemma4, phi4mm)
## Description

`Generator.rewind_to(0)` fails with an attention mask / KV cache inconsistency error for multimodal model types (`gemma4`, `phi4mm`). This prevents Generator reuse for batch loglikelihood evaluation, forcing a new Generator per sample.

## Reproduction

```python

model = og.Model(og.Config("path/to/gemma4-e2b-it-onnx"))
params.set_search_options(max_length=4096, past_present_share_buffer=False)

gen = og.Generator(model, params)
gen.append_tokens([[2, 100, 200, 300, 400]])
logits = gen.get_output("logits")  # Works

gen.rewind_to(0)  # Appears to succeed
gen.append_tokens([[2, 101, 201]])  # CRASHES
```

**Error:**
```
RuntimeError: inconsistent total_sequence_length (between attn_mask and past_key and past_value)
```

## Root Causes

Three issues:

### 1. `MultiModalPipelineState::RewindTo` is a no-op (inherits base `State::RewindTo` which does nothing)
`MultiModalPipelineState` does not override `RewindTo`, so the decoder's KV cache and attention mask are never reset. Fix: delegate to `decoder_state_->RewindTo(index)`.

### 2. `DecoderState::RewindTo` is also a no-op (same reason)
`DecoderState` (used inside multimodal pipeline) does not override `RewindTo`. Fix: implement it to delegate to `position_inputs_`, `kv_cache_`, and `recurrent_state_`.

### 3. `DefaultKeyValueCache::RewindTo(0)` fails bounds check
Line 411: `shape_[2] <= static_cast<int>(index)` evaluates to `0 <= 0` = true, throwing an error even for `index=0`. This is because `shape_[2]` starts at 0 before any KV cache updates. Fix: skip bounds check when `index == 0`.

## Performance Motivation

Generator reuse via `rewind_to(0)` would eliminate the need to create a new `og.Generator` per sample in lm-eval benchmarks. Current benchmark throughput:
- **With new Generator per call (ortgenai)**: 9.5 req/s
- **Direct ORT (no Generator overhead)**: 142 req/s
- **PyTorch HF**: 185 req/s

## Fix

PR forthcoming with the three fixes above. Tested on Gemma4 E2B-IT with CUDA EP — 5 consecutive rewind_to(0) + append_tokens cycles succeed.

justinchuby:
## Summary

Fix `Generator.rewind_to(0)` for multimodal model types (`gemma4`, `phi4mm`). Previously, `rewind_to(0)` appeared to succeed but the next `append_tokens` call would crash with an attention mask / KV cache inconsistency.

### 1. `MultiModalPipelineState::RewindTo` was a no-op
The base `State::RewindTo` does nothing. Added override that delegates to `decoder_state_->RewindTo(index)` and resets `is_prompt_`.

### 2. `DecoderState::RewindTo` was also a no-op
Added override matching `DecoderOnly_State::RewindTo` pattern — delegates to `position_inputs_`, `kv_cache_`, and `recurrent_state_`.

### 3. `DefaultKeyValueCache::RewindTo(0)` bounds check failure
`shape_[2]` starts at 0 before any KV updates, so `shape_[2] <= 0` triggered `RewindTo(0)` to throw. Fixed by allowing `index == 0` to bypass the bounds check (rewinding to empty is always valid).

## Testing

Tested on Gemma4 E2B-IT with CUDA EP:
- 5 consecutive `rewind_to(0)` cycles with varying sequence lengths
- Different input tokens each cycle
- All produce correct logits

## Motivation

Generator reuse via `rewind_to(0)` eliminates the overhead of creating a new `og.Generator` per sample in lm-eval benchmarks. Without this fix, the `ortgenai` lm-eval backend is 15× slower than direct ORT session for multimodal models.

Fixes #2140

Depends on https://github.com/microsoft/onnxruntime-genai/pull/2123

### DIFF
--- a/src/models/kv_cache.cpp
+++ b/src/models/kv_cache.cpp
@@ -299,7 +299,7 @@ void DefaultKeyValueCache::Update(DeviceSpan<int32_t> beam_indices, int total_le
 void DefaultKeyValueCache::RewindTo(size_t index) {
   if (past_present_share_buffer_) {
     return;
-  } else if (shape_[2] <= static_cast<int>(index)) {
+  } else if (index > 0 && shape_[2] <= static_cast<int>(index)) {
     throw std::runtime_error("Requested length of rewind is greater than the current length.");
   }
 
--- a/src/models/multi_modal.cpp
+++ b/src/models/multi_modal.cpp
@@ -62,6 +62,14 @@ int64_t GetImageFeatureBatchSize(const std::vector<ExtraInput>& extra_inputs) {
 
 MultiModalLanguageModel::MultiModalLanguageModel(std::unique_ptr<Config> config, OrtEnv& ort_env, bool vision, bool speech)
     : Model(std::move(config)) {
+  // Override p_device_inputs_ to CPU for multimodal models on CUDA.
+  // ORT's InsertedPrecisionFreeCast (F32→F16) is always placed on CPU.
+  // When GenAI allocates inputs on GPU, the CPU Cast reads GPU memory → crash.
+  // Keeping inputs on CPU avoids this; the embedding session handles the transfer.
+  if (p_device_->GetType() == DeviceType::CUDA || p_device_->GetType() == DeviceType::NvTensorRtRtx) {
+    p_device_inputs_ = GetDeviceInterface(DeviceType::CPU);
+    p_device_scoring_ = GetDeviceInterface(DeviceType::CPU);
+  }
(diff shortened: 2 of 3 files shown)


baijumeswani:
Closing this. Please re-open if still relevant.

PR #2145, opened 2026-05-08
Expose mutable sampling parameters on live Generator

qjia7:
## Summary

- Add  and  C API functions to change sampling parameters (temperature, top_k, top_p, repetition_penalty, do_sample) on a live generator without destroying/recreating it
- Generator stores optional overrides that take precedence over the original  values
-  is recomputed when relevant parameters change
- Python binding exposes  with kwargs

## Motivation

Sampling parameters (temperature, top_k, top_p, do_sample, repetition_penalty) only affect logit post-processing and are independent of the KV-cache. Previously, changing any parameter required destroying the generator and recreating it, discarding the entire KV-cache. This is wasteful for scenarios like Edge's Prompt API where mid-conversation parameter changes are needed.

## Test plan

- [x] 6 unit tests in  covering greedy-to-sampling switch, sampling-to-greedy, temperature change, top_k change, invalid name error, and isolation between generators
- [x] 1 model test in  verifying KV-cache preservation after mid-generation parameter switch
- [x] Python e2e test: multi-turn conversation with parameter changes between turns produces coherent, contextually-aware responses
- [x] All 7 C++ tests pass, Python test passes with phi4 model

### DIFF
--- a/src/generators.cpp
+++ b/src/generators.cpp
@@ -552,6 +552,8 @@ void Generator::SetRuntimeOption(const char* key, const char* value) {
 }
 
 void Generator::SetSearchNumber(const char* name, double value) {
+  if (!search_)
+    throw std::runtime_error("SetSearchNumber is not supported for this model type");
   std::string_view n{name};
   if (n == "temperature") {
     temperature_override_ = static_cast<float>(value);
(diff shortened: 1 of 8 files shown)


baijumeswani:
Let's revisit this offline. I think we should do this work, but perhaps we can design something slightly differently.

PR #2147, opened 2026-05-08
Add VideoChat-Flash (OpenGVLab) language model support

anilmartha:
## Summary

Adds support for [OpenGVLab/VideoChat-Flash-Qwen2_5-7B_InternVideo2-1B](https://huggingface.co/OpenGVLab/VideoChat-Flash-Qwen2_5-7B_InternVideo2-1B), a video-language model whose architecture is `VideoChatFlashQwenForCausalLM`. The language backbone is standard Qwen2.5-7B (flat config, standard weight keys, 2D RoPE, `rope_theta=1e6`, 28 layers / 28 heads / 4 KV heads / hidden=3584). Because it does **not** use MRoPE, the new builder inherits from `QwenModel` directly rather than `Qwen25VLTextModel`.

### Python builder (`src/python/py/models/`)
- **`builders/qwen.py`** — New `VideoChatFlashQwenModel` subclass of `QwenModel`.
  - Overrides `make_genai_config()` to bypass the model's custom remote code (which imports `av`, `cv2`, `decord`, `imageio` even for config-only loads) by writing a temporary `Qwen2Config`-based `config.json` and patching `genai_config.json["model"]["type"]` back to `videochat_flash_qwen`.
  - Overrides `load_weights()` to load the LM backbone directly via `Qwen2ForCausalLM.from_pretrained()`, avoiding video deps.
  - Sets `model_type = "videochat_flash_qwen"` and `hf_remote = False`.
- **`builders/__init__.py`** — Exports `VideoChatFlashQwenModel`.
- **`builder.py`** —
  - Detects `VideoChatFlashQwenForCausalLM` by peeking at raw `config.json` (local dir or via `hf_hub_download`) before calling `AutoConfig`, then loads as `Qwen2Config` to avoid pulling in video libraries.
  - Routes `VideoChatFlashQwenForCausalLM` → `VideoChatFlashQwenModel` and defaults `exclude_embeds=True` (the text decoder receives `inputs_embeds` from the embedding merger). `exclude_embeds` is now opt-in (guarded with `not in extra_options`) so standalone `input_ids` export still works.

### C++ runtime (`src/`)
- **`models/videochat_flash_processor.{h,cpp}`** *(new)* — `VideoChatFlashProcessor` runs ORT-Extensions image preprocessing, transposes HWC→CHW, reshapes to `[1, num_frames, C, H, W]`, casts pixel values to the session input dtype, builds `input_ids` by replacing each `<|vision_start|>…<|vision_end|>` block with a fixed number of `<|image_pad|>` tokens, and emits `image_grid_thw` so `GetImageFeatureBatchSize` can determine `num_images` after the `pixel_values` name remap.
- **`models/model.cpp`** — Registers `"videochat_flash_qwen" → VideoChatFlashProcessor` in `MultiModalProcessor`.
- **`models/model_type.h`** — Adds `"videochat_flash_qwen"` to the `IsVLM()` list (size 6 → 7).
- **`config.{h,cpp}`** — Adds `Vision::num_visual_tokens` (default `0`) — fixed visual tokens per image; `0` falls back to computing from `image_grid_thw`. Required (`>0`) by the new processor.

## Files changed

| File | Change |
| --- | --- |
| `src/config.cpp` | Parse `num_visual_tokens` from vision config |
| `src/config.h` | Add `num_visual_tokens` field on `Vision` |
| `src/models/model.cpp` | Register `VideoChatFlashProcessor` |
| `src/models/model.h` | Header / license update |
| `src/models/model_type.h` | Add `videochat_flash_qwen` to VLM list |
| `src/models/videochat_flash_processor.cpp` | **New** — image processor |
| `src/models/videochat_flash_processor.h` | **New** — header |
| `src/python/py/models/builder.py` | Detect VCF arch, bypass remote-code video deps, route to new builder |
| `src/python/py/models/builders/__init__.py` | Export `VideoChatFlashQwenModel` |
| `src/python/py/models/builders/qwen.py` | Add `VideoChatFlashQwenModel` |

## Validation

- Text-only export of the Qwen2.5-7B backbone via `python -m onnxruntime_genai.models.builder -m OpenGVLab/VideoChat-Flash-Qwen2_5-7B_InternVideo2-1B -o <dir> -p fp32 -e cpu --extra_options exclude_embeds=false` succeeds **without** `av`/`cv2`/`decord`/`imageio` installed.
- Standalone inference on the exported `vcf-oga-fp32-standalone` model produces correct answers across 4 batched QA prompts (e.g. *Paris*, *56*, ONNX Runtime description) using `generator.append_tokens()` on the Qwen2.5-7B backbone.

## Test plan

- [ ] Build OGA with the new processor compiled in and confirm `IsVLM("videochat_flash_qwen")` returns `true`.
- [ ] Export the text decoder with `exclude_embeds=false` and run text-only inference end-to-end.
- [ ] Confirm `genai_config.json` contains `"type": "videochat_flash_qwen"` (full VLM) or `"type": "qwen2"` (text-only standalone).
- [ ] Image-input smoke test once `vision.onnx` and `embedding.onnx` land.


VishalX:
Olive Recipe: https://github.com/microsoft/olive-recipes/pull/406

anilmartha:
@microsoft-github-policy-service agree company="AMD"

src/python/py/models/builders/__init__.py
```diff
@@ -28,7 +28,7 @@
     Phi4MMModel,
     PhiModel,
 )
-from .qwen import Qwen3Model, Qwen3VLTextModel, Qwen25VLTextModel, Qwen35TextModel, QwenModel
+from .qwen import Qwen3Model, Qwen25VLTextModel, Qwen3VLTextModel, QwenModel, VideoChatFlashQwenModel
```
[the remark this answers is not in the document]
kunal-vaishnavi: This is the right fix.

src/python/py/models/builders/__init__.py
```diff
@@ -61,6 +61,7 @@
     "Qwen25VLTextModel",
     "Qwen35TextModel",
     "QwenModel",
+    "VideoChatFlashQwenModel",
```
kunal-vaishnavi: Can you insert this in alphabetical order?

anilmartha: Added the change.  @kunal-vaishnavi  Could you please review agian?

anilmartha:
Hi @kunal-vaishnavi 
Could you please review it again?

src/models/videochat_flash_processor.cpp
```diff
@@ -0,0 +1,225 @@
...
+    named_tensors->emplace(std::string(Config::Defaults::NumImageTokens),
+                           std::make_shared<Tensor>(std::move(num_img_tokens)));
+    return named_tensors;
+  }
+
+  // Run ORT Extensions image preprocessing (Decode → Resize → Rescale → Normalize)
+  ort_extensions::OrtxObjectPtr<OrtxTensorResult> result;
+  CheckResult(OrtxImagePreProcess(processor_.get(), images->images_.get(), result.ToBeAssigned()));
+
+  OrtxTensor* pixel_values = nullptr;
```
kunal-vaishnavi: This appears to be a memory leak. Can you fix this? There was a [recent PR](https://github.com/microsoft/onnxruntime-genai/pull/2168) that fixes this issue.

src/models/videochat_flash_processor.cpp
```diff
@@ -0,0 +1,225 @@
...
+    num_imgs = pv_shape[0];
+    height = pv_shape[1];
+    width = pv_shape[2];
+    channels = pv_shape[3];
+  } else {
+    throw std::runtime_error("VideoChatFlashProcessor: unexpected pixel_values rank " +
+                             std::to_string(pv_ndims) + " (expected 3 or 4)");
+  }
+
+  // Transpose HWC → CHW and reshape to [1, num_frames, C, H, W]
```
kunal-vaishnavi: Let's do this transpose as another step inside ORT Extensions before returning the tensor to ORT GenAI.

src/models/videochat_flash_processor.cpp
```diff
@@ -0,0 +1,225 @@
...
+    for (int64_t c = 0; c < channels; ++c) {
+      for (int64_t h = 0; h < height; ++h) {
+        for (int64_t w = 0; w < width; ++w) {
+          dst_img[c * height * width + h * width + w] = src_img[h * width * channels + w * channels + c];
+        }
+      }
+    }
+  }
+
+  auto converted_pv = ConvertPixelValues(*float_tensor, pixel_values_type_, allocator);
```
kunal-vaishnavi: Rather than creating a new method called `ConvertPixelValues`, can we re-use existing APIs that already convert dtypes? Here is an example.

https://github.com/microsoft/onnxruntime-genai/blob/ca9ecd2aa18cf0b7376dd404aa5ad699b2dd9d8c/src/models/whisper_processor.cpp#L33-L39

src/models/videochat_flash_processor.cpp
```diff
@@ -0,0 +1,225 @@
...
+  // rank-based lookup in GetImageFeatureBatchSize won't match; it falls through
+  // to image_grid_thw whose name is not remapped.
+  auto grid_thw = OrtValue::CreateTensor<int64_t>(allocator, std::vector<int64_t>{num_imgs, 3});
+  auto* grid_ptr = grid_thw->GetTensorMutableData<int64_t>();
+  for (int64_t i = 0; i < num_imgs; ++i) {
+    grid_ptr[i * 3 + 0] = 1;
+    grid_ptr[i * 3 + 1] = height;
+    grid_ptr[i * 3 + 2] = width;
+  }
+  named_tensors->emplace("image_grid_thw",
```
kunal-vaishnavi: If there are new input names, let's add their default names in the config (e.g. `Config::Defaults::ImageGridThw`) and reference them here.

src/python/py/models/builder.py
```diff
@@ -45,6 +46,7 @@
     Qwen25VLTextModel,
     Qwen35TextModel,
     QwenModel,
+    VideoChatFlashQwenModel,
```
kunal-vaishnavi: Let's preserve any existing alphabetical orders when inserting new items.

src/python/py/models/builders/qwen.py
```diff
@@ -909,6 +912,83 @@ def load_weights(self, input_path):
...
+        # Pre-configure before super().__init__() so the base class (Model)
+        # establishes these attributes on first assignment instead of us
+        # overwriting inherited values.
+        #
+        # The custom remote code requires video libraries (av, cv2, decord,
+        # imageio) which are not needed to export the LM backbone. Force
+        # hf_remote=False so base class helpers (make_genai_config,
+        # save_processing) use standard, non-remote-code paths.
+        extra_options = dict(extra_options) if extra_options else {}
+        extra_options["hf_remote"] = False
```
kunal-vaishnavi: To be consistent with other models, let's set any default extra options flags during architecture parsing. For example:

https://github.com/microsoft/onnxruntime-genai/blob/ca9ecd2aa18cf0b7376dd404aa5ad699b2dd9d8c/src/python/py/models/builder.py#L292-L299

src/python/py/models/builders/qwen.py
```diff
@@ -909,6 +912,83 @@ def load_weights(self, input_path):
...
+
+        # Model.__init__ sets self.model_type = config.architectures[0]. The
+        # HF architecture string ("VideoChatFlashQwenForCausalLM") has already
+        # served its purpose in the dispatch table in builder.py; swap it for
+        # the runtime identifier expected by genai_config.json and the C++
+        # runtime registration in model.cpp.
+        config.architectures = ["videochat_flash_qwen"]
```
kunal-vaishnavi: You can update the runtime identifier and the runtime registration via `self.model_type` instead of changing this.

https://github.com/microsoft/onnxruntime-genai/blob/ca9ecd2aa18cf0b7376dd404aa5ad699b2dd9d8c/src/python/py/models/builders/qwen.py#L693-L696

src/python/py/models/builders/qwen.py
```diff
@@ -909,6 +912,83 @@ def load_weights(self, input_path):
...
+    so the decoder receives inputs_embeds from the embedding merger model, which
+    fuses the InternVideo2 visual tokens with text embeddings.
+    """
+
+    def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
+        # Pre-configure before super().__init__() so the base class (Model)
+        # establishes these attributes on first assignment instead of us
+        # overwriting inherited values.
+        #
+        # The custom remote code requires video libraries (av, cv2, decord,
```
kunal-vaishnavi: It's fine if video libraries need to be temporarily installed. These libraries are on a per-model basis and are based on Hugging Face's requirements for loading a model. For example, Phi-4 mm requires PEFT for loading the model and we don't remove those library dependencies even though they are not technically needed for producing the ONNX model.

src/python/py/models/builders/qwen.py
```diff
@@ -909,6 +912,83 @@ def load_weights(self, input_path):
...
+        # Model.__init__ sets self.model_type = config.architectures[0]. The
+        # HF architecture string ("VideoChatFlashQwenForCausalLM") has already
+        # served its purpose in the dispatch table in builder.py; swap it for
+        # the runtime identifier expected by genai_config.json and the C++
+        # runtime registration in model.cpp.
+        config.architectures = ["videochat_flash_qwen"]
+
+        super().__init__(config, io_dtype, onnx_dtype, ep, cache_dir, extra_options)
+
+    def make_genai_config(self, model_name_or_path, extra_kwargs, out_dir):
```
kunal-vaishnavi: Same comment as [here](https://github.com/microsoft/onnxruntime-genai/pull/2147/changes#r3270339314). We should not need to override this method at all.

src/python/py/models/builders/qwen.py
```diff
@@ -909,6 +912,83 @@ def load_weights(self, input_path):
...
+        # make_genai_config in base.py calls AutoConfig with trust_remote_code,
+        # which triggers the video library imports. Instead, write a clean
+        # Qwen2-compatible config.json to a temp dir and let base class read it.
+        import json as _json
+        import shutil
+        import tempfile
+
+        from transformers import Qwen2Config
+
+        vcf_config = Qwen2Config.from_pretrained(model_name_or_path, token=self.hf_token, **extra_kwargs)
```
kunal-vaishnavi: If you want to avoid loading the video library imports by using a Qwen2Config for importing, can you create a generic `load_config` method in the base class (similar to `load_weights` below)? Then, we can call that method in the various locations that require a config. That method can then be overridden in this class to be Qwen-2 specific.

src/python/py/models/builder.py
```diff
@@ -191,7 +193,48 @@ def create_model(
     hf_token = parse_hf_token(extra_options.get("hf_token", "true"))
     hf_remote = extra_options.get("hf_remote", True)
 
-    config = AutoConfig.from_pretrained(hf_name, token=hf_token, trust_remote_code=hf_remote, **extra_kwargs)
+    # VideoChat-Flash uses custom remote code that imports heavy video libraries
```
kunal-vaishnavi: All of the below changes should be reverted. They are too specific for this model and this section needs to remain generic. We can try the `load_config` suggestion from earlier to resolve this.

src/python/py/models/builder.py
```diff
@@ -289,6 +332,11 @@ def create_model(
         onnx_model = Phi4MMModel(config, io_dtype, onnx_dtype, execution_provider, cache_dir, extra_options)
     elif config.architectures[0] == "Qwen2ForCausalLM":
         onnx_model = QwenModel(config, io_dtype, onnx_dtype, execution_provider, cache_dir, extra_options)
+    elif config.architectures[0] == "VideoChatFlashQwenForCausalLM":
```
kunal-vaishnavi, line 294: Can we remove the `if` condition?

https://github.com/microsoft/onnxruntime-genai/blob/f32444df344f2dbf657c87026c80a14e299daa66/src/python/py/models/builder.py#L292-L299

Multi-modal models do not need to check this since the embedding ONNX model will go from `input_ids` to `inputs_embeds`.

anilmartha:
Hi @kunal-vaishnavi 
I have addressed the review comments. Could you please take another look?

src/python/py/models/builders/qwen.py
```diff
@@ -913,6 +916,39 @@ def load_weights(self, input_path):
...
+        # Override model_type for the C++ runtime registration in model.cpp
+        # and genai_config.json. Same pattern as Qwen3VLTextModel.
+        # Base class transforms this to "videochat_flash_qwen" via:
+        #   model_type[:model_type.find("For")].lower()
+        self.model_type = "VideoChat_Flash_QwenForCausalLM"
+
+    def load_weights(self, input_path):
+        # The LM backbone is identical to Qwen2ForCausalLM. Load it directly
+        # to avoid the custom remote code (which requires video libraries).
+        from transformers import Qwen2ForCausalLM
```
kunal-vaishnavi: Can we move this import to the global imports at the top of this file?

kunal-vaishnavi (dismissed):

src/python/py/models/builders/qwen.py
anilmartha: moved it to global imports

kunal-vaishnavi (approved):

PR #2150, opened 2026-05-08
Nvidia Parakeet Tdt ASR support

nenad1002:
Add Parakeet TDT speech recognition model support
Adds end-to-end support for **NVIDIA's Parakeet TDT ASR** model in onnxruntime-genai.

Code here will support both v2 (English-only) and v3 (multilingual) model - both CPU and CUDA.

Note: Only for batch inference - can support any length of the audio.

**What's included**
New model type parakeet_tdt with encoder / decoder / joint ONNX subgraphs wired through Model::Create.

TDT decoding loop in src/models/parakeet.cpp: joint network emits [token_logits | duration_logits]; per-frame argmax over tokens and durations with the standard TDT constraints:
Blank + duration 0 is forbidden (prevents single-frame loop-hang).
max_symbols_per_frame cap forces a one-frame skip if the model wants to stay on a frame indefinitely.
Processor (src/models/parakeet_processor.cpp / .h): full-audio log-mel computed up front via the audio extension op; encoder consumes the mel in chunked fashion.
Config plumbing (src/config.h, src/config.cpp): adds tdt_durations, blank_id, max_symbols_per_frame, etc., matching the genai_config.json

Sample: examples/python/parakeet.py - minimal CLI to transcribe an audio file.
Tests:
C++: test/c_api_tests.cpp - model creation + decode smoke test.
Python: test/python/test_parakeet_tdt.py - end-to-end transcription on jfk.flac (short) and tedlium_long_120s.flac fixture (~2 MB) for long-audio coverage.

Long-audio handling: encoder is **run on overlapping up to 10 s chunks with both left and right context** (configurable via chunk_samples / left_context_samples / right_context_samples in genai_config.json.
Only the chunk-proper region is decoded; context frames just provide receptive field at boundaries. This keeps per-step latency bounded while supporting arbitrary-length audio alongside with much higher RTFx than for native true streaming models.

The model is multilingual at just 0.6B. 
The original model link:
https://huggingface.co/nvidia/parakeet-tdt-0.6b-v3


src/models/parakeet.cpp
```diff
@@ -0,0 +1,444 @@
...
+  auto& cpu_allocator = model_.allocator_cpu_;
+  auto& cpu_device = *GetDeviceInterface(DeviceType::CPU);
+  auto& inference_device = *model_.p_device_inputs_;
+  auto run_options = OrtRunOptions::Create();
+
+  const int num_durations = static_cast<int>(cfg_.tdt_durations.size());
+  const int blank_id = cfg_.blank_id;
+  const int max_sym = cfg_.max_symbols_per_step;
+  const int64_t hidden_dim = cfg_.hidden_dim;
+  const int64_t dec_dim = cfg_.decoder_lstm_dim;
```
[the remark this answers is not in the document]
nenad1002: Not possible for num durations to be 0, and other stuff are validated already

src/models/parakeet.cpp
```diff
@@ -0,0 +1,444 @@
...
+  const char* dec_input_names[] = {
+      cfg_.dec_in_targets.c_str(), cfg_.dec_in_targets_length.c_str(),
+      cfg_.dec_in_lstm_hidden_state.c_str(), cfg_.dec_in_lstm_cell_state.c_str()};
+  OrtValue* dec_inputs[] = {targets.get(), targets_length.get(),
+                            dec_.state_h.get(), dec_.state_c.get()};
+
+  const char* dec_output_names[] = {
+      cfg_.dec_out_outputs.c_str(), cfg_.dec_out_outputs_length.c_str(),
+      cfg_.dec_out_lstm_hidden_state.c_str(), cfg_.dec_out_lstm_cell_state.c_str()};
+
```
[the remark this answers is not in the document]
nenad1002: If those keys are missing from genai_config.json the JSON parser will throw an error.

src/models/parakeet.cpp
```diff
@@ -0,0 +1,486 @@
...
+  join_out_logits = jo.outputs.logits;
+
+  dec_in_targets = dec.inputs.targets;
+  dec_in_targets_length = dec.inputs.targets_length;
+  dec_in_lstm_hidden_state = dec.inputs.lstm_hidden_state;
+  dec_in_lstm_cell_state = dec.inputs.lstm_cell_state;
+  dec_out_outputs = dec.outputs.outputs;
+  dec_out_outputs_length = dec.outputs.outputs_length;
+  dec_out_lstm_hidden_state = dec.outputs.lstm_hidden_state;
+  dec_out_lstm_cell_state = dec.outputs.lstm_cell_state;
```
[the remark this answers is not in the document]
nenad1002: Not applicable, we dont do this with other models either

test/python/test_parakeet_tdt.py
```diff
@@ -0,0 +1,148 @@
...
+        "help the poorest two billion live better lives. But energy and climate are "
+        "extremely important to these people. In fact, more important than to anyone "
+        "else on the planet. The climate getting worse means that many years their "
+        "crops won't grow. There'll be too much rain, not enough rain. Things will "
+        "change in ways that their fragile environment simply can't support. And that "
+        "leads to starvation, it leads to uncertainty, it leads to unrest. The climate changes "
+        "will be terrible for them. Also, the price of energy is very important to them. In "
+        "fact, if you could pick just one thing to lower the price of, to reduce "
+        "poverty, by far you would pick energy. Now, the price of energy has come "
+        "down over time. Really, advanced civilization is based on advances in in"
```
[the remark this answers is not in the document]
nenad1002: Not a typo

src/models/parakeet.cpp
```diff
@@ -0,0 +1,486 @@
...
+}
+
+int32_t ParakeetTdtState::EmitNextToken() {
+  auto& cpu_allocator = model_.allocator_cpu_;
+  auto run_options = OrtRunOptions::Create();
+
+  const int num_durations = static_cast<int>(cfg_.tdt_durations.size());
+  const int blank_id = cfg_.blank_id;
+  const int max_sym = cfg_.max_symbols_per_step;
+  const int64_t hidden_dim = cfg_.hidden_dim;
```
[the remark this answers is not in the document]
nenad1002: Not possible

examples/python/parakeet.py
```diff
@@ -0,0 +1,99 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+"""
+Parakeet TDT speech recognition
+
+  python parakeet.py --model_path <model_dir> --audio_file <audio>
```
kunal-vaishnavi, line 6: Rather than creating new example files for each speech model, let's find a way to support all of them in a more generic and scalable way. Otherwise, we will need to keep creating, maintaining, and validating each model's example in each language.

src/config.cpp
```diff
@@ -878,6 +882,10 @@ struct Speech_Element : JSON::Element {
     }
   }
 
+  Element& OnArray(std::string_view name) override {
+    throw JSON::unknown_value_error{};
```
kunal-vaishnavi: Why is this needed?

src/models/parakeet.cpp
```diff
@@ -0,0 +1,486 @@
...
+}
+
+ParakeetTdtState::ParakeetTdtState(const ParakeetTdtModel& model, const GeneratorParams& params)
+    : State{params, model},
+      model_{model},
+      cfg_{model.parakeet_config_} {
+  logits_size_ = model_.config_->model.vocab_size;
+
+  // Allocate the persistent logits buffer (CPU-resident, one-hot).
+  logits_buffer_.assign(static_cast<size_t>(logits_size_), 0.0f);
```
kunal-vaishnavi: Why is there a need to manually create and allocate a logits buffer? ORT GenAI will already do that for you with the right state.

src/models/parakeet.cpp
```diff
@@ -0,0 +1,486 @@
...
+  // Prime the decoder with the blank token to obtain the initial decoder_output.
+  StepDecoder(static_cast<int32_t>(cfg_.blank_id));
+  dec_.last_token = cfg_.blank_id;
+}
+
+void ParakeetTdtState::StepDecoder(int32_t token_id) {
+  // Tiny scalar inputs stay on CPU; ORT bridges them automatically when the
+  // decoder session is on a different EP.
+  auto& cpu_allocator = model_.allocator_cpu_;
+  auto run_options = OrtRunOptions::Create();
```
kunal-vaishnavi: Run options should be read from the GenAI config.

src/models/parakeet.cpp
```diff
@@ -0,0 +1,486 @@
...
+      cfg_.dec_in_targets.c_str(), cfg_.dec_in_targets_length.c_str(),
+      cfg_.dec_in_lstm_hidden_state.c_str(), cfg_.dec_in_lstm_cell_state.c_str()};
+  OrtValue* dec_inputs[] = {targets.get(), targets_length.get(),
+                            dec_.state_h.get(), dec_.state_c.get()};
+
+  const char* dec_output_names[] = {
+      cfg_.dec_out_outputs.c_str(), cfg_.dec_out_outputs_length.c_str(),
+      cfg_.dec_out_lstm_hidden_state.c_str(), cfg_.dec_out_lstm_cell_state.c_str()};
+
+  auto dec_outputs = model_.session_decoder_->Run(
```
kunal-vaishnavi: Let's use the GenAI infra via `State::Run` for handling all of this. We shouldn't have to do all this manually.

https://github.com/microsoft/onnxruntime-genai/blob/3ef7ab1200e3881fc2f27d8d1f6b88d6bec71e67/src/models/nemotron_speech.cpp#L170-L249

src/models/parakeet.cpp
```diff
@@ -0,0 +1,486 @@
...
+  auto signal_length = OrtValue::CreateTensor(cpu_allocator, len_shape, ONNX_TENSOR_ELEMENT_DATA_TYPE_INT64);
+  *signal_length->GetTensorMutableData<int64_t>() = num_mel_frames;
+
+  const char* enc_input_names[] = {cfg_.enc_in_audio.c_str(), cfg_.enc_in_length.c_str()};
+  OrtValue* enc_inputs[] = {processed_signal.get(), signal_length.get()};
+  const char* enc_output_names[] = {cfg_.enc_out_encoded.c_str(), cfg_.enc_out_length.c_str()};
+
+  auto run_options = OrtRunOptions::Create();
+
+  auto enc_outputs = model_.session_encoder_->Run(
```
kunal-vaishnavi: Same comment as [here](https://github.com/microsoft/onnxruntime-genai/pull/2150/changes#r3258446594)

src/models/parakeet.cpp
```diff
@@ -0,0 +1,486 @@
...
+    }
+    const float* dec_src = dec_.decoder_output->GetTensorData<float>();
+    std::memcpy(decoder_frame_->GetTensorMutableData<float>(), dec_src,
+                static_cast<size_t>(dec_dim) * sizeof(float));
+
+    const char* join_input_names[] = {cfg_.join_in_encoder.c_str(), cfg_.join_in_decoder.c_str()};
+    OrtValue* join_inputs[] = {encoder_frame_.get(), decoder_frame_.get()};
+    const char* join_output_names[] = {cfg_.join_out_logits.c_str()};
+
+    auto join_outputs = model_.session_joiner_->Run(
```
kunal-vaishnavi: Same comment as [here](https://github.com/microsoft/onnxruntime-genai/pull/2150/changes#r3258446594)

src/models/parakeet.cpp
```diff
@@ -0,0 +1,486 @@
...
+    int best_token = 0;
+    float best_score = logits_data[0];
+    for (int i = 1; i < num_tok_logits; ++i) {
+      if (logits_data[i] > best_score) {
+        best_score = logits_data[i];
+        best_token = i;
+      }
+    }
+
+    // Duration argmax. When the predicted token is blank, forbid duration
```
kunal-vaishnavi, line 471: There seems to be a lot of logits processing being done on these speech models. Can we find a way to generalize their techniques so that we can add new settings in the `search` section of the GenAI config to apply these processing steps on any speech model?

src/models/parakeet.cpp
```diff
@@ -0,0 +1,486 @@
...
+}
+
+void ParakeetTdtState::SetExtraInputs(const std::vector<ExtraInput>& extra_inputs) {
+  if (mel_loaded_) return;
+
+  // Locate the pre-computed mel-features tensor produced by ParakeetTdtProcessor.
+  // Expected shape: [1, num_mels, total_frames], already globally normalized.
+  const Tensor* mel_tensor = nullptr;
+  for (const auto& ei : extra_inputs) {
+    if (ei.name == "mel_features") {
```
kunal-vaishnavi: Can we re-use existing names such as `audio_features`?

src/models/parakeet.cpp
```diff
@@ -0,0 +1,486 @@
...
+  // when the audio is fully consumed).
+  if (!initialized_) {
+    InitializeDecoderState();
+    EncodeNextChunk();
+    initialized_ = true;
+  }
+
+  int32_t next_token = EmitNextToken();
+
+  // Encode the emitted token id as a one-hot logits row that the standard
```
kunal-vaishnavi: We shouldn't have to deal with copying the logits buffer across device memory. Let's find another way to add this.

src/config.h
```diff
@@ -346,6 +352,9 @@ struct Config {
         std::string targets;
         std::string lstm_hidden_state;
         std::string lstm_cell_state;
+
+        // Parakeet TDT decoder (prediction network) extra inputs
+        std::string targets_length;
```
kunal-vaishnavi, line 357: For Nemotron speech, we were able to remove the target length and output length from the input/output structure of the produced ONNX models. Can the same be done for Parakeet via a static-shaped exported model or is a dynamic-shaped exported model better?

nenad1002: Parakeet is not static, and removing it does not bring any benefit. If anything, I've found issues with removing this in Nemotron, and we'll probably need to reintroduce it back to Nemotron. But I'll have another PR for Nemotron.

src/models/parakeet.cpp
nenad1002: It's hard to generalize as every model might be doing something slightly different.
ASRs are not LLMs, and what's done in one model is different in another. There are many different types of decoders, and even within a single decoder many different ways to process.

src/models/parakeet.cpp
nenad1002: logits_buffer_ itself does not have to exist, that's true, so removing it, but we do need allocate logits_device_buffer.

Cause generator expects logits, Parakeet does not output pure logits like Nemotron or Whisper, but instead multidimensional array of [token_logits | duration_logits]. The alternative would be to slicing token logits out, but compute/memory wise would be the same.

src/models/parakeet.cpp
nenad1002: Ok, removed logits buffer

src/models/parakeet.cpp
nenad1002: I introduced TranducerState to unify some of Nemotron and Parakeet logic. But no new search param needed here.

examples/python/parakeet.py
nenad1002: We can delete both parakeet.py and nemotron_speech.py, and have one asr.py that can handle any model based on a selector? Even Whisper could be inluded.

Can do in the Cohere PR when we'll have more models to unify to keep this one focused on the runtime changes.

src/config.h
nenad1002: To clarify better. 

The Nemotron streaming encoder consumes a fixed-size chunk every step (always exactly chunk_samples mel frames), so length is constant and was safe to bake into the graph. However, this is what we found that empirically works well, not something that is correct, in fact it's very much wrong. This will not work for multilingual Nemotron cause it's not what Nemotron with NeMo does (which has first chunk different size). Also, the finetuned models perform much worse with static exports. 
This is why most probably we'll need to either:
1) Revert Nemotron to use only dynamic shape.
2) Keep Nemotron with static shape, but optionally enable dynamic, I prefer this option as it's backwards compatible, and gives us support for both dynamic and static exports.

Parakeet's encoder also supports variable per-chunk length. The first chunk has no left context, middle chunks carry full left + chunk + right, and the last chunk is whatever is left of the audio. That's three different time dimensions in a single utterance. To make the encoder static we'd have to pad every chunk to the max window and mask, that would most probably corrupt decoder as there is no good way to pad with no accuracy loss.

src/models/parakeet_processor.cpp
```diff
@@ -0,0 +1,102 @@
...
+
+  // 1. Decode audio file to float32 mono PCM at the model's sample rate with downmixing.
+  ort_extensions::OrtxObjectPtr<OrtxTensorResult> decoded;
+  CheckResult(OrtxDecodeAudio(audios->audios_.get(), 0,
+                              static_cast<int64_t>(m.sample_rate),
+                              /*stereo_to_mono=*/1,
+                              decoded.ToBeAssigned()));
+
+  OrtxTensor* pcm_tensor = nullptr;
+  CheckResult(OrtxTensorResultGetAt(decoded.get(), 0, &pcm_tensor));
```
kunal-vaishnavi: This appears to be a memory leak. Can you fix this and any other locations that follow this pattern? There was a [recent PR](https://github.com/microsoft/onnxruntime-genai/pull/2168) that fixes this issue.

baijumeswani (dismissed):

kunal-vaishnavi (dismissed):

baijumeswani (approved):

PR #2156, opened 2026-05-12
Add `kv_cache_fixed_to_max_length` search option (AMD RyzenAI)

akholodnamdcom:
## Summary

  Adds a new `kv_cache_fixed_to_max_length` boolean to the `search` section of `genai_config.json`. Some AMD RyzenAI models require the kv-cache to be allocated as a fixed buffer sized to `max_length`; this option marks the model as having that constraint so OGA can preserve the
  invariant at runtime.

  ## Behavior

  - When `kv_cache_fixed_to_max_length` is true, runtime overrides of `max_length` via `OgaGeneratorParamsSetSearchNumber` (and equivalents in C#/Python/Java/Objective-C) that would change the value are dropped with a yellow `warning` log entry, preserving the pre-sized buffer.
  - Same-value overrides are silent no-ops.
  - When the option is unset (default), behavior is unchanged.

  ## Why a warning instead of an error

  Throwing on every override would break clients that pass `max_length` through defensively (e.g. `model_benchmark`'s `--max_length` flag, which is set by default to `prompt_length + generation_length`). Logging a warning and ignoring the override keeps existing tools working
  while still surfacing that the value was not applied.

  ## Files changed

  - `src/config.h` — new `bool kv_cache_fixed_to_max_length{}` field at the end of `Config::Search`.
  - `src/config.cpp` — JSON parser branch for the new key (last in the chain, before the unknown-key throw), and a guard at the top of `SetSearchNumber` that warns and drops differing `max_length` overrides.

  ## Test plan

  - [x] `model_benchmark -ml 1024` with `kv_cache_fixed_to_max_length: true` and `max_length: 131072` in `genai_config.json` → warning printed, benchmark runs with `max_length=131072`.
  - [x] `model_benchmark -ml 131072` (matches config) → no warning, benchmark runs.
  - [x] `model_benchmark -ml -1` (no override call) → no warning, benchmark runs.
  - [x] Verified on Windows x64, RelWithDebInfo, CPU EP, with `Llama-3.2-1B-Instruct` (Q4F16).

  Sample warning output:

  ```
  [warning]  Ignoring max_length=1024 override: kv_cache_fixed_to_max_length is enabled in genai_config.json (kv-cache pinned to max_length=131072).
  ```

### DIFF
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -1343,9 +1343,11 @@ void SetSearchNumber(Config::Search& search, std::string_view name, double value
   if (name == "max_length" && search.kv_cache_fixed_to_max_length) {
     const int new_value = static_cast<int>(value);
     if (new_value != search.max_length) {
-      Log("warning", "Ignoring max_length=" + std::to_string(new_value) +
-                         " override: kv_cache_fixed_to_max_length is enabled in genai_config.json (kv-cache pinned to max_length=" +
-                         std::to_string(search.max_length) + ").");
+      if (g_log.enabled && g_log.warning) {
+        Log("warning", "Ignoring max_length=" + std::to_string(new_value) +
+                           " override: kv_cache_fixed_to_max_length is enabled in genai_config.json (kv-cache pinned to max_length=" +
+                           std::to_string(search.max_length) + ").");
+      }
     }
     return;
   }
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -1264,6 +1264,8 @@ struct Search_Element : JSON::Element {
       v_.early_stopping = JSON::Get<bool>(value);
     } else if (name == "blank_penalty") {
       v_.blank_penalty = static_cast<float>(JSON::Get<double>(value));
+    } else if (name == "kv_cache_fixed_to_max_length") {
+      v_.kv_cache_fixed_to_max_length = JSON::Get<bool>(value);
     } else {
       throw JSON::unknown_value_error{};
     }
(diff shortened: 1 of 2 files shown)


akholodnamdcom:
@baijumeswani @kunal-vaishnavi could you please take a look at the fix

baijumeswani:
This pull-request introduces a new config setting `kv_cache_fixed_to_max_length` that controls whether or not to use fixed shape for the kv-cache regardless of what the user may set with `max_length`. A few issues I think this introduces:

  - This seems like a temporary work-around. The longer-term solution might be to update the models to support dynamic kv-cache sizes. The problem with this temporary work-around is that it is user facing. Removing this setting in the future will be difficult.
  - It ignores the `max_length` set by the user. Moreover, it is prone to errors in case the user/application updates the value of `kv_cache_fixed_to_max_length` themselves.
  - It makes the settings more clunky.

Here are some alternative proposals:

1. Update the model to accept dynamic axes for the kv-cache. This might be a good solution if the fixed shape is not a hard requirement but a by-product of the maximum number of tokens the model can support. If the max number of tokens the model can support is fixed, you could consider making the context length = 4096 which would limit the number of tokens generated + input tokens to 4096.
2. Read the session information to derive the fixed length of the kv-cache from graph outputs. If the model has a fixed sequence length index, use that to make necessary buffer allocations instead of exposing the setting through the genai_config.json.

baijumeswani:
Closing in favor of https://github.com/microsoft/onnxruntime-genai/pull/2166

PR #2157, opened 2026-05-12
Add text-only mode support for Qwen 3.5 model builder

apsonawane:
## Description

Adds support for running Qwen 3.5 as a standalone text-only LLM (without vision/embedding pipelines).

### Changes

- **2D position_ids support**: When `exclude_embeds=false`, the builder creates a 2D `[B, S]` position_ids graph input and internally expands it to 3D `[3, B, S]` for mRoPE compatibility. This allows the standard onnxruntime-genai runtime to provide position_ids without requiring the multimodal pipeline.
- **Tokenizer regex fix**: Added `save_processing` override that patches unsupported `\p{M}` (Unicode Mark category) from tokenizer regex patterns after export. The C++ `std::regex` engine in onnxruntime-extensions does not support this Unicode property class.
- **genai_config correctness**: The internal 3D expanded tensor name is stored in a separate `_pos_ids_3d` attribute so `self.input_names["position_ids"]` remains as `"position_ids"` — ensuring genai_config.json references the actual graph input.

### Usage

```bash
python3 -m onnxruntime_genai.models.builder \
  --model_name Qwen/Qwen3.5-2B \
  --precision int4 \
  --execution_provider cuda \
  --output /path/to/output \
  --extra_options exclude_embeds=false prune_lm_head=true int4_algo_config=k_quant_last

### CODE REVIEW

File: src/python/py/models/builders/qwen.py [hunk 1]
```diff
@@ -1077,6 +1085,35 @@ def _setup_hybrid_cache_io(self):
...
+                unsq_name,
+                [pos_2d, "/model/constants/INT64/[0]"],
+                ir.DataType.INT64,
+                [1, "batch_size", "sequence_length"],
+            )
+            expand_name = "/model/position_ids_expand/Expand"
+            expand_output = f"{expand_name}/output_0"
+            self.make_expand(
+                expand_name,
+                [unsq_output, "/model/constants/INT64/[3, 1, 1]"],
```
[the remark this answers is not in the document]
apsonawane: This is an incorrect comment.

src/models/multi_modal.cpp
```diff
@@ -650,8 +650,13 @@ DecoderState::DecoderState(const MultiModalLanguageModel& model, DeviceSpan<int3
...
-  if (model_.session_info_.HasInput(model_.config_->model.decoder.inputs.input_ids)) {
+  // Some multimodal decoders (e.g., Gemma4) require input_ids alongside inputs_embeds.
+  // Use a decoder-only SessionInfo to avoid false positives from the embedding session
+  // (which always has input_ids), preventing incorrect injection into decoders that only
+  // accept inputs_embeds (e.g., Qwen3.5, Mistral3/Pixtral).
+  SessionInfo decoder_only_info;
+  decoder_only_info.Add(*model_.decoder_session_);
+  if (decoder_only_info.HasInput(model_.config_->model.decoder.inputs.input_ids)) {
     decoder_input_ids_ = std::make_unique<DefaultInputIDs>(*this);
     decoder_input_ids_->Add();
```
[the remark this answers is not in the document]
apsonawane: This concern doesn't apply — SessionInfo::Add() is only called during model initialization (constructor), not on the inference hot path. The cached session_info_ is already reused across all DecoderState instances

src/python/py/models/builders/qwen.py
```diff
@@ -1320,10 +1357,10 @@ def _get_shared_l2_eps(self):
     def _make_mrope_cos_sin(self, basename):
         """Build interleaved mRoPE cos/sin from pre-computed cache + position_ids.
 
-        Input: position_ids [3, B, S]
+        Input: position_ids [3, B, S] (from self._pos_ids_3d)
         Output: cos [B, S, rdim_half], sin [B, S, rdim_half]
         """
-        pos_ids = self.input_names["position_ids"]
+        pos_ids = self._pos_ids_3d
         cos_cache = "model.rotary_emb.cos_cache"
```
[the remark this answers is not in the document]
apsonawane: Fixed

src/python/py/models/builders/qwen.py
```diff
@@ -2012,3 +2051,27 @@ def make_genai_config(self, model_name_or_path, extra_kwargs, out_dir):
         del self.input_names["past_key_values.value"]
         del self.output_names["present.key"]
         del self.output_names["present.value"]
+
+    def save_processing(self, model_name_or_path, extra_kwargs, out_dir):
+        super().save_processing(model_name_or_path, extra_kwargs, out_dir)
+        # Patch tokenizer regex: remove \p{M} (Unicode Mark category).
+        # Qwen3.5 is the only model family whose tokenizer uses \p{M}.
+        # onnxruntime-extensions has hand-coded matchers for known regex
+        # sub-patterns (e.g. [\p{Lu}\p{Lt}\p{Lm}\p{Lo}\p{M}]) but Qwen3.5's
```
kunal-vaishnavi: We should make these changes in ORT Extensions to support the tokenizer instead of making these changes in ORT GenAI.

src/python/py/models/builders/qwen.py
```diff
@@ -926,6 +928,8 @@ class Qwen35TextModel(Model):
 
     def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
         # Qwen3.5 is a VL model. The decoder takes inputs_embeds.
+        # When exclude_embeds is explicitly set to False, build as a standalone LLM.
+        self.is_text_only = extra_options.get("exclude_embeds") is False
```
kunal-vaishnavi: Can you set the default value for `get()`?

https://github.com/microsoft/onnxruntime-genai/blob/a936152ae58fa23b8f5bf78b80af75ef4e8e8140/src/python/py/models/builders/base.py#L403

src/python/py/models/builders/qwen.py
```diff
@@ -969,8 +973,14 @@ def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
         # SkipSimplifiedLayerNormalization can be used directly.
         self.layernorm_attrs["add_offset"] = 1
 
-        # 3D position_ids for mRoPE: [3, batch_size, sequence_length]
-        self.input_shapes["position_ids"] = [3, "batch_size", "sequence_length"]
+        # Position IDs input.
+        # In text-only mode the runtime provides standard 2D [B, S] position_ids.
+        # We expand them to 3D [3, B, S] inside the graph so mRoPE works unchanged.
+        # In VL mode the pipeline provides 3D position_ids directly.
```
kunal-vaishnavi, line 977: Why do the position ids have to differ in text-only mode vs. vision-language mode? It should be an identical decoder besides the `input_ids` vs `inputs_embeds` input.

apsonawane: The difference exists because of who provides them, not the decoder itself.

1. VL mode: The multimodal pipeline computes 3D [3, B, S] position IDs externally — image patches need distinct spatial (H/W) positions vs text tokens (temporal only).
2. Text-only mode: The onnxruntime-genai runtime only generates standard 2D [B, S] position IDs. For pure text, all 3 mRoPE dimensions are identical (sequential), so we Tile [B, S] → [3, B, S] in the graph.

To eliminate the Tile, we'd need to:

1. Add ```qwen3_5_text``` to ```IsQwenVLFamily()``` in ```model_type.h```
2. Guard vision token config reads in ```Qwen2VLPositionInputs``` for the no-vision case in position_inputs.cpp
3. Remove the ```is_text_only``` branching in the Python builder

There is no perf regression, adding it in runtime will add code complexity I would say. Also, regarding above changes keeping text model out of VL model type makes sense to me.

src/python/py/models/builders/qwen.py
apsonawane: Updated created this PR in extensions: https://github.com/microsoft/onnxruntime-extensions/pull/1063

src/python/py/models/builders/qwen.py
apsonawane: Updated

src/python/py/models/builders/qwen.py
```diff
@@ -1077,6 +1085,35 @@ def _setup_hybrid_cache_io(self):
         self.output_names["present.key"] = filtered_key_outputs
         self.output_names["present.value"] = filtered_value_outputs
 
+    def make_inputs_and_outputs(self):
+        super().make_inputs_and_outputs()
+
+        if self.is_text_only:
+            # The graph input is 2D position_ids [B, S].
+            # Expand to 3D [3, B, S] for mRoPE by stacking 3 copies.
```
kunal-vaishnavi, line 1091: Can we move this logic to happen in an override of `make_position_ids_reformatting` instead since these changes are position ids related?

https://github.com/microsoft/onnxruntime-genai/blob/21192daaf005caa55d1c1ddd58efcf8dad5b6886/src/python/py/models/builders/base.py#L4744-L4747

kunal-vaishnavi (approved):

PR #2161, opened 2026-05-13
Limit the CUDA cmake architectures to 86 for CI builds

baijumeswani:


kunal-vaishnavi (dismissed):

test/python/requirements.txt
```diff
@@ -8,4 +8,3 @@ pytest
 onnx
 onnx_ir>=0.1.3
 transformers<5.0.0
-huggingface_hub[cli,hf_xet]
```
kunal-vaishnavi, line 11: Why remove this? We use the `cli` scope to download models from the Hugging Face Hub and the `hf_xet` scope to download large files.

baijumeswani: We don't download any models/files in the CI pipelines. We use models already embedded in the vm image.

The huggingface_hub requirement here was causing problems because of transformers<5.0.0 requirement above. This caused the python package manager to look repeatedly for a correct version and was causing installation issues.

baijumeswani:
Closing-reopening to trigger all CIs again.

kunal-vaishnavi (approved):

baijumeswani:
Thanks for the review. :)

PR #2162, opened 2026-05-13
Gate leaked-object error reporting in Shutdown() to debug builds or when logging is enabled

baijumeswani:
`Generators::Shutdown()` walks the LeakChecked<T> counters and prints messages of the form:

```
OGA Error: N instances of <type> were leaked.
    Please see the documentation for the API being used to ensure proper cleanup.
```

to stderr whenever any tracked object is still alive at process exit.

FoundryLocal offers a model.unload. But if users do not unload the model, they will run into this std::cerr error on their console.

With this pull-request, the leak dump is now gated:

- Debug builds (NDEBUG not defined): leaks are reported as before
- Release builds: leaks are reported only when logging is enabled.

### DIFF
--- a/src/generators.cpp
+++ b/src/generators.cpp
@@ -87,11 +87,28 @@ GetOrtGlobals() {
   return globals;
 }
 
-// Used by Shutdown() to display the counts and types of any leaked objects
+// Used by Shutdown() to display the counts and types of any leaked objects.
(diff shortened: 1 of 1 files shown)


kunal-vaishnavi (approved):

baijumeswani:
Thanks for the review. :)

PR #2165, opened 2026-05-14
Add AMDGPU execution provider

aditya-dl:
This PR adds the AMDGPU execution provider to ONNX Runtime GenAI. The AMDGPU EP picks a concrete backend at runtime and OGA drives it as a single `DeviceType::AMDGPU`, so callers never manage the backend directly. It supersedes the earlier MIGraphX-only framing (#2093 and the previous version of this PR).

Users select it as `"amdgpu"`. We also accept `"AMDGPUExecutionProvider"`, the catalog name the AMD Windows ML EP MSIX advertises, so harnesses that match against WinML-discovered names work without special-casing. Both normalize to `"AMDGPU"`.

## What's in the three commits

**1. AMDGPU EP with GPU-resident KV cache.** Adds the EP wiring (`DeviceType::AMDGPU`, the dispatch-table entry, provider-name normalization, and the `src/amdgpu/` sources) plus a GPU-resident `DeviceInterface` that keeps the KV cache on the device, removing the per-token CPU/GPU roundtrip. Cross-device copies go through backend-agnostic opaque `DeviceBuffer` handles. It also emits the static-shape prefill config (`ep.migraphx.static_pad_*` and `hip_graph_enable`) so the backend compiles the prefill once and reuses a captured graph.

**2. Host-accessible decode inputs.** The small decode inputs (input_ids, position_ids, attention_mask) are allocated from a host-accessible allocator that the CPU can write and the GPU can read, so the CPU updates them in place with no per-step copy. The KV cache and scoring stay on the default device interface. If no host-accessible allocator is available, it falls back to the default input path.

**3. Logits on CPU.** Logits are GPU-written and CPU-read, the opposite of the decode inputs. On AMDGPU those inputs live in a host-accessible heap that is not coherent for CPU reads, so reading logits from it would return stale data. This commit routes logits to the CPU interface through a new `p_logits_` member, leaving only the decode inputs pinned.

## Configuration

```json
"provider_options": [{ "amdgpu": {} }]
```

or

```python
config.append_provider("amdgpu")
```

`"AMDGPUExecutionProvider"` also works (the catalog form).

## Design notes

Graph capture is always on for AMDGPU, so OGA sizes the attention mask and KV cache itself. The EP is told to pad the prefill token axis through the static-pad session-config entries, while OGA keeps ownership of mask and KV sizing. The host-accessible and logits routing are both backend-aware and fall back safely to the default path when the allocator isn't there.

## Known limitations

Beam search isn't supported (it needs `past_present_share_buffer=true`, which needs `num_beams=1`). The host-accessible allocator currently assumes `device_id=0`, so multi-GPU is a follow-up.

### DIFF
--- a/src/models/kv_cache.cpp
+++ b/src/models/kv_cache.cpp
@@ -432,7 +432,10 @@ DefaultKeyValueCache::DefaultKeyValueCache(State& state)
       }
 
       presents_.push_back(OrtValue::CreateTensor(Allocator(), tensor_shape, type_));
-      if (Device().GetType() != DeviceType::WEBGPU) {
+      // WebGPU has no Zero() implementation; AMDGPU skips it for the same reason (Stage A
+      // DeviceInterface throws on Zero — KV is fresh per-Generator so zeroing is optional).
+      if (Device().GetType() != DeviceType::WEBGPU &&
+          Device().GetType() != DeviceType::AMDGPU) {
         ByteWrapTensor(Device(), *presents_.back()).Zero();
       }
     }
--- a/src/models/kv_cache.cpp
+++ b/src/models/kv_cache.cpp
@@ -432,10 +432,8 @@ DefaultKeyValueCache::DefaultKeyValueCache(State& state)
       }
 
       presents_.push_back(OrtValue::CreateTensor(Allocator(), tensor_shape, type_));
-      // WebGPU has no Zero() implementation; AMDGPU skips it for the same reason (Stage A
-      // DeviceInterface throws on Zero — KV is fresh per-Generator so zeroing is optional).
-      if (Device().GetType() != DeviceType::WEBGPU &&
-          Device().GetType() != DeviceType::AMDGPU) {
+      // WebGPU has no Zero() implementation; every other backend (incl. AMDGPU) zero-inits the KV.
+      if (Device().GetType() != DeviceType::WEBGPU) {
         ByteWrapTensor(Device(), *presents_.back()).Zero();
       }
     }
--- a/src/models/kv_cache.cpp
+++ b/src/models/kv_cache.cpp
@@ -517,10 +517,8 @@ DefaultKeyValueCache::DefaultKeyValueCache(State& state)
       }
 
       presents_.push_back(OrtValue::CreateTensor(Allocator(), tensor_shape, type_));
-      // WebGPU has no Zero() implementation; AMDGPU skips it for the same reason (Stage A
-      // DeviceInterface throws on Zero — KV is fresh per-Generator so zeroing is optional).
-      if (Device().GetType() != DeviceType::WEBGPU &&
-          Device().GetType() != DeviceType::AMDGPU) {
+      // WebGPU has no Zero() implementation; every other backend (incl. AMDGPU) zero-inits the KV.
+      if (Device().GetType() != DeviceType::WEBGPU) {
         ByteWrapTensor(Device(), *presents_.back()).Zero();
       }
     }
--- a/src/models/kv_cache.cpp
+++ b/src/models/kv_cache.cpp
@@ -517,7 +517,10 @@ DefaultKeyValueCache::DefaultKeyValueCache(State& state)
       }
 
       presents_.push_back(OrtValue::CreateTensor(Allocator(), tensor_shape, type_));
-      if (Device().GetType() != DeviceType::WEBGPU) {
+      // WebGPU has no Zero() implementation; AMDGPU skips it for the same reason (Stage A
+      // DeviceInterface throws on Zero — KV is fresh per-Generator so zeroing is optional).
+      if (Device().GetType() != DeviceType::WEBGPU &&
+          Device().GetType() != DeviceType::AMDGPU) {
         ByteWrapTensor(Device(), *presents_.back()).Zero();
       }
     }
--- a/src/models/kv_cache.cpp
+++ b/src/models/kv_cache.cpp
@@ -517,7 +517,7 @@ DefaultKeyValueCache::DefaultKeyValueCache(State& state)
       }
 
       presents_.push_back(OrtValue::CreateTensor(Allocator(), tensor_shape, type_));
-      // WebGPU has no Zero() implementation; every other backend (incl. AMDGPU) zero-inits the KV.
+      // WebGPU has no Zero() implementation; every other backend zero-inits the KV.
       if (Device().GetType() != DeviceType::WEBGPU) {
         ByteWrapTensor(Device(), *presents_.back()).Zero();
       }
--- a/src/models/logits.cpp
+++ b/src/models/logits.cpp
@@ -13,7 +13,13 @@ Logits::Logits(State& state)
     : state_{state},
       shape_{static_cast<int64_t>(state_.params_->BatchBeamSize()), 0, model_.config_->model.vocab_size},
       type_{model_.session_info_.GetOutputDataType(model_.config_->model.decoder.outputs.logits)} {
-  output_raw_ = std::make_unique<Tensor>(model_.p_device_inputs_, type_);
+  // Logits is GPU-written / CPU-read (the sampler). For AMDGPU, p_device_inputs_ points at the
+  // host-accessible allocator (DML WRITE_COMBINE is not CPU-read-coherent), so route logits to the
+  // CPU interface instead; other backends keep logits on their device-inputs interface.
+  p_logits_ = (model_.p_device_->GetType() == DeviceType::AMDGPU)
+                  ? GetDeviceInterface(DeviceType::CPU)
+                  : model_.p_device_inputs_;
+  output_raw_ = std::make_unique<Tensor>(p_logits_, type_);
 
   input_sequence_lengths.resize(state_.params_->search.batch_size);
 
@@ -42,10 +48,10 @@ DeviceSpan<float> Logits::Get() {
     const size_t num_beams = state_.params_->search.num_beams;
 
     // create new OrtValue for logits_of_last_token and use output_last_tokens_ to hold it
-    output_last_tokens_ = OrtValue::CreateTensor(model_.p_device_inputs_->GetAllocator(), shape_last, type_);
+    output_last_tokens_ = OrtValue::CreateTensor(p_logits_->GetAllocator(), shape_last, type_);
 
     if (type_ == Ort::TypeToTensorType<Ort::Float16_t> || type_ == Ort::TypeToTensorType<Ort::BFloat16_t>)
-      logits_of_last_token_fp32_ = OrtValue::CreateTensor<float>(model_.p_device_inputs_->GetAllocator(), shape_);
+      logits_of_last_token_fp32_ = OrtValue::CreateTensor<float>(p_logits_->GetAllocator(), shape_);
 
     logits_of_last_token = output_last_tokens_.get();
 
@@ -53,7 +59,7 @@ DeviceSpan<float> Logits::Get() {
     size_t vocab_index = 0;  // Simpler math to have this index go up by vocab_size for every logit chunk we process
 
     auto logits_raw = output_raw_->GetByteSpan();
-    auto logits_last_tokens = ByteWrapTensor(*model_.p_device_inputs_, *logits_of_last_token);
+    auto logits_last_tokens = ByteWrapTensor(*p_logits_, *logits_of_last_token);
 
     for (int batch_index = 0; batch_index < state_.params_->search.batch_size; batch_index++) {
       // Find the first non pad token from the end
@@ -71,12 +77,12 @@ DeviceSpan<float> Logits::Get() {
 
   // Convert from float16/bfloat16 to float32 if necessary
   if (type_ == Ort::TypeToTensorType<Ort::Float16_t> || type_ == Ort::TypeToTensorType<Ort::BFloat16_t>) {
-    Cast(*logits_of_last_token, logits_of_last_token_fp32_, *model_.p_device_inputs_, Ort::TypeToTensorType<float>);
+    Cast(*logits_of_last_token, logits_of_last_token_fp32_, *p_logits_, Ort::TypeToTensorType<float>);
     logits_of_last_token = logits_of_last_token_fp32_.get();
   }
 
   if (logits_.empty() || logits_of_last_token->GetTensorMutableRawData() != logits_.Span().data())
-    logits_ = WrapTensor<float>(*model_.p_device_inputs_, *logits_of_last_token);
+    logits_ = WrapTensor<float>(*p_logits_, *logits_of_last_token);
 
   return logits_;
 }
--- a/src/models/logits.cpp
+++ b/src/models/logits.cpp
@@ -13,13 +13,7 @@ Logits::Logits(State& state)
     : state_{state},
       shape_{static_cast<int64_t>(state_.params_->BatchBeamSize()), 0, model_.config_->model.vocab_size},
       type_{model_.session_info_.GetOutputDataType(model_.config_->model.decoder.outputs.logits)} {
-  // Logits is GPU-written / CPU-read (the sampler). For AMDGPU, p_device_inputs_ points at the
-  // host-accessible allocator (DML WRITE_COMBINE is not CPU-read-coherent), so route logits to the
-  // CPU interface instead; other backends keep logits on their device-inputs interface.
-  p_logits_ = (model_.p_device_->GetType() == DeviceType::AMDGPU)
-                  ? GetDeviceInterface(DeviceType::CPU)
-                  : model_.p_device_inputs_;
-  output_raw_ = std::make_unique<Tensor>(p_logits_, type_);
+  output_raw_ = std::make_unique<Tensor>(model_.p_device_logits_, type_);
 
   input_sequence_lengths.resize(state_.params_->search.batch_size);
 
@@ -48,10 +42,10 @@ DeviceSpan<float> Logits::Get() {
     const size_t num_beams = state_.params_->search.num_beams;
 
     // create new OrtValue for logits_of_last_token and use output_last_tokens_ to hold it
-    output_last_tokens_ = OrtValue::CreateTensor(p_logits_->GetAllocator(), shape_last, type_);
+    output_last_tokens_ = OrtValue::CreateTensor(model_.p_device_logits_->GetAllocator(), shape_last, type_);
 
     if (type_ == Ort::TypeToTensorType<Ort::Float16_t> || type_ == Ort::TypeToTensorType<Ort::BFloat16_t>)
-      logits_of_last_token_fp32_ = OrtValue::CreateTensor<float>(p_logits_->GetAllocator(), shape_);
+      logits_of_last_token_fp32_ = OrtValue::CreateTensor<float>(model_.p_device_logits_->GetAllocator(), shape_);
 
     logits_of_last_token = output_last_tokens_.get();
 
@@ -59,7 +53,7 @@ DeviceSpan<float> Logits::Get() {
     size_t vocab_index = 0;  // Simpler math to have this index go up by vocab_size for every logit chunk we process
 
     auto logits_raw = output_raw_->GetByteSpan();
-    auto logits_last_tokens = ByteWrapTensor(*p_logits_, *logits_of_last_token);
+    auto logits_last_tokens = ByteWrapTensor(*model_.p_device_logits_, *logits_of_last_token);
 
     for (int batch_index = 0; batch_index < state_.params_->search.batch_size; batch_index++) {
       // Find the first non pad token from the end
@@ -77,12 +71,12 @@ DeviceSpan<float> Logits::Get() {
 
   // Convert from float16/bfloat16 to float32 if necessary
   if (type_ == Ort::TypeToTensorType<Ort::Float16_t> || type_ == Ort::TypeToTensorType<Ort::BFloat16_t>) {
-    Cast(*logits_of_last_token, logits_of_last_token_fp32_, *p_logits_, Ort::TypeToTensorType<float>);
+    Cast(*logits_of_last_token, logits_of_last_token_fp32_, *model_.p_device_logits_, Ort::TypeToTensorType<float>);
     logits_of_last_token = logits_of_last_token_fp32_.get();
   }
 
   if (logits_.empty() || logits_of_last_token->GetTensorMutableRawData() != logits_.Span().data())
-    logits_ = WrapTensor<float>(*p_logits_, *logits_of_last_token);
+    logits_ = WrapTensor<float>(*model_.p_device_logits_, *logits_of_last_token);
 
   return logits_;
 }
--- a/src/models/logits.h
+++ b/src/models/logits.h
@@ -31,12 +31,6 @@ struct Logits {
 
   std::unique_ptr<Tensor> output_raw_;  // Raw logits output from model
 
-  // Device interface for the logits output + last-token/fp32 scratch. Normally p_device_inputs_, but
-  // logits is GPU-written / CPU-read (opposite of the pinned decode inputs). For AMDGPU the inputs use
-  // a host-accessible allocator whose heap is NOT CPU-read-coherent (DML CUSTOM/L0/WRITE_COMBINE), so
-  // logits go on the CPU interface instead; only the 3 decode inputs use the pinned allocator.
-  DeviceInterface* p_logits_{};
-
   std::vector<int> input_sequence_lengths;
   // OrtValue wrapped in a DeviceMemory object to make it universal
   DeviceSpan<float> logits_;
--- a/src/models/logits.h
+++ b/src/models/logits.h
@@ -31,6 +31,12 @@ struct Logits {
 
   std::unique_ptr<Tensor> output_raw_;  // Raw logits output from model
 
+  // Device interface for the logits output + last-token/fp32 scratch. Normally p_device_inputs_, but
+  // logits is GPU-written / CPU-read (opposite of the pinned decode inputs). For AMDGPU the inputs use
+  // a host-accessible allocator whose heap is NOT CPU-read-coherent (DML CUSTOM/L0/WRITE_COMBINE), so
+  // logits go on the CPU interface instead; only the 3 decode inputs use the pinned allocator.
+  DeviceInterface* p_logits_{};
+
   std::vector<int> input_sequence_lengths;
   // OrtValue wrapped in a DeviceMemory object to make it universal
   DeviceSpan<float> logits_;
--- a/src/models/model.cpp
+++ b/src/models/model.cpp
@@ -508,13 +508,19 @@ void EnsureDeviceOrtInit(DeviceInterface& device, const Config& config) {
   // if unavailable, creation returns null and callers fall back to the default device inputs path.
   if (!allocator.host_accessible_allocator_ && type == DeviceType::AMDGPU) {
     try {
-      // device_id is hardcoded 0, so this is correct for single-GPU only. On multi-GPU the
-      // device_id differs and GetSharedAllocator returns null (falls back, perf loss not error).
-      // TODO(multi-gpu): query EpDevice::MemoryInfo(HOST_ACCESSIBLE) instead of reconstructing.
-      auto host_info = OrtMemoryInfo::CreateV2("pinned", OrtMemoryInfoDeviceType_GPU,
-                                               /*vendor_id=*/0x1002, /*device_id=*/0,
-                                               OrtDeviceMemoryType_HOST_ACCESSIBLE, OrtDeviceAllocator);
-      allocator.host_accessible_allocator_ = GetOrtEnv().GetSharedAllocator(*host_info);
+      // Query the AMDGPU EP's advertised HOST_ACCESSIBLE memory-info instead of reconstructing it with
+      // hardcoded vendor/device ids, so it carries the real ids for this machine (multi-GPU safe). If
+      // none is advertised, leave it unset -> callers fall back to the default device inputs path.
+      const OrtMemoryInfo* host_info = nullptr;
+      for (const OrtEpDevice* ep_device : FindRegisteredEpDevices("AMDGPUExecutionProvider")) {
+        if (const OrtMemoryInfo* mi =
+                Ort::api->EpDevice_MemoryInfo(ep_device, OrtDeviceMemoryType_HOST_ACCESSIBLE)) {
+          host_info = mi;
+          break;
+        }
+      }
+      if (host_info)
+        allocator.host_accessible_allocator_ = GetOrtEnv().GetSharedAllocator(*host_info);
     } catch (const Ort::Exception&) {
       allocator.host_accessible_allocator_ = nullptr;
     }
--- a/src/models/model.cpp
+++ b/src/models/model.cpp
@@ -623,6 +623,9 @@ Model::Model(std::unique_ptr<Config> config) : config_{std::move(config)} {
       p_device_inputs_ = pinned;
   }
 
+  // Logits are CPU-read; AMDGPU host-accessible inputs aren't CPU-read-coherent, so route to CPU.
+  p_device_logits_ = (p_device_->GetType() == DeviceType::AMDGPU) ? GetDeviceInterface(DeviceType::CPU) : p_device_inputs_;
+
   // Search and sampling are performed on the CPU for all device types,
   // except for CUDA and NvTensorRtRtx, where this is performed on the device.
   if (p_device_->GetType() == DeviceType::CUDA ||
--- a/src/models/model.cpp
+++ b/src/models/model.cpp
@@ -467,18 +467,25 @@ void EnsureDeviceOrtInit(DeviceInterface& device, const Config& config) {
   const auto trivial_model = GetTrivialModel();
   allocator.session_ = OrtSession::Create(GetOrtEnv(), trivial_model.data(), trivial_model.size(), session_options.get());
 
-  // Names for the device memory types used by 'OrtMemoryInfo::Create'
-  // AMDGPU device memory-type name is "Hip" (the MIGraphX backend registers
-  // CreateMemoryInfo_V2("Hip", GPU, ...); a V1 CreateMemoryInfo("Hip") memory-info matches it
-  // by device attributes). This mirrors the original B003 GPU-KV path (commit 64d1b3a5).
+  // Names for the device memory types used by 'OrtMemoryInfo::Create'. AMDGPU is "Hip".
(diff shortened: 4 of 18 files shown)


aditya-dl:
@kunal-vaishnavi @baijumeswani could you help review this PR?

src/models/position_inputs.h
```diff
@@ -30,7 +30,7 @@ struct DefaultPositionInputs : PositionInputs {
   template <typename T>
   void InitializeSequenceLengths(std::array<int64_t, 2> shape, cpu_span<int32_t> sequence_lengths_unk);
   template <typename T>
-  void CreateAndInitializePositionIDs(DeviceSpan<int32_t> next_tokens, std::array<int64_t, 2> shape);
+  void CreateAndInitializePositionIDs(DeviceSpan<int32_t> next_tokens, std::array<int64_t, 2> shape, int seq_length);
   template <typename T>
   void CreateAndInitializeAttentionMask(DeviceSpan<int32_t> next_tokens, std::array<int64_t, 2> shape);
```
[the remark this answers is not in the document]
aditya-dl: The template declaration is on line 32, immediately above this one (`template <typename T>`) - the templating is intact. This compiles cleanly.

src/models/model.cpp
```diff
@@ -154,9 +154,52 @@ void State::Run(OrtSession& session, bool graph_capture_this_run) {
     run_options_->AddConfigEntry("disable_synchronize_execution_providers", "1");
   }
 
+  std::unique_ptr<OrtValue> new_input_ids;
+
+  if (prompt_gen_) {
+    for (int i = 0; i < inputs_.size(); i++) {
+      std::string input_name = input_names_[i];
+
+      if (input_name == "input_ids") {
```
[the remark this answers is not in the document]
aditya-dl: Good catch. Fixed in the latest push: the original `OrtValue*` and its slot index are saved before overwrite, and `inputs_[i]` is restored before `new_input_ids.reset()`, so the slot no longer dangles after `session.Run()`.

src/models/model.cpp
```diff
@@ -154,9 +154,52 @@ void State::Run(OrtSession& session, bool graph_capture_this_run) {
...
+        int64_t batch_size = params_->search.batch_size;
+        int64_t padded_seq_len = static_cast<int64_t>(params_->search.max_length);
+        std::vector<int64_t> new_shape{batch_size, padded_seq_len};
+
+        OrtValue* value = inputs_[i];
+        auto info = value->GetTensorTypeAndShapeInfo();
+        ONNXTensorElementDataType elem_type = info->GetElementType();
+        std::vector<int64_t> dims = info->GetShape();
```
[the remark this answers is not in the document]
aditya-dl: Fixed in the latest push: switched from `params_->search.batch_size` to `dims[0]` so the copy bounds are always pinned to the source tensor's actual batch dimension.

src/models/position_inputs.cpp
```diff
@@ -252,28 +256,34 @@ void DefaultPositionInputs::UpdateAttentionMask(int total_length, int new_kv_len
...
+  std::fill(position_data, position_data + shape[0] * shape[1], T{0});
+
   // If batch_size is 1 we have no padding, so we do simple ascending
   if (shape[0] == 1) {
-    for (int i = 0; i < shape[1]; ++i) {
+    for (int i = 0; i < seq_length; ++i) {
       position_data[i] = static_cast<T>(i);
     }
-    position_data_next[0] = static_cast<T>(shape[1]) - 1;
+    position_data_next[0] = static_cast<T>(seq_length) - 1;
```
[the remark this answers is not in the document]
aditya-dl: Acknowledged. In practice `seq_length == 0` cannot reach this function: `Generator::AppendTokens` rejects empty `input_ids` with a `runtime_error` upstream (`generators.cpp`). Happy to add a defensive `if (seq_length == 0) return;` here if you'd prefer belt-and-suspenders.

src/models/position_inputs.cpp
```diff
@@ -252,28 +256,34 @@ void DefaultPositionInputs::UpdateAttentionMask(int total_length, int new_kv_len
...
     // Otherwise we iterate backwards as to not misinterpret any right pad tokens
   } else {
-    const auto* word_id = const_cast<DeviceSpan<int32_t>&>(next_tokens).CpuSpan().data() + shape[0] * shape[1] - 1;
-    auto* position = position_data + shape[0] * shape[1] - 1;
+    // Use seq_length to index into next_tokens safely (it may be shorter than shape[1])
+    const auto* word_id = const_cast<DeviceSpan<int32_t>&>(next_tokens).CpuSpan().data() + shape[0] * seq_length - 1;
```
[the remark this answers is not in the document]
aditya-dl: Noted. The `const_cast` was inherited from the original code on this line (same shape, only the operands changed for the `seq_length` fix). Happy to clean it up as a follow-up - flagging here so as not to widen this PR's scope.

src/models/model.cpp
```diff
@@ -154,9 +154,52 @@ void State::Run(OrtSession& session, bool graph_capture_this_run) {
...
+  if (prompt_gen_) {
+    for (int i = 0; i < inputs_.size(); i++) {
+      std::string input_name = input_names_[i];
+
+      if (input_name == "input_ids") {
+        int64_t batch_size = params_->search.batch_size;
+        int64_t padded_seq_len = static_cast<int64_t>(params_->search.max_length);
+        std::vector<int64_t> new_shape{batch_size, padded_seq_len};
+
+        OrtValue* value = inputs_[i];
```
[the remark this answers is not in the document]
aditya-dl: Same issue as the parent thread on line 167 - addressed in the latest push (commit `d5638019`). The original `OrtValue*` is now saved before overwrite and restored before `new_input_ids.reset()`.

src/models/model.cpp
```diff
@@ -154,9 +154,52 @@ void State::Run(OrtSession& session, bool graph_capture_this_run) {
...
+        auto info = value->GetTensorTypeAndShapeInfo();
+        ONNXTensorElementDataType elem_type = info->GetElementType();
+        std::vector<int64_t> dims = info->GetShape();
+        int64_t orig_seq_len = std::min(dims[1], padded_seq_len);
+
+        new_input_ids = OrtValue::CreateTensor(model_.allocator_cpu_, new_shape, elem_type);
```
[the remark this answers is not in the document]
aditya-dl: Same issue as the parent thread on line 167 - addressed in the latest push (commit `d5638019`). The original `OrtValue*` is now saved before overwrite and restored before `new_input_ids.reset()`.

src/models/model.cpp
```diff
@@ -154,9 +154,52 @@ void State::Run(OrtSession& session, bool graph_capture_this_run) {
...
+        }
+        inputs_[i] = new_input_ids.get();
+        break;
+      }
+    }
+  }
   session.Run(run_options_.get(), input_names_.data(), inputs_.data(), input_names_.size(),
               output_names_.data(), outputs_.data(), output_names_.size());
 
+  new_input_ids.reset();
```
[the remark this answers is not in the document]
aditya-dl: Same issue as the parent thread on line 167 - addressed in the latest push (commit `d5638019`). The original `OrtValue*` is now saved before overwrite and restored before `new_input_ids.reset()`.

src/models/model.cpp
```diff
@@ -154,9 +154,52 @@ void State::Run(OrtSession& session, bool graph_capture_this_run) {
...
+        new_input_ids = OrtValue::CreateTensor(model_.allocator_cpu_, new_shape, elem_type);
+
+        if (elem_type == ONNX_TENSOR_ELEMENT_DATA_TYPE_INT64) {
+          auto* src = value->GetTensorMutableData<int64_t>();
+          auto* dst = new_input_ids->GetTensorMutableData<int64_t>();
+          std::fill(dst, dst + batch_size * padded_seq_len, int64_t{0});
+          for (int64_t b = 0; b < batch_size; ++b) {
+            std::copy(src + b * orig_seq_len, src + b * orig_seq_len + orig_seq_len,
+                      dst + b * padded_seq_len);
+          }
```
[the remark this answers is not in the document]
aditya-dl: Same issue as the parent thread on line 171 - addressed in the latest push (commit `d5638019`). The copy loop now uses `dims[0]` instead of `params_->search.batch_size`.

src/models/position_inputs.cpp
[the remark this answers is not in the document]
aditya-dl: Defensive guard added in the latest push (commit `d5638019`). When `seq_length <= 0` we now skip both per-token population branches and just leave the zero-filled tensor in place, with `position_data_next[0] = 0`. The pointer arithmetic that previously could form `buffer + ... - 1` is no longer reached in that case.

src/models/position_inputs.cpp
```diff
@@ -252,28 +256,34 @@ void DefaultPositionInputs::UpdateAttentionMask(int total_length, int new_kv_len
...
     bool found_first_non_pad = false;
     for (int i = static_cast<int>(shape[0] - 1); i >= 0; i--) {
-      T abs_position = static_cast<T>(shape[1] - 1);
+      T abs_position = static_cast<T>(seq_length - 1);
       found_first_non_pad = false;
-      for (int j = static_cast<int>(shape[1] - 1); j >= 0; j--, word_id--, position--) {
+      auto* position = position_data + i * shape[1] + seq_length - 1;
+      for (int j = seq_length - 1; j >= 0; j--, word_id--, position--) {
```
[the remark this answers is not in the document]
aditya-dl: Same as the adjacent thread - guarded by the `seq_length <= 0` early branch added in commit `d5638019`.

src/models/logits.cpp
```diff
@@ -107,6 +107,9 @@ void Logits::Update(const DeviceSpan<int32_t>& next_tokens, size_t new_kv_length
   }
 
   shape_[1] = new_kv_length;
+  if (state_.prompt_gen_) {
```
kunal-vaishnavi: Continuing the conversation from [here](https://github.com/microsoft/onnxruntime-genai/pull/2093#discussion_r3118880224), I think it would be good to have better architectural separation. This feels hacky to me.

src/models/logits.cpp
```diff
@@ -107,6 +107,9 @@ void Logits::Update(const DeviceSpan<int32_t>& next_tokens, size_t new_kv_length
...
+    shape_[1] = state_.params_->search.max_length;
+  }
```
kunal-vaishnavi: Continuing the conversation from [here](https://github.com/microsoft/onnxruntime-genai/pull/2093#discussion_r3118866947), why can't the MIGraphX EP handle static shapes through its own mechanism?

src/models/model.cpp
```diff
@@ -156,9 +154,65 @@ void State::Run(OrtSession& session, bool graph_capture_this_run) {
     run_options_->AddConfigEntry("disable_synchronize_execution_providers", "1");
   }
 
+  std::unique_ptr<OrtValue> new_input_ids;
+  // Track the original input_ids pointer + slot so we can restore it after
+  // session.Run() — otherwise inputs_[i] would dangle once new_input_ids is reset.
+  OrtValue* original_input_ids = nullptr;
+  int input_ids_idx = -1;
+
+  if (prompt_gen_) {
```
kunal-vaishnavi: We should find a way to design this so that we don't need to introduce more `if A, then B, else C` logic for EP-specific scenarios (even if this means making an architectural change for handling how inputs are updated).

src/migraphx/session_options.cpp
```diff
@@ -0,0 +1,34 @@
...
+  // V2 path: MSIX (and the future plugin EP) registers as "MIGraphXExecutionProvider".
+  if (!AppendExecutionProviderV2(session_options, provider_options,
+                                 DeviceType::CPU, "MIGraphXExecutionProvider")) {
+    // V1 fallback: legacy ORT shared provider
+    std::vector<const char*> keys, values;
+    for (auto& option : provider_options.options) {
+      keys.emplace_back(option.first.c_str());
+      values.emplace_back(option.second.c_str());
+    }
+    session_options.AppendExecutionProvider("MIGraphX", keys.data(), values.data(), keys.size());
```
kunal-vaishnavi: Continuing the conversation from [here](https://github.com/microsoft/onnxruntime-genai/pull/2093#discussion_r3139459861), why is this done?

src/models/logits.cpp
rlegithub: Hi @kunal-vaishnavi, just wanted to send our ACK on your comments. Currently Aditya is OoO for July 4 long weekend break. He will address your comments as soon as he is back on July 6. Thank you. Roto Le/AMD

aditya-dl:
Hi @kunal-vaishnavi. Apologies for the long delay getting back to you. In the meantime, we've been reworking the architecture of the AMD GPU integration, and your comments on this PR shaped that direction. 
The short version: **we're dropping MIGraphX as a user-facing execution provider.** Going forward, the **umbrella "AMDGPU" EP is the single user-facing EP for all AMD GPUs**; it selects the backend internally, so users target one EP regardless of the underlying stack. 

This resolves the concerns you raised here:
- **"Why can't the MIGraphX EP handle static shapes through its own mechanism?" / the `logits.cpp` hackiness**: Agreed, and that's now the design. Static input-shape padding has moved **out of OGA's shared paths and into the EP** (the EP pads/slices internally and reuses the compiled program). OGA no longer special-cases it.
- **"Don't introduce EP-specific if/else logic for how inputs are updated"**: Removed. The umbrella EP owns the backend-specific behavior; OGA drives it through generic provider options instead of branching on the EP in shared code. The `NeedsStaticInputShapes` special-case goes away.
- **`session_options.cpp` "why is this done?"**: Superseded by the umbrella architecture; it won't carry forward in its current form.

**We'll push the new-architecture updates to this same PR** rather than opening a new one retitling/rescoping it to the umbrella EP as we go. Thanks for the patience and for the feedback that pushed us here.

src/models/logits.cpp
aditya-dl: Agreed. This is removed in the umbrella EP, handles static shapes internally, so this OGA-side handling goes away.

src/models/logits.cpp
aditya-dl: That's exactly the new design, static-shape padding now lives in the EP, not OGA.

src/models/model.cpp
aditya-dl: Removed. Under the umbrella EP the backend-specific behavior lives in the EP and OGA drives it via generic provider options, no EP branching in shared input-update code.

src/migraphx/session_options.cpp
aditya-dl: Superseded by the umbrella architecture; this handling won't carry forward in its current form.

aditya-dl:
Hi @kunal-vaishnavi. The new-architecture updates are now pushed to this PR (three commits on top of latest main), and I've updated the description to match. The EP-specific if/else logic and the logits.cpp hackiness from the previous version are both gone. Whenever you have a chance, could you take another look? Happy to walk through any of it. Thanks again for the earlier feedback.

src/models/kv_cache.cpp
```diff
@@ -432,7 +432,10 @@ DefaultKeyValueCache::DefaultKeyValueCache(State& state)
       }
 
       presents_.push_back(OrtValue::CreateTensor(Allocator(), tensor_shape, type_));
-      if (Device().GetType() != DeviceType::WEBGPU) {
+      // WebGPU has no Zero() implementation; AMDGPU skips it for the same reason (Stage A
+      // DeviceInterface throws on Zero — KV is fresh per-Generator so zeroing is optional).
```
kunal-vaishnavi: From Copilot:

The rationale here looks stale: `src/amdgpu/interface.cpp` now implements `GpuMemory::Zero()`, but this branch still skips zeroing for `DeviceType::AMDGPU` and the comment says AMDGPU "throws on Zero".

Can we either:
1. remove AMDGPU from this exclusion and rely on its `Zero()` implementation, or
2. document the specific reason AMDGPU still must skip zeroing despite having `GpuMemory::Zero()`?

As written, the comment and behavior disagree, and skipping initialization in KV-cache code is risky unless we're certain every backend/path fully overwrites these buffers before first use.

src/models/model.cpp
```diff
@@ -499,6 +503,24 @@ void EnsureDeviceOrtInit(DeviceInterface& device, const Config& config) {
...
   device.InitOrt(*Ort::api, *allocator.allocator_);
+
+  // Host-accessible allocator for decode inputs (AMDGPU only). Request it via GetSharedAllocator;
+  // if unavailable, creation returns null and callers fall back to the default device inputs path.
+  if (!allocator.host_accessible_allocator_ && type == DeviceType::AMDGPU) {
+    try {
+      // device_id is hardcoded 0, so this is correct for single-GPU only. On multi-GPU the
+      // device_id differs and GetSharedAllocator returns null (falls back, perf loss not error).
+      // TODO(multi-gpu): query EpDevice::MemoryInfo(HOST_ACCESSIBLE) instead of reconstructing.
+      auto host_info = OrtMemoryInfo::CreateV2("pinned", OrtMemoryInfoDeviceType_GPU,
```
kunal-vaishnavi: From Copilot:

This host-accessible allocator lookup feels too brittle because it reconstructs the memory identity with hardcoded values (`vendor_id=0x1002`, `device_id=0`) instead of using the active EP/device’s advertised metadata.

That means the fast path depends on our guess matching exactly how the EP registered the allocator. On systems with different device IDs, multi-GPU setups, or different allocator registration details, this could silently fail back to the default path or even pick the wrong allocator.

Can we source this `OrtMemoryInfo` from the actual EP/device metadata instead of rebuilding it here?

src/models/session_options.cpp
```diff
@@ -13,6 +13,7 @@
 #include "../openvino/session_options.h"
 #include "../qnn/session_options.h"
 #include "../ryzenai/session_options.h"
+#include "../amdgpu/session_options.h"
```
kunal-vaishnavi, line 16: Can we make the insertions in this file such that alphabetical order is maintained?

src/models/logits.h
```diff
@@ -31,6 +31,12 @@ struct Logits {
 
   std::unique_ptr<Tensor> output_raw_;  // Raw logits output from model
 
+  // Device interface for the logits output + last-token/fp32 scratch. Normally p_device_inputs_, but
+  // logits is GPU-written / CPU-read (opposite of the pinned decode inputs). For AMDGPU the inputs use
+  // a host-accessible allocator whose heap is NOT CPU-read-coherent (DML CUSTOM/L0/WRITE_COMBINE), so
+  // logits go on the CPU interface instead; only the 3 decode inputs use the pinned allocator.
+  DeviceInterface* p_logits_{};
```
kunal-vaishnavi: Can we define `p_logits_` here instead and name it as `p_device_logits_`?

https://github.com/microsoft/onnxruntime-genai/blob/b8659f2b813e8dd6254a92d5885fd72b838283d2/src/models/model.h#L174-L177

src/models/session_options.cpp
aditya-dl: Thanks, fixed.

src/models/logits.h
aditya-dl: Done. Moved `p_device_logits_` into the Model device-family (`model.h`/`model.cpp`, set alongside `p_device_inputs_`) and consumers now read `model_.p_device_logits_`.

src/models/kv_cache.cpp
aditya-dl: Went with option 1. Removed AMDGPU from the exclusion, it now relies on GpuMemory::Zero().

src/models/model.cpp
aditya-dl: Fixed. No longer reconstructs the mem-info with hardcoded ids, it's sourced from the EP's advertised metadata via `EpDevice_MemoryInfo`. Also correlated it (and the compute allocator) to the selected device id read from EP metadata, so multi-GPU picks the right device instead of assuming 0; single-GPU resolves to 0 unchanged.

src/models/model.cpp
```diff
@@ -592,6 +630,18 @@ Model::Model(std::unique_ptr<Config> config) : config_{std::move(config)} {
   else
     p_device_inputs_ = GetDeviceInterface(DeviceType::CPU);
 
+  // Host-accessible decode inputs (AMDGPU): route the small decode inputs through a
+  // host-accessible interface so the CPU updates them in place with no per-step roundtrip.
+  // KV cache and scoring stay on the default interface. Falls back if no allocator.
+  if (p_device_->GetType() == DeviceType::AMDGPU &&
```
kunal-vaishnavi: Instead of adding a new `if` case here, can we add a new `else if` case above?

src/models/model.cpp
```diff
@@ -499,6 +510,33 @@ void EnsureDeviceOrtInit(DeviceInterface& device, const Config& config) {
     throw std::runtime_error("Unexpected failure to create device memory allocator for " + std::string(name));
   }
   device.InitOrt(*Ort::api, *allocator.allocator_);
+
+  // Host-accessible allocator for decode inputs (AMDGPU only). Request it via GetSharedAllocator;
+  // if unavailable, creation returns null and callers fall back to the default device inputs path.
+  if (!allocator.host_accessible_allocator_ && type == DeviceType::AMDGPU) {
```
kunal-vaishnavi: This feels like it should be inside the interface since it's AMD GPU specific. Can we make that happen?

src/models/kv_cache.cpp
```diff
@@ -432,6 +432,7 @@ DefaultKeyValueCache::DefaultKeyValueCache(State& state)
       }
 
       presents_.push_back(OrtValue::CreateTensor(Allocator(), tensor_shape, type_));
+      // WebGPU has no Zero() implementation; every other backend (incl. AMDGPU) zero-inits the KV.
```
kunal-vaishnavi:
```suggestion
      // WebGPU has no Zero() implementation; every other backend zero-inits the KV.
```

src/models/model.cpp
```diff
@@ -466,15 +467,25 @@ void EnsureDeviceOrtInit(DeviceInterface& device, const Config& config) {
...
 
-  // Names for the device memory types used by 'OrtMemoryInfo::Create'
-  static const char* device_memory_type_names[] = {"CPU (Not used, see above)", "Cuda", "DML", "WebGPU_Buf", "QnnHtpShared", "QnnHtpShared", "OpenVINO (Not used, see above)", "Cuda", "Cpu"};
+  // Names for the device memory types used by 'OrtMemoryInfo::Create'. AMDGPU is "Hip".
+  static const char* device_memory_type_names[] = {"CPU (Not used, see above)", "Cuda", "DML", "WebGPU_Buf", "QnnHtpShared", "QnnHtpShared", "OpenVINO (Not used, see above)", "Cuda", "Cpu", "Hip"};
   static_assert(std::size(device_memory_type_names) == static_cast<size_t>(DeviceType::MAX));
 
   // Get the allocator from the OrtSession for the DeviceType (it's called 'AllocatorCreate' but it's really 'AllocatorGet')
   auto name = device_memory_type_names[static_cast<int>(type)];
+  // AMDGPU: use the selected device's id rather than a hardcoded 0. Single-GPU resolves to 0.
```
kunal-vaishnavi: Are you planning to run ORT GenAI on machines with multiple AMD GPUs?

src/models/model.cpp
```diff
@@ -466,15 +467,25 @@ void EnsureDeviceOrtInit(DeviceInterface& device, const Config& config) {
...
+  if (type == DeviceType::AMDGPU) {
```
kunal-vaishnavi: Could we move this into the interface and have it set to 0 by default for non AMD GPU EPs?

src/models/onnxruntime_api.h
```diff
@@ -827,6 +832,12 @@ struct OrtSession {
 struct OrtMemoryInfo {
   static std::unique_ptr<OrtMemoryInfo> CreateCpu(OrtAllocatorType type, OrtMemType mem_type1);
   static std::unique_ptr<OrtMemoryInfo> Create(const char* name, OrtAllocatorType type, int id, OrtMemType mem_type);
+  // Wraps CreateMemoryInfo_V2 — lets us request a specific OrtDeviceMemoryType (e.g.
+  // HOST_ACCESSIBLE) so an EP that registers both DEFAULT and HOST_ACCESSIBLE allocators
+  // (MIGraphX HipPinned, DML CUSTOM/L0) hands back the host-accessible one.
+  static std::unique_ptr<OrtMemoryInfo> CreateV2(const char* name, OrtMemoryInfoDeviceType device_type,
```
baijumeswani: Since this function is internal to ort-genai, can we just call it Create and overload the existing Create?
We can keep the comment as is.

src/models/model.cpp
```diff
@@ -592,6 +630,18 @@ Model::Model(std::unique_ptr<Config> config) : config_{std::move(config)} {
   else
     p_device_inputs_ = GetDeviceInterface(DeviceType::CPU);
 
+  // Host-accessible decode inputs (AMDGPU): route the small decode inputs through a
+  // host-accessible interface so the CPU updates them in place with no per-step roundtrip.
+  // KV cache and scoring stay on the default interface. Falls back if no allocator.
+  if (p_device_->GetType() == DeviceType::AMDGPU &&
+      p_device_->GetHostAccessibleAllocator() != nullptr) {
+    if (auto* pinned = GetAMDGPUPinnedInputsInterface())
+      p_device_inputs_ = pinned;
```
baijumeswani: Why not use the real device allocator for p_device_inputs? Why should the inputs be allocated through a host accessible allocator?

Another question is why not these have be on CPU if the size of these inputs is small?

src/models/model.h
```diff
@@ -173,6 +173,7 @@ struct Model : std::enable_shared_from_this<Model>, LeakChecked<Model>, External
 
   DeviceInterface* p_device_{};          // The device we're running on (matches device_type_) used for things that work the same on all devices
   DeviceInterface* p_device_inputs_{};   // For some model inputs, the device might be the CPU device (all but KV cache currently for WebGPU and DML)
+  DeviceInterface* p_device_logits_{};   // Logits are CPU-read; on AMDGPU the host-accessible input allocator isn't CPU-read-coherent, so logits use CPU. Others match p_device_inputs_.
```
baijumeswani: What would it take for the host accessible allocator's memory to be cpu-read coherent? Does that require an ort change?

src/models/model.cpp
```diff
@@ -466,15 +467,25 @@ void EnsureDeviceOrtInit(DeviceInterface& device, const Config& config) {
   const auto trivial_model = GetTrivialModel();
   allocator.session_ = OrtSession::Create(GetOrtEnv(), trivial_model.data(), trivial_model.size(), session_options.get());
 
-  // Names for the device memory types used by 'OrtMemoryInfo::Create'
-  static const char* device_memory_type_names[] = {"CPU (Not used, see above)", "Cuda", "DML", "WebGPU_Buf", "QnnHtpShared", "QnnHtpShared", "OpenVINO (Not used, see above)", "Cuda", "Cpu"};
+  // Names for the device memory types used by 'OrtMemoryInfo::Create'. AMDGPU is "Hip".
```
baijumeswani: Could we remove specialized comments for AMD? I think from the code, we should be able to tell AMD maps to Hip?

src/models/model.cpp
aditya-dl: Not on the near-term roadmap, but the EP advertises its memory-info per device, so hardcoded 0 is wrong if someone runs on a multi-GPU box. 

Reading the id from EP's metadata is nearly free and resolves to 0 on a single-GPU machine.

src/models/model.cpp
aditya-dl: Both alternatives were measured:
| decode inputs | vs CPU baseline |
| - | - |
| CPU | - |
| host-accessible (pinned) | ~+7% |
| real device allocator (VRAM) | ~-18% |

On device allocator: The 3 inputs are rewritten by the CPU every decode step. In VRAM that means an explicit host-to-device transfer per step + synchronization around it; which is the regression. 

On CPU: the size is what makes the copy cheap, but it is not what removes it. The EP still has to get the data onto the device each step, so the transfer just moves inside the EP rather than disappearing. 

Host-accessible is less a third place to put the data than a way to delete the transfer: one mapped allocation is CPU-writable and GPU-readable, so `UpdatePositionIds` / `UpdateAttentionMask` write in place and the model reads the same memory. No copy in either direction.

src/models/model.h
aditya-dl: Tested it - the routing is required. Given there are multiple backends in the EP, there is one backend which EP advertises host-accessible memory for and it isn't CPU-read-coherent, the output goes to garbage if logits are placed there. 

We will working on that for the EP and push a new PR to address logits to host-accessible.

src/models/onnxruntime_api.h
aditya-dl: Deleting it as it wasn't required after the rebase

src/models/model.cpp
aditya-dl: Done

src/models/model.cpp
aditya-dl: Done

src/models/model.cpp
aditya-dl: Done.

kunal-vaishnavi (dismissed):

kunal-vaishnavi:
Approving this PR. We should update the main README to show that AMD GPU is supported instead of on the roadmap. Looks like there is still a merge conflict to resolve as well.

src/models/model.h
```diff
@@ -188,6 +188,7 @@ struct Model : std::enable_shared_from_this<Model>, LeakChecked<Model>, External
 
   DeviceInterface* p_device_{};          // The device we're running on (matches device_type_) used for things that work the same on all devices
   DeviceInterface* p_device_inputs_{};   // For some model inputs, the device might be the CPU device (all but KV cache currently for WebGPU and DML)
+  DeviceInterface* p_device_logits_{};   // Logits are read back on the CPU every step
```
baijumeswani: This comment is misleading. Logits can be on the device for cuda and nvtensorrtrtx eps.

src/models/model.cpp
```diff
@@ -586,15 +594,24 @@ Model::Model(std::unique_ptr<Config> config) : config_{std::move(config)} {
   CreateSessionOptions();
   EnsureDeviceOrtInit(*p_device_, *config_);
 
+  // Inputs-only interface backed by a host-accessible allocation, so the CPU updates the small
+  // decode inputs in place with no per-step roundtrip. Null if the device offers no such allocator.
+  DeviceInterface* p_host_accessible_inputs = GetAMDGPUPinnedInputsInterface();
```
baijumeswani: The variable reads p_host_accessible_inputs  but queries a EP specialized function `GetAMDGPUPinnedInputsInterface`. Could we move this behind an abstraction such as p_device->GetHostAccessibleDevice()?

Now, that we have so many device interface variables, we could consider moving all behind p_device->GetInputDevice(), p_device->GetLogitsDevice() so the code can have ep specific overrides instead of us having to add EP specialized functions at the top level.

baijumeswani:
A few thoughts:

1. Could you explain how the different EPs are to be thought of. Right now, we have 5 AMD EPs: `VitisAI`, `RyzenAI`, `MigraphX`, `rocm`, `AMDGPU`. How are they related? And if they are not related, what mental model should we have on which scenario each ep specializes in.
2. It would be good if we can keep EP abstractions away from the main code flow and push them behind device interfaces as much as possible.
3. Do we expect new models to show up in foundry-local for amdgpu ep? Is there a timeline? Is there anything needed from us to enable those models?

src/models/model.h
aditya-dl: Fixed. It now reads `Matches p_device_inputs_ unless the device reads logits back on the CPU.`

src/models/model.cpp
aditya-dl: Moved it behind a `GetHostAccessibleDevice()` virtual, so `model.cpp` no longer calls anything EP-specific. On folding the other device-family members behind `GetInputDevice()/GetLogitsDevice()`, agreed in principle, but those are existing members with a lot of call sites, so I'd rather do that in a separate PR. Let me know if that works.

kunal-vaishnavi (approved):

aditya-dl:
> A few thoughts:
> 
> 1. Could you explain how the different EPs are to be thought of. Right now, we have 5 AMD EPs: `VitisAI`, `RyzenAI`, `MigraphX`, `rocm`, `AMDGPU`. How are they related? And if they are not related, what mental model should we have on which scenario each ep specializes in.
> 2. It would be good if we can keep EP abstractions away from the main code flow and push them behind device interfaces as much as possible.
> 3. Do we expect new models to show up in foundry-local for amdgpu ep? Is there a timeline? Is there anything needed from us to enable those models?

**1. How to think about the 5 EPs**

They're not 5 parallel/independent providers. They split into NPU-facing and GPU-facing, and a couple are layers of the same stack rather than separate user-facing EPs:
- **AMDGPU EP**: the single user-facing GPU EP for AMD GPUs, and the path forward for Foundry Local and Windows ML. It's an "umbrella" EP: applications register and configure one EP, and it internally routes to the best backend based on perf/compatibility/ISV heuristics. Callers should treat it as a black box and don't need to know or select the active backend.
- **MIGraphX**: our graph compiler/inference engine. Going forward it's being retired as a user-facing EP in Windows ML; it continues internally as one of the backends behind the AMDGPU EP.
- **ROCm EP**: EOL. Removed upstream as of ORT 1.23 (ROCm 7.1+), with MIGraphX EP as the recommended replacement. Note ROCm-the-stack (HIP runtime, rocBLAS/hipBLASLt, MIOpen, rocMLIR, etc.) is still the foundation everything else is built on; only the ROCMExecutionProvider that's retired.
- **VitisAI**: NPU EP.
- **RyzenAI**: NPU EP (Ryzen AI NPU path). Runs in two modes: hybrid (NPU + iGPU) and NPU-only. Built on top of the VitisAI. 

The short mental model: one user-facing GPU EP (AMDGPU), one user-facing NPU path (VitisAI/RyzenAI), and everything else is either an internal backend (MIGraphX) or retired (ROCm EP).

**2. Keeping EP abstractions behind device interfaces**

Agreed.

**3. New models in Foundry Local for the AMDGPU EP**

Yes, for Foundry Local and Windows ML, the AMDGPU EP is the go-forward user-facing GPU EP for AMD GPUs, so new-model enablement flows through it. I'll follow up separately with the current onboarding timeline and any specific steps needed on our side, since that side is moving separately.

baijumeswani (approved):

baijumeswani:
@aditya-dl could you please agree to the license terms above?

aditya-dl:
@microsoft-github-policy-service agree company="Advanced Micro Devices"

baijumeswani (approved):

PR #2166, opened 2026-05-18
Auto-detect fixed kv-cache shape in DefaultKeyValueCache

akholodnamdcom:
## Summary

  Some compiled backends (e.g. AMD RyzenAI, and likely future NPU/accelerator targets) emit decoder models where the `past_key_values.N.{key,value}` ONNX inputs declare a **static positive integer** in the `seq_len` dimension instead of the usual symbolic `past_sequence_length` parameter. Such models require the kv-cache tensor to be allocated to exactly that size and reused as a shared past/present buffer — ORT will reject any cache tensor whose shape doesn't match the model's static dim, and the runtime cannot pick a size from `search.max_length` after the fact.

  This PR makes OGA notice the static dim at cache construction and size the kv-cache accordingly, with no config schema change, no public API change, and no client cooperation required.

  ## How it works

  In `DefaultKeyValueCache`'s constructor, right after the existing per-layer `head_dim` auto-detection (which already reads `SessionInfo::GetInputShape`), walk each layer's `past_key.N` input and look at the second-to-last dim:

  - All layers report a **positive integer** for that dim and the values agree → treat the model as fixed-shape; let `fixed_kv_seq_len = that value`.
  - Any layer reports `<= 0` (symbolic, typically `-1`) or values disagree → leave detection off; behavior unchanged.

  When detection fires:

  - `past_present_share_buffer_` is forced `true` (model demands it; the existing share-buffer code path already does the right thing for a pre-sized buffer reused as both past and present).
  - The cache is allocated to `fixed_kv_seq_len`, not `search.max_length`. A new local `cache_seq_len = fixed_kv_seq_len > 0 ? fixed_kv_seq_len : search.max_length` replaces the direct read of `search.max_length` in the share-buffer branch, so dynamic-shape models are byte-equivalent to before.
  - Beam search (`num_beams != 1`) is rejected with a clear `std::runtime_error` — beam picking reshuffles past tensors, which is incompatible with a model-mandated fixed shape.
  - An `info` log records the detected size (gated by `g_log.enabled`, consistent with surrounding logs).
  - If `search.max_length > fixed_kv_seq_len`, a `warning` is emitted once per process (`std::call_once`) noting the cache is sized to the model's limit. We **do not clamp** `search.max_length` because `state_.params_` is `std::shared_ptr<const GeneratorParams>` and is owned by the caller; clamping would require a public API change.

  ## Files changed

  - `src/models/kv_cache.cpp` — `+67/-2`. One detection block in `DefaultKeyValueCache::DefaultKeyValueCache`, one `shape_[2]` source change in the share-buffer branch, and `#include <mutex>` for `std::call_once`.

  No header changes. No public API changes. No config schema changes. No changes to `Combined/Windowed/LFM2/Cross/ModelManaged` cache classes (out of scope for this PR — AMD RyzenAI decoder-only models route through `DefaultKeyValueCache`).

  ## Reused infrastructure

  - `SessionInfo::GetInputShape(name)` — `src/models/model.cpp:533-538`. Same call already used at `kv_cache.cpp:217` for `head_dim` auto-detection. Returns the ONNX-declared shape with `<= 0` for symbolic dims (confirmed by sibling checks in `src/models/multi_modal.cpp` and `src/models/recurrent_state.cpp`).
  - `Log("warning"|"info", ...)` — `src/logging.h:82`. Same pattern as the existing `Log` calls in `kv_cache.cpp` (lines 156, 201, 239).
  - The share-buffer data-path — `kv_cache.cpp:301-310` (shape), `:357-361` (Add: inputs = presents), `:366-367` (Update no-op), `:409-410` (RewindTo no-op). All already correctly handle a pre-sized buffer.

  ## Verification

  Tested end-to-end on Windows x64, RelWithDebInfo, CPU EP, with `Llama-3.2-1B-Instruct` (Q4F16). Since no existing test model under `test/test_models/` declares a static kv-cache `seq_len`, I produced a patched copy of the Llama model whose past_key/value (and present.key/value)  ONNX shapes were rewritten in place — only graph metadata, no weight reload — from symbolic `past_sequence_length` to a static `dim_value = 128`. With those static dims in place, ORT will reject any cache tensor that isn't exactly `[1, 8, 128, 64]`; bound-and-run success is therefore proof that the cache was sized correctly.

  | # | Model | `-ml` flag | `search.max_length` | Detection | Outcome |
  |---|---|---|---|---|---|
  | 1 | Fixed-shape (patched, static `[1,8,128,64]`) | `-1` (use config) | `4096` (JSON) | fires | ✅ warning printed, model bound + ran 32 tokens against the 128-slot cache. Without detection the cache would be `[1,8,4096,64]` → bind failure. |
  | 2 | Fixed-shape | `1024` (runtime override) | `1024` | fires | ✅ warning printed, cache stayed at 128, ran 16 tokens. Confirms runtime `SetSearchOption` overrides of `max_length` are ignored for cache sizing. |
  | 3 | Original dynamic-shape Llama | `-1` → `131072` | `131072` | does not fire | ✅ no warning, no `info`, normal run. Regression preserved. |
  | 4 | Fixed-shape, multi-iter (`-w 1 -r 3` → 4 cache constructions) | `-1` | `4096` | fires once | ✅ exactly one warning across the run (`std::call_once`). |

  Verbatim warning emitted in test 1:

  ```
  [warning]  Model has fixed kv-cache seq_len=128 but search.max_length=4096; cache is sized to the model's limit, so generation beyond it will fail.
  ```

  ## Backward compatibility

  For every model that ships today with symbolic kv-cache dims (i.e. essentially all decoder models built via the OGA model builder or the Microsoft ORT pipelines), `seq_dim <= 0` at every layer → detection short-circuits → cache construction is byte-identical to `main`. No behavior change is exposed to existing users.

### DIFF
--- a/src/models/kv_cache.cpp
+++ b/src/models/kv_cache.cpp
@@ -257,6 +257,16 @@ DefaultKeyValueCache::DefaultKeyValueCache(State& state)
   // When that's the case the cache must be allocated to exactly that size and
   // reused as a shared past/present buffer; max_length cannot drive the size
   // because ORT rejects any tensor that doesn't match the model's static dim.
+  //
+  // Limitation: only uniform per-layer static sizes are recognised here —
+  // every past_key layer must declare the same fixed seq_len. Models that
+  // declare *different* static seq_lens per layer (e.g. a mix of full-attention
+  // and sliding-window layers with distinct static caps) fall through to
+  // dynamic handling. Lifting this restriction would extend the existing
+  // layer_shapes_ infrastructure used for per-layer head_dim detection above:
+  // store the per-layer detected seq_len into layer_shapes_[i][2] instead of
+  // a single scalar, and let the share-buffer branch's per-layer loop do the
+  // rest. Deferred until a model in the wild actually needs it.
   int64_t fixed_kv_seq_len = 0;
   {
     bool all_fixed_and_uniform = layer_count_ > 0;
--- a/src/models/kv_cache.cpp
+++ b/src/models/kv_cache.cpp
@@ -8,7 +8,6 @@
 #include "../openvino/interface.h"
 #include "../qnn/interface.h"
 #include <algorithm>
-#include <mutex>
 
 namespace Generators {
 
@@ -156,7 +155,7 @@ namespace {
 //   - reject beam search (num_beams != 1),
 //   - force past_present_share_buffer on,
 //   - log info on the detected size,
-//   - warn once per process if search.max_length exceeds the detected size.
+//   - warn if search.max_length exceeds the detected size.
 // Returns the detected static seq_len, or 0 if the model has symbolic
 // kv-cache dims or per-layer static sizes disagree.
 //
@@ -210,16 +209,13 @@ int64_t DetectAndConfigureFixedKvShape(const SessionInfo& session_info,
                     std::to_string(common_seq_len) +
                     "; allocating shared past/present buffer to that size.");
   }
-  if (search.max_length > static_cast<int>(common_seq_len)) {
-    // De-duplicate across repeated Generator constructions (e.g. benchmark loops).
-    static std::once_flag warned_once;
-    std::call_once(warned_once, [&] {
-      Log("warning", "Model has fixed kv-cache seq_len=" +
-                         std::to_string(common_seq_len) +
-                         " but search.max_length=" +
-                         std::to_string(search.max_length) +
-                         "; cache is sized to the model's limit, so generation beyond it will fail.");
-    });
+  if (search.max_length > static_cast<int>(common_seq_len) &&
+      g_log.enabled && g_log.warning) {
+    Log("warning", "Model has fixed kv-cache seq_len=" +
+                       std::to_string(common_seq_len) +
+                       " but search.max_length=" +
+                       std::to_string(search.max_length) +
+                       "; cache is sized to the model's limit, so generation beyond it will fail.");
   }
   return common_seq_len;
 }
--- a/src/models/kv_cache.cpp
+++ b/src/models/kv_cache.cpp
@@ -8,6 +8,7 @@
 #include "../openvino/interface.h"
 #include "../qnn/interface.h"
 #include <algorithm>
+#include <mutex>
 
 namespace Generators {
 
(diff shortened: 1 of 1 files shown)


akholodnamdcom:
@baijumeswani, Does this one look like what we agreed on? Thanks!

src/models/kv_cache.cpp
```diff
@@ -250,6 +251,65 @@ DefaultKeyValueCache::DefaultKeyValueCache(State& state)
     }
   }
 
+  // Auto-detect a fixed kv-cache shape from the model's past_key input shapes.
+  // Some compiled backends (e.g. AMD RyzenAI) emit models where the kv-cache
+  // seq_len dimension is a static positive integer instead of a symbolic dim.
+  // When that's the case the cache must be allocated to exactly that size and
+  // reused as a shared past/present buffer; max_length cannot drive the size
+  // because ORT rejects any tensor that doesn't match the model's static dim.
+  int64_t fixed_kv_seq_len = 0;
```
kunal-vaishnavi: Currently, these changes assume that the KV cache is the same static size per layer. How would you support per-layer static cache lengths for models that use different sizes per layer?

baijumeswani: I think to unblock RyzenAI models, this should be an ok limitation for now. Maybe we can add a comment highlighting that limitation?

src/models/kv_cache.cpp
```diff
@@ -250,6 +251,65 @@ DefaultKeyValueCache::DefaultKeyValueCache(State& state)
...
+      static std::once_flag warned_once;
+      std::call_once(warned_once, [&] {
+        Log("warning", "Model has fixed kv-cache seq_len=" +
+                           std::to_string(fixed_kv_seq_len) +
+                           " but search.max_length=" +
+                           std::to_string(state_.params_->search.max_length) +
+                           "; cache is sized to the model's limit, so generation beyond it will fail.");
+      });
+    }
+  }
```
baijumeswani: Can all of this logic be put into a helper function?

src/models/kv_cache.cpp
```diff
@@ -8,6 +8,7 @@
 #include "../openvino/interface.h"
 #include "../qnn/interface.h"
 #include <algorithm>
+#include <mutex>
```
baijumeswani: Do we need this include?

akholodnamdcom: That's for `std::call_once`.

src/models/kv_cache.cpp
akholodnamdcom: Good catch, thanks. Going with @baijumeswani's suggestion — keeping the uniform-only path in this PR and added a comment in the source highlighting the limitation and the path forward (see updated lines around `// Limitation:`).

Concretely, lifting the restriction is a small follow-up: the constructor already builds a `layer_shapes_` vector when it detects per-layer `head_dim` variation a few lines above this block, and the share-buffer branch already iterates `layer_shapes_` when populating cache tensors. The per-layer static-seq_len version would store the detected value into `layer_shapes_[i][2]` instead of a single scalar `fixed_kv_seq_len`, and the rest of the data path falls through unchanged. Happy to do that in a follow-up PR once a model that actually mixes static cache sizes per layer shows up.

src/models/kv_cache.cpp
akholodnamdcom: Done in 52cc2ba4 — pulled the detection + consequence handling into a file-local helper `DetectAndConfigureFixedKvShape` in an anonymous namespace; the constructor now just reads its return value.

src/models/kv_cache.cpp
```diff
@@ -148,6 +149,83 @@ void CombinedKeyValueCache::PickPastState(DeviceSpan<int32_t> beam_indices, int
...
+        "shape (model expects seq_len=" +
+        std::to_string(common_seq_len) + ").");
+  }
+  past_present_share_buffer = true;
+  if (g_log.enabled) {
+    Log("info", "DefaultKeyValueCache: auto-detected fixed kv-cache seq_len=" +
+                    std::to_string(common_seq_len) +
+                    "; allocating shared past/present buffer to that size.");
+  }
+  if (search.max_length > static_cast<int>(common_seq_len)) {
```
kunal-vaishnavi: Why is this last `if` condition needed for repeated generator constructions? Wouldn't it be better to re-use the same generator and use the rewind API?

src/models/kv_cache.cpp
```diff
@@ -148,6 +149,83 @@ void CombinedKeyValueCache::PickPastState(DeviceSpan<int32_t> beam_indices, int
...
+    // De-duplicate across repeated Generator constructions (e.g. benchmark loops).
+    static std::once_flag warned_once;
+    std::call_once(warned_once, [&] {
+      Log("warning", "Model has fixed kv-cache seq_len=" +
```
baijumeswani: I think you had it correct earlier. You need to check if log is enabled, otherwise it will error out.

src/models/kv_cache.cpp
```diff
@@ -148,6 +149,83 @@ void CombinedKeyValueCache::PickPastState(DeviceSpan<int32_t> beam_indices, int
...
+  }
+  past_present_share_buffer = true;
+  if (g_log.enabled) {
+    Log("info", "DefaultKeyValueCache: auto-detected fixed kv-cache seq_len=" +
+                    std::to_string(common_seq_len) +
+                    "; allocating shared past/present buffer to that size.");
+  }
+  if (search.max_length > static_cast<int>(common_seq_len)) {
+    // De-duplicate across repeated Generator constructions (e.g. benchmark loops).
+    static std::once_flag warned_once;
```
baijumeswani: For applications where multiple models may be loaded in the same process (Foundry Local), this would mean that we would log this warning only once for the entire process (even though multiple models may have been loaded).

I think we should avoid the once_flag here. Logging the warning every time should be good imo.

akholodnamdcom: Okay. Will update.

src/models/kv_cache.cpp
akholodnamdcom: Fixing, thanks.

baijumeswani (approved):

src/models/kv_cache.cpp
akholodnamdcom: Good point — agreed and removed the dedup machinery in 3c9481e6 . The right pattern for repeated generations against the same model is exactly what you described: reuse one `Generator` and call `RewindTo()` between runs. `model_benchmark` exposes that as `--reuse_generator`. With the `once_flag` gone the warning now fires once per cache construction, which lines up naturally with one-warning-per-model in both single-model and multi-model hosts.

kunal-vaishnavi (approved):

PR #2171, opened 2026-05-20
Multilingual Streaming Nemotron ASR + CUDA support

nenad1002:
Adds support for NVIDIA's multilingual streaming Nemotron ASR model (nemotron_speech) with both CPU and CUDA execution providers.

CUDA support is functional and validated end-to-end, but might not yet be fully optimized.

E.g. run of the new model
```

python /home/nebanfic/onnxruntime-genai/examples/python/nemotron_speech.py   
--model_path /datadisks/disk2/nebanfic/multilingual_onnx_fp32   --audio_file /home/nebanfic/data_german/audio_test/test_000282.wav   --language de   -e cuda
Setting model to cuda
  Language: de (lang_id=9)
  Use VAD: false
------------------------------------------------------------
 Genau wieder munte er der anzieht und die gezeichneten Verursacht, übt auch die Milchstraße Anziehungskräfte auf die Sagittariusgalaxie aus
============================================================
  Genau wieder munte er der anzieht und die gezeichneten Verursacht, übt auch die Milchstraße Anziehungskräfte auf die Sagittariusgalaxie aus
```

### DIFF
--- a/src/models/nemotron_speech.cpp
+++ b/src/models/nemotron_speech.cpp
@@ -14,10 +14,13 @@ namespace Generators {
 
 namespace {
 // Picks the DeviceInterface that matches the OrtValue's actual memory location.
-// Required because ORT may place outputs on CPU even when the EP is CUDA.
-DeviceInterface& DeviceFor(const OrtValue& v) {
-  const bool on_gpu = v.GetTensorMemoryInfo().GetDeviceType() == OrtMemoryInfoDeviceType_GPU;
-  return *GetDeviceInterface(on_gpu ? DeviceType::CUDA : DeviceType::CPU);
+// Required because ORT may place outputs on CPU even when the EP is on a device
+// (e.g. CUDA, DML, QNN). For non-CPU placements we fall back to the model's
+// own input device interface rather than hardcoding a specific EP, so this
+// works across any backend the model was built with.
+DeviceInterface& DeviceFor(const OrtValue& v, DeviceInterface& model_device) {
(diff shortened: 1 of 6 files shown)


src/models/nemotron_speech.cpp
```diff
@@ -11,6 +12,18 @@
 
 namespace Generators {
 
+namespace {
+// Picks the DeviceInterface that matches the OrtValue's actual memory location.
+// Required because ORT may place outputs on CPU even when the EP is on a device
```
baijumeswani, line 17: ORT places the output on CPU when the output was not pre-allocated and passed into session run. If we know the shape of the output ahead of time, we should allocate it on the device we want it to be on. The ort will just populate that buffer and will not allocate the output on cpu.

nenad1002: I think I tried exactly that and it did not work cause I had mixed placement, causing ORT inference to crash. 

I can try again, will give to Copilot to follow your exact instructions, maybe I was manually doing something wrong.

nenad1002: Let me leave CUDA optimization for the future PR, as long as we don't have bugs / memory leaks, we should be fine, CUDA is running ~15 RTFx compared to CPU which is ~8 RTFx. 

I want to do CUDA optimization properly without just letting Copilot/AI spawn random fixes, and can't do that in one day. We should also try to place as many nodes on CUDA as possible which is another axis of optimization.

baijumeswani (approved):

kunal-vaishnavi (approved):

rui-ren (approved):

PR #2176, opened 2026-05-22
Add Csharp binding for Multi-lingual ASR

rui-ren:
Add the C# Binding for `Nemotron ASR`

- This is essential for Core and FL.


baijumeswani (dismissed):

hanbitmyths (approved):

kunal-vaishnavi (approved):

kunal-vaishnavi:
We should also update the [runtime options doc](https://github.com/microsoft/onnxruntime-genai/blob/main/docs/RuntimeOptions.md) so users know that the language id is an option for Nemotron speech.

PR #2177, opened 2026-05-22
Forward user provider options to allocator init session

qjia7:
## Summary

- `EnsureDeviceOrtInit` creates a dummy ORT session to obtain a device allocator before the real user session is created. For WebGPU EP, this first session triggers `WebGpuContextFactory::CreateContext`, which creates a singleton `WebGpuContext` whose Dawn device toggles (including `skip_validation`) are immutable after creation via `std::call_once`.
- Previously the dummy session was created with empty provider options, so the WebGpuContext was always initialized with the default `validation_mode=Basic`, ignoring the user's `validationMode: "disabled"` from `genai_config.json`. When the real session arrived with the user's setting, the singleton already existed and the setting was silently dropped.
- This caused Dawn to run the expensive `TransformIndirectDispatchBuffer` validation path for every indirect dispatch (~4,199 per inference run), because the `skip_validation` toggle was never enabled on the Dawn device.
- Forward the user's provider options from `config.model.decoder.session_options.provider_options` to the dummy session, excluding `enableGraphCapture` (which would fail on the trivial model). This ensures singleton resources like WebGpuContext are initialized with the user's intended settings on the very first creation.

## Test plan

- [x] Verified via runtime debug logging that `CreateContext` is now called with `validation_mode=0` (Disabled) on the first call, and `skip_validation` toggle is included in Dawn device toggles
- [x] Verified model correctness with `qwen3-1.7b-graph-prune` (graph capture enabled)
- [x] Run e2e benchmark to measure performance improvement from skipping indirect dispatch validation

### Benchmark: phi4-graph-prune (Vulkan, graph capture ON, RTX 5080)

Same ORT Vulkan DLL used for both runs. Only GenAI differs (`main` vs this PR).

| Metric | `main` | This PR | Delta |
|--------|--------|---------|-------|
| Prefill (tps) | 11,374 | 11,654 | +2.5% |
| Token Gen (tps) | 101 | 150 | **+49%** |

Params: `-w 2 -r 2 --reuse_generator -l 1024 -g 300 -e follow_config`


guschmue (dismissed):

src/models/model.cpp
```diff
@@ -406,7 +406,27 @@ void EnsureDeviceOrtInit(DeviceInterface& device, const Config& config) {
   // Create an OrtSessionOptions and set the options to use the DeviceType we're using here
   auto session_options = OrtSessionOptions::Create();
   std::vector<Config::ProviderOptions> provider_options_list;
-  provider_options_list.emplace_back(Config::ProviderOptions{device_type_names[static_cast<int>(type)], {}});
+  const char* provider_name = device_type_names[static_cast<int>(type)];
+  Config::ProviderOptions dummy_provider_options{provider_name, {}};
+
+  // Forward the user's provider options (e.g. validationMode) to the dummy session
```
baijumeswani: The DeviceInterface (and the corresponding allocator) lives in global scope and is only created once per process.

Instead of forwarding user options for a particular model, can we define some defaults that should work across multiple models irrespective of what model gets loaded first?

qjia7: Updated the forwarding logic to a whitelist of only the 7 global/singleton options that affect the process-wide WebGpuContext: deviceId, webgpuInstance, webgpuDevice, dawnBackendType, powerPreference, validationMode, dawnProcTable. Per-session options like enableGraphCapture, bufferCacheMode are now excluded.

Will also add warnings in ORT's WebGpuContextFactory::CreateContext so that if a later session requests different singleton option values, ORT logs a warning instead of silently dropping them.

qjia7: >Instead of forwarding user options for a particular model, can we define some defaults that should work across multiple models irrespective of what model gets loaded first?

If the users don't reconfigure those parameters, ort will use the default ones. But ort also supports users to overrides those values. So for webgpu context level parameters, only the first model's parameters take effect. We can add some warnings in ort to alert users.

src/models/model.cpp
```diff
@@ -406,7 +406,37 @@ void EnsureDeviceOrtInit(DeviceInterface& device, const Config& config) {
...
+  if (type == DeviceType::WEBGPU) {
+    static const std::unordered_set<std::string> global_options = {
+        "deviceId",
+        "webgpuInstance",
+        "webgpuDevice",
+        "dawnBackendType",
+        "powerPreference",
+        "validationMode",
+        "dawnProcTable",
+    };
```
baijumeswani: Can use a constexpr std::array here:

```
constexpr std::array<std::string_view, 7> webgpu_global_options = {
        "deviceId",
        "webgpuInstance",
        "webgpuDevice",
        "dawnBackendType",
        "powerPreference",
        "validationMode",
        "dawnProcTable",
    };
```

And then search using std::find instead of .count.

src/models/model.cpp
```diff
@@ -406,7 +406,37 @@ void EnsureDeviceOrtInit(DeviceInterface& device, const Config& config) {
   // Create an OrtSessionOptions and set the options to use the DeviceType we're using here
   auto session_options = OrtSessionOptions::Create();
   std::vector<Config::ProviderOptions> provider_options_list;
-  provider_options_list.emplace_back(Config::ProviderOptions{device_type_names[static_cast<int>(type)], {}});
+  const char* provider_name = device_type_names[static_cast<int>(type)];
+  Config::ProviderOptions dummy_provider_options{provider_name, {}};
```
baijumeswani: I think this variable name is misleading. It doesn't seem like a dummy. Could we instead call it user_provided_webgpu_global_options or something like that?

baijumeswani (approved):

kunal-vaishnavi (approved):
The PR looks good to me. In the future, we should consider routing these initial session settings through the EP interface to avoid adding more EP-specific code in the base `model.cpp` file.

qjia7:
> The PR looks good to me. In the future, we should consider routing these initial session settings through the EP interface to avoid adding more EP-specific code in the base `model.cpp` file.

Merged it. Will put it in my todo list. Thanks.

PR #2179, opened 2026-05-22
Fix: Skip CPU provider in SetProviderSessionOptions

jiafatom:
### Problem
When users call `config.append_provider('cpu')`, onnxruntime-genai 0.13.x throws:
`
RuntimeError: [ErrorCode:InvalidArgument] Unknown provider name 'cpu'. Possible options: ...
`

This worked in v0.12.1 but regressed in v0.13.1.

### Root Cause
PR #2038 (*Decouple plugin execution providers from USE_WINML*, March 2026) introduced `src/models/session_options.cpp` with a fallthrough path that passes unrecognized provider names to ORT's `AppendExecutionProviderV1`. ORT's strict whitelist (added in commit `4d03aeff0e`) rejects `cpu` since CPU EP is always implicitly registered and never needs an explicit append call.

### Fix
In `SetProviderSessionOptions`, detect CPU provider names (case-insensitive: `cpu`, `cpuexecutionprovider`) and:
- **If no provider options are set**: skip silently (backward-compatible no-op)
- **If provider options are present**: throw a clear error explaining that CPU EP does not support provider options and does not need explicit registration

This ensures:
1. Existing code using `append_provider('cpu')` continues to work
2. Misconfigurations (setting CPU provider options that would be silently lost) are caught early with a helpful error message

### Testing
Added regression test `TEST(CAPITests, AppendCpuProvider)` that verifies `cpu`, `CPU`, and `CPUExecutionProvider` all succeed without error.

### DIFF
--- a/src/models/session_options.cpp
+++ b/src/models/session_options.cpp
@@ -173,6 +173,12 @@ DeviceInterface* SetProviderSessionOptions(OrtSessionOptions& session_options,
   }
 
   for (const auto& provider : providers_list) {
+    // CPU EP is always implicitly registered, skip it to avoid passing it
+    // to ORT's AppendExecutionProvider which does not accept "cpu" as a name.
+    if (provider == "cpu" || provider == "CPU" || provider == "CPUExecutionProvider") {
+      continue;
+    }
+
     auto provider_options_it = std::find_if(provider_options_list.begin(), provider_options_list.end(),
                                             [&provider](const Config::ProviderOptions& po) { return po.name == provider; });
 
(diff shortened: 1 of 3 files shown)


src/models/session_options.cpp
baijumeswani, line 169: Can you register a function for cpu? Also, let's assume that the cpu ep has a genai canonical name: https://github.com/microsoft/onnxruntime-genai/blob/38e6eb621f1b7b68cb70237d932083be0a338cb1/src/config.cpp#L19

I think these changes add some overhead that is not necessary imo. But if we need to make this work, let's add cpu names to genai where needed.

baijumeswani (approved):

PR #2183, opened 2026-05-26
Update main to version 0.15.0

baijumeswani:


kunal-vaishnavi (dismissed):

kunal-vaishnavi (dismissed):

.github/workflows/win-webgpu-x64-build.yml
```diff
@@ -105,6 +108,14 @@ jobs:
         run: |
           python test/python/test_onnxruntime_genai.py --cwd "test\python" --test_models "test\models" --e2e
 
+      - name: Run the C# Tests
```
kunal-vaishnavi, line 113: Can we make sure that this CI runs the same tests as the other CIs?

https://github.com/microsoft/onnxruntime-genai/blob/b8a784854b5bb08e26d1678092a91ef206e3e049/.github/workflows/win-cpu-x64-build.yml#L130-L145

test/python/cpu/ort/requirements.txt
```diff
@@ -1,2 +1 @@
--i https://aiinfra.pkgs.visualstudio.com/PublicPackages/_packaging/ORT-Nightly/pypi/simple/
-onnxruntime==1.25.0.dev20260126001
\ No newline at end of file
+onnxruntime==1.26.0
```
kunal-vaishnavi, line 1: If this PR isn't merged before ORT 1.27 is released, can we use 1.27 instead?

test/python/models/test_qwen_fara_models.py
```diff
@@ -729,16 +729,32 @@ def test_qwen3_5_hybrid_text_generation_cuda(test_data_path):
 # ---------------------------------------------------------------------------
 
 
+def _register_webgpu_plugin():
```
kunal-vaishnavi: Rather than defining this method here, can we have a shared helper file that we can import this method from (similar to `common.py` in our examples)?

test/python/test_onnxruntime_genai_e2e.py
```diff
@@ -14,6 +14,25 @@
 log = logging.getLogger("onnxruntime-genai-tests")
 
 
+def _register_webgpu_plugin():
```
kunal-vaishnavi: We can import the method from the helper file and use it here.

test/main.cpp
```diff
@@ -11,6 +15,94 @@
 // Global variable to store custom model base path
 std::string g_custom_model_path;
 
+namespace {
+
+namespace fs = std::filesystem;
+
+// Shared file-name stem of every ONNX Runtime execution provider plugin library, e.g.
+// "onnxruntime_providers_webgpu.dll" or "libonnxruntime_providers_cuda.so".
+#if defined(_WIN32)
```
kunal-vaishnavi, line 25: Can we move all of the `constexpr const char*` variables used throughout all of our C++ tests into a shared file? It's easier for readability and reusability purposes if we can see all predefined constants.

.github/workflows/win-webgpu-x64-build.yml
```diff
@@ -17,7 +17,10 @@ env:
   ORT_NIGHTLY_REST_API: "https://feeds.dev.azure.com/aiinfra/PublicPackages/_apis/packaging/Feeds/ORT-Nightly/packages?packageNameQuery=Microsoft.ML.OnnxRuntime&api-version=6.0-preview.1"
   ORT_PACKAGE_NAME: "Microsoft.ML.OnnxRuntime"
   binaryDir: 'build/cpu/win-x64'
-  TEST_WEBGPU: 'true'
+  WEBGPU_EP_PACKAGE_NAME: 'Microsoft.ML.OnnxRuntime.EP.WebGpu'
+  WEBGPU_EP_PACKAGE_VERSION: '0.1.0'
```
qjia7: 0.1.0 is not newer enough. The is one webgpu test failed.
```
[  FAILED  ] CAPITests.RewindQwen25CAPI

File: cmake/ortlib.cmake [hunk 7]
```diff
@@ -77,16 +80,15 @@ if(ORT_HOME)
   endif()
 else()
   # If ORT_HOME is not specified, download the onnxruntime headers and libraries from the nightly feed
-  set(ORT_VERSION "1.25.0-dev-20260125-1205-727db0d3dc")
+  set(ORT_VERSION "1.26.0")
```
qjia7, line 83: The new API [SessionReleaseCapturedGraph](https://github.com/microsoft/onnxruntime/blob/main/include/onnxruntime/core/session/onnxruntime_c_api.h#L7496) was added since Version 1.27. So can we upgrade it into 1.27.dev?

.github/workflows/win-webgpu-x64-build.yml
baijumeswani: For this PR, I am just focussing on updating our testing to use webgpu plugin ep and updating the versions of ort/ort-genai.

Since I added the C# infra for plugin ep, I added the C# test as well. But the C# example, and C# tool calling example are not updated in this PR. We can enable those separately.

cmake/ortlib.cmake
baijumeswani: Can we do that in your PR that calls into this API? Maybe we can consider default build with 1.26.0. And if user brings 1.27.0, we compile with that api in your PR. This way, we are not forced into using that ort version if we don't need it.

test/main.cpp
baijumeswani: I think it is probably better if the constexpr are localized to where they are used. Unless they need to be shared across multiple files, I think putting them in the *.cpp file is better.

But I will clean up this file.

test/python/cpu/ort/requirements.txt
baijumeswani: I think ort 1.27.0 might be release next week. Let's update this next week. One thing to be careful about is that the 1.27 ort is now built with cuda 13. So the gpu package is not friendly to our pipelines. We should update our pipelines as well once ort 1.27.0 goes out.

.github/workflows/win-webgpu-x64-build.yml
baijumeswani: The error was coming from webgpu graph capture. The graph executor code was creating an ort session with v1 api which does not work with plugin ep.

kunal-vaishnavi (dismissed):

.github/workflows/win-webgpu-x64-build.yml
```diff
@@ -105,6 +110,14 @@ jobs:
         run: |
           python test/python/test_onnxruntime_genai.py --cwd "test\python" --test_models "test\models" --e2e
 
+      - name: Run the C# Tests
+        run: |
+          # ONNX Runtime comes from the ORT-Nightly ADO feed. Passing EPDir is the single switch that
+          # turns on the EP-plugin registration code path; the tests enumerate and register every EP
+          # plugin in that directory (the one the pipeline downloaded them to).
+          cd test\csharp
+          dotnet test /p:Configuration=Release /p:NativeBuildOutputDir="$env:GITHUB_WORKSPACE\$env:binaryDir\Release" /p:EPDir="$env:GITHUB_WORKSPACE\ort\ep" --verbosity normal
```
qjia7, line 119: The webgpu CI fails that in `Run the C# Tests`. You may need to copy 1.28.0-dev onnxruntime.dll over the NuGet v1.26 DLL to avoid the upper-bound check.

kunal-vaishnavi (approved):

PR #2191, opened 2026-06-01
Fix webgpu shared buffer issue

apsonawane:
This pull request refactors the management of recurrent state buffers in `RecurrentState` to better support both graph capture (e.g., TRT-RTX) and general execution providers. The main change is to conditionally use shared or separate buffers for past/present state, improving compatibility and correctness across execution environments.

**Recurrent state buffer management:**

* Added a `share_buffers_` flag to `RecurrentState` to determine whether to use the same buffer for both input and output (for graph capture) or separate buffers with swapping (for general EPs like WebGPU). [[1]](diffhunk://#diff-b1c111b3b3b54cd6c983b851715f3e073b1331c1464b255bfe697e6fabe81337L91-R150) [[2]](diffhunk://#diff-0684b0daefd911bd4e5114b4f14d4d261fbcf5367c77138b2491698b0058dd64R30-R38)
* Allocates and manages `pasts_` and `presents_` vectors conditionally based on `share_buffers_`, ensuring correct buffer usage and initialization.
* Refactored the `Add` and `Update` methods to handle buffer aliasing and swapping according to the execution mode, maintaining stable buffer addresses for graph capture and proper buffer updates otherwise.
* Updated the `RewindTo` method to zero and rebind buffers as needed, supporting both shared and separate buffer modes.

**Internal state tracking:**

* Introduced `input_index_` and `output_index_` members to track the range of recurrent state buffers within the global input/output vectors, enabling efficient buffer rebinding after rewinding or updating. [[1]](diffhunk://#diff-b1c111b3b3b54cd6c983b851715f3e073b1331c1464b255bfe697e6fabe81337L91-R150) [[2]](diffhunk://#diff-0684b0daefd911bd4e5114b4f14d4d261fbcf5367c77138b2491698b0058dd64R30-R38)

### DIFF
--- a/src/models/recurrent_state.cpp
+++ b/src/models/recurrent_state.cpp
@@ -94,11 +94,11 @@ RecurrentState::RecurrentState(State& state)
         "Set past_present_share_buffer to true in genai_config.json.");
   }
 
-  // Graph capture (TRT-RTX) requires stable buffer addresses across steps —
-  // bind the same tensor as both input and output. Otherwise use separate
-  // past/present buffers with swap, which is compatible with all EPs including
-  // WebGPU (which prohibits aliasing a buffer as both read-only and read-write).
-  share_buffers_ = state_.params_->use_graph_capture;
+  // WebGPU prohibits binding the same buffer as both read-only (input) and
+  // read-write (output) storage in the same compute pass, so it must use
+  // separate past/present buffers with swap. All other EPs share buffers
+  // for stable addresses (required by TRT-RTX graph replay, beneficial elsewhere).
+  share_buffers_ = model_.p_device_kvcache_->GetType() != DeviceType::WEBGPU;
 
   if (!share_buffers_) {
     pasts_.resize(num_layers * 2);
@@ -130,8 +130,8 @@ void RecurrentState::Add() {
 
   const int num_layers = static_cast<int>(layer_indices_.size());
   for (int i = 0; i < num_layers * 2; ++i) {
-    // Graph capture: alias input=output for stable addresses.
-    // Otherwise: separate past/present buffers.
+    // Shared: alias input=output for stable addresses.
+    // WebGPU: separate past/present buffers to avoid aliasing violation.
     state_.inputs_.push_back(share_buffers_ ? presents_[i].get() : pasts_[i].get());
     state_.input_names_.push_back(input_name_strings_[i].c_str());
     state_.outputs_.push_back(presents_[i].get());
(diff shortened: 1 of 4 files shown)


src/models/recurrent_state.cpp
```diff
@@ -88,44 +88,66 @@ RecurrentState::RecurrentState(State& state)
...
     throw std::runtime_error(
         "RecurrentState requires past_present_share_buffer=true. "
         "Set past_present_share_buffer to true in genai_config.json.");
   }
 
+  // Graph capture (TRT-RTX) requires stable buffer addresses across steps —
+  // bind the same tensor as both input and output. Otherwise use separate
+  // past/present buffers with swap, which is compatible with all EPs including
+  // WebGPU (which prohibits aliasing a buffer as both read-only and read-write).
+  share_buffers_ = state_.params_->use_graph_capture;
```
baijumeswani: We enable buffer sharing when use_graph_capture is enabled? It seems non-intuitive that the setting in the genai_config.json for enabling buffer-sharing is not the one used to control whether or not the buffer is shared.

Is it that the past_present_buffer_share should only be used for the default kv cache and recurrent state kv cache should not use the same setting in the config?

src/models/recurrent_state.cpp
```diff
@@ -139,10 +161,19 @@ void RecurrentState::RewindTo(size_t index) {
                          ") is a no-op. Recurrent states cannot be partially rewound.");
     return;
```
baijumeswani: This should be an error imo not a warning that is only logged if the log is enabled.

src/models/recurrent_state.cpp
apsonawane: Using deviceType as a check instead of graph capture

src/models/recurrent_state.cpp
```diff
@@ -88,61 +88,89 @@ RecurrentState::RecurrentState(State& state)
...
     throw std::runtime_error(
         "RecurrentState requires past_present_share_buffer=true. "
         "Set past_present_share_buffer to true in genai_config.json.");
   }
 
+  // WebGPU prohibits binding the same buffer as both read-only (input) and
+  // read-write (output) storage in the same compute pass, so it must use
+  // separate past/present buffers with swap. All other EPs share buffers
+  // for stable addresses (required by TRT-RTX graph replay, beneficial elsewhere).
+  share_buffers_ = model_.p_device_kvcache_->GetType() != DeviceType::WEBGPU;
```
baijumeswani, line 103: Can you add a todo comment mentioning that this is pending webgpu ep LinearAttention kernel support for past present buffer sharing.

apsonawane: Done, added the todo comment

baijumeswani (approved):

PR #2198, opened 2026-06-08
 Add real-model integration test pipeline

baijumeswani:
Adds an ADO pipeline that validates the ORT GenAI native layer against real models pulled from the `foundrylocalmodels` storage account.

## What it does

For every (model, execution provider) pair, the pipeline:
1. Builds an ORT GenAI wheel from source (one build per OS / arch).
2. Spawns a separate ADO test job that downloads only its model from blob, installs the wheel, and runs pytest.

## Coverage

| OS / arch        | Providers tested     |
|------------------|----------------------|
| Windows x64      | CPU, CUDA, WebGPU    |
| Linux x64        | CPU, CUDA            |
| macOS arm64      | CPU, WebGPU          |

## Suites

The hardwired model catalog lives in `test/python/integration/models.py`:
- `pr` (6 small models, one per architecture family): runs on every PR.
- `all` (22 models, broader coverage): runs on every merge to main.

See `test/python/integration/README.md` for the local workflow and the decision table for when to add a new model to `pr` vs `all`.

### DIFF
--- a/.pipelines/integration-tests.yml
+++ b/.pipelines/integration-tests.yml
@@ -96,6 +96,9 @@ parameters:
   default: true
 
 # Suite contents - keep in sync with test/python/integration/models.py.
+# The validate_pipeline_in_sync stage runs test/python/integration/check_models_in_sync.py
+# at the start of every run and fails the pipeline if these defaults drift
+# from the pr / all_ lists in models.py.
 - name: pr_models
   type: object
   default:
@@ -135,6 +138,10 @@ parameters:
 stages:
 - template: stages/integration-stage.yml
   parameters:
+    # Pipeline yaml's lists are passed through so the in-sync checker can
+    # compare them to models.py without parsing yaml itself.
+    pr_models: ${{ parameters.pr_models }}
+    all_models: ${{ parameters.all_models }}
     # Resolve the suite's model list at compile time so each model
     # becomes its own ADO job (fans out via ${{ each }} in the template).
     ${{ if eq(parameters.suite, 'pr') }}:
--- a/.pipelines/integration-tests.yml
+++ b/.pipelines/integration-tests.yml
@@ -1,0 +1,162 @@
+# ONNX Runtime GenAI - integration tests
+#
+# Purpose
+#   Validate the ORT GenAI native layer against real models, end to end.
+#   The Python binding is the entry point, but the goal is to exercise the
+#   C++ core: model loading, tokenizer, generator loop, and each supported
+#   execution provider.
+#
+# When it runs
+#   - On every pull request to main: the `pr` suite (small fast set).
+#   - On every merge to main:        the `all` suite (broader coverage).
+#   - Manually queued:               choose the suite and which OS/EP combos.
+#
+# What gets built
+#   ORT GenAI is built from source on each run, once per (os, arch):
+#     - Windows x64: built with --use_cuda (CUDA 12.8, sm_86 for A10). The
+#                    same wheel exercises the CPU, CUDA, and WebGPU lanes.
+#     - Linux x64:   built with --use_cuda (CUDA 12.8, sm_90 for H100)
+#                    inside the manylinux container. The same wheel
+#                    exercises the CPU and CUDA lanes.
+#     - macOS arm64: built without CUDA. The same wheel exercises the CPU
+#                    and WebGPU lanes.
+#
+# What gets tested
+#   The hardwired model catalog lives in test/python/integration/models.py.
+#   The same catalog is mirrored below in the `pr_models` and `all_models`
+#   parameter defaults so the pipeline can fan one job out per model.
+#   Keep them in sync - when adding a new model, update both.
+#
+# How models are delivered
+#   Foundry Local publishes models to the `foundrylocalmodels` storage
+#   account. Each test job (one per model × ep combination) uses the
+#   agent's managed identity to azcopy just its own model directory, then
+#   runs pytest against that local copy.
+#
+# Pipeline structure
+#   Per (os, arch):
+#     Stage 1: build the wheel.
+#     Stage 2: test jobs - one ADO job per (model, ep). Models run in
+#              parallel on separate agents, so each agent only needs disk
+#              for one model.
+
+trigger:
+  branches:
+    include:
+    - main
+
+pr:
+  branches:
+    include:
+    - main
+
+parameters:
+- name: suite
+  displayName: 'Test suite. "auto" picks pr for PRs, all for main merges.'
+  type: string
+  default: 'auto'
+  values:
+  - 'auto'
+  - 'pr'
+  - 'all'
+
+- name: windows_x64_cpu
+  displayName: 'Run Windows x64 CPU tests'
+  type: boolean
+  default: true
+
+- name: windows_x64_cuda
+  displayName: 'Run Windows x64 CUDA tests'
+  type: boolean
+  default: true
+
+- name: windows_x64_webgpu
+  displayName: 'Run Windows x64 WebGPU tests'
+  type: boolean
+  default: true
+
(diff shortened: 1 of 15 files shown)


test/python/integration/test_integration_text.py
```diff
@@ -0,0 +1,94 @@
...
+    Returns True if registration succeeded (or had already happened).
+    """
+    if getattr(_register_webgpu_plugin_once, "_done", False):
+        return True
+    try:
+        import onnxruntime_ep_webgpu as webgpu_ep  # noqa: PLC0415
+    except ImportError:
+        return False
+    og.register_execution_provider_library("webgpu", webgpu_ep.get_library_path())
+    _register_webgpu_plugin_once._done = True
```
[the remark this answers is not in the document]
baijumeswani: not true. It works with webgpu as well.

.pipelines/stages/jobs/integration-build-job.yml
```diff
@@ -0,0 +1,185 @@
...
+    - task: CopyFiles@2
+      displayName: 'Stage wheel'
+      inputs:
+        SourceFolder: '$(Build.Repository.LocalPath)/build/Linux/RelWithDebInfo/wheel'
+        Contents: '*manylinux*.whl'
+        TargetFolder: '$(Build.ArtifactStagingDirectory)/wheel'
+
+  - ${{ if eq(parameters.os, 'win') }}:
+    - task: AzureCLI@2
+      displayName: 'Download CUDA 12.8'
```
kunal-vaishnavi, line 104: Can we test on CUDA 13 as well?

.pipelines/integration-tests.yml
```diff
@@ -0,0 +1,162 @@
...
+  displayName: 'Run macOS arm64 CPU tests'
+  type: boolean
+  default: true
+
+- name: macos_arm64_webgpu
+  displayName: 'Run macOS arm64 WebGPU tests'
+  type: boolean
+  default: true
+
+# Suite contents - keep in sync with test/python/integration/models.py.
```
kunal-vaishnavi, line 98: Rather than manually keep the contents in sync, can we have a helper script to automate keeping them in sync? We can run the helper script in CIs to validate that the lists match. An error can be returned if they don't. This would ensure the contents are always in sync.

test/python/integration/models.py
```diff
@@ -0,0 +1,127 @@
...
+    "Phi-3.5-mini-instruct":              {"cpu", "cuda", "webgpu"},
+    "Phi-4":                              {"cpu", "cuda", "webgpu"},
+    "Phi-4-mini-instruct":                {"cpu", "cuda", "webgpu"},
+    "Phi-4-mini-reasoning":               {"cpu", "cuda", "webgpu"},
+    "Phi-4-reasoning":                    {"cpu", "cuda", "webgpu"},
+    "deepseek-r1-distill-llama-8b":       {"cpu", "cuda", "webgpu"},
+    "deepseek-r1-distill-qwen-1.5b":      {"cpu", "cuda", "webgpu"},
+    "deepseek-r1-distill-qwen-14b":       {"cpu", "cuda", "webgpu"},
+    "deepseek-r1-distill-qwen-7b":        {"cpu", "cuda", "webgpu"},
+    "gpt-oss-20b":                        {"cpu", "webgpu"},
```
kunal-vaishnavi: We have a CUDA version but I presume it has been excluded here due to size constraints?

test/python/integration/models.py
```diff
@@ -0,0 +1,127 @@
...
+    "qwen2.5-coder-3b-instruct":          {"cpu", "cuda", "webgpu"},
+    "qwen2.5-coder-7b-instruct":          {"cpu", "cuda", "webgpu"},
+    "qwen3-0.6b":                         {"cpu", "cuda", "webgpu"},
+    "qwen3-0.6b-pp-finetuned":            {"cpu"},
+    "qwen3-0.6b-pp-finetuned-mtt":        {"cpu"},
+    "qwen3-1.7b":                         {"cpu", "cuda", "webgpu"},
+    "qwen3-14b":                          {"cpu", "cuda", "webgpu"},
+    "qwen3-4b":                           {"cpu", "cuda", "webgpu"},
+    "qwen3-8b":                           {"cpu", "cuda", "webgpu"},
+    "qwen3.5-0.8b":                       {"cpu", "cuda", "webgpu"},
```
kunal-vaishnavi, line 66: The Qwen-3.5 series of models are VLMs so they are image-text-to-text.

tools/python/integration/suite_paths.py
```diff
@@ -0,0 +1,37 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+"""Print the storage subpath for a single (model, device) pair, used by
+the integration pipeline's per-model fetch step.
+
+Prints nothing (empty string) if the model doesn't support the device.
+
+Usage:
+    python tools/python/integration/suite_paths.py --model qwen3-0.6b --device cuda
```
kunal-vaishnavi: Can this be put with the other Python files in `test/python/integration`?

.pipelines/stages/jobs/integration-build-job.yml
baijumeswani: I'll move to CUDA 13 by default in a separate PR (moving away from cuda 12 support at that point). I think the next release of ort-genai, maybe we should support cuda 13 by default.

test/python/integration/models.py
baijumeswani: I think that model was not uploaded to the foundrylocalmodels storage account. I did not find it there. We can add cuda once we add that model there. But right now, this model is not tested against.

test/python/integration/models.py
baijumeswani: From my testing they seem to work for text to text scenarios as well. We can add a multimodality test suite in subsequent PRs to enable integration tests against vision/speech models.

test/python/integration/check_models_in_sync.py
```diff
@@ -0,0 +1,96 @@
...
+def main(argv: list[str] | None = None) -> int:
+    parser = argparse.ArgumentParser(description=__doc__)
+    parser.add_argument(
+        "--pr",
+        required=True,
+        help="Comma-separated pr_models list from the pipeline yaml.",
+    )
+    parser.add_argument(
+        "--all",
+        dest="all_",
```
kunal-vaishnavi: Should the default be `all_models` here?

```suggestion
        dest="all_models",
```

test/python/integration/models.py
kunal-vaishnavi: Yes, all VLMs should work with text-to-text scenarios as well. The comments at the beginning of this dictionary suggest that only text-to-text models should be listed here. I presumed that meant that VLMs couldn't be listed here.

test/python/integration/models.py
kunal-vaishnavi: It is available [here](https://huggingface.co/onnxruntime/gpt-oss-20b-onnx). I can put the CUDA variant there so we can add it as an option.

test/python/integration/check_models_in_sync.py
baijumeswani: Can change to `all_models` if that is clearer.

test/python/integration/models.py
baijumeswani: I think the intention of this file was to define all types of models that should be run on pr and merge (text-to-text, vlm, speech...). This PR only enables text to text.

test/python/integration/models.py
baijumeswani: I'll update this optimistically in the hopes that the cuda model is added to the foundrylocalmodels storage account.

kunal-vaishnavi (approved):

PR #2208, opened 2026-06-09
Prefer using CMake path variables where available

jaeyoonjung:
The hardcoded paths would fail if the build system uses different include or lib paths.


jaeyoonjung:
@microsoft-github-policy-service agree company="LGE"

kunal-vaishnavi:
Closing this PR in favor of https://github.com/microsoft/onnxruntime-genai/pull/2310. Please add any comments or concerns on the new PR.

jaeyoonjung:
> Closing this PR in favor of #2310. Please add any comments or concerns on the new PR.

#2310 doesn't seem to address the issue that this PR does. The point is to use CMAKE_INSTALL_INCLUDEDIR and CMAKE_INSTALL_LIBDIR instead of hardcoded paths "include" and "lib" respectively.

PR #2209, opened 2026-06-09
Fix CUDA QMoE INT4 export in Qwen and Gpt-OSS models

tianleiwu:
## Summary

INT4 CUDA exports of Qwen3.5/3.6-MoE produced incoherent output even though the fp16 export was correct. The root cause was the QMoE expert-weight encoding: the builder shipped weights in a layout/scale convention the CUDA `fpA_intB` mixed-GEMM kernel does not consume. This PR makes the model builder emit CUTLASS-prepacked INT4 QMoE weights with signed blockwise scales, and keeps the routing-critical matmuls out of quantization.

### `builders/base.py` — QMoE weight encoding
- **New `_cutlass_prepacked_blockwise_quantize`**: quantizes each expert weight `[N, K]` with ONNX Runtime's blockwise `quantize_matmul_4bits` on the transposed `[K, N]` weight, **keeps the signed scales**, and offline CUTLASS-prepacks via `pack_weights_for_cuda_mixed_gemm(..., force_arch=80)`. `force_arch=80` is correct for all SM ≥ 80 (Ampere/Ada/Hopper). This is the exact encoding validated by the `com.microsoft` QMoE CUDA parity tests.
  - The previous path applied `abs()` to the scales. `quantize_matmul_4bits` folds the block-anchor sign into the scale and the kernel dequantizes as `(q − 8) · scale`, so `abs()` corrupts every block whose anchor element is negative — about half the blocks — yielding garbage weights that *look* like INT4 quality loss.
- **`make_qmoe_weights`**: routes CUDA QMoE through the new prepacked path and **asserts `block_size ∈ {64, 128}`** (the only block sizes the CUDA QMoE kernel supports).
- **`weights_prepacked` attribute** (tri-state): default `None` omits the attribute (= the kernel's prepacked default); override via `extra_options["qmoe_weights_prepacked"]` (e.g. `0` to ship raw `[E, N, K/pack]` weights for the runtime PrePack hook).

### `builders/qwen.py` — Qwen3.5/3.6 MoE
- **Exclude routing matmuls from INT4/INT8**: the MoE router and shared-expert gate are tiny matmuls, but 4-bit rounding of their weights flips top-k expert selection (measured ~1.4 of 8 experts change per token), injecting large error into every MoE layer. They are now added to `nodes_to_exclude` for INT4/INT8.
- **Fix shared-expert node-name collision**: the gated `Mul` and the gate-up `Mul` both used `.../Mul`, producing a duplicate value name and a `ShapeInferenceError`. The final gated mul is renamed to `.../gate/Mul`.

## Validation

- INT4 next-token correlation vs. HF reference: **0.22 → 0.965**, argmax matches the reference, and greedy decoding produces coherent text (previously garbage).
- fp16 export was already correct and is unchanged.
- Op-level parity: the CUDA QMoE output reconstructs the original (un-quantized) expert weights with corr ≈ 0.99 using signed scales (≈ 0.05 with `abs()`), confirming the sign was the bug.

## Testing Notes

Export an INT4 CUDA model for a Qwen3.5/3.6-MoE checkpoint and confirm coherent generation:

```bash
python -m onnxruntime_genai.models.builder -m <qwen3.5_or_3.6_moe_id> -o out -p int4 -e cuda \
  --extra_options int4_block_size=128
```

Verify the exported QMoE nodes have no `weights_prepacked` attribute (prepacked default), `block_size=128`, and that generation is coherent.


src/python/py/models/builders/base.py
```diff
@@ -3243,6 +3263,45 @@ def make_qmoe_weights(self, weights):
...
+
+        # When the model targets the raw (un-prepacked) QMoE weight layout
+        # (weights_prepacked == 0), produce weights with ONNX Runtime's
+        # MatMulNBits-compatible blockwise quantizer. This is the exact
+        # encoding the QMoE PrePack hook (com.microsoft QMoE) expects: raw
+        # [N, K/pack] bytes + positive blockwise scales, which it then lays out
+        # into the CUTLASS fpA_intB format at load time. The builder's own
+        # _symmetric_blockwise_quantize uses a different scale/packing
+        # convention that the kernel cannot consume.
+        if self.moe_attrs.get("weights_prepacked") == 0 and self.qmoe_block_size > 0:
```
tianleiwu: This shall also check self.ep == "cuda"

src/python/py/models/builders/base.py
```diff
@@ -3243,6 +3263,45 @@ def make_qmoe_weights(self, weights):
...
 
+        # CUDA QMoE consumes CUTLASS-prepacked expert weights (the kernel's
+        # fpA_intB mixed GEMM layout). Produce them offline so the QMoE op's
+        # default (weights_prepacked=auto=prepacked) reads them directly: quantize
+        # with ONNX Runtime's blockwise quantizer, keep the signed scales, then run
+        # pack_weights_for_cuda_mixed_gemm. This is the encoding validated by the
+        # com.microsoft QMoE CUDA parity tests. The builder's own
+        # _symmetric_blockwise_quantize uses a different scale/packing convention
+        # the kernel cannot consume.
+        if self.ep == "cuda" and self.moe_attrs.get("weights_prepacked") is None and self.qmoe_block_size > 0:
```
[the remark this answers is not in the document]
tianleiwu: Fixed in 0c8d4e58: `make_qmoe_weights` now produces CUTLASS-prepacked weights for `weights_prepacked` in {None, 1}, so an explicit `qmoe_weights_prepacked=1` no longer falls through to the raw path while the op emits `weights_prepacked=1`.

src/python/py/models/builders/base.py
```diff
@@ -3243,6 +3263,45 @@ def make_qmoe_weights(self, weights):
...
+            block_size = int(self.qmoe_block_size)
+            # The CUDA QMoE fpA_intB mixed-GEMM kernel only supports block sizes
+            # of 64 or 128 for block-wise quantization.
+            assert block_size in (64, 128), (
+                f"CUDA QMoE only supports block_size 64 or 128, got {block_size}."
+            )
```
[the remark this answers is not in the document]
tianleiwu: Fixed in 0c8d4e58: replaced the `assert` with a `ValueError` (which `python -O` does not strip) and applied the check on all CUDA QMoE paths.

src/python/py/models/builders/base.py
```diff
@@ -3221,6 +3229,18 @@ def make_qmoe_op(self, name, **kwargs):
...
             extra_kwargs["block_size"] = self.moe_attrs["block_size"]
 
+        # The builder ships raw, un-prepacked [E, N, K/pack] quantized expert
+        # weights (as produced by _symmetric_blockwise_quantize / quantize_matmul).
+        # The CUDA QMoE op defaults to interpreting weights as already
+        # CUTLASS-prepacked (weights_prepacked=-1/auto), so the raw layout must be
+        # flagged explicitly with weights_prepacked=0 to trigger the in-PrePack
+        # layout transform. This attribute requires an ONNX Runtime build that
+        # includes the QMoE PrePack hook (com.microsoft QMoE weights_prepacked
+        # tri-state); it is only meaningful for integer (INT4/INT8) QMoE.
```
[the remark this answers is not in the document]
tianleiwu: Fixed in 0c8d4e58: rewrote the comment to describe the tri-state accurately (None/1 = prepacked, which is what the builder ships for CUDA; 0 = raw for the runtime PrePack hook).

src/python/py/models/builders/base.py
```diff
@@ -3287,6 +3346,92 @@ def make_qmoe_weights(self, weights):
...
+    def _cutlass_prepacked_blockwise_quantize(self, weights, block_size):
+        """Quantize a single expert's weights and CUTLASS-prepack them for the
+        CUDA QMoE fpA_intB mixed-GEMM kernel.
+
+        ``weights`` has logical shape ``[N, K]`` (quantized along ``K``). Returns
+        ``(qweight, scales)`` where ``qweight`` is the prepacked uint8 tensor of
+        shape ``[K, N/pack]`` (``pack`` = 2 for INT4, 1 for INT8) and ``scales``
+        is ``[N, K/block_size]`` positive float scales. Stacking the per-expert
+        results yields ``fc_weights`` ``[E, K, N/pack]`` and ``scales``
+        ``[E, N, K/block_size]`` — the layout the QMoE op reads when
```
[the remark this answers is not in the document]
tianleiwu: Fixed in 0c8d4e58: the docstring now states the scales are SIGNED and explains the CUDA dequant `(q - 2^(bits-1)) * scale` requires the sign (abs() would corrupt blocks with a negative anchor). A regression test asserts the scales are signed.

src/python/py/models/builders/base.py
```diff
@@ -3243,6 +3263,45 @@ def make_qmoe_weights(self, weights):
...
+            block_size = int(self.qmoe_block_size)
+            # The CUDA QMoE fpA_intB mixed-GEMM kernel only supports block sizes
+            # of 64 or 128 for block-wise quantization.
+            assert block_size in (64, 128), (
+                f"CUDA QMoE only supports block_size 64 or 128, got {block_size}."
+            )
+            try:
+                qweight, scales = self._cutlass_prepacked_blockwise_quantize(weights, block_size)
+                self.moe_attrs["block_size"] = block_size
+                return qweight, scales.to(torch.float16)
```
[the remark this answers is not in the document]
tianleiwu: Added `test/python/models/test_qmoe_weights.py` in 0c8d4e58 covering path dispatch (None/0/1), CUDA-EP gating, block-size validation, and a signed-scale regression guard (catches an accidental `abs(scales)`).

src/python/py/models/builders/base.py
tianleiwu: Done in 0c8d4e58: both the raw (`weights_prepacked=0`) quantization path and the emitted `weights_prepacked` attribute are now gated on `self.ep == "cuda"`.

src/python/py/models/builders/base.py
```diff
@@ -3243,6 +3266,50 @@ def make_qmoe_weights(self, weights):
...
         qweight, scales = None, None
 
+        weights_prepacked = self.moe_attrs.get("weights_prepacked")
+
+        # CUDA QMoE consumes CUTLASS-prepacked expert weights (the kernel's
+        # fpA_intB mixed GEMM layout). Produce them offline so the QMoE op reads
+        # them directly: quantize with ONNX Runtime's blockwise quantizer, keep
+        # the signed scales, then run pack_weights_for_cuda_mixed_gemm. This is
+        # the encoding validated by the com.microsoft QMoE CUDA parity tests. The
+        # builder's own _symmetric_blockwise_quantize uses a different
```
kunal-vaishnavi, line 3268: Previously, we added channel-wise quantization via TRT-LLM for GPT-OSS when running with the CUDA EP. Now that we have block-wise quantization supported on the CPU, CUDA, and WebGPU EPs, can we remove any channel-wise quantization and use the default block-wise quantization only? That would stay consistent with using a block-wise quantization approach for MatMul ops as well.

src/python/py/models/builders/base.py
```diff
@@ -3287,6 +3354,94 @@ def make_qmoe_weights(self, weights):
...
+        shape ``[K, N/pack]`` (``pack`` = 2 for INT4, 1 for INT8) and ``scales``
+        is ``[N, K/block_size]`` SIGNED float scales. The sign is required: the
+        CUDA kernel dequantizes as ``(q - 2^(bits-1)) * scale``, so abs() would
+        corrupt every block whose max-magnitude element is negative. Stacking the
+        per-expert results yields ``fc_weights`` ``[E, K, N/pack]`` and ``scales``
+        ``[E, N, K/block_size]`` — the layout the QMoE op reads when
+        ``weights_prepacked`` is left at its prepacked default.
+        """
+        import numpy as np
+        from onnxruntime.capi import _pybind_state as _ortpyb
```
kunal-vaishnavi, line 3334: Are there no higher-level APIs available in ORT's quantizer tools for calling these?

src/python/py/models/builders/base.py
tianleiwu: Done

src/python/py/models/builders/base.py
tianleiwu: No. quantize_matmul_4bits/8bits are re-exported by onnxruntime.quantization.matmul_nbits_quantizer, but pack_weights_for_cuda_mixed_gemm exists only in onnxruntime.capi._pybind_state — there's no high-level wrapper. The high-level MatMulNBitsQuantizer operates on ONNX graph MatMul nodes, not raw per-expert weight tensors, so it can't be used here.

src/python/py/models/builders/base.py
```diff
@@ -304,6 +304,10 @@ def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
         top_k_experts = config.num_experts_per_tok if hasattr(config, "num_experts_per_tok") else 0
         expert_weight_bits = 8 if extra_options.get("use_8bits_moe", False) else 4
         swiglu_limit = config.swiglu_limit if hasattr(config, "swiglu_limit") else None
+        # For CUDA QMoE the builder ships expert weights already CUTLASS-prepacked (offline via pack_weights_for_cuda_mixed_gemm, see make_qmoe_weights),
+        # so the QMoE op's default interpretation (weights_prepacked=-1/auto = prepacked) is exactly what we want and the attribute is omitted.
+        # Override via extra_options["qmoe_weights_prepacked"] (e.g. 0 to ship raw [E, N, K/pack] weights and let the runtime PrePack hook transform them).
+        weights_prepacked = int(extra_options["qmoe_weights_prepacked"]) if "qmoe_weights_prepacked" in extra_options else -1
```
kunal-vaishnavi:
```suggestion
        weights_prepacked = int(extra_options.get("qmoe_weights_prepacked", -1))
```

src/python/py/models/builders/base.py
```diff
@@ -330,7 +335,7 @@ def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
         # Quantization-specific variables (INT4, INT8, etc.)
         algo_config = self.make_algo_config(extra_options.get("int4_algo_config", "default"))
         self.matmul_block_size = int(extra_options.get("int4_block_size", 32))
-        self.qmoe_block_size = int(extra_options.get("qmoe_block_size", 128 if self.ep in {"cuda", "trt-rtx"} else 32))
+        self.qmoe_block_size = int(extra_options.get("qmoe_block_size", 128 if self.ep in {"trt-rtx"} else 32))
         self.quant_attrs = {
             "accuracy_level": int(extra_options.get("int4_accuracy_level", 4 if self.ep in ["cpu", "webgpu"] else 0)),
             "qmoe_block_size": int(self.qmoe_block_size),
```
kunal-vaishnavi:
```suggestion
            "qmoe_block_size": self.qmoe_block_size,
```

src/python/py/models/builders/base.py
```diff
@@ -3241,10 +3258,47 @@ def make_qmoe_op(self, name, **kwargs):
...
+        # them directly: quantize with ONNX Runtime's blockwise quantizer, keep
+        # the signed scales, then run pack_weights_for_cuda_mixed_gemm. This is
+        # the encoding validated by the com.microsoft QMoE CUDA parity tests. The
+        # builder's own _symmetric_blockwise_quantize uses a different
+        # scale/packing convention the kernel cannot consume.
+        #
+        # Both weights_prepacked=-1 (auto, the op's prepacked default) and
+        # weights_prepacked=1 (explicitly prepacked) mean the op reads prepacked
+        # weights, so the builder must produce prepacked weights for both.
+        if self.ep == "cuda" and weights_prepacked in (-1, 1) and self.qmoe_block_size > 0:
```
kunal-vaishnavi, line 3274: Can we reference `self.qmoe_block_size` directly? It has already been casted to an `int`.

src/python/py/models/builders/base.py
```diff
@@ -3241,10 +3258,47 @@ def make_qmoe_op(self, name, **kwargs):
...
+                return qweight, scales.to(torch.float16)
+            except Exception as e:
+                raise RuntimeError(f"CUTLASS-prepacked QMoE quantization failed with block_size={block_size}: {e}")
+
+        # weights_prepacked == 0: ship raw [N, K/pack] weights with ONNX Runtime's
+        # MatMulNBits-compatible blockwise quantizer. This is the exact encoding
+        # the CUDA QMoE PrePack hook (com.microsoft QMoE) expects: raw bytes +
+        # blockwise scales, which it lays out into the CUTLASS fpA_intB format at
+        # load time. Only valid on the CUDA EP.
+        if self.ep == "cuda" and weights_prepacked == 0 and self.qmoe_block_size > 0:
```
kunal-vaishnavi, line 3290: Same question as [here](https://github.com/microsoft/onnxruntime-genai/pull/2209#discussion_r3417441647)

test/python/builder/test_qmoe_weights.py
```diff
@@ -0,0 +1,118 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License
+
+"""Unit tests for CUDA QMoE expert-weight quantization in the model builder.
```
kunal-vaishnavi, line 4: Can you move this into `test/python/builder` instead?

https://github.com/microsoft/onnxruntime-genai/blob/7e4bc0d1e24c8a717461559dad385287eb083f03/test/README.md?plain=1#L11

src/python/py/models/builders/base.py
tianleiwu: Done in 55b8442 - switched to `int(extra_options.get("qmoe_weights_prepacked", -1))`.

src/python/py/models/builders/base.py
tianleiwu: Done in 55b8442 - `self.qmoe_block_size` is already cast to `int`, so the redundant cast is removed.

src/python/py/models/builders/base.py
tianleiwu: Done in 55b8442 - dropped the redundant `int()`; `block_size = self.qmoe_block_size` (already an `int`).

src/python/py/models/builders/base.py
tianleiwu: Done in 55b8442 - same fix as above (removed the redundant `int()`), and unified the block-size error message to match.

test/python/builder/test_qmoe_weights.py
tianleiwu: Done in 55b8442 - moved to `test/python/builder/test_qmoe_weights.py` and switched it to load the builder module from source (like `test_hunyuan.py`), so it validates the in-repo code instead of an installed package.

src/python/py/models/builders/base.py
```diff
@@ -3221,6 +3226,18 @@ def make_qmoe_op(self, name, **kwargs):
         if "block_size" in self.moe_attrs:
             extra_kwargs["block_size"] = self.moe_attrs["block_size"]
 
+        # weights_prepacked is a tri-state CUDA QMoE attribute describing the expert-weight layout (see make_qmoe_weights, which produces the matching bytes):
```
kunal-vaishnavi, line 3229: For the other EPs that implement `QMoE`, how do they handle `weights_prepacked`? If the same ONNX model needs to be shared for both the CPU EP and CUDA EP, can the attribute be processed correctly in both EPs?

src/python/py/models/builders/base.py
```diff
@@ -3241,10 +3258,47 @@ def make_qmoe_op(self, name, **kwargs):
 
     def make_qmoe_weights(self, weights):
         dtype = torch.quint4x2 if self.moe_attrs["expert_weight_bits"] == 4 else torch.int8
-        qweight, scales = None, None
+        weights_prepacked = self.moe_attrs.get("weights_prepacked")
+
+        # CUDA QMoE consumes CUTLASS-prepacked expert weights (the kernel's
```
kunal-vaishnavi, line 3263: Can we combine both cases like this?

```py
if self.ep == "cuda" and self.qmoe_block_size > 0:
    # Options for packing QMoE weights on CUDA EP:
    #
    # CUDA QMoE consumes CUTLASS-prepacked expert weights (the kernel's
    # fpA_intB mixed GEMM layout). Produce them offline so the QMoE op reads
    # them directly: quantize with ONNX Runtime's blockwise quantizer, keep
    # the signed scales, then run pack_weights_for_cuda_mixed_gemm. This is
    # the encoding validated by the com.microsoft QMoE CUDA parity tests. The
    # builder's own _symmetric_blockwise_quantize uses a different
    # scale/packing convention the kernel cannot consume.
    #
    # Both weights_prepacked=-1 (auto, the op's prepacked default) and
    # weights_prepacked=1 (explicitly prepacked) mean the op reads prepacked
    # weights, so the builder must produce prepacked weights for both.
    #
    # weights_prepacked == 0: ship raw [N, K/pack] weights with ONNX Runtime's
    # MatMulNBits-compatible blockwise quantizer. This is the exact encoding
    # the CUDA QMoE PrePack hook (com.microsoft QMoE) expects: raw bytes +
    # blockwise scales, which it lays out into the CUTLASS fpA_intB format at
    # load time. Only valid on the CUDA EP.
    quantize_method = self._matmulnbits_blockwise_quantize if weights_prepacked == 0 else self._cutlass_prepacked_blockwise_quantize
    descriptor = "MatMulNBits-compatible" if weights_prepacked == 0 else "CUTLASS-prepacked"

    if self.qmoe_block_size not in (32, 64, 128):
        raise ValueError(f"CUDA QMoE only supports block_size 32, 64, or 128, got {block_size}.")
    try:
        qweight, scales = quantize_method(weights)
        self.moe_attrs["block_size"] = self.qmoe_block_size
        return qweight, scales.to(torch.float16)
    except Exception as e:
        raise RuntimeError(f"{descriptor} QMoE quantization failed with block_size={block_size}: {e}")
```

The block size is already accessible in both methods via `self.qmoe_block_size` so we don't need to pass it in as a parameter.

PR #2215, opened 2026-06-11
Add Tool Calling and Reasoning Token Metadata to `genai_config.json` with Fallback Map

sayanshaw24:
# Add tool-calling and reasoning token IDs to GenAI Tokenizer

## Summary

Adds four new token ID fields to the `model` section of `genai_config.json` — `bot_token_id`, `eot_token_id`, `bor_token_id`, `eor_token_id` — and exposes them on the **Tokenizer** (alongside `bos`, `eos`, `pad`), with a backward-compatible fallback for older model packages.

This establishes a new naming convention for generation-level special tokens:

| Abbreviation | Expansion | Purpose |
|---|---|---|
| **bot** | beginning of tool (call) | marks start of tool-call content |
| **eot** | end of tool (call) | marks end of tool-call content |
| **bor** | beginning of reasoning | marks start of reasoning/thinking content |
| **eor** | end of reasoning | marks end of reasoning/thinking content |

The naming mirrors `bos`/`eos`/`pad` — short, unambiguous, and follows the same access pattern (`GetBotTokenId()` alongside `GetBosTokenId()`).

### Config parsing (`config.h` / `config.cpp`)
- Added to `Config::Model`: `bot_token_id`, `eot_token_id`, `bor_token_id`, `eor_token_id` (int, default -1)
- Parsed in `Model_Element::OnValue()` alongside existing token ID fields (bos, eos, pad)

### Tokenizer (`models/model.h` / `models/model.cpp`)
- New private members: `int32_t bot_token_id_`, `eot_token_id_`, `bor_token_id_`, `eor_token_id_`
- New public getters: `GetBotTokenId()`, `GetEotTokenId()`, `GetBorTokenId()`, `GetEorTokenId()`
- Initialized in the Tokenizer constructor from config (identical pattern to bos/eos/pad)
- **Fallback logic**: if any ID is -1 after reading config, attempts to resolve it by encoding known model-family-specific token strings through the vocabulary. This provides backward compatibility for Foundry Local consuming older model packages that predate these config fields.
  - Fallback map keyed by `model.type`:
    - `qwen2`/`qwen3`/`phi3` → `<tool_call>` / `</tool_call>`
    - `gptoss` → `<|start|>` / `<|call|>`
    - `qwen3` → `<think>` / `</think>` (reasoning)

### Public C API (`ort_genai_c.h` / `ort_genai_c.cpp`)
- `OgaTokenizerGetBotTokenId(tokenizer, &token_id)` — returns BOT token id or -1
- `OgaTokenizerGetEotTokenId(tokenizer, &token_id)` — returns EOT token id or -1
- `OgaTokenizerGetBorTokenId(tokenizer, &token_id)` — returns BOR token id or -1
- `OgaTokenizerGetEorTokenId(tokenizer, &token_id)` — returns EOR token id or -1

Same pattern as `OgaTokenizerGetBosTokenId` / `OgaTokenizerGetPadTokenId`.

### C++ wrapper (`ort_genai.h`)
- `int32_t OgaTokenizer::GetBotTokenId()`
- `int32_t OgaTokenizer::GetEotTokenId()`
- `int32_t OgaTokenizer::GetBorTokenId()`
- `int32_t OgaTokenizer::GetEorTokenId()`

### C# (`Tokenizer.cs`)
- `int Tokenizer.GetBotTokenId()`
- `int Tokenizer.GetEotTokenId()`
- `int Tokenizer.GetBorTokenId()`
- `int Tokenizer.GetEorTokenId()`

### Java (`Tokenizer.java`)
- `int tokenizer.getBotTokenId()`
- `int tokenizer.getEotTokenId()`
- `int tokenizer.getBorTokenId()`
- `int tokenizer.getEorTokenId()`

### Objective-C (`ort_genai_objc.h`)
- `- (int32_t)getBotTokenId:(NSError**)error`
- `- (int32_t)getEotTokenId:(NSError**)error`
- `- (int32_t)getBorTokenId:(NSError**)error`
- `- (int32_t)getEorTokenId:(NSError**)error`

### Python (`python.cpp`)
- `tokenizer.bot_token_id` (read-only property)
- `tokenizer.eot_token_id`
- `tokenizer.bor_token_id`
- `tokenizer.eor_token_id`

### Tests (`test/c_api_tests.cpp`)
- `TagId_Unknown` — verifies `gpt2` model (not in fallback map, no config IDs) returns -1 for all four
- `TagId_FromConfig` — creates a temp model dir with token IDs in the `model` section, verifies correct parsing via the tokenizer

## genai_config.json format

```json
{
  "model": {
    "type": "qwen3",
    "bos_token_id": 151643,
    "eos_token_id": 151645,
    "bot_token_id": 151657,
    "eot_token_id": 151658,
    "bor_token_id": 151659,
    "eor_token_id": 151660,
    "decoder": { ... }
  }
}
```

All four fields are optional. If absent, the fallback map attempts to encode the known string via the tokenizer vocabulary. If the model type isn't in the fallback map either, -1 is returned (model doesn't support tool calling / reasoning).

## Motivation

This is required for the Foundry Local Catalog v2 migration where tool/reasoning metadata is being removed from catalog.

Also, the Foundry Local C++ SDK currently detects tool-call and reasoning tokens by **double-decoding every token** through both a normal and a special-token tokenizer stream, then doing `string.find("tool_call")` / `string.find("think")` per token. This is expensive.

With token IDs exposed by GenAI on the Tokenizer, FL can do a single integer comparison per token in the decode loop — eliminating the special stream entirely for models with configured IDs.

### Why on the Tokenizer (not Model)?
- BOS, EOS, and PAD token IDs are already accessed from the Tokenizer
- Token IDs are a tokenizer-level concept — they describe the vocabulary
- Follows existing pattern: `Tokenizer::GetBosTokenId()` → `Tokenizer::GetBotTokenId()`
- No lazy init needed; resolved once in the constructor

### Why bot/eot/bor/eor naming?
- Mirrors `bos`/`eos`/`pad` — short, memorable, consistent
- Config field names also follow this pattern: `bot_token_id` alongside `bos_token_id`

### Fallback map
- Exists specifically for Foundry Local backward compatibility with older model packages
- Only triggered when config fields are not set
- Resolved eagerly in the Tokenizer constructor (no lazy cache needed since tokenizer is already available)

## Related PRs
- **Foundry-Local**: [Optimize tool/reasoning detection with token ID comparison in decode loop](https://github.com/microsoft/Foundry-Local/pull/838)


src/config.cpp
```diff
@@ -1556,13 +1584,17 @@ struct Root_Element : JSON::Element {
     if (name == "model") return model_element_;
     if (name == "search") return search_element_;
     if (name == "engine") return engine_element_;
+    if (name == "tool_calling") return tool_calling_element_;
```
kunal-vaishnavi: Let's not introduce these new sections in the GenAI config. These are model-specific attributes, and they can be stored inside the `model` section. For example,

```json
"model": {
    "bos_token_id": 1,
    "eos_token_id": 2,
    "pad_token_id": 0,
    "reasoning_start_token_id": 5,
    "reasoning_end_token_id": 6,
    "tool_call_start_token_id": 10,
    "tool_call_end_token_id": 11
}
```

The `bos_token_id`, `eos_token_id`, and `pad_token_id` fields already exist in the `model` section and have the `token_id` ending suffix since they map to an integer. We can use a similar approach and store the integer representation of these new special tokens in the `model` section of the GenAI config.

src/ort_genai_c.h
```diff
@@ -406,6 +406,19 @@ OGA_EXPORT OgaResult* OGA_API_CALL OgaModelGetType(const OgaModel* model, const
...
+/**
+ * \brief Returns a tag value for this model by name.
+ *
+ * Known tag names: "tool_call_start", "tool_call_end", "reasoning_start", "reasoning_end".
+ * Returns an empty string if the model doesn't define the requested tag.
+ *
+ * \param[in] model The model to query.
+ * \param[in] tag_name The name of the tag to retrieve.
+ * \param[out] out The tag value string. Must be destroyed with OgaDestroyString.
+ * \return OgaResult containing the error message if the call failed.
```
kunal-vaishnavi: This says an error message is returned if the call failed but the method returns an empty string.

src/config.cpp
sayanshaw24: updated!

src/ort_genai_c.h
sayanshaw24: fixed!

src/models/model.cpp
```diff
@@ -820,6 +827,64 @@ bool Model::IsPruned() const {
   return logits_shape[1] == 1;
 }
 
+namespace {
+
+// Fallback map for models whose genai_config.json doesn't yet have token IDs in the model section.
```
kunal-vaishnavi: Can we modify this comment to mention that this is a fallback method for Foundry Local specifically? That way, future readers know why the fallback exists.

src/models/model.cpp
```diff
@@ -820,6 +827,64 @@ bool Model::IsPruned() const {
...
+  auto type_it = fallback_map.find(model_type);
+  if (type_it == fallback_map.end()) return nullptr;
+  auto tag_it = type_it->second.find(tag_name);
+  if (tag_it == type_it->second.end()) return nullptr;
+  return &tag_it->second;
+}
+
+}  // namespace
+
+void Model::InitTagIdCache() const {
```
kunal-vaishnavi: Can these token ids be accessed from the tokenizer instead (e.g. `Tokenizer::GetTokenId`)? The BOS, EOS, and PAD token ids are accessed from the tokenizer.

https://github.com/microsoft/onnxruntime-genai/blob/388b009c4c24cbd4a50dee8914b7046300a6487f/src/models/model.cpp#L301-L303

src/models/model.cpp
sayanshaw24: updated!

src/models/model.cpp
sayanshaw24: done!

src/models/model.cpp
```diff
@@ -297,16 +299,63 @@ const std::string& TokenizerStream::Decode(int32_t token) {
   return chunk_;
 }
 
+namespace {
```
kunal-vaishnavi: Why do we need an anonymous namespace? Can we remove it?

src/models/model.cpp
```diff
@@ -297,16 +299,63 @@ const std::string& TokenizerStream::Decode(int32_t token) {
...
   // Resolve tokenizer_dir (may be empty, relative, absolute, or "package:"-scheme).
   const fs::path tokenizer_dir = config.ResolvePath(config.model.tokenizer_dir);
   CheckResult(OrtxCreateTokenizerWithOptions(tokenizer_.Address(), tokenizer_dir.string().c_str(), keys, values, 2));
+
+  // Fallback: if bot/eot/bor/eor were not explicitly set in genai_config.json (-1),
+  // attempt to resolve them by encoding well-known token strings for the model type.
+  // This provides backward compatibility for Foundry Local when consuming older model
+  // packages that predate the bot/eot/bor/eor config fields.
+  if (bot_token_id_ < 0 || eot_token_id_ < 0 || bor_token_id_ < 0 || eor_token_id_ < 0) {
+    auto try_fallback = [&](int32_t& id, const std::string& tag_name) {
```
kunal-vaishnavi: Can all this be merged inside `GetFallbackTag` so that `try_fallback` doesn't need to exist as a method?

src/ort_genai_c.cpp
```diff
@@ -667,6 +667,34 @@ OgaResult* OGA_API_CALL OgaTokenizerGetPadTokenId(const OgaTokenizer* tokenizer,
   OGA_CATCH
 }
 
+OgaResult* OGA_API_CALL OgaTokenizerGetBotTokenId(const OgaTokenizer* tokenizer, int32_t* out) {
```
kunal-vaishnavi, line 687: Can we add the APIs to the following language bindings as well?

- C#
- Java
- Objective-C
- Python

src/models/model.cpp
sayanshaw24: removed - replaced with a static function `ResolveFallbackTagId()` that combines the lookup and resolution in one call.

src/models/model.cpp
sayanshaw24: done - merged into a single static void `ResolveFallbackTagId(int32_t& id, model_type, tag_name, tokenizer)` that handles the full flow (check if already set, lookup fallback string, encode via tokenizer). no more lambda.

src/ort_genai_c.cpp
sayanshaw24: done! each follows the exact same pattern as `GetBosTokenId`/`GetPadTokenId` in that language.

rymikula:
The phi3 entry in the fallback map looks wrong, and the catalog metadata this migration replaces confirms it.

The map resolves phi3 to <tool_call> and </tool_call>. But the current Foundry Local catalog entry for Phi-4-mini-instruct has toolCallStart <|tool_call|> and toolCallEnd <|/tool_call|> (with pipes), and those are what the model's added_tokens.json actually defines (ids 200025 and 200026). The string <tool_call> without pipes does not exist anywhere in Phi-4-mini's vocabulary, so TokenToTokenId("<tool_call>") cannot resolve and the IDs stay -1. Phi-4-mini is exactly the model this fallback exists for: its genai_config.json has model.type "phi3" and no bot/eot token IDs.

Consequence once the catalog metadata is removed (microsoft/Foundry-Local#838): with the IDs unresolved, the enrichment step writes no tool-call marker strings into ModelInfo and no longer marks the model as supporting tool calling. ToolCallStreamAccumulator degrades to passthrough when its markers are empty, so the <|tool_call|>[...] output Phi-4-mini produces under tool_choice "required" would come through as plain text in content with an empty tool_calls array instead of structured tool calls. That path works today because the markers come from the catalog, so this would be a regression.

Suggested fix: map phi3 to <|tool_call|> and <|/tool_call|>, matching the catalog metadata being migrated.

Separate note: under tool_choice "auto", Phi-4-mini emits its tool call as a bare JSON array with no markers at all, because the request's tools are never rendered into its prompt (microsoft/Foundry-Local#874). So marker detection alone cannot cover the auto case even with corrected IDs.

sayanshaw24:
> The phi3 entry in the fallback map looks wrong, and the catalog metadata this migration replaces confirms it.
> 
> The map resolves phi3 to <tool_call> and </tool_call>. But the current Foundry Local catalog entry for Phi-4-mini-instruct has toolCallStart <|tool_call|> and toolCallEnd <|/tool_call|> (with pipes), and those are what the model's added_tokens.json actually defines (ids 200025 and 200026). The string <tool_call> without pipes does not exist anywhere in Phi-4-mini's vocabulary, so TokenToTokenId("<tool_call>") cannot resolve and the IDs stay -1. Phi-4-mini is exactly the model this fallback exists for: its genai_config.json has model.type "phi3" and no bot/eot token IDs.
> 
> Consequence once the catalog metadata is removed ([microsoft/Foundry-Local#838](https://github.com/microsoft/Foundry-Local/pull/838)): with the IDs unresolved, the enrichment step writes no tool-call marker strings into ModelInfo and no longer marks the model as supporting tool calling. ToolCallStreamAccumulator degrades to passthrough when its markers are empty, so the <|tool_call|>[...] output Phi-4-mini produces under tool_choice "required" would come through as plain text in content with an empty tool_calls array instead of structured tool calls. That path works today because the markers come from the catalog, so this would be a regression.
> 
> Suggested fix: map phi3 to <|tool_call|> and <|/tool_call|>, matching the catalog metadata being migrated.
> 
> Separate note: under tool_choice "auto", Phi-4-mini emits its tool call as a bare JSON array with no markers at all, because the request's tools are never rendered into its prompt ([microsoft/Foundry-Local#874](https://github.com/microsoft/Foundry-Local/issues/874)). So marker detection alone cannot cover the auto case even with corrected IDs.

Great catch! Phi-4-mini uses `<|tool_call|>` / `<|/tool_call|>` (ids 200025/200026), not the pipe-less variants. Fixed in the latest commit: `{"phi3", {{"tool_call_start", "<|tool_call|>"}, {"tool_call_end", "<|/tool_call|>"}}}`.

Re: the `tool_choice: "auto"` note - agreed that's a separate issue (tools never rendered into the prompt), tracked in [#874](https://github.com/microsoft/Foundry-Local/issues/874).

kunal-vaishnavi (dismissed):

src/ort_genai.h
```diff
@@ -342,6 +342,32 @@ struct OgaTokenizer : OgaAbstract {
     return token_id;
   }
 
+  // Tool-calling and reasoning token IDs (bot/eot/bor/eor).
+  // Returns -1 if the model does not define the token.
```
baijumeswani: Why return -1 instead of throwing?

test/c_api_tests.cpp
```diff
@@ -1969,6 +1969,74 @@ TEST(CAPITests, ParakeetTdtTranscribeLong) {
   EXPECT_FALSE(transcription.empty());
 }
 
+// Test that bot/eot/bor/eor return -1 for models without these tokens configured
+TEST(CAPITests, TagId_Unknown) {
```
baijumeswani: Change to TokenId to be consistent with the api?

src/config.h
```diff
@@ -145,6 +145,16 @@ struct Config {
...
 
+    // Tool-calling and reasoning token IDs.
+    // Follows the bos/eos/pad naming convention:
+    //   bot = beginning of tool (call), eot = end of tool (call)
+    //   bor = beginning of reasoning,   eor = end of reasoning
+    // -1 means the model does not define this token.
+    int bot_token_id{-1};
+    int eot_token_id{-1};
+    int bor_token_id{-1};
+    int eor_token_id{-1};
```
baijumeswani: Could these be std::optional instead of -1 by default?

src/models/model.cpp
```diff
@@ -306,16 +308,53 @@ const std::string& TokenizerStream::Decode(int32_t token) {
   return chunk_;
 }
 
+// Fallback: if the given token ID is unset (-1), attempt to resolve it by looking up
+// a well-known token string for the model type in the tokenizer vocabulary.
+// This provides backward compatibility for Foundry Local when consuming older model
+// packages that predate the bot/eot/bor/eor config fields.
+// Keyed by model.type string from genai_config.json.
+static void ResolveFallbackTagId(int32_t& id, const std::string& model_type,
```
baijumeswani: Why define the fallback here instead of ort-extensions?

baijumeswani: could this return int instead of void?

baijumeswani: if this absolutely needs to be in ort-genai, can this be moved to a file of its own? Maybe tokenizer_tag_utils.h/cpp?

src/ort_genai.h
sayanshaw24: Good point — changed to throw. The getters now throw `std::runtime_error` if the model doesn't define the token. Callers should only query these if they know the model supports tool calling/reasoning.

test/c_api_tests.cpp
sayanshaw24: Done — renamed tests to `TokenId_Unsupported` and `TokenId_FromConfig`.

src/config.h
sayanshaw24: Done — config fields and Tokenizer members are now `std::optional`. Nullopt = not supported, getters throw.

src/models/model.cpp
sayanshaw24: The fallback map depends on model.type from genai_config.json — a GenAI-level concept that Extensions doesn't have access to. Moving it to Extensions would require threading the model type into the tokenizer construction API plus a cross-repo change + commit update. Kept in GenAI but moved to its own file (`tokenizer_tag_utils.h/cpp`). The GLOB in cmake picks it up automatically.

sayanshaw24: Done — `ResolveFallbackTokenId` now returns `std::optional<int32_t>` instead of mutating by reference.

src/ort_genai.h
baijumeswani: Could you update the comment above in the code?

src/ort_genai_c.h
```diff
@@ -728,6 +728,38 @@ OGA_EXPORT OgaResult* OGA_API_CALL OgaTokenizerGetEosTokenIds(const OgaTokenizer
...
+/**
+ * \brief Return the BOR (beginning of reasoning) token id, or -1 if the model does not define one.
+ * \param[in] tokenizer The tokenizer to read from
+ * \param[out] token_id The BOR token id
+ * \return OgaResult containing the error message if the call fails.
+ */
+OGA_EXPORT OgaResult* OGA_API_CALL OgaTokenizerGetBorTokenId(const OgaTokenizer* tokenizer, int32_t* token_id);
+
+/**
+ * \brief Return the EOR (end of reasoning) token id, or -1 if the model does not define one.
```
baijumeswani: Does this return -1 or return an error?

src/models/tokenizer_tag_utils.cpp
```diff
@@ -0,0 +1,29 @@
...
+      {"phi3", {{"tool_call_start", "<|tool_call|>"}, {"tool_call_end", "<|/tool_call|>"}}},
+      {"gptoss", {{"tool_call_start", "<|start|>"}, {"tool_call_end", "<|call|>"}}},
+  };
+
+  auto type_it = fallback_map.find(model_type);
+  if (type_it == fallback_map.end()) return std::nullopt;
+  auto tag_it = type_it->second.find(tag_name);
+  if (tag_it == type_it->second.end()) return std::nullopt;
+
+  int32_t resolved = tokenizer.TokenToTokenId(tag_it->second.c_str());
```
baijumeswani: Why do we need to do this? Could we not create the map with token ids instead of tags to avoid the call to TokenToTokenId? Since the map is model specific anyway, would it matter if we hardcoded to map to token id instead of the tag?

sayanshaw24: It's a style preference - I went with strings for runtime validation (if a model variant has a different vocab layout, TokenToTokenId returns nullopt rather than silently using a stale hardcoded ID).

The cost is negligible — 4 vocab lookups once at tokenizer construction. Happy to switch to hardcoded IDs if you prefer though.

src/ort_genai_c.h
sayanshaw24: agent missed updating the comments, fixed!

src/ort_genai.h
sayanshaw24: agent missed updating the comments, fixed!

src/models/tokenizer_tag_utils.cpp
baijumeswani: > TokenToTokenId returns nullopt

Could you explain this? `TokenToTokenId` seems to be returning an `int` here. What does nullopt map to?

src/models/model.cpp
```diff
@@ -308,14 +311,44 @@ const std::string& TokenizerStream::Decode(int32_t token) {
...
   // Default tokenizer options
   const char* keys[] = {"add_special_tokens", "skip_special_tokens"};
   const char* values[] = {"false", "true"};
 
   // Resolve tokenizer_dir (may be empty, relative, absolute, or a "sha256:" shared-asset reference).
   const fs::path tokenizer_dir = config.ResolvePath(config.model.tokenizer_dir);
   CheckResult(OrtxCreateTokenizerWithOptions(tokenizer_.Address(), tokenizer_dir.string().c_str(), keys, values, 2));
+
+  // Resolve any unset bot/eot/bor/eor IDs via model-type fallback strings.
+  if (!bot_token_id_) bot_token_id_ = ResolveFallbackTokenId(config.model.type, "tool_call_start", *this);
```
baijumeswani: Any reason to define two different terms? `"tool_call_start"` and `"bot_token_id"`? Could we standardize this in our code and use a single file (probably config.h) to define the names? Are the tag names such as `tool_call_start` standard? Any reason to use them over `bot_token`?

This could become:

```cpp
  if (!bot_token_id_) bot_token_id_ = ResolveFallbackTokenId(config.model.type, Config::Defaults::bot_token_id", *this);
```

src/models/tokenizer_tag_utils.cpp
sayanshaw24: `TokenToTokenId` returns `int32_t` — it returns a negative value when the token string doesn't exist in the vocabulary. This is the established convention from ORT Extensions' underlying tokenizer API (`OrtxTokenize`), where negative = "not found." Since `TokenToTokenId` is already a public API on the Tokenizer, we can't change its return type to `std::optional` without a breaking change. Instead, our wrapper `ResolveFallbackTokenId` converts that convention cleanly: if `TokenToTokenId` returns negative (not in vocab), the wrapper returns `std::nullopt`, leaving the field unset. So the full chain is: fallback map gives us a token string → `TokenToTokenId` tries to resolve it in the vocabulary → negative means not found → wrapper returns `std::nullopt` → the optional field stays empty → getter throws if called.

src/models/model.cpp
sayanshaw24: great catch, updated! the map keys and config field names are now aligned — both use bot/eot/bor/eor. The keys only appear in the fallback map definition and the 4 call sites in the Tokenizer constructor (same file). Since they're purely internal (never exposed in any API or config surface), I think inline string literals are clear enough here without adding a constants header. Happy to add constexpr constants if you prefer though — just feels like indirection for 4 strings in 2 adjacent locations.

i think adding a Config::Defaults namespace with string constants creates indirection for something that's trivially readable inline. It would look like:

```
// config.h
namespace Config::Defaults {
  constexpr const char* bot = "bot";
  constexpr const char* eot = "eot";
  // ...
}
```

src/models/tokenizer_tag_utils.cpp
```diff
@@ -0,0 +1,40 @@
...
+  // qwen2/qwen3| eot | </tool_call>    | 151658
+  // qwen2/qwen3| bor | <think>         | 151667
+  // qwen2/qwen3| eor | </think>        | 151668
+  // phi3        | bot | <|tool_call|>   | 200025
+  // phi3        | eot | <|/tool_call|>  | 200026
+  // clang-format off
+  static const std::unordered_map<std::string, std::unordered_map<std::string, int32_t>> fallback_map = {
+      {"qwen2", {{"bot", 151657}, {"eot", 151658}, {"bor", 151667}, {"eor", 151668}}},
+      {"qwen3", {{"bot", 151657}, {"eot", 151658}, {"bor", 151667}, {"eor", 151668}}},
+      {"phi3",  {{"bot", 200025}, {"eot", 200026}}},
```
baijumeswani: Could you use the names from the config in the map as well so it is cosistent?

baijumeswani (approved):

PR #2216, opened 2026-06-11
Use FoundryLocal-GPU-5090 agent for integration tests

baijumeswani:


baijumeswani:
the agent seems to have long wait times. Will close this for now.

PR #2221, opened 2026-06-11
Add upper bounds to dependencies

apsonawane:
This pull request adds stricter validation for generation parameters in the `Generator` constructor to prevent invalid or out-of-bounds values.

Parameter validation improvements:

* Enforced that `batch_size` must be between 1 and 256, updating the error message to reflect the valid range.
* Added a new check to ensure `num_beams` is also between 1 and 256, throwing a descriptive error if the value is out of bounds.


src/generators.cpp
```diff
@@ -380,8 +380,20 @@ Generator::Generator(const Model& model, const GeneratorParams& params) : model_
     throw std::runtime_error("search max_length is 0");
   if (params.search.max_length > model.config_->model.context_length)
     throw std::runtime_error("max_length (" + std::to_string(params.search.max_length) + ") cannot be greater than model context_length (" + std::to_string(model.config_->model.context_length) + ")");
-  if (params.search.batch_size < 1)
-    throw std::runtime_error("batch_size must be 1 or greater, is " + std::to_string(params.search.batch_size));
+
+  constexpr int kMaxBatchSize = 256;
+  constexpr int kMaxNumBeams = 256;
```
kunal-vaishnavi: How were these values determined? Can we set the defaults to something such as 32 for batch size and beam size?

apsonawane: Updated to 32

kunal-vaishnavi (approved):

PR #2222, opened 2026-06-11
Add path traversal validation

apsonawane:
# Add path traversal validation for `custom_ops_library`

## Summary

Hardens `Model::CreateSessionOptionsFromConfig` in `src/models/model.cpp` so that the `session_options.custom_ops_library` value from `genai_config.json` can only resolve to a shared library inside an explicit allowlist. Prevents a malicious config from pointing `RegisterCustomOpsLibrary` at an arbitrary path on disk.

## Motivation

Before this change, `custom_ops_library` was passed straight through path-resolution and then to `OrtSessionOptions::RegisterCustomOpsLibrary`, which calls `dlopen` / `LoadLibrary`. A crafted config value (e.g. an absolute path to a system DLL, or `../../…/evil.so`) would load arbitrary code during `og.Model(path)` construction.

## Validation rules

The value of `custom_ops_library` is validated in this order. Any failure raises `std::runtime_error`; nothing is loaded.

1. **Path traversal rejected up-front.** If the raw string contains `..`, load fails.
2. Path is classified as **relative** or **absolute** via `std::filesystem::path::is_relative()`.
3. **Relative paths** are joined against each allowed directory in order until one exists on disk; the first match is canonicalized and used.
4. **Absolute paths** are canonicalized as-is.
5. The canonicalized result must sit under one of the allowed directories (prefix match with a directory-separator boundary). Otherwise, load fails.

## Allowed paths

A load succeeds only if the canonical resolved path is inside one of these directories:

| # | Directory | Source | Purpose |
|---|-----------|--------|---------|
| 1 | **GenAI model folder** | `Config::config_path` (the directory passed to `OgaCreateModel`) | Custom op libraries shipped alongside the model |
| 2 | **EP library directories** | Parent directory of every `library_path` in `Ort::GetEpDeviceMetadata` for each `OrtEpDevice` returned by `Ort::GetEpDevices` | Custom ops registered by an installed execution provider |
| 3 | **Current working directory** | `getcwd` / `_getcwd` at the time `Model` is constructed | Development / portable-app scenarios |

Both relative and absolute inputs are supported — the value is normalized with `std::filesystem::canonical` before the prefix check, so symlinks and mixed separators resolve to the on-disk form.

### Examples that load

Assuming model folder `/opt/models/foo`, EP lib dir `/usr/lib/ort_eps`, cwd `/home/user/app`:

| `custom_ops_library` value | Resolves to | Result |
|---|---|---|
| `custom_ops.so` | `/opt/models/foo/custom_ops.so` (if present) | Load |
| `libs/custom_ops.so` | `/opt/models/foo/libs/custom_ops.so` (if present) | Load |
| `custom_ops.so` (only present in cwd) | `/home/user/app/custom_ops.so` | Load |
| `/opt/models/foo/custom_ops.so` | Same, canonicalized | Load |
| `/usr/lib/ort_eps/custom_ops.so` | Same, canonicalized | Load |

## Rejected paths

Loads fail with a `std::runtime_error` for any of these:

| Reason | Example values | Error message |
|---|---|---|
| Contains `..` anywhere in the raw string | `../evil.so`, `libs/../../evil.so`, `..\evil.dll`, `C:\models\..\evil.dll` | `custom_ops_library must not contain path traversal (..): <value>` |
| Absolute path outside every allowed directory | `/etc/malicious.so`, `C:\Windows\System32\evil.dll`, `/tmp/evil.so` | `custom_ops_library path is not within an allowed directory: <value>` |
| Relative path that doesn't exist under any allowed directory | `missing.so` (not in model dir, EP lib dirs, or cwd) | Falls through to the allowlist check with the unresolved value and is rejected |
| Canonicalization fails (e.g. broken symlink to unreachable target) | any path where `std::filesystem::canonical` sets `ec` | `Failed to canonicalize custom_ops_library path: <value>` |
| Symlink whose target lies outside every allowed directory | `custom_ops.so` in model dir → `/tmp/evil.so` | `custom_ops_library path is not within an allowed directory: <resolved target>` |


src/models/model.cpp
```diff
@@ -636,8 +636,18 @@ void Model::CreateSessionOptionsFromConfig(const Config::SessionOptions& config_
 
     std::string custom_library_file_prefix = config_session_options.custom_ops_library.value();
 
-    // If relative path, try to resolve using multiple search locations
+    // Reject absolute paths and any rooted paths (e.g., Windows drive-relative "C:foo.dll"),
```
kunal-vaishnavi: I have seen a mix of absolute and relative paths used for the path to the custom ops library. Is it possible to support both?

apsonawane: Yes, we now support both absolute and relative paths. However, to prevent arbitrary library loading (e.g., a malicious C:\evil\malware.dll), we validate that the final resolved path falls within one of the trusted directories: the model folder, the EP library directory, or the current working directory. Absolute paths that point outside these locations are rejected. Path traversal (..) is also blocked upfront. This gives users flexibility while maintaining security.

kunal-vaishnavi (approved):

PR #2225, opened 2026-06-12
Add check for malformed audio files

apsonawane:
This pull request improves the robustness of audio buffer handling by adding a minimum size check for audio data and introduces a regression test to ensure malformed buffers are properly rejected. These changes prevent crashes due to undersized or malformed audio buffers and improve error reporting.

**Audio buffer validation:**
* Added a minimum buffer size check (`kMinAudioBufferSize = 44`) in `LoadAudiosFromBuffers` to reject audio buffers that are too small to contain a valid audio header, throwing a clear runtime error if violated.

**Testing improvements:**
* Added a regression test (`LoadAudiosFromBuffersRejectsTooSmallBuffer`) to ensure that audio buffers smaller than the minimum required size are rejected with a descriptive error, preventing heap-buffer-overflow crashes.

Onnxruntime-extensions validation check: https://github.com/microsoft/onnxruntime-extensions/pull/1079


src/models/processor.cpp
```diff
@@ -58,9 +58,17 @@ std::unique_ptr<Audios> LoadAudiosFromBuffers(std::span<const void*> audio_data,
   if (audio_data.size() != audio_data_sizes.size())
     throw std::runtime_error("Number of audio data buffers does not match the number of audio data sizes");
 
+  // Conservative minimum buffer size to avoid decoder header parsing reading past the end of the buffer.
+  // 44 bytes matches the standard WAV header size; this is a safety check, not full format validation.
```
kunal-vaishnavi: This is too WAV-specific. The API supports other audio file types so a generic sanity check such as `> 0` may be better. Otherwise, valid inputs may be rejected.

apsonawane: Updated this check to be more generic

kunal-vaishnavi (approved):

PR #2227, opened 2026-06-12
Load models from ONNX Runtime model packages

jambayk:
Adds support for loading a model from an ONNX Runtime *model package* in addition to the existing flat directory layout. A package bundles multiple build variants of the same model and lets ONNX Runtime select the variant matching the local hardware (and any execution provider hints) at load time.

## Public API

* C: `OgaCreateConfig` / `OgaCreateModel` continue to work on flat directories. When given a model package they auto-detect the execution provider; an ambiguous package returns a clear error directing the caller to the new `OgaCreateConfigFromPackage` / `OgaCreateModelFromPackage` entry points, which take an explicit ep argument and require the path to be a model package.
* C++: a two-argument `OgaConfig::Create(path, ep)` / `OgaModel::Create(path, ep)` overload routes through the `FromPackage` entry points.
* Python: `og.Model(path)` / `og.Config(path)` handle both flat directories and packages; an `ep=` keyword selects a variant.

The variant's `genai_config.json` remains the sole source of truth for the providers and provider options used at session creation time. The genai loader does not inject anything based on the package metadata.

## Single-component packages

The genai loader expects exactly one component per package. The component can be named anything; the loader takes whatever name ORT reports. Multi-component packages return a clear error.

## Path resolution

`Config` gains a `package_root` field (empty when not loaded from a package) and a `ResolvePath()` helper that interprets path-shaped strings from `genai_config.json`:

| Form | Resolution |
| --- | --- |
| empty | `config_path` (variant directory) |
| `package:<relative>` | joined with `package_root` (errors when no package root is set) |
| anything else | joined with `config_path` |

A new optional `model.tokenizer_dir` field uses this resolver. The existing `model.vision.config_filename` and `model.speech.config_filename` fields go through it too, including when they fall back to their default basenames.

## Tests + fixtures

* New `test/model_package_test.cpp` covers package detection, single/explicit/ambiguous EP variant selection, `ResolvePath` scheme handling, package-only contract of the `FromPackage` C entry points, and rejection of an ep argument on a flat directory.
* Two tiny hand-authored fixtures under `test/test_models/` (`tiny-cpu.ortpackage`, `tiny-ambiguous.ortpackage`) with placeholder `.onnx` bytes; no real model files are checked in.

## Docs

`docs/model_package.md` walks through the package layout, the `tokenizer_dir` field, the path scheme, and the public API surface.

### DIFF
--- a/src/filesystem.h
+++ b/src/filesystem.h
@@ -29,19 +29,6 @@
 #include <unistd.h>
 #endif
 
-// Internal C++ helpers are not part of the genai DLL's public C API exports on Windows.
-// BUILDING_ORT_GENAI_DLL is set privately on the onnxruntime-genai target so the genai
-// build exports these symbols while consumers (e.g. unit_tests) see dllimport.
-#if defined(_WIN32)
-#ifdef BUILDING_ORT_GENAI_DLL
-#define ORT_GENAI_INTERNAL_API __declspec(dllexport)
-#else
-#define ORT_GENAI_INTERNAL_API __declspec(dllimport)
-#endif
-#else
-#define ORT_GENAI_INTERNAL_API
-#endif
-
 namespace fs {
 
 class path {
--- a/src/filesystem.h
+++ b/src/filesystem.h
@@ -29,6 +29,19 @@
 #include <unistd.h>
 #endif
 
+// Internal C++ helpers are not part of the genai DLL's public C API exports on Windows.
+// BUILDING_ORT_GENAI_DLL is set privately on the onnxruntime-genai target so the genai
+// build exports these symbols while consumers (e.g. unit_tests) see dllimport.
+#if defined(_WIN32)
+#ifdef BUILDING_ORT_GENAI_DLL
+#define ORT_GENAI_INTERNAL_API __declspec(dllexport)
+#else
+#define ORT_GENAI_INTERNAL_API __declspec(dllimport)
+#endif
+#else
+#define ORT_GENAI_INTERNAL_API
+#endif
+
 namespace fs {
 
 class path {
--- a/src/filesystem.h
+++ b/src/filesystem.h
@@ -40,6 +40,11 @@ class path {
 #endif
   };
 
+#ifdef _WIN32
+  // Construct from a wide (UTF-16) path, e.g. an ORTCHAR_T path from ONNX Runtime.
+  path(const std::wstring& wpath) : path_(to_utf8(wpath)), wpath_(wpath) {};
+#endif
+
   static constexpr char separator =
 #ifdef _WIN32
       '\\';
@@ -175,6 +180,31 @@ class path {
     MultiByteToWideChar(codePage, 0, path_.data(), iSource, out.data(), iTarget);
 
     // Return as a string
+    return out;
+  }
+
+  // Convert a wide (UTF-16) path to UTF-8, the inverse of to_wstring().
+  static std::string to_utf8(const std::wstring& wpath) {
+    if (wpath.empty()) {
+      return {};
+    }
+
+    int iSource;  // convert to int because Wc2Mb requires it.
+    SizeTToInt(wpath.size(), &iSource);
+
+    // Ask how much space we will need.
+    SetLastError(0);
+    const auto iTarget =
+        WideCharToMultiByte(CP_UTF8, 0, wpath.data(), iSource, nullptr, 0, nullptr, nullptr);
+
+    size_t cchNeeded;
+    IntToSizeT(iTarget, &cchNeeded);
+
+    // Allocate ourselves some space and convert for real.
+    std::string out;
+    out.resize(cchNeeded);
+    WideCharToMultiByte(CP_UTF8, 0, wpath.data(), iSource, out.data(), iTarget, nullptr, nullptr);
+
     return out;
   }
 #endif  // _WIN32
--- a/src/generators.h
+++ b/src/generators.h
@@ -176,10 +176,8 @@ OrtEnv& GetOrtEnv();
 std::shared_ptr<Model> CreateModel(OrtEnv& ort_env, const char* config_path, const RuntimeSettings* settings = nullptr);
 std::shared_ptr<Model> CreateModel(OrtEnv& ort_env, std::unique_ptr<Config> config);
 
-// Construct a Config from `config_path`. When `config_path` is a model package the
-// package is opened and a variant is selected (auto-detected when `ep` is null or empty,
-// or chosen by the supplied `ep`). When `config_path` is a flat directory the existing
-// path-based Config constructor is used; `ep` must be null/empty in that case.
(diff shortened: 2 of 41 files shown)


src/models/model_package.cpp
```diff
@@ -0,0 +1,175 @@
...
+
+}  // namespace
+
+bool IsModelPackage(const fs::path& path) {
+  std::error_code ec;
+  const std::filesystem::path root = AsStdPath(path);
+  if (!std::filesystem::is_directory(root, ec) || ec) {
+    return false;
+  }
+  return std::filesystem::is_regular_file(root / kManifestFilename, ec) && !ec;
```
baijumeswani, line 67: If the only reference to kManifestFilename is here, can you define it inside this function?

src/models/model_package.h
```diff
@@ -0,0 +1,51 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+#pragma once
+
+#include <string>
+
+#include "../filesystem.h"
+#include "onnxruntime_c_api.h"
```
baijumeswani: Use onnxruntime_api.h that offers RAII instead of using onnxruntime_c_api.h here.

src/models/model_package.h
```diff
@@ -0,0 +1,51 @@
...
+
+// ORT_GENAI_HAS_MODEL_PACKAGE is set when the ORT C API version exposes the
+// OrtModelPackageApi experimental functions (introduced in API version 28) and the
+// experimental header is available on the build's include path. Some Apple/iOS
+// toolchains ship the core C API but no experimental header; those builds compile with
+// model-package support disabled.
+#if !defined(ORT_GENAI_HAS_MODEL_PACKAGE)
+#if defined(ORT_API_VERSION) && ORT_API_VERSION >= 28 && \
+    (!defined(__has_include) || __has_include("onnxruntime_experimental_c_api.h"))
+#define ORT_GENAI_HAS_MODEL_PACKAGE 1
```
baijumeswani: These can be defined in onnxruntime_api.h as HAS_EXPERIMENTAL_API &&ORT_API_VERSION > 28

src/models/model_package.cpp
```diff
@@ -0,0 +1,175 @@
...
+std::string ToLower(std::string s) {
+  std::transform(s.begin(), s.end(), s.begin(),
+                 [](unsigned char c) { return static_cast<char>(std::tolower(c)); });
+  return s;
+}
+
+// Maps a user-supplied EP name to its canonical ORT form for matching against the EP
+// names ORT returns from GetVariantEpName. Accepts short aliases used in genai_config.json
+// (e.g. "cuda") and the full ORT names. Unknown names pass through unchanged.
+std::string NormalizeEpName(const std::string& ep_name) {
```
baijumeswani: this mapping exists in config.h if I remember correctly. Can you ruse that?

src/models/model_package.cpp
```diff
@@ -0,0 +1,175 @@
...
+    std::string ep = pkg_ctx.GetVariantEpName(component_name.c_str(), variant.c_str());
+    if (!ep.empty()) {
+      result.insert(std::move(ep));
+    }
+  }
+  return result;
+}
+
+// Builds a fresh OrtSessionOptions with `ep` appended, used only by ORT to pick a variant.
+// CUDA needs the dedicated V2 entry point; every other EP goes through the generic API.
```
baijumeswani: This is not a good assumption to make. Any reason for not supporting plugin eps here?

src/generators.h
```diff
@@ -175,6 +175,15 @@ OrtEnv& GetOrtEnv();
...
 std::shared_ptr<Model> CreateModel(OrtEnv& ort_env, const char* config_path, const RuntimeSettings* settings = nullptr);
 std::shared_ptr<Model> CreateModel(OrtEnv& ort_env, std::unique_ptr<Config> config);
+
+// Construct a Config from `config_path`. When `config_path` is a model package the
+// package is opened and a variant is selected (auto-detected when `ep` is null or empty,
+// or chosen by the supplied `ep`). When `config_path` is a flat directory the existing
+// path-based Config constructor is used; `ep` must be null/empty in that case.
+std::unique_ptr<Config> CreateConfig(OrtEnv& ort_env, const char* config_path,
+                                     const char* ep = nullptr,
+                                     std::string_view json_overlay = {});
```
baijumeswani, line 183: overlay seems to not be used anywhere in the code.

baijumeswani: sorry, seems like it is in use.

src/filesystem.h
```diff
@@ -29,6 +29,19 @@
 #include <unistd.h>
 #endif
 
+// Internal C++ helpers are not part of the genai DLL's public C API exports on Windows.
+// BUILDING_ORT_GENAI_DLL is set privately on the onnxruntime-genai target so the genai
+// build exports these symbols while consumers (e.g. unit_tests) see dllimport.
```
baijumeswani: Is this only for unit tests? Why is this needed?

src/models/model_package.cpp
```diff
@@ -0,0 +1,131 @@
...
+
+namespace Generators {
+
+namespace {
+
+// fs::path -> std::filesystem::path. The wrapper's c_str() returns the platform-correct
+// UTF-16/UTF-8 form.
+std::filesystem::path AsStdPath(const fs::path& p) {
+  return std::filesystem::path{p.c_str()};
+}
```
baijumeswani, line 23: fs::path has a str() so you should be able to do p.str().c_str().

src/filesystem.h
```diff
@@ -177,6 +182,31 @@ class path {
     // Return as a string
     return out;
   }
+
+  // Convert a wide (UTF-16) path to UTF-8, the inverse of to_wstring().
+  static std::string to_utf8(const std::wstring& wpath) {
```
baijumeswani, line 187: can this be private?

baijumeswani (approved):
A couple of small comments.

src/models/model_package.cpp
jambayk: `fs::path::c_str()` returns the wide (UTF-16) `wchar_t*` on Windows (and `char*`/UTF-8 elsewhere), which `std::filesystem::path` stores natively. `string().c_str()` is always a narrow UTF-8 `char*`; on Windows `std::filesystem::path` reinterprets a narrow string under the active code page (not UTF-8), which corrupts non-ASCII paths — that separate UTF-16 `wpath_` exists precisely to avoid this. So `c_str()` is the correct source here. (Also, the accessor is `string()`, not `str()`.)

src/filesystem.h
jambayk: It already is — `to_utf8` sits under the `private:` label (next to `to_wstring`) and is only used internally by the wide-string constructor.

PR #2232, opened 2026-06-18
Route init-session provider-option shaping through DeviceInterface

qjia7:
## Summary

- Move EP-specific shaping of the `EnsureDeviceOrtInit` trivial-model session options out of `model.cpp` and onto a new optional virtual `DeviceInterface::ShapeInitSessionProviderOptions`. WebGPU and QNN provide overrides; all other EPs inherit the default no-op. The base path becomes EP-agnostic: it looks up the user-supplied entry once, then dispatches.
- Expand the WebGPU global-options whitelist from 7 to 14 entries so every key consumed by `ParseWebGpuContextConfig` (`preserveDevice`, `maxStorageBufferBindingSize`, `maxNumPendingDispatches`, and the four `*BufferCacheMode` keys) actually reaches the process-wide `WebGpuContext` singleton.
- Make `config.h` self-sufficient (adds the standard and project headers it already implicitly depended on) so it can be included from `smartptrs.h` without a transitive chain. A one-line comment in `config.h` warns against introducing a circular include back to `smartptrs.h`/`generators.h`.

## Test plan

- [x] Build `onnxruntime-genai` with `USE_WEBGPU=ON` against a local ORT (`onnxruntime-genai.dll`, `*.pyd`, wheel, `unit_tests.exe`, `model_benchmark.exe` all produced).
- [ ] End-to-end WebGPU smoke: confirm a known-good model still loads and generates correctly via `verify_model_correctness.py` and `verify_multi_gen.py` (covers the original reason `EnsureDeviceOrtInit` exists).
- [ ] QNN compile/load check on hardware (if available) — confirms `enable_htp_shared_memory_allocator=1` is still injected.
- [ ] CPU/CUDA/DML/OpenVINO/RyzenAI/NvTensorRtRtx: compile-only check is sufficient (default no-op virtual; previously never executed the WebGPU/QNN branches either).


src/config.h
```diff
@@ -4,6 +4,16 @@
...
 
+// Do not include "smartptrs.h" or "generators.h" from this header: smartptrs.h
+// includes this file, so the reverse direction would form a cycle.
+#include <optional>
+#include <string>
+#include <string_view>
+#include <utility>
+#include <vector>
+#include "filesystem.h"
+#include "models/onnxruntime_api.h"
```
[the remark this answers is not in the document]
qjia7: Resolved in commit 17745368: the 12-line include dump was removed entirely. `DeviceFilteringOptions` and `ProviderOptions` were extracted to a new self-sufficient `src/provider_options.h` which declares all its own dependencies. `config.h` now just adds `#include "provider_options.h"` and uses transparent type aliases, so `uint32_t`, `std::unordered_map`, etc. are no longer declared in `config.h` itself.

src/config.h
```diff
@@ -4,6 +4,20 @@
 // Portions of this file consist of AI generated content.
 #pragma once
 
+// Do not include "smartptrs.h" or "generators.h" from this header: smartptrs.h
+// includes this file, so the reverse direction would form a cycle.
+#include <cstddef>
+#include <cstdint>
+#include <optional>
+#include <span>
+#include <string>
```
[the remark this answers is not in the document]
qjia7: Resolved in commit 17745368: `#include <span>` was removed from `config.h` along with the rest of the include dump. The project span wrapper (`span.h`) is included only from `provider_options.h` and `smartptrs.h` which already used it, so C++17 toolchains are unaffected.

src/smartptrs.h
```diff
@@ -3,11 +3,14 @@
 //
 // Modifications Copyright(C) 2026 Advanced Micro Devices, Inc. All rights reserved.
 #pragma once
+#include <algorithm>
 #include <assert.h>
 #include <atomic>
 #include <memory>
+#include <type_traits>
```
kunal-vaishnavi: Why is this import needed?

src/smartptrs.h
```diff
@@ -3,11 +3,14 @@
 //
 // Modifications Copyright(C) 2026 Advanced Micro Devices, Inc. All rights reserved.
 #pragma once
+#include <algorithm>
```
kunal-vaishnavi: Why is this import needed?

qjia7: `<algorithm>` is needed for `std::copy` used in the free function `Generators::copy` at the bottom of this file. It was previously pulled in transitively, which Copilot flagged as fragile, so I added it explicitly in commit 28212984. Added a `// for std::copy` marker in d7921352 to match the existing convention used by the other includes in this file.

src/smartptrs.h
qjia7: `<type_traits>` is needed for `std::remove_const_t` used at `DeviceInterface::WrapMemory` (line 117 of this file) to strip `const` before passing to `WrapMemoryBase(void*, ...)`. It was previously pulled in transitively, which Copilot flagged as fragile, so I added it explicitly in commit 28212984. Added a `// for std::remove_const_t` marker in d7921352 to match the existing convention.

kunal-vaishnavi (approved):

PR #2233, opened 2026-06-19
Add base speculative decoding implementation (v0)

samsat701:
### Description

This PR adds speculative decoding to ONNX Runtime GenAI, decoder-only LLMs. A small draft model proposes K candidate tokens autoregressively (cheap); the large target model verifies all K in one parallel forward pass; a rejection-sampling acceptance rule keeps the output distributionally identical to running the target alone. Literature reported speedups are 2-3x for well-matched pairs.

A round consists of:

1. Propose - draft produces K tokens autoregressively
2. Verify - target scores all K in one pass (K+1 distributions)
3. Accept / correct / bonus - walk draft tokens left to right:
- Greedy: accept token i if it equals target's argmax at position i
- Sampling: accept with probability min(1, p_target/p_draft)
- First rejection -> correction token from normalize(max(0, p_target-p_draft))
- All K accepted -> 1 bonus token from the target's trailing distribution
4. Re-anchor: rewind target KV cache to accepted prefix, advance on committed token

Each round commits 1...K+1 tokens from a single target call. A round is computed up front and tokens are buffered and emitted one per GenerateNextToken().

**Public API/config:**
- Model type model.type = "speculative" + sibling model.draft decoder block
- speculative.max_draft_tokens (K), default 4, range [1,16]
- C ABI: OgaGeneratorParams{Set,Get}SpeculativeNumber, OgaGenerator_GetSpeculativeStats + OgaSpeculativeStats
- C++: OgaGeneratorParams::SetSpeculativeOption/GetSpeculativeNumber, OgaGenerator::GetSpeculativeStats
- Python: params.set_speculative_options(...)/get_speculative_options(), generator.get_speculative_stats()
- Stats: rounds, draft proposed/accepted, correction/bonus counts, avg draft & target ms/token, acceptance rate, effective tokens per round

**Main additions:**
- Config, API, sampling math: speculative_sampling.h (ComputeAcceptProb, BuildCorrectionDistribution, SampleFromDistribution, Softmax, SamplingDistributionFromProbs) + the config/ABI/Python surface
- Refactor sampling helpers: nucleus + top-k/top-p moved from search.cpp to shared sampling_distribution.h (ComputeSampledCategorical, SampledCategorical, ScatterToFullVocab); standard decode is behavior-preserving
- Pluggable DecodingStrategy: GenerateNextToken()'s body extracted into StandardDecodingStrategy/TransducerDecodingStrategy, selected via MakeDecodingStrategy; adds GetStats() and Reset() hooks. Behavior-preserving.
- SpeculativeDecodingModel/State: composes target + draft DecoderOnly_Model (each its own cloned Config), shared tokenizer/search; Run() does prefill and caches the draft's pending distribution
- SpeculativeDecodingStrategy + BaseSpeculativeStrategy + tests: the propose->verify->accept->re-anchor loop, one-token-per-call buffering

### Motivation and Context
Currently, LLM inference latency is based on per-token forward pass of a large model. For autoregressive generation of n tokens, we have n expensive forward passes. Many of the tokens are “easy” to predict. A small model would predict the same next token as a large one. We pay full per-token cost for the target model even when a much cheaper model would have produced the same token. ORT-GenAI has no current support for this technique. 

### Known Issues

All guarded with a clear throw (fail fast, never silently wrong). Deliberate v0 cuts:
- Pipeline/multimodal model support
- Guidance/constrained decoding
- repetition_penalty != 1.0 and min_length > 0
- Sliding-window / LFM2 (hybrid) KV cache / legacy combined KV exports (requires separate KV cache format)
- batch_size > 1, num_beams > 1
- Continuous decoding (RewindToLength -> GenerateNextToken isn't supported yet, current workaround is rewind -> AppendTokens (re-prefill) -> GenerateNextToken)
- CPU is main target provider for v0, robust GPU/NPU functionality testing will be conducted in v1
- Cross EP draft/target (eg. draft on CPU, target on GPU) (currently must use the same EP)

### Testing
- C++ unit tests (speculative_sampling_tests.cpp): accept-prob, correction distribution, CDF sampling, shared ComputeSampledCategorical
- Python e2e tests (test_speculative_decoding.py): config guards, self-speculative generation (draft==target so output is comparable to plain greedy), state guards, rewind


samsat701:
@microsoft-github-policy-service agree company="Microsoft"

kunal-vaishnavi:
Here are some Copilot-reviewed comments.


- **Blocking: need statistical equivalence validation for sampling correctness**  
  This change introduces acceptance/rejection correction logic intended to preserve the target distribution, but I don’t see strong statistical validation proving parity vs target-only sampling. Please add a Monte Carlo style test suite (multiple prompts + many seeds) that compares token frequency distributions between speculative and baseline decoding, with explicit coverage for: (1) first rejection early in the K window, (2) rejection late in the window, and (3) all-K-accepted bonus-token path.  
  **Why blocking:** Without empirical equivalence checks, subtle bias can slip in even when formulas look correct.

- **Blocking: prove non-speculative decode behavior is unchanged after sampling refactor**  
  Since top-k/top-p/nucleus helpers were moved out of `search.cpp` into shared sampling utilities, we need parity tests demonstrating standard decoding behavior remains identical. Please add regression tests that assert same outputs/logits decisions for fixed seeds and identical settings before/after refactor (including boundary settings like low `top_p`, small `top_k`, temperature interactions).  
  **Why blocking:** This PR alters shared sampling code used outside speculative mode; behavior drift here would be a silent global regression.

- **Blocking: KV cache re-anchor invariants are under-specified/tested**  
  The rewind/re-anchor step is correctness-critical. Please add focused tests asserting cache/state invariants across mixed accept/reject rounds, including edge cases: reject at token 1, reject at token K, and all accepted + bonus token. Also verify the next target step observes exactly the committed prefix and no stale draft influence.  
  **Why blocking:** Cache misalignment can produce incorrect generations that are hard to detect and debug.

- **Blocking: stats semantics need strict contract + regression tests**  
  New speculative stats are useful, but semantics look potentially ambiguous (e.g., how avg ms/token and effective tokens/round are computed under partial rounds/corrections). Please document exact formulas and add tests that pin expected values for deterministic toy scenarios.  
  **Why blocking:** If these metrics are not stable and precisely defined, perf analysis/monitoring will be misleading and future regressions may go unnoticed.

- **Blocking: unsupported-mode guards must guarantee fail-fast without state corruption**  
  The PR correctly lists unsupported combinations (e.g., `batch_size > 1`, beams, repetition penalty/min_length constraints, cross-EP). Please add tests that each unsupported config throws a clear actionable error *before* generation proceeds, and that generator/model state remains unchanged after the throw (no partial cache/token mutations).  
  **Why blocking:** Partial mutation on failure can create non-deterministic follow-on bugs and invalid outputs.

kunal-vaishnavi:
Here is a Copilot-generated approval checklist for the PR.

---

## ✅ Speculative Decoding (v0)

### 1) Correctness gates (must pass)
- [ ] **Distributional correctness (sampling mode):** Statistical parity validated vs target-only decoding across multiple prompts/seeds (including early reject, late reject, all-accepted+bonus paths).
- [ ] **Greedy correctness:** Accepted token sequence matches target-only greedy decoding for supported settings.
- [ ] **Correction distribution validity:** `normalize(max(0, p_target - p_draft))` behavior tested for numeric stability and edge probabilities.
- [ ] **Acceptance probability validity:** `min(1, p_target / p_draft)` guarded for zero/near-zero draft probs and finite outputs.
- [ ] **Round commit semantics:** Each round commits exactly `1..K+1` tokens, never 0, never > `K+1`.
- [ ] **Buffer emission semantics:** One-token-per-`GenerateNextToken()` contract preserved with correct precomputed buffering.
- [ ] **KV re-anchor invariants:** Rewind/advance logic verified for:
  - [ ] reject at position 1  
  - [ ] reject at position K  
  - [ ] all accepted + bonus token  
- [ ] **No stale state leakage:** Draft/target cache state after correction round is equivalent to committed-prefix target-only progression.
- [ ] **Unsupported mode fail-fast:** Every declared unsupported combo throws deterministic, actionable errors before mutation.

---

### 2) Performance gates (CPU v0 target)
- [ ] **Benchmark methodology documented:** model pair, prompts, token count, warmup, thread config, EP, seed policy.
- [ ] **Baseline vs speculative comparison included:** tokens/sec, latency per token, acceptance rate, effective tokens/round.
- [ ] **Overhead characterization:** performance at low acceptance rates evaluated (including possible fallback guidance).
- [ ] **No major regression in standard decode path:** non-speculative throughput/latency unchanged within agreed tolerance.
- [ ] **Memory overhead reviewed:** dual-model/session/cache impact measured and acceptable for v0 target scenarios.
- [ ] **Perf reproducibility artifact provided:** script/command and expected output format available in repo/docs.

---

### 3) API & UX gates (C ABI, C++, Python)
- [ ] **Config surface complete and consistent:**
  - [ ] `model.type = "speculative"` + sibling `model.draft` schema validated  
  - [ ] `speculative.max_draft_tokens` default/range enforced (`default=4`, `[1,16]`)
- [ ] **Language parity:** C ABI, C++, Python option names and behavior aligned.
- [ ] **Stats API parity:** `GetSpeculativeStats` fields consistently defined across C/C++/Python.
- [ ] **Error messaging quality:** unsupported/invalid configs produce clear remediation guidance.
- [ ] **Backward compatibility:** existing models/configs run unchanged without speculative settings.
- [ ] **Docs updated:** user-facing setup, limitations, and examples added for each surface.

---

### 4) Regression & test gates
- [ ] **Unit test coverage added for speculative math utilities** (`Softmax`, acceptance, correction, sampling helpers).
- [ ] **Behavior-preserving refactor tests** for moved shared sampling logic (top-k/top-p/nucleus parity).
- [ ] **Deterministic seeded tests** for standard and speculative modes where applicable.
- [ ] **Negative tests** for all unsupported combinations listed in PR description.
- [ ] **Integration tests** exercise end-to-end propose→verify→accept/correct→re-anchor loop.
- [ ] **CI matrix coverage** includes required platforms/providers for v0 claims (at minimum CPU; others gated or explicitly deferred).
- [ ] **Flake check:** new tests run stable across repeated CI runs.
- [ ] **No hidden coupling regressions:** transducer/standard strategies validated after DecodingStrategy extraction.

---

### 5) Code quality & maintainability gates
- [ ] **Strategy boundaries are clean:** `Standard/Transducer/Speculative` responsibilities clearly separated.
- [ ] **Naming and comments reflect algorithm paper terms** (propose/verify/accept/correct/bonus/re-anchor).
- [ ] **Invariants documented in code** at critical points (round boundaries, cache length assumptions, token ownership).
- [ ] **Telemetry/stats lifecycle clear:** reset/init semantics tested and documented.
- [ ] **Known v0 constraints tracked:** explicit follow-up issues created for deferred items (GPU/NPU robustness, cross-EP, continuous decoding, etc.).

---

### 6) Merge requirements
- [ ] **All correctness gates pass**
- [ ] **All required tests green**
- [ ] **Perf results attached and acceptable**
- [ ] **API/docs reviewed and approved**
- [ ] **No unresolved blocking review comments**

---

test/python/models/test_speculative_decoding.py
```diff
@@ -0,0 +1,332 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License
+
+"""End-to-end tests for CPU speculative decoding (target verifies draft proposals).
```
kunal-vaishnavi, line 4: The Python test files have been reorganized into different subfolders. Please place this test script in the appropriate subfolder. The [README](https://github.com/microsoft/onnxruntime-genai/blob/main/test/README.md) has more details on the subfolders.

src/models/speculative_decoding.cpp
```diff
@@ -0,0 +1,185 @@
...
+        return false;
+    }
+  }
+
+  return true;
+}
+
+void ValidateLogitsDimensionsMatch(const DecoderOnly_Model& target,
+                                   const DecoderOnly_Model& draft) {
+  const auto& tn = target.config_->model.decoder.outputs.logits;
```
kunal-vaishnavi: Can we expand out the variable names instead of using very compact abbreviations? This will help future readers to understand what the variables represent.

src/models/speculative_decoding.cpp
```diff
@@ -0,0 +1,185 @@
...
+
+  // No support for num_beams > 1 (beam search) in the speculative loop in this release.
+  if (params.search.num_beams != 1)
+    throw std::runtime_error(
+        "Speculative decoding does not support num_beams > 1 (beam search). Got num_beams=" +
+        std::to_string(params.search.num_beams) + ".");
+
+  // No support for repetition_penalty and min_length; 
+  // needs cross-position bookkeeping that isn't implemented yet.
+  if (params.search.repetition_penalty != 1.0f)
```
kunal-vaishnavi: How do you plan to support new search options as they are added to ORT GenAI? We don't want to have to reimplement how each search option is applied in the speculative decoding case where possible.

src/base_speculative_strategy.cpp
```diff
@@ -0,0 +1,116 @@
...
+
+#include "generators.h"
+#include "search.h"
+#include "speculative_sampling.h"
+#include "models/speculative_decoding.h"
+
+namespace Generators {
+
+BaseSpeculativeStrategy::BaseSpeculativeStrategy(Generator& g)
+    : spec_state_{*dynamic_cast<SpeculativeDecodingState*>(g.state_.get())} {}
```
kunal-vaishnavi: `SpeculativeDecodingState&` is a good fit in the header file (required, non-null dependency), but ctor should safely validate the downcast before binding the reference.

The main risk here is in construction: `*dynamic_cast<SpeculativeDecodingState*>(g.state_.get())` dereferences immediately so if wiring is wrong, the current `*dynamic_cast<...>` can dereference null on wiring regressions and this can crash/UB.

Could we make this fail-fast and explicit by checking the cast first, then binding the reference in the constructor? Example pattern: cast to pointer → `if (!ptr) throw std::runtime_error("BaseSpeculativeStrategy requires SpeculativeDecodingState")` → bind/use reference. That preserves the non-null reference design while giving a clear diagnostic instead of a hard crash.

src/config.cpp
```diff
@@ -1208,6 +1209,7 @@ struct Model_Element : JSON::Element {
   Config::Model& v_;
   Encoder_Element encoder_{v_.encoder};
   Decoder_Element decoder_{v_.decoder};
+  Decoder_Element draft_{v_.draft};
```
kunal-vaishnavi: Can we add a comment here that draft is an extension point and uses `Decoder_Element` since it only supports such models currently?

src/decoding_strategy.cpp
```diff
@@ -0,0 +1,31 @@
...
+SpeculativeStats DecodingStrategy::GetStats() const {
+  return SpeculativeStats{};
+}
+
+// Factory
+std::unique_ptr<DecodingStrategy> MakeDecodingStrategy(Generator& generator) {
+  const auto& model_type = generator.model_->config_->model.type;
+  if (ModelType::IsTransducer(model_type))
+    return std::make_unique<TransducerDecodingStrategy>(generator);
+  if (model_type == "speculative")
```
kunal-vaishnavi:
```suggestion
  if (ModelType::IsSpeculative(model_type))
```

src/generators.h
```diff
@@ -44,6 +44,27 @@ struct TransducerState;
...
+
+}  // namespace Generators
+
+#include "decoding_strategy.h"
+
+namespace Generators {
+
+// Instrumentation snapshot returned by Generator::GetSpeculativeStats().
+// All fields are zero for non-speculative models.
+struct SpeculativeStats {
```
kunal-vaishnavi: Can this be defined in a separate header file? This doesn't seem like the place to define these stats. We can import the header as needed.

src/ort_genai_c.h
```diff
@@ -619,6 +637,32 @@ OGA_EXPORT size_t OGA_API_CALL OgaGenerator_GetSequenceCount(const OgaGenerator*
  */
 OGA_EXPORT const int32_t* OGA_API_CALL OgaGenerator_GetSequenceData(const OgaGenerator* generator, size_t index);
 
+/**
+ * \brief POD struct holding speculative-decoding instrumentation.
+ *        All fields are zero for non-speculative generators.
+ */
+typedef struct OgaSpeculativeStats {
```
kunal-vaishnavi: This requires us to define two copies of the same struct (one for C++ and one for C). Can we find a way to only have one copy?

src/speculative_decoding_strategy.h
```diff
@@ -0,0 +1,109 @@
...
+  // (greedy). Otherwise probs[i] is the draft's probability distribution for token i, and the
+  // token is accepted with probability min(1, p_target/p_draft).
+  struct Proposal {
+    std::vector<int32_t> tokens;
+    std::vector<std::vector<float>> probs;
+  };
+
+  // Sampling settings worked out by the base class and passed to Propose. greedy -> accept by
+  // matching the target's argmax (empty probs); else sample from the top-k / top-p distribution.
+  struct SamplingConfig {
```
kunal-vaishnavi: I’m concerned `SpeculativeDecodingStrategy::SamplingConfig` will become a parallel sampling config surface. As we add new knobs, this requires touching the struct and every `Propose(...)` implementation, which can drift from canonical sampling behavior. 

Can we change this to pass in the original config instead of a predefined struct? We can add validation checks for unsupported params.

src/speculative_decoding_strategy.cpp
```diff
@@ -0,0 +1,429 @@
...
+  }
+
+  SamplingConfig cfg{};
+  cfg.greedy = (g.sampling_method_ == Generator::SamplingMethod::kGreedy);
+  cfg.top_k = params.search.top_k;
+  cfg.top_p = params.search.top_p;
+  cfg.temperature = params.search.temperature;
+
+  SampledCategorical sampled;
+  auto to_dist = [&cfg, &sampled, vocab_size](std::span<const float> logits) -> std::vector<float> {
```
kunal-vaishnavi: Can we make this a named helper function instead of an anonymous lambda function?

src/speculative_decoding_strategy.cpp
```diff
@@ -0,0 +1,429 @@
...
+  // K target distributions.
+  const std::string& logits_name =
+      spec_state->spec_model().target_model().config_->model.decoder.outputs.logits;
+  OrtValue* raw_ort = spec_state->target_state().GetOutput(logits_name.c_str());
+  if (!raw_ort)
+    throw std::runtime_error(
+        "Speculative decoding: target state has no logits output named '" + logits_name + "'.");
+
+  auto raw_shape = raw_ort->GetTensorTypeAndShapeInfo()->GetShape();
+  const bool is_multi =
```
kunal-vaishnavi: It seems that `is_multi` means if the target model returned logits for a sequence length greater than 1 in the forward pass. Can we rename this variable to be clearer on what it represents (e.g. `is_multiple_tokens_`)?

src/speculative_decoding_strategy.cpp
```diff
@@ -0,0 +1,429 @@
...
+      (raw_shape.size() >= 2 && raw_shape[1] == static_cast<int64_t>(verify_width));
+  is_multi_ = is_multi;
+
+  // Runtime vocab-size safety net (once per generator lifetime).
+  if (!vocab_check_done_) {
```
kunal-vaishnavi, line 434: Can this be done when the ONNX models are created? The output buffers are pre-allocated ahead of time and the logits are of shape `[batch_size, sequence_length, vocab_size]`, so the dimensions of the output tensors should be known before running a round.

src/speculative_sampling.h
```diff
@@ -0,0 +1,103 @@
...
+#include <limits>
+#include <random>
+#include <vector>
+#include "span.h"
+#include "sampling_distribution.h"
+
+namespace Generators {
+
+// Numerically-stable softmax over a contiguous fp32 row.
+inline std::vector<float> Softmax(std::span<const float> logits) {
```
kunal-vaishnavi: There are existing files for calculating softmax. Let's re-use existing implementations rather than adding new ones.

src/speculative_sampling.h
```diff
@@ -0,0 +1,103 @@
...
+    // Distributions matched exactly (shouldn't happen on a rejection): fall back to target
+    for (size_t i = 0; i < out.size(); i++) {
+      out[i] = p_target[i];
+    }
+  }
+}
+
+// Input: distribution probs; rng state
+// Output: index of sampled token
+inline int SampleFromDistribution(std::span<const float> probs, std::mt19937& rng) {
```
kunal-vaishnavi: Same comment as [here](https://github.com/microsoft/onnxruntime-genai/pull/2233#discussion_r3498103202) about re-using existing code

src/transducer_decoding_strategy.cpp
```diff
@@ -0,0 +1,19 @@
...
+// Licensed under the MIT License.
+#include "transducer_decoding_strategy.h"
+
+#include "generators.h"
+#include "models/transducer_state.h"
+
+namespace Generators {
+
+TransducerDecodingStrategy::TransducerDecodingStrategy(Generator& g)
+    : transducer_state_{*dynamic_cast<TransducerState*>(g.state_.get())} {}
```
kunal-vaishnavi: Same comment as [here](https://github.com/microsoft/onnxruntime-genai/pull/2233#discussion_r3497647389)

test/python/models/test_speculative_decoding.py
samsat701: Done in https://github.com/microsoft/onnxruntime-genai/pull/2233/changes/1a69ef32d2780e7867e528a37e06e1c102b38f54 - moved to `test/python/models/test_speculative_decoding.py` to follow the `test/README.md` convention (`test/python/models/*` holds `test_*.py` scripts that test ONNX models). Verified all 21 tests still collect from the new location.

src/models/speculative_decoding.cpp
samsat701: Done in https://github.com/microsoft/onnxruntime-genai/pull/2233/changes/d762196245d470b59e1a8327c1fb5c803776db54

src/base_speculative_strategy.cpp
samsat701: Done in https://github.com/microsoft/onnxruntime-genai/pull/2233/changes/e4120153cb694f4a842bb3892472d2723093077e - added a static RequireSpeculativeState(Generator&) helper that casts to a pointer -> throws "BaseSpeculativeStrategy requires SpeculativeDecodingState" if it's null -> returns the dereferenced reference.

src/config.cpp
samsat701: Done in https://github.com/microsoft/onnxruntime-genai/pull/2233/changes/a625891925324b0ba9a68b7119e7fbbc9330378b

src/generators.h
samsat701: Done https://github.com/microsoft/onnxruntime-genai/pull/2233/changes/d45e90f8ac00a430cc524f6aa725d9030e87f280 - moved SpeculativeStats into dedicated src/speculative_stats.h.

src/speculative_decoding_strategy.h
samsat701: Fixed - https://github.com/microsoft/onnxruntime-genai/pull/2233/changes/5f0ce2840667a6fcf8bf9a9e6f7fcf2f48ebf7a0

Removed the SamplingConfig struct. Propose/RunRound now read the canonical config directly (g.search_->params_->search for top_k/top_p/temperature). New public Generator::IsGreedySampling() accessor for the greedy-vs-sample decision.

src/speculative_decoding_strategy.cpp
samsat701: Fixed - https://github.com/microsoft/onnxruntime-genai/pull/2233/changes/2d888909c08b8d393e320bcc0fa31d1a365f6157

Created new helper function and updated calls.

src/speculative_decoding_strategy.cpp
samsat701: Fixed- https://github.com/microsoft/onnxruntime-genai/pull/2233/changes/ed8cdbf664c0d680e8bda09d68aa070d18d6a6c9

src/speculative_sampling.h
samsat701: Fixed - https://github.com/microsoft/onnxruntime-genai/pull/2233/changes/9f323e4dabb5087107f4c317e8f8ac2fc95256e0

Removed the SampleFromDistribution wrapper and inlined std::discrete_distribution at the four call sites (matching search.cpp); dropped the 3 unit tests that only exercised std::discrete_distribution

src/speculative_sampling.h
samsat701: Fixed - https://github.com/microsoft/onnxruntime-genai/commit/9f323e4dabb5087107f4c317e8f8ac2fc95256e0

Removed the speculative softmax reimplementation; call the existing Softmax(span, temperature) from softmax.h directly (inlined).

src/transducer_decoding_strategy.cpp
samsat701: Done - https://github.com/microsoft/onnxruntime-genai/pull/2233/changes/a3e157f657179a15319460c536379fc5525e2c00

Applied the same fix as in BaseSpeculativeStrategy; added a static RequireTransducerState(Generator&) helper that casts to a pointer -> throws "TransducerDecodingStrategy requires TransducerState" if null -> returns the dereferenced reference.

src/models/speculative_decoding.cpp
```diff
@@ -0,0 +1,189 @@
...
+  if (uses_combined_kv(config_->model.decoder) || uses_combined_kv(config_->model.draft))
+    throw std::runtime_error(
+        "Speculative decoding requires decoder-only target and draft models that use the separate "
+        "key/value KV-cache format (past_key_names/past_value_names). Combined-KV / legacy formats "
+        "such as the original gpt2 graph (past_%d/present_%d) are not supported in this release.");
+
+  // Sliding-window and LFM2 caches discard old K/V by design, so RewindTo throws.
+  // Speculative decoding rewinds on every rejection, so reject these up front (checked on both target and draft).
+  auto uses_sliding_kv = [](const Config::Model::Decoder& d) {
+    return d.sliding_window.has_value() && d.sliding_window->slide_key_value_cache;
```
kunal-vaishnavi: The `sliding_window` dict is only used on the TRT-RTX EP.

https://github.com/microsoft/onnxruntime-genai/blob/0be0a44cd45833e03219be9d84da7ab03b1a7b31/src/python/py/models/builders/base.py#L785-L796

Can you use other identifiers for sliding window attention on other EPs?

src/generators.h
```diff
@@ -44,6 +45,13 @@ struct TransducerState;
 struct Search;
 struct Tokenizer;
 struct ConstrainedLogitsProcessor;
+
+}  // namespace Generators
+
+#include "decoding_strategy.h"
```
kunal-vaishnavi: This seems odd. Why not put the header at the top of the file? Why re-initialize the Generators namespace?

src/ort_genai_c.cpp
```diff
@@ -628,6 +642,23 @@ const int32_t* OGA_API_CALL OgaGenerator_GetSequenceData(const OgaGenerator* gen
   return generator->GetSequence(static_cast<int>(index)).CopyDeviceToCpu().data();
 }
 
+OgaResult* OGA_API_CALL OgaGenerator_GetSpeculativeStats(const OgaGenerator* generator, OgaSpeculativeStats* out_stats) {
```
kunal-vaishnavi: Can we change this to use `ReturnUnique` like the following?

https://github.com/microsoft/onnxruntime-genai/blob/236a7000bcf73c47faf8ce3550c6349c14755ecb/src/ort_genai_c.cpp#L245-L251

src/speculative_decoding_strategy.h
```diff
@@ -0,0 +1,106 @@
...
+  // Update the draft model's own state (its KV cache, n-gram tables, ...) after the base class
+  // commits the accepted tokens. n_direct = how many proposed tokens were accepted (excludes the
+  // correction/bonus token).
+  virtual void Advance(Generator& g,
+                       const Proposal& proposal,
+                       int n_direct,
+                       int32_t final_token,
+                       int seed_length) = 0;
+
+  // Stats accumulators.
```
kunal-vaishnavi: Why duplicate if `speculative_stats.h` already exists?

kunal-vaishnavi:
From Copilot:

- **Blocking: add explicit statistical parity evidence for sampling-mode correctness (target-only vs speculative).**  
  The new unit/e2e tests are strong, but this PR still needs a clear, reproducible parity report that demonstrates speculative decoding remains distributionally equivalent in practice across multiple prompts and seeds. Please attach results (or CI artifact) that compare token frequency distributions for target-only vs speculative and explicitly cover: (1) early rejection, (2) late rejection, and (3) all-accepted + bonus path.

- **Blocking: prove standard (non-speculative) decoding behavior is unchanged after shared sampling refactor.**  
  Since `search.cpp` now routes top-k/top-p/top-k+top-p through `sampling_distribution.h`, we need explicit regression parity evidence for standard decoding to avoid silent global drift. Please add/attach deterministic before-vs-after parity outputs for representative boundary settings (e.g., low top_p, top_k=1/2, high/low temperature) and fixed seeds.

- **Blocking: pin speculative stats semantics with deterministic contract tests and docs alignment.**  
  Stats are now well-structured, but fields like `effective_speedup`, `avg_target_ms_per_token`, and fold-path behavior can still be misinterpreted without strict, user-facing contract guarantees. Please add/point to deterministic tests and docs that unambiguously validate formulas and edge behavior (including when fold path is active and re-anchor timing inputs differ), so telemetry is stable and trustworthy for perf decisions.

src/models/model.cpp
```diff
@@ -851,6 +852,9 @@ std::unique_ptr<Config> CreateConfig(OrtEnv& ort_env, const char* config_path, c
 }
 
 std::shared_ptr<Model> CreateModel(OrtEnv& ort_env, std::unique_ptr<Config> config) {
+  // Speculative decoding wraps two decoder-only models (target + draft) under single model.
+  if (config->model.type == "speculative")
```
baijumeswani: Can we determine the model type from the config instead of the model.type?

src/config.h
```diff
@@ -405,6 +405,8 @@ struct Config {
 
     } decoder;
 
+    Decoder draft;
```
baijumeswani: Can this be std::optional?

src/models/speculative_decoding.cpp
samsat701: Fixed in https://github.com/microsoft/onnxruntime-genai/pull/2233/changes/9b9de62635bba2f8cf24b79b86343b1d65a51435. Speculative decoding now rejects only physically sliding, non-rewindable KV caches. Other sliding-window attention implementations remain supported.

src/generators.h
samsat701: Fixed in https://github.com/microsoft/onnxruntime-genai/pull/2233/changes/9b9de62635bba2f8cf24b79b86343b1d65a51435. Moved decoding_strategy.h to the standard include section and removed the unnecessary namespace reopening.

src/ort_genai_c.cpp
samsat701: Fixed in https://github.com/microsoft/onnxruntime-genai/pull/2233/changes/9b9de62635bba2f8cf24b79b86343b1d65a51435. Speculative stats now use an owned opaque object created with ReturnUnique (with destroy and accessor APIs).

src/speculative_decoding_strategy.h
samsat701: Fixed in https://github.com/microsoft/onnxruntime-genai/pull/2233/changes/9b9de62635bba2f8cf24b79b86343b1d65a51435. Consolidated counters in SpeculativeStats; strategy now keeps only the timing accumulators needed for derived values.

src/csharp/Generator.cs
```diff
@@ -89,6 +89,12 @@ public ReadOnlySpan<int> GetSequence(ulong index)
             }
         }
 
+        public SpeculativeStats GetSpeculativeStats()
```
baijumeswani, line 96: Could we add code comments explaining what these functions do?

src/csharp/GeneratorParams.cs
```diff
@@ -46,6 +46,17 @@ public bool GetSearchBool(string searchOption)
             return value;
         }
 
+        public void SetSpeculativeNumber(string name, double value)
```
baijumeswani, line 54: Could we add code comments explaining what these functions do?

src/csharp/GeneratorParams.cs
```diff
@@ -46,6 +46,17 @@ public bool GetSearchBool(string searchOption)
...
+        {
+            Result.VerifySuccess(NativeMethods.OgaGeneratorParamsSetSpeculativeNumber(_generatorParamsHandle, StringUtils.ToUtf8(name), value));
+        }
+
+        public double GetSpeculativeNumber(string name)
```
baijumeswani, line 64: Could we add code comments explaining what these functions do?

src/csharp/NativeMethods.cs
```diff
@@ -186,6 +196,28 @@ internal class NativeLib
...
+
+        [DllImport(NativeLib.DllName, CallingConvention = CallingConvention.Winapi)]
+        public static extern IntPtr /* OgaResult* */ OgaSpeculativeStatsGetNumber(IntPtr /* const OgaSpeculativeStats* */ stats,
+                                                                                 byte[] /* const char* */ name,
+                                                                                 out double value);
+
+        [DllImport(NativeLib.DllName, CallingConvention = CallingConvention.Winapi)]
+        public static extern IntPtr /* OgaResult* */ OgaSpeculativeStatsGetBool(IntPtr /* const OgaSpeculativeStats* */ stats,
+                                                                               byte[] /* const char* */ name,
+                                                                               out bool value);
```
baijumeswani: Returning bools can be complicated. See https://github.com/microsoft/onnxruntime-genai/pull/2327 on how it should be done.

src/csharp/SpeculativeStats.cs
```diff
@@ -0,0 +1,49 @@
...
+    public sealed class SpeculativeStats : SafeHandle
+    {
+        internal SpeculativeStats(IntPtr nativeHandle) : base(IntPtr.Zero, true)
+        {
+            SetHandle(nativeHandle);
+        }
+
+        public override bool IsInvalid => handle == IntPtr.Zero;
+
+        public ulong GetCount(string name)
```
baijumeswani, line 33: Please add cpde comments for each function regardless of how trivial it might be.

src/cuda/search_cuda.h
```diff
@@ -52,10 +52,15 @@ struct GreedySearch_Cuda : Search_Cuda {
   void SampleTopK(int k, float t) override { SampleTopKTopP(k, 1.0, t); }
   void SampleTopP(float p, float t) override { SampleTopKTopP(-1, p, t); }
   void SampleTopKTopP(int k, float p, float t) override;
+  void CommitToken(int32_t token) override;
```
baijumeswani, line 64: CommitToken seems to not be called from outside search. Does it need to be public?

src/cuda/search_cuda.cpp
```diff
@@ -160,24 +160,41 @@ void GreedySearch_Cuda::SampleTopKTopP(int k, float p, float temperature) {
...
+  FinalizeGeneratedTokens();
+}
 
-  // Append tokens
-  CUDA_CHECK(cudaStreamSynchronize(GetStream()));
-  if (!*done_cpu_) {
-    cuda::Launch_AppendNextTokensToSequences(next_tokens_buffer_.Span(), sequences_.GetSequences().Span(), params_->BatchBeamSize(), sequences_.GetSequenceLength(), sequences_.max_length_, GetStream());
-    sequences_.AfterAppendNextTokens(next_tokens_buffer_, params_->BatchBeamSize());
-  }
+void GreedySearch_Cuda::AppendTokensToSequences(DeviceSpan<int32_t>& tokens) {
```
baijumeswani, line 235: Are the changes in search and sampling purely a refactor or are there meaningful changes here? If there are, can you explain why the changes were needed?

src/models/model.h
```diff
@@ -38,8 +38,8 @@ struct State {
 
   void ClearIO();  // Clear all inputs/outputs
 
-  void SetActiveAdapter(Adapters* adapters, const std::string& adapter_name);
-  void SetRunOption(const char* key, const char* value);
+  virtual void SetActiveAdapter(Adapters* adapters, const std::string& adapter_name);
+  virtual void SetRunOption(const char* key, const char* value);
```
baijumeswani, line 42: Why was this change needed?

src/models/model_type.h
```diff
@@ -61,6 +61,16 @@ struct ModelType {
...
 
+  // Speculative decoding wraps two decoder-only models (target + draft) under a single model.
+  inline static bool IsSpeculative(const std::string& model_type) {
+    return model_type == "speculative";
+  }
+
+  inline static bool UsesDraftModelSpeculation(const std::string& model_type,
+                                               const std::string& draft_filename) {
+    return IsSpeculative(model_type) || !draft_filename.empty();
+  }
```
baijumeswani: One of the two checks should be sufficient. Probably the latter. We should not need IsSpeculative. Although, the name `IsSpeculative` is more apt for this.

src/models/speculative_decoding.cpp
```diff
@@ -0,0 +1,224 @@
...
+    throw std::runtime_error(
+        "model.type is \"speculative\" but model.draft.filename is not set in genai_config.json.");
+
+  if (!ProviderConfigurationMatches(config_->model.decoder.session_options,
+                                    config_->model.draft.session_options))
+    throw std::runtime_error(
+        "Target and draft must use the same execution provider. "
+        "Cross-EP speculative decoding is not supported in this release.");
+
+  // v0 scope: target and draft must be plain decoder-only LLMs. No Pipeline/multimodal (reject upfront).
```
baijumeswani: Avoid planning/design comments in the code.

src/models/speculative_decoding.cpp
```diff
@@ -0,0 +1,224 @@
...
+  return true;
+}
+
+void ValidateLogitsDimensionsMatch(const DecoderOnly_Model& target,
+                                   const DecoderOnly_Model& draft) {
+  const auto& target_logits_name = target.config_->model.decoder.outputs.logits;
+  const auto& draft_logits_name = draft.config_->model.decoder.outputs.logits;
+  if (!target.session_info_.HasOutput(target_logits_name) ||
+      !draft.session_info_.HasOutput(draft_logits_name))
+    return;
```
baijumeswani: Is this not an error condition? Should it not throw here? Same question for all the remainder of early returns from this function.

Or can you specify in code comments when ValidateLogitsDimensionsMatch should throw and when it should not?

src/models/speculative_decoding.cpp
```diff
@@ -0,0 +1,224 @@
...
+  // The inner states are constructed directly as DecoderOnly_Model, which requires the modern
+  // separate key/value KV-cache format (past_key_names/past_value_names).
+  auto uses_combined_kv = [](const Config::Model::Decoder& d) {
+    return !d.inputs.past_names.empty() || !d.outputs.present_names.empty();
+  };
+  if (uses_combined_kv(config_->model.decoder) || uses_combined_kv(config_->model.draft))
+    throw std::runtime_error(
+        "Speculative decoding requires decoder-only target and draft models that use the separate "
+        "key/value KV-cache format (past_key_names/past_value_names). Combined-KV / legacy formats "
+        "such as the original gpt2 graph (past_%d/present_%d) are not supported in this release.");
```
baijumeswani: I think you can safely assume that DecoderOnlyModels use the default key value cache. This check should not be necessary.

src/models/speculative_decoding.cpp
```diff
@@ -0,0 +1,224 @@
...
+  // Reject only windowed caches that discard KV history.
+  if (UsesNonRewindableWindowedKeyValueCache(*this, config_->model.decoder) ||
+      UsesNonRewindableWindowedKeyValueCache(*this, config_->model.draft))
+    throw std::runtime_error(
+        "Speculative decoding does not support physically sliding KV caches in this release; "
+        "WindowedKeyValueCache cannot rewind after discarding KV history.");
+  if (!config_->model.decoder.layer_types.empty() || !config_->model.draft.layer_types.empty())
+    throw std::runtime_error(
+        "Speculative decoding does not support LFM2 (hybrid SSM/attention) models in this release; "
+        "their rolling convolution state cannot be rewound.");
```
baijumeswani: Similarly, this check should not be necessary. DecoderOnlyModels do not construct the windowed key value caches.

src/models/speculative_decoding.cpp
```diff
@@ -0,0 +1,224 @@
...
+  ValidateLogitsDimensionsMatch(*target_model_, *draft_model_);
+  session_info_.Add(*target_model_->session_decoder_);
+}
+
+std::unique_ptr<State> SpeculativeDecodingModel::CreateState(DeviceSpan<int32_t> sequence_lengths,
+                                                             const GeneratorParams& params) const {
+  return std::make_unique<SpeculativeDecodingState>(*this, sequence_lengths, params);
+}
+
+// SpeculativeDecodingState — construction
```
baijumeswani: I think we can avoid such comments since the constructor definition is obvious.

baijumeswani (commented):
First rounds of review comments. Will do another pass for the remainder of the files soon.

src/csharp/NativeMethods.cs
samsat701: Fixed in https://github.com/microsoft/onnxruntime-genai/pull/2233/changes/8b41245f228858fd2fe5aa0e726a418feb00d71d. Added [MarshalAs(UnmanagedType.I1)].

src/models/model.cpp
samsat701: Addressed. Speculative decoding is now determined from the presence of the configured model.draft.filename, rather than requiring model.type == "speculative". This allows existing model types such as phi3 to enable speculative decoding without replacing their original model type (and affecting rope factor handling).

src/config.h
samsat701: Fixed in - https://github.com/microsoft/onnxruntime-genai/pull/2233/changes/76bfd1df70394be8fc020acd31d0da9b9997bd7e. Changed model.draft to std::optional<Decoder>. 

The config parser now lazily constructs the draft decoder only when a model.draft object is present and all draft-model consumers handle the optional value explicitly.

src/csharp/Generator.cs
samsat701: Fixed in https://github.com/microsoft/onnxruntime-genai/pull/2233/changes/76bfd1df70394be8fc020acd31d0da9b9997bd7e.

src/csharp/GeneratorParams.cs
samsat701: Fixed in https://github.com/microsoft/onnxruntime-genai/pull/2233/changes/76bfd1df70394be8fc020acd31d0da9b9997bd7e.

src/csharp/GeneratorParams.cs
samsat701: Fixed in https://github.com/microsoft/onnxruntime-genai/pull/2233/changes/76bfd1df70394be8fc020acd31d0da9b9997bd7e.

src/models/speculative_decoding.cpp
samsat701: Fixed in https://github.com/microsoft/onnxruntime-genai/pull/2233/changes/76bfd1df70394be8fc020acd31d0da9b9997bd7e.

src/models/speculative_decoding.cpp
samsat701: Fixed in https://github.com/microsoft/onnxruntime-genai/pull/2233/changes/76bfd1df70394be8fc020acd31d0da9b9997bd7e.

src/csharp/SpeculativeStats.cs
samsat701: Fixed in https://github.com/microsoft/onnxruntime-genai/pull/2233/changes/76bfd1df70394be8fc020acd31d0da9b9997bd7e.

src/models/model_type.h
samsat701: Fixed in - https://github.com/microsoft/onnxruntime-genai/pull/2233/changes/76bfd1df70394be8fc020acd31d0da9b9997bd7e. Removed IsSpeculative and UsesDraftModelSpeculation. Draft-model speculative decoding is now selected solely when model.draft is configured (the original model architecture type is preserved).

src/models/speculative_decoding.cpp
samsat701: Fixed in https://github.com/microsoft/onnxruntime-genai/pull/2233/changes/80c2dc566c526082dc4c0b6811555892e7bcdc69. Removed all early returns from logits validation. Missing outputs, rankless outputs, and non-static or non-positive vocabulary dimensions now throw actionable errors. 

Only success when both concrete vocabulary dimensions match. Added tests for missing logits outputs and mismatched vocabulary dimensions.

src/models/speculative_decoding.cpp
samsat701: Fixed in https://github.com/microsoft/onnxruntime-genai/pull/2233/changes/80c2dc566c526082dc4c0b6811555892e7bcdc69. Removed the speculative-specific combined/legacy KV-cache precheck and its corresponding tests.

src/models/speculative_decoding.cpp
samsat701: Rechecked this against the current implementation. DecoderOnly_State calls CreateKeyValueCache, which explicitly constructs WindowedKeyValueCache when slide_key_value_cache is enabled. WindowedKeyValueCache::RewindTo throws, while speculative reconciliation requires rewinding both child states (same applies to LFM2Cache, state cannot be rewound). 

Should we include this as speculation requires rewind (better to fail during model creation)?

src/cuda/search_cuda.h
samsat701: CommitToken is called outside search implementation by SpeculativeDecodingStrategy::EmitToken through Search interface. 

Spec decoding has already selected the accepted/correction token (needs public search operation that commits the token while we preserve the relevant bookkeeping for EOS/max-len).

src/cuda/search_cuda.cpp
samsat701: Have mix of refactoring + meaningful changes here:
- Added CommitToken so speculative decoding can append an already-selected token (while reusing EOS/max-len handling)
- Extracted CPU top-k/top-p sampling logic into shared helper so normal/speculative decoding calculate probabilities consistently
- Shared min-length and repetition penalties helpers with speculative decoding
- Factored out finalization logic so normal sampling and CommitToken reuse it

src/models/model.h
samsat701: Speculative decoding has 2 model states (target/draft). Methods kept virtual so speculative state can forward the adapter to the target and runtime options to both models (without it, options would only be applied to outer wrapper + not reach models that run inference).

src/models/speculative_decoding.h
```diff
@@ -0,0 +1,77 @@
...
+  // Accessors for the strategy layer.
+  const SpeculativeDecodingModel& spec_model() const { return model_; }
+  State& target_state() { return *target_state_; }
+  State& draft_state() { return *draft_state_; }
+
+  const std::vector<float>& draft_pending_logits() const { return draft_pending_logits_; }
+  void set_draft_pending_logits(std::vector<float> logits) {
+    draft_pending_logits_ = std::move(logits);
+    draft_pending_valid_ = true;
+  }
```
baijumeswani: Is this function used? Could we remove it if it is not needed?

src/speculative_sampling.h
```diff
@@ -0,0 +1,117 @@
...
+                                                        int top_k, float top_p,
+                                                        float temperature) {
+  std::vector<float> scores(probs.size());
+  for (size_t i = 0; i < probs.size(); i++) {
+    scores[i] = std::log(std::max(probs[i], std::numeric_limits<float>::min()));
+  }
+  SampledCategorical c;
+  ComputeSampledCategorical({scores.data(), scores.size()}, top_k, top_p, temperature, c);
+  return ScatterToFullVocab(c, static_cast<int>(probs.size()));
+}
```
baijumeswani: Is this function used? If not, can we remove it? Or maybe we could move it to the tests file where it is in use?

baijumeswani (dismissed):
Few small comments.

src/models/speculative_decoding.cpp
baijumeswani: Let's leave this as is.

src/csharp/NativeMethods.cs
```diff
@@ -130,6 +130,16 @@ internal class NativeLib
                                                                                      byte[] /* const char* */ searchOption,
                                                                                      [MarshalAs(UnmanagedType.I1)] out bool /* const bool* */ value);
 
+        [DllImport(NativeLib.DllName, CallingConvention = CallingConvention.Winapi)]
+        public static extern IntPtr /* OgaResult* */ OgaGeneratorParamsSetSpeculativeNumber(IntPtr /* OgaGeneratorParams* */ generatorParams,
+                                                                                           byte[] /* const char* */ name,
```
kunal-vaishnavi, line 135: This doesn't have to be done in this PR but we should fix the parameter spacings in these C# methods to mirror the other methods.

src/cuda/search_cuda.cpp
kunal-vaishnavi: We should add docs explaining how the new internal structuring of ORT GenAI works in a separate PR.

src/models/speculative_decoding.cpp
```diff
@@ -0,0 +1,224 @@
...
+    return output;
+  return draft_state_->GetOutput(name);
+}
+
+void SpeculativeDecodingState::SetActiveAdapter(Adapters* adapters, const std::string& adapter_name) {
+  // Apply the adapter only to the target; the draft may be a different, incompatible model.
+  target_state_->SetActiveAdapter(adapters, adapter_name);
+}
+
+void SpeculativeDecodingState::SetRunOption(const char* key, const char* value) {
```
kunal-vaishnavi, line 198: This method looks like it can be simplified but you can do it in another PR.

src/models/speculative_decoding.h
kunal-vaishnavi: Looks like it is unused and can be removed.

src/models/speculative_decoding.cpp
```diff
@@ -0,0 +1,224 @@
...
+    throw std::runtime_error(
+        "Target and draft logit dimensions don't match. Target vocab: " +
+        std::to_string(target_vocab_size) + ", Draft vocab: " + std::to_string(draft_vocab_size) +
+        ". Target and draft must share the same vocabulary.");
+}
+
+}  // namespace
+
+// SpeculativeDecodingModel
+SpeculativeDecodingModel::SpeculativeDecodingModel(std::unique_ptr<Config> config, OrtEnv& ort_env)
```
kunal-vaishnavi, line 106: From Copilot:

Could we validate all unsupported generator settings before constructing `target_state_` and `draft_state_`?

The current initializer list creates both child states—and therefore initializes their model state/KV caches—before the constructor body checks `batch_size`, `num_beams`, and `no_repeat_ngram_size`. If one of these settings is unsupported, construction can perform substantial partial initialization before throwing, which does not satisfy the fail-fast/no-state-mutation contract described by the PR.

Please move these validations ahead of child-state construction, for example into a pre-validation helper used before the initializer list, and add negative tests that verify each unsupported configuration throws without partially initializing either child state or its KV cache.

src/config.cpp
```diff
@@ -1318,6 +1324,30 @@ struct Search_Element : JSON::Element {
...
+struct Speculative_Element : JSON::Element {
+  explicit Speculative_Element(Config::Speculative& v) : v_{v} {}
+
+  // K (draft tokens per round) must be within [kMinK, kMaxK].
+  static constexpr int kMinK = 1;
+  static constexpr int kMaxK = 16;
+
+  void OnValue(std::string_view name, JSON::Value value) override {
+    if (name == "max_draft_tokens") {
+      int k = static_cast<int>(JSON::Get<double>(value));
```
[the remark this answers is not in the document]
kunal-vaishnavi: This is a valid issue. We need this change or else an IcM ticket will get raised.

src/speculative_decoding_strategy.cpp
```diff
@@ -0,0 +1,1080 @@
...
+//   1. Propose: the draft produces K draft tokens.
+//   2. Verify: the target scores all K tokens in a single pass.
+//   3. Accept each token in order - greedy keeps it if it's the target's top choice, sampling
+//      keeps it with probability min(1, p_target/p_draft). The first rejected token is swapped
+//      for a "correction" token; if all K are accepted we append one extra bonus token.
+//   4. Buffer the accepted (+ correction/bonus) tokens for DrainOne to emit one at a time.
+// Re-anchoring the target and the subclass's Advance are deferred to FinalizeRound.
+// Note: grab the target's distribution for the first token before the verify Run - afterwards
+// GetLogits() holds the last token's row, not the first.
+void SpeculativeDecodingStrategy::RunRound(Generator& g) {
```
kunal-vaishnavi, line 291: From Copilot:

When `use_anchor` is true, `target_input` contains `K + 1` tokens: the pending anchor token followed by the `K` proposals. However, this call still passes `seed_length + K` as the total length:

```cpp
spec_state->target_state().Run(seed_length + K, target_input, {});
```

Could this use a total length consistent with the number of tokens being appended, including the anchor? Otherwise the fold path can assign incorrect target positions or leave the target KV cache length out of sync with the input. Please also add a focused anchor/fold test covering the resulting logits and cache-length invariants.

kunal-vaishnavi, line 291: From Copilot:

Could we keep the CPU copy alive in a local variable instead of taking `.data()` from the temporary returned by `CopyDeviceToCpu()`?

For example:

```cpp
auto verify_logits_cpu = verify_logits.CopyDeviceToCpu();
const float* data = verify_logits_cpu.data();
```

The current expression may leave `data` pointing into a buffer whose owning temporary has already been destroyed, making the subsequent `GetLogitsRow(data, ...)` accesses undefined behavior. The guidance path appears to use the same pattern, so please apply the same lifetime fix there as well.

src/config.h
```diff
@@ -454,6 +456,10 @@ struct Config {
     float blank_penalty{};             // Penalty applied to blank token logits in CTC/RNNT decoding. Default 0 means no penalty.
   } search;
 
+  struct Speculative {
+    int max_draft_tokens{4};  // Number of tokens the draft proposes per round.
```
kunal-vaishnavi: We should document why 4 is selected as the default.

samsat701: Documented in config.h - https://github.com/microsoft/onnxruntime-genai/pull/2233/changes/8f3e8787136f10adcfd1b1bca6b1d6bb644a3982. 

4 is a conservative default that amortizes target verification without excessive draft work (the optimal val remains model- and EP-dependent)

src/speculative_decoding_strategy.cpp
samsat701: Fixed - https://github.com/microsoft/onnxruntime-genai/pull/2233/changes/8f3e8787136f10adcfd1b1bca6b1d6bb644a3982.

Addressed in both standard and guidance paths. The returned CPU span is now stored in a local variable before taking .data().

samsat701: Kept seed_length + K. In the folded path, the target cache starts one token behind seed_length; appending the anchor + K proposals ends at seed_length + K.

src/models/speculative_decoding.cpp
samsat701: Fixed in https://github.com/microsoft/onnxruntime-genai/pull/2233/changes/8f3e8787136f10adcfd1b1bca6b1d6bb644a3982. 

Added pre-validation before constructing either target/draft state. Existing negative tests cover all three settings.

src/speculative_sampling.h
samsat701: Removed in https://github.com/microsoft/onnxruntime-genai/pull/2233/changes/8f3e8787136f10adcfd1b1bca6b1d6bb644a3982.

src/models/speculative_decoding.h
samsat701: Removed in https://github.com/microsoft/onnxruntime-genai/pull/2233/changes/8f3e8787136f10adcfd1b1bca6b1d6bb644a3982.

src/config.cpp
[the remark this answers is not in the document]
samsat701: Fixed in config.cpp - https://github.com/microsoft/onnxruntime-genai/pull/2233/changes/8f3e8787136f10adcfd1b1bca6b1d6bb644a3982. Parsing now uses SafeDoubleToInt.

kunal-vaishnavi (approved):

src/models/speculative_decoding.cpp
samsat701: Will address in future PR.

src/csharp/NativeMethods.cs
samsat701: Will address in future PR.

src/cuda/search_cuda.cpp
samsat701: Will address in future PR.

src/speculative_decoding_strategy.cpp
samsat701: Will address in future PR.

src/ort_genai_c.h
samsat701: Will address further in future PR.

PR #2229, opened 2026-06-15
Add option of fp4 QMoE for gpt-oss in model builder

tianleiwu:
Add option to build gpt-oss model with FP4 QMoE.

PR #2234, opened 2026-06-19
Update model builder for gpt-oss

tianleiwu:
# Update model builder for GPT-OSS (QMoE encoding fix, decoupled INT4/INT8 quantization, QK-norm fusion)

## Description

This PR overhauls the Python model builder to correctly export GPT-OSS (and related MoE/hybrid) models. It fixes a CUDA QMoE expert-weight encoding bug that produced garbage INT4 weights, decouples the INT4 quantization *method* from INT8 bit *placement* (so any base method can be combined with any INT8 upgrade policy), adds per-channel QMoE quantization, and introduces optional Q/K-norm fusion into `GroupQueryAttention`. It supersedes #2209.

### QMoE expert-weight quantization (CUDA)

| File | Change |
|------|--------|
| [src/python/py/models/builders/base.py](src/python/py/models/builders/base.py) | New `_cutlass_prepacked_blockwise_quantize` quantizes each expert `[N, K]` with ONNX Runtime's blockwise `quantize_matmul_4bits`/`8bits` on the transposed `[K, N]` weight, **keeps the signed scales**, and offline CUTLASS-prepacks via `pack_weights_for_cuda_mixed_gemm(..., force_arch=80)` — the exact encoding validated by the `com.microsoft` QMoE CUDA parity tests. |
| [src/python/py/models/builders/base.py](src/python/py/models/builders/base.py) | New `_matmulnbits_blockwise_quantize` emits raw `[N, K/pack]` weights + blockwise scales for the runtime PrePack hook (used when `qmoe_weights_prepacked=0`). |
| [src/python/py/models/builders/base.py](src/python/py/models/builders/base.py) | New `_symmetric_per_channel_quantize` / `_cuda_per_channel_quantize` add per-channel QMoE quantization (`qmoe_block_size <= 0`), omitting the `block_size` attribute. |
| [src/python/py/models/builders/base.py](src/python/py/models/builders/base.py) | `make_qmoe_weights` routes CUDA QMoE through the prepacked/raw/per-channel paths, validates `block_size ∈ {32, 64, 128}` and `N % pack == 0` with real exceptions, and removes the deprecated column-wise (TensorRT-LLM) path. |
| [src/python/py/models/builders/base.py](src/python/py/models/builders/base.py) | New tri-state `weights_prepacked` QMoE attribute (`-1` auto/omit, `1` prepacked, `0` raw), CUDA-only, validated to `{-1, 0, 1}`; override via `extra_options["qmoe_weights_prepacked"]`. |
| [src/python/py/models/builders/gptoss.py](src/python/py/models/builders/gptoss.py) | Always transpose HF expert weights to output-major `[E, N, K]` for **all** EPs (the old CUDA-only special case swapped `N`/`K` and silently corrupted quantized weights/scales). |

**Root cause fixed:** the previous CUDA path applied `abs()` to the blockwise scales. `quantize_matmul_4bits` folds the block-anchor sign into the scale and the kernel dequantizes as `(q − 8) · scale`, so `abs()` corrupted every block whose anchor element is negative (~half the blocks), yielding garbage weights that merely *looked* like INT4 quality loss.

### Decoupled INT4 method vs. INT8 bit placement

| File | Change |
|------|--------|
| [src/python/py/models/builder.py](src/python/py/models/builder.py) | New orthogonal flags `last_matmul_weight_int8`, `int8_mixed_layers`, `int8_linear_attn`; `int4_algo_config` reduced to base methods `default`/`rtn`/`k_quant` with help text updated. |
| [src/python/py/models/builders/base.py](src/python/py/models/builders/base.py) | New `resolve_int4_quant_config` splits `int4_algo_config` into a base method + placement flags; legacy compound names (`rtn_last`, `k_quant_last`, `k_quant_mixed`, `k_quant_linear`) are kept as aliases. |
| [src/python/py/models/builders/base.py](src/python/py/models/builders/base.py) | New `make_bit_placement_config` builds the per-node INT8 `customized_weight_config`; `make_algo_config` now takes a base method + config and raises on unknown methods. |
| [src/python/py/models/builders/base.py](src/python/py/models/builders/base.py) | `to_int4` supports INT8 placement on top of the DEFAULT method via a two-pass scheme (INT4 body first, then an **RTN** INT8 pass for designated nodes, since DEFAULT 8-bit QOperator output is not consumed correctly by the runtime kernel). |
| [src/python/py/models/quantization.md](src/python/py/models/quantization.md) | New doc explaining base methods, INT8 placement flags, legacy aliases, accuracy notes (MMLU numbers), and examples. |

### Q/K-norm fusion into GroupQueryAttention

| File | Change |
|------|--------|
| [src/python/py/models/builders/base.py](src/python/py/models/builders/base.py) | New `is_fused_qk_norm_gqa_supported` and `q_norm_weight`/`k_norm_weight` GQA inputs to pass Q/K-norm weights directly into `GroupQueryAttention` (CUDA/WebGPU). Gated by the `fuse_qk_norm_gqa` extra option (default off). |
| [src/python/py/models/builder.py](src/python/py/models/builder.py) | Register `fuse_qk_norm_gqa` extra option. |
| [src/python/py/models/README.md](src/python/py/models/README.md) | Document the QK-norm fusion toggle. |

### Robustness / misc

- [src/python/py/models/builders/base.py](src/python/py/models/builders/base.py): `enable_skip_layer_norm_strict_mode` is now version-gated — disabled on ONNX Runtime ≥ 1.27 (FP32 accumulation by default, faster) and kept enabled on older supported versions to avoid an accuracy regression (new `onnxruntime_version_at_least` helper).
- Error chaining (`raise ... from e`) added to QMoE quantization failures.

### Tests

- [test/python/builder/test_qmoe_weights.py](test/python/builder/test_qmoe_weights.py): New unit tests covering CUDA path dispatch (prepacked/raw/per-channel), EP gating, unsupported block-size rejection, per-channel shapes, and a regression guard asserting blockwise scales stay **signed**. Includes a hardened dependency stub that no longer shadows a real `onnxruntime` wheel.

## Testing

- Run the new builder unit tests:
  ```bash
  python -m pytest -sv test/python/builder/test_qmoe_weights.py
  ```
  CUDA-only shape/sign regression tests auto-skip when the ONNX Runtime CUDA pybind (`quantize_matmul_4bits`, `pack_weights_for_cuda_mixed_gemm`) is unavailable.
- End-to-end: build a GPT-OSS QMoE model on CUDA and verify generation quality (the encoding fix restores expected INT4 accuracy vs. the previous garbage output).
- Backward compatibility: legacy `int4_algo_config` compound values (`rtn_last`, `k_quant_last`, `k_quant_mixed`, `k_quant_linear`) still produce identical models via the alias mapping.

## Motivation and Context

Replaces #2209. The CUDA QMoE export path was producing weights the `com.microsoft` QMoE kernel could not consume correctly (sign-stripped scales + wrong `N`/`K` orientation), which manifested as severe quality loss on GPT-OSS / Qwen3.5 MoE models. Alongside the fix, the quantization configuration surface is reworked so INT4 rounding method and INT8 placement are independent, which is needed for the mixed-precision recipes these models rely on.

## Checklist

- [x] Tests added/updated
- [x] Documentation updated (`quantization.md`, `README.md`)
- [x] No breaking changes (legacy `int4_algo_config` values preserved as aliases)
- [ ] CI passes


src/python/py/models/builders/base.py
```diff
@@ -123,7 +123,7 @@ def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
             "cpu": {},
             "cuda": {
                 "enable_cuda_graph": "1" if extra_options.get("enable_cuda_graph", False) else "0",
-                "enable_skip_layer_norm_strict_mode": "1",
+                "enable_skip_layer_norm_strict_mode": "0",
```
kunal-vaishnavi: Why disable strict mode by default?

src/python/py/models/builders/base.py
```diff
@@ -304,6 +304,10 @@ def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
         top_k_experts = config.num_experts_per_tok if hasattr(config, "num_experts_per_tok") else 0
         expert_weight_bits = 8 if extra_options.get("use_8bits_moe", False) else 4
         swiglu_limit = config.swiglu_limit if hasattr(config, "swiglu_limit") else None
+        # For CUDA QMoE the builder ships expert weights already CUTLASS-prepacked (offline via pack_weights_for_cuda_mixed_gemm, see make_qmoe_weights),
+        # so the QMoE op's default interpretation (weights_prepacked=-1/auto = prepacked) is exactly what we want and the attribute is omitted.
+        # Override via extra_options["qmoe_weights_prepacked"] (e.g. 0 to ship raw [E, N, K/pack] weights and let the runtime PrePack hook transform them).
+        weights_prepacked = int(extra_options["qmoe_weights_prepacked"]) if "qmoe_weights_prepacked" in extra_options else -1
```
kunal-vaishnavi: Looks like some changes from #2209 haven't been brought over such as this one

src/python/py/models/builders/base.py
```diff
@@ -123,7 +123,8 @@ def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
             "cpu": {},
             "cuda": {
                 "enable_cuda_graph": "1" if extra_options.get("enable_cuda_graph", False) else "0",
-                "enable_skip_layer_norm_strict_mode": "1",
+                # Since ORT 1.27.0, we use FP32 accumulation so no need to use the strict mode workaround anymore.
+                "enable_skip_layer_norm_strict_mode": "0",
             },
```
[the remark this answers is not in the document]
tianleiwu: Good point on the ORT version floor. Fixed in d462d62 by version-gating the default: `enable_skip_layer_norm_strict_mode` is now set to `0` only when the installed onnxruntime is >= 1.27.0 (where SkipLayerNorm uses FP32 accumulation), and falls back to `1` for older versions still supported by the package (onnxruntime-gpu>=1.20.1), avoiding the accuracy regression.

src/python/py/models/builders/base.py
```diff
@@ -690,59 +716,153 @@ def save_processing(self, model_name_or_path, extra_kwargs, out_dir):
...
+        if algo in self.LEGACY_INT4_ALGO_ALIASES:
+            base_method, implied = self.LEGACY_INT4_ALGO_ALIASES[algo]
+        elif algo == "k_quant":
+            # Historically `k_quant` always upgraded the LM head to int8.
+            implied = {"last_matmul_weight_int8": True}
+
+        placement = {}
+        for flag in self.INT8_PLACEMENT_FLAGS:
+            placement[flag] = bool(extra_options.get(flag, implied.get(flag, False)))
+        return base_method, placement
```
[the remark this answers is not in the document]
tianleiwu: Agreed — bare `k_quant` should not imply an INT8 LM head. Fixed in d462d62 by removing the implicit `last_matmul_weight_int8=True` for `k_quant`; only the legacy `k_quant_last`/`k_quant_mixed`/`k_quant_linear` aliases (and the explicit `last_matmul_weight_int8` flag) upgrade the LM head now, matching the CLI help and README. This restores the prior `main` behavior where bare `k_quant` keeps the LM head at INT4.

src/python/py/models/builders/base.py
```diff
@@ -3240,52 +3415,217 @@ def make_qmoe_op(self, name, **kwargs):
...
+        # is negative and yields garbage weights.
+
+        # CUTLASS-prepack with force_arch=80: the fpA_intB kernel always expects
+        # the SM80-style interleaved layout for SM >= 80 (validated on SM89/SM90),
+        # so 80 is the correct universal choice regardless of build/runtime GPU.
+        q_reshaped = qweight.reshape(n, -1)  # [N, K/pack]
+        packed = _ortpyb.pack_weights_for_cuda_mixed_gemm(q_reshaped, n, k, bits, 80)
+        out_cols = n // pack
+        packed = np.asarray(packed).view(np.uint8).reshape(k, out_cols)  # [K, N/pack]
+
```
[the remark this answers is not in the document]
tianleiwu: Fixed in d462d62: added an explicit `n % pack != 0` check (pack=2 for INT4) that raises a `ValueError`, mirroring the existing `k % block_size` validation, so an odd `N` fails deterministically instead of silently mis-shaping.

test/python/builder/test_qmoe_weights.py
```diff
@@ -0,0 +1,277 @@
...
+        tensor_adapters = types.ModuleType("onnx_ir.tensor_adapters")
+        tensor_adapters.TorchTensor = object
+        tensor_adapters.to_torch_dtype = lambda dtype: dtype
+        sys.modules["onnx_ir"] = onnx_ir
+        sys.modules["onnx_ir.tensor_adapters"] = tensor_adapters
+
+    if not _module_available("onnxruntime.quantization.matmul_nbits_quantizer"):
+        onnxruntime = sys.modules.setdefault("onnxruntime", types.ModuleType("onnxruntime"))
+        quantization = types.ModuleType("onnxruntime.quantization")
+        matmul_nbits_quantizer = types.ModuleType("onnxruntime.quantization.matmul_nbits_quantizer")
```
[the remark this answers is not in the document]
tianleiwu: Fixed in d462d62: the stub now prefers the real `onnxruntime` when it is importable and only fabricates a top-level module when the package truly is not installed. It reuses the real `onnxruntime.quantization` submodule if present and only injects the missing `matmul_nbits_quantizer` stub, so a real wheel is no longer shadowed for the rest of the session.

jiafatom (dismissed):

src/python/py/models/builders/base.py
```diff
@@ -33,6 +33,21 @@
 )
 
 
+def onnxruntime_version_at_least(major, minor):
```
kunal-vaishnavi: Can we remove this? Model builder can be assumed to work with ORT nightly and we don't have to worry about supporting older versions. Otherwise it will become complicated very quickly.

src/python/py/models/builders/base.py
```diff
@@ -123,7 +138,11 @@ def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
             "cpu": {},
             "cuda": {
                 "enable_cuda_graph": "1" if extra_options.get("enable_cuda_graph", False) else "0",
-                "enable_skip_layer_norm_strict_mode": "1",
+                # Since ORT 1.27.0, SkipLayerNorm uses FP32 accumulation by default, so the strict-mode
+                # workaround is unnecessary (and disabling it is faster). For older ORT versions that the
+                # package still supports (onnxruntime-gpu>=1.20.1), keep strict mode enabled to avoid an
+                # accuracy regression.
+                "enable_skip_layer_norm_strict_mode": "0" if onnxruntime_version_at_least(1, 27) else "1",
```
kunal-vaishnavi: We can remove this altogether.

src/python/py/models/builders/base.py
```diff
@@ -328,12 +357,19 @@ def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
         self.make_lm_head_init(config)
 
         # Quantization-specific variables (INT4, INT8, etc.)
-        algo_config = self.make_algo_config(extra_options.get("int4_algo_config", "default"))
+        # Decouple the int4 quantization *method* (how weights are rounded) from the
+        # int8 bit *placement* (which MatMuls are upgraded to 8 bits). The base method
+        # is one of {"default", "rtn", "k_quant"}; bit placement is controlled by the
+        # orthogonal flags resolved in `resolve_int4_quant_config`. Legacy compound
+        # names (e.g. "rtn_last", "k_quant_mixed") are still accepted as aliases.
+        self.int4_base_method, self.int4_bit_placement = self.resolve_int4_quant_config(extra_options)
```
kunal-vaishnavi: I have a [separate PR](https://github.com/microsoft/onnxruntime-genai/pull/2237) that conflicts with these changes. Can we merge the other PR first and then update this PR accordingly?

src/python/py/models/builders/base.py
```diff
@@ -474,6 +510,14 @@ def is_packed_attn_supported(self) -> bool:
     def is_fused_rope_supported(self):
         return self.ep not in ["dml"]
 
+    def is_fused_qk_norm_gqa_supported(self):
+        return (
+            self.attention_attrs["op_type"] == "GroupQueryAttention"
+            and self.ep in {"cuda", "webgpu"}
+            and self.extra_options.get("fuse_qk_norm_gqa", False)
```
kunal-vaishnavi: Rather than introduce a new extra option for `fuse_qk_norm_gqa`, can we automatically do this for EPs that support it and disable the fusion for EPs that don't support it?

src/python/py/models/builders/base.py
```diff
@@ -2446,22 +2581,42 @@ def make_group_query_attention(self, name, **kwargs):
...
+            "num_heads": self.num_attn_heads,
+            "kv_num_heads": self.num_kv_heads,
+            "scale": self.attention_attrs["scale"],
+            "local_window_size": self.window_size,
+            "softcap": self.attention_attrs["softcap"],
+            "do_rotary": self.attention_attrs["use_rope_in_attn"],
+            "rotary_interleaved": self.rope_attrs["interleaved"],
+        }
+        if q_norm_weight and k_norm_weight:
+            attributes["qk_norm_epsilon"] = kwargs.get(
```
kunal-vaishnavi: Can we always add the attribute and set it to a default value rather than having attributes conditionally added?

src/python/py/models/builders/base.py
```diff
@@ -3240,52 +3430,219 @@ def make_qmoe_op(self, name, **kwargs):
...
+    def _cuda_per_channel_quantize(self, weights, prepack):
+        """Quantize per-channel QMoE weights and optionally CUTLASS-prepack them.
+
+        ``weights_prepacked=0`` uses the raw schema layout from
+        ``_symmetric_per_channel_quantize`` so the CUDA QMoE PrePack hook can do
+        the layout transform. The default CUDA path ships the same weights already
+        packed for the SM80 fpA_intB MoE GEMM layout, matching the QMoE op's
+        default ``weights_prepacked=-1`` contract.
+        """
+        import numpy as np
```
kunal-vaishnavi: Can we move any local imports such as these to be global imports?

src/python/py/models/builders/base.py
tianleiwu: It's for experiment purpose (have a way to compare the benefit of the fusion). I can set the default to be True.

src/python/py/models/builders/base.py
tianleiwu: Why need we do that? If there is no qk_norm, the epsilon is not useful for op.

src/python/py/models/builders/base.py
```diff
@@ -287,6 +289,13 @@ def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
         top_k_experts = config.num_experts_per_tok if hasattr(config, "num_experts_per_tok") else 0
         expert_weight_bits = 8 if extra_options.get("use_8bits_moe", False) else 4
         swiglu_limit = config.swiglu_limit if hasattr(config, "swiglu_limit") else None
+        # weights_prepacked is a CUDA-only QMoE layout contract. Non-CUDA EPs omit the attribute and use
+        # their normal blockwise QMoE encoding, so CUDA-prepacked exports are not intended to be shared
+        # with CPU/WebGPU/TRT-RTX. Override via extra_options["qmoe_weights_prepacked"] (e.g. 0 to ship
+        # raw [E, N, K/pack] weights and let the CUDA runtime PrePack hook transform them).
+        weights_prepacked = int(extra_options.get("qmoe_weights_prepacked", -1))
```
kunal-vaishnavi: Why does the user need to know how the weights are packed? Is this not something we can auto-detect inside the model builder by examining the weights and setting the attribute appropriately on behalf of the user?

src/python/py/models/builders/base.py
```diff
@@ -311,17 +321,25 @@ def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
...
+        self.int4_base_method, self.int4_bit_placement = self.resolve_int4_quant_config(extra_options)
+        self.int4_customized_weight_config = self.make_bit_placement_config(self.int4_bit_placement)
+        algo_config = self.make_algo_config(self.int4_base_method, self.int4_customized_weight_config)
         self.matmul_block_size = int(extra_options.get("int4_block_size", 32))
-        self.qmoe_block_size = int(extra_options.get("qmoe_block_size", 128 if self.ep in {"cuda", "trt-rtx"} else 32))
+        self.qmoe_block_size = int(extra_options.get("qmoe_block_size", 128 if self.ep in {"trt-rtx"} else 32))
         self.quant_attrs = {
             "accuracy_level": int(extra_options.get("int4_accuracy_level", 4 if self.ep in ["cpu", "webgpu"] else 0)),
-            "qmoe_block_size": int(self.qmoe_block_size),
+            "qmoe_block_size": self.qmoe_block_size,
```
kunal-vaishnavi, line 361: The `int` casting is done in case the block size is interpreted as a string.

src/python/py/models/builders/base.py
```diff
@@ -734,59 +764,150 @@ def save_processing(self, model_name_or_path, extra_kwargs, out_dir):
...
+        "k_quant_last": ("k_quant", {"last_matmul_weight_int8": True}),
+        "k_quant_mixed": ("k_quant", {"last_matmul_weight_int8": True, "int8_mixed_layers": True}),
+        "k_quant_linear": ("k_quant", {"last_matmul_weight_int8": True, "int8_linear_attn": True}),
+    }
+
+    # Orthogonal int8 bit-placement flags (each can be combined with any base method).
+    INT8_PLACEMENT_FLAGS = ("last_matmul_weight_int8", "int8_mixed_layers", "int8_linear_attn")
+
+    def resolve_int4_quant_config(self, extra_options):
+        """Split `int4_algo_config` into a base method and int8 bit-placement flags.
```
kunal-vaishnavi: The idea of having algo aliases was to summarize how the algorithm works in one flag. What is the benefit to splitting?

src/python/py/models/builders/base.py
```diff
@@ -734,59 +764,150 @@ def save_processing(self, model_name_or_path, extra_kwargs, out_dir):
...
+        quant_format = QuantFormat.QDQ if self.quant_attrs["use_qdq"] else QuantFormat.QOperator
+        nodes_to_exclude = list(self.quant_attrs["nodes_to_exclude"])
+        customized_weight_config = getattr(self, "int4_customized_weight_config", {}) or {}
+        base_method = getattr(self, "int4_base_method", "default")
+
+        # The DEFAULT quantizer applies a single global bit-width per pass and ignores
+        # `customized_weight_config`. To support int8 bit placement (e.g. the int8 last
+        # MatMul) on top of the DEFAULT method, quantize the int4 body first (excluding
+        # the int8-designated nodes), keeping the int4 body byte-identical to a plain
+        # DEFAULT model. Then upgrade just the designated nodes to int8 with a second
```
kunal-vaishnavi: Is this something we can fix inside ORT for the future? It seems odd that we have to make two full passes through the model when an algorithm config allows us to specify which nodes are at which bits.

src/python/py/models/builders/base.py
tianleiwu: It could have more combinations. 3 algos with 2 flags have 3 x 2 x 2 = 12 combinations. 
Previously, you have 3 algos + 3 alias = 6 combinations.

src/python/py/models/builders/base.py
tianleiwu: It is an experimental flag to control whether we prepack the weights in model builder. Just in case you use a CPU-only machine to run model builder, and you can still build a model that can run in GPU machine. For production model, we shall always prepack.

src/python/py/models/builders/base.py
kunal-vaishnavi: That's the case for other attributes but they are anyways added so that the list of attributes stays the same and we avoid adding extra conditional logic. Its values determine whether an attribute is activated or not.

src/python/py/models/builders/base.py
tianleiwu: It's possible. Need some code change.

src/python/py/models/builders/gptoss.py
```diff
@@ -607,64 +729,75 @@
...
+                        down_proj_qweight_list.append(qweight2)
+                        down_proj_scales_list.append(scales2)
+
+                    gate_up_proj_qweight_tensor = torch.stack(gate_up_proj_qweight_list, dim=0).to(torch.uint8)
+                    gate_up_proj_scales_tensor = torch.stack(gate_up_proj_scales_list, dim=0)
+                    down_proj_qweight_tensor = torch.stack(down_proj_qweight_list, dim=0).to(torch.uint8)
+                    down_proj_scales_tensor = torch.stack(down_proj_scales_list, dim=0)
+
+                gate_up_proj_qweight_shape, down_proj_qweight_shape = self.make_qmoe_weight_initializer_shapes(
+                    gate_up_proj_qweight_list if not has_quark_experts else None,
```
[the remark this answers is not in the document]
tianleiwu: Fixed in 39b9c90. `gate_up_proj_qweight_list`/`down_proj_qweight_list` are now initialized to `None` before the Quark/non-Quark branch, so the shape helper never references an undefined name (the `... if not has_quark_experts else None` guard already prevented a real runtime use).

src/python/py/models/builders/gptoss.py
```diff
@@ -607,64 +729,75 @@
...
+                    down_proj_qweight_list if not has_quark_experts else None,
```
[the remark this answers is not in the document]
tianleiwu: Fixed in 39b9c90 (same as the `gate_up_proj_qweight_list` case): both per-expert weight lists are now initialized to `None` up front.

src/python/py/models/builders/base.py
```diff
@@ -734,72 +798,232 @@ def save_processing(self, model_name_or_path, extra_kwargs, out_dir):
...
+
+    def make_algo_config(self, quant_method: str, customized_weight_config=None):
+        """Create the MatMulNBitsQuantizer algo config for a *base* method.
+
+        `quant_method` is one of {"default", "rtn", "k_quant"}. Per-node int8 bit
+        placement is supplied via `customized_weight_config`. The "default" method
+        returns ``None`` (MatMulNBitsQuantizer's built-in DEFAULT quantizer), which
+        cannot apply per-node bits in a single pass; `to_nbits` performs a second
+        RTN pass to honor any int8 placement for that method.
+        """
```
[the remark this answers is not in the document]
tianleiwu: Fixed in 39b9c90. The docstring now says the DEFAULT method runs a second DEFAULT int8 pass (`algo_config=None`, `bits=8`) rather than an RTN pass, matching the implementation in `to_nbits` and its inline comment.

src/python/py/models/builders/cuda_quantizer.py
```diff
@@ -0,0 +1,477 @@
...
+    def _matmulnbits_blockwise_quantize_impl(
+        weights: torch.Tensor,
+        bits: int,
+        block_size: int,
+        *,
+        symmetric: bool,
+        abs_scales: bool,
+        unsigned_full_range: bool,
+    ) -> tuple[torch.Tensor, torch.Tensor, torch.Tensor]:
+        """Quantize ``weights`` with MatMulNBits pybinds and return unflattened storage."""
```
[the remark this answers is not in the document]
tianleiwu: Fixed in 39b9c90. The docstring now states that the symmetric path uses the local NumPy MLAS-style implementation (to keep signed per-block scales) and only the asymmetric path delegates to the MatMulNBits ORT pybinds.

src/python/py/models/builder.py
```diff
@@ -475,8 +497,15 @@ def get_args():
                     Use this option when you want to use quantize-dequantize ops. For example, you will have a quantized MatMul op instead of the MatMulNBits op.
                 use_8bits_moe = Use 8-bit quantization for MoE layers. Default is false.
                     If true, the QMoE op will use 8-bit quantization. If false, the QMoE op will use 4-bit quantization.
+                use_fp4_moe = Use FP4 (MXFP4) quantization for MoE layers on the CUDA EP. Default is false.
+                    If true, the QMoE op uses MXFP4 weights (quant_type="fp4", expert_weight_bits=4, block_size=32):
+                    4-bit e2m1 weights with ue8m0 (float8e8m0) block scales and a per-expert float32 global scale.
+                    Requires an ONNX Runtime build with onnxruntime_USE_FP4_QMOE=ON.
+                    Mutually exclusive with use_8bits_moe. Only supported on the CUDA EP.
```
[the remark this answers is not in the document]
tianleiwu: Fixed in 39b9c90. Added a "Use FP4 Quantization in QMoE" section to `README.md` (plus a TOC entry) documenting `use_fp4_moe`, its MXFP4 encoding, CUDA-only/INT4 constraints, the `onnxruntime_USE_FP4_QMOE=ON` build requirement, mutual exclusivity with `use_8bits_moe`, and a CLI example.

src/python/py/models/builders/base.py
tianleiwu: Done in 7fc7837 — `qk_norm_epsilon` is now added unconditionally to the GroupQueryAttention attributes (alongside the other always-present attributes), so the attribute list stays consistent regardless of whether Q/K norm is active. The value is only consumed by the op when the q_norm/k_norm weight inputs are present.

src/python/py/models/builders/base.py
```diff
@@ -32,6 +32,8 @@
     GenerationConfig,
 )
 
+from .cuda_quantizer import CudaQuantizer
+
```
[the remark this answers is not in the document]
tianleiwu: Fixed in 7fc7837 — removed the whitespace-only line after the `CudaQuantizer` import.

src/python/py/models/builders/gptoss.py
```diff
@@ -20,6 +21,120 @@ def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
...
+            with open(index_path) as f:
+                self._mxfp4_weight_map = json.load(f)["weight_map"]
+        else:
+            safetensors_files = sorted(snapshot_dir.glob("*.safetensors"))
+            if len(safetensors_files) != 1:
+                raise RuntimeError(
+                    f"Could not locate original GPT-OSS MXFP4 safetensors index in {snapshot_dir}."
+                )
+            self._mxfp4_weight_map = None
+        return self._mxfp4_weight_map
```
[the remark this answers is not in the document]
tianleiwu: Fixed in 7fc7837 — added an explicit `_mxfp4_weight_map_loaded` flag so the no-index case (`weight_map = None`) is cached as well, instead of re-globbing the snapshot directory on every tensor load.

src/python/py/models/builders/cuda_quantizer.py
```diff
@@ -0,0 +1,685 @@
...
+        )
+    except ImportError as e:
+        raise ImportError(
+            "CUDA blockwise quantization requires quantize_matmul_4bits and quantize_matmul_8bits from onnxruntime."
+        ) from e
+
+    return quantize_matmul_4bits, quantize_matmul_8bits
+
+
+class CudaQuantizer:
```
kunal-vaishnavi, line 213: If we use ORT's new `CudaQuantizer`, can this be removed? If yes, let's use the latest ORT nightly build in the CIs and pair it with this PR to make CUDA quantization work.

src/python/py/models/builders/base.py
```diff
@@ -102,11 +104,17 @@ def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
 
         # EP-specific variables
         self.ep = ep
+        use_fp4_moe = extra_options.get("use_fp4_moe", False)
```
kunal-vaishnavi: Are the [FP4 QMoE PR changes](https://github.com/microsoft/onnxruntime-genai/pull/2229) being merged into this PR?

src/python/py/models/builders/cuda_quantizer.py
tianleiwu: Yes, it can be removed. 

This one is latest. Need update the ORT one, then remove the copy from genai later.

src/python/py/models/builders/base.py
tianleiwu: I think so. That PR is not needed.

src/python/py/models/builders/base.py
```diff
@@ -102,11 +104,17 @@ def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
...
+        if use_fp4_moe and extra_options.get("use_8bits_moe", False):
```
kunal-vaishnavi: Can we move these validations for the extra options to `check_extra_options` instead?

https://github.com/microsoft/onnxruntime-genai/blob/1735d564894c4e5b3977ae372c08b5340e0d6079/src/python/py/models/builder.py#L60-L63

src/python/py/models/builders/base.py
```diff
@@ -264,6 +272,9 @@ def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
             "rope": True,                                    # Use rotary embeddings in attention subgraph
             "q_norm": False,                                 # LayerNorm after MatMul in Q path
             "k_norm": False,                                 # LayerNorm after MatMul in K path
+            "q_norm_weight": "",                             # Q norm weight input to GroupQueryAttention
```
kunal-vaishnavi: `q_norm`/`k_norm` and `q_norm_weight`/`k_norm_weight` seem like similar settings. Can we simplify to use just the boolean settings? We can assume what the naming convention will be for the QK norm weights. For example, the `sinks` tensor is handled as follows.

https://github.com/microsoft/onnxruntime-genai/blob/506e04e0978b1830527efc912757185e3f48f552/src/python/py/models/builders/base.py#L3102-L3106

It only requires a boolean setting here.

src/python/py/models/builders/base.py
```diff
@@ -285,8 +296,18 @@ def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
...
+        # FP4 (MXFP4) is a 4-bit format; INT8 path is disabled when FP4 is requested.
+        expert_weight_bits = 4 if use_fp4_moe else (8 if extra_options.get("use_8bits_moe", False) else 4)
+        # QMoE quantization type: "int" (default INT4/INT8) or "fp4" (MXFP4).
+        moe_quant_type = "fp4" if use_fp4_moe else "int"
         swiglu_limit = config.swiglu_limit if hasattr(config, "swiglu_limit") else None
+        # weights_prepacked is a CUDA-only QMoE layout contract. Non-CUDA EPs omit the attribute and use
+        # their normal blockwise QMoE encoding, so CUDA-prepacked exports are not intended to be shared
+        # with CPU/WebGPU/TRT-RTX. Override via extra_options["qmoe_weights_prepacked"] (e.g. 0 to ship
+        # raw [E, N, K/pack] weights and let the CUDA runtime PrePack hook transform them).
+        weights_prepacked = int(extra_options.get("qmoe_weights_prepacked", -1))
```
kunal-vaishnavi: Same comment as [this](https://github.com/microsoft/onnxruntime-genai/pull/2234/changes#r3575645400)

src/python/py/models/builders/base.py
```diff
@@ -311,17 +334,39 @@ def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
...
         self.matmul_block_size = int(extra_options.get("int4_block_size", 32))
-        self.qmoe_block_size = int(extra_options.get("qmoe_block_size", 128 if self.ep in {"cuda", "trt-rtx"} else 32))
+        # matmulnbits_weights_prepacked is a CUDA-only MatMulNBits (int4/int8) layout selector. It offline
+        # prepacks the weights into the fpA_intB mixed-GEMM layout so the kernel can consume them directly:
+        # 0 = off (raw blockwise layout), 1 = SM80/Ampere fpA_intB layout (weight_prepacked=1),
+        # 2 = SM90/Hopper fpA_intB layout (weight_prepacked=2). Only meaningful on the CUDA EP; other EPs
+        # keep the raw blockwise layout. Override via extra_options["matmulnbits_weights_prepacked"].
+        self.matmulnbits_weights_prepacked = int(extra_options.get("matmulnbits_weights_prepacked", 0))
+        if self.matmulnbits_weights_prepacked not in (0, 1, 2):
+            raise ValueError(
```
kunal-vaishnavi: Same comment as [this](https://github.com/microsoft/onnxruntime-genai/pull/2234/changes#r3575645400)

src/python/py/models/builders/base.py
```diff
@@ -311,17 +334,39 @@ def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
         self.make_lm_head_init(config)
 
         # Quantization-specific variables (INT4, INT8, etc.)
+        # Decouple the int4 quantization *method* (how weights are rounded) from the
+        # int8 bit *placement* (which MatMuls are upgraded to 8 bits). The base method
+        # is one of {"default", "rtn", "k_quant"}; bit placement is controlled by the
+        # orthogonal flags resolved in `resolve_int4_quant_config`. Legacy compound
+        # names (e.g. "rtn_last", "k_quant_mixed") are still accepted as aliases.
         self.algo_config_name = extra_options.get("int4_algo_config", "default")
+        self.int4_base_method, self.int4_bit_placement = self.resolve_int4_quant_config(extra_options)
```
kunal-vaishnavi: There is [another PR](https://github.com/microsoft/onnxruntime-genai/pull/2275) that will introduce int8 support and generalize the extra options for both int4 and int8 by removing the `int4` prefix. Can we generalize here as well?

src/python/py/models/builders/base.py
```diff
@@ -734,72 +791,233 @@ def save_processing(self, model_name_or_path, extra_kwargs, out_dir):
         print(f"Saving processing files in {out_dir} for GenAI")
         tokenizer.save_pretrained(out_dir)
 
-    def make_algo_config(self):
-        # Create a quantization configuration based on the algorithm name.
+    # Legacy compound `int4_algo_config` names map to a base method plus a set of
+    # int8 bit-placement flags so existing recipes keep producing identical models.
+    LEGACY_INT4_ALGO_ALIASES = {
```
kunal-vaishnavi: Can we move these constants into `make_quant_init` and invoke all resolutions inside there?

src/python/py/models/builders/base.py
```diff
@@ -457,6 +502,14 @@ def is_packed_attn_supported(self) -> bool:
     def is_fused_rope_supported(self):
         return self.ep not in ["dml"]
 
+    def is_fused_qk_norm_gqa_supported(self):
+        return (
+            self.attention_attrs["op_type"] == "GroupQueryAttention"
+            and self.ep in {"cuda", "webgpu"}
```
kunal-vaishnavi, line 506: Is the CPU EP not supported yet?

src/python/py/models/builders/base.py
kunal-vaishnavi: The `resolve_int4_quant_config` method can define `self.int4_base_method` and `self.int4_bit_placement` as class attributes so we don't need to return them from the method.

src/python/py/models/builders/base.py
```diff
@@ -311,17 +334,39 @@ def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
...
 
         # Quantization-specific variables (INT4, INT8, etc.)
+        # Decouple the int4 quantization *method* (how weights are rounded) from the
+        # int8 bit *placement* (which MatMuls are upgraded to 8 bits). The base method
+        # is one of {"default", "rtn", "k_quant"}; bit placement is controlled by the
+        # orthogonal flags resolved in `resolve_int4_quant_config`. Legacy compound
+        # names (e.g. "rtn_last", "k_quant_mixed") are still accepted as aliases.
         self.algo_config_name = extra_options.get("int4_algo_config", "default")
+        self.int4_base_method, self.int4_bit_placement = self.resolve_int4_quant_config(extra_options)
+        self.int4_customized_weight_config = self.make_bit_placement_config(self.int4_bit_placement)
```
kunal-vaishnavi: Same thing here with `make_bit_placement_config` returning `int4_customized_weight_config`.

src/python/py/models/builders/base.py
```diff
@@ -547,18 +600,22 @@ def make_tied_quantized_embedding_input_names(self):
         # where rtn* = rtn, rtn_last
         #       k_quant* = k_quant, k_quant_last, k_quant_linear, k_quant_mixed
 
-        bits = 8 if self.algo_config_name in {"k_quant_mixed", "k_quant_last", "k_quant_linear", "rtn_last"} else 4
+        base_method, placement = self.resolve_int4_quant_config(getattr(self, "extra_options", {}))
```
kunal-vaishnavi: Why do we have to call this twice?

src/python/py/models/builders/base.py
```diff
@@ -734,72 +791,233 @@ def save_processing(self, model_name_or_path, extra_kwargs, out_dir):
...
+    # int8 bit-placement flags so existing recipes keep producing identical models.
+    LEGACY_INT4_ALGO_ALIASES = {
+        "rtn_last": ("rtn", {"last_matmul_weight_int8": True}),
+        "k_quant_last": ("k_quant", {"last_matmul_weight_int8": True}),
+        "k_quant_mixed": ("k_quant", {"last_matmul_weight_int8": True, "int8_mixed_layers": True}),
+        "k_quant_linear": ("k_quant", {"last_matmul_weight_int8": True, "int8_linear_attn": True}),
+    }
+
+    # Orthogonal int8 bit-placement flags (each can be combined with any base method).
+    INT8_PLACEMENT_FLAGS = ("last_matmul_weight_int8", "int8_mixed_layers", "int8_linear_attn")
```
kunal-vaishnavi: After the [extra options renaming](https://github.com/microsoft/onnxruntime-genai/pull/2234/changes#r3581096559), I wonder if we can specify the number of bits to use on a flag rather than specifying int4-specific or int8-specific flags.

Separately, I'm still wondering whether specifying bit placement via an extra option per node location is ideal. I feel that having one extra option to specify the exact bit placement for all nodes is easier for the user to understand instead of needing to specify many extra options that describe bit placement per node.

src/python/py/models/builders/base.py
```diff
@@ -734,72 +791,233 @@ def save_processing(self, model_name_or_path, extra_kwargs, out_dir):
...
+        # verified end-to-end, so RTN is no longer needed for the int8 nodes.)
+        if base_method == "default" and customized_weight_config:
+            int8_nodes = list(customized_weight_config.keys())
+
+            quant_int4 = MatMulNBitsQuantizer(
+                model=ir.to_proto(self.model),
+                block_size=self.quant_attrs["qdq_block_size"],
+                is_symmetric=self.quant_attrs["is_symmetric"],
+                accuracy_level=self.quant_attrs["accuracy_level"],
+                nodes_to_exclude=nodes_to_exclude + int8_nodes,
```
kunal-vaishnavi: Did we ever fix this inside the latest ORT PRs to only require constructing one quantizer and processing once? Or do we still need 2 separate quantizations?

src/python/py/models/builders/base.py
```diff
@@ -734,72 +791,233 @@ def save_processing(self, model_name_or_path, extra_kwargs, out_dir):
...
+            model_proto = quant.model.model
+
+        # Offline CUDA weight prepacking is a pure weight *layout* conversion for the
+        # fpA_intB mixed-GEMM kernel and is independent of the quantization method or bit
+        # width above, so it is applied here as an orthogonal post-pass over the quantized
+        # MatMulNBits nodes.
+        self.prepack_matmulnbits_weights(model_proto)
+        return ir.from_proto(model_proto)
+
+    def prepack_matmulnbits_weights(self, model_proto):
```
kunal-vaishnavi, line 991: Same comment as [this](https://github.com/microsoft/onnxruntime-genai/pull/2234#discussion_r3575073782)

src/python/py/models/builders/base.py
```diff
@@ -3234,6 +3507,74 @@ def make_base_moe_op(self, name, **kwargs):
         )
         self.make_value(output, self.io_dtype, shape=["batch_size", "sequence_length", self.hidden_size])
 
+    def make_fp8e8m0_initializer(self, scales_uint8, name):
```
kunal-vaishnavi, line 3561: Should these new FP* methods be inside `cuda_quantizer.py` or are they generic enough for any EP?

src/python/py/models/builders/base.py
```diff
@@ -3234,6 +3507,74 @@ def make_base_moe_op(self, name, **kwargs):
...
+        ir_tensor = ir.Tensor(np.ascontiguousarray(arr), dtype=ir.DataType.FLOAT8E8M0, name=name)
+        value = self.make_value(name, ir_tensor.dtype, ir_tensor.shape)
+        value.const_value = ir_tensor
+        self.model.graph.register_initializer(value)
+
+    # Positive FP4 e2m1 representable magnitudes (codes 0-7); negatives use codes 8-15.
+    _FP4_E2M1_POS_VALUES = (0.0, 0.5, 1.0, 1.5, 2.0, 3.0, 4.0, 6.0)
+    _FP4_E2M1_MAX = 6.0
+
+    def make_mxfp4_weights(self, weight, block_size=32):
```
kunal-vaishnavi, line 3578: I'm guessing this can be moved inside `cuda_quantizer.py` or put inside ORT's quantizer tools?

src/python/py/models/builders/base.py
tianleiwu: Right. CPU support is not done yet.

src/python/py/models/builders/base.py
tianleiwu: See the comments above. We have to do two passes.

src/python/py/models/builders/base.py
tianleiwu: Done in 73fa03f485. Removed `q_norm_weight`/`k_norm_weight` from `attention_attrs`; only the `q_norm`/`k_norm` booleans remain. The initializer names are now derived by convention in `get_qk_norm_weight_names(layer_id)` (same pattern as `sinks`), and they are only fed into GroupQueryAttention when the fused path is active (`q_norm and k_norm and is_fused_qk_norm_gqa_supported()`). The non-fused path (`make_qk_norm`) still emits separate SimplifiedLayerNorm nodes and passes no norm weights into the op.

src/python/py/models/builders/base.py
tianleiwu: Done in 73fa03f485. The `qmoe_weights_prepacked`/`matmulnbits_weights_prepacked` range checks now live in `check_extra_options` (which now also takes `precision`), and the duplicate checks were removed from `Model.__init__`. The FP4 precision/EP validation moved along with the `moe_quant_type` rework (f873b2823c): `check_extra_options` validates `moe_quant_type` (int4/int8/mxfp4) and enforces that `mxfp4` is CUDA-only and requires an int4 symmetric build.

src/python/py/models/builders/base.py
tianleiwu: Partly generalized. `use_fp4_moe`/`use_8bits_moe` are replaced by a single `moe_quant_type` (`int4`/`int8`/`mxfp4`), keeping backward-compat for the now-deprecated `use_8bits_moe` (f873b2823c). Mixed precision is now expressed as `matmul_mixed_precision=selector:quant_type` using quant-type names (`int4`/`int8`) instead of int-specific boolean flags (ef36112bcd, f9fc9aa39e), so int8 placement no longer needs a dedicated flag. I kept the existing `int4_*` prefixed options (`int4_algo_config`, `int4_block_size`, `int4_is_symmetric`, ...) as-is for now to avoid churn/conflicts with #2275 — happy to do the prefix rename in coordination with that PR so we land one consistent naming.

src/python/py/models/builders/base.py
tianleiwu: Done in 73fa03f485. The alias/selector constants (`LEGACY_INT4_ALGO_ALIASES`, `MATMUL_MIXED_PRECISION_SELECTORS`, `MATMUL_MIXED_PRECISION_QUANT_TYPES`) now sit next to `make_quant_init`, and `make_quant_init` invokes all the resolution (`resolve_int4_quant_config` -> `make_matmul_mixed_precision` -> `make_algo_config`) instead of `__init__` doing it inline.

src/python/py/models/builders/base.py
tianleiwu: Done in 73fa03f485. `resolve_int4_quant_config` no longer returns anything; it sets `self.quantization_algo` and `self.matmul_mixed_precision` directly.

src/python/py/models/builders/base.py
tianleiwu: Done in 73fa03f485 (renamed to `make_matmul_mixed_precision` in ef36112bcd). It now sets `self.int4_customized_weight_config` instead of returning it.

src/python/py/models/builders/base.py
tianleiwu: Fixed in 73fa03f485. `make_tied_quantized_embedding_input_names` no longer re-resolves the config on every call — it reuses the cached `self.quantization_algo`/`self.matmul_mixed_precision` (populated by `make_quant_init` during `__init__`) and only calls `resolve_int4_quant_config` if those attributes aren't set yet.

src/python/py/models/builders/base.py
tianleiwu: This is now handled by `matmul_mixed_precision` (ef36112bcd, f9fc9aa39e): a single extra option maps node-group selectors to a quant type, e.g. `matmul_mixed_precision=last_matmul:int8,linear_attn:int8`. Because it uses quant-type names (`int4`/`int8`, extensible to fp8/fp4) instead of int-specific boolean flags, the number of bits is specified per selector, and all mixed-precision placement is expressed through this one option rather than several per-node flags.

src/python/py/models/builders/base.py
tianleiwu: Same as https://github.com/microsoft/onnxruntime-genai/pull/2234#discussion_r3575073782 — this vendored `cuda_quantizer.py` is the current/latest copy. Once ORT's `CudaQuantizer` is updated to match, we'll drop the genai copy and import it from ORT.

src/python/py/models/builders/base.py
tianleiwu: These helpers register initializers into the builder's IR graph (`self.model.graph` / `self.make_value`), so they're currently tied to the builder rather than the pure quantizer. The FP4/ue8m0 encoding math itself is EP-agnostic and could move into `cuda_quantizer.py` (or ORT's quantizer tools) later, leaving just the graph-initializer registration here; MXFP4 QMoE is CUDA-only today. I'd prefer to do that extraction as a follow-up once the ORT-side `CudaQuantizer` update lands so we don't fork the logic twice.

src/python/py/models/builders/base.py
tianleiwu: Same as r3581639158: `make_mxfp4_weights` produces the FP4 e2m1 codes + ue8m0 block scales; the encoding math is EP-agnostic and a good candidate to move into `cuda_quantizer.py` / ORT's quantizer tools, while the initializer registration stays in the builder. I'll fold this into the same follow-up that consolidates the FP helpers once the ORT `CudaQuantizer` update is in.

src/python/py/models/builders/base.py
```diff
@@ -32,6 +32,9 @@
     GenerationConfig,
 )
 
+from .cuda_quantizer import CudaQuantizer
```
kunal-vaishnavi, line 35: Can we verify that this import style works for both `python -m onnxruntime_genai.models.builder` and `python builder.py` when the model builder is called? Typically, we have used try-except handling for this.

https://github.com/microsoft/onnxruntime-genai/blob/87047e6d77d55d20b54751e3f2da9d9883603c62/src/python/py/models/builders/base.py#L3719-L3722

src/python/py/models/builders/base.py
```diff
@@ -497,7 +544,46 @@ def make_attention_init(self, config):
     def make_lm_head_init(self, config):
         pass
 
+    # `matmul_mixed_precision` maps a node-group selector to the quantization type applied to that group
+    # (instead of the int4 body). Using a quant-type name rather than a bare bit count lets new schemes
+    # such as fp8/fp4 be added without introducing a new option. The legacy compound `int4_algo_config`
+    # names (e.g. "rtn_last", "k_quant_mixed") are desugared into these selectors by
+    # `quant_config.desugar_int4_algo_config`, the single source of truth for both surfaces.
+    MATMUL_MIXED_PRECISION_SELECTORS = ("last_matmul", "mixed_layers", "linear_attn")
+    MATMUL_MIXED_PRECISION_QUANT_TYPES = {
```
kunal-vaishnavi: Instead of storing these constants like this, can we keep them defined inside an init method (e.g. `make_quant_init`)?

src/python/py/models/builders/base.py
```diff
@@ -497,7 +544,46 @@ def make_attention_init(self, config):
...
+        "int4": 4,
+        "int8": 8,
+    }
+
+    @staticmethod
+    def onnx_dtype_to_precision(onnx_dtype):
```
kunal-vaishnavi, line 548: Instead of having this method, can we use `dtype.name` or `str(dtype)`?

src/python/py/models/builders/base.py
```diff
@@ -738,72 +833,274 @@ def save_processing(self, model_name_or_path, extra_kwargs, out_dir):
...
+        ``matmul_mixed_precision`` entries take precedence over the defaults implied by a legacy
+        compound name. An unknown base method is passed through unchanged and rejected later
+        (in `make_algo_config` / `make_tied_quantized_embedding_input_names`).
+        """
+        base_method, placement = desugar_int4_algo_config(extra_options)
+        self.quantization_algo = base_method
+        self.matmul_mixed_precision = placement
+
+    @classmethod
+    def normalize_matmul_mixed_precision(cls, value):
```
kunal-vaishnavi: I don't see this method used anywhere.

src/python/py/models/builders/gptoss.py
```diff
@@ -20,6 +21,122 @@ def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
...
+        for tensor_file in candidate_files:
+            with safe_open(tensor_file, framework="pt", device="cpu") as f:
+                if tensor_name in f.keys():
+                    tensor = f.get_tensor(tensor_name)
+                    self._mxfp4_tensor_cache[tensor_name] = tensor
+                    return tensor
+
+        raise RuntimeError(f"Original GPT-OSS MXFP4 tensor '{tensor_name}' was not found in {snapshot_dir}.")
+
+    @staticmethod
```
kunal-vaishnavi: Can we reduce the need for any `staticmethod` or `classmethod` decorators except when loading a model with `from_pretrained`?

src/python/py/models/builders/base.py
tianleiwu: yes. both works

src/python/py/models/builders/base.py
```diff
@@ -501,7 +544,39 @@ def make_attention_init(self, config):
     def make_lm_head_init(self, config):
         pass
 
+    @staticmethod
```
kunal-vaishnavi, line 547: Can we remove the static method decorator?

src/python/py/models/builders/base.py
```diff
@@ -3246,6 +3558,74 @@ def make_base_moe_op(self, name, **kwargs):
...
+        and tag it FLOAT8E8M0 (the encoding the CUDA QMoE FP4 kernel expects).
+        """
+        arr = scales_uint8.detach().cpu().numpy().astype(np.uint8) if isinstance(scales_uint8, torch.Tensor) else np.asarray(scales_uint8, dtype=np.uint8)
+        ir_tensor = ir.Tensor(np.ascontiguousarray(arr), dtype=ir.DataType.FLOAT8E8M0, name=name)
+        value = self.make_value(name, ir_tensor.dtype, ir_tensor.shape)
+        value.const_value = ir_tensor
+        self.model.graph.register_initializer(value)
+
+    # Positive FP4 e2m1 representable magnitudes (codes 0-7); negatives use codes 8-15.
+    _FP4_E2M1_POS_VALUES = (0.0, 0.5, 1.0, 1.5, 2.0, 3.0, 4.0, 6.0)
```
kunal-vaishnavi, line 3575: Can we register these constants in the constructor as `self._fp4_e2m1_descriptor` instead of putting them here?

kunal-vaishnavi (approved):

PR #2236, opened 2026-06-22
Cuda Plugin EP integration

tianleiwu:


baijumeswani:
Closing this since plugin ep now has the same name as the provider bridge ep. It should work out of the box now.

PR #2237, opened 2026-06-23
Fix tied embeddings

### LINKED ISSUE #2185: OGA export generating an invalid model when a lm_head unquantized model is exported
**Describe the bug**
OGA export generating an invalid model when a lm_head unquantized model is exported

**To Reproduce**
1. Please access the quantized model from here https://huggingface.co/amd/granite-4.0-1b-awq-grp128-exclude_lmhead_quark
2. Create OGA environment, I have tested from 0.12.1 to latest 0.13.2. I have been using `transformers==4.57.6` and `torch==2.12.0`
3. Export using this command

```
python -m onnxruntime_genai.models.builder `
    -i "granite_awq_lmhead_exclude" `
    -o "granite_oga_int4_lmhead_exclude" `
    -p int4 `
    -e cpu `
    --extra_options exclude_embeds=true int4_nodes_to_exclude=/lm_head/MatMul
```

4. See the message of invalid model

```
2026-05-26 13:21:43,041 onnxruntime.quantization.matmul_nbits_quantizer [INFO] - skip to quantize /model/constant_nodes/FLOAT/0.125 ...
2026-05-26 13:21:43,041 onnxruntime.quantization.matmul_nbits_quantizer [INFO] - skip to quantize /lm_head/Mul ...
2026-05-26 13:21:43,057 onnx_ir.serde [WARNING] - Input 'lm_head.MatMul.weight_Q4G32' of node '/model/embed_tokens/Reshape' (::Reshape:) cannot be found in any scope. The model is invalid but we will still create a new input for the node (current depth: 1)
2026-05-26 13:21:43,057 onnx_ir.serde [WARNING] - Input 'lm_head.MatMul.weight_scale' of node '/model/embed_tokens/GatherBlockQuantized' (com.microsoft::GatherBlockQuantized:) cannot be found in any scope. The model is invalid but we will still create a new input for the node (current depth: 1)
```

kunal-vaishnavi:
### Description

This PR rewrites how tied embeddings are determined in the model builder. It also adds a unit test file to ensure that the different cases for tied embeddings are tested in the CIs.

### Motivation and Context

There are still some small edge cases that haven't been validated for tied embeddings.


tianleiwu (commented):
Thanks for tightening up the tied-embeddings logic — splitting the decision into explicit `tied_quantized_embeddings` / `tied_unquantized_embeddings` flags (and gating on `exclude_embeds`/`exclude_lm_head`/`prune_lm_head` up front) is a clear improvement, and the new `make_tied_quantized_embedding_input_names` helper plus `test_tied_embeddings.py` give good coverage. This also fixes a latent mismatch where `make_algo_config` forced `/lm_head/MatMul` to 8 bits for plain `k_quant` while `make_embedding` used a 4-bit initializer name.

The two earlier bot threads ("MatMul quantized but Gather not" and `exclude_lm_head`) look addressed by the new guards on the current head. A couple of small items remain — see inline comments. Additionally: `test_onnxruntime_genai.py` drops the explicit `run_gemma4_vision_tests` / `run_qwen_fara_vision_tests` calls in favor of `pytest ... builder models`; since those helpers are `run_*` (not `test_*`) functions, please confirm they are still exercised somewhere so vision coverage isn't silently lost.

Non-blocking overall.

src/python/py/models/quantized_model.py
```diff
@@ -1258,6 +1258,7 @@ def __init__(self, quant_type, input_path, quant_attrs, q_size, kv_size, interme
         for i, layer in enumerate(self.layers):
             if i >= self.num_layers:
                 break
+            print(f"Unpacking and repacking layer {i}")
```
tianleiwu, line 1261: This looks like a leftover debug `print`. It sits inside the per-layer loop and will emit one line per layer on stdout during quantized model loading. Suggest removing it (or routing through a logger). Note there are two identical pre-existing `print(f"Unpacking and repacking layer {i}")` statements earlier in this file (~lines 1104 and 1179) — a good opportunity to clean up all three.

src/python/py/models/builder.py
```diff
@@ -192,7 +195,7 @@ def create_model(
     # Load model config
     extra_kwargs = {} if os.path.isdir(input_path) else {"cache_dir": cache_dir}
     hf_name = input_path if os.path.isdir(input_path) else model_name
-    hf_token = parse_hf_token(extra_options.get("hf_token", "true"))
+    hf_token = extra_options.get("hf_token", True)
```
tianleiwu, line 198: `parse_hf_token` now runs only inside `check_extra_options` (the CLI/`parse_extra_options` path). Both this line and `Model.__init__` (`self.hf_token = extra_options.get("hf_token", True)`) now assume the value is already parsed.

`create_model` is a public entry point imported by `examples/python/awq-quantized-model.py` and `tools/python/model_validation/validation_tool.py`. A programmatic caller passing `extra_options={"hf_token": "false"}` as a raw string will no longer have it converted to `None` — the literal string `"false"` is forwarded to `AutoConfig.from_pretrained(token=...)`, which treats any non-empty string as a real token and silently does **not** disable auth. Previously `parse_hf_token` handled this here. Consider normalizing `hf_token` inside `create_model`/`Model` too, or documenting that callers must pre-parse it.

src/python/py/models/builders/base.py
```diff
@@ -519,17 +519,33 @@ def make_quant_init(self, config):
...
+            and not self.exclude_lm_head
+            and not self.prune_lm_head
+        )
+
+        if shared_embeddings:
+            self.tied_quantized_embeddings = self.quantized_embeds and self.quantized_lm_head
+            self.tied_unquantized_embeddings = not self.tied_quantized_embeddings
+        else:
+            self.tied_unquantized_embeddings = False
+            self.tied_quantized_embeddings = False
```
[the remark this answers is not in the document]
tianleiwu: Follow-up on the current head: the new `make_tied_quantized_embedding_input_names` helper now emits `default`-algo-specific names (`lm_head.MatMul.weight_Q4` / `lm_head.MatMul.weight`) rather than the `_Q4G*` names, and the `exclude_lm_head=True` case is already guarded (`and not self.exclude_lm_head`), so those two parts appear resolved. One concrete gap remains: for `int4_algo_config=default` with `int4_is_symmetric=false`, the helper returns empty scale and zero-point names, so `make_embedding` builds `GatherBlockQuantized` with only `[weight, input_ids]`. `scales` is a required input of `GatherBlockQuantized`, so that combination would produce an invalid model. Worth either guarding it (fall back to no tying) or confirming it is unreachable.

tianleiwu (approved):

src/python/py/models/builder.py
tianleiwu: User will need `hf auth login` by default.  This change seems fine. If they want to disable it, they need pass hf_token as False instead of "false". Not a blocking issue.

PR #2242, opened 2026-06-25
Implement no_repeat_ngram_size for CPU search

### LINKED ISSUE #2241: no_repeat_ngram_size search option is parsed but never applied
## Description
The `no_repeat_ngram_size` search option is parsed from `genai_config.json` and is settable at runtime (via `set_search_options(...)` in Python and `OgaGeneratorParamsSetSearchNumber` in the C API), but it is never applied during generation. No logits processor enforces it, so the option silently has no effect.

In `src/config.h` the field was even annotated `// Unused param`.

## Impact
Greedy decoding can emit unbounded repeated n-grams. For example, encoder-decoder translation models can fall into a degenerate loop on noisy/degenerate input and repeat the same n-gram until they hit `max_length`, producing output like `当然当然当然...` / `谢谢谢...` instead of terminating.

## Repro
Using any model with greedy decoding (`do_sample=false`, `top_k=1`) and `no_repeat_ngram_size=3`, the generated sequence still contains repeated 3-grams. Setting the option has no observable effect on the output.

## Expected
`no_repeat_ngram_size=N` should ban any token that would complete an already-seen n-gram of size N, matching HuggingFace `no_repeat_ngram_size` semantics.

## Environment
- onnxruntime-genai 0.14.0.dev0
- CPU execution provider

mustjab:
The no_repeat_ngram_size search option was parsed from genai_config.json and settable via set_search_options / OgaGeneratorParamsSetSearchNumber, but no logits processor enforced it, so it had no effect. Greedy decoding could emit unbounded repeated n-grams (e.g. Marian translation looping to max_length on degenerate input).

Add a CPU logits processor that bans any token which would complete an already-seen n-gram of the configured size, matching HuggingFace no_repeat_ngram_size semantics.

Changes:
- search.h: virtual ApplyNoRepeatNgram no-op default (CUDA/beam unaffected), Search_Cpu override.
- search.cpp: ApplyNoRepeatNgram implementation.
- generators.cpp, engine/request.cpp: invoke after ApplyRepetitionPenalty.
- config.h: document the field.
- sampling_tests.cpp: NoRepeatNgramCorrectnessCpu unit test.


src/search.cpp
```diff
@@ -580,4 +580,43 @@ void Search_Cpu::ApplyRepetitionPenalty(float penalty) {
   }
 }
 
+void Search_Cpu::ApplyNoRepeatNgram(int ngram_size) {
```
kunal-vaishnavi, line 587: Do you plan to add a CUDA implementation for `no_repeat_ngram_size`? If not, can we add a one-time warning somewhere so that if `no_repeat_ngram_size` is attempted, the gap is documented?

mustjab: Wasn't planning to add CUDA implementation but did update change with documentation. Let me know if you think that we should implement this for CUDA as well.

kunal-vaishnavi: We should implement this for CUDA as well. Otherwise, there won't be feature parity across EPs.

kunal-vaishnavi (dismissed):

src/search.h
```diff
@@ -28,6 +30,22 @@ struct Search : LeakChecked<Search> {
   // Scoring features
   virtual void ApplyMinLength(int min_length) = 0;
   virtual void ApplyRepetitionPenalty(float penalty) = 0;
+  // Bans tokens that would complete an already-seen n-gram of size ngram_size.
+  // Backends that don't implement it (e.g. CUDA) fall back to this default, which
+  // warns once if the option is set so the request is not silently ignored. Defined
+  // inline (like the other Search defaults) so it links into the standalone CUDA
+  // shared library, which does not compile search.cpp.
+  virtual void ApplyNoRepeatNgram(int ngram_size) {
```
baijumeswani, line 33: Why warn instead of hard fail? If user requested it and it does not work on cuda, I would expect it to fail.

mustjab: Agreed. I changed the default backend implementation to throw when a positive no_repeat_ngram_size is requested. CUDA will now fail during GenerateNextToken() instead of warning and continuing without enforcing the constraint. The CPU implementation is unchanged.

baijumeswani (approved):

PR #2244, opened 2026-06-26
webgpu: fix RecurrentState graph capture with shared buffer aliasing

qjia7:
## Summary

- `RecurrentState::share_buffers_` now derives from `past_present_share_buffer` config (same pattern as `DefaultKeyValueCache`) instead of being hardcoded to `false` for WebGPU
- Add a graph-capture guard that throws when `share_buffers_` is false and graph capture is requested, matching the existing guard in `kv_cache.cpp`
- Add `Ort::Experimental::Get_*_Fn` accessor shim to `onnxruntime_api.h` so genai builds against ORT builds where `onnxruntime_experimental_cxx_api.h` cannot be included directly

## Motivation

WebGPU graph capture freezes GPU buffer handles at the first captured decode step. `RecurrentState` was using separate past/present buffers for WebGPU with a per-step C++ pointer swap — but graph capture ignores those swaps on replay, so every decode step after the first read a stale buffer, producing garbage output.

With `past_present_share_buffer: true`, the ORT `LinearAttention` and `CausalConvWithState` kernels detect the aliased buffers via `initial_state_in_present_state` / `conv_state_in_present_state` and switch to a single `read_write` binding, which satisfies the WebGPU spec constraint that originally motivated the per-device split. Buffer handles stay stable across graph capture replays.

The original comment in the code noted a TODO to remove the WebGPU special case once the ORT WebGPU EP kernels natively supported past/present buffer sharing — that support is already present, so this PR resolves the TODO.

## Test plan

- [x] `Qwen3.5-0.8B-webgpu-fused` correctness test: all queries pass with `enableGraphCapture=1`
- [x] `Qwen3.5-0.8B-webgpu-fused` multi-gen test: sequential and overlapping generators pass with `enableGraphCapture=1`
- [x] Both tests also pass with `enableGraphCapture=0` (no regression on non-graph-capture path)
- [x] clang-format: no issues on changed files
- [x] Build: succeeds against local ORT WebGPU EP build


src/models/onnxruntime_api.h
```diff
@@ -88,6 +88,25 @@ p_session_->Run(nullptr, input_names, inputs, std::size(inputs), output_names, o
...
+  inline OrtExperimental_##NAME##_SinceV##VER##_Fn Get_##NAME##_SinceV##VER##_Fn(      \
+      const OrtApi* api) {                                                             \
+    return reinterpret_cast<OrtExperimental_##NAME##_SinceV##VER##_Fn>(                \
+        api->GetExperimentalFunction(kOrtExperimental_##NAME##_SinceV##VER##_FnName)); \
+  }
+#include "onnxruntime_experimental_c_api.inc"
+#undef ORT_EXPERIMENTAL_API
+}  // namespace Experimental
+}  // namespace Ort
+#endif
```
qjia7: Above is to fix a build break when GenAI main is compiled against an ORT main that exposes ORT_API_VERSION >= 28.

Without it, MSVC fails on every Model Package call site in onnxruntime_inline.h.
Compiler output (MSVC, 30+ identical-shape errors at src/models/onnxruntime_inline.h lines 1539–1561):
```
error C2653: 'Experimental': is not a class or namespace name
error C3861: 'Get_OrtModelPackageApi_CreateModelPackageOptionsFromSessionOptions_SinceV28_Fn': identifier not found
error C3861: 'Get_OrtModelPackageApi_ReleaseModelPackageOptions_SinceV28_Fn': identifier not found
error C3861: 'Get_OrtModelPackageApi_CreateModelPackageContext_SinceV28_Fn': identifier not found
error C3861: 'Get_OrtModelPackageApi_ReleaseModelPackageContext_SinceV28_Fn': identifier not found
error C3861: 'Get_OrtModelPackageApi_ModelPackage_GetComponentCount_SinceV28_Fn': identifier not found
error C3861: 'Get_OrtModelPackageApi_ModelPackage_GetComponentNames_SinceV28_Fn': identifier not found
error C3861: 'Get_OrtModelPackageApi_ModelPackage_GetVariantCount_SinceV28_Fn': identifier not found
error C3861: 'Get_OrtModelPackageApi_ModelPackage_GetVariantNames_SinceV28_Fn': identifier not found
error C3861: 'Get_OrtModelPackageApi_ModelPackage_GetVariantEpName_SinceV28_Fn': identifier not found
error C3861: 'Get_OrtModelPackageApi_SelectComponent_SinceV28_Fn': identifier not found
error C3861: 'Get_OrtModelPackageApi_ReleaseModelPackageComponentContext_SinceV28_Fn': identifier not found
error C3861: 'Get_OrtModelPackageApi_ModelPackageComponent_GetSelectedVariantFolderPath_SinceV28_Fn': identifier not found
```

src/models/onnxruntime_api.h
```diff
@@ -88,6 +88,25 @@ p_session_->Run(nullptr, input_names, inputs, std::size(inputs), output_names, o
 #define ORT_GENAI_HAS_EXPERIMENTAL_C_API 1
 #endif
 
+// Expose Ort::Experimental::Get_*_Fn accessors used by onnxruntime_inline.h without pulling
```
kunal-vaishnavi: From Copilot:

I’m a bit concerned about this approach. We’re reopening `Ort::Experimental` in a central header and expanding `onnxruntime_experimental_c_api.inc` ourselves to recreate the `Get_*_Fn` accessors. That works around the include conflict, but it also means we now have a local re-implementation of part of ORT’s C++ experimental surface that has to stay in sync with upstream.

Could we instead move this behind a small genai-private compat shim/helper for just the experimental functions we actually need, rather than adding new ad-hoc definitions under `namespace Ort` here?

qjia7: Done

src/models/onnxruntime_api.h
baijumeswani: Can you explain why the issue shows up in your build but not in the CI? Should we use ORT_GENAI_HAS_MODEL_PACKAGE macro in case there is a version issue here?

qjia7: > Can you explain why the issue shows up in your build but not in the CI?

CI without `ORT_HOME` uses the download path in `cmake/ortlib.cmake`, which pins `ORT_VERSION = 1.26.0`. That release ships `ORT_API_VERSION = 26`, so the entire ModelPackageApi block in `onnxruntime_inline.h` (guarded by `#if ORT_API_VERSION >= 28 && ORT_GENAI_HAS_EXPERIMENTAL_C_API`) is preprocessed out — the `Ort::Experimental::Get_*_Fn` references are never seen by MSVC and the build succeeds.

My local build uses `--ort_home` pointing to ORT `main`, which is at `ORT_API_VERSION = 28`. The gate opens, MSVC compiles the block, and fails with C2653/C3861 on `Ort::Experimental::...` because the C++ experimental header (`onnxruntime_experimental_cxx_api.h`) is where that namespace lives and we can't include it — it transitively pulls in `onnxruntime_cxx_api.h`, which conflicts with genai's vendored `Ort::` wrappers in `src/models/onnxruntime_api.h`.

The fix in this PR side-steps that: `GENAI_MP_V28_FN(NAME)` reinterprets `api->GetExperimentalFunction(k...)` to the C typedef directly, so we only need `onnxruntime_experimental_c_api.h`, which genai already includes safely.

> Should we use `ORT_GENAI_HAS_MODEL_PACKAGE` macro in case there is a version issue here?

Good suggestion — the version gate is already there (`onnxruntime_api.h` defines `ORT_GENAI_HAS_MODEL_PACKAGE` as exactly `ORT_API_VERSION >= 28 && ORT_GENAI_HAS_EXPERIMENTAL_C_API`), but `onnxruntime_inline.h` was restating the raw condition rather than using the macro. Pushed 3897d65d to route both files through the single named macro so they can't drift.

src/models/onnxruntime_inline.h
```diff
@@ -1541,30 +1541,29 @@ inline const ModelPackageApi& GetModelPackageApi() {
...
-    f.ModelPackage_GetVariantEpName =
-        Experimental::Get_OrtModelPackageApi_ModelPackage_GetVariantEpName_SinceV28_Fn(api);
-    f.SelectComponent =
-        Experimental::Get_OrtModelPackageApi_SelectComponent_SinceV28_Fn(api);
-    f.ReleaseModelPackageComponentContext =
-        Experimental::Get_OrtModelPackageApi_ReleaseModelPackageComponentContext_SinceV28_Fn(api);
+    // Local resolver for experimental function pointers. Kept inline here so we don't have
+    // to include onnxruntime_experimental_cxx_api.h (which transitively includes
+    // onnxruntime_cxx_api.h and conflicts with genai's vendored Ort wrappers).
+#define GENAI_MP_V28_FN(NAME)                                                \
```
baijumeswani, line 1549: Why do we need to change this part of the code? It seems with the macro, we shouldn't need to

```
#define GENAI_MP_V28_FN(NAME)
...
#undef GENAI_MP_V28_FN
```

qjia7: Good question — the two macros serve different purposes and both are needed:

- **`ORT_GENAI_HAS_MODEL_PACKAGE`** is the *outer gate*. It decides whether this whole block is compiled at all (`ORT_API_VERSION >= 28` AND the experimental C header is available).
- **`GENAI_MP_V28_FN`** is the *inner resolver*. It decides *how* the function pointers are looked up once we're inside the gate.

The old code called `Ort::Experimental::Get_OrtModelPackageApi_*_SinceV28_Fn(api)`. Those `Get_*_Fn` accessors are C++ inline functions defined **only** in `onnxruntime_experimental_cxx_api.h`. We can't include that header here because it transitively pulls in `onnxruntime_cxx_api.h`, which redefines the `Ort::` types that genai has vendored in `src/models/onnxruntime_api.h` — you'd get a wall of redefinition errors.

So the macro replaces the C++ accessor calls with a direct `reinterpret_cast` over `api->GetExperimentalFunction(k...SinceV28_FnName)`, which only depends on the C typedefs and name constants from `onnxruntime_experimental_c_api.h` (already included). Dropping the macro and going back to `Experimental::Get_*_Fn(api)` would re-introduce the exact ORT_HOME ≥ v28 build break this PR fixes — `ORT_GENAI_HAS_MODEL_PACKAGE` alone doesn't help, because inside the gate the C++ accessors are still unreachable without the cxx header.

Also tightened the comment above the `#define` in 0a19108e to name the specific accessors being replaced so this trade-off is visible at the call site.

baijumeswani (approved):

PR #2245, opened 2026-06-26
webgpu: bypass manual mRoPE for text-only Qwen3.5 when GQA fuses RoPE

qjia7:
## Summary

For text-only Qwen3.5, multi-head RoPE (mRoPE) collapses to standard 1D RoPE:
`Qwen3_5TextRotaryEmbedding` expands a 2D `position_ids` into 3 identical axes
and `apply_interleaved_mrope` returns `freqs[0]` unchanged. The manual mRoPE
subgraph (Shape → Expand → interleaved cos/sin caches → custom kernel) is
therefore equivalent to a plain fused-RoPE pass inside GQA.

When the GQA operator supports fused RoPE (`use_rope_in_attn=True`, e.g. on
WebGPU), this PR detects the text-only case and routes through the fused path,
bypassing the manual mRoPE subgraph entirely. This removes the `Shape →
Memcpy` node that reads a dynamic tensor shape at runtime — the path that
prevents WebGPU graph capture on Qwen3.5 text-only models.

**Changes** (`src/python/py/models/builders/qwen.py` only):
- Add `use_text_only_fused_rope` flag: true when `is_text_only` and `use_rope_in_attn`.
- When flag is set: call `make_rotary_embedding_caches()` (standard 2D cos/sin for GQA), skip mRoPE config, leave `use_rope_in_attn=True`.
- When flag is not set: keep existing mRoPE path unchanged (VL mode and non-fused-RoPE EPs).
- `make_position_ids_reformatting`: early-return `None` when fused RoPE is active (no `position_ids` tensor on the data flow).

## Test plan

- [ ] Verify text-only Qwen3.5-0.8B generates correct output on WebGPU with graph capture enabled
- [ ] Verify multimodal Qwen3.5 (VL mode) is unaffected — still uses manual mRoPE path
- [ ] Run existing Qwen integration tests: `qwen3-0.6b`, `qwen2.5-0.5b-instruct`

### DIFF
--- a/src/python/py/models/builders/qwen.py
+++ b/src/python/py/models/builders/qwen.py
@@ -1014,31 +1014,45 @@ class Qwen35TextModel(Model):
         # SkipSimplifiedLayerNormalization can be used directly.
         self.layernorm_attrs["add_offset"] = 1
 
-        # Position IDs input.
-        # In text-only mode the runtime provides standard 2D [B, S] position_ids.
-        # We expand them to 3D [3, B, S] inside the graph so mRoPE works unchanged.
-        # In VL mode the pipeline provides 3D position_ids directly.
-        if self.is_text_only:
-            self.input_shapes["position_ids"] = ["batch_size", "sequence_length"]
+        # Text-only mRoPE collapses to standard 1D RoPE because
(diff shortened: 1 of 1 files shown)


qjia7:
Close it since we have mrope support.

PR #2251, opened 2026-07-01
[model-builder] Cap max_length at 4096 to reduce KV cache memory usage

daijh:
This change limits the default max_length in `genai_config.json` to **4096** tokens (or the model's native context_length if smaller) to reduce memory consumption for KV cache allocation, especially important resource-constrained environments.

For models with very large context windows (e.g., 131K tokens), this significantly reduces initial memory allocation when using `past_present_share_buffer=true`.

### DIFF
--- a/src/python/py/models/builders/base.py
+++ b/src/python/py/models/builders/base.py
@@ -618,7 +618,8 @@ class Model:
                 "do_sample": config.do_sample if hasattr(config, "do_sample") else False,
                 "early_stopping": True,
                 "length_penalty": config.length_penalty if hasattr(config, "length_penalty") else 1.0,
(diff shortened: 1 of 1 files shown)


daijh:
@guschmue @qjia7 @huningxin

kunal-vaishnavi:
The size of the pre-allocated KV cache can be managed by overriding the `max_length` value in the GenAI config by [setting generator params](https://github.com/microsoft/onnxruntime-genai/blob/97227263e8d4407d3895f43c19fb11afc1289dd9/examples/python/common.py#L627-L629). This change is not needed.

daijh:
Thanks. Close as resolved.

PR #2252, opened 2026-07-01
[model_qa.py] Warn when an unrecognized EP name falls back to CPU

daijh:
A provider name that is neither a built-in provider nor a registered plugin EP is registered via the legacy append path with a CPU device, leaving model inputs and the KV cache in CPU memory. This was silent and could cause a large, hard-to-diagnose slowdown. Log a warning (primary session only, when V2 plugin registration fails) suggesting the canonical genai provider name.

Also add "webgpu" as an accepted -e choice in the Python example, mirroring the existing cuda/CUDAExecutionProvider pairing.

### DIFF
--- a/src/models/session_options.cpp
+++ b/src/models/session_options.cpp
@@ -216,9 +216,8 @@ DeviceInterface* SetProviderSessionOptions(OrtSessionOptions& session_options,
         // instead of silently degrading performance.
         if (is_primary_session_options && g_log.enabled && g_log.warning) {
           Log("warning", "Execution provider '" + provider_options.name +
-                             "' is not a built-in provider or a registered plugin EP; "
-                             "the model will run on CPU, which may be slow. For a built-in "
-                             "provider, use its canonical genai name (e.g. \"webgpu\", \"cuda\", \"dml\").");
+                             "' is not a built-in provider or a registered plugin EP; GenAI will keep model inputs and the KV cache in CPU memory, which may be slow. If you intended a built-in "
+                             "provider, use its canonical GenAI name (e.g. \"webgpu\", \"cuda\", \"dml\").");
         }
         AppendExecutionProviderV1(session_options, provider_options);
       }
--- a/src/models/session_options.cpp
+++ b/src/models/session_options.cpp
@@ -205,8 +205,21 @@ DeviceInterface* SetProviderSessionOptions(OrtSessionOptions& session_options,
         device = session_device;  // Set the device if not already set by a previous provider
       }
     } else {
+      // Not a built-in provider: register it as a plugin EP (V2), falling back to the
+      // legacy append (V1). Neither path returns a genai DeviceInterface, so the
+      // model's inputs and KV cache stay in CPU memory for this session.
       if (!AppendExecutionProviderV2(session_options, provider_options,
                                      DeviceType::CPU, provider_options.name)) {
+        // The name is neither a built-in provider nor a registered plugin EP. The
+        // usual cause is a misspelled built-in name (e.g. an ORT-style name that
+        // NormalizeProviderName does not map). Warn so the CPU fallback is visible
+        // instead of silently degrading performance.
+        if (is_primary_session_options && g_log.enabled && g_log.warning) {
+          Log("warning", "Execution provider '" + provider_options.name +
+                             "' is not a built-in provider or a registered plugin EP; "
+                             "the model will run on CPU, which may be slow. For a built-in "
+                             "provider, use its canonical genai name (e.g. \"webgpu\", \"cuda\", \"dml\").");
+        }
         AppendExecutionProviderV1(session_options, provider_options);
       }
     }
--- a/examples/python/common.py
+++ b/examples/python/common.py
@@ -569,6 +569,7 @@ def get_ep_args(parser: argparse.ArgumentParser) -> None:
         "OpenVINOExecutionProvider",       # Intel IHV EP
         "QNNExecutionProvider",            # Qualcomm IHV EP
         "VitisAIExecutionProvider",        # AMD IHV EP
+        "webgpu",                          # GenAI canonical name for WebGPU EP
         "WebGpuExecutionProvider",         # WebGPU EP
     ]
 


daijh:
model-qa.py with `-e WebGpuExecutionProvider` is ~x11 slower than without.

```shell
python examples\python\model-qa.py -m model -e WebGpuExecutionProvider
```

```shell
python examples\python\model-qa.py -m model
```

daijh:
# Root Cause & Fix: WebGPU Slowdown with `past_present_share_buffer`

## The symptom

Running the **same model** with `past_present_share_buffer: true` on WebGPU:

| Invocation | New tokens/sec |
|---|---|
| `model-qa.py` (no `-e`, i.e. `follow_config`) | ~30 tps |
| `model-qa.py -e WebGpuExecutionProvider` | ~1.8 tps |

It looked like "Python is slow" or "the config flag is bad." Both were red herrings.
The flag is only an *amplifier*. The real trigger is the `-e` flag value.

---

## Root cause: a provider name that doesn't resolve to the WebGPU device

### The mechanism, step by step

**1. What `-e WebGpuExecutionProvider` does.** `get_config` clears the config's
providers and appends the name you passed verbatim:

```python
config.clear_providers()
config.append_provider("WebGpuExecutionProvider")   # the literal -e value
```

**2. genai maps that name to a device via a fixed dispatch table** keyed on the
**canonical** name `"WebGPU"` (`src/models/session_options.cpp`).

**3. Name normalization is supposed to bridge the two spellings** —
`NormalizeProviderName` (`src/config.cpp`) lowercases and maps aliases to the canonical
key. **But in the installed genai 0.14.1, it does not strip the `ExecutionProvider`
suffix.** So:

- `"webgpu"` → `"WebGPU"` ✅ matches the table
- `"WebGpuExecutionProvider"` → (unchanged in 0.14.1) ❌ misses the table

Proven with a controlled three-way test (same process, `model_input_values` logging):

```
append_provider("WebGPU")                  -> KV cache Location: GPU
append_provider("webgpu")                  -> KV cache Location: GPU
append_provider("WebGpuExecutionProvider") -> KV cache Location: CPU   <-- !!
```

**4. The miss falls through to a plugin-EP registration path** that registers the
provider with a CPU device. The model's device stays CPU, so genai **allocates the KV
cache in CPU memory** (confirmed by the `Location: CPU` logging above).

**5. `past_present_share_buffer: true` turns the misplacement into a catastrophe.**
With the shared buffer on, the KV cache isn't grown incrementally — it's pre-allocated
once to the **full `max_length` (4096)** and reused. So you have a 4096-long KV buffer
across all 32 layers, fp16, living in CPU memory, that must be marshalled across the
CPU<->GPU boundary every single decode step. That host/device round-trip per token is
the ~16x collapse.

### Why each earlier observation happened

- **Fast originally** — ran without `-e`, i.e. `follow_config`, which keeps the config's
  name `"webgpu"` → normalizes to `"WebGPU"` → GPU device.
- **Slow with `-e WebGpuExecutionProvider`** — hits the dispatch-table miss → CPU device.
- **`past_present_share_buffer: false` fixed it** — removes the giant pre-sized buffer.
  The cache then grows incrementally, so even on the (wrong) CPU device the per-step
  copy is small and cheap.

### The one-sentence root cause

> In genai 0.14.1, `"WebGpuExecutionProvider"` is not normalized to the canonical
> `"WebGPU"`, so it misses the EP dispatch table and is registered as a CPU-device
> plugin EP; the KV cache then lives in CPU memory, and `past_present_share_buffer=true`
> (which pre-sizes that buffer to `max_length`) makes every decode step pay a full
> CPU<->GPU copy.

`past_present_share_buffer` is the **amplifier**; the **wrong device from the unresolved
name** is the root.

---

## The fix

### Examples accept the canonical name (verified)

`examples/python/common.py`:

```python
"VitisAIExecutionProvider",        # AMD IHV EP
"webgpu",                          # GenAI canonical name for WebGPU EP   <-- added
"WebGpuExecutionProvider",         # WebGPU EP
```

The CLI previously offered *only* `WebGpuExecutionProvider` — the one spelling that
breaks on 0.14.1. Adding `webgpu` mirrors how CUDA already exposes both `cuda` and
`CUDAExecutionProvider`.

**Verified on 0.14.1:** `-e webgpu` → **31.49 tps** (vs 1.8 with the ORT name). Works
with no rebuild, because `webgpu` normalizes correctly even on the old binary.

### Core warns instead of silently falling back to CPU

`src/models/session_options.cpp` — when a provider name is neither a built-in provider
nor a registered plugin EP, it is registered via the legacy path with a CPU device
(leaving model inputs + the KV cache in CPU memory). It now logs a warning so that
fallback is visible:

```cpp
} else {
  // Not a built-in provider: register it as a plugin EP (V2), falling back to the
  // legacy append (V1). Neither path returns a genai DeviceInterface, so the
  // model's inputs and KV cache stay in CPU memory for this session.
  if (!AppendExecutionProviderV2(session_options, provider_options,
                                 DeviceType::CPU, provider_options.name)) {
    // The name is neither a built-in provider nor a registered plugin EP. The
    // usual cause is a misspelled built-in name (e.g. an ORT-style name that
    // NormalizeProviderName does not map). Warn so the CPU fallback is visible
    // instead of silently degrading performance.
    if (is_primary_session_options && g_log.enabled && g_log.warning) {
      Log("warning", "Execution provider '" + provider_options.name +
                         "' is not a built-in provider or a registered plugin EP; "
                         "the model will run on CPU, which may be slow. For a built-in "
                         "provider, use its canonical genai name (e.g. \"webgpu\", \"cuda\", \"dml\").");
    }
    AppendExecutionProviderV1(session_options, provider_options);
  }
}
```

The warning fires only when V2 plugin registration fails (so it does not spam on
legitimate `--ep_path` plugin EPs) and only for the primary session. This is the
**diagnostic** fix: the entire investigation existed because the CPU fallback was
silent. Note it is symbol-checked against the codebase but **not runtime-verified**
(C++, needs a genai rebuild).

---

daijh:
@guschmue @qjia7 @huningxin

src/models/session_options.cpp
```diff
@@ -205,8 +205,20 @@ DeviceInterface* SetProviderSessionOptions(OrtSessionOptions& session_options,
         device = session_device;  // Set the device if not already set by a previous provider
       }
     } else {
+      // Not a built-in provider: register it as a plugin EP (V2), falling back to the
+      // legacy append (V1). Neither path returns a genai DeviceInterface, so the
+      // model's inputs and KV cache stay in CPU memory for this session.
       if (!AppendExecutionProviderV2(session_options, provider_options,
                                      DeviceType::CPU, provider_options.name)) {
+        // The name is neither a built-in provider nor a registered plugin EP. The
```
kunal-vaishnavi, line 213: This doesn't make sense as the legacy EP registration via V1 is still attempted after the `if` condition below.

PR #2261, opened 2026-07-02
Avoid disclosing config file contents in parse error messages

jiafatom:
## Description

When `genai_config.json` fails to parse, `ParseConfig` (in `src/config.cpp`) rethrew the raw JSON parser message. That message can embed verbatim content from the file being parsed — for example, a JSON key name (`Unknown value "<key>"`). If the config path is influenced by untrusted input, this echoes contents of the targeted file back through the error string.

## Change

- In `ParseConfig`, on a JSON parse failure, keep the file path in the error message (so the failing file can still be identified) but no longer append the parser's raw message content.
- Added `CAPITests.ConfigParseErrorDoesNotLeakContent` (no model required): writes a `genai_config.json` containing an unknown key, calls `OgaConfig::Create`, and asserts the resulting error names `genai_config.json` and contains the parse-error prefix, but does **not** contain the key from the file. Uses a unique temp directory to avoid parallel-run collisions.

## Scope / non-goals

This intentionally does not alter path handling. Accepting an arbitrary filesystem path is the intended behavior of `OgaCreateConfig` / `OgaCreateModel`; validating which paths are permissible is the responsibility of the calling application, and adding path/`../`/UNC restrictions here would break legitimate use cases (relative paths, models on network shares).

## Testing

Built and ran the affected tests:

```
[ RUN      ] CAPITests.Config
[       OK ] CAPITests.Config
[ RUN      ] CAPITests.ConfigParseErrorDoesNotLeakContent
[       OK ] CAPITests.ConfigParseErrorDoesNotLeakContent
[  PASSED  ] 2 tests.
```

Replaces #2258 (recreated as a non-fork branch so CI can run).

### DIFF
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -1600,10 +1600,13 @@ void ParseConfig(const fs::path& filename, std::string_view json_overlay, Config
   RootObject_Element root_object{root};
   try {
     JSON::Parse(root_object, std::string_view(buffer.data(), buffer.size()));
-  } catch (const std::exception& message) {
-    std::ostringstream oss;
-    oss << "Error encountered while parsing '" << filename.string() << "' " << message.what();
-    throw std::runtime_error(oss.str());
+  } catch (const std::exception&) {
+    // Intentionally omit the parser's message here: it can echo verbatim
+    // content (such as JSON key names) from the file being parsed. When the
+    // path is influenced by untrusted input this would disclose file contents
+    // through the error message. The path is retained so the failing file can
+    // still be identified.
+    throw std::runtime_error("Error encountered while parsing '" + filename.string() + "'");
   }
 
   if (!json_overlay.empty()) {
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -1600,13 +1600,19 @@ void ParseConfig(const fs::path& filename, std::string_view json_overlay, Config
   RootObject_Element root_object{root};
   try {
     JSON::Parse(root_object, std::string_view(buffer.data(), buffer.size()));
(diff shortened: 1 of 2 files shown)


src/config.cpp
```diff
@@ -1600,10 +1600,13 @@ void ParseConfig(const fs::path& filename, std::string_view json_overlay, Config
   RootObject_Element root_object{root};
   try {
     JSON::Parse(root_object, std::string_view(buffer.data(), buffer.size()));
-  } catch (const std::exception& message) {
-    std::ostringstream oss;
-    oss << "Error encountered while parsing '" << filename.string() << "' " << message.what();
-    throw std::runtime_error(oss.str());
+  } catch (const std::exception&) {
+    // Intentionally omit the parser's message here: it can echo verbatim
```
kunal-vaishnavi: Sometimes, it helps to know which key name was not parsed. That way, we know that we need to add parsing for that key in our JSON parser.

Can we print that information but gate it under specific build conditions (e.g. RelWithDebInfo or Debug builds) and omit it in other build conditions (e.g. Release)?

baijumeswani: This change feels overly restrictive. We should be able to tell what failed imo (even in non-debug builds).
What is the concern with the error message containing the contents of the config? Most of these configs come from models that we build.

kunal-vaishnavi: I agree that we should tell what failed in all cases but this looks to be resolving the [following IcM ticket](https://portal.microsofticm.com/imp/v5/incidents/details/31000000580572/summary) filed by MSRC.

PR #2262, opened 2026-07-03
Moonshine Model Streaming Test Development

themason2011:
Adding C++ and Python tests to support the new Moonshine model. Used pre-existing test infrastructure from the Nemotron model

### DIFF
--- a/test/c_api_tests.cpp
+++ b/test/c_api_tests.cpp
@@ -1551,6 +1551,20 @@ class CAPITests : public ::testing::TestWithParam<StreamingASRModel> {
   size_t ChunkSamples() const { return GetParam().chunk_samples; }
 };
 
+// Helper: a streaming ASR model has VAD enabled by default only when its genai_config.json
+// declares a "vad" section AND the referenced silero_vad.onnx file is present in the model dir.
+static bool ModelHasVad(const std::string& model_path) {
+  const auto vad_path = std::filesystem::path(model_path) / "silero_vad.onnx";
+  if (!std::filesystem::exists(vad_path))
+    return false;
+  std::ifstream config_file(std::filesystem::path(model_path) / "genai_config.json");
+  if (!config_file)
+    return false;
+  const std::string config_json((std::istreambuf_iterator<char>(config_file)),
+                                std::istreambuf_iterator<char>());
+  return std::regex_search(config_json, std::regex(R"("vad"\s*:\s*\{)"));
+}
+
 // Helper: if mel is not null, set inputs and run the decode loop
 static void DecodeInputs(OgaGenerator& generator, OgaNamedTensors* mel) {
   if (mel) {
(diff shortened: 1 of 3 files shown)


test/c_api_tests.cpp
```diff
@@ -1533,13 +1534,22 @@ TEST(CAPITests, RewindQwen25CAPI) {
 }
 #endif  // TEST_QWEN_2_5
 
-#ifndef STREAMING_ASR_PATH
-#define STREAMING_ASR_PATH MODEL_PATH "nemotron-speech-streaming"
-#endif
+// Streaming ASR models exercised by the StreamingASR* tests below, each paired
+// with its chunk size (samples per Process() call).
+struct StreamingASRModel {
+  const char* subdir;
```
nenad1002: any reason not to just use string as you construct it later anyway?

test/python/test_onnxruntime_genai_api.py
```diff
@@ -1084,15 +1084,24 @@ def test_streaming_asr_sine_wave(nemotron_speech_model_path):
...
 
 
-def test_streaming_asr_config_model_type(nemotron_speech_model_path):
-    """Test that a nemotron_speech model reports the correct type."""
-    model = og.Model(nemotron_speech_model_path)
-    assert model.type == "nemotron_speech"
+# Expected model "type" reported by each ASR streaming model's genai_config.json.
+_ASR_MODEL_TYPES = {
+    "nemotron-speech-streaming": "nemotron_speech",
+    "moonshine-streaming-small-official": "streaming_enc_dec_asr",
```
nenad1002: -official is just the name of the model i have in hf, but it shoudn't have that suffix in code

test/python/test_onnxruntime_genai_api.py
```diff
@@ -1166,7 +1175,15 @@ def normalize(text):
     return d[len(r)][len(h)] / max(len(r), 1)
 
 
-def test_streaming_asr_transcription_quality(nemotron_speech_model_path, test_data_path):
+# Per-model word-error-rate thresholds for the transcription-quality test.
+_ASR_WER_THRESHOLDS = {
```
nenad1002, line 1179: does it consistently pass?

test/c_api_tests.cpp
themason2011: you're right, changed to string

test/python/test_onnxruntime_genai_api.py
themason2011: done

test/python/test_onnxruntime_genai_api.py
themason2011: Yes, ran 20 times and passed every time, indicating > 95% success threshold at least (at least one would have failed to be below 95%)

test/c_api_tests.cpp
```diff
@@ -1712,36 +1764,134 @@ TEST(CAPITests, StreamingASRVadSetGetOption) {
...
+  const size_t chunk_samples = static_cast<size_t>(ReadConfigNumber(model_path, "chunk_samples", 8000));
+  const float sample_rate = static_cast<float>(ReadConfigNumber(model_path, "sample_rate", 16000));
+  const double seconds_per_memory_frame = ReadConfigNumber(model_path, "seconds_per_memory_frame", 0.02);
+  const int min_segment_memory_frames =
+      static_cast<int>(ReadConfigNumber(model_path, "min_segment_memory_frames", 250));
+
+  constexpr float frequency = 440.0f;
+  std::vector<float> speech(chunk_samples);
+  for (size_t i = 0; i < chunk_samples; ++i)
+    speech[i] = 0.5f * std::sin(2.0f * 3.14159265f * frequency * static_cast<float>(i) / sample_rate);
```
nenad1002: Can this be reused for future tests? If so, pull into a separate method

nenad1002 (approved):

test/c_api_tests.cpp
themason2011: Pulled in a separate method, thank you

PR #2264, opened 2026-07-06
fix(ci): use GHCR instead of ACR for workflow docker caches

sanaa-hamel-microsoft:
Azure resources should be read-only from GH workflows.


tools/ci_build/get_docker_image.py
```diff
@@ -64,7 +64,7 @@ def main():
         log.info("No container registry will be used")
 
     full_image_name = (
-        f"{args.container_registry}.azurecr.io/{args.repository}:latest"
+        f"{args.container_registry}/{args.repository}:latest"
```
tianleiwu: The semantics of `--container-registry` change here from "ACR short name" (`<name>` -> `<name>.azurecr.io/<repo>`) to "full registry host + namespace" (`ghcr.io/microsoft/onnxruntime-genai`, `onnxruntimebuildcache.azurecr.io`). All in-repo callers were updated, but the argparse `description` (line 24, "push it to a remote Azure Container Registry") and this flag's `help` (line 37, "The Azure container registry name") still describe the old ACR-only contract. A future caller who follows the help and passes a bare `onnxruntimebuildcache` now silently gets an invalid ref `onnxruntimebuildcache/<repo>:latest` instead of a clear error. Suggest updating the help/description to state that the full registry host is expected, and optionally failing fast when the value contains no `.` or `/` (i.e. a bare ACR name).

.github/workflows/linux-gpu-x64-build.yml
```diff
@@ -102,12 +105,14 @@ jobs:
       - name: Get Docker Image
         run: |
           set -e -x
-          az login --identity --object-id b44538a9-6545-4353-907e-da46df88a2ab
+          az login --identity --object-id 161bbca8-9899-4f61-950c-4d556fbcf830
           az acr login --name onnxruntimebuildcache --subscription 00c06639-6ee4-454e-8058-8d8b1703bd87
+          docker login --username "$GITHUB_ACTOR" --password '${{ secrets.GITHUB_TOKEN }}' ghcr.io
```
tianleiwu, line 110: Security nit: prefer `--password-stdin` over `--password` on the CLI. `docker login ... --password '<token>'` puts the token in the process argument list (visible to other processes on the shared self-hosted runner), and Docker itself warns against it. Since this step runs under `set -x`, the argument is also echoed to the trace. Suggest:

```bash
echo '${{ secrets.GITHUB_TOKEN }}' | docker login --username "$GITHUB_ACTOR" --password-stdin ghcr.io
```

(The single quotes around `${{ secrets.GITHUB_TOKEN }}` are fine as-is: GitHub substitutes `${{ }}` expressions before the shell runs, so quoting does not affect expansion; it just keeps the shell from re-interpreting the token value.)

.github/workflows/linux-cpu-arm64-build.yml
```diff
@@ -73,12 +76,14 @@ jobs:
       - name: Download Docker Image
         run: |
           set -e -x
-          az login --identity --object-id b44538a9-6545-4353-907e-da46df88a2ab
+          az login --identity --object-id 161bbca8-9899-4f61-950c-4d556fbcf830
           az acr login --name onnxruntimebuildcache --subscription 00c06639-6ee4-454e-8058-8d8b1703bd87
+          docker login --username "$GITHUB_ACTOR" --password '${{ secrets.GITHUB_TOKEN }}' ghcr.io
```
tianleiwu, line 81: Same `--password-stdin` suggestion as the CUDA workflow: `docker login ... --password '<token>'` exposes the token in the process argument list on the shared self-hosted runner (and under `set -x`). Prefer:

```bash
echo '${{ secrets.GITHUB_TOKEN }}' | docker login --username "$GITHUB_ACTOR" --password-stdin ghcr.io
```

tianleiwu (commented):
Reviewed the switch from ACR to GHCR for the GitHub-hosted docker build caches. The overall design is coherent and correct: GHCR gets read/write (push cache) via job-scoped `packages: write`, while the ADO pipelines keep pushing to ACR (`onnxruntimebuildcache.azurecr.io`), and all five in-repo `--container-registry` call sites were updated consistently for the new full-hostname semantics. Scoping `permissions: {contents: read, packages: write}` per job is a nice least-privilege touch.

One clarification for the record: the `az acr login` retained in the two GitHub workflows is **not** redundant. The Dockerfiles pull their base images `FROM onnxruntimebuildcache.azurecr.io/...` (e.g. `Dockerfile.manylinux2_28_cuda_12.8` uses `ARG BASEIMAGE=onnxruntimebuildcache.azurecr.io/...`, and the aarch64 CPU Dockerfile has a hard-coded ACR `FROM`), so ACR *pull* access is still required. Removing those `az login`/`az acr login` steps would break the base-image pull. Keeping ACR pull-only (via the read-only managed identity, which the new object-id `161bbca8...` appears to be) is exactly what "Azure resources read-only from GH workflows" means, so this looks intentional and fine.

Minor: the new `.ignore` file is missing a trailing newline and is unrelated to the CI cache change. Worth confirming which tool is meant to consume it (ripgrep/fd read `.ignore` and skip dot-dirs by default, so the negated entries do have an effect there, but the file reads as scope creep for this PR).

Suggestions inline. Nothing blocking.

.github/workflows/linux-gpu-x64-build.yml
sanaa-hamel-microsoft: Proposed fix is broken. Secret is still echoed by `-x`  (expect logging system to censor it as it is a known secret), still visible in `echo`'s arg list to other procs (though briefly, but racing doesn't give security). Also the PW is written to an unencrypted file in HOME, and we assume an attacker has code exec on github workflows anyways.

Note: Single quote is insufficient to correctly escape all values. Only safe method of transfer is via env-vars, which would be even more visible.

baijumeswani (dismissed):

baijumeswani (approved):

tianleiwu (approved):

kunal-vaishnavi (approved):

PR #2266, opened 2026-07-06
Validate eos_token_id against vocab_size to prevent OOB write in ApplyMinLength

jiafatom:
## Summary

Fixes an out-of-bounds write (CWE-787 / CWE-129) reachable from a malicious `genai_config.json`.

`Search_Cpu::ApplyMinLength` suppresses EOS tokens while the sequence is shorter than `min_length` by writing `std::numeric_limits<float>::lowest()` into the per-beam score row at each configured `eos_token_id`:

```cpp
std::span<float> const beam_token_scores = GetScores(i);       // vocab_size elements
for (auto token_id : params_->config.model.eos_token_id)
  beam_token_scores[token_id] = std::numeric_limits<float>::lowest();  // unchecked index
```

`eos_token_id` comes straight from config (`src/config.cpp`) with only a `static_cast<int>` and is never bounded against `vocab_size`. A model whose `eos_token_id >= vocab_size` (or `< 0`) drives the store past the end of the `vocab_size`-sized `GetScores()` subspan — a heap-buffer-overflow write on the first decode step (reached via `Generator::GenerateNextToken` -> `ApplyMinLength`). The same pattern exists in the CUDA backend (`Search_Cuda::ApplyMinLength`).

## Changes

- **`src/generators.cpp`**: reject any `eos_token_id` outside `[0, vocab_size)` in the `Generator::Generator` validator, mirroring the existing `top_k <= vocab_size` check (added in #2224).
- **`src/search.cpp`** and **`src/cuda/search_cuda.cpp`**: defense-in-depth — skip out-of-range ids at the sink in both backends.
- **`test/sampling_tests.cpp`**: add `EosTokenIdExceedsVocabSizeThrowsCpu`, which overlays an out-of-range `eos_token_id` and asserts `OgaGenerator::Create` throws instead of proceeding to the OOB write.

## Testing

New regression test follows the existing `SamplingTests` overlay pattern (`tiny-random-gpt2-fp32` + `vocab_size`/`eos_token_id` overlay). Local build isn't available in my environment; relying on CI to build and run the C++ unit tests.

### DIFF
--- a/src/cuda/search_cuda.cpp
+++ b/src/cuda/search_cuda.cpp
@@ -258,7 +258,8 @@ void Search_Cuda::ApplyMinLength(int min_length) {
     return;
 
   for (auto eos_token_id : params_->config.model.eos_token_id)
-    cuda::LaunchSetScoreProcessor(GetScores().data(), params_->BatchBeamSize(), params_->config.model.vocab_size, eos_token_id, std::numeric_limits<float>::lowest(), GetStream());
+    if (eos_token_id >= 0 && eos_token_id < params_->config.model.vocab_size)
+      cuda::LaunchSetScoreProcessor(GetScores().data(), params_->BatchBeamSize(), params_->config.model.vocab_size, eos_token_id, std::numeric_limits<float>::lowest(), GetStream());
 }
 
 void Search_Cuda::ApplyRepetitionPenalty(float penalty) {
--- a/src/cuda/search_cuda.cpp
+++ b/src/cuda/search_cuda.cpp
@@ -258,8 +258,7 @@ void Search_Cuda::ApplyMinLength(int min_length) {
     return;
 
   for (auto eos_token_id : params_->config.model.eos_token_id)
-    if (eos_token_id >= 0 && eos_token_id < params_->config.model.vocab_size)
-      cuda::LaunchSetScoreProcessor(GetScores().data(), params_->BatchBeamSize(), params_->config.model.vocab_size, eos_token_id, std::numeric_limits<float>::lowest(), GetStream());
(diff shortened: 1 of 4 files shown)


src/search.cpp
```diff
@@ -548,7 +548,8 @@ void Search_Cpu::ApplyMinLength(int min_length) {
   for (int i = 0; i < batch_beam_size; i++) {
     std::span<float> const beam_token_scores = GetScores(i);
     for (auto token_id : params_->config.model.eos_token_id)
-      beam_token_scores[token_id] = std::numeric_limits<float>::lowest();
+      if (token_id >= 0 && token_id < params_->config.model.vocab_size)
```
tianleiwu: Can we validate eos_token_id during loading params_->config.model once?
 
It impacts performance to check in this inference.

tianleiwu: I saw you also changed src/generators.cpp. If eos_token_id failed there, it will not be able to call into search.cpp since generator failed to create.

src/cuda/search_cuda.cpp
```diff
@@ -258,7 +258,8 @@ void Search_Cuda::ApplyMinLength(int min_length) {
     return;
 
   for (auto eos_token_id : params_->config.model.eos_token_id)
-    cuda::LaunchSetScoreProcessor(GetScores().data(), params_->BatchBeamSize(), params_->config.model.vocab_size, eos_token_id, std::numeric_limits<float>::lowest(), GetStream());
+    if (eos_token_id >= 0 && eos_token_id < params_->config.model.vocab_size)
```
tianleiwu: Same here. We shall not check this in inference. Check it in configure loading once.

kunal-vaishnavi (dismissed):

kunal-vaishnavi (approved):

PR #2267, opened 2026-07-06
Drop CWD fallback in custom_ops_library resolution

apsonawane:
custom_ops_library no longer resolves relative paths against the process working directory. EP-library-dir resolution and the platform loader's search path are unchanged. Defense-in-depth only; no functional change for models that specify absolute paths or bare library names that live on the loader search path.


src/models/model.cpp
```diff
@@ -669,6 +669,28 @@ void Model::CreateSessionOptionsFromConfig(const Config::SessionOptions& config_
...
+    // that typically ships alongside the model (including models downloaded from public registries
+    // such as HuggingFace Hub) and the search order below happily picks up a library packaged with
+    // the model, honoring this setting by default is a supply-chain code-execution risk.
+    //
+    // Require an explicit, process-wide opt-in via environment variable before loading anything.
+    // Callers who trust the model can set ORTGENAI_ALLOW_CUSTOM_OPS_LIBRARY=1 to restore the
+    // previous behavior. Absolute paths are subject to the same opt-in requirement because the
+    // attacker also controls the config value.
+    bool custom_ops_library_allowed = false;
+    GetEnv("ORTGENAI_ALLOW_CUSTOM_OPS_LIBRARY", custom_ops_library_allowed);
```
baijumeswani: Rejecting the load of the custom op library outright seems overly harsh and restrictive. We should probably allow loading the library if the library is already on the search path or part of the execution provider library directory. Requiring applications to set the env variable is not desirable.

We probably need to check existing genai_config.json files for all FL models and understand if this is a common field in the genai_config.json. If it is, we need to offer a mitigation/migration path for loading the custom op library.

baijumeswani (approved):

PR #2270, opened 2026-07-06
Fix use-after-free by making GeneratorParams co-own its Model

jiafatom:
## Summary

Fixes a use-after-free (CWE-416) reachable through the public C ABI.

`GeneratorParams` stored only a bare `const Config&` aliasing the model-owned `Config` (`src/generators.h`), relying on the documented-but-unenforced assumption that "the model outlives the GeneratorParams". It did **not** take co-ownership of the model. The following ordinary C API sequence therefore triggers a UAF:

1. `OgaCreateModel` — allocates the `Model` and its `std::unique_ptr<Config> config_`.
2. `OgaCreateGeneratorParams` — binds `config{*model.config_.get()}`, a bare reference; the model's refcount is unchanged.
3. `OgaDestroyModel` — `ExternalRelease` drops the last owner and frees the `Model` and its `Config`.
4. `OgaCreateGenerator(model, params)` — `Generator::Generator` runs `model_{model.shared_from_this()}` on the **freed** model → heap-use-after-free read.

`Generator` already co-owns the model via `std::shared_ptr<const Model> model_`; `GeneratorParams` did not follow the same discipline, leaving the window open.

## Fix

- **`src/generators.h`**: add `std::shared_ptr<const Model> model_owner_` to `GeneratorParams`, declared **before** the `config` reference so it is initialized first.
- **`src/generators.cpp`**: initialize `model_owner_{model.shared_from_this()}` and bind `config{*model_owner_->config_.get()}`. Now the model (and its `Config`) cannot be freed while a params handle is live, and a generator can never be constructed from a freed model. The benchmark-only `GeneratorParams(const Config&)` path leaves `model_owner_` null by design.
- **`test/c_api_tests.cpp`**: add `CreateGeneratorAfterDestroyModel`, reproducing the PoC ordering (destroy model handle, then create generator) and asserting success without a UAF.

## Testing

Local build isn't available in my environment (root-owned build dir / needs the jiafa-dev container); relying on CI to build and run the C++ unit tests, ideally under ASan.

### DIFF
--- a/src/generators.h
+++ b/src/generators.h
@@ -73,7 +73,10 @@ struct GeneratorParams : std::enable_shared_from_this<GeneratorParams>, LeakChec
   GeneratorParams(const Config& config);  // This constructor is only used for internal generator benchmarks
   GeneratorParams(const Model& model);
 
-  const Config& config;                  // The model outlives the GeneratorParams
+  // Co-owns the model so the aliased Config below cannot be freed while this
+  // params object is alive. Null for the benchmark-only Config constructor.
+  std::shared_ptr<const Model> model_owner_;
+  const Config& config;                  // Aliases model-owned Config; kept alive by model_owner_
   Config::Search search{config.search};  // Copy of the search parameters from the config
 
   // Query the params to get the value set for a param
--- a/src/generators.h
+++ b/src/generators.h
@@ -75,8 +75,8 @@ struct GeneratorParams : std::enable_shared_from_this<GeneratorParams>, LeakChec
 
   // Co-owns the model so the aliased Config below cannot be freed while this
   // params object is alive. Null for the benchmark-only Config constructor.
-  std::shared_ptr<const Model> model_owner_;
-  const Config& config;                  // Aliases model-owned Config; kept alive by model_owner_
+  std::shared_ptr<const Model> model_;
+  const Config& config;                  // Aliases model-owned Config; kept alive by model_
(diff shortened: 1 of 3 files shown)


src/generators.h
```diff
@@ -73,7 +73,10 @@ struct GeneratorParams : std::enable_shared_from_this<GeneratorParams>, LeakChec
   GeneratorParams(const Config& config);  // This constructor is only used for internal generator benchmarks
   GeneratorParams(const Model& model);
 
-  const Config& config;                  // The model outlives the GeneratorParams
+  // Co-owns the model so the aliased Config below cannot be freed while this
+  // params object is alive. Null for the benchmark-only Config constructor.
+  std::shared_ptr<const Model> model_owner_;
```
kunal-vaishnavi: Can we rename this to `model_`? Also, we already have access to `config` via `model_->config_.get()` so why do we need to store it now?

jiafatom: Done — renamed `model_owner_` to `model_` (8284fe9b).

On keeping `config`: the other constructor `GeneratorParams(const Config&)` (used for internal generator benchmarks) has no `Model`, so `model_` is null there and `config` cannot be universally derived from `model_->config_.get()`. Keeping `config` as an alias also avoids churning the many `params.config` call sites across the codebase. The new `model_` is purely for lifetime co-ownership so the aliased Config stays valid.

baijumeswani (approved):

PR #2273, opened 2026-07-06
Contain enable_profiling path to model directory (path traversal hardening)

jiafatom:
## Summary

Hardens the `enable_profiling` field of `genai_config.json` against path traversal (CWE-22).

The field was read from the (untrusted) model config and passed **verbatim** to OnnxRuntime's `EnableProfiling()`:

```cpp
// src/models/model.cpp
fs::path profile_file_prefix{config_session_options.enable_profiling.value()};
session_options.EnableProfiling(profile_file_prefix.c_str());  // honored verbatim
```

ORT writes a `<prefix>_<timestamp>.json` profiling file when the session closes. A malicious model package could set e.g. `"enable_profiling": "../../../../etc/cron.d/ort"` (Linux) or a `...\Startup\evil` path (Windows) to write to arbitrary locations.

## Fix

- **`src/models/model.cpp`**: add `ResolveContainedConfigPath`, which rejects absolute paths and any parent-directory (`..`) component, then resolves the profiling prefix relative to the model directory (`config_->config_path`). Applied at the `enable_profiling` sink.
- **`test/model_tests.cpp`**: add `EnableProfilingRejectsPathTraversal` and `EnableProfilingRejectsAbsolutePath`.

The custom `fs::path` in this repo has no canonicalization primitives, so containment is enforced lexically (reject absolute + `..`), which is sufficient to keep the resolved path within the model directory subtree.

## Notes / scope

- The **host-controlled** runtime option (`SetRuntimeOption("enable_profiling", ...)`, `model.cpp` / `State::SetRunOption`) is intentionally left unchanged — that path is not attacker-controlled.
- Behavior change: config-sourced relative profiling paths now resolve relative to the model directory (previously the process CWD). This matches how other config paths (e.g. `custom_ops_library`) resolve and is the contained/expected default.
- As the reporter notes, the co-located `custom_ops_library` field is a higher-value target but reflects the documented model-package trust boundary; this PR scopes to the `enable_profiling` hardening.

## Testing

New regression tests use the existing `OgaConfig::Overlay` pattern on `tiny-random-gpt2-fp32`. Local build isn't available in my environment (root-owned build dir / jiafa-dev container needed); relying on CI to build and run the C++ unit tests.

### DIFF
--- a/src/models/model.cpp
+++ b/src/models/model.cpp
@@ -53,6 +53,33 @@ namespace {
 constexpr const char* kOrtSessionOptionsModelExternalInitializersFileFolderPath =
     "session.model_external_initializers_file_folder_path";
 
+// Validates that a path-like string from the (untrusted) model config stays
+// within the model directory, then resolves it relative to that directory.
+// Rejects absolute paths and any parent-directory ("..") component to prevent
+// path traversal / arbitrary file writes (CWE-22).
+fs::path ResolveContainedConfigPath(const fs::path& model_dir, const std::string& value, const char* field_name) {
+  fs::path candidate{value};
+  if (!candidate.is_relative()) {
+    throw std::runtime_error(std::string(field_name) + " (" + value +
+                             ") must be a relative path contained within the model directory");
+  }
+
+  size_t start = 0;
+  while (start <= value.size()) {
+    size_t sep = value.find_first_of("/\\", start);
+    const std::string component = value.substr(start, sep == std::string::npos ? std::string::npos : sep - start);
+    if (component == "..") {
+      throw std::runtime_error(std::string(field_name) + " (" + value +
+                               ") must not contain parent-directory (\"..\") components");
+    }
+    if (sep == std::string::npos)
+      break;
+    start = sep + 1;
+  }
+
+  return model_dir / value;
+}
+
 }  // namespace
 
 State::State(const GeneratorParams& params, const Model& model)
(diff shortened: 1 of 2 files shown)


src/models/model.cpp
```diff
@@ -53,6 +53,40 @@ namespace {
 constexpr const char* kOrtSessionOptionsModelExternalInitializersFileFolderPath =
     "session.model_external_initializers_file_folder_path";
 
+// Validates that a path-like string from the (untrusted) model config stays
+// within the model directory, then resolves it relative to that directory.
+// Rejects absolute paths and any current/parent-directory (".", "..") component
```
kunal-vaishnavi, line 58: Same question as [here](https://github.com/microsoft/onnxruntime-genai/pull/2222#discussion_r3405695915) in the other PR

jiafatom: Good question. Unlike the custom-ops library path in #2222, `enable_profiling` here is being deliberately hardened against path traversal (CWE-22) per the security intent of this PR: the value comes from the (untrusted) `genai_config.json`, and profiling output is a file the process *writes*. Allowing absolute paths would re-open the arbitrary-file-write vector this change is meant to close, so absolute paths are intentionally rejected and the value is resolved relative to the model directory. Relative paths within the model dir remain fully supported. If a caller genuinely needs an arbitrary profiling location, that should be an explicit, trusted API input rather than something sourced from the model config. Happy to revisit if you feel the containment is too strict for this field.

PR #2274, opened 2026-07-06
Fix UAF in adapter lifecycle: serialize Adapters ops with a mutex

apsonawane:
The Adapter::ref_count_ field was a plain int32_t modified without

synchronization, and Adapters::UnloadAdapter() performed a check-then-erase

on the underlying std::unordered_map without a lock. Concurrent

OgaSetActiveAdapter / OgaUnloadAdapter calls could therefore:

  * race on ref_count_ (data race, undefined behavior, lost updates), and

  * pass the RefCount() > 0 guard in UnloadAdapter while another thread

    was mid-AcquireRef(), destroying the adapter while a reference was

    still held (use-after-free).

Add a mutex to Adapters and take it in Load/Unload/Acquire/ReleaseAdapter.

Since AcquireRef/ReleaseRef/RefCount on Adapter are only reachable through

these Adapters entry points, the single mutex now serializes both map

mutation and per-adapter ref counting, closing the TOCTOU window and the

data race in one place. ReleaseRef also restores the counter before

throwing so a caught exception does not leave it in a negative state.

### DIFF
--- a/src/models/adapters.cpp
+++ b/src/models/adapters.cpp
@@ -10,14 +10,14 @@ Adapter::Adapter(const char* adapter_file_path, Ort::Allocator* allocator)
     : adapter_{OrtLoraAdapter::Create(fs::path(adapter_file_path).c_str(), *allocator)} {}
 
 const OrtLoraAdapter* Adapter::AcquireRef() {
-  // Caller (Adapters::AcquireAdapter) holds Adapters::mutex_, which
-  // serializes all access to ref_count_.
+  // Private; only callable by Adapters (friend), which holds Adapters::mutex_
+  // and therefore serializes all access to ref_count_.
   ref_count_++;
   return adapter_.get();
 }
 
 void Adapter::ReleaseRef() {
-  // Caller (Adapters::ReleaseAdapter) holds Adapters::mutex_.
+  // Private; only callable by Adapters (friend), which holds Adapters::mutex_.
   ref_count_--;
   if (ref_count_ < 0) {
     // Restore invariant so a caller catching the exception doesn't leave the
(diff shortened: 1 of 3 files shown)


src/models/adapters.h
```diff
@@ -13,13 +13,21 @@ struct Adapter {
 
   Adapter(const char* adapter_file_path, Ort::Allocator* allocator);
 
+ private:
```
kunal-vaishnavi, line 16: From Copilot:

Could we make `AcquireRef`, `ReleaseRef`, and `RefCount` private and only expose them to `Adapters` via friendship? Right now the locking invariant relies on convention, which means a future caller could bypass `Adapters::mutex_` and reintroduce the race/UAF.

src/models/adapters.cpp
```diff
@@ -10,25 +10,32 @@ Adapter::Adapter(const char* adapter_file_path, Ort::Allocator* allocator)
     : adapter_{OrtLoraAdapter::Create(fs::path(adapter_file_path).c_str(), *allocator)} {}
 
 const OrtLoraAdapter* Adapter::AcquireRef() {
+  // Private; only callable by Adapters (friend), which holds Adapters::mutex_
```
kunal-vaishnavi, line 13: From Copilot:

Could we also tighten the API contract in the header so `UnloadAdapter()` can’t be bypassed by future ref-count access outside `Adapters`? The mutex only protects the current call paths, so making the ref-count methods private would help enforce the locking invariant and avoid reintroducing the race/UAF later.

src/models/adapters.h
apsonawane: Changes are already done

src/models/adapters.cpp
apsonawane: Changes are already done, looks like copilot is reviewing the merge main commit

kunal-vaishnavi (approved):

PR #2276, opened 2026-07-07
Add multi_rotary_cache_concat_offset for Phi-4 model in TRT-RTX EP 

yen-shi:
## Summary

Add the OGA-side Phi-4 LongRoPE changes required to fix TRT-RTX perf regression. The regression was coming from the If node change in https://github.com/microsoft/onnxruntime-genai/pull/1851.

Phi-4 exports previously used an ONNX `If` node to select short or long rotary caches. That path keeps accuracy across the LongRoPE boundary, but it prevents TRT-RTX from using the fast GQA/CUDA graph path. This change exports a concatenated short + long rotary cache for TRT-RTX and writes `multi_rotary_cache_concat_offset` into the TRT-RTX provider options so the plugin can select the correct cache rows in-kernel.

## Changes

- Treat TRT-RTX like other EPs that do not use the ONNX `If` rotary-cache subgraph.
- Emit concatenated short + long cosine/sine caches for TRT-RTX.
- Add `multi_rotary_cache_concat_offset` to TRT-RTX provider options.
- Keep Phi rewind enabled for TRT-RTX only when the concatenated-cache offset is present, so KV is recomputed with the long LongRoPE factors after crossing the short-cache boundary.

## Companion Changes

This branch requires the matching TRT-RTX EP ABI and TRT-RTX plugin updates for support of `multi_rotary_cache_concat_offset` attribute.


src/python/py/models/builders/base.py
```diff
@@ -1876,55 +1876,21 @@ def make_rotary_embedding_multi_cache(self, **kwargs):
...
-                gather_name = "/model/attn_mask_reformat/attn_mask_subgraph/Gather"
-            else:
-                gather_name = "/model/attn_mask_reformat/attn_mask_subgraph/Gather_2"
-
-            greater_name = f"{basename}/Greater"
-            greater_inputs = [f"{gather_name}/output_0", f"/model/constants/INT64/{self.original_context_length}"]
-            self.make_greater(greater_name, greater_inputs, shape=[])
-
-            # Create split If nodes and return early
-            self.make_split_if_nodes(
```
kunal-vaishnavi, line 1916: Can you make the corresponding changes after removing this `make_split_if_nodes` method? Here are some example locations that require updating.

https://github.com/microsoft/onnxruntime-genai/blob/87877c20c12ad5f20efa43ba5e31ded24ea25e02/src/python/py/models/builders/base.py#L332-L358

https://github.com/microsoft/onnxruntime-genai/blob/87877c20c12ad5f20efa43ba5e31ded24ea25e02/src/python/py/models/builders/expansions/trt_rtx.py#L197-L326

src/python/py/models/builders/base.py
```diff
@@ -1876,55 +1876,21 @@ def make_rotary_embedding_multi_cache(self, **kwargs):
         self.eps_without_if_support = ["dml"]
```
kunal-vaishnavi: Instead of adding an `if` condition, let's add `trt-rtx` in the original list.

src/python/py/models/builders/base.py
yen-shi: Thanks for pointing out, wasn't aware of these, will update.

src/generators.cpp
```diff
@@ -408,15 +432,20 @@ Generator::Generator(const Model& model, const GeneratorParams& params) : model_
...
+                                                   : 0;
+  const bool ep_uses_single_rope_factor = device_type == DeviceType::DML ||
+                                          (device_type == DeviceType::NvTensorRtRtx && trt_rtx_multi_rope_cache_offset == 0);
 
   // Phi3 ROPE factor rewind threshold: 4097 for phi3/phimoe, 8193 for phi3small, 0 otherwise
   // TODO: Extend to support batch size > 1, num beams > 1, and multimodal models
   const auto& model_type = model_->config_->model.type;
   if (params.BatchBeamSize() == 1 && !ep_uses_single_rope_factor) {
-    if (model_type == "phi3" || model_type == "phimoe")
+    if (trt_rtx_multi_rope_cache_offset > 0 && trt_rtx_multi_rope_cache_offset < model_->config_->model.context_length)
```
kunal-vaishnavi: Given that we have the Phi multi-RoPE cases hardcoded inside ORT GenAI, why is this `if` condition necessary to know the threshold values and why do we need to know what the offset value is here from the provider options?

yen-shi: Updated `HasValidTrtRtxMultiRotaryCacheConcatOffset()` to return true or false to avoid propagating the actual threshold.

src/python/py/models/builders/base.py
yen-shi: Sure, done.

src/python/py/models/builders/base.py
yen-shi: Remove these unused functions.

yen-shi:
@kunal-vaishnavi I've addressed the comments, please review again, thanks!

src/generators.cpp
```diff
@@ -62,6 +63,27 @@ static OrtLoggingLevel GetDefaultOrtLoggingLevel() {
   return ort_verbose_logging ? OrtLoggingLevel::ORT_LOGGING_LEVEL_VERBOSE : OrtLoggingLevel::ORT_LOGGING_LEVEL_ERROR;
 }
 
+static bool HasValidTrtRtxMultiRotaryCacheConcatOffset(const Config& config) {
```
kunal-vaishnavi: Can this be moved into the device interface since it's TRT-RTX specific?

yen-shi: Sure, moved this to be under `struct NvTensorRtRtxInterfaceImpl`

yen-shi:
@kunal-vaishnavi I've addressed last comment, please review again, thanks!

src/cuda/interface.cpp
```diff
@@ -186,6 +188,27 @@ struct CudaInterfaceImpl final : CudaInterfaceImplBase {
 
 struct NvTensorRtRtxInterfaceImpl final : CudaInterfaceImplBase {
   DeviceType GetType() const override { return DeviceType::NvTensorRtRtx; }
```
kunal-vaishnavi, line 190: These changes are fine to keep here for now. In a future PR, we should move the TRT-RTX interface to a standalone `src/nvtensorrtrtx/interface.cpp` file.

kunal-vaishnavi (approved):

kunal-vaishnavi:
The Linux CPU arm64 CI is failing to get the Docker image because of a CI change. The fix was merged in earlier today so you can sync with the main branch to get the fix.

src/cuda/interface.cpp
yen-shi: Ok thanks for pointing out!

yen-shi:
@kunal-vaishnavi I've updated to ToT main branch, and looks like the workflows need approval from a maintainer.

PR #2278, opened 2026-07-07
Fix QNN interface assert parameter

nsubaru:


kunal-vaishnavi:
I see the same typo fix in [this PR](https://github.com/microsoft/onnxruntime-genai/pull/2259). Let's get that PR in since it looks ready to me and we can close this one.

PR #2279, opened 2026-07-07
Support both FP16 and FP32 I/O for Nemotron

p3achyjr:
NvTensorRtRtx is much better when running with FP16. We would like to enable this in OGA runtime.

### DIFF
--- a/src/models/model.cpp
+++ b/src/models/model.cpp
@@ -934,14 +934,16 @@ void Cast(OrtValue& input, std::unique_ptr<OrtValue>& output, DeviceInterface& d
   auto input_info = input.GetTensorTypeAndShapeInfo();
   auto shape = input_info->GetShape();
 
-  if (output && shape != output->GetTensorTypeAndShapeInfo()->GetShape())
-    output = nullptr;
+  if (output) {
+    auto output_info = output->GetTensorTypeAndShapeInfo();
+    if (shape != output_info->GetShape() || output_type != output_info->GetElementType())
+      output = nullptr;
+  }
   if (!output)
     output = OrtValue::CreateTensor(device.GetAllocator(), shape, output_type);
 
   auto input_type = input_info->GetElementType();
   auto element_count = input_info->GetElementCount();
-
   if (element_count != output->GetTensorTypeAndShapeInfo()->GetElementCount())
     throw std::runtime_error("Cast: input and output element count mismatch");
 
--- a/src/models/nemotron_speech.cpp
+++ b/src/models/nemotron_speech.cpp
@@ -7,6 +7,7 @@
 #include <vector>
 
 #include "../generators.h"
+#include "../search.h"
 #include "nemo_mel_spectrogram.h"
 #include "nemotron_speech.h"
 
@@ -22,7 +23,47 @@ DeviceInterface& DeviceFor(const OrtValue& v, DeviceInterface& model_device) {
   const bool on_cpu = v.GetTensorMemoryInfo().GetDeviceType() == OrtMemoryInfoDeviceType_CPU;
   return on_cpu ? *GetDeviceInterface(DeviceType::CPU) : model_device;
 }
+
+template <typename T, typename Convert>
+int NemotronArgMaxImpl(const OrtValue& logits_value, int blank_id, float blank_penalty, Convert convert) {
+  const auto count = logits_value.GetTensorTypeAndShapeInfo()->GetElementCount();
+  auto logits = std::span<const T>(logits_value.GetTensorData<T>(), count);
+  return ArgMax<T>(logits, [blank_id, blank_penalty, convert](T value, size_t index) {
+    return convert(value) - (static_cast<int>(index) == blank_id ? blank_penalty : 0.0f);
+  });
+}
 }  // namespace
+
+int NemotronArgMax(const OrtValue& logits, int blank_id, float blank_penalty) {
+  const auto type = logits.GetTensorTypeAndShapeInfo()->GetElementType();
+  if (type == ONNX_TENSOR_ELEMENT_DATA_TYPE_FLOAT) {
+    return NemotronArgMaxImpl<float>(logits, blank_id, blank_penalty,
+                                     [](float value) { return value; });
+  }
+  if (type == ONNX_TENSOR_ELEMENT_DATA_TYPE_FLOAT16) {
+    return NemotronArgMaxImpl<Ort::Float16_t>(logits, blank_id, blank_penalty,
+                                              [](Ort::Float16_t value) {
+                                                return ToFloat32(value);
+                                              });
+  }
+  throw std::runtime_error("NemotronArgMax only supports float32 or float16 logits");
+}
+
+ONNXTensorElementDataType ValidateNemotronFloatType(std::span<const ONNXTensorElementDataType> types) {
+  if (types.empty())
+    throw std::runtime_error("Nemotron ASR floating-point type list must not be empty");
+
+  const auto type = types[0];
+  if (type != ONNX_TENSOR_ELEMENT_DATA_TYPE_FLOAT &&
+      type != ONNX_TENSOR_ELEMENT_DATA_TYPE_FLOAT16)
+    throw std::runtime_error("Nemotron ASR only supports homogeneous float32 or float16 I/O");
+
+  for (auto candidate : types) {
+    if (candidate != type)
+      throw std::runtime_error("Nemotron ASR requires homogeneous floating-point I/O");
+  }
+  return type;
+}
 
 void NemotronConfig::PopulateFromConfig(const Config& config) {
   const auto& enc = config.model.encoder;
@@ -176,6 +217,24 @@ NemotronSpeechModel::NemotronSpeechModel(std::unique_ptr<Config> config, OrtEnv&
   session_info_.Add(*session_encoder_);
   session_info_.Add(*session_decoder_);
   session_info_.Add(*session_joiner_);
+
+  auto float_types = std::array<ONNXTensorElementDataType, 14>{
+      session_info_.GetInputDataType(nemotron_config_.enc_in_audio),
+      session_info_.GetInputDataType(nemotron_config_.enc_in_cache_channel),
+      session_info_.GetInputDataType(nemotron_config_.enc_in_cache_time),
+      session_info_.GetOutputDataType(nemotron_config_.enc_out_encoded),
+      session_info_.GetOutputDataType(nemotron_config_.enc_out_cache_channel),
+      session_info_.GetOutputDataType(nemotron_config_.enc_out_cache_time),
+      session_info_.GetInputDataType(nemotron_config_.dec_in_lstm_hidden),
+      session_info_.GetInputDataType(nemotron_config_.dec_in_lstm_cell),
+      session_info_.GetOutputDataType(nemotron_config_.dec_out_outputs),
+      session_info_.GetOutputDataType(nemotron_config_.dec_out_lstm_hidden),
+      session_info_.GetOutputDataType(nemotron_config_.dec_out_lstm_cell),
+      session_info_.GetInputDataType(nemotron_config_.join_in_encoder),
+      session_info_.GetInputDataType(nemotron_config_.join_in_decoder),
+      session_info_.GetOutputDataType(nemotron_config_.join_out_logits),
+  };
+  float_type_ = ValidateNemotronFloatType(float_types);
 }
 
 std::unique_ptr<State> NemotronSpeechModel::CreateState(DeviceSpan<int32_t> /*sequence_lengths*/,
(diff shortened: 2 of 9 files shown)


src/models/model.cpp
```diff
@@ -903,33 +909,48 @@ std::shared_ptr<GeneratorParams> CreateGeneratorParams(const Config& config) {
...
-    auto output_span = ByteWrapTensor(device, *output);
+    auto output_span = ByteWrapTensor(device, output);
     input_data = input_span.CopyDeviceToCpu().data();
     output_data = output_span.CopyDeviceToCpu().data();
     GetDeviceInterface(DeviceType::CPU)->Cast(input_data, output_data, input_type, output_type, element_count);
     output_span.CopyCpuToDevice();
   }
 }
 
+void Cast(OrtValue& input, std::unique_ptr<OrtValue>& output, DeviceInterface& device, ONNXTensorElementDataType output_type) {
```
kunal-vaishnavi: This seems inefficient to have a `Cast` and a `CastInto` method. Let's keep the existing `Cast` method and merge any new changes into it.

src/models/nemotron_speech.cpp
```diff
@@ -388,10 +388,23 @@ NemotronSpeechState::NemotronSpeechState(const NemotronSpeechModel& model,
...
   auto enc_out_type = model_.session_info_.GetOutputDataType(nemotron_config_.enc_out_encoded);
+  auto joiner_encoder_type = model_.session_info_.GetInputDataType(nemotron_config_.join_in_encoder);
   auto frame_shape = std::array<int64_t, 3>{1, 1, nemotron_config_.hidden_dim};
   encoder_frame_ = OrtValue::CreateTensor(model_.allocator_cpu_, frame_shape, enc_out_type);
+  if (enc_out_type != joiner_encoder_type)
+    joiner_encoder_frame_ = OrtValue::CreateTensor(model_.allocator_cpu_, frame_shape, joiner_encoder_type);
+
+  auto decoder_output_type = model_.session_info_.GetOutputDataType(nemotron_config_.dec_out_outputs);
+  auto joiner_decoder_type = model_.session_info_.GetInputDataType(nemotron_config_.join_in_decoder);
+  if (decoder_output_type != ONNX_TENSOR_ELEMENT_DATA_TYPE_FLOAT ||
```
kunal-vaishnavi: Can we always use the dtype from the session info when creating an input or output tensor rather than throwing errors? We can leave it up to the user to pick whether the dtype should be FP16 or FP32 in their ONNX models.

src/models/nemotron_speech.cpp
```diff
@@ -536,6 +540,12 @@ void NemotronSpeechState::StepToken() {
     // Copy current encoder frame
     auto src_frame = enc_span.subspan(static_cast<size_t>(time_step_) * frame_bytes, frame_bytes);
     frame_span.CopyFrom(src_frame);
+    OrtValue* encoder_frame = encoder_frame_.get();
+    std::unique_ptr<OrtValue> encoder_frame_cast;
+    if (enc_type != joiner_encoder_type) {
+      Cast(*encoder_frame_, encoder_frame_cast, cpu_dev, joiner_encoder_type);
+      encoder_frame = encoder_frame_cast.get();
+    }
```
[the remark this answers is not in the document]
kunal-vaishnavi: This looks like a valid issue. Casting per-frame can add overhead over time. If you are using an FP16 encoder, can you use an FP16 decoder and FP16 joiner then to avoid the need for casting?

src/models/nemotron_streaming_processor.cpp
```diff
@@ -2,13 +2,71 @@
...
 namespace Generators {
 
+namespace detail {
+namespace {
+
+template <typename T>
+T FromFloat(float value);
+
+template <>
+float FromFloat<float>(float value) {
```
kunal-vaishnavi: There already exists casting methods between different dtypes. Let's re-use those instead of introducing new ones. If you need to introduce a new casting method, let's add it in [`utils.cpp`](https://github.com/microsoft/onnxruntime-genai/blob/main/src/models/utils.cpp).

src/models/nemotron_streaming_processor.cpp
```diff
@@ -2,13 +2,71 @@
...
+  return value;
+}
+
+template <>
+Ort::Float16_t FromFloat<Ort::Float16_t>(float value) {
+  return Ort::Float16_t{FastFloat32ToFloat16(value)};
+}
+
+template <typename T>
+void PopulateMelTensorImpl(T* output, std::span<const float> cache,
```
kunal-vaishnavi, line 13: These are useful helper functions! Can we remove the two namespaces, `detail` and anonymous, that they're in? The `Generators` namespace should be sufficient.

src/models/nemotron_streaming_processor.cpp
```diff
@@ -97,47 +155,25 @@ std::unique_ptr<OrtValue> NemotronStreamingProcessor::BuildMelTensor(const float
...
-                mel_pre_encode_cache_.data(),
-                (cache_size - first_run) * num_mels * sizeof(float));
-  }
+  detail::PopulateMelTensor(*processed_signal, mel_pre_encode_cache_, cache_pos_,
+                            mel_data, num_frames, num_mels);
 
-  // Transpose mel from [num_mels, num_frames] directly into output tensor after cache
-  float* out_ptr = signal_data + cache_size * num_mels;
-  for (int t = 0; t < num_frames; ++t) {
+  // Update the FP32 ring buffer from the frequency-major mel output.
```
kunal-vaishnavi: Can we move the ring buffer updating logic into a helper function as well?

src/models/nemotron_streaming_processor.cpp
```diff
@@ -97,47 +155,25 @@ std::unique_ptr<OrtValue> NemotronStreamingProcessor::BuildMelTensor(const float
...
+  int frames_to_cache = std::min(num_frames, cache_size);
+  int first_frame_to_cache = num_frames - frames_to_cache;
+  for (int t = 0; t < frames_to_cache; ++t) {
+    int destination_t = (cache_pos_ + t) % cache_size;
```
kunal-vaishnavi: Can we change the suffix here from `_t` to something else so that the C++ syntax highlighter doesn't treat `destination_t` as a type name?

src/models/nemotron_speech.cpp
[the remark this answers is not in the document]
p3achyjr: hey kunal, I've updated the MR to support either all FP16 I/O, or all FP32 I/O. can you PTAL again?

src/models/nemotron_speech.cpp
p3achyjr: updated MR

src/models/nemotron_streaming_processor.cpp
p3achyjr: updated

src/models/nemotron_speech.cpp
```diff
@@ -22,8 +22,58 @@ DeviceInterface& DeviceFor(const OrtValue& v, DeviceInterface& model_device) {
   const bool on_cpu = v.GetTensorMemoryInfo().GetDeviceType() == OrtMemoryInfoDeviceType_CPU;
   return on_cpu ? *GetDeviceInterface(DeviceType::CPU) : model_device;
 }
+
+template <typename T, typename Convert>
+int NemotronArgMaxImpl(const OrtValue& logits_value, int blank_id, float blank_penalty, Convert convert) {
```
kunal-vaishnavi, line 28: Let's find a way to re-use existing logic for argmax and other search + sampling methods implemented in this file rather than re-implementing all of them from scratch.

p3achyjr: admittedly this MR is done by codex, and by codex's suggestion it is hard to reuse existing implementations because they 1) assume FP32 and 2) are tied to implementation-specific state. I've asked codex to clean up the implementation but there is still a new, stateless argmax implementation.

src/models/nemotron_streaming_processor.cpp
```diff
@@ -2,13 +2,68 @@
...
+    PopulateMelTensorImpl(output.GetTensorMutableData<float>(), cache, cache_pos,
+                          mel, num_frames, num_mels, [](float value) { return value; });
+  } else if (output_type == ONNX_TENSOR_ELEMENT_DATA_TYPE_FLOAT16) {
+    PopulateMelTensorImpl(output.GetTensorMutableData<Ort::Float16_t>(), cache, cache_pos,
+                          mel, num_frames, num_mels, [](float value) {
+                            return Ort::Float16_t{FastFloat32ToFloat16(value)};
+                          });
+  } else {
+    throw std::runtime_error("PopulateMelTensor: output must be float32 or float16");
+  }
```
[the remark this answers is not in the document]
p3achyjr: changed

src/models/model.cpp
p3achyjr: updated

src/models/nemotron_streaming_processor.cpp
p3achyjr: updated

src/models/nemotron_streaming_processor.cpp
p3achyjr: believe this is updated

src/models/nemotron_streaming_processor.cpp
p3achyjr: updated

yen-shi:
Hi @kunal-vaishnavi,
Can you review this PR again when you get a chance? Thanks!

src/models/nemotron_speech.cpp
kunal-vaishnavi: This is fine for now but we should clean this up in a future PR.

kunal-vaishnavi (dismissed):
Approving for now but we should fix [this issue](https://github.com/microsoft/onnxruntime-genai/pull/2279#discussion_r3573830987) in the future. Please sync with the main branch to get the latest CI changes.

yen-shi:
Hi @p3achyjr,
I tried running this PR with olive-recipes. To run with TRT-RTX EP, looks like I need to apply this patch so nemo speech script can run, can you check if this makes sense and applies to this PR? Thanks!


Verified with
```
python ~/oga_pr2279/examples/python/nemotron_speech.py \
  --model_path /home/nvidia/spark_bench/models/multilingual_fp16_trtrtx \
  --audio_file /home/nvidia/asr_ex2/default_ipa.wav \
  -e NvTensorRTRTXExecutionProvider \
  --ep_path "$LIBS/ep/libonnxruntime_providers_nv_tensorrt_rtx.so" --language en
```


Git patch:
```
diff --git a/examples/python/nemotron_speech.py b/examples/python/nemotron_speech.py
index 0bcb2dc1..ccec71b5 100644
--- a/examples/python/nemotron_speech.py
+++ b/examples/python/nemotron_speech.py
@@ -8,7 +8,7 @@ import sys
 import time
 import numpy as np
-from common import get_config
+from common import get_config, get_ep_args, register_ep
 
 # Maps short language codes / locale tags to (lang_id, human-readable name).
 # Restricted to the languages officially supported by
@@ -130,12 +130,17 @@ def decode_tokens(generator, tokenizer_stream):
     return text
 
 
-def simulate_microphone(model_path, audio_path, execution_provider, use_vad=None, language=None):
+def simulate_microphone(model_path, audio_path, execution_provider, use_vad=None, language=None,
+                        ep_path="", use_winml=False):
     """Stream audio through Generator + StreamingProcessor API."""
     sample_rate, chunk_samples = load_config(model_path)
     audio = load_audio(audio_path, sample_rate)
     duration = len(audio) / sample_rate
 
+    # Register an external plugin EP (e.g. the NvTensorRtRtx plugin) when --ep_path is given.
+    # --execution_provider must be the ORT-expected name (e.g. "NvTensorRTRTXExecutionProvider");
+    # get_config normalizes it to the genai canonical provider name when selecting it.
+    register_ep(execution_provider, ep_path, use_winml)
     config = get_config(model_path, execution_provider, None)
     selected_lang = None
     if language is not None:
@@ -217,9 +222,7 @@ def main():
                 "Pass 'auto' to let the model detect the language. Supported codes:\n" \
                 + "\n".join(f"  {code:<7} {name}" for code, (_, name) in sorted(LANG_TO_ID.items()))
     parser.add_argument("--language", "-l", type=str, default=None, help=lang_help)
-    parser.add_argument("-e", "--execution_provider", type=str, required=False, default="follow_config",
-                        choices=["cpu", "cuda", "dml", "follow_config"],
-                        help="Execution provider to run with. Defaults to follow_config.")
+    get_ep_args(parser)
     args = parser.parse_args()
     if not os.path.exists(args.audio_file):
         print(f"Error: {args.audio_file} not found")
@@ -228,7 +231,8 @@ def main():
     if args.use_vad is not None:
         use_vad_override = args.use_vad == "true"
     simulate_microphone(args.model_path, args.audio_file, args.execution_provider,
-                        use_vad=use_vad_override, language=args.language)
+                        use_vad=use_vad_override, language=args.language,
+                        ep_path=args.ep_path, use_winml=args.use_winml)
 
 
 if __name__ == "__main__":

```

p3achyjr:
@yen-shi cannot reply directly to your comment, but can confirm. Updated the MR.

p3achyjr:
@kunal-vaishnavi the failures all look like CI flakes, can we rerun?

kunal-vaishnavi (dismissed):

p3achyjr:
@kunal-vaishnavi sorry I encountered merge conflicts, could you PTAL again?

baijumeswani (approved):

PR #2282, opened 2026-07-07
Inherit top-level session options in pipeline model components

qti-mattsinc:
### Description
* Previously, for decoder pipelines, the top-level session options were ignored for any component model which already specified its own session options, even when those options did not conflict with the top-level ones.
* To fix this, when initializing pipeline component session options, start by cloning the top-level options as a base for each pipeline component's options.

### Motivation and Context
* QNN uses pipeline models to run LLMs. APIs like `OgaConfigSetDecoderProviderOptionsHardwareDeviceType` only apply to the top-level session options, which means that e.g. top-level device filtering options can get ignored if any pipeline component has its own session options with unrelated entries (like "enable_profiling") and leads to the model not actually running with the desired device.


src/models/model.cpp
```diff
@@ -491,6 +492,41 @@ void EnsureDeviceOrtInit(DeviceInterface& device, const Config& config) {
...
+      child_provider_option_names.clear();
+      for (const auto& opt : child_provider_option_it->options) {
+        child_provider_option_names.insert(opt.first);
+      }
+
+      for (const auto& parent_opt : parent_provider_option.options) {
+        if (child_provider_option_names.find(parent_opt.first) == child_provider_option_names.end()) {
+          child_provider_option_it->options.emplace_back(parent_opt);
+        }
+      }
```
qti-mattsinc, line 566: Alternatively, we could define the semantics to be that if a pipeline model has its own `"provider_options"`, then it's taken as-is and not combined with the top-level options.


So, for example:
```json5
      {
        "model": {
          // ...
          "decoder": {
            "session_options": {
              "provider_options": [ { "qnn": { "foo": "bar" } } ]
            },
            // ...
            "pipeline": [
                "m1": {
                  // ...
                  "session_options": {
                    "provider_options": [ { "qnn": { "baz": "qux" } } ]
                  },
                  // ...
                }
              }
            ]
          },
          // ...
        },
        // ...
      }
```

As of the current state of this PR, model m1 will end up with ` "provider_options": [ { "qnn": { "foo": "bar", "baz": "qux" } } ]`, but it would also be reasonable to expect it to be ` "provider_options": [ { "qnn": { "baz": "qux" } } ]`.

src/models/model.cpp
```diff
@@ -500,6 +501,41 @@ void EnsureDeviceOrtInit(DeviceInterface& device, const Config& config) {
...
+
+      // Merge parent EP provider options
+      child_provider_option_names.clear();
+      for (const auto& opt : child_provider_option_it->options) {
+        child_provider_option_names.insert(opt.first);
+      }
+
+      for (const auto& parent_opt : parent_provider_option.options) {
+        if (child_provider_option_names.find(parent_opt.first) == child_provider_option_names.end()) {
+          child_provider_option_it->options.emplace_back(parent_opt);
```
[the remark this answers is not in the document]
qti-mattsinc: Done

src/models/model.cpp
```diff
@@ -766,23 +808,33 @@ void Model::CreateSessionOptionsFromConfig(const Config::SessionOptions& config_
...
   for (auto& pipeline_model : config_->model.decoder.pipeline) {
     if (pipeline_model.session_options.has_value()) {
-      auto emplaced = pipeline_session_options_.emplace(pipeline_model.model_id, OrtSessionOptions::Create());
-      CreateSessionOptionsFromConfig(*pipeline_model.session_options, *emplaced.first->second, false);
+      // Update config ProviderOptions for pipeline model with values from top-level options
+      InheritParentProviderOptions(config_->model.decoder.session_options.provider_options,
+                                   pipeline_model.session_options->provider_options);
+
+      // Clone the main OrtSessionOptions and then overlay any options explicitly set for this pipeline model
+      auto emplaced = pipeline_session_options_.emplace(pipeline_model.model_id, session_options_->Clone());
```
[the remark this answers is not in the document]
qti-mattsinc: The current pipeline model test is for cuda, and I don't see any QNN coverage under `tests/` currently. Testing QNN would be nice, but would require including a model exported for QNN to `tests/models/` (not sure what the policy is for that, esp. with regard to size limitations). I would be interested in hearing what an OGA maintainer thinks, because I do think QNN coverage would be valuable.

qti-mattsinc:
@kunal-vaishnavi / @baijumeswani, PTAL

johnpaultaken:
Hi @kunal-vaishnavi  @baijumeswani   we are looking forward to your input on this PR. Basically, are the session options in genai_config meant to inherit or override ? The LLM-profiling EpCertification test case from MSFT is assuming a behavior and failing due to the current genai behavior on the Qualcomm EP with GPU backend. This PR is to address that issue. We need this fix before the 10D MSFT certification that starts its process within a week or two. As IHVs we would appreciate cooperation from the MSFT side.

kunal-vaishnavi:
> Hi kunal-vaishnavi Baiju Meswani (@baijumeswani) we are looking forward to your input on this PR. Basically, are the session options in genai_config meant to inherit or override ? The LLM-profiling EpCertification test case from MSFT is assuming a behavior and failing due to the current genai behavior on the Qualcomm EP with GPU backend. This PR is to address that issue. We need this fix before the 10D MSFT certification that starts its process within a week or two. As IHVs we would appreciate cooperation from the MSFT side.

Options (session, provider, run, etc) are scoped to the component that they are defined in. If a component does not define any options, the default is to inherit from the decoder component's definitions.

https://github.com/microsoft/onnxruntime-genai/blob/11233b81fd333b64e5c8455728d8745a360356f4/src/models/model.cpp#L828-L843

Settings in the GenAI config override any defaults inside ORT GenAI. Settings in code override any defaults in both the GenAI config and inside ORT GenAI.

baijumeswani:
> QNN uses pipeline models to run LLMs. APIs like OgaConfigSetDecoderProviderOptionsHardwareDeviceType only apply to the top-level session options, which means that e.g. top-level device filtering options can get ignored if any pipeline component has its own session options with unrelated entries (like "enable_profiling") and leads to the model not actually running with the desired device.

Could the pipeline components in the genai_config.json not be updated to use the correct session options for that device? As Kunal mentioned, the top level (at the decoder level) session options are the fallback session options. Those get used if a pipeline component does not define its own session options. Conflating the two might result in unexpected outcomes because the individual pipeline components then lose their ability to be authoritative. I may be misunderstanding the motivation here.

qti-mattsinc:
> Could the pipeline components in the genai_config.json not be updated to use the correct session options for that device? 

`OgaConfigSetDecoderProviderOptionsHardwareDeviceType` and friends only apply to the top-level session options. So, if we have a pipeline model where one component has some unrelated session option, then the app can't change actually the device type used by that component with that API function. We could change `OgaConfigSetDecoderProviderOptionsHardwareDeviceType` and similar functions to apply the new device filtering option to all members of the pipeline, but that doesn't handle the case where an app decides to overlay some session options manually.

We are actually seeing this behavior in a MSFT test app (EpCert, not sure if you are familiar with it), where it decides to write the "enable_profiling" option to all members of the pipeline. This causes the top-level options to be completely ignored.

> As Kunal mentioned, the top level (at the decoder level) session options are the fallback session options. Those get used if a pipeline component does not define its own session options.

If you have a pipeline model where none of the components define their own session options, then they all use the top-level one. But if you add e.g. just "log_severity_level" to one of the components, then all the top-level options are now ignored for that component. This seems strange to me? And given that a MSFT test app is doing the same thing, it's maybe not obvious that this is the expected behavior.

> Conflating the two might result in unexpected outcomes because the individual pipeline components then lose their ability to be authoritative. I may be misunderstanding the motivation here.

Maybe an option could be added to signal that a pipeline component should completely override the top-level ones (current behavior), so that they retain the ability to be authoritative? Then, when exporting our QNN GPU models, we can turn this option off in genai_config.json (because our exports do not need/want this behavior). But if you want to stick with the current design, then I think we at least need to:

1. Update `OgaConfigSetDecoderProviderOptionsHardwareDeviceType` and similar fns to apply to pipeline components.
2. Document the expectations around the fallback so that app knows what do to if manually overlays some session options in the config.

baijumeswani:
Ok, thanks for the context. It seems like if ort-genai exposed a way for applications to change the pipeline components dynamically, that would address the issues?

Do you think ort-genai should expose fine-grained control over the pipeline components? Or is it better to return a json string, let the application manipulate it the way they want, pass it back in to the oga::Config, and then load model based on updated config from the application?

qti-mattsinc:
`OgaConfigOverlay` allows the app to manipulate the config json, right? Or, it can always just load genai_config.json from disk, modify it, and write it back before loading the model. So, I don't think we'd necessarily need a new API for changing pipeline components.

I guess I'm wondering whether you think there is a benefit to allowing the components to inherit the session options as in this PR instead of the current all-or-nothing fallback. A QNN GPU exported model looks like this:

```json5
{
    "model": {
        "decoder": {
            "session_options": { "provider_options": [ { "qnn": { "backend_type": "gpu", "enable_dx12_shared_memory_allocator": "1" } } ] },
            // ...
            "pipeline": [
                {
                    "context": {
                        "filename": "context.onnx",
                        // ... (no session_options here)
                    },
                    "iterator": {
                        "filename": "iterator.onnx",
                        // ... (no session_options here)
                    }
                }
            ]
        },
        // ...
    },
    // ...
}
```

No matter what APIs are provided, if the app decides to e.g. add "enable_profiling" to the context component, then suddenly it no longer runs on QNN GPU because it has stopped using the top-level ones. So either the app must know that it shouldn't do that (which requires adding some documentation and the app developer reading it), or we can just let the component inherit the top-level options (we can guard this behind an option like "enable_session_option_inheritance": true, which we would start putting in our QNN GPU exports). I would prefer the latter. But if you disagree with that approach and want to treat this as an app-issue rather than an ort-genai issue, then I'm OK with that, and in that case I could recommend an app-level fix for situations where we have encountered this problem. What do you think?

kunal-vaishnavi:
The inheritance of options was mainly implemented for when nothing is provided in a standalone section. Given that the pipeline section is within the decoder section, I think it is fine to inherit the top-level options from the decoder section for the pipeline sections without even a guard flag. Only QNN models use the decoder-pipeline architecture.

baijumeswani:
My slight preference is that pipeline components define an explicit opt-in to inherit top level session options. Some setting like `inherit_session_options` at the `PipelineModel` level. That way, both scenarios can work.

So, then QNN models can define inherit_session_options at the pipeline component level and then the top level session options can be overlaid on top of the pipeline component level session options.

Why I think blanker inheritance may not be a good idea. Some models run across multiple execution providers. For example, the default at the top level could be CPU, while the individual components could be CUDA. Session options for CPU should not be directly applicable to CUDA as an example.

qti-mattsinc:
Thank you both for the feedback. I've added an "inherit_session_options" key for the config following Baiju's suggestion. We will plan to update our Olive recipes to enable this option for our model exports when this change lands. Let me know if there are any other changes you'd like to see.

src/models/whisper.cpp
```diff
@@ -8,7 +8,9 @@ namespace Generators {
 WhisperModel::WhisperModel(std::unique_ptr<Config> config, OrtEnv& ort_env)
     : Model{std::move(config)} {
   encoder_session_options_ = OrtSessionOptions::Create();
-  CreateSessionOptionsFromConfig(config_->model.encoder.session_options.has_value() ? config_->model.encoder.session_options.value() : config_->model.decoder.session_options, *encoder_session_options_, true);
+  const auto& encoder_session_options = config_->model.encoder.session_options.has_value() ? config_->model.encoder.session_options.value() : config_->model.decoder.session_options;
+  CreateSessionOptionsFromConfig(encoder_session_options, *encoder_session_options_);
+  AppendSessionProviders(encoder_session_options, *encoder_session_options_, true);
```
baijumeswani: Could you explain why these changes are necessary? None of these use the decode-pipeline components.

src/models/model.h
```diff
@@ -197,12 +197,19 @@ struct Model : std::enable_shared_from_this<Model>, LeakChecked<Model>, External
...
-                                      bool is_primary_session_options,
-                                      bool disable_graph_capture = false);
+                                      bool cloned_from_parent = false) const;
+
+  /// Append the execution providers from the config onto the session options
+  /// and update p_device_.
+  void AppendSessionProviders(const Config::SessionOptions& config_session_options,
+                              OrtSessionOptions& session_options,
+                              bool is_primary_session_options,
+                              bool disable_graph_capture = false);
```
baijumeswani: Can you explain why this split was necessary?

qti-mattsinc: Because I'm using `OrtSessionOptions::Clone()` for the inherit path ([src/models/model.cpp:815-817](https://github.com/microsoft/onnxruntime-genai/pull/2282/changes#diff-3da792c72182817b1bdb08f4ec7eb7687fa19d9219aea8ccfdb310a1061a5c7bR815-R817)), I want to delay calling `AppendExecutionProvider_V2` on the base session options until after it has been cloned. Otherwise, we would be appending EPs twice for the components that inherit, which seems undesirable.

src/models/whisper.cpp
qti-mattsinc: It's necessary because I separated the `SetProviderSessionOptions` step out of `Model::CreateSessionOptionsFromConfig` into it's own function (`Model::AppendSessionProviders`). As for why that is needed, see my reply to your other comment: https://github.com/microsoft/onnxruntime-genai/pull/2282#discussion_r3806158353.

src/models/model.h
baijumeswani: 2 phase split api is easy to misuse. callers must remember to call both. could we instead create a single-entry point and hide the implementation details from the invoker?

src/models/model.cpp
```diff
@@ -504,6 +505,41 @@ void EnsureDeviceOrtInit(DeviceInterface& device, const Config& config) {
...
+        child_provider_options.end(),
+        [&parent_provider_option](const Config::ProviderOptions& po) { return po.name == parent_provider_option.name; });
+    if (child_provider_option_it == child_provider_options.end()) {
+      // Child has no provider option for this provider, so just copy from parent.
+      child_provider_options.emplace_back(parent_provider_option);
+    } else {
+      // Use device filtering options from parent if not already set
+      if (!child_provider_option_it->device_filtering_options.has_value()) {
+        child_provider_option_it->device_filtering_options = parent_provider_option.device_filtering_options;
+      }
```
baijumeswani, line 554: If both child and parent define device filtering options, then the child doesn't merge the parent's device filtering options. Is that intentional?

src/models/model.cpp
```diff
@@ -504,6 +505,41 @@ void EnsureDeviceOrtInit(DeviceInterface& device, const Config& config) {
   device.InitOrt(*Ort::api, *allocator.allocator_);
 }
 
+// Update provider options using values from a parent if they are not already specified in the child.
+// Used for pipeline models that opt in to inheritance to ensure that top-level provider options are
+// communicated into the options for those component models.
+static void InheritParentProviderOptions(const std::vector<Config::ProviderOptions>& parent_provider_options,
+                                         std::vector<Config::ProviderOptions>& child_provider_options) {
```
baijumeswani, line 540: Avoid mutating the original config. The config should be thought of as const. Instead make a copy in case the intention is to use the config to create the session option.

src/models/model.cpp
qti-mattsinc: Yes, this was intentional. I think it's better for the device filtering options to be taken atomically. Because otherwise you could have a situation where e.g. you specify device_type CPU but inherit the DML vendor ID, which is probably not what you wanted.

I was also wondering the same thing about the provider_options (https://github.com/microsoft/onnxruntime-genai/pull/2282#discussion_r3540020734). Right now they are merged with the parent, but I could see the argument for them being all-or-nothing too.

src/models/model.cpp
qti-mattsinc: Done. Switched to copying the config

src/models/model.h
qti-mattsinc: Switched to one entry point with an optional arg that can be used to skip the append EPs step. That way, the only callsite that needs to change is in model.cpp and the others (in marian.cpp, multi_modal.cpp) don't need to know about it. Is this what you had in mind?

baijumeswani (approved):

PR #2284, opened 2026-07-09
LFM2: add shared-buffer KV cache support to LFM2Cache

SanjayAMD:
### Summary
Adds a shared (pre-allocated, fixed-size) KV-cache mode to `LFM2Cache`, mirroring
`DefaultKeyValueCache`. This lets LFM2 use `past_present_share_buffer`, where the
past and present KV alias a single fixed-size buffer — a layout some execution
providers/kernels require.

### Problem
LFM2 is a hybrid (attention + conv) model and uses a dedicated `LFM2Cache`. That
class only implemented a **dynamic / growing** KV cache: it allocated `seq=0`, fed
an empty past, and reallocated/swapped the present tensors every step. In addition,
the LFM2 model builder hard-forced `past_present_share_buffer = False` in the
generated `genai_config.json`, so the shared-buffer path could never be selected —
even for backends that only support a fixed, shared past/present buffer.

### Changes
- Add `kv_share_buffer_` to `LFM2Cache`, gated the same way as
  `DefaultKeyValueCache` (via `IsPastPresentShareBufferEnabled` and
  `DetectAndConfigureFixedKvShape`).
- In shared-buffer mode: pre-allocate the KV buffer once (sized to the model's fixed
  kv `seq_len` when the graph pins it, otherwise to `search.max_length`), bind
  past == present (the same `OrtValue`), and make `Update()` a no-op for KV (no
  grow/swap).
- Preserve the existing dynamic/growing path unchanged when share-buffer is off.
- Stop overriding `past_present_share_buffer` to `False` in the LFM2 model builder,
  so the runtime can enable the shared-KV path when appropriate.

### Testing
- Exported **LFM2.5-350M** and **LFM2.5-1.2B-Instruct** via Olive using the OGA
  model builder; verified `past_present_share_buffer: true` in the generated
  `genai_config.json`.
- Ran inference across a range of prompt lengths; generation is stable and outputs
  remain coherent.

### DIFF
--- a/src/models/kv_cache.cpp
+++ b/src/models/kv_cache.cpp
@@ -1,5 +1,6 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
+// Modifications Copyright(C) 2026 Advanced Micro Devices, Inc. All rights reserved.
 
 #include "../generators.h"
 #include "model.h"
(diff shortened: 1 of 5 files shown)


VishalX:
@baijumeswani / @kunal-vaishnavi could you pls review this?

src/python/py/models/builders/lfm2.py
```diff
@@ -275,8 +276,9 @@ def make_genai_config(self, model_name_or_path, extra_kwargs, out_dir):
         decoder["inputs"]["past_conv_names"] = "past_conv.%d"
         decoder["outputs"]["present_conv_names"] = "present_conv.%d"
 
-        # Hybrid models don't support shared past/present buffer
-        genai_config["search"]["past_present_share_buffer"] = False
+        # past_present_share_buffer: LFM2Cache pre-allocates KV to search.max_length when
```
kunal-vaishnavi: You can remove these comments about past-present share buffer now that it is supported.

src/models/kv_cache.cpp
```diff
@@ -205,7 +207,7 @@ int64_t DetectAndConfigureFixedKvShape(const SessionInfo& session_info,
   }
   past_present_share_buffer = true;
   if (g_log.enabled) {
-    Log("info", "DefaultKeyValueCache: auto-detected fixed kv-cache seq_len=" +
+    Log("info", std::string(cache_name) + ": auto-detected fixed kv-cache seq_len=" +
```
kunal-vaishnavi, line 210: Is the `kv-cache` portion of the error message still correct when it's a conv cache? Should the error message be further adjusted?

kunal-vaishnavi:
Since there are several LFM models under 3B params that are small enough to run in the CIs, can we add some unit testing that runs them? We can add any Hugging Face repo ids here.

https://github.com/microsoft/onnxruntime-genai/blob/ff53d6b9ba67331b38545f585ce800ef3e19e5ea/test/python/_test_utils.py#L147-L156

src/models/kv_cache.cpp
SanjayAMD: DetectAndConfigureFixedKvShape only ever inspects the attention KV inputs (kv_input_name_strings_) — the LFM2 conv-state cache is allocated on a separate path and never calls this helper, so "kv-cache" is accurate in both the DefaultKeyValueCache and LFM2Cache cases.

SanjayAMD:
> Since there are several LFM models under 3B params that are small enough to run in the CIs, can we add some unit testing that runs them? We can add any Hugging Face repo ids here.
> 
> https://github.com/microsoft/onnxruntime-genai/blob/ff53d6b9ba67331b38545f585ce800ef3e19e5ea/test/python/_test_utils.py#L147-L156

Ran the two newly added LFM2 model entries through the CI test flow locally (build INT4 CPU via the model builder + `run_model` batch generation from `test_onnxruntime_genai_e2e.py`):

| Model | Build | Generation |
|---|---|---|
| `lfm2.5-350m` (`LiquidAI/LFM2.5-350M`) | ✅ | ✅ |
| `lfm2.5-1.2b` (`LiquidAI/LFM2.5-1.2B-Instruct`) | ✅ | ✅ |

Both models build cleanly to INT4/CPU and generate valid sequences (batch of 3 prompts, `max_length=200`).

src/python/py/models/builders/lfm2.py
SanjayAMD: done, I have removed the comments in latest commit 
Thanks

kunal-vaishnavi (dismissed):

kunal-vaishnavi:
Please sync with the main branch to get the latest CI changes.

SanjayAMD:
hey @kunal-vaishnavi 
The CI failure here is a transformers-version mismatch, not a code issue.

The LiquidAI/LFM2.5-350M model was published with transformers >= 5, so its tokenizer_config.json uses the new "tokenizer_class": "TokenizersBackend". OGA's Python test suite pins transformers<5.0.0 (test/python/requirements.txt), and transformers 4.x doesn't recognize that class, so AutoTokenizer.from_pretrained fails during save_processing:

ValueError: Tokenizer class TokenizersBackend does not exist or is not currently imported.
The model builds fine (the ONNX export and GenAI config succeed) — it only fails at the tokenizer/processing step.

We should either disable the tests for these models or update the transformers version in the CI (verified and passing in my local env)

kunal-vaishnavi:
> hey @kunal-vaishnavi The CI failure here is a transformers-version mismatch, not a code issue.
> 
> The LiquidAI/LFM2.5-350M model was published with transformers >= 5, so its tokenizer_config.json uses the new "tokenizer_class": "TokenizersBackend". OGA's Python test suite pins transformers<5.0.0 (test/python/requirements.txt), and transformers 4.x doesn't recognize that class, so AutoTokenizer.from_pretrained fails during save_processing:
> 
> ValueError: Tokenizer class TokenizersBackend does not exist or is not currently imported. The model builds fine (the ONNX export and GenAI config succeed) — it only fails at the tokenizer/processing step.
> 
> We should either disable the tests for these models or update the transformers version in the CI (verified and passing in my local env)

Let's update the transformers version in the CIs.

https://github.com/microsoft/onnxruntime-genai/blob/a8072b2a5294dc9bbdafab234fab6b8c4ed10c05/test/python/requirements.txt#L10

kunal-vaishnavi (approved):

PR #2287, opened 2026-07-09
Update base speculative decoding implementation (v1)

samsat701:
### Description

This PR is the v1 follow-up to[ #2233](https://github.com/microsoft/onnxruntime-genai/pull/2233) v0 implementation.

[#2233](https://github.com/microsoft/onnxruntime-genai/pull/2233) introduced the core speculative decoding implementation - a draft model proposes K tokens, the target verifies them, and accepted/correction/bonus tokens are buffered and emitted one per `GenerateNextToken()` call.

This v1 PR focuses on closing the main functional and API gaps between speculative decoding and regular single-sequence decoder-only generation.

### What changed from v0

| Area | v0 | v1 |
|---|---|---|
| Core algorithm | Draft proposal, batched target verification, correction/bonus tokens, buffered delivery, deferred re-anchoring, and next-round anchor folding | Retained and hardened |
| Sampling | Supported `do_sample`, `top_k`, `top_p`, `temperature`, and `random_seed` | Preserved with expanded test coverage |
| Search penalties | Rejected `repetition_penalty != 1.0` and `min_length > 0` | Supports `repetition_penalty` and `min_length` using helpers shared with regular decoding |
| Continuous decoding | Rewind followed by appended tokens worked, but direct rewind-to-generation and mid-round state changes were unsupported | Supports direct rewind-and-resume, repeated append/rewind flows, and child-cache reconciliation |
| Guidance | Rejected | Supports guided greedy and sampling generation |
| External APIs | Speculative composite forwarding and buffered-state handling were incomplete | Supports named I/O, extra inputs, target adapters, runtime options, profiling, cancellation, and external logits |
| Statistics | Exposed 10 fields | Exposes 29 statistics through a GenAI-defined object with typed, name-based getters |
| Model recognition | Required explicit `model.type = "speculative"` | Recognizes model.type = `speculative` or any model configuration with a draft decoder |
| Language bindings | C, C++, and Python | Adds C#, Java/JNI, and Objective-C |
| CPU-side overhead | Stored dense target distributions and used one-hot logits for token commits | Uses sparse target rows, direct token commits, reusable buffers, raw pending logits, and lazy correction buffers |

### Guidance

Guidance is implemented by giving draft lookahead an independent clone of the grammar cursor.

This allows the draft model to propose several tokens without prematurely advancing the externally visible guidance state. During proposal and verification:

- Guidance masks are applied using the same mask/commit operations as regular generation.
- Search penalties are applied after guidance masking.
- Accepted and emitted tokens are committed to the main grammar state.
- Grammar fast-forward tokens are included in the target verification batch rather than being verified through separate target calls.
- Both greedy and sampling paths are supported.

This preserves speculative lookahead while keeping the externally visible grammar state aligned with the tokens returned to the caller.

### Continuous decoding

The speculative implementation can internally evaluate several tokens ahead of the sequence visible to the caller. Consequently, rewind or append operations may occur while accepted tokens are still buffered or while the target and draft caches are at different internal positions.

v1 tracks and reconciles these states by:

- Discarding pending speculative work that is no longer valid after a rewind or append.
- Reconciling both target and draft caches with the externally visible sequence.
- Re-establishing the logits and draft anchor required for the next speculative round.
- Supporting direct `RewindTo()` followed by `GenerateNextToken()`.
- Supporting append operations performed while speculative tokens are buffered.
- Supporting repeated append, rewind, and generation cycles.
- Re-prefilling caches when supported Phi3-family models cross a long-context ROPE transition.

This is same-generator continuous decoding. Engine batching and continuous batching remain separate features (not included in this PR).

### Search parameters

v1 supports the parameters applicable to regular single-sequence greedy and sampling generation:

- `do_sample`
- `top_k`
- `top_p`
- `temperature`
- `random_seed`
- `repetition_penalty`
- `min_length`
- `max_length`
- EOS handling

`early_stopping` and `length_penalty` apply to beam search (speculative decoding does not currently support).

`no_repeat_ngram_size` and `diversity_penalty` are not implemented by regular decoding either.

### External API behavior

The speculative model now handles the regular generator APIs consistently across its target and draft children:

- Resolves named inputs and outputs.
- Propagates extra inputs.
- Applies adapters to the target model.
- Forwards runtime options, profiling, and termination.
- Provides observational `GetLogits()` behavior without changing speculative state.
- Supports `SetLogits()` while safely invalidating speculative work derived from the previous logits.

### Statistics

The speculative statistics contract was expanded from 10 to 29 fields.

The updated statistics distinguish between:

- Proposed, accepted, rejected, and discarded draft tokens.
- Target and draft tokens evaluated.
- Tokens emitted to the caller.
- Tokens currently buffered.
- Completed and interrupted speculative rounds.
- Target and draft forward passes.
- Cache reconciliation and re-prefill work.
- Draft, target, sampling, and total generation time.
- Acceptance rate, throughput, and formula-based speedup estimates.

This avoids counting internally evaluated or buffered tokens as if they had already been emitted.

### Language bindings

The speculative options and statistics APIs are now exposed through:

- C
- C++
- Python
- C#
- Java/JNI
- Objective-C

C, C++, and Python support existed in v0. v1 expands their statistics coverage and adds C#, Java/JNI, and Objective-C support.

### Supported scope and current limitations for v1

- Target and draft must use the same execution-provider configuration.
- Target and draft vocabularies must be compatible.
- Models must use supported, separately rewindable KV-cache layouts.
- Physically sliding KV caches, LFM2 hybrid state, pipeline models, multimodal models, and other excluded model/cache layouts are not supported.
- Beam search and multiple return sequences are not supported.
- Engine batching and continuous batching are not supported.
- Performance gains depend on the model pair, hardware, prompt, sampling configuration, draft length, and acceptance rate.

### Validation

Validation performed for this PR includes:

- The full speculative Python test suite.
- Native C++ unit tests.
- C# Release build and focused speculative tests.
- Java formatting, compilation, JNI build, and focused speculative tests.
- Objective-C API and test source review; Objective-C runtime validation was not available on Windows.

The speculative Python suite expanded from 25 test functions in v0 to 80 in v1, covering penalties, sampling, guidance, continuation, external APIs, statistics, and Phi3-family long-context transitions. 

Model-backed speculative performance benchmarking is maintained separately on the following branch:
https://github.com/microsoft/onnxruntime-genai/tree/t-sshrimali/speculative-decoding-benchmarking

### Related PRs

- Base speculative decoding v0: #2233


src/csharp/NativeMethods.cs
```diff
@@ -129,6 +163,16 @@ internal class NativeLib
                                                                                      byte[] /* const char* */ searchOption,
                                                                                      out bool /* const bool* */ value);
 
+        [DllImport(NativeLib.DllName, CallingConvention = CallingConvention.Winapi)]
+        public static extern IntPtr /* OgaResult* */ OgaGeneratorParamsSetSpeculativeNumber(IntPtr /* OgaGeneratorParams* */ generatorParams,
```
kunal-vaishnavi, line 134: Earlier, you called the method as `SetSpeculativeOption` and here you call it as `SetSpeculativeNumber`. There is an API naming mismatch.

src/cuda/search_cuda.cpp
```diff
@@ -178,6 +178,26 @@ void GreedySearch_Cuda::SampleTopKTopP(int k, float p, float temperature) {
   }
 }
 
+void GreedySearch_Cuda::CommitToken(int32_t token) {
```
kunal-vaishnavi: Can you document and explain how this differs from `AppendTokens` besides that it adds one token to the sequences?

https://github.com/microsoft/onnxruntime-genai/blob/3bd384b0cff741fe74f713f100fde409b93a4156/src/cuda/search_cuda.cpp#L241-L257

There is also some duplicate logic in both files. Let's refactor and use shared commonality where possible.

src/csharp/NativeMethods.cs
```diff
@@ -8,6 +8,40 @@ namespace Microsoft.ML.OnnxRuntimeGenAI
 {
     internal static class NativeMethods
     {
+        [StructLayout(LayoutKind.Sequential)]
+        internal struct OgaSpeculativeStats
```
kunal-vaishnavi: Can we follow the approach used for getting and setting generator params across language bindings? That way, we don't have to enumerate each attribute within the struct.

src/csharp/SpeculativeStats.cs
```diff
@@ -0,0 +1,71 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+namespace Microsoft.ML.OnnxRuntimeGenAI
+{
+    public readonly struct SpeculativeStats
+    {
+        internal SpeculativeStats(NativeMethods.OgaSpeculativeStats stats)
```
kunal-vaishnavi: Let's see if there is another way we can do this so that we don't have to enumerate each attribute. It is fine to return a GenAI-defined object such as `OgaSpeculativeStats` and then let that object have its own APIs for obtaining these stats.

src/java/src/main/java/ai/onnxruntime/genai/GeneratorParams.java
```diff
@@ -121,4 +151,10 @@ private native void setSearchOptionBool(long nativeHandle, String optionName, bo
   private native double getSearchNumber(long nativeHandle, String optionName) throws GenAIException;
 
   private native boolean getSearchBool(long nativeHandle, String optionName) throws GenAIException;
+
+  private native void setSpeculativeOptionNumber(long nativeHandle, String optionName, double value)
```
kunal-vaishnavi: Let's keep the API names consistent to `setSpeculativeNumber` and `getSpeculativeNumber`.

src/java/src/main/java/ai/onnxruntime/genai/SpeculativeStats.java
```diff
@@ -0,0 +1,186 @@
+/*
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+package ai.onnxruntime.genai;
+
+/** Speculative decoding work, delivery, timing, and speedup statistics. */
```
kunal-vaishnavi: Same comment as [this](https://github.com/microsoft/onnxruntime-genai/pull/2287/changes#r3574250930)

src/models/model.cpp
```diff
@@ -853,7 +853,11 @@ std::unique_ptr<Config> CreateConfig(OrtEnv& ort_env, const char* config_path, c
 
 std::shared_ptr<Model> CreateModel(OrtEnv& ort_env, std::unique_ptr<Config> config) {
   // Speculative decoding wraps two decoder-only models (target + draft) under single model.
-  if (config->model.type == "speculative")
+  const bool is_phi3_speculative =
```
kunal-vaishnavi: Why Phi-3 specific?

src/objectivec/include/ort_genai_objc.h
```diff
@@ -47,6 +47,38 @@ typedef NS_ENUM(NSInteger, OGAElementType) {
   OGAElementTypeUint64,   // maps to c type uint64_t
 };
 
+typedef struct OGASpeculativeStats {
```
kunal-vaishnavi: Same comment as [here](https://github.com/microsoft/onnxruntime-genai/pull/2287/changes#r3574250930)

src/objectivec/include/ort_genai_objc.h
```diff
@@ -311,6 +343,25 @@ typedef NS_ENUM(NSInteger, OGAElementType) {
...
 - (BOOL)getSearchBool:(NSString*)key
                 error:(NSError**)error;
+
+/**
+ * Set a numerical speculative decoding option.
+ * @param key The option key.
+ * @param value The option value.
+ * @param error Optional error information set if an error occurs.
+ */
+- (BOOL)setSpeculativeOption:(NSString*)key
```
kunal-vaishnavi: Same comment as [here](https://github.com/microsoft/onnxruntime-genai/pull/2287#discussion_r3574277100)

src/base_speculative_strategy.cpp
```diff
@@ -41,45 +47,129 @@ SpeculativeDecodingStrategy::Proposal BaseSpeculativeStrategy::Propose(
...
+  std::unique_ptr<ConstrainedLogitsProcessor> draft_grammar;
+  DeviceSpan<float> guidance_buf;
+  if (guidance_active) {
+    draft_grammar = g.guidance_logits_processor_->Clone();
+    guidance_buf = params.p_device->Allocate<float>(static_cast<size_t>(vocab_size));
+  }
+  if (penalties_active || guidance_active)
+    row_buf.resize(static_cast<size_t>(vocab_size));
+
+  auto is_eos = [&](int32_t t) {
```
kunal-vaishnavi: We can just call `std::find` directly instead of wrapping it in `is_eos`.

src/decoding_strategy.cpp
```diff
@@ -23,7 +23,12 @@ std::unique_ptr<DecodingStrategy> MakeDecodingStrategy(Generator& generator) {
   const auto& model_type = generator.model_->config_->model.type;
   if (ModelType::IsTransducer(model_type))
     return std::make_unique<TransducerDecodingStrategy>(generator);
-  if (ModelType::IsSpeculative(model_type))
+  const auto& model = generator.model_->config_->model;
+  const bool is_phi3_speculative =
```
kunal-vaishnavi: Why Phi-3?

src/sampling_distribution.h
```diff
@@ -14,6 +14,51 @@
...
 namespace Generators {
 
+// --- Logits penalties (min-length + repetition) -------------------------------------------
+// Shared CPU logits-processing helpers used by both standard search (Search_Cpu) and speculative
+// decoding, so the two paths apply identical penalties to a single logits row.
+
+// Mask the end-of-stream tokens (set to lowest float) while the sequence is shorter than
+// min_length, so EOS cannot be selected yet. Mirrors Search_Cpu::ApplyMinLength for one row.
+// current_length is the sequence length *before* the token this row predicts is appended.
+inline void ApplyMinLengthToLogits(std::span<float> logits, int current_length, int min_length,
```
kunal-vaishnavi, line 24: The search files already implement applying min length and repetition penalty. Why are we re-implementing that here? Have we already migrated the implementations over?

src/speculative_decoding_strategy.cpp
```diff
@@ -13,41 +14,246 @@
...
-  if (greedy) {
-    // Greedy accept/correct/bonus only needs argmax.
-    std::vector<float> onehot(static_cast<size_t>(vocab_size), 0.0f);
-    const int32_t idx = static_cast<int32_t>(std::max_element(logits.begin(), logits.end()) - logits.begin());
-    onehot[static_cast<size_t>(idx)] = 1.0f;
-    return onehot;
-  }
-  ComputeSampledCategorical(logits, search.top_k, search.top_p, search.temperature, sampled);
-  return ScatterToFullVocab(sampled, vocab_size);
+namespace {
```
kunal-vaishnavi, line 17: Why do we need an empty namespace? Can we remove it?

src/speculative_decoding_strategy.cpp
```diff
@@ -121,6 +328,80 @@ void SpeculativeDecodingStrategy::RunRound(Generator& g) {
...
+  std::vector<float> row_buf;
+  if (penalties_active) {
+    auto seed = g.search_->GetSequence(0).CopyDeviceToCpu();
+    seed_prefix.assign(seed.begin(), seed.end());
+    row_buf.resize(static_cast<size_t>(vocab_size));
+  }
+
+  // Apply penalties (when active) to a logits row in place and return a view of the (possibly
+  // rewritten) row. No allocation/copy on the default no-penalty path.
+  auto penalize = [&](std::span<const float> logits, int current_length,
```
kunal-vaishnavi: This `penalize` looks the same as in `base_decoding_strategy.cpp`. Let's refactor and re-use common logic.

src/speculative_decoding_strategy.cpp
```diff
@@ -121,6 +328,80 @@ void SpeculativeDecodingStrategy::RunRound(Generator& g) {
...
+    auto row = penalize(logits, current_length, prefix);
+    if (greedy_accept) {
+      pos0_argmax = RowArgmax(row);
+    } else {
+      ComputeSampledCategorical(row, search.top_k, search.top_p, search.temperature, scratch_sc);
+      pos0_sparse = {scratch_sc.indices, scratch_sc.probs};
+    }
+  };
+  // Store target prediction row i (the target's prediction after proposal token i).
+  auto store_row = [&](int i, std::span<const float> logits, int current_length,
```
kunal-vaishnavi: There is quite high usage of lambda functions in this PR. I'm assuming they are AI-generated. Let's remove them where not needed and use named helper functions instead.

src/csharp/NativeMethods.cs
samsat701: Fixed in https://github.com/microsoft/onnxruntime-genai/pull/2287/changes/04f612a321739aafa86a2d5b0f2fe8a57c1d175c - all API names are consistent with SetSpeculativeNumber/GetSpeculativeNumber (language-appropriate casing).

src/java/src/main/java/ai/onnxruntime/genai/GeneratorParams.java
samsat701: Fixed in https://github.com/microsoft/onnxruntime-genai/pull/2287/changes/04f612a321739aafa86a2d5b0f2fe8a57c1d175c - all API names are consistent with SetSpeculativeNumber/GetSpeculativeNumber (language-appropriate casing).

src/objectivec/include/ort_genai_objc.h
samsat701: Fixed in https://github.com/microsoft/onnxruntime-genai/pull/2287/changes/04f612a321739aafa86a2d5b0f2fe8a57c1d175c - all API names are consistent with SetSpeculativeNumber/GetSpeculativeNumber (language-appropriate casing).

src/base_speculative_strategy.cpp
samsat701: Fixed - https://github.com/microsoft/onnxruntime-genai/pull/2287/changes/f6c5c62be0634fd64f331b3c9fe4b8f41d6aef6e. I removed is_eos and used std::find directly.

src/speculative_decoding_strategy.cpp
samsat701: Fixed - https://github.com/microsoft/onnxruntime-genai/commit/f6c5c62be0634fd64f331b3c9fe4b8f41d6aef6e. Moved the repeated penalty state + row-processing logic into the shared LogitsPenaltyProcessor (now used by both draft and target speculative paths).

src/speculative_decoding_strategy.cpp
samsat701: Fixed - https://github.com/microsoft/onnxruntime-genai/commit/f6c5c62be0634fd64f331b3c9fe4b8f41d6aef6e. Replaced lambdas with named helper functions + a shared helper object for the penalty processing.

src/cuda/search_cuda.cpp
samsat701: Fixed in https://github.com/microsoft/onnxruntime-genai/pull/2287/changes/1fe0808538e716c3b2f7906dd1b06a1d670b2f92. Refactored to share sequence append, max-length, and generated-token finalization logic. 

CommitToken commits an already-selected generated token and applies EOS, padding, and max-length handling

AppendTokens ingests prompt/continuation tokens. It skips generated-token EOS/padding handling but does max-length handling.

src/sampling_distribution.h
samsat701: Yes, the existing CPU search implementations were migrated into the shared ApplyMinLengthToLogits and ApplyRepetitionPenaltyToLogits helpers; this is used by both standard and speculative decoding. Also consolidated the repeated speculative-path setup into LogitsPenaltyProcessor.

src/speculative_decoding_strategy.cpp
samsat701: Unnamed namespace now holds the helper functions and types used only in this file from lambda function refactor.

src/models/model.cpp
samsat701: Fixed in https://github.com/microsoft/onnxruntime-genai/pull/2287/changes/b80f517d1b7dd811b3f025ddc0be35e62fe45d89. 

Removed Phi-3-specific speculative detection. Base speculative when model.type == "speculative" or a draft decoder is configured. Preserves original model type for existing Phi-3/LongRoPE handling.

src/decoding_strategy.cpp
samsat701: Fixed in https://github.com/microsoft/onnxruntime-genai/pull/2287/changes/b80f517d1b7dd811b3f025ddc0be35e62fe45d89. 

Removed Phi-3-specific speculative detection. Base speculative when model.type == "speculative" or a draft decoder is configured. Preserves original model type for existing Phi-3/LongRoPE handling.

src/csharp/NativeMethods.cs
samsat701: Fixed - https://github.com/microsoft/onnxruntime-genai/pull/2287/changes/0b78f7166cd0fa26a26edbae3d160ee57c96dd50. 

Stats now provided through GenAI object with typed name-based getters. The lang bindings only wrap this object (no longer duplicate or enumerate the full statistics field list).

src/csharp/SpeculativeStats.cs
samsat701: Fixed - https://github.com/microsoft/onnxruntime-genai/pull/2287/changes/0b78f7166cd0fa26a26edbae3d160ee57c96dd50. 

Stats now provided through GenAI object with typed name-based getters. The lang bindings only wrap this object (no longer duplicate or enumerate the full statistics field list).

src/java/src/main/java/ai/onnxruntime/genai/SpeculativeStats.java
samsat701: Fixed - https://github.com/microsoft/onnxruntime-genai/pull/2287/changes/0b78f7166cd0fa26a26edbae3d160ee57c96dd50. 

Stats now provided through GenAI object with typed name-based getters. The lang bindings only wrap this object (no longer duplicate or enumerate the full statistics field list).

src/objectivec/include/ort_genai_objc.h
samsat701: Fixed - https://github.com/microsoft/onnxruntime-genai/pull/2287/changes/0b78f7166cd0fa26a26edbae3d160ee57c96dd50. 

Stats now provided through GenAI object with typed name-based getters. The lang bindings only wrap this object (no longer duplicate or enumerate the full statistics field list).

src/csharp/GeneratorParams.cs
```diff
@@ -46,6 +46,17 @@ public bool GetSearchBool(string searchOption)
             return value;
         }
 
+        public void SetSpeculativeNumber(string speculativeOption, double value)
```
kunal-vaishnavi: Can we rename `speculativeOption` to a simpler phrase such as `name`?

src/models/model.cpp
```diff
@@ -853,7 +853,7 @@ std::unique_ptr<Config> CreateConfig(OrtEnv& ort_env, const char* config_path, c
 
 std::shared_ptr<Model> CreateModel(OrtEnv& ort_env, std::unique_ptr<Config> config) {
   // Speculative decoding wraps two decoder-only models (target + draft) under single model.
-  if (config->model.type == "speculative")
+  if (ModelType::UsesDraftModelSpeculation(config->model.type, config->model.draft.filename))
```
kunal-vaishnavi, line 874: What about `ModelType::IsSpeculative`?

src/speculative_decoding_strategy.h
```diff
@@ -88,6 +139,28 @@ struct SpeculativeDecodingStrategy : DecodingStrategy {
...
+
+  // How many leading committed tokens were part of the target verify batch (so their KV is already
+  // built). FinalizeGuidanceRound only replays committed tokens past this point.
+  int saved_verify_prefix_{};
+
+  // The last verify row for the committed prefix, reused as the next round's pos0 when nothing needs
+  // replaying (avoids an extra target pass).
+  std::vector<float> saved_last_row_;
+
+  // Set while a round has left the inner KV caches out of sync with the committed sequence (mid round/deferred fold).
```
kunal-vaishnavi: Rather than introducing so many boolean and int flags to keep track of states, can we refactor into a shared struct (e.g. `SpeculativeRoundState`) even if the struct is minimal and can we document phase transitions as comments?

```cpp
struct SpeculativeRoundState {
  enum Phase { Idle, ProposalReady, VerifyComplete, Draining, Finalize };
  Phase phase_;  // ... all round-specific state
};

class SpeculativeDecodingStrategy {
  SpeculativeRoundState round_state_;
  
  // Document valid transitions:
  // Idle → ProposalReady → VerifyComplete → Draining → Finalize → Idle
  // Any phase may jump to Reconciling if external op occurs
};
```

This is a similar approach to how the current status is tracked inside the generator.

https://github.com/microsoft/onnxruntime-genai/blob/ac7df6f369c6863750bc2c1995410cd1afb202a7/src/generators.h#L133-L136

Currently, it is difficult for the reader to infer these flags' valid combinations, guess which transitions are allowed, and hope that they never land in an impossible state.

src/generators.cpp
```diff
@@ -523,6 +523,11 @@ void Generator::AppendTokens(cpu_span<const int32_t> input_ids) {
     set_extra_inputs_ = false;
   }
```
kunal-vaishnavi, line 664: Couple of documentation-related points:

1. Can you create a new Markdown file explaining all of your changes and store it in `docs`? Ex:

```md
## Architecture Overview

### Data Flow During a Speculative Round

1. **PROPOSAL PHASE**: Draft generates K tokens (logits → sampling/argmax)
2. **VERIFY PHASE**: Target verifies all K tokens in one batched pass
3. **ACCEPT PHASE**: Compare draft vs target logits, decide accept/reject
4. **EMIT PHASE**: Buffer accepted tokens, emit one per GenerateNextToken()
5. **FINALIZE PHASE**: Re-anchor target KV cache for next round

### Cache State Tracking

Two caches are maintained:
- **Target KV cache**: Current length = verified_length
- **Draft KV cache**: Current length = proposed_length
- **Visible sequence**: Emitted tokens only

When caches_out_of_sync = true:
  - Target KV ahead of visible by at most K tokens
  - Draft KV ahead of visible by at most K tokens
  - External ops (AppendTokens/SetLogits) must reconcile
```

2. Can you create a state interaction matrix in a comment block that explains how guidance and continuous decoding work together? Ex:

```cpp
/*
 * GUIDANCE + CONTINUOUS DECODING INTERACTION MATRIX:
 * 
 * Operation     | Guidance Active | Effect
 * --------------|-----------------|----------------------------------
 * AppendTokens  | No              | Reconcile caches, preserve committed
 * AppendTokens  | Yes (mid-round) | ERROR: Cannot reconcile with active grammar
 * AppendTokens  | Yes (idle)      | OK: Grammar state is consistent
 * RewindTo(0)   | Yes             | Reset grammar to initial state
 * RewindTo(N>0) | Yes             | UNSUPPORTED: llguidance cannot roll back
 * SetLogits     | Yes (mid-round) | ERROR: Would break grammar consistency
 * 
 * ff_carry_ is discarded on any external op that resets the round.
 */
```

Key invariants should also be noted. Ex:

```cpp
// INVARIANTS when round_state_.phase == Draining:
// - pending_tokens is non-empty
// - caches_out_of_sync == true (verify moved them ahead)
// - cached_target_logits holds rows for direct tokens
// - Grammar (if active) is advanced past emitted tokens
// - Next GenerateNextToken() call may call TryGetExternalLogits()
```

src/speculative_decoding_strategy.cpp
```diff
@@ -13,41 +14,325 @@
...
+namespace {
+// A target verify row kept cheaply (item #2): greedy needs only the argmax id; sampling keeps the
+// truncated categorical (kept ids + renormalized probs) -- far smaller than a dense vocab-sized
+// vector -- and densifies it on demand only for the single correction/bonus row.
+struct SparseRow {
+  std::vector<int32_t> indices;
+  std::vector<float> probs;
+};
+
+int32_t RowArgmax(std::span<const float> logits) {
```
kunal-vaishnavi: I have seen this defined in `base_speculative_strategy.cpp` as well. Can we refactor and reuse?

src/speculative_decoding_strategy.cpp
```diff
@@ -13,41 +14,325 @@
 
 namespace Generators {
 
-// Turn a logits row into the probability distribution the accept step compares against.
-static std::vector<float> LogitsToDistribution(std::span<const float> logits, bool greedy,
-                                               const Config::Search& search, int vocab_size,
-                                               SampledCategorical& sampled) {
+namespace {
+// A target verify row kept cheaply (item #2): greedy needs only the argmax id; sampling keeps the
```
kunal-vaishnavi: From Copilot:

## Documentation gap: Sparse row storage tradeoff

The **`SparseRow` struct and lookup functions** (`GetSparseProbability`, `DensifySparseRow`) implement an important optimization that should be explicitly documented.

### What's happening

The code replaces v0's dense target logits storage (`K * vocab_size` floats ≈ 2MB per round) with sparse rows:
- **Greedy**: stores only the argmax ID (1 int per row) via `RowArgmax()`
- **Sampling**: stores truncated categorical (top_k IDs + probabilities, ~50 floats vs 128K)
- **Correction**: lazily densifies ONE row (via `DensifySparseRow()`) only on rejection

### The tradeoff

This trades **O(1) lookup** for **O(top_k) linear scan** in `GetSparseProbability()`:

```cpp
// v0: O(1) direct index, but stores 2MB
const float p_t = target_dist[proposal.tokens[i]];

// v1: O(50) linear scan, but stores ~1KB
const float p_t = GetSparseProbability(sr, proposal.tokens[i]);
```

**Why it's worthwhile:**
- Lookup cost: ~50 comparisons × 4 tokens/round = 400 nanoseconds per round (negligible)
- Memory saved: 2MB → 1KB per round (avoids allocation overhead + cache pressure)
- GPU bottleneck anyway: millisecond-scale target forward passes dwarf microsecond-scale CPU lookups

### Suggested documentation

Add a comment block above `GetSparseProbability()` or `SparseRow`:

```cpp
// Sparse target-verify storage (item #2 from PR description):
//   - Greedy: only the argmax ID (1 int)
//   - Sampling: truncated categorical (~50 items, not vocab_size)
//   - Correction: single on-demand densification
//
// Per-round memory: 2MB (v0 dense) → 1KB (v1 sparse)
// Per-lookup cost: O(1) → O(top_k) ≈ 400ns/round (acceptable vs allocation overhead)
```

src/speculative_decoding_strategy.cpp
```diff
@@ -13,41 +14,325 @@
...
+    // to start verify (pending_anchor_token_). After RewindToLength -> stale, so replay
+    // the boundary token like StandardDecodingStrategy - ComputeLogits -> Run refreshes both the
+    // target logits and draft_pending_logits_ before RunRound.
+    if (!g.computed_logits_ && !pending_anchor_token_.has_value()) {
+      if (g.search_->GetSequenceLength() == 0)
+        throw std::runtime_error(
+            "Speculative decoding: GenerateNextToken called with no prior state. Please call "
+            "AppendTokens, SetLogits, or SetInputs before calling GenerateNextToken.");
+      auto next_tokens = g.search_->GetNextTokens();
+      if (g.last_action_ == Generator::Action::rewound)
```
kunal-vaishnavi, line 65: Can we make the `Action` enum public so that you can do `Action::rewound`?

src/speculative_decoding_strategy.cpp
```diff
@@ -13,41 +14,325 @@
...
   DrainOne(g);
 }
 
 void SpeculativeDecodingStrategy::Reset() {
-  pending_.clear();
+  DiscardPendingTokens();
+  ClearPendingExternalLogits();
+  reanchor_pending_ = false;
+  pending_anchor_token_.reset();
+  round_dirty_ = false;
```
kunal-vaishnavi: Examples of some booleans for [this issue](https://github.com/microsoft/onnxruntime-genai/pull/2287#discussion_r3664067569)

src/speculative_decoding_strategy.cpp
```diff
@@ -387,53 +788,369 @@ void SpeculativeDecodingStrategy::FinalizeRound(Generator& g) {
...
+                             logits_name + "'.");
+  auto raw_shape = raw_ort->GetTensorTypeAndShapeInfo()->GetShape();
+  const bool is_multiple_tokens =
+      (raw_shape.size() >= 2 && raw_shape[1] == static_cast<int64_t>(K));
+  if (is_multiple_tokens) {
+    auto& target_device = *spec_state->spec_model().target_model().p_device_inputs_;
+    const auto elem_type = raw_ort->GetTensorTypeAndShapeInfo()->GetElementType();
+    DeviceSpan<float> verify_logits;
+    if (elem_type == Ort::TypeToTensorType<float>) {
+      verify_logits = WrapTensor<float>(target_device, *raw_ort);
```
kunal-vaishnavi: The logits casting is duplicated on lines 517-522 as well. Let's put this into a helper function.

src/speculative_decoding_strategy.cpp
```diff
@@ -387,53 +788,369 @@ void SpeculativeDecodingStrategy::FinalizeRound(Generator& g) {
...
+  int n_evaluated = 0;
+  // leading committed tokens that came from the verify batch
+  int verify_prefix = 0;  
+  bool eos_hit = false, rejected = false;
+  // repetition-penalty context, grows with committed tokens
+  std::vector<int32_t> dist_prefix; 
+  if (penalty_processor.IsActive()) dist_prefix = seed_prefix;
+
+  int i = 0;
+  for (; i < K; i++) {
```
kunal-vaishnavi: Any reason for not doing the following?


```suggestion
  for (int i = 0; i < K; i++) {
```

samsat701: Fixed in https://github.com/microsoft/onnxruntime-genai/pull/2287/changes/7eae6bc81c68c4057272dacaf42573294e9d2261.

src/speculative_decoding_strategy.cpp
samsat701: Fixed in https://github.com/microsoft/onnxruntime-genai/pull/2287/changes/7eae6bc81c68c4057272dacaf42573294e9d2261. Made Generator::Action a public scoped enum so callers can use Generator::Action::rewound (while keeping last_action_ private).

src/csharp/GeneratorParams.cs
samsat701: Fixed in https://github.com/microsoft/onnxruntime-genai/pull/2287/changes/7eae6bc81c68c4057272dacaf42573294e9d2261. Renamed speculativeOption to name consistently in the C# setter, getter, and native declarations.

src/speculative_decoding_strategy.cpp
samsat701: Fixed in https://github.com/microsoft/onnxruntime-genai/pull/2287/changes/7eae6bc81c68c4057272dacaf42573294e9d2261. Applied the shared target-row selection approach from advanced PR #2315 here as well; removing the duplicate verifier-side RowArgmax and adding focused unit tests.

src/speculative_decoding_strategy.cpp
samsat701: Fixed in https://github.com/microsoft/onnxruntime-genai/pull/2287/changes/7eae6bc81c68c4057272dacaf42573294e9d2261. Extracted the FP32 wrapping/casting logic into GetFloatVerifyLogits() and reused it in both normal + guidance verification.

src/speculative_decoding_strategy.h
samsat701: Fixed in https://github.com/microsoft/onnxruntime-genai/pull/2287/changes/7eae6bc81c68c4057272dacaf42573294e9d2261. Consolidated the round-specific flags and context into RoundState with explicit phases and documented transitions.

src/speculative_decoding_strategy.cpp
samsat701: Fixed - https://github.com/microsoft/onnxruntime-genai/pull/2287#discussion_r3664067569.

src/models/model.cpp
samsat701: IsSpeculative only checks model.type == "speculative". We use UsesDraftModelSpeculation (which also uses IsSpeculative) because draft-model speculation also supports regular model types configured with draft.filename (eg. Phi3-family models are supported whose original model type must be preserved for RoPE-factor handling).

PR #2290, opened 2026-07-11
Merge pipeline config overlays by model ID

### LINKED ISSUE #2004: Config::overlay() on pipeline session_options causes "File doesn't exist" on Model load
**Describe the bug**
Config::overlay() appends pipeline entries instead of merging by model_id, causing "File doesn't exist" error on Model load

When calling `Config.overlay()` with a JSON that contains a `model.decoder.pipeline` array,
the overlay **appends new pipeline entries** instead of **merging into existing entries by `model_id`**.
This leaves the pipeline vector with extra entries that have empty `filename` fields,
causing `og.Model(config)` to fail with:

**To Reproduce**
## Reproduction Steps
1. Load a `decoder-pipeline` model config (e.g. one with `embeddings`, `context`, `iterator`, `lm_head` in the pipeline)
2. Apply an overlay that targets pipeline sub-model `session_options`:

```python

config = og.Config("C:/path/to/model/")
overlay = '''{
  "model": {
    "decoder": {
      "pipeline": [
        {
          "submodel":  { "session_options": { "enable_profiling": "test" } }
        }
      ]
    }
  }
}'''
config.overlay(overlay)
model = og.Model(config)  # ← RuntimeError here

**Expected behavior**
The overlay should find the existing pipeline entry by model_id ("context", "iterator") and merge/update its fields in-place, consistent with how provider_options overlays work (matched by name).


**Additional context**
I guess the root cause is In src/config.cpp, PipelineModelObject_Element::OnObject() always calls emplace_back() without checking for an existing entry with the same model_id.

hogeheer499-commits:
## Description

Fixes #2004.

`Config::overlay()` currently appends decoder and vision pipeline entries even when an entry with the same `model_id` already exists. A partial component overlay can therefore create a duplicate entry with an empty `filename`, causing model loading to fail.

This change:
- reuses the matching pipeline entry and only appends unknown model IDs;
- preserves existing component filenames and `session_options` while applying a partial overlay;
- resets the active parser element before an append can reallocate the pipeline vector;
- applies the same semantics to decoder and vision pipelines.

The existing replacement behavior for `run_options` is unchanged.

## Testing

- Added `ConfigTests.PipelineOverlayMergesByModelId`, covering:
  - two existing IDs, with only the second targeted;
  - preservation of the untouched entry, filename, log ID, and provider options;
  - addition of a new ID;
  - decoder and vision pipelines.
- Confirmed the regression test fails against current `main` (four entries instead of three) and passes with this change.
- Built the CPU `RelWithDebInfo` targets and ran the full CTest suite: 100% passed.
- Verified end to end with the Python binding: the existing ID now creates one profiled session, while a new ID still adds a second session.
- `clang-format`, lintrunner, and `git diff --check` pass.

## AI assistance

OpenAI Codex (GPT-5) was used to assist with investigation, implementation, review, and testing. The resulting behavior was validated against current `main` with both unit and end-to-end tests.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -183,6 +183,7 @@ target_link_directories(onnxruntime-genai-obj PUBLIC ${ORT_LIB_DIR})
 target_link_libraries(onnxruntime-genai-obj PUBLIC Threads::Threads)
 
 if(WIN32)
+  target_compile_definitions(onnxruntime-genai-obj PRIVATE BUILDING_ORT_GENAI_INTERNALS)
   add_library(onnxruntime-genai SHARED "${GENERATORS_ROOT}/dll/onnxruntime-genai.rc")
   target_compile_definitions(onnxruntime-genai PRIVATE VERSION_INFO=\"${VERSION_INFO}\")
   target_compile_definitions(onnxruntime-genai PRIVATE VERSION_MAJOR=${VERSION_MAJOR})
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -458,7 +458,9 @@ struct PipelineModel_Element : JSON::Element {
 
   JSON::Element& OnObject(std::string_view name) override {
     if (name == "session_options") {
-      v_.session_options = Config::SessionOptions{};
+      if (!v_.session_options) {
+        v_.session_options.emplace();
+      }
       session_options_ = std::make_unique<SessionOptions_Element>(*v_.session_options);
       return *session_options_;
     }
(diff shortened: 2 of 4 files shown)


hogeheer499-commits:
Closing this as part of consolidating my open contributions — I would rather keep a few PRs I can follow through on than many parallel ones. As far as I can tell the underlying issue (#2004) and the approach here are still valid; happy to reopen or hand this off if the maintainers consider it useful.

PR #2292, opened 2026-07-13
Enable re-init after OgaShutdown

skottmckay:
Enable re-init after OgaShutdown. State returns to 'just loaded' post-OgaShutdown.
Allows user to control lifetime of the OrtEnv instance that is pinned in OgaGlobals.

See the plan document for details.

Adds some additional library load validation/diagnostic info (copied from ORT) so we stay in a clean state if an EP can't be loaded.

### DIFF
--- a/src/generators.cpp
+++ b/src/generators.cpp
@@ -77,6 +77,11 @@ OrtGlobals::OrtGlobals()
 // The single genai global. Lazily (re)created by GetOrtGlobals() and reset by Shutdown(), so that
 // genai can be torn down and re-initialized in-process (e.g. a host recreating its wrapper).
 static std::unique_ptr<OrtGlobals> g_ort_globals;
+// Guards lazy (re)creation and teardown of g_ort_globals so concurrent first-use (or first-use
+// after a shutdown) cannot double-construct the globals, and a concurrent Shutdown() cannot race
+// creation. Re-init after shutdown must remain possible, so this is a plain mutex rather than
+// std::call_once (which is process-once).
+static std::mutex g_ort_globals_mutex;
 // Set by the process-exit EnsureShutdown destructor to prevent GetOrtGlobals() from resurrecting
 // the globals during static destruction.
 static bool g_process_exiting = false;
@@ -94,14 +99,11 @@ std::unique_ptr<OrtGlobals>& GetOrtGlobals() {
   // Registered once; its destructor runs at process exit (before g_ort_globals is destroyed) and
   // performs the final Shutdown().
   static EnsureShutdown ensure_shutdown;
-  // Guard lazy (re)creation so concurrent first-use (or first-use after a shutdown) cannot
-  // double-construct the globals. Re-init after shutdown must remain possible, so this is a plain
-  // mutex rather than std::call_once (which is process-once). The OrtGlobals constructor does not
-  // re-enter GetOrtGlobals() (it bootstraps the CPU interface via the OrtGlobals member accessor),
-  // so acquiring this lock here cannot deadlock. The lock is released before the returned
-  // reference is dereferenced by callers, so it does not nest with device_interfaces_mutex_.
-  static std::mutex creation_mutex;
-  std::scoped_lock lock{creation_mutex};
+  // The OrtGlobals constructor does not re-enter GetOrtGlobals() (it bootstraps the CPU interface
+  // via the OrtGlobals member accessor), so acquiring this lock here cannot deadlock. The lock is
+  // released before the returned reference is dereferenced by callers, so it does not nest with
+  // device_interfaces_mutex_.
+  std::scoped_lock lock{g_ort_globals_mutex};
   if (!g_ort_globals && !g_process_exiting)
     g_ort_globals = std::make_unique<OrtGlobals>();
 
@@ -137,9 +139,12 @@ void Shutdown() {
     std::cerr << "    Please see the documentation for the API being used to ensure proper cleanup." << std::endl;
   }
 
-  // Delete now because on process exit is too late. ~OrtGlobals tears down the device interfaces
-  // (including the RyzenAI EP shutdown) and unloads the genai add-on libraries.
-  GetOrtGlobals().reset();
+  // Reset g_ort_globals directly (rather than through GetOrtGlobals(), which would lazily construct
+  // the globals just to immediately tear them down). If genai was never initialized there is nothing
+  // to do. Delete now because on process exit is too late. ~OrtGlobals tears down the device
+  // interfaces (including the RyzenAI EP shutdown) and unloads the genai add-on libraries.
+  std::scoped_lock lock{g_ort_globals_mutex};
+  g_ort_globals.reset();
 }
 
 OrtEnv& GetOrtEnv() {
--- a/src/generators.cpp
+++ b/src/generators.cpp
@@ -89,8 +89,18 @@ static bool g_process_exiting = false;
 // Ensure Shutdown() has been called before process exit
 struct EnsureShutdown {
   ~EnsureShutdown() {
-    g_process_exiting = true;
-    if (g_ort_globals)
+    // Set the process-exit flag under the mutex so GetOrtGlobals() (which reads it under the same
+    // lock) can never race with this write. Capture whether teardown is needed while holding the
+    // lock, then call Shutdown() outside it -- Shutdown() re-acquires the mutex, so holding it here
+    // would self-deadlock (the mutex is non-recursive).
+    bool needs_shutdown = false;
+    {
+      std::scoped_lock lock{g_ort_globals_mutex};
+      g_process_exiting = true;
+      needs_shutdown = static_cast<bool>(g_ort_globals);
+    }
+
+    if (needs_shutdown)
       Shutdown();
   }
 };
--- a/src/generators.cpp
+++ b/src/generators.cpp
@@ -70,24 +70,42 @@ OrtGlobals::OrtGlobals()
   Ort::Allocator& allocator_cpu{Ort::Allocator::GetWithDefaultOptions()};
   env_->CreateAndRegisterAllocator(allocator_cpu.GetInfo(), *arena_config);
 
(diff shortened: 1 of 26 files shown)


src/ryzenai/interface.cpp
skottmckay, line 1: @akholodnamdcom would you be able to validate the RyzenAI changes?

test/ep_registration.h
```diff
@@ -0,0 +1,175 @@
...
+                  << f.path.string() << std::endl;
+        OgaRegisterExecutionProviderLibrary(f.registration_name.c_str(), f.path.string().c_str());
+        registered_handles_.push_back(f.registration_name);
+        if (!f.provider_name.empty())
+          providers_.push_back(f.provider_name);
+      } catch (const std::exception& e) {
+        std::cerr << "Warning: failed to register execution provider library '" << f.registration_name
+                  << "': " << e.what() << std::endl;
+      }
+    }
```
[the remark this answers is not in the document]
skottmckay: EPs can be registered directly with the OrtEnv and GenAI doesn't control that path.

src/qnn/interface.cpp
```diff
@@ -108,16 +112,12 @@ struct GpuInterfaceImpl : QnnInterfaceBase {
 
 }  // namespace QNN
 
-DeviceInterface* GetQNNInterface(DeviceType device_type) {
-  assert(type == DeviceType::QnnHtp || type == DeviceType::QnnGpu);
```
kunal-vaishnavi, line 112: This assert is getting fixed in [this PR](https://github.com/microsoft/onnxruntime-genai/pull/2259). It seems that Qualcomm would like to keep the assert so can we put the same fix in this PR?

src/ryzenai/interface.cpp
```diff
@@ -117,15 +117,32 @@ struct Interface : RyzenAIInterface {
       // fallback to current working directory
       ep_path_ = std::filesystem::current_path(ec) / ep_filename_;
 
-    Ort::ThrowOnError(Ort::api->RegisterExecutionProviderLibrary(GetOrtGlobals()->env_.get(), ep_name_, ep_path_.native().c_str()));
+    // Register on this env, tolerating the case where it is already registered. An OrtEnv can outlive
+    // a genai OgaShutdown (the host may hold a reference), so on re-init the same env may already have
+    // the library; ORT reports that as "library is already registered under ...", which is benign here.
+    if (auto status = std::unique_ptr<OrtStatus>{
+            Ort::api->RegisterExecutionProviderLibrary(&env_, ep_name_, ep_path_.native().c_str())}) {
```
kunal-vaishnavi: Can we use a method from `onnxruntime_inline.h` instead of calling `Ort::api` APIs directly?

https://github.com/microsoft/onnxruntime-genai/blob/2ef64f99339fc6f21831827c24f4dc86206699d6/src/models/onnxruntime_inline.h#L188-L194

src/ryzenai/interface.cpp
```diff
@@ -136,7 +153,7 @@ struct Interface : RyzenAIInterface {
       const OrtEpDevice* const* devices = nullptr;
       size_t ndevices = 0;
 
-      Ort::ThrowOnError(Ort::api->GetEpDevices(&GetOrtEnv(), &devices, &ndevices));
+      Ort::ThrowOnError(Ort::api->GetEpDevices(&env_, &devices, &ndevices));
```
kunal-vaishnavi: Same comment here

https://github.com/microsoft/onnxruntime-genai/blob/2ef64f99339fc6f21831827c24f4dc86206699d6/src/models/onnxruntime_inline.h#L138-L140

src/ryzenai/interface.cpp
```diff
@@ -157,7 +174,8 @@ struct Interface : RyzenAIInterface {
 
       // this call merges provider_options into session_options
       Ort::ThrowOnError(Ort::api->SessionOptionsAppendExecutionProvider_V2(&session_options,
```
kunal-vaishnavi: Same comment here

https://github.com/microsoft/onnxruntime-genai/blob/2ef64f99339fc6f21831827c24f4dc86206699d6/src/models/onnxruntime_inline.h#L782-L793

src/generators.cpp
```diff
@@ -212,30 +245,106 @@ struct LibraryHandle {
...
-
-    return cuda_interface;
+    return reinterpret_cast<decltype(&GetInterface)>(
+        cuda_library_->GetSymbol("GetInterface"))(&g_genai, to_string(type).c_str());
   } catch (const std::exception& e) {
     throw std::runtime_error("Cuda interface not available: " + std::string(e.what()));
   }
 }
 
+DeviceInterface* OrtGlobals::GetDeviceInterface(DeviceType type) {
```
kunal-vaishnavi, line 299: Should this be `std::unique_ptr<DeviceInterface>` to follow the consistency used in this PR?

src/generators.cpp
```diff
@@ -261,27 +370,7 @@ std::string to_string(DeviceType device_type) {
 }
 
 DeviceInterface* GetDeviceInterface(DeviceType type) {
```
kunal-vaishnavi, line 372: Can we remove the wrapper method? Seems like it would be clearer to the reader if they are forced to write `GetOrtGlobals()->GetDeviceInterface(type)` directly.

src/generators.cpp
```diff
@@ -212,30 +245,106 @@ struct LibraryHandle {
 };
 #endif
 
-DeviceInterface* GetCudaInterface(DeviceType type) {
+OrtGlobals::~OrtGlobals() {
```
kunal-vaishnavi, line 248: From Copilot:

Can we consolidate all teardown paths into a single canonical reset helper?  
Right now, `OgaShutdown` and init-failure cleanup can drift over time. A shared reset routine would ensure we always return to the same “just loaded” baseline and avoid partial-state regressions.

src/generators.cpp
```diff
@@ -70,24 +70,54 @@ OrtGlobals::OrtGlobals()
...
-std::unique_ptr<OrtGlobals>&
-GetOrtGlobals() {
-  static auto globals = std::make_unique<OrtGlobals>();
-  static auto validate = std::make_unique<EnsureShutdown>();  // Must be after the above line so the destructor runs before the above destructor
-  return globals;
+std::unique_ptr<OrtGlobals>& GetOrtGlobals() {
+  // Registered once; its destructor runs at process exit (before g_ort_globals is destroyed) and
+  // performs the final Shutdown().
+  static EnsureShutdown ensure_shutdown;
+  // The OrtGlobals constructor does not re-enter GetOrtGlobals() (it bootstraps the CPU interface
```
kunal-vaishnavi, line 112: From Copilot:

Could we add/confirm synchronization around lifecycle transitions (`loaded → initialized → shutdown`)?  
Given re-init after shutdown is now supported, concurrent init/shutdown calls could race and expose transient global state. A single lock/guard around transitions would make this deterministic.

src/ort_genai_c.cpp
```diff
@@ -1145,9 +1146,31 @@ void OGA_API_CALL OgaDestroyRuntimeSettings(OgaRuntimeSettings* p) { delete stat
...
+// translation unit). The exception propagates and is caught by callers normally.
+#pragma warning(disable : 4297)
+#endif
 void OGA_API_CALL OgaRegisterExecutionProviderLibrary(const char* registration_name, const char* library_path) {
+  // Pre-flight the provider library before registering it. A failed ORT registration (e.g. a
+  // hardware EP whose native runtime cannot initialize because the GPU was removed or disabled) can
+  // leave ONNX Runtime with a half-registered library that crashes at environment teardown. By
+  // verifying the library and its native dependencies load first, a broken EP fails cleanly here
+  // with a descriptive error instead of destabilizing the process (see CanLoadLibrary).
+  if (!CanLoadLibrary(library_path))
```
kunal-vaishnavi, line 1163: From Copilot:

On EP/provider load failure, can we explicitly guarantee no partially-initialized objects remain reachable from globals? It would be great to confirm env/provider/diagnostic state is fully cleared before returning error, so the next init attempt starts from a truly clean state.

src/generators.cpp
skottmckay: I wouldn't have said so for the getter. User doesn't need to care if it's coming from a unique_ptr or not. 

The 'create' functions consistently return a unique_ptr but that's an ownership transfer.

src/generators.cpp
skottmckay: I agree but that change will touch 59 call sites in 24 files. Maybe better as a follow up PR?

src/generators.cpp
skottmckay: Isn't this all wrapped up by OgaGlobals and all OgaShutdown is doing is releasing that instance?

src/generators.cpp
skottmckay: We use `g_ort_globals_mutex` to synchronize. Is there something missing from that usage?

src/ort_genai_c.cpp
skottmckay: What partially-initialized objects exist? Device interfaces are lazily created when required by a model but that requires successful EP load.

src/generators.cpp
kunal-vaishnavi: I believe so too. I think Copilot was reviewing an out-of-date version of the PR.

src/generators.cpp
kunal-vaishnavi: I don't think so. I think Copilot was reviewing an out-of-date version of the PR.

src/ort_genai_c.cpp
kunal-vaishnavi: None should exist. I think Copilot was reviewing an out-of-date version of the PR.

kunal-vaishnavi (dismissed):
I think we can merge this in and if any changes from this PR break any of the IHV EPs, then we can open PR fixes for them accordingly.

kunal-vaishnavi (approved):

PR #2275, opened 2026-07-06
Add int8 precision support to model builder

danielsongmicrosoft:
The model builder currently supports the following precisions for weight quantization: int4, bf16, fp16, fp32

This change adds support for int8.

PR #2299, opened 2026-07-16
docs: unified model builder quantization config design

tianleiwu:
## Summary

Adds a design proposal (`src/python/py/models/quantization_config_design.md`) for a
*unified* quantization configuration for the ONNX Runtime GenAI model builder. It is
design-only — no builder code is changed. The document defines a single quant-dtype
vocabulary and a structured, target-oriented `quantization` JSON config to replace the
organically-grown, entangled `int4_*` / `--precision` / `moe_quant_type` options.

## Motivation

The current quantization surface is inconsistent and hard to extend:

- `int4_`-prefixed options (`int4_block_size`, `int4_is_symmetric`, …) actually apply to
  any bit-width, so the prefix is misleading once int8/fp8/mxfp4 land.
- `--precision` conflates model I/O (activation) dtype with weight quant type — there is
  no clean way to say "int8 weights with bf16 I/O".
- Quant type is encoded three different ways (a `--precision` string, a bare `{"bits": N}`,
  and quant-type names), none of which express fp8 variants, mxfp4, or nvfp4 uniformly.
- Per-node control is limited, there is no KV-cache quantization surface, and
  `extra_options` is flat `KEY=VALUE` only (losing Olive/olive-recipes per-node overrides).

## Key Changes

| Area | Content |
| --- | --- |
| Analysis | Documents today's CLI, MatMul/QMoE `extra_options`, and how PR #2275 (int8 precision) fits in. |
| Unified dtype vocabulary | One descriptor table covering `fp32`/`fp16`/`bf16`, `int8`/`int4` (+ future `int2`/`int6`), `fp8_e4m3`/`fp8_e5m2`, `fp4_e2m1`, `mxfp4`, `mxfp8_*`, and `nvfp4`. |
| Structured `quantization` config | Target-oriented JSON: `io_dtype`, `weights`, `embeddings`, `lm_head`, `moe`, `kv_cache` (future), `runtime`, with ordered per-node/per-layer `overrides`. |
| I/O decoupling | `io_dtype` set independently of `weights.type`, fixing the `--precision` conflation. |
| Auxiliary models / MTP | `auxiliary_models` (e.g. Qwen3.6 `mtp.onnx`) inherit resolved parent targets and can override per-target quant policy, with source-namespace resolution rules. |
| Back-compat | Existing flat options desugar into the new config as thin aliases so current recipes/CLIs produce byte-identical models. |

## Testing Notes

Documentation-only change. To review, read
`src/python/py/models/quantization_config_design.md` and verify the proposed option
surface against the current behavior described in `quantization.md` and the referenced
PR #2275. No code paths are affected and no builds/tests are required.

### DIFF
--- a/src/python/py/models/quantization_config_design.md
+++ b/src/python/py/models/quantization_config_design.md
@@ -163,7 +163,7 @@ empty object means "no quantization, I/O follows `io_dtype`".
     "lm_head":    { "type": "int8" },                        // role shortcut
 
     "moe": {                        // MoE expert weights (QMoE)
-      "type": "mxfp4",             // int4 | int8 | mxfp4 | nvfp4 | fp8_* ...
+      "type": "mxfp4",             // any dtype from §4, or "none" for dense MoE
       "block_size": 32,
       "weights_prepacked": -1       // CUDA QMoE layout (-1 auto | 0 | 1)
     },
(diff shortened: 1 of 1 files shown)


src/python/py/models/quantization_config_design.md
```diff
@@ -0,0 +1,520 @@
...
+- **Extensibility.** Adding a dtype = adding one row to a descriptor table; no new option
+  name is introduced. This is the property today's `matmul_mixed_precision` already aims
+  for by using type *names* rather than bit counts.
+
+## 5. The `quantization` config
+
+A single object, `quantization`, is the primary surface. Every field is optional; an
+empty object means "no quantization, I/O follows `io_dtype`".
+
+```jsonc
```
kunal-vaishnavi, line 145: Here are some ideas that I am thinking of. I wonder if any of them make sense to implement.

1. We build a light, optional, QA-based utility that walks a user through the types of quantization settings supported and users can respond to the questions with selections. Then, this config file is automatically generated for them so that they don't need to know custom settings. For users who want the most granular control, they can go through the utility to construct a simple version and customize it manually or they can manually construct the quant config themselves.
2. We could use a similar approach that AMD's Quark uses for creating a quant config. Here are some resources for reference.
- [Full list of quantization configuration features](https://quark.docs.amd.com/latest/onnx/appendix_full_quant_config_features.html)
- [Quantization config details](https://quark.docs.amd.com/latest/onnx/config/quantization_configuration_serialization_and_deserialization.html)
3. For the config itself, what if we mirror it to be like the GenAI config for the parent sections? Then the JSON file containing a quant config can describe how it's done for multiple ONNX models.
```jsonrc
{
    "vision": {
        ...
    },
    "embedding": {
        ...
    },
    "decoder": {
        "io_dtype": "fp16",
        "runtime": {
            "use_qdq": false,
            "matmulnbits_weights_prepacked": 0,
            "qmoe_weights_prepacked": 0,
        },
        "embeddings": {
            ...
        },
        "attn": {
            ...
        },
        "mlp": {
            ...
        },
        "moe": {
            ...
        },
        "lm_head": {
            ...
        }
    },
    "mtp": {
        ...
    }
}
```

PR #2301, opened 2026-07-16
Add Nemotron Parse support

ttyio:
Enabled Nemotron Parse support, with some optimization: export tensorScatter for inplace kv cache, export Q/DQ for INT4 WoQ instead of the contrib op MatMulNBits. Validate on Spark improved steady decode from 3.919 to 1.589 ms/step and the 128-token decode loop from 498.499 to 202.277 ms on RTX backend.


ttyio:
@microsoft-github-policy-service agree company="Nvidia"

src/python/py/models/builders/nemotron_parse.py
```diff
@@ -0,0 +1,859 @@
...
+            if hasattr(target, "target_height"):
+                target.target_height = self.image_height
+            if hasattr(target, "target_width"):
+                target.target_width = self.image_width
+            if hasattr(target, "final_size"):
+                target.final_size = [self.image_height, self.image_width]
+            if hasattr(target, "size"):
+                target.size = {"height": self.image_height, "width": self.image_width}
+        print(f"Saving processing files in {out_dir} for Nemotron Parse")
+        processor.save_pretrained(out_dir)
```
yen-shi: When I tried to run this PR with claude, it says that `auto_map` field is not propagated correctly, so I have to apply a patch to run.

The command I used to export the model:
```
python3 builder.py -i <hf_model_dir> -o <out> -p fp16 -e NvTensorRtRtx \
  --extra_options decoder_cache_mode=tensor_scatter torch_dtype=fp16 hf_remote=True export_device=cuda
```

The patch
```
diff --git a/src/python/py/models/builders/nemotron_parse.py b/src/python/py/models/builders/nemotron_parse.py
index 002eafc8..1a439b08 100644
--- a/src/python/py/models/builders/nemotron_parse.py
+++ b/src/python/py/models/builders/nemotron_parse.py
@@ -857,3 +857,34 @@ class NemotronParseModel:
                 target.size = {"height": self.image_height, "width": self.image_width}
         print(f"Saving processing files in {out_dir} for Nemotron Parse")
         processor.save_pretrained(out_dir)
+
+        # transformers' save_pretrained does not re-emit the `auto_map` (nor the
+        # do_rescale/rescale_factor) keys for this remote-code image processor, so
+        # AutoProcessor.from_pretrained(out_dir, trust_remote_code=True) would fail
+        # with "Unrecognized image processor". Re-attach them from the source
+        # config so the exported package stays self-contained and reloadable.
+        try:
+            from transformers.utils import cached_file
+
+            src_config = cached_file(
+                model_name_or_path,
+                "preprocessor_config.json",
+                token=self.hf_token,
+                _raise_exceptions_for_missing_entries=False,
+            )
+            out_config = os.path.join(out_dir, "preprocessor_config.json")
+            if src_config and os.path.isfile(out_config):
+                with open(src_config) as src_file:
+                    source = json.load(src_file)
+                with open(out_config) as out_file:
+                    exported = json.load(out_file)
+                restored = False
+                for key in ("auto_map", "do_rescale", "rescale_factor"):
+                    if key in source and key not in exported:
+                        exported[key] = source[key]
+                        restored = True
+                if restored:
+                    with open(out_config, "w") as out_file:
+                        json.dump(exported, out_file, indent=2)
+        except Exception as exc:  # never let this optional fix-up fail the export
+            print(f"Warning: could not preserve preprocessor auto_map: {exc}")
```

If I don't add `auto_map` fix, the export succeeds but inference of examples/python/nemotron_parse.py failed with
```
  File ".../examples/python/nemotron_parse.py", line 109, in run_parse
    processor = AutoProcessor.from_pretrained(model_dir, trust_remote_code=True)
  File ".../transformers/models/auto/processing_auto.py", line 324, in from_pretrained
    return processor_class.from_pretrained(...)
  File ".../hf_nemotron_parse_processor.py", line 382, in from_pretrained
    image_processor = AutoImageProcessor.from_pretrained(...)
  File ".../transformers/models/auto/image_processing_auto.py", line 692, in from_pretrained
    raise ValueError(
ValueError: Unrecognized image processor in <export_dir>. Should have a `image_processor_type` key
in its preprocessor_config.json of config.json, or one of the following `model_type` keys in its
config.json: aria, beit, bit, blip, bridgetower, chameleon, chinese_clip, clip, ... [long built-in list]
```

The inference command:
```
python ~/oga_pr2279/examples/python/nemotron_parse.py \
  --model_path /home/nvidia/spark_bench/models/nemotron_parse_fp16_trtrtx \
  --image_file /home/nvidia/asr_ex2/equations.png \
  -e NvTensorRTRTXExecutionProvider --ep_path "$LIBS/ep/libonnxruntime_providers_nv_tensorrt_rtx.so" \
  --max_new_tokens 128
```

ttyio:

src/python/py/models/builders/nemotron_parse.py
ttyio: Thank you @yen-shi , updated the code

yen-shi:
Hi @kunal-vaishnavi,
Can you review this MR or add relevant reviewers for Nemotron Parse support with NVTensorRTRtx EP? Thanks!

src/models/nemotron_parse.h
```diff
@@ -0,0 +1,27 @@
...
+                                     const GeneratorParams& params) const override;
+
+  std::unique_ptr<OrtSession> encoder_session_;
+  std::unique_ptr<OrtSession> prefill_session_;
+  std::unique_ptr<OrtSession> decoder_session_;
+
+  std::unique_ptr<OrtSessionOptions> encoder_session_options_;
+  std::unique_ptr<OrtSessionOptions> prefill_session_options_;
+
+  SessionInfo encoder_session_info_;
```
kunal-vaishnavi: It looks like this model now stores per-session metadata (`encoder_session_info_`, `prefill_session_info_`, `decoder_session_info_`) instead of relying on the base `Model::session_info_` pattern used in places like `whisper.h`. Can we change it to the Whisper style instead of needing separate `SessionInfo` objects?

src/python/py/models/builders/nemotron_parse.py
```diff
@@ -0,0 +1,859 @@
+# Copyright (c) Microsoft Corporation.  All rights reserved.
```
kunal-vaishnavi: This file is using the exporter + standard `onnx` helper APIs rather than using the model builder infra. Please rewrite this file to integrate with the existing infra. You can see `whisper.py` as an example for adding multi-model support.

Since you already have the ONNX graphs after exporting, you can point the graphs to AI and have it generate the node structures and code conventions needed. You can also provide it with screenshots of the graph structures for better analysis.

examples/python/nemotron_parse.py
```diff
@@ -0,0 +1,197 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
```
kunal-vaishnavi: Nemotron Parse is a vision-language model. Please integrate any runtime changes needed inside the various examples in C/C++/C#/Python that are labeled as `model-mm` or `model_mm`. We should not need a new standalone example as the `mm` examples are for multi-modal models.

src/models/model.cpp
```diff
@@ -891,6 +892,8 @@ std::shared_ptr<Model> CreateModel(OrtEnv& ort_env, std::unique_ptr<Config> conf
     return std::make_shared<NemotronSpeechModel>(std::move(config), ort_env);
   if (ModelType::IsTDT(config->model.type))
     return std::make_shared<ParakeetTdtModel>(std::move(config), ort_env);
+  if (config->model.type == "nemotron_parse")
```
kunal-vaishnavi: We should find a way to route Nemotron Parse inside `ModelType::IsVLM` and not need another case.

src/models/nemotron_parse.cpp
```diff
@@ -0,0 +1,809 @@
...
+#include "../generators.h"
+#include "nemotron_parse.h"
+#include "input_ids.h"
+#include "kv_cache.h"
+#include "logits.h"
+
+namespace Generators {
+namespace {
+
+constexpr const char* kNvProfileMinShapes =
```
kunal-vaishnavi, line 14: Anything that is EP-specific such as these session options should go inside the EP's session options and be activated that way. We should not be putting them where the model is implemented.

src/models/nemotron_parse.cpp
```diff
@@ -0,0 +1,809 @@
...
+  for (int64_t dimension : dimensions) {
+    if (!first_dimension) {
+      profile << 'x';
+    }
+    first_dimension = false;
+    profile << dimension;
+  }
+}
+
+void SetFixedNvTensorRtRtxProfile(OrtSessionOptions& options,
```
kunal-vaishnavi: The same thing applies to the profile logic being moved to the EP-specific folder.

src/models/nemotron_parse.cpp
```diff
@@ -0,0 +1,809 @@
...
+                        const std::string& name) {
+  if (actual != expected) {
+    throw std::runtime_error(name + " has type " + TypeToString(actual) +
+                             ", expected " + TypeToString(expected));
+  }
+}
+
+class DynamicAttentionMask {
+ public:
+  DynamicAttentionMask(State& state, const SessionInfo& session_info)
```
kunal-vaishnavi: Why aren't the existing attention mask options inside `position_inputs.cpp` sufficient? There already exists support for both static and dynamic attention masks. Let's re-use existing logic instead of rewriting from scratch.

src/models/nemotron_parse.cpp
```diff
@@ -0,0 +1,809 @@
...
+  }
+
+ private:
+  State& state_;
+  std::string name_;
+  ONNXTensorElementDataType type_;
+  Tensor value_;
+};
+
+class TensorScatterKeyValueCache {
```
kunal-vaishnavi: Any new KV cache classes should be added to `kv_cache.cpp` for generic applicability across other models. Their usage should be documented for clear future use.

src/python/py/models/builder.py
```diff
@@ -614,6 +625,19 @@ def get_args():
                     Use this option to create quantized ONNX models that use BF16 precision.
                 adapter_path = Path to folder on disk containing the adapter files (adapter_config.json and adapter model weights).
                     Use this option for LoRA models.
+                image_height / image_width = Fixed image size used by Nemotron Parse ONNX export.
```
kunal-vaishnavi: Do we need all of these extra options? We can just keep default values in the creation of Nemotron Parse. ORT GenAI will anyways best support the options that these are set by default to.

src/config.h
```diff
@@ -316,9 +316,13 @@ struct Config {
 
     struct Decoder {
       std::string filename;
+      std::string prefill_filename;
       SessionOptions session_options;
       std::optional<RunOptions> run_options;
 
+      std::string cache_update_mode;
+      int prefill_sequence_length{};
```
kunal-vaishnavi, line 359: Why do we need a reference to the prefill graph and its sequence length inside the decoder section of the GenAI config?

src/config.h
```diff
@@ -378,6 +383,7 @@ struct Config {
         std::string present_key_names{Defaults::PresentKeyName};
         std::string present_value_names{Defaults::PresentValueName};
         std::string present_names;  // When key/value pairs are combined
+        std::string cross_present_key_names, cross_present_value_names;
```
kunal-vaishnavi, line 452: Can you share what the graphs look like and why this is needed? Typically, cross-attention KV cache outputs are pre-computations that only need to be calculated once. As a result, commonly they make more sense to compute during prefill inside an encoder model and not during decoding inside a decoder model. That way, you don't have to re-compute the same outputs after each decoding step.

You can take a look at how Whisper is implemented [here](https://github.com/microsoft/onnxruntime-genai/blob/0be0a44cd45833e03219be9d84da7ab03b1a7b31/src/python/py/models/builders/whisper.py#L143) with cross-attention KV cache outputs in its encoder for reference.

src/models/nemotron_parse.h
ttyio: Updated, now the three model-specific SessionInfo fields were removed. Thanks!

src/python/py/models/builders/nemotron_parse.py
ttyio: Updated. Prefill and decode now use `NemotronParseDecoderComponent`, derived from the shared builder Model and ONNX IR infrastructure. The RADIO encoder remains a thin PyTorch export because there is no corresponding shared encoder component. Thanks!

examples/python/nemotron_parse.py
ttyio: Updated, The standalone examples/python/nemotron_parse.py was removed, thanks!

src/models/model.cpp
ttyio: Updated. Nemotron Parse is now registered in `ModelType::IsVLM`. Thanks!

src/models/nemotron_parse.cpp
ttyio: Updated, now the generic TRT-RTX provider setup now live in `src/nvtensorrtrtx/session_options.cpp`. However, Nemotron Parse’s fixed profile construction is still in `src/models/nemotron_parse.cpp`, including `MakePrefillProfile`, `MakeDecodeProfile`, `SetFixedProfile`. This is because Nemotron Parse creates separate encoder, prefill, and decode sessions, while the existing EP session-options entry point receives the overall model configuration but no explicit indication of which graph/session is being configured.

src/models/nemotron_parse.cpp
ttyio: Updated same as above reason. Thanks!

src/models/nemotron_parse.cpp
ttyio: Updated. Both phases now use `DefaultPositionInputs`, thanks!

src/models/nemotron_parse.cpp
ttyio: Updated. Now moved the classes to `kv_cache.cpp`, thanks!

src/python/py/models/builder.py
ttyio: Updated,  Now the `cache-update mode` was removed,  The remaining options control static export dimensions or partial component export. Thanks!

src/config.h
ttyio: Nemotron Parse exports three graphs:

1. `encoder.onnx` converts the image into visual features.
2. `decoder_prefill.onnx` runs the mBART decoder over the initial prompt and produces the first logits plus the initial self-attention and cross-attention KV caches.
3. `decoder.onnx` runs the same decoder architecture one token at a time using those caches.

Because `decoder_prefill.onnx` and `decoder.onnx` are two execution forms of the text decoder, their metadata is currently grouped under the decoder section.
The runtime needs `prefill_filename` to locate and create the separate decoder-prefill session. It needs `prefill_sequence_length` before creating that TRT-RTX session because the prefill graph has a fixed input shape and the optimization profile must be configured during session creation. The value cannot be obtained from the created session afterward because the profile is required to create that session in the first place.

Thanks!

src/config.h
ttyio: The decode graph does not recompute cross-attention KV caches. `cross_present_key_names` and `cross_present_value_names` identify outputs of `decoder_prefill.onnx`, which runs once. Those tensors are retained by the runtime and bound to the `cross_past_key_names` and `cross_past_value_names` inputs of `decoder.onnx`. The per-token decode graph outputs only logits and updated self-attention KV caches; it has no cross-present outputs. The fields are currently under the shared decoder configuration because they describe the handoff between its prefill and decode phases. Thanks!

yen-shi:
Hi @kunal-vaishnavi,
Can you review again when available? Thanks!

ttyio:
@kunal-vaishnavi , thank you for your review and valuable suggestions, follow up our offline discussion, we made following changes: 
- remove toch.onnx.export
- use 2 onnx model with 3 profile.
- the perf still 1.520 ms / token for decoder on Spark.

src/models/nemotron_parse.cpp
```diff
@@ -0,0 +1,370 @@
...
+    cross_cache.AddInputs(*this);
+  }
+
+  DeviceSpan<float> Run(int total_length, DeviceSpan<int32_t>& next_tokens,
+                        DeviceSpan<int32_t>) override {
+    const size_t new_length = next_tokens.size() / params_->BatchBeamSize();
+    const bool is_prompt = first_run_;
+    if (is_prompt) {
+      if (total_length !=
+              model_.config_->model.decoder.prefill_sequence_length ||
```
baijumeswani: The processor accepts any prompt length, but this check requires exactly `prefill_sequence_length` tokens. The default task happens to be eight tokens, while most custom prompts will fail. A package exported with another documented prefill length, such as 16, will also reject the default task.

Could we allow dynamic prompt lengths on providers that support them, and pad plus mask shorter prompts for a fixed TRT-RTX profile? At minimum, please reject unsupported prompts in the processor with a clear message and avoid presenting `--user_prompt` as unrestricted text. Tests should cover the default prompt, short and long custom prompts, and a non-default prefill setting.

src/models/nemotron_parse.cpp
```diff
@@ -0,0 +1,370 @@
...
+  DeviceSpan<float> Run(int total_length, DeviceSpan<int32_t>& next_tokens,
+                        DeviceSpan<int32_t> next_indices) override {
+    if (encoder_state_) {
+      encoder_state_->RunEncoder();
+      encoder_state_.reset();
+    }
+    return decoder_state_.Run(total_length, next_tokens, next_indices);
+  }
+
+  void RewindTo(size_t) override {
```
baijumeswani, line 279: `Generator::RewindToLength` rewinds the search tokens before it calls this method. Throwing here therefore leaves the search sequence shortened while the cache and attention mask remain at the old length.

Please reject Nemotron Parse at the start of `Generator::RewindToLength`, before any state changes. A general `SupportsRewind()` check would be cleaner than adding more model-name checks.

src/models/io/tensor_scatter_kv_cache.cpp
```diff
@@ -0,0 +1,149 @@
...
+    std::fill(cpu_values.begin(), cpu_values.end(), write_index);
+    values.CopyCpuToDevice();
+  } else {
+    auto values = cache_write_indices_->GetDeviceSpan<int64_t>();
+    auto cpu_values = values.CpuSpan();
+    std::fill(cpu_values.begin(), cpu_values.end(),
+              static_cast<int64_t>(write_index));
+    values.CopyCpuToDevice();
+  }
+  current_length_ = total_length;
```
baijumeswani, line 144: This marks the new length as complete before the ONNX run succeeds. If the run fails or is stopped, retrying at the same length is rejected, even though the cache may be incomplete. The decoder's first-run flag is also changed before execution.

Could we prepare the write index first, run the model, and commit the new cache length only after success? Since past and present share memory, a stopped run may partially update the cache; if recovery cannot be guaranteed, the generator should be marked unusable after that error.

src/models/nemotron_parse.cpp
```diff
@@ -0,0 +1,370 @@
...
+
+  void SetExtraInputs(const std::vector<ExtraInput>& extra_inputs) override {
+    if (!encoder_state_) {
+      throw std::runtime_error(
+          "Nemotron Parse inputs cannot be changed after prompt processing");
+    }
+    encoder_state_->SetExtraInputs(extra_inputs);
+  }
+
+  DeviceSpan<float> Run(int total_length, DeviceSpan<int32_t>& next_tokens,
```
baijumeswani, line 260: This wrapper state owns no session that actually runs. Runtime options are currently applied to the wrapper's `OrtRunOptions`, not to `encoder_state_` or `decoder_state_`. As a result, options such as `terminate_session` and run-level profiling will not affect the active inference call.

Please override the runtime-option handling and forward options to the active inner states, following the composite-state pattern used by speculative decoding. Input/output inspection and adapter handling should also be forwarded or explicitly rejected.

src/python/py/models/builders/nemotron_parse.py
```diff
@@ -0,0 +1,340 @@
...
+            )
+        return options
+
+    def _torch_dtype(self):
+        dtype = self.extra_options.get("torch_dtype")
+        if dtype == "fp32":
+            return torch.float32
+        if dtype == "bf16":
+            return torch.bfloat16
+        return torch.float16
```
baijumeswani: This defaults every model load to FP16, even when the requested export is BF16 or FP32. Those exports then contain values that were already rounded to FP16. INT4 scales are also calculated from the rounded weights.

Please default to the requested precision, the checkpoint dtype, or `torch_dtype="auto"`. If FP16 loading is needed to reduce memory, it should be an explicit option. A small test can verify the dtype passed to `AutoModel.from_pretrained` for FP16, BF16, FP32, and INT4.

src/python/py/models/builders/nemotron_parse.py
```diff
@@ -0,0 +1,340 @@
...
+            prefill_sequence_length=self.prefill_sequence_length,
+            cache_sequence_length=self.cache_sequence_length,
+        )
+
+    def make_model(self, input_path):
+        self._model = self._load_model(input_path)
+
+    def save_model(self, output_dir):
+        try:
+            if "encoder" in self.export_components:
```
baijumeswani, line 211: `export_components` allows only the encoder or decoder to be written, but `make_genai_config()` always references both `encoder.onnx` and `decoder.onnx`. A component-only export into a clean directory therefore produces a package that cannot load.

Please either skip package configuration for component-only output, require the other graph to already exist, or remove component-only export from the supported public contract.

test/python/builder/test_nemotron_parse.py
```diff
@@ -0,0 +1,908 @@
...
+            )
+        )
+        hidden_states = hidden_states + feed_forward
+        caches.append((self_key, self_value, cross_key, cross_value))
+
+    hidden_states = decoder.layer_norm(hidden_states)
+    return model.lm_head(hidden_states), caches
+
+
+class NemotronParseBuilderTests(TestCase):
```
baijumeswani, line 432: These tests contain important numerical coverage, but the current CPU and GPU workflows appear to run named test files rather than discover `test/python/builder`. That means this suite may not gate the PR.

Please add this directory to required CI and make sure ONNX Runtime is installed so the numerical tests do not skip. General builder-test discovery would also prevent future tests from being checked in without being run.

src/models/preprocessing/nemotron_parse_processor.cpp
```diff
@@ -0,0 +1,230 @@
...
+  }
+  if (payload.audios) {
+    throw std::runtime_error("Nemotron Parse does not accept audio input");
+  }
+
+  Ort::Allocator& allocator{Ort::Allocator::GetWithDefaultOptions()};
+  auto named_tensors = std::make_unique<NamedTensors>();
+  named_tensors->emplace(
+      std::string(Config::Defaults::InputIdsName),
+      std::make_shared<Tensor>(BuildInputIds(
```
baijumeswani, line 232: The multimodal processor's list overload stores text in `payload.prompts`, but this code reads only `payload.prompt`. A one-item prompt list is therefore silently ignored and the default task is used.

Please accept exactly one list item or reject the list overload clearly. Tests should compare string input, a one-item list, an empty list, and a multi-item list.

src/models/nemotron_parse.cpp
```diff
@@ -0,0 +1,370 @@
...
+                                 /*disable_graph_capture=*/true);
+
+  if (p_device_->GetType() == DeviceType::NvTensorRtRtx) {
+    // Reuse one ONNX file, but create independently optimized prefill and
+    // decode sessions. A single dynamic engine is materially slower at Q=1.
+    prefill_decoder_session_options_ = OrtSessionOptions::Create();
+    CreateSessionOptionsFromConfig(
+        decoder.session_options, *prefill_decoder_session_options_,
+        /*is_primary_session_options=*/false);
+    SpecializeDecoderSession(*session_options_, *config_, 1);
```
baijumeswani: These are two separate decoder sessions, each with one fixed profile. Generic `State::Run` can still select profile 1 after the first call when `nv_multi_profile_enable=1`, but the one-token session appears to have only profile 0.

Please reject multi-profile mode for this model or disable the generic profile-index switch for these sessions. Tracking first-run/profile state per session would be the more general solution.

src/models/nemotron_parse.cpp
```diff
@@ -0,0 +1,370 @@
...
+class NemotronParseState : public State {
+ public:
+  NemotronParseState(const NemotronParseModel& model,
+                     DeviceSpan<int32_t> sequence_lengths,
+                     const GeneratorParams& params)
+      : State{params, model},
+        model_{model},
+        encoder_state_{std::make_unique<EncoderState>(model, params)},
+        decoder_state_{model, sequence_lengths, params} {
+    if (params_->search.batch_size != 1 || params_->search.num_beams != 1) {
```
baijumeswani: The decoder state, including the full KV cache, is constructed before this batch and beam check runs. An unsupported large batch can allocate a large amount of device memory, or fail allocation, before receiving the intended validation error.

Please validate batch size, beam count, and maximum length in `CreateState` or another step that runs before `DecoderState` is constructed.

src/python/py/models/builders/nemotron_parse.py
```diff
@@ -0,0 +1,340 @@
...
+                        "cache_write_indices": "cache_write_indices",
+                    },
+                    "outputs": {
+                        "logits": "logits",
+                        "present_key_names": "present.%d.key",
+                        "present_value_names": "present.%d.value",
+                    },
+                },
+            },
+            "search": {
```
baijumeswani, line 308: The generated search settings do not carry over the checkpoint's generation defaults. The official Nemotron Parse configuration uses a repetition penalty, while omission here gives ORT GenAI's default of 1.0 and can change long OCR or Markdown output.

Could we load the supported values from the checkpoint `GenerationConfig`, at least `repetition_penalty`, and add a test for the emitted setting?

baijumeswani (commented):
## Overall overview

This PR adds end-to-end Nemotron Parse support. The builder exports a RADIO image encoder and a shared mBART decoder graph, moves the cross-attention cache work into the encoder, and uses TensorScatter to update a fixed self-attention cache during generation. The runtime adds native image and prompt processing, separate prefill and one-token TRT-RTX sessions, configuration support, and integration with the existing C, C#, and Python multimodal examples.

There is a lot to like here. Computing the image and cross-attention data once is the right shape for this model. Reusing one decoder graph for prefill and decode avoids maintaining two graph implementations. The TRT-RTX profiles are focused on the two important query sizes, remote Hugging Face code remains opt-in, and the builder tests include useful numerical comparisons for the encoder and decoder rather than checking node names alone. Most shared behavior is also gated by model type or provider, which limits risk to existing models.

I found several correctness and usability issues that are worth addressing before merge. The most important are the fixed prompt-length contract, handling of failed or stopped runs, forwarding runtime options through the wrapper state, export precision, and component-only packages. I left focused comments with suggested fixes below.

src/python/py/models/builders/nemotron_parse.py
```diff
@@ -0,0 +1,340 @@
...
+        self.io_dtype = (
+            ir.DataType.FLOAT16 if io_dtype is None else ir.DataType(io_dtype)
+        )
+        self.onnx_dtype = (
+            self.io_dtype if onnx_dtype is None else ir.DataType(onnx_dtype)
+        )
+        self.ep = ep
+        self.cache_dir = cache_dir
+        self.extra_options = dict(extra_options)
+        self.extra_options.setdefault("block_size", 32)
```
baijumeswani, line 35: The intended TRT-RTX INT4 graph uses Q/DQ, but `use_qdq=true` is only shown in the documentation and is not selected or required here. A normal TRT-RTX INT4 export can therefore produce `MatMulNBits` and miss the path this PR is optimizing.

Please enable Q/DQ automatically for TRT-RTX INT4 or reject that combination unless `use_qdq=true` is supplied, with a clear explanation.

src/python/py/models/builders/nemotron_parse_encoder.py
```diff
@@ -0,0 +1,673 @@
...
+            component_options,
+        )
+        self.graph.name = "nemotron_parse_radio_encoder"
+        self._validate_architecture()
+
+    def _radio_model(self):
+        model_encoder = self.encoder.model_encoder
+        return getattr(model_encoder, "radio_model", model_encoder)
+
+    @staticmethod
```
baijumeswani: The model-builder review rules ask us not to add `staticmethod` methods other than `from_pretrained`, and they also ask us to avoid leading-underscore method names. This file and the decoder add many such helpers and duplicate several common graph-building operations.

Please remove this static method, use the repository's normal method naming, and check whether the linear, normalization, attention, and shape helpers can extend shared `Model` behavior. If some duplication is required for RADIO, documenting that boundary would make the design easier to maintain.

src/models/preprocessing/nemotron_parse_processor.cpp
```diff
@@ -0,0 +1,230 @@
...
+#include <algorithm>
+#include <array>
+#include <cmath>
+#include <utility>
+#include <vector>
+
+namespace Generators {
+namespace {
+
+constexpr std::array<float, 3> kClipMean{0.48145466f, 0.4578275f, 0.40821073f};
```
baijumeswani, line 19: These values appear to match the current official checkpoint, but the exporter discards the source processor's resize, rescale, normalization, mean, standard deviation, interpolation, and padding settings. A future checkpoint or fine-tune with different preprocessing would export successfully but receive the wrong pixels.

Please save the resolved preprocessing settings in `processor_config.json` and read them here, or reject source configurations that do not match the supported contract. A pixel-level comparison with the Hugging Face processor would be a useful test.

test/python/builder/test_nemotron_parse.py
```diff
@@ -0,0 +1,908 @@
...
+        feeds["decoder_input_ids"] = next_token.numpy()
+        feeds["decoder_attention_mask"] = np.array(
+            [[1] * full_input_ids.shape[1] + [0] * 3],
+            dtype=np.int64,
+        )
+        feeds["cache_write_indices"] = np.array(
+            [input_ids.shape[1]], dtype=np.int64
+        )
+        for layer_id in range(builder.config.decoder.decoder_layers):
+            feeds[f"past_key_values.{layer_id}.key"] = prefill_outputs[
```
baijumeswani, line 687: This test feeds each `present` output back as a separate `past` input. Production binds past and present to the same buffer through `OrtIoBinding`, so this does not exercise the most important cache path. It also checks only one decode step.

Please add a small end-to-end runtime test that uses the normal model/generator API, generates several tokens, and checks the active cache plus untouched cache slots after every step. CPU and CUDA coverage would catch provider-specific aliasing problems.

test/python/builder/test_nemotron_parse.py
```diff
@@ -0,0 +1,908 @@
...
+            reshape_heads = next(
+                initializer
+                for initializer in serialized_model.graph.initializer
+                if initializer.name.endswith("/reshape_heads")
+            )
+            self.assertNotEqual(
+                reshape_heads.data_location,
+                onnx.TensorProto.EXTERNAL,
+            )
+            if ort is not None:
```
baijumeswani, line 759: Loading this graph with the CPU provider proves that the model is well formed, but it does not show that the advertised TRT-RTX INT4 path compiles, stays on the target provider, or produces reasonable output.

Please add target-provider session creation and a small FP-versus-INT4 numerical comparison with an explicit tolerance. The test should fail rather than silently fall back to CPU.

src/models/nemotron_parse.cpp
ttyio: Thank you for your review, now we have a static profile for the 8 control token path (it is the fast path), and a dynamic profile for other control token length (1 to context_length -1), and a validation check for invalid input length.

src/models/nemotron_parse.cpp
ttyio: Override the `SupportsRewind()` and return false, thank you!

src/models/io/tensor_scatter_kv_cache.cpp
ttyio: Now current_length_ is now updated in Commit(), called only after the ONNX run and logits retrieval succeed.  Thank you!

src/models/nemotron_parse.cpp
ttyio: Now `SetRunOption()` update the `encoder_state_/decoder_state_` and `GetInput()/GetOutput()` use them, thank you!

src/python/py/models/builders/nemotron_parse.py
ttyio: Now updated, the load dtype now follows the export precision (FP16, FP32, or BF16), defaults to "auto", thank you!

src/python/py/models/builders/nemotron_parse.py
ttyio: Component-only exports are now rejected with a clear error. Thank you!

src/models/model.h
```diff
@@ -81,7 +82,7 @@ struct State {
 
  protected:
   void Run(OrtSession& session, bool graph_capture_this_run = false, int graph_capture_length = 1,
-           int graph_capture_variant = 0);
+           int graph_capture_variant = 0, OrtIoBinding* io_binding = nullptr);
```
baijumeswani: Why is IO binding necessary? ort-genai already support past present buffer sharing through passing in the same buffer for inputs and outputs. Any reason that same infrastructure cannot work for nemotron-parse?

test/python/builder/test_nemotron_parse.py
ttyio: The shared runner includes builder and models in its pytest, so this suite is discovered and test failures fail those jobs. Thank you!

src/models/preprocessing/nemotron_parse_processor.cpp
ttyio: One-item prompt lists now use the supplied prompt, matching string input. Empty and multi-item lists are explicitly rejected. Thank you!

src/models/nemotron_parse.cpp
ttyio: Now explicitly rejects `nv_multi_profile_enable=1` during model construction, with an additional guard during state creation. Each decoder session uses a single profile. Thank you!

src/models/nemotron_parse.cpp
ttyio: Now we added the validation in the `CreateState`, thank you!

src/python/py/models/builders/nemotron_parse.py
ttyio: Now we uses the model’s loaded `generation_config`, or loads it from the checkpoint. Thank you!

src/python/py/models/builders/nemotron_parse.py
ttyio: The shared builder automatically sets use_qdq=True for NvTensorRtRtx. A normal TRT-RTX INT4 CLI export therefore uses Q/DQ without requiring an explicit option. Thank you!

src/python/py/models/builders/nemotron_parse_encoder.py
ttyio: removed, thank you!

src/models/preprocessing/nemotron_parse_processor.cpp
ttyio: The exporter now validates the source processor against the supported native preprocessing contract and rejects incompatible settings. Thank  you!

test/python/builder/test_nemotron_parse.py
ttyio: Add e2e test test_in_place_cache_across_multiple_steps(), thank you!

test/python/builder/test_nemotron_parse.py
ttyio: Now we have `test_trt_rtx_int4_matches_fp_reference_without_cpu_fallback` and `test_trt_rtx_dynamic_int4_matches_fp_reference_without_cpu_fallback` test int4 vs. fp32 accuracy without cpu fallback to test the accuracy. Thank you!

src/models/model.h
ttyio: Thank you for the catch, now removed this redundant io_binding.

examples/c/src/model_mm.cpp
```diff
@@ -16,6 +16,10 @@
 
 OgaGenerator* g_generator = nullptr;
 
+// Conversational-model fallback only. An omitted Nemotron Parse prompt uses
+// the native task controls: </s><s><predict_bbox><predict_classes><output_markdown>.
+constexpr const char* kDefaultUserPrompt = "What color is the sky?";
```
kunal-vaishnavi, line 21: Why is this being moved up here instead of keeping it where it originally was?

examples/c/src/model_mm.cpp
```diff
@@ -88,11 +96,25 @@ void CXX_API(
...
+    } else {
+      text = GetUserPrompt(effective_user_prompt, interactive);
+    }
     signal(SIGINT, TerminateGeneration);
     if (text == "quit()") {
       break;  // Exit the loop
     }
+    if (is_nemotron_parse) {
+      if (num_images != 1) {
+        throw std::runtime_error("Nemotron Parse requires exactly one image");
```
kunal-vaishnavi: Can we let Nemotron Parse's procesosr handle this rather than adding model-specific logic in this file?

examples/c/src/model_mm.cpp
```diff
@@ -45,6 +49,10 @@ void CXX_API(
 
   if (verbose) std::cout << "Creating model..." << std::endl;
   auto model = OgaModel::Create(*config);
+  const bool is_nemotron_parse =
```
kunal-vaishnavi: Why do we need to change the prompt if it's Nemotron Parse or not? Can we keep the default user prompt and use that automatically?

examples/c/src/model_mm.cpp
```diff
@@ -132,11 +154,16 @@ void CXX_API(
 
     // Apply chat template
     std::string prompt;
-    try {
-      bool add_generation_prompt = true;
-      prompt = ApplyChatTemplate(model_path, *tokenizer, messages, add_generation_prompt, tools);
-    } catch (...) {
+    if (is_nemotron_parse) {
```
kunal-vaishnavi: Same comment here about adding model-specific logic. We should avoid needing to add such logic in these model-generic examples.

examples/csharp/ModelMM/Program.cs
```diff
@@ -6,6 +6,10 @@
 using System.CommandLine;
 using System.Text.Json;
 
+// Conversational-model fallback only. An omitted Nemotron Parse prompt uses
+// the native task controls: </s><s><predict_bbox><predict_classes><output_markdown>.
+const string DefaultUserPrompt = "What color is the sky?";
```
kunal-vaishnavi, line 11: Same question as [this one](https://github.com/microsoft/onnxruntime-genai/pull/2301#discussion_r4076516451)

examples/csharp/ModelMM/Program.cs
```diff
@@ -120,15 +147,22 @@ bool verbose
         if (verbose) Console.WriteLine("Generator created");
 
         // Apply chat template
-        string prompt = "";
-        try
+        string prompt;
+        if (isNemotronParse)
```
kunal-vaishnavi, line 151: Same questions in this C# file like with the C++ file

examples/python/model-mm.py
```diff
@@ -22,6 +22,10 @@
     set_logger,
 )
 
+# Conversational-model fallback only. An omitted Nemotron Parse prompt uses
+# the native task controls: </s><s><predict_bbox><predict_classes><output_markdown>.
+DEFAULT_USER_PROMPT = "What color is the sky?"
```
kunal-vaishnavi, line 27: Same questions in this Python file like with the C++ file

examples/python/model-mm.py
```diff
@@ -258,5 +297,4 @@ def main(args):
     get_guidance_args(parser)
 
     args = parser.parse_args()
-    args.max_length = args.max_length if hasattr(args, "max_length") else 7680
```
kunal-vaishnavi, line 261: Max length is intentionally capped at 7680 because otherwise multi-modal models will preallocate their KV caches to their max length, which can be up to 256K for some models. This can cause OOM issues during inference, causing users to believe that there is an error in the code.

src/models/io/default_position_inputs.cpp
```diff
@@ -20,10 +20,20 @@ void DispatchOnType(ONNXTensorElementDataType type, Func&& func) {
     func.template operator()<int64_t>();
 }
 
-DefaultPositionInputs::DefaultPositionInputs(const Model& model, State& state, DeviceSpan<int32_t> sequence_lengths_unk, const std::string& attention_mask_name)
+DefaultPositionInputs::DefaultPositionInputs(
+    const Model& model, State& state,
+    DeviceSpan<int32_t> sequence_lengths_unk,
+    const std::string& attention_mask_name,
+    AttentionMaskOptions attention_mask_options)
```
kunal-vaishnavi, line 27: Similar to how we now have different classes for the various types of key-value caches (static, dynamic, windowed, etc.), can we do the same with the position inputs rather than add attention mask options (e.g. static, dynamic, etc.)?

src/models/io/default_position_inputs.h
```diff
@@ -4,8 +4,26 @@
 
 namespace Generators {
 
+enum class AttentionMaskMode {
```
kunal-vaishnavi: Can we convert these into different classes that each have their own file? Other EPs may want to customize logic per class.

src/models/preprocessing/nemotron_parse_processor.cpp
```diff
@@ -0,0 +1,252 @@
...
```
kunal-vaishnavi, line 19: Can these all go in the vision processing config instead of hardcoding them?

https://github.com/microsoft/onnxruntime-genai/blob/2b363be4a283963e166c4889b112f14e3131dd72/src/config.h#L280

You can add it in the GenAI config as follows.

```json
"vision": {
    "config_filename": "vision_processing.json"
}
```

src/models/preprocessing/nemotron_parse_processor.cpp
```diff
@@ -0,0 +1,252 @@
...
+                                        int64_t required_prompt_length,
+                                        int context_length,
+                                        Ort::Allocator& allocator) {
+  const auto task_prompt = prompt.empty() ? kDefaultTaskPrompt : prompt;
+  auto prompt_ids = tokenizer.Encode(std::string(task_prompt).c_str());
+  const int32_t tokenizer_bos = tokenizer.TokenToTokenId("<s>");
+  const int32_t tokenizer_eos = tokenizer.TokenToTokenId("</s>");
+
+  // AutoProcessor tokenization uses add_special_tokens=True. OGA tokenizers
+  // deliberately disable automatic special tokens, so reproduce the checkpoint's
```
kunal-vaishnavi, line 122: You can change these settings by using the tokenizer's `UpdateOptions` API added [here](https://github.com/microsoft/onnxruntime-genai/pull/1785). Can we instead use this API to use `add_special_tokens=True` or is this logic still needed?

src/models/preprocessing/multi_modal_processor.cpp
```diff
@@ -55,7 +57,7 @@ std::unique_ptr<NamedTensors> MultiModalProcessor::Process(const std::string& pr
 }
 
 std::unique_ptr<NamedTensors> MultiModalProcessor::Process(std::span<const char*> prompts, const Images* images, const Audios* audios) const {
-  Payload payload{"", prompts, images, audios};
+  Payload payload{"", prompts, images, audios, true};
```
kunal-vaishnavi, line 60: There already exists an overloaded `MultiModalProcessor::Process` API for when there is a single prompt vs multiple prompts.

https://github.com/microsoft/onnxruntime-genai/blob/2b363be4a283963e166c4889b112f14e3131dd72/src/models/preprocessing/multi_modal_processor.cpp#L52-L60

Can we use that instead of adding this new boolean variable?

src/models/model.cpp
```diff
@@ -951,8 +952,11 @@ std::shared_ptr<Model> CreateModel(OrtEnv& ort_env, std::unique_ptr<Config> conf
     return std::make_shared<ParakeetTdtModel>(std::move(config), ort_env);
   if (ModelType::IsALM(config->model.type))
     return std::make_shared<WhisperModel>(std::move(config), ort_env);
-  if (ModelType::IsVLM(config->model.type))
+  if (ModelType::IsVLM(config->model.type)) {
+    if (config->model.type == "nemotron_parse")
+      return std::make_shared<NemotronParseModel>(std::move(config), ort_env);
```
kunal-vaishnavi, line 957: Now that I think about it some more, can we instead make a `MultiModalLanguageModel` and then introduce a factory in there to create a `NemotronParseModel`?

src/models/nemotron_parse.cpp
kunal-vaishnavi: Are these profile settings not used in all models?

src/models/nemotron_parse.cpp
```diff
@@ -0,0 +1,452 @@
...
+  AppendProfileShape(profile, first, decoder.inputs.cache_write_indices, {1});
+  return profile.str();
+}
+
+void ConfigureDecoderSession(OrtSessionOptions& session_options,
+                             const Config& config,
+                             int min_length, int opt_length, int max_length) {
+  // Shape overrides enable ORT's static optimizations on the fast paths.
+  // Only custom-prompt prefill retains a symbolic sequence dimension.
+  session_options.AddFreeDimensionOverrideByName("batch_size", 1);
```
kunal-vaishnavi, line 103: Are you changing the symbolic shapes in the ONNX model through these calls? If so, why can't these be set on the ONNX model when produced?

src/models/nemotron_parse.h
```diff
@@ -0,0 +1,29 @@
...
+
+struct NemotronParseModel : Model {
+  NemotronParseModel(std::unique_ptr<Config> config, OrtEnv& ort_env);
+
+  std::unique_ptr<State> CreateState(DeviceSpan<int32_t> sequence_lengths,
+                                     const GeneratorParams& params) const override;
+
+  std::unique_ptr<OrtSession> encoder_session_;
+  std::unique_ptr<OrtSession> decoder_session_;
+  std::unique_ptr<OrtSession> prefill_decoder_session_;
```
kunal-vaishnavi, line 21: If you're using the new encoder-decoder setup, what are the extra prefill sessions for?

src/python/py/models/builders/base.py
```diff
@@ -2262,6 +2264,35 @@ def make_transpose(self, name, root_input, dtype, shape, perm):
...
+        updates,
+        indices,
+        dtype,
+        shape,
+        *,
+        output=None,
+        axis=-2,
+        mode="linear",
+    ):
+        """Update slices of a tensor with the standard ONNX TensorScatter op."""
```
kunal-vaishnavi, line 2280: If TensorScatter requires opset 24 at minimum, then let's increase the default opset to be 24 for all models instead. You will need to check that other ops haven't changed signatures when you change the default opset.

src/python/py/models/builders/base.py
```diff
@@ -44,6 +44,8 @@
 
 
 class Model:
+    external_data_size_threshold_bytes = 0
```
kunal-vaishnavi: Can we undo this? Looks like you don't need this change.

src/python/py/models/builders/nemotron_parse.py
```diff
@@ -0,0 +1,414 @@
...
+
+import onnx_ir as ir
+import torch
+from transformers import AutoModel, AutoProcessor, GenerationConfig
+
+from .nemotron_parse_decoder import NemotronParseDecoderComponent
+from .nemotron_parse_encoder import NemotronParseEncoderComponent
+
+
+class NemotronParseModel:
```
kunal-vaishnavi, line 19: Can you inherit the base `Model` class and simplify a lot of the work done in this file?

src/python/py/models/builders/nemotron_parse_decoder.py
```diff
@@ -0,0 +1,782 @@
...
+
+from .base import Model
+
+
+class NemotronParseDecoderComponent(Model):
+    """Build the unified Nemotron Parse mBART prefill/decode graph."""
+
+    # ORT shape inference needs static Reshape inputs in the model protobuf.
+    # Keep small graph constants embedded while externalizing decoder weights.
+    external_data_size_threshold_bytes = 1024
```
kunal-vaishnavi, line 20: Can you explain this some more? We construct the graph so we decide how shape inference should be. Why can this not be set to the default value of 0?

src/python/py/models/builders/nemotron_parse_decoder.py
```diff
@@ -0,0 +1,782 @@
...
+        for layer_id in range(self.num_layers):
+            for name, shape in (
+                (f"present.{layer_id}.key", head_shape),
+                (f"present.{layer_id}.value", head_shape),
+            ):
+                self.graph.outputs.append(
+                    self.make_value(name, self.io_dtype, shape)
+                )
+
+    def make_constants(self):
```
kunal-vaishnavi, line 181: You should not have to do all this work. The model builder is [designed](https://github.com/microsoft/onnxruntime-genai/blob/main/src/python/py/models/DESIGN.md#constants) to do it for you. As long as you use the constants namespace, it will be saved.

src/python/py/models/builders/nemotron_parse_decoder.py
```diff
@@ -0,0 +1,782 @@
...
+            self.constant_names["query_position_axes"],
+        )
+        self.make_initializer(
+            torch.arange(
+                self.cache_sequence_length, dtype=torch.int64
+            ).reshape(1, 1, 1, self.cache_sequence_length),
+            self.constant_names["key_positions"],
+        )
+
+    def make_attention_mask(self):
```
kunal-vaishnavi, line 203: Let's make sure that we're following the [namespaces conventions](https://github.com/microsoft/onnxruntime-genai/blob/main/src/python/py/models/DESIGN.md#namespaces) when creating node names. Most of the nodes created here are not.

src/python/py/models/builders/nemotron_parse_encoder.py
```diff
@@ -0,0 +1,669 @@
...
+                    position,
+                    size=target,
+                    mode="bilinear",
+                    align_corners=True,
+                )
+        return position.permute(0, 2, 3, 1).reshape(
+            1, self.patch_count, self.radio_hidden_size
+        )
+
+    def make_radio_block(self, layer_id, block, hidden_states):
```
kunal-vaishnavi, line 301: Same comment [as this](https://github.com/microsoft/onnxruntime-genai/pull/2301#discussion_r4077711252) throughout this file as well

src/python/py/models/README.md
```diff
@@ -236,6 +238,117 @@ python -m onnxruntime_genai.models.builder -m model_name -o path_to_output_folde
...
+decoder emits the standard main-domain `TensorScatter` operator from opset 24.
+Both graph components reuse `Model` for weight serialization, quantization,
+MatMul/bias construction, and shape/operator wrappers. Their linear helpers only
+compose the shared MatMul and bias builders. Standalone LayerNormalization keeps
+each source module's epsilon and tensor shape without the base decoder's residual
+state bookkeeping. RADIO's packed QKV, patch geometry, and compression neck, plus
+mBART's absolute positions, cross-attention, and fixed TensorScatter cache, remain
+model-specific; they do not use the base decoder's fused-attention/cache pipeline.
+
+Its model-specific options are:
```
kunal-vaishnavi, line 255: If these are fixed values that should be used for this model, then let's keep them in a processor config or in the GenAI config rather than storing them here as extra options.

src/python/py/models/README.md
```diff
@@ -236,6 +238,117 @@ python -m onnxruntime_genai.models.builder -m model_name -o path_to_output_folde
 python builder.py -m model_name -o path_to_output_folder -p precision -e execution_provider -c cache_dir_for_hf_files --extra_options hf_remote=true
 ```
 
+#### Nemotron Parse Options
```
kunal-vaishnavi, line 241: We can create a new folder at `src/python/py/models/docs` to store model-specific information rather than add them in the main README. We will need to eventually clean up the main README as well for other model-specific options that were added there.

src/config.cpp
```diff
@@ -549,6 +551,10 @@ struct DecoderOutputs_Element : JSON::Element {
       v_.present_value_scale_names = JSON::Get<std::string_view>(value);
     } else if (name == "present_names") {
       v_.present_names = JSON::Get<std::string_view>(value);
+    } else if (name == "cross_present_key_names") {
```
kunal-vaishnavi, line 554: Based on the discussion we had offline, we should be having these in the encoder's outputs and not the decoder's outputs.

examples/c/src/model_mm.cpp
ttyio: This is mostly because we have not same default user prompt for nemotron parser vs. other VLM. For nemotron parser, the prompt is a control string, when user did not provide, we need this "<predict_bbox><predict_classes><output_markdown>", we cannot use the "What color is the sky" as default, thanks!

examples/c/src/model_mm.cpp
ttyio: Same here, the default prompt is "<predict_bbox><predict_classes><output_markdown>" for nemotron parser. Thanks!

examples/csharp/ModelMM/Program.cs
ttyio: Same here, we need default prompt "<predict_bbox><predict_classes><output_markdown>".

examples/csharp/ModelMM/Program.cs
ttyio: Same here, we need use prompt like "<predict_bbox><predict_classes><output_markdown>", by default we pass an empty string, which nemotron parser interprets as "<predict_bbox><predict_classes><output_markdown>". thanks!

examples/python/model-mm.py
ttyio: Same here, we need use prompt like "<predict_bbox><predict_classes><output_markdown>", by default we pass an empty string, which nemotron parser interprets as "<predict_bbox><predict_classes><output_markdown>". thanks!

examples/c/src/model_mm.cpp
ttyio: I think again, and will add a new `model.default_user_prompt` in the `genai_config.json`, for nemotron parser, it contains `<predict_bbox><predict_classes><output_markdown>`, Other VLMs without that field: the example falls back to "What color is the sky?".

examples/c/src/model_mm.cpp
ttyio: same here, will add a new `model.default_user_prompt` in the `genai_config.json`.

examples/c/src/model_mm.cpp
ttyio: this is fixed after add a new `model.default_user_prompt` in the `genai_config.json`.

examples/csharp/ModelMM/Program.cs
ttyio: this is fixed after add a new `model.default_user_prompt` in the `genai_config.json`.

examples/csharp/ModelMM/Program.cs
ttyio: this is fixed after add a new `model.default_user_prompt` in the `genai_config.json`. thanks!

examples/python/model-mm.py
ttyio: this is fixed after add a new `model.default_user_prompt` in the `genai_config.json`. thanks!

examples/python/model-mm.py
ttyio: previous I moved the place of this code because of the default prompt is different, now this is fixed after add a new `model.default_user_prompt` in the `genai_config.json`. thanks!

src/models/io/default_position_inputs.h
ttyio: Split into StaticAttentionMask and DynamicAttentionMask, they are composition in the DefaultPositionInputs class, this is more clear, thank you!

src/models/io/default_position_inputs.cpp
ttyio: Split into StaticAttentionMask and DynamicAttentionMask, they are composition in the DefaultPositionInputs class, this is more clear, thank you!

src/models/preprocessing/nemotron_parse_processor.cpp
ttyio: moved, now `nemotron_parse.py` will generate this `vision_processing.json`, including those values, this is more clean, thank you!

src/models/preprocessing/multi_modal_processor.cpp
ttyio: fixed, and I make sure we fail aloud when prompt list is given before tokenization or image preprocessing

src/python/py/models/builders/base.py
ttyio: updated to opset 24, thanks!

src/python/py/models/builders/nemotron_parse.py
ttyio: Thanks, now updated to use base `Model`, it is more clean!

src/python/py/models/builders/nemotron_parse_decoder.py
ttyio: Follow your suggestion to use [designed](https://github.com/microsoft/onnxruntime-genai/blob/main/src/python/py/models/DESIGN.md#constants), then we no more need this. Thanks!

src/python/py/models/builders/nemotron_parse_decoder.py
ttyio: changed, this is more clean, thank you!

src/config.cpp
ttyio: removed, thank you for the catch!

src/python/py/models/README.md
ttyio: updated, thank you!

src/python/py/models/builders/base.py
ttyio: removed, thank you!

src/python/py/models/builders/nemotron_parse_decoder.py
ttyio: updated, thank you for the catch!

src/python/py/models/builders/nemotron_parse_encoder.py
ttyio: updated the entire file, thank you!

src/models/nemotron_parse.h
ttyio: For nemotron parse, image and prompt are not concat together. Image features are feed into decoder's cross-attention k/v, so the decoder need the full prompt to output the first token. For performance, we create two prefill sessions for the decoder onnx: a static session for the default prompt length (8 tokens), and a dynamic session for other prompt lengths. and `decoder_session_` handles the one token input with k/v cached. All three decoder sessions are from the same onnx. Thank you!

src/models/nemotron_parse.cpp
ttyio: they are not used in other VLM models, other VLM concat the image feature and the prompt token. so we can clear separate the decoder.onnx; but nemotron parse, the image feature is cross attention k/v input, so the decoder.onnx itself also include the prefill, decoding. the sessions we created is different than other VLM models.

examples/c/src/model_mm.cpp
ttyio: removed, thank you!

src/python/py/models/README.md
ttyio: removed those option, moved to the config, thank you!

src/models/model.cpp
ttyio: Now we have `MultiModalLanguageModel::Create()`, and the `NemotronParseModel` is crated there. but we don't share the `MultiModalLanguageModel`, because the nemotron parse need feed prompt toke into cross attention k/v, it is diff compare to other VLM models. Thanks!

src/models/preprocessing/nemotron_parse_processor.cpp
ttyio: Updated to use UpdateOptions() with add_special_tokens=true. Thank you!

src/models/nemotron_parse.cpp
ttyio: Thanks, now we only have the `sequence_length` here, others like `batch_size` are move to the exporter.

PR #2302, opened 2026-07-17
Add Moonshine streaming ASR model support

nenad1002:
Adds a new moonshine_streaming model type: a 5-session ONNX pipeline (frontend / encoder / adapter / cross-KV / decoder-KV) driving an AED decoder with bidirectional cross-attention over accumulating memory. The main advantage of this model is its speed and relatively low CPU utilization, making it suitable for devices where Nemotron would require too much compute or memory. Both HF small and tiny version will be exported and run with onnxruntime-genai.

Highlights:

New model MoonshineStreamingModel and state (src/models/moonshine_streaming.{h,cpp}) built on the shared TransducerState interface, so it plugs into the existing streaming pump alongside Nemotron/Parakeet.

Chunked streaming decode with lookahead hold-back, LCP-based commit across chunks (increases accuracy by more than 5% WER by delaying the emitting until the same text is confirmed twice), per-chunk token cap, and BOS re-decode. AR loop reuses a pre-allocated [1,1] token tensor, and the committed prefix is teacher-forced in a single parallel decoder call. 

Incremental cross-KV cache - new memory frames are projected once and concatenated into the cached k_cross / v_cross instead of recomputing from scratch.

VAD-aware segmentation - hard segment cap, min-segment gating, and mid-segment silence handling.

Python example: renames nemotron_speech.py to streaming_asr.py so one entry point covers all streaming ASR models.
Tests: extends c_api_tests.cpp and updates the Python API tests / conftest for the new model type.

**Moonshine Streaming Tiny (kquant8) - WER on Open ASR Leaderboard (streaming)**

Model size: ~95 MB on disk (INT8 k-quant ORT build) — ~34M parameters.

| Dataset | Split | WER |
|---|---|---|
| AMI | test | 24.06% |
| Earnings22 | test | 27.15% |
| GigaSpeech | test | 15.16% |
| LibriSpeech | test.clean | 8.01% |
| LibriSpeech | test.other | 17.12% |
| VoxPopuli | test | 19.03% |
| TED-LIUM | test | 11.56% |
| SPGISpeech | test | 10.45% |
| **Average** | — | **16.57%** |


test/python/test_onnxruntime_genai_api.py
```diff
@@ -1166,7 +1123,15 @@ def normalize(text):
     return d[len(r)][len(h)] / max(len(r), 1)
 
 
-def test_streaming_asr_transcription_quality(nemotron_speech_model_path, test_data_path):
+# Per-model word-error-rate thresholds for the transcription-quality test.
+_ASR_WER_THRESHOLDS = {
+    "nemotron-speech-streaming": 0.15,
+    "moonshine-streaming-small": 0.2,
+    "moonshine-streaming-tiny": 0.25,
+}
```
themason2011, line 1121: Is Nemotron's WER threshold supposed to be less than 0.1?

nenad1002: We want this test to pass consistently, so it’s probably better to keep it a bit loose.

themason2011:
> Adds a new moonshine_streaming model type: a 5-session ONNX pipeline (frontend / encoder / adapter / cross-KV / decoder-KV) driving an AED decoder with bidirectional cross-attention over accumulating memory. The main advantage of this model is its speed and relatively low CPU utilization, making it suitable for devices where Nemotron would require too much compute or memory. Both HF small and tiny version will be exported and run with onnxruntime-genai.
> 
> Highlights:
> 
> New model MoonshineStreamingModel and state (src/models/moonshine_streaming.{h,cpp}) built on the shared TransducerState interface, so it plugs into the existing streaming pump alongside Nemotron/Parakeet.
> 
> Chunked streaming decode with lookahead hold-back, LCP-based commit across chunks (increases accuracy by more than 5% WER by delaying the emitting until the same text is confirmed twice), per-chunk token cap, and BOS re-decode. AR loop reuses a pre-allocated [1,1] token tensor, and the committed prefix is teacher-forced in a single parallel decoder call.
> 
> Incremental cross-KV cache - new memory frames are projected once and concatenated into the cached k_cross / v_cross instead of recomputing from scratch.
> 
> VAD-aware segmentation - hard segment cap, min-segment gating, and mid-segment silence handling.
> 
> Python example: renames nemotron_speech.py to streaming_asr.py so one entry point covers all streaming ASR models. Tests: extends c_api_tests.cpp and updates the Python API tests / conftest for the new model type.
> 
> **Moonshine Streaming Tiny (kquant8) - WER on Open ASR Leaderboard (streaming)**
> 
> Model size: ~95 MB on disk (INT8 k-quant ORT build) — ~34M parameters.
> 
> Dataset	Split	WER
> AMI	test	24.06%
> Earnings22	test	27.15%
> GigaSpeech	test	15.16%
> LibriSpeech	test.clean	8.01%
> LibriSpeech	test.other	17.12%
> VoxPopuli	test	19.03%
> TED-LIUM	test	11.56%
> SPGISpeech	test	10.45%
> **Average**	—	**16.57%**

What is the cause of AMI and Earning22 having the highest WERs? I know AMI has multiple languages and that may be a root cause for that Dataset, but I haven't heard of Earnings22.

src/models/preprocessing/streaming_processor.cpp
```diff
@@ -121,4 +130,13 @@ std::string StreamingProcessor::GetOption(const char* key) const {
   }
 }
 
+// Factory / dispatch point: maps a model type to its concrete streaming
+// processor.
+std::unique_ptr<StreamingProcessor> CreateStreamingProcessor(Model& model) {
```
themason2011, line 131: Do we want to make the default processor Nemotron, or should we specify which model types use Moonshine/Nemotron and then all models that are not explicitly supported throw an error on this method?

src/models/preprocessing/streaming_processor.cpp
```diff
@@ -35,6 +37,13 @@ void StreamingProcessor::EnableVadFromModel() {
   consecutive_silence_chunks_ = 0;
 }
 
+bool StreamingProcessor::IsChunkSilent(const float* chunk_data, size_t chunk_size) {
```
themason2011, line 40: Could be worth refactoring this into ShouldDropChunk, which does the exact same thing in the beginning. I.e. have ShouldDropChunk call IsChunkSilent instead of repeating same code.

src/models/moonshine_streaming.cpp
```diff
@@ -0,0 +1,843 @@
...
+}
+}  // namespace
+
+void MoonshineConfig::PopulateFromConfig(const Config& config) {
+  const auto& m = config.model;
+  const auto& enc = m.encoder;
+  const auto& dec = m.decoder;
+  const auto& ms = m.moonshine;
+
+  chunk_samples = m.chunk_samples;
```
themason2011, line 52: Might want to have range checks for certain parameters, ex. for chunk_samples should be positive and non-zero

src/config.cpp
```diff
@@ -1173,6 +1216,8 @@ struct Model_Element : JSON::Element {
       v_.sample_rate = SafeDoubleToInt(JSON::Get<double>(value), name);
     } else if (name == "chunk_samples") {
       v_.chunk_samples = SafeDoubleToInt(JSON::Get<double>(value), name);
+    } else if (name == "overlap_samples") {
+      v_.overlap_samples = SafeDoubleToInt(JSON::Get<double>(value), name);
```
themason2011: is overlap_samples ever used?

examples/python/model-asr.py
```diff
@@ -1,5 +1,18 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License.
+"""Streaming ASR example using the onnxruntime-genai StreamingProcessor API.
```
themason2011, line 3: Have been using AI to assist with reviewing, it brought up this solution to a CI failure but I can't seem to find the failure logs when searching manually in Github. Not sure if this is still applicable, but worth mentioning:

C# nullable warnings are failing checks. src/csharp/StreamingProcessor.cs uses NamedTensors? on lines 24 and 42, but the file isn't in a #nullable context. Multiple CI jobs (windows-cpu, windows-cuda, linux_cpu) flag: "The annotation for nullable reference types should only be used in code within a '#nullable' annotations context." Wrap the file in #nullable enable (or enable it project-wide) to clear these.

src/models/moonshine_streaming_processor.h
```diff
@@ -0,0 +1,42 @@
...
+struct MoonshineStreamingProcessor : StreamingProcessor {
+  explicit MoonshineStreamingProcessor(Model& model);
+  ~MoonshineStreamingProcessor() override;
+
+  std::unique_ptr<NamedTensors> Process(const float* audio_data, size_t num_samples) override;
+  std::unique_ptr<NamedTensors> Flush() override;
+
+ private:
+  Model& model_;
+  MoonshineConfig config_;
```
themason2011: Could use a reference for config_ to reduce copying (there three other places this could be done in moonshine_streaming.h)

src/models/preprocessing/streaming_processor.cpp
nenad1002: We can do throw.

src/config.cpp
nenad1002: Good catch

src/models/moonshine_streaming.cpp
nenad1002: While we should prob do it, we don't do it for most part to save on inference time. To be consistent skipping this

nenad1002:
> > Adds a new moonshine_streaming model type: a 5-session ONNX pipeline (frontend / encoder / adapter / cross-KV / decoder-KV) driving an AED decoder with bidirectional cross-attention over accumulating memory. The main advantage of this model is its speed and relatively low CPU utilization, making it suitable for devices where Nemotron would require too much compute or memory. Both HF small and tiny version will be exported and run with onnxruntime-genai.
> > Highlights:
> > New model MoonshineStreamingModel and state (src/models/moonshine_streaming.{h,cpp}) built on the shared TransducerState interface, so it plugs into the existing streaming pump alongside Nemotron/Parakeet.
> > Chunked streaming decode with lookahead hold-back, LCP-based commit across chunks (increases accuracy by more than 5% WER by delaying the emitting until the same text is confirmed twice), per-chunk token cap, and BOS re-decode. AR loop reuses a pre-allocated [1,1] token tensor, and the committed prefix is teacher-forced in a single parallel decoder call.
> > Incremental cross-KV cache - new memory frames are projected once and concatenated into the cached k_cross / v_cross instead of recomputing from scratch.
> > VAD-aware segmentation - hard segment cap, min-segment gating, and mid-segment silence handling.
> > Python example: renames nemotron_speech.py to streaming_asr.py so one entry point covers all streaming ASR models. Tests: extends c_api_tests.cpp and updates the Python API tests / conftest for the new model type.
> > **Moonshine Streaming Tiny (kquant8) - WER on Open ASR Leaderboard (streaming)**
> > Model size: ~95 MB on disk (INT8 k-quant ORT build) — ~34M parameters.
> > Dataset	Split	WER
> > AMI	test	24.06%
> > Earnings22	test	27.15%
> > GigaSpeech	test	15.16%
> > LibriSpeech	test.clean	8.01%
> > LibriSpeech	test.other	17.12%
> > VoxPopuli	test	19.03%
> > TED-LIUM	test	11.56%
> > SPGISpeech	test	10.45%
> > **Average**	—	**16.57%**
> 
> What is the cause of AMI and Earning22 having the highest WERs? I know AMI has multiple languages and that may be a root cause for that Dataset, but I haven't heard of Earnings22.

Both AMI and Earnings22 are English, but noisy datasets from meetings, so it has higher WER.

themason2011 (dismissed):

examples/python/model-asr.py
kunal-vaishnavi, line 3: Can we make similar generic changes for the C/C++ and C# examples? Separately, should we rename this file to something like `model-asr.py` to stay consistent with the other generic examples' filenames?

src/csharp/StreamingProcessor.cs
```diff
@@ -3,6 +3,8 @@
 
 using System;
 
+#nullable enable
```
kunal-vaishnavi, line 6: Why do we need this?

src/models/moonshine_streaming.cpp
```diff
@@ -0,0 +1,842 @@
...
+
+void MoonshineStreamingState::RunPipeline() {
+  // Reset the per-chunk queue so any early return (dropped silence, no stable
+  // frames yet) leaves the chunk with zero committed tokens.
+  pending_tokens_.clear();
+  pending_idx_ = 0;
+  last_tokens_.clear();
+
+  // Deferred reset from a previous segment close (hard cap / VAD / Flush).
+  if (needs_reset_) {
```
kunal-vaishnavi, line 536: From Copilot:

The segment reset clears the frontend/audio/memory state, but it leaves `previous_pass_tokens_`, `emitted_count_`, and `previous_memory_frames_` intact. The subsequent `memory_frames < previous_memory_frames_` check is not sufficient to detect a new segment: the next utterance can be equal to or longer than the previous one. In that case, decoding starts with the previous segment's commit history, which can suppress tokens from the beginning of the new utterance or compare unrelated token sequences for LCP. Please reset the token-commit bookkeeping atomically with `ResetAccumulation()` (or set an explicit `new_segment_` flag) and add a regression test covering a second segment at least as long as the first.

test/cpp/c_api_tests.cpp
```diff
@@ -2272,36 +2340,163 @@ TEST(CAPITests, StreamingASRVadSetGetOption) {
...
+// is_silent flag, and ALL segmentation (accumulate / flush past min_segment / reset / drop
+// pre-utterance silence) happens downstream in the State's RunPipeline. This test therefore:
+//   * asserts the thin-processor contract  (a full chunk is never dropped -> Process() != nullptr),
+//   * checks the is_silent verdict that drives segmentation (speech -> 0, silence -> 1), and
+//   * drives the speech -> silence -> speech pattern through the Generator so the State's
+//     segmentation routing actually runs end to end (flush + reset + restart).
+// Silero VAD only fires on real human speech, so "speech" uses a synthetic tone with VAD disabled
+// (forces is_silent=0, routing through the same accumulation path real speech would) while VAD is
+// enabled for the silence that drives the segment boundary.
+TEST(StreamingASRTests, MoonshineVadSegmentation) {
```
kunal-vaishnavi, line 2410: From Copilot:

This test exercises a VAD-triggered reset, but it does not verify the emitted transcript/token sequence after the reset, nor does it cover a second segment whose memory is equal to or longer than the first. Please extend it to assert that the post-reset segment starts from a fresh BOS/LCP state—for example, feed two segments where the second is at least as long as the first and verify that tokens from the second segment are emitted rather than being suppressed by the previous segment's `emitted_count_` and `previous_pass_tokens_`.

src/models/moonshine_streaming.cpp
```diff
@@ -0,0 +1,842 @@
...
+// ---- frontend --------------------------------------------------------------
+
+MoonshineFrontendSubState::MoonshineFrontendSubState(const MoonshineStreamingModel& model,
+                                                     const GeneratorParams& params)
+    : State{params, model},
+      model_{model},
+      config_{model.moonshine_config_} {
+  AllocateStateBuffers();
+
+  // Inputs: audio_chunk (set per-run) + the 5 persistent state buffers.
```
kunal-vaishnavi, line 215: Can we read the input and output names from the GenAI config? If a future model decides to change a name, the pipeline will break.

src/models/moonshine_streaming.cpp
```diff
@@ -0,0 +1,842 @@
...
+  output_names_.push_back("conv1_buffer_out");
+  outputs_.push_back(nullptr);
+  output_names_.push_back("conv2_buffer_out");
+  outputs_.push_back(nullptr);
+  output_names_.push_back("frame_count_out");
+  outputs_.push_back(nullptr);
+}
+
+void MoonshineFrontendSubState::AllocateStateBuffers() {
+  auto& alloc = model_.allocator_cpu_;
```
kunal-vaishnavi, line 257: This looks like it allocates inputs for CPU only. Can we use `p_device_inputs_->GetAllocator()` to support non-CPU EPs?

From Copilot:

The pipeline currently allocates all persistent state and intermediate tensors with `model_.allocator_cpu_`, including the frontend buffers, encoder/adapter inputs, cross-KV cache, self-KV cache, and decoder inputs.

Although these sessions inherit the configured session options and may register a non-CPU execution provider, the tensors remain CPU-resident. This can introduce CPU↔device copies around nearly every session invocation, and some execution providers may not accept these manually allocated CPU tensors as intended.

Could we either document Moonshine as CPU-only for this implementation, or make the pipeline device-aware by using the selected input/KV-cache allocators and device-safe KV-cache concatenation? If non-CPU execution is expected, please also add an EP-specific test covering the full streaming pipeline.

src/models/moonshine_streaming.cpp
```diff
@@ -0,0 +1,842 @@
...
+  {
+    auto shape = std::array<int64_t, 2>{1, config_.sample_buffer_size};
+    sample_buffer_ = OrtValue::CreateTensor(alloc, shape, ONNX_TENSOR_ELEMENT_DATA_TYPE_FLOAT);
```
kunal-vaishnavi: Can we get the input and output types from the session info? Ex: 

https://github.com/microsoft/onnxruntime-genai/blob/af166b1903227d2384424a356068efaea197184a/src/models/nemotron_speech.cpp#L222-L235

src/models/moonshine_streaming.cpp
```diff
@@ -0,0 +1,842 @@
...
+  return {};
+}
+
+// ---- encoder ---------------------------------------------------------------
+
+MoonshineEncoderSubState::MoonshineEncoderSubState(const MoonshineStreamingModel& model,
+                                                   const GeneratorParams& params)
+    : State{params, model},
+      model_{model} {
+  input_names_.push_back("features");
```
kunal-vaishnavi: Same comment as [here](https://github.com/microsoft/onnxruntime-genai/pull/2302#discussion_r3938025338)

src/models/moonshine_streaming.cpp
```diff
@@ -0,0 +1,842 @@
...
+                                                   const GeneratorParams& params)
+    : State{params, model},
+      model_{model} {
+  auto pos_shape = std::array<int64_t, 1>{1};
+  pos_tensor_ = OrtValue::CreateTensor(model_.allocator_cpu_, pos_shape,
+                                       ONNX_TENSOR_ELEMENT_DATA_TYPE_INT64);
+  *pos_tensor_->GetTensorMutableData<int64_t>() = 0;
+
+  encoded_input_idx_ = inputs_.size();
+  input_names_.push_back("encoded");
```
kunal-vaishnavi: Same comment as [here](https://github.com/microsoft/onnxruntime-genai/pull/2302#discussion_r3938025338)

src/models/moonshine_streaming.cpp
```diff
@@ -0,0 +1,842 @@
...
+  return {};
+}
+
+// ---- cross_kv --------------------------------------------------------------
+
+MoonshineCrossKvSubState::MoonshineCrossKvSubState(const MoonshineStreamingModel& model,
+                                                   const GeneratorParams& params)
+    : State{params, model},
+      model_{model} {
+  input_names_.push_back("memory");
```
kunal-vaishnavi: Same comment as [here](https://github.com/microsoft/onnxruntime-genai/pull/2302#discussion_r3938025338)

src/models/moonshine_streaming.cpp
```diff
@@ -0,0 +1,842 @@
...
+}
+
+// ---- decoder_kv ------------------------------------------------------------
+
+MoonshineDecoderKvSubState::MoonshineDecoderKvSubState(const MoonshineStreamingModel& model,
+                                                       const GeneratorParams& params)
+    : State{params, model},
+      model_{model} {
+  token_input_idx_ = inputs_.size();
+  input_names_.push_back("token");
```
kunal-vaishnavi: Same comment as [here](https://github.com/microsoft/onnxruntime-genai/pull/2302#discussion_r3938025338)

src/models/moonshine_streaming.cpp
```diff
@@ -0,0 +1,842 @@
...
+}
+
+// ---- adapter ---------------------------------------------------------------
+
+MoonshineAdapterSubState::MoonshineAdapterSubState(const MoonshineStreamingModel& model,
+                                                   const GeneratorParams& params)
+    : State{params, model},
+      model_{model} {
+  auto pos_shape = std::array<int64_t, 1>{1};
+  pos_tensor_ = OrtValue::CreateTensor(model_.allocator_cpu_, pos_shape,
```
kunal-vaishnavi: Same comment as [here](https://github.com/microsoft/onnxruntime-genai/pull/2302#discussion_r3938113504)

src/models/moonshine_streaming.cpp
```diff
@@ -0,0 +1,842 @@
...
+  adapter_state_ = std::make_unique<MoonshineAdapterSubState>(model, params);
+  cross_kv_state_ = std::make_unique<MoonshineCrossKvSubState>(model, params);
+  decoder_kv_state_ = std::make_unique<MoonshineDecoderKvSubState>(model, params);
+
+  // Idle until SetExtraInputs() supplies a chunk.
+  chunk_done_ = true;
+
+  // Pre-allocate single-token int64 tensor [1, 1].
+  auto tok_shape = std::array<int64_t, 2>{1, 1};
+  token_tensor_ = OrtValue::CreateTensor(model.allocator_cpu_, tok_shape,
```
kunal-vaishnavi: Same comment as [here](https://github.com/microsoft/onnxruntime-genai/pull/2302#discussion_r3938113504)

src/models/moonshine_streaming.cpp
```diff
@@ -0,0 +1,842 @@
...
+  ResetSelfKv();
+}
+
+MoonshineStreamingState::~MoonshineStreamingState() = default;
+
+void MoonshineStreamingState::ResetSelfKv() {
+  auto shape = std::array<int64_t, 5>{
+      config_.num_decoder_layers, 1, config_.num_decoder_heads, 0,
+      config_.decoder_head_size};
+  k_self_ = OrtValue::CreateTensor(moonshine_model_.allocator_cpu_, shape,
```
kunal-vaishnavi, line 488: Same comment as [here](https://github.com/microsoft/onnxruntime-genai/pull/2302#discussion_r3938113504)

src/models/moonshine_streaming.cpp
```diff
@@ -0,0 +1,842 @@
...
+  auto& alloc = moonshine_model_.allocator_cpu_;
+  const int encoder_dim = config_.encoder_dim;
+  const int decoder_dim = config_.decoder_dim;
+  DeviceSpan<int32_t> dummy_tokens;
+
+  // ----- frontend -------------------------------------------------------
+  // Run only when there is audio to feed (frontend audio dim must be > 0).
+  if (num > 0) {
+    auto audio_shape = std::array<int64_t, 2>{1, static_cast<int64_t>(num)};
+    auto audio_tensor =
```
kunal-vaishnavi, line 630: Same comment as [here](https://github.com/microsoft/onnxruntime-genai/pull/2302#discussion_r3938113504) for all tensors below

src/models/moonshine_streaming.h
```diff
@@ -0,0 +1,386 @@
...
+#include "transducer_state.h"
+
+namespace Generators {
+
+struct MoonshineConfig {
+  // Audio framing.
+  int chunk_samples{};  // e.g. 8000 = 500ms at 16 kHz
+
+  // Tokens.
+  int bos_token_id{};
```
kunal-vaishnavi, line 48: Can we use the existing GenAI config attributes rather than storing these?

src/models/moonshine_streaming.h
```diff
@@ -0,0 +1,386 @@
...
+  // trigger a segment break. Below this threshold, silent chunks are dropped
+  // (treated as pre/inter-utterance pause) but do NOT cut the segment — we
+  // want enough committed speech behind us that an early-segment short
+  // silence (e.g. comma-pause, breath) doesn't fragment the transcript.
+  // After this threshold, any silent chunk triggers an is_final flush + reset
+  // (e.g. 250 frames = 5s @ 50fps). Only active when VAD is enabled; set <=0
+  // to disable VAD-based segmentation entirely (hard cap still applies).
+  int min_segment_memory_frames{};
+
+  std::string frontend_filename;
```
kunal-vaishnavi, line 90: Same comment as [here](https://github.com/microsoft/onnxruntime-genai/pull/2302#discussion_r3938290568)

src/models/preprocessing/moonshine_streaming_processor.cpp
```diff
@@ -0,0 +1,96 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
kunal-vaishnavi, line 1: Can we move this file and the header file inside `src/models/preprocessing`?

test/cpp/c_api_tests.cpp
```diff
@@ -2087,11 +2112,29 @@ static void DecodeInputs(OgaGenerator& generator, OgaNamedTensors* mel) {
...
+// Uses frequency and sample_rate to produce the minimum length of speech detectable by the model.
+static std::vector<float> GenerateSineWave(size_t num_samples, float frequency, float sample_rate) {
+  std::vector<float> samples(num_samples);
+  for (size_t i = 0; i < num_samples; ++i)
+    samples[i] = 0.5f * std::sin(2.0f * 3.14159265f * frequency * static_cast<float>(i) / sample_rate);
+  return samples;
+}
+
+#ifndef STREAMING_ASR_PATH
+#define STREAMING_ASR_PATH MODEL_PATH "nemotron-speech-streaming"
```
kunal-vaishnavi, line 2129: Is it just Nemotron in the unit tests? Can we have other ASR models tested as well?

test/cpp/c_api_tests.cpp
```diff
@@ -2087,11 +2112,29 @@ static void DecodeInputs(OgaGenerator& generator, OgaNamedTensors* mel) {
...
-TEST(CAPITests, StreamingASRCreate) {
-  if (!std::filesystem::exists(STREAMING_ASR_PATH))
-    GTEST_SKIP() << "Streaming ASR model not found at " << STREAMING_ASR_PATH;
-  auto model = OgaModel::Create(STREAMING_ASR_PATH);
+// Helper: generate a synthetic sine wave of num_samples at the given frequency and sample rate.
+// Uses frequency and sample_rate to produce the minimum length of speech detectable by the model.
+static std::vector<float> GenerateSineWave(size_t num_samples, float frequency, float sample_rate) {
+  std::vector<float> samples(num_samples);
+  for (size_t i = 0; i < num_samples; ++i)
+    samples[i] = 0.5f * std::sin(2.0f * 3.14159265f * frequency * static_cast<float>(i) / sample_rate);
```
kunal-vaishnavi: Can we use [existing definitions](https://stackoverflow.com/a/1727896) for `pi`?

test/python/conftest.py
```diff
@@ -114,13 +114,25 @@ def path_for_model(request):
     return functools.partial(get_path_for_model, request.config.getoption("--test_models"))
 
 
-@pytest.fixture
-def nemotron_speech_model_path(request):
-    """Return the path to a nemotron_speech model directory, or skip if not available."""
+# ASR streaming model directories exercised by the streaming-ASR tests.
+ASR_SPEECH_MODELS = [
```
kunal-vaishnavi, line 118: Can we do a similar approach in the C++ tests for getting the model path?

test/cpp/c_api_tests.cpp
nenad1002: Not in scope

src/csharp/StreamingProcessor.cs
nenad1002: Not strictly needed for compiling, but recommended as a good practice because Process returns NamedTensors with a nullable annotation which only compiles when nullable is enabled, and the project has no global setting so each file opts in the same way as OnnxRuntimeGenAIChatClient.cs.

src/models/moonshine_streaming.cpp
```diff
@@ -0,0 +1,948 @@
...
+      if (!encode_chunk(audio, num, /*is_final=*/false)) return;
+    }
+  } else {
+    // Speech (or VAD disabled).
+    if (!encode_chunk(audio, num, /*is_final=*/false)) return;
+    // Hard-cap: once memory crosses the segment cap, schedule a reset for the
+    // next chunk (which also commits all tokens — see below).
+    if (config_.max_segment_memory_frames > 0 &&
+        memory_frames_ >= config_.max_segment_memory_frames) {
+      needs_reset_ = true;
```
baijumeswani, line 611: At the hard segment limit, the normal encode path has still held back `total_lookahead` frames. The next chunk resets the state and clears those frames, so part of the audio can be lost. Please release the held-back frames here using the same final encode path used by Flush and VAD, refresh cross-KV, and then decode and reset. Please also add a test that reaches the hard limit and checks that the boundary audio is kept.

src/config.h
```diff
@@ -345,6 +345,100 @@ struct Config {
       std::optional<RunOptions> run_options;
     } vad;
 
+    struct Moonshine {
+      std::string frontend_filename;
+      std::string encoder_filename;
+      std::string adapter_filename;
+      std::string cross_kv_filename;
+      std::string decoder_kv_filename;
```
baijumeswani, line 357: Please add all five new Moonshine model filenames to `ValidateModelPaths()`. Other model filenames reject absolute paths and `..`, but these fields currently skip that check. Validating them keeps every model file inside the selected model directory and matches the existing config behavior.

src/generator/generators.cpp
```diff
@@ -600,9 +600,11 @@ Generator::Generator(const Model& model, const GeneratorParams& params)
...
     ValidateNGramDecoding(model, params);
 
-  // RNNT and TDT models don't use the traditional search/logits pipeline,
-  // so skip the standard validations and just create the state.
-  if (ModelType::IsTransducer(model.config_->model.type)) {
+  // RNNT/TDT/streaming-enc-dec-ASR models don't use the traditional
+  // search/logits pipeline; skip the standard validations and create the
+  // TransducerState-derived state directly.
+  const auto& model_type = model.config_->model.type;
+  if (ModelType::IsTransducer(model_type) || ModelType::IsStreamingEncDecASR(model_type)) {
```
baijumeswani, line 607: This path returns without creating `search_`, but `RewindToLength()` still uses `search_` without checking it. Calling the public rewind API on a Moonshine generator will crash. Please either support rewind for this state or return a clear unsupported-operation error before `search_` is accessed.

examples/c/CMakeLists.txt
```diff
@@ -26,7 +26,7 @@ option(MODEL_CHAT "Build the Model Chat example" OFF)
 option(MODEL_QA "Build the Model Q&A example" OFF)
 option(MODEL_MM "Build the Model Multimodal example" OFF)
 option(WHISPER "Build the Whisper example" OFF)
-option(NEMOTRON_SPEECH "Build the Nemotron Speech Streaming example" OFF)
+option(MODEL_ASR "Build the Model ASR (streaming) example" OFF)
```
baijumeswani, line 29: The repository build workflows still pass `-DNEMOTRON_SPEECH=ON`. After this rename, CMake accepts that old option but does not build `model_asr`, so CI no longer compiles this example. Please update all workflows to use `MODEL_ASR`, or keep `NEMOTRON_SPEECH` as a temporary compatibility alias.

src/models/moonshine_streaming.cpp
```diff
@@ -0,0 +1,948 @@
...
+int MoonshineStreamingState::RunDecoderForward(const std::vector<int64_t>& tokens) {
+  const int64_t n = static_cast<int64_t>(tokens.size());
+  // Build a [1, N] token tensor for the teacher-forcing pass over the
+  // committed prefix at the start of each chunk. Type comes from the decoder
+  // session so it matches the graph's declared input dtype.
+  auto shape = std::array<int64_t, 2>{1, n};
+  auto tok_type = moonshine_model_.session_info_.GetInputDataType(
+      moonshine_model_.moonshine_config_.decoder_kv.in_token);
+  auto tok_tensor = OrtValue::CreateTensor(moonshine_model_.allocator_cpu_, shape, tok_type);
+  std::memcpy(tok_tensor->GetTensorMutableData<int64_t>(),
```
baijumeswani, line 906: This tensor is allocated with the graph-declared token type, but the data is always copied as `int64_t`. An int32 token input would receive an oversized write. Please either validate that all Moonshine integer inputs are int64 when the model loads, or write values using their declared type. The same check is needed for `sample_len`, `frame_count`, and `pos_offset`.

baijumeswani (commented):
Thanks for adding Moonshine streaming ASR support. The overall processor/state design fits the existing streaming API well. I found one case that can lose audio at segment boundaries, plus a few integration and compatibility gaps. I added clear fix suggestions inline.

src/models/moonshine_streaming.cpp
nenad1002: Not a bug

baijumeswani (approved):

PR #2308, opened 2026-07-20
for vlm model, add AppendInputs for multi-turn chat

lishixlnx:
For current oga, for vlm model, the image input is only allowed in first-turn conversation via SetInputs.
In following chat, only text message is allowed to input via the function of AppendTokens().
But this is not the limitation of vlm model, but the oga platform.

This PR extend the oga function by adding new API of AppendInputs similar to AppendTokens but allow input images in following chat.  
Currently, it only supports qwen 3.5 vl model (other vlm model not test yet)

### DIFF
--- a/src/generators.cpp
+++ b/src/generators.cpp
@@ -557,6 +557,42 @@ void Generator::SetInputs(const NamedTensors& named_tensors) {
   }
 }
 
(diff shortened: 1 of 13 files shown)


lishixlnx:
> @microsoft-github-policy-service agree [company="{your company}"]

@microsoft-github-policy-service agree company="AMD"

PR #2314, opened 2026-07-21
Add n-gram speculative decoding implementation (v0)

samsat701:
### Description

This PR adds v0 support for n-gram speculative decoding to ONNX Runtime GenAI. It builds on the shared speculative decoding lifecycle introduced in #2233 and expanded in #2287.

Unlike draft-model speculative decoding, n-gram decoding does not require a second model. It uses repeated token patterns in the already committed sequence to propose likely continuations, then verifies those tokens with the target model in a batched forward pass.

A round consists of:

1. Lookup - match the current (N-1)-token suffix against committed history
2. Propose - copy up to K tokens from the most recent matching continuation
3. Verify - run the target model over the proposal sequence in one multi-token decode pass
4. Accept / correct / bonus:
   - Greedy: accept each proposal token while it matches the target argmax
   - Sampling: sample from each target distribution and accept while the sampled token matches the deterministic proposal
   - First rejection: commit the target-selected correction token
   - All proposals accepted: commit a bonus token from the trailing target distribution
5. Re-anchor - reconcile the target KV cache with the committed sequence when required

Each successful speculative round commits between 1 and K+1 tokens from one target verification call. Tokens are buffered internally and remain externally compatible with the existing one-token-per-GenerateNextToken() API.

### Motivation and context

Autoregressive generation normally performs one target-model execution for every generated token. Repeated phrases, formatting patterns, code structures, and other locally predictable sequences can often be recovered directly from previously generated context.

N-gram speculative decoding uses this history as a model-free deterministic proposer. When continuations are accepted, multiple output tokens are produced from one target execution without:

- Loading a separate draft model
- Running draft-model forward passes
- Maintaining a second model's KV cache
- Changing the target output distribution

Performance depends on repetition in the prompt/output, proposal length, acceptance rate, model behavior, and execution provider.

### Public API and configuration

N-gram decoding is configured through the existing speculative options:

```python
params.set_speculative_options(
    ngram_size=3,
    max_draft_tokens=4,
)
```

It can also be configured in genai_config.json:

```json
{
  "speculative": {
    "ngram_size": 3,
    "max_draft_tokens": 4
  }
}
```

- ngram_size=0 disables n-gram decoding
- Enabled values must be between 2 and 16
- max_draft_tokens controls the maximum number of proposed tokens per round
- N-gram decoding and draft-model speculative decoding cannot be enabled together
- ngram_size is exposed through the existing C/C++ and Python speculative-option APIs

#### Incremental exact n-gram lookup

This PR adds NGramLookup, which:

- Indexes only verified and committed tokens
- Uses exact token vectors as keys, with equality checks preserving collision safety
- Retains the most recent occurrence of each key
- Proposes a contiguous continuation from the most recent matching history
- Indexes only newly committed suffixes instead of rescanning the full sequence each round
- Clears and rebuilds lookup state when generation rewinds
- Never inserts unverified proposal tokens into lookup history

A lookup miss uses the shared standard single-token decoding path.

#### N-gram decoding strategy

NGramDecodingStrategy integrates lookup-based proposals with the existing speculative lifecycle:

- Multi-token target verification
- Greedy and sampling acceptance
- Correction and bonus tokens
- KV-cache rewind and re-anchoring
- Internal token buffering
- Append, rewind, resume, and external-logits lifecycle handling
- Existing speculative statistics

The target model remains the only model executed by this strategy so draft_forward_passes remains zero.

#### Deterministic proposal verification

The shared speculative implementation now represents proposal semantics explicitly:

- Greedy draft-model proposal
- Probabilistic draft-model proposal
- Deterministic n-gram proposal

For n-gram sampling, each proposal is checked against an actual draw from the processed target distribution. The implementation applies:

- top_k
- top_p
- temperature
- min_length
- repetition_penalty

If the target draw matches the deterministic proposal, the proposal is accepted. Otherwise, that draw becomes the correction token. This preserves the target distribution without constructing a dense one-hot draft distribution.

#### Canonical RNG ownership

RNG ownership is centralized in the owning Generator; engine requests maintain their corresponding request-owned RNG.

Standard decoding, draft-model speculative decoding, and n-gram decoding use the same canonical stream. This prevents an n-gram lookup miss from unexpectedly switching RNG state or behavior when it falls back to standard decoding.

The implementation also ensures that:

- Rows after the first rejection do not consume random draws
- A bonus draw occurs only after full proposal acceptance
- RNG state associated with buffered tokens is applied as those tokens become externally visible
- Discarded speculative work does not leave the canonical stream advanced for unobserved tokens
- CUDA sampling is seeded from the owning generator/request RNG rather than maintaining an independently seeded stream

CPU and CUDA retain different sampling implementations (fixed-seed equality is only expected within the same sampling path, not across different CPU and CUDA samplers).

#### Shared speculative lifecycle updates

The shared speculative strategy was generalized so both draft-model and n-gram proposers can reuse:

- Target-row selection
- Greedy and sampling verification
- Correction and bonus handling
- Token buffering and emission
- Cache reconciliation
- Append, rewind, and external-logits invalidation
- Statistics collection

A CUDA nonzero-rewind boundary issue found during validation was also fixed so CUDA preserves the same replay token as CPU when resuming generation.

### Statistics

N-gram decoding uses the existing speculative statistics API, including:

- Speculative rounds
- Completed and interrupted rounds
- Proposed, evaluated, and accepted tokens
- Correction and bonus tokens
- Buffered, emitted, and discarded tokens
- Target and draft forward passes
- Acceptance rate
- Proposal, verification, and re-anchor timing

For n-gram decoding, the draft-related token counters describe lookup proposals, while draft_forward_passes remains zero because no draft model is executed.

### Supported v0 scope

v0 supports:

- Decoder-only text models (with rewindable KV state and full multi-token logits)
- batch_size=1, num_beams=1 and num_return_sequences=1
- Greedy and regular non-beam sampling modes
- Top-k, top-p, temperature and random-seed handling
- Minimum length and repetition penalty
- EOS and maximum-length handling
- Shared and non-shared KV buffers
- Prefill chunking through existing model-state abstractions
- Same-generator continuous decoding, including append, rewind/resume and GetLogits/SetLogits
- fp32 and fp16 target execution

### Current limitations

v0 exclusions include:

- Guidance 
- Batch size greater than one
- Beam search and multiple return sequences
- Engine and continuous-batching integration
- Multimodal, audio, and pipeline models
- Sliding-window KV caches
- Hybrid and model-managed state
- Pruned last-token-only logits
- NPU/fixed-shape runtimes that cannot accept variable-width multi-token decode inputs or rewind state correctly
- Fixed n-gram order and maximum proposal length for the generator lifetime; adaptive K and rejection cooldown not yet implemented
- Chained lookup and multiple n-gram-order fallback not yet implemented
- N-gram specific instrumentation statistics not yet implemented

### Benchmarking

Benchmarking scripts and model-backed performance evaluation are intentionally maintained separately from this implementation PR in the following branch:

https://github.com/microsoft/onnxruntime-genai/tree/t-sshrimali/speculative-decoding-benchmarking

### Related PRs

- Base speculative decoding v0: #2233
- Base speculative decoding v1: #2287


src/cuda/search_cuda.cpp
```diff
@@ -40,16 +40,12 @@ void GreedySearch_Cuda::EnsureSamplingData() {
...
-                                       ? params_->search.random_seed
-                                       : std::random_device{}();
-
   // Allocate a single buffer for all sampling data
   size_t sampling_buffer_size = cuda::SamplingData::CalculateTotalSize(params_->search.batch_size, params_->config.model.vocab_size, GetStream());
   sampling_buffer_ = params_->p_device->Allocate<uint8_t>(sampling_buffer_size);
 
   // Create SamplingData with the externally allocated buffer
-  samplingdata_ = std::make_unique<cuda::SamplingData>(random_seed, params_->search.batch_size, params_->config.model.vocab_size, GetStream(),
+  samplingdata_ = std::make_unique<cuda::SamplingData>(0, params_->search.batch_size, params_->config.model.vocab_size, GetStream(),
```
kunal-vaishnavi: Can you document why a random seed is being replaced with 0 here?

src/cuda/search_cuda.cpp
```diff
@@ -389,7 +392,7 @@ void BeamSearch_Cuda::AppendTokens(DeviceSpan<int32_t>& next_tokens) {
 void GreedySearch_Cuda::RewindTo(size_t index) {
   ResetDone();
   if (index > 0)
-    cuda::Launch_GetLastTokens(next_tokens_.data(), sequences_.GetSequences().Span().data(), static_cast<int>(params_->BatchBeamSize()), static_cast<int>(index), sequences_.max_length_, GetStream());
+    cuda::Launch_GetLastTokens(next_tokens_.data(), sequences_.GetSequences().Span().data(), static_cast<int>(params_->BatchBeamSize()), static_cast<int>(index + 1), sequences_.max_length_, GetStream());
```
kunal-vaishnavi, line 400: Can you document why +1?

src/models/decoder_only.cpp
```diff
@@ -37,18 +37,23 @@ DeviceSpan<float> DecoderOnly_State::Run(int total_length, DeviceSpan<int32_t>&
...
-    UpdateInputsOutputs(next_tokens, next_indices, total_length);
-    if (model_.config_->model.decoder.run_options.has_value()) {
-      State::SetRunOptions(model_.config_->model.decoder.run_options.value());
-    }
+  }
 
-    // Graph capture enabled for token generation case, allowing it to repeat the same graph for each token.
-    bool graph_capture_this_run = params_->use_graph_capture && input_ids_.GetShape()[1] == 1;
-    State::Run(*model_.session_decoder_, graph_capture_this_run);
+  return RunUnchunked(total_length, next_tokens, next_indices);
```
kunal-vaishnavi: Why is the default `Run` method insufficient? That is the unchunked version.

src/config.cpp
```diff
@@ -1358,6 +1358,42 @@ struct Speculative_Element : JSON::Element {
...
+    } else if (name == "ngram_size") {
+      const int ngram_size = SafeDoubleToInt(JSON::Get<double>(value), name);
+      if (ngram_size != 0 && (ngram_size < 2 || ngram_size > kMaxK))
+        throw std::runtime_error(
+            "speculative.ngram_size must be 0 or between 2 and " + std::to_string(kMaxK) +
+            ". Got: " + std::to_string(ngram_size) + ".");
+      v_.ngram_size = ngram_size;
+    } else if (name == "ngram_chained_lookup_bool") {
+      const int ngram_chained_lookup_bool = SafeDoubleToInt(JSON::Get<double>(value), name);
+      if (ngram_chained_lookup_bool != 0 && ngram_chained_lookup_bool != 1)
```
kunal-vaishnavi: Let's remove the `bool` suffixes from these variables. We can use the same approach that other settings such as `past_present_share_buffer` use in the `search` section.

src/config.cpp
```diff
@@ -1358,6 +1358,42 @@ struct Speculative_Element : JSON::Element {
...
+            std::to_string(ngram_chained_lookup_bool) + ".");
+      v_.ngram_chained_lookup_bool = ngram_chained_lookup_bool;
+    } else if (name == "adaptive_k_bool") {
+      const int adaptive_k_bool = SafeDoubleToInt(JSON::Get<double>(value), name);
+      if (adaptive_k_bool != 0 && adaptive_k_bool != 1)
+        throw std::runtime_error(
+            "speculative.adaptive_k_bool must be 0 or 1. Got: " +
+            std::to_string(adaptive_k_bool) + ".");
+      v_.adaptive_k_bool = adaptive_k_bool;
+    } else if (name == "adaptive_k_min") {
```
kunal-vaishnavi: Can we use `min_adaptive_k` instead to mirror variables such as `min_length` and `max_length`?

src/n_gram_decoding_strategy.h
```diff
@@ -0,0 +1,98 @@
...
+#include "speculative_decoding_strategy.h"
+
+namespace Generators {
+
+struct GeneratorParams;
+struct Model;
+
+struct NGramDecodingCapabilities {
+  int batch_size{1};
+  int num_beams{1};
```
kunal-vaishnavi: We should be able to read these params (batch size, num beams, num return sequences) from the search section instead of storing them here.

src/sampling_distribution.h
```diff
@@ -188,6 +190,22 @@ struct SampledCategorical {
...
+                                      std::mt19937& rng) {
+  if (indices.empty() || indices.size() != probs.size())
+    throw std::invalid_argument(
+        "Categorical sampling requires equally sized, non-empty token and probability arrays.");
+
+  std::discrete_distribution<int> distribution(probs.begin(), probs.end());
+  return indices[static_cast<size_t>(distribution(rng))];
+}
+
+inline int32_t SampleCategoricalToken(const SampledCategorical& categorical,
```
kunal-vaishnavi: Can we simplify this to one API instead of maintaining two APIs for `SampleCategoricalToken`?

src/speculative_decoding_strategy.h
```diff
@@ -15,17 +20,247 @@
...
+    effective_k_ = min_k_;
+    estimates_ = {};
+    probe_origin_k_ = 0;
+    probe_observations_ = 0;
+    stable_observations_ = 0;
+    probe_cooldown_ = 0;
+  }
+
+ private:
+  static constexpr float kEwmaAlpha = 0.25f;
```
kunal-vaishnavi, line 122: Can you document how these values were set?

src/speculative_decoding_strategy.cpp
```diff
@@ -517,62 +665,84 @@ void SpeculativeDecodingStrategy::RunRound(Generator& g) {
...
+    final_token = result.final_token;
+    round_.pending_rng_states.assign(
+        std::make_move_iterator(states_after_draw.begin()),
+        std::make_move_iterator(states_after_draw.end()));
+    if (result.used_bonus)
+      bonuses_++;
+    else
+      corrections_++;
+  } else {
+    for (int i = 0; i < K; i++) {
```
kunal-vaishnavi, line 718: From Copilot:

## RNG Stream Reproducibility Issue in Draft-Sampling Path

In the `else` branch (lines 688–735), when verifying draft-model proposals with sampling, RNG draws are consumed directly from `g.rng_` during the acceptance loop:

```cpp
accepted = (uni(g.rng_) < ComputeAcceptProb(p_t, p_d));
```

and later for corrections/bonuses:

```cpp
final_token = dist(g.rng_);
```

If this round is interrupted later (via `Append`, `Rewind`, or `SetLogits`) and its buffered tokens are discarded, **the canonical RNG stream will have advanced for tokens that were never emitted**. This violates the stated design contract: *"advance RNG only as tokens become externally visible."*

The deterministic-proposal path (lines 673–686) avoids this by capturing RNG state snapshots in `pending_rng_states` and replaying them during emission. The draft-sampling path should do the same to preserve reproducibility across interrupted rounds.

**Suggested fix:** Capture the RNG state after each accept/rejection decision in the draft-sampling loop (similar to the deterministic path), and restore those states as tokens are emitted in `DrainOne`.

src/config.h
```diff
@@ -466,8 +466,14 @@ struct Config {
 
   struct Speculative {
     // Four is a conservative default that amortizes target verification without excessive draft
-    // work; the best value depends on the model pair and execution provider.
+    // work; the best value depends on repetition/acceptance and the execution provider. When
```
kunal-vaishnavi: From Copilot:

## Clarify `max_draft_tokens` Behavior Under Adaptive K

The comment for `max_draft_tokens` does not clarify its semantics when adaptive K is enabled. Specifically:

- Is `max_draft_tokens` a **hard cap** that adaptive K cannot exceed?
- Or does it become the **initial probe width**, with adaptive K allowed to grow beyond it (with a separate, undocumented hard cap)?

When `adaptive_k_bool` is enabled, the effective K can exceed the configured `max_draft_tokens` value. The current comment reads:

> "Four is a conservative default that amortizes target verification without excessive draft work; the best value depends on the model pair and execution provider."

This implies a constraint, but doesn't specify whether it's a hard limit or a starting point. **Please clarify in-code:** does adaptive K respect `max_draft_tokens` as a ceiling, or is there a separate maximum when adaptive probing is active?

src/speculative_decoding_strategy.cpp
```diff
@@ -719,23 +902,26 @@ void SpeculativeDecodingStrategy::FinalizeRound(Generator& g) {
...
 }
 
-// Runs one guidance round over the K draft tokens - mask the target with the grammar, accept/verify
-// each token, commit it, and add forced tokens. Greedy commits the draft's matching token; sampling
-// uses speculative sampling on the masked distributions. Caches are re-anchored in FinalizeRound.
+// Runs one guidance round over the K proposal tokens - mask the target with the grammar,
+// accept/verify each token, commit it, and add forced tokens. Draft-model sampling uses p/q
+// verification; deterministic sampling draws from the masked target and compares that draw with
+// the proposal. Caches are re-anchored in FinalizeRound.
 void SpeculativeDecodingStrategy::RunGuidanceRound(Generator& g, const Proposal& proposal,
```
kunal-vaishnavi, line 944: From Copilot:

## Consider Simplifying Guidance RNG State Capture Logic

The guidance round implementation (`RunGuidanceRound`, lines 917–1178) is ~260 lines that interleave three concerns:

1. **Grammar filtering**: mask proposal tokens through the constrained logits processor
2. **Forced-token handling**: accept grammar-forced continuations and feed them to the next position
3. **RNG state capture** (for deterministic n-gram proposals): capture the RNG state after each sampling draw (lines 1089, 1107) and correlate it with committed tokens

This interweaving creates subtle coupling. For example:

- Line 1088 calls `SampleTargetToken(target_selection, g.rng_)`, which mutates `g.rng_`
- Line 1089 immediately saves the post-draw state: `deterministic_rng_states.push_back(g.rng_)`
- Lines 1105–1108: forced tokens that follow a rejection each also append their RNG state

The interaction between rejection boundaries, forced-token loops, and RNG snapshots is intricate and hard to verify at a glance. A regression here could silently corrupt the RNG stream that's restored in `DrainOne`.

**Suggestion:** Extract deterministic sampling verification (lines 1087–1111) into a dedicated helper function—something like `VerifyGuidanceTokenDeterministic(...)` that encapsulates the accept/reject/forced-token logic and returns both the committed tokens and their corresponding RNG states as a bundle. This would isolate the RNG coupling and make it easier to test and reason about.

src/speculative_decoding_strategy.cpp
```diff
@@ -881,19 +1078,44 @@ void SpeculativeDecodingStrategy::RunGuidanceRound(Generator& g, const Proposal&
...
         if (i == 0) {
           committed.push_back(ttok);
-          stats_.correction_tokens++;
+          corrections_++;
           if (!IsEosToken(eos_ids, ttok))
             for (int32_t fwd : CommitGuidanceToken(proc, ttok)) committed.push_back(fwd);
         }
         break;
       }
+    } else if (deterministic_sampling) {
```
kunal-vaishnavi, line 1117: From Copilot:

## RNG State Capture Complexity in Guidance Deterministic-Sampling Path

The guidance round's deterministic-sampling branch (lines 1087–1111) interleaves RNG state capture with forced-token handling, creating subtle alignment risks.

**The concern:**

When a token is accepted after deterministic sampling (lines 1088–1100), the code captures RNG state:

```cpp
const int32_t target_token = SampleTargetToken(target_selection, g.rng_);
deterministic_rng_states.push_back(g.rng_);  // Line 1089 - capture after draw
committed.push_back(target_token);
if (target_token == proposal.tokens[i]) {
  verify_prefix++;
  n_direct++;
  // ... process acceptance ...
  for (int32_t fwd : CommitGuidanceToken(proc, target_token))
    pending_forced.push_back(fwd);
}
```

Then, if the token is **rejected**, forced tokens from the grammar follow (lines 1105–1108):

```cpp
} else {
  rejected = true;
  corrections_++;
  if (!IsEosToken(eos_ids, target_token)) {
    for (int32_t fwd : CommitGuidanceToken(proc, target_token)) {
      committed.push_back(fwd);
      deterministic_rng_states.push_back(g.rng_);  // Line 1107 - capture for forced tokens
    }
  }
  break;
}
```

**Why it's risky:**

1. **RNG state vs. token mismatch:** The code pushes one RNG state after the initial draw (line 1089), but **for each forced token that follows a rejection, it pushes g.rng_ again** (line 1107). If a forced token doesn't consume RNG (no sampling), the state captured at line 1107 may be stale or misaligned.

2. **Forced tokens from accepted proposals:** When an accepted token triggers forced tokens (line 1099–1100), **no RNG state is captured for those forced tokens**. This creates an asymmetry: forced tokens after rejection get RNG snapshots, but forced tokens after acceptance don't.

3. **Validation happens after buffering:** The check at lines 1156–1158 validates that `deterministic_rng_states.size() == committed.size()`, but by then tokens have already been pushed. If there's a mismatch, it throws—but the round's state may already be partially corrupted.

**Potential failure mode:**

If forced tokens are emitted in `DrainOne()` and their corresponding RNG states are missing or stale, the canonical RNG stream will diverge from the committed sequence. This can silently break reproducibility.

**Recommendation:**

Add **clarifying comments** that document the RNG-to-token correspondence invariant:
- "One RNG state per committed token; forced tokens share the RNG state of the token that triggered them."
- Or: "Each committed token gets a fresh RNG snapshot, including forced tokens."

Then **add a stricter validation** before returning from `RunGuidanceRound()`:
- Ensure every committed token (including forced tokens) has a corresponding RNG state OR
- Ensure forced tokens explicitly inherit the RNG state of their triggering token (not a new snapshot).

Finally, **extend test coverage** for the deterministic-sampling guidance path to verify:
- Acceptance + forced-token sequences
- Rejection + forced-token sequences
- RNG reproducibility when rounds are interrupted mid-forced-sequence

This is not a blocking bug (the check at 1156–1158 will catch size mismatches), but the complexity warrants defensive instrumentation to prevent future regressions.

src/speculative_decoding_strategy.cpp
```diff
@@ -604,10 +781,15 @@ void SpeculativeDecodingStrategy::DrainOne(Generator& g) {
 
   const int32_t tok = round_.pending.front();
   round_.pending.pop_front();
+  if (!round_.pending_rng_states.empty()) {
```
kunal-vaishnavi: From Copilot:

## Add Defensive RNG State Validation in DrainOne

Currently, RNG state capture is validated at the end of round construction (lines 742–745):

```cpp
if (!round_.pending_rng_states.empty() &&
    round_.pending_rng_states.size() != round_.pending.size())
  throw std::runtime_error(
      "N-gram decoding captured an inconsistent number of sampling RNG states.");
```

This catches mismatches **after tokens have already been buffered**. Add an earlier, defensive check in `DrainOne()` when tokens are actually being emitted, to catch corruptions sooner:

```cpp
void SpeculativeDecodingStrategy::DrainOne(Generator& g) {
  // ... existing checks ...
  
  const int32_t tok = round_.pending.front();
  round_.pending.pop_front();
  
  // NEW: Defensive check - if we're capturing RNG states, 
  // every emitted token must have one.
  if (!round_.pending_rng_states.empty()) {
    if (round_.pending_rng_states.empty())
      throw std::runtime_error(
          "N-gram decoding: RNG state missing for emitted token. "
          "pending_rng_states was non-empty at round start but is now exhausted.");
  }
  
  if (!round_.pending_rng_states.empty()) {
    g.rng_ = round_.pending_rng_states.front();
    round_.pending_rng_states.pop_front();
  }
  // ... rest of method ...
}
```

This catches two classes of bugs:

1. **RNG state underrun:** If a code path discards an RNG state without a corresponding token, the mismatch is caught at emission rather than silently causing RNG divergence.

2. **Silent async divergence:** If a future change removes RNG state capture from some code path (e.g., guidance forced tokens), the check will fail immediately instead of producing silent reproducibility bugs.

The check is cheap (one condition per emitted token) and dramatically improves debuggability of RNG-related issues.

src/speculative_decoding_strategy.cpp
kunal-vaishnavi, line 1117: From Copilot:

## Extract Deterministic-Sampling Logic from RunGuidanceRound

The `RunGuidanceRound` method is already ~260 lines. The deterministic-sampling branch (lines 1087–1111) compounds this complexity by interleaving:

- RNG state capture (`g.rng_` sampling)
- Token acceptance/rejection decisions
- Forced-token handling from the grammar processor
- RNG state snapshots for each committed token

**Current structure (hard to follow):**

```cpp
} else if (deterministic_sampling) {
  const int32_t target_token = SampleTargetToken(target_selection, g.rng_);
  deterministic_rng_states.push_back(g.rng_);
  committed.push_back(target_token);
  if (target_token == proposal.tokens[i]) {
    verify_prefix++;
    n_direct++;
    // ... acceptance path ...
    for (int32_t fwd : CommitGuidanceToken(proc, target_token))
      pending_forced.push_back(fwd);
  } else {
    rejected = true;
    corrections_++;
    if (!IsEosToken(eos_ids, target_token)) {
      for (int32_t fwd : CommitGuidanceToken(proc, target_token)) {
        committed.push_back(fwd);
        deterministic_rng_states.push_back(g.rng_);
      }
    }
    break;
  }
}
```

**Suggested refactor:**

Extract this into a helper function:

```cpp
bool AcceptGuidanceTokenDeterministic(
    ConstrainedLogitsProcessor& proc,
    const TargetTokenSelection& target_selection,
    int32_t proposal_token,
    std::mt19937& rng,
    std::vector<int32_t>& committed,
    std::vector<std::mt19937>& rng_states,
    int& verify_prefix,
    int& n_direct,
    int& corrections,
    bool& rejected) {
  
  const int32_t target_token = SampleTargetToken(target_selection, rng);
  rng_states.push_back(rng);
  committed.push_back(target_token);
  
  if (target_token == proposal_token) {
    verify_prefix++;
    n_direct++;
    for (int32_t fwd : CommitGuidanceToken(proc, target_token)) {
      // Handle forced tokens...
    }
    return true; // accepted
  } else {
    rejected = true;
    corrections++;
    // Handle rejection + forced tokens...
    return false; // rejected
  }
}
```

Then call it from `RunGuidanceRound`:

```cpp
} else if (deterministic_sampling) {
  if (!AcceptGuidanceTokenDeterministic(
      proc, target_selection, proposal.tokens[i], g.rng_,
      committed, deterministic_rng_states, verify_prefix, n_direct,
      corrections, rejected)) {
    break; // rejection ends the round
  }
}
```

**Benefits:**

1. **Reduces visual complexity:** `RunGuidanceRound` becomes ~20 lines shorter and easier to follow.
2. **Isolates RNG state logic:** All RNG capture happens in one place, making it easier to audit for correctness.
3. **Testability:** The helper can be unit-tested independently of the full guidance round.
4. **Maintainability:** Future changes to forced-token or RNG handling are localized to one function.

This is a structural improvement, not a functional fix, but it significantly improves code reviewability and future maintainability.

src/n_gram_lookup.h
```diff
@@ -0,0 +1,103 @@
...
+#include <algorithm>
+#include <cstdint>
+#include <span>
+#include <stdexcept>
+#include <unordered_map>
+#include <vector>
+
+namespace Generators {
+
+class NGramLookup {
```
kunal-vaishnavi, line 21: From Copilot:

## Clarify NGramLookup Memory and Collision-Safety Semantics

Add documentation to the class comment explaining memory usage and collision handling:

```cpp
class NGramLookup {
 public:
  // Incremental exact n-gram lookup using token vector keys.
  // 
  // Memory model: Uses exact token sequences (not hashed) as keys in an unordered_map.
  // Each new n-gram suffix appended is indexed. Collisions are NOT compressed—if the same
  // n-gram pattern appears multiple times, only the most recent position is retained.
  // Memory complexity: O(sequence_length * ngram_size) for the history and occurrence map.
  // Note: A full re-index only occurs on rewind (e.g., during continuous decoding with
  // Append/SetLogits), so incremental append dominates during normal generation.
  //
  // Proposal guarantee: Returns a contiguous suffix from the most recent matching history.
  // If no match is found, returns an empty proposal (caller falls back to standard decoding).
  explicit NGramLookup(int ngram_size)
      : ngram_size_{ngram_size}, key_length_{static_cast<size_t>(ngram_size - 1)} {
    if (ngram_size < 2)
      throw std::runtime_error("NGramLookup requires ngram_size >= 2.");
  }
  // ... rest of class ...
```

**Why this matters:**

Users and maintainers need to understand:

1. **Memory overhead:** O(sequence_length * ngram_size) can be significant for long sequences with large n-gram sizes. This helps operators understand when n-gram decoding is feasible on memory-constrained hardware.

2. **Collision semantics:** "Only most recent position retained" is intentional and correct, but non-obvious. Without documentation, someone might assume collisions are bucketed (multiple positions per key) or compressed somehow.

3. **Performance profile:** The note about re-index rarity justifies the incremental append design. Without this, readers might worry that every rewind is expensive.

4. **Proposal behavior:** Explicitly stating "contiguous suffix" sets expectations and prevents misunderstanding about chained vs. non-chained proposals.

This documentation helps future maintainers reason about trade-offs (e.g., "can we use a smaller n-gram to save memory?") and helps users make deployment decisions.

src/models/decoder_only.cpp
samsat701: Run() is unchunked only when chunking is disabled/input fits within configured chunk_size. Otherwise it calls RunWithChunking() (returned logits contain only final chunk). 

Spec verify step needs one logits row for every proposed token (so it must bypass prefill chunking). 

This can be improved in a future PR.

src/cuda/search_cuda.cpp
samsat701: Added a comment explaining this - https://github.com/microsoft/onnxruntime-genai/pull/2314/changes/f548094b0edbe9f7dc681279e86cf0f9e48ed24d. 

RNG ownership now belongs to the Generator/engine Request. SamplingData still needs initialized cuRAND state when it is created so 0 is used only as an initial value. Before stochastic sampling -> state is replaced seed from the caller-owned RNG.

src/cuda/search_cuda.cpp
samsat701: Added a comment explaining this - https://github.com/microsoft/onnxruntime-genai/pull/2314/changes/f548094b0edbe9f7dc681279e86cf0f9e48ed24d. 

RewindTo(index) preserves sequence[index] so generation can replay that boundary token. 

Launch_GetLastTokens reads sequence_length - 1 -> passing index + 1 selects the intended token.

src/config.cpp
samsat701: Fixed in commit- https://github.com/microsoft/onnxruntime-genai/commit/f548094b0edbe9f7dc681279e86cf0f9e48ed24d. 

Settings are now real Boolean fields named ngram_chained_lookup, adaptive_k, and cooldown. I also added speculative Boolean setter/getter APIs across the C, C++, Python, C#, Java, and Objective-C bindings and updated parsing and tests.

src/config.cpp
samsat701: Fixed in commit- https://github.com/microsoft/onnxruntime-genai/commit/f548094b0edbe9f7dc681279e86cf0f9e48ed24d.

Renamed to min_adaptive_k. Now follows the same min_* naming style as min_length.

src/n_gram_decoding_strategy.h
samsat701: Fixed in commit- https://github.com/microsoft/onnxruntime-genai/commit/f548094b0edbe9f7dc681279e86cf0f9e48ed24d.

Removed batch_size, num_beams, and num_return_sequences from NGramDecodingCapabilities. Validation now reads these values directly from Config::Search.

src/sampling_distribution.h
samsat701: Fixed in - https://github.com/microsoft/onnxruntime-genai/commit/f548094b0edbe9f7dc681279e86cf0f9e48ed24d.

The span-based API is now the single implementation and standard search passes indices + probs directly.

src/speculative_decoding_strategy.h
samsat701: Documented in https://github.com/microsoft/onnxruntime-genai/commit/f548094b0edbe9f7dc681279e86cf0f9e48ed24d.

src/config.h
samsat701: Addressed in https://github.com/microsoft/onnxruntime-genai/pull/2314/changes/f548094b0edbe9f7dc681279e86cf0f9e48ed24d.

src/speculative_decoding_strategy.cpp
samsat701: Addressed in https://github.com/microsoft/onnxruntime-genai/pull/2314/changes/c93ff8ddd5f66175f43f09c7dc36d952adbe9664. 

The draft-sampling path no longer advances g.rng_ while it precomputes a buffered round. It uses the round-local RNG and records one post-output checkpoint per pending token. DrainOne advances the canonical stream from those checkpoints (discarded buffered outputs do not commit their later RNG work).

src/speculative_decoding_strategy.cpp
samsat701: Fixed in - https://github.com/microsoft/onnxruntime-genai/pull/2314/changes/30f8ca2bbb58c878f8753f2933dbde0fff0f8ee9. 

Extracted the deterministic guidance sampling and grammar-commit work into VerifyDeterministicGuidanceToken.

src/speculative_decoding_strategy.cpp
samsat701: Fixed in - https://github.com/microsoft/onnxruntime-genai/pull/2314/changes/30f8ca2bbb58c878f8753f2933dbde0fff0f8ee9, https://github.com/microsoft/onnxruntime-genai/pull/2314/changes/c93ff8ddd5f66175f43f09c7dc36d952adbe9664. Every queued output in a sampling round now has exactly one RNG checkpoint. Grammar-forced tokens consume no random draw -> inherit the unchanged checkpoint of the sampled token that triggered them (rule is used consistently for accepted/rejected paths).

src/speculative_decoding_strategy.cpp
samsat701: Fixed in - https://github.com/microsoft/onnxruntime-genai/commit/30f8ca2bbb58c878f8753f2933dbde0fff0f8ee9, https://github.com/microsoft/onnxruntime-genai/commit/c93ff8ddd5f66175f43f09c7dc36d952adbe9664. 

RoundState records whether the round uses RNG checkpoints. More reliable than inferring the mode from whether the checkpoint queue is empty as empty queue cannot distinguish a non-sampling round from an underrun.

src/speculative_decoding_strategy.cpp
samsat701: Fixed in - https://github.com/microsoft/onnxruntime-genai/commit/30f8ca2bbb58c878f8753f2933dbde0fff0f8ee9, https://github.com/microsoft/onnxruntime-genai/commit/c93ff8ddd5f66175f43f09c7dc36d952adbe9664. 

Sampling/grammar-commit operation isolated in VerifyDeterministicGuidanceToken and token/checkpoint insertion is centralized in AppendGuidanceSamplingOutput.

src/n_gram_lookup.h
samsat701: Fixed in - https://github.com/microsoft/onnxruntime-genai/commit/30f8ca2bbb58c878f8753f2933dbde0fff0f8ee9.

Class comment now documents incremental indexing, latest-occurrence replacement, reset behavior, contiguous and chained proposal behavior, and O(sequence_length * ngram_size) worst-case storage.

Also clarifies collision safety.

src/speculative_decoding_strategy.cpp
```diff
@@ -517,62 +665,84 @@ void SpeculativeDecodingStrategy::RunRound(Generator& g) {
...
+      bool accepted = false;
+      const TargetTokenSelection& target =
+          (i == 0) ? pos0_selection : target_selections[static_cast<size_t>(i - 1)];
+      if (target_uses_greedy) {
+        accepted = (target.greedy_token == proposal.tokens[i]);
+      } else {
+        const float p_t = GetTargetTokenProbability(target, proposal.tokens[i]);
+        const float p_d = proposal.probs[i][proposal.tokens[i]];
+        accepted = (uni(g.rng_) < ComputeAcceptProb(p_t, p_d));
+      }
```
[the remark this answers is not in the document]
samsat701: Fixed in https://github.com/microsoft/onnxruntime-genai/pull/2314/changes/30f8ca2bbb58c878f8753f2933dbde0fff0f8ee9, https://github.com/microsoft/onnxruntime-genai/pull/2314/changes/c93ff8ddd5f66175f43f09c7dc36d952adbe9664. 

Draft-sampling rounds now use a round-local copy of the Generator RNG. Acceptance checks and correction/bonus sampling doesn't advance g.rng_ while the outputs are buffered.

Interrupting a round discards checkpoints for unseen outputs (preserves reproducibility).

kunal-vaishnavi:
Here are some Copilot-generated review comments.

---

## **BLOCKING ISSUE 1: RNG Stream Reproducibility in Draft-Sampling Path**

`src/speculative_decoding_strategy.cpp`
Lines: 688–735

````markdown
## RNG Stream Reproducibility Issue in Draft-Sampling Path

In the non-greedy draft-model sampling branch (lines 688–735), RNG draws for accept/reject decisions and correction/bonus sampling are consumed directly from `g.rng_` while multiple tokens are buffered internally:

```cpp
accepted = (uni(g.rng_) < ComputeAcceptProb(p_t, p_d));
```

and later:

```cpp
final_token = dist(g.rng_);
```

**The problem:** If this round is later interrupted via `Append()`, `Rewind()`, or `SetLogits()` and its buffered tokens are discarded, the canonical RNG stream **will already have advanced for tokens that were never emitted**. This violates the stated design contract: *"advance RNG only as tokens become externally visible."*

The deterministic-proposal path (lines 673–686) avoids this by capturing RNG state snapshots in `pending_rng_states` and replaying them during emission in `DrainOne()`. The draft-sampling path should follow the same pattern.

**Suggested fix:**
- Capture the RNG state after each accept/rejection decision in the sampling loop (similar to line 1089 in the guidance path)
- Store these snapshots in a transactional queue (like `pending_rng_states`)
- Restore each snapshot as its corresponding token is emitted in `DrainOne()`

This ensures reproducibility is maintained even when rounds are interrupted mid-emission.
````

**Why this matters:**
The RNG centralization effort only works if *all* code paths respect the contract that RNG state advances synchronously with token emission. If draft-sampling can advance the RNG for unobserved tokens, it breaks determinism across interrupted sessions (e.g., append/resume workflows) and contradicts the PR's stated goal of canonical RNG ownership. This is silent—no exception is thrown—so it could produce subtle, hard-to-debug non-reproducibility bugs.

---

## **BLOCKING ISSUE 2: Ambiguous `max_draft_tokens` Semantics Under Adaptive K**

File: `src/config.h`
Lines: ~150–160 (comment block for `max_draft_tokens`)

````markdown
## Clarify `max_draft_tokens` Behavior When Adaptive K is Enabled

The comment for `max_draft_tokens` does not clarify its semantics when `adaptive_k` is enabled:

- Is `max_draft_tokens` a **hard cap** that adaptive K cannot exceed?
- Or does it become the **initial probe width**, with adaptive K allowed to grow beyond it?
- If the latter, what is the hard maximum?

Current comment:
> "Four is a conservative default that amortizes target verification without excessive draft work; the best value depends on the model pair and execution provider."

This implies a constraint, but doesn't specify whether it's inviolable or a starting point. When `adaptive_k` is `true`, the effective K can exceed `max_draft_tokens`, making this ambiguity a real configuration trap.

**Please clarify in-code:**

```cpp
// max_draft_tokens: Maximum number of tokens to propose per speculative round.
// When adaptive_k is disabled: hard cap on proposal length.
// When adaptive_k is enabled: initial probe width; adaptive K may grow beyond this
// value up to [SPECIFY_HARD_LIMIT_HERE]. Four is conservative; best value depends on
// model pair and execution provider.
int max_draft_tokens = 4;
```

If there is a hard ceiling under adaptive K (e.g., 2x `max_draft_tokens`), document it explicitly so users and operators can make informed configuration choices.
````

**Why this matters:**
Without clear semantics, users might configure `max_draft_tokens=4` expecting a hard memory/latency bound, then find adaptive K silently exceeds it, causing unexpected resource usage. Maintainers could also misunderstand which field to adjust when tuning adaptive behavior. This is a documentation issue now, but if semantics change later (e.g., enforcing a hard cap), the ambiguity could cause silent bugs.

---

## **BLOCKING ISSUE 3: RNG State Capture Complexity in Guidance Deterministic-Sampling Path**

File: `src/speculative_decoding_strategy.cpp`
Lines: 1087–1111 (deterministic-sampling branch in `RunGuidanceRound`)

````markdown
## RNG State Capture Asymmetry in Guidance Deterministic-Sampling Path

The guidance round's deterministic-sampling branch (lines 1087–1111) interleaves RNG state capture with forced-token handling, creating alignment risks:

**Acceptance path (lines 1095–1101):**
```cpp
if (target_token == proposal.tokens[i]) {
  verify_prefix++;
  n_direct++;
  // ... acceptance logic ...
  for (int32_t fwd : CommitGuidanceToken(proc, target_token))
    pending_forced.push_back(fwd);
    // NOTE: No RNG state captured for forced tokens here
}
```

**Rejection path (lines 1105–1108):**
```cpp
} else {
  rejected = true;
  corrections_++;
  for (int32_t fwd : CommitGuidanceToken(proc, target_token)) {
    committed.push_back(fwd);
    deterministic_rng_states.push_back(g.rng_);  // Line 1107
  }
}
```

**The issue:** 
- Forced tokens from a **rejection** each get an RNG state snapshot (line 1107)
- Forced tokens from an **acceptance** do NOT get RNG state snapshots (lines 1099–1100)
- This asymmetry means the RNG-to-token correspondence breaks if acceptance-triggered forced tokens are later emitted

The validation at lines 1156–1158 catches size mismatches *after* tokens are buffered:

```cpp
if (!round_.pending_rng_states.empty() &&
    round_.pending_rng_states.size() != committed.size())
  throw std::runtime_error("...");
```

If there's a mismatch, the round's state is already partially corrupted.

**Fix:**
Document the RNG-to-token invariant clearly:
- Option A: "One RNG state per committed token; forced tokens share the RNG state of their triggering token."
- Option B: "Each committed token, including forced tokens, gets its own RNG snapshot."

Then enforce it consistently:
- If Option A: store only one RNG per rejection (not per forced token), and reference the parent's state during emission
- If Option B: capture RNG states for forced tokens from acceptance too (line 1099–1101)

Add stricter validation *before* round completion to detect asymmetries early.
````

**Why this matters:**
Forced-token handling is intricate. If the RNG-to-token correspondence is silently broken (e.g., because a future change adds forced tokens in a new path), reproducibility will fail without any exception. The validation catches size mismatches, but if a future change *adds* RNG captures somewhere else, the sizes could re-align while the semantics diverge. Explicit documentation of the invariant and symmetric capture logic prevent this class of silent regression.

---

## **IMPROVEMENT 1: Configuration API Naming Inconsistency**

File: `src/config.h`
Lines: (wherever `adaptive_k_bool`, `cooldown_bool`, etc. are defined)

````markdown
## Normalize Boolean Configuration Naming

Variables like `adaptive_k_bool` and `cooldown_bool` use `_bool` suffixes. The existing codebase convention (e.g., `past_present_share_buffer` in the search section) avoids this suffix, treating booleans as ordinary fields.

**Rename to align with convention:**

```cpp
// Before:
bool adaptive_k_bool = false;
bool cooldown_bool = false;

// After:
bool adaptive_k = false;
bool cooldown = false;
```

This maintains consistency across the codebase and reduces redundancy (the type is already explicit in the declaration).
````

**Why this matters:**
Consistency in naming convention reduces cognitive load for developers working across the codebase. The `_bool` suffix is redundant and inconsistent with established patterns. While this is a style issue, inconsistency can lead to confusion and makes future refactoring harder.

---

## **IMPROVEMENT 2: Duplicate `SampleCategoricalToken` API**

File: `src/sampling_distribution.h`
Lines: (wherever both overloads are declared)

````markdown
## Consolidate Duplicate `SampleCategoricalToken` Overloads

Two overloads of `SampleCategoricalToken` are maintained:
1. One accepting an external RNG reference (`std::mt19937&`)
2. One without an RNG parameter (legacy?)

**Suggestion:** Consolidate to a single API that always requires an external RNG:

```cpp
// Keep only this:
int32_t SampleCategoricalToken(
    const float* logits, int vocab_size,
    std::mt19937& rng);

// Remove the no-RNG overload or clearly document which callers use which
```

If some callers genuinely need different behavior, add a well-named parameter (e.g., `use_default_rng=false`) rather than maintaining two separate function signatures. This reduces the maintenance surface and makes the RNG dependency explicit at every call site.
````

**Why this matters:**
Duplicate APIs increase maintenance burden and create implicit behavioral divergence. If the two overloads differ in subtle ways (e.g., which RNG they use), callers might pick the wrong one. Having one clear API with explicit RNG ownership makes the call sites self-documenting and easier to audit for correctness.

---

## **IMPROVEMENT 3: Extract Deterministic-Sampling Logic from RunGuidanceRound**

File: `src/speculative_decoding_strategy.cpp`
Lines: 1087–1111

````markdown
## Extract Deterministic-Sampling Verification into a Helper Function

`RunGuidanceRound` is already ~260 lines. The deterministic-sampling branch (lines 1087–1111) compounds complexity by interleaving:
- RNG state capture
- Acceptance/rejection decisions
- Forced-token handling from the grammar processor
- RNG snapshots for committed tokens

This interweaving makes the function hard to follow and audit for correctness. Extract into a dedicated helper:

```cpp
bool AcceptGuidanceTokenDeterministic(
    ConstrainedLogitsProcessor& proc,
    const TargetTokenSelection& target_selection,
    int32_t proposal_token,
    std::mt19937& rng,
    std::vector<int32_t>& committed,
    std::vector<std::mt19937>& rng_states,
    int& verify_prefix,
    int& n_direct,
    int& corrections,
    bool& rejected) {
  
  const int32_t target_token = SampleTargetToken(target_selection, rng);
  rng_states.push_back(rng);
  committed.push_back(target_token);
  
  if (target_token == proposal_token) {
    verify_prefix++;
    n_direct++;
    for (int32_t fwd : CommitGuidanceToken(proc, target_token)) {
      // Handle forced tokens from acceptance
    }
    return true; // accepted
  } else {
    rejected = true;
    corrections++;
    // Handle rejection + forced tokens
    return false; // rejected
  }
}
```

Then call it from `RunGuidanceRound`:

```cpp
} else if (deterministic_sampling) {
  if (!AcceptGuidanceTokenDeterministic(
      proc, target_selection, proposal.tokens[i], g.rng_,
      committed, deterministic_rng_states, verify_prefix, n_direct,
      corrections, rejected)) {
    break;
  }
}
```

**Benefits:**
- Reduces `RunGuidanceRound` by ~20 lines
- Isolates RNG state capture logic in one place
- Makes the function independently testable
- Improves maintainability of future forced-token or RNG changes
````

**Why this matters:**
Large functions are hard to review and maintain. Extracting well-scoped helpers doesn't change functionality but dramatically improves code clarity and reduces the risk of subtle bugs (like the RNG state asymmetry issue above). The helper is a candidate for unit testing in isolation.

---

## **IMPROVEMENT 4: Avoid Redundant Parameter Storage in NGramDecodingStrategy**

File: `src/n_gram_decoding_strategy.h` and `.cpp`
Lines: (constructor and field definitions)

````markdown
## Eliminate Redundant Parameter Duplication

`NGramDecodingStrategy` stores `batch_size`, `num_beams`, and `num_return_sequences` as member fields. These values are already available through:
- The owning `Search` object (check accessible fields/methods)
- The config passed to the constructor
- The generator/request

**Suggestion:** Remove these fields and read from `search` or config on demand:

```cpp
// Before:
class NGramDecodingStrategy {
  int batch_size_;
  int num_beams_;
  int num_return_sequences_;
  // ... other fields ...
};

// After:
class NGramDecodingStrategy {
  Search* search_;  // or pass search as reference to methods needing it
  // ... only store strategy-specific state ...
};

// In methods that need batch_size:
int batch_size = search_->batch_size; // or search_->GetBatchSize()
```

This reduces state duplication, eases maintenance (no need to keep fields in sync), and makes dependencies explicit at the call site.
````

**Why this matters:**
Duplicated state creates maintenance burden and can become stale if the source is updated but the copy is not. By reading from authoritative sources (Search, config), the code becomes more maintainable and the source of truth is always clear.

---

## **IMPROVEMENT 5: Add Clarifying Documentation to NGramLookup**

File: `src/n_gram_lookup.h`
Lines: (class definition comment)

````markdown
## Document NGramLookup Memory Model and Collision Semantics

Add detailed class-level documentation explaining memory usage, collision handling, and proposal behavior:

```cpp
/// Incremental exact n-gram lookup using token vector keys.
///
/// **Memory model:** Uses exact token sequences (not hashed) as keys in an
/// unordered_map. Each new n-gram suffix appended is indexed. Collisions are
/// NOT compressed—if the same n-gram pattern appears multiple times, only the
/// most recent position is retained. Memory complexity: O(sequence_length *
/// ngram_size) for the history buffer and occurrence map.
///
/// **Update pattern:** Incremental append dominates during normal generation
/// (only newly committed suffixes are indexed). A full re-index occurs only on
/// rewind (e.g., during continuous decoding with Append/SetLogits), so
/// performance typically remains amortized linear.
///
/// **Proposal guarantee:** Returns a contiguous suffix from the most recent
/// matching history. If no match is found, returns an empty proposal (caller
/// falls back to standard decoding). Does not support chained lookups or
/// multiple fallback n-gram orders yet.
///
/// **Invariant:** Only verified and committed tokens are indexed. Unverified
/// proposal tokens are never inserted, ensuring lookup history is accurate
/// for downstream decisions.
class NGramLookup {
 public:
  explicit NGramLookup(int ngram_size)
      : ngram_size_{ngram_size},
        key_length_{static_cast<size_t>(ngram_size - 1)} {
    if (ngram_size < 2)
      throw std::runtime_error("NGramLookup requires ngram_size >= 2.");
  }
  // ... rest of class ...
};
```

This documentation helps:
- **Operators** understand memory overhead and when n-gram decoding is feasible on memory-constrained hardware
- **Maintainers** reason about trade-offs (e.g., "can we use a smaller n-gram to save memory?") and justify the incremental append design
- **Future work** provide context for enhancements (e.g., chained lookups or adaptive n-gram order)
````

**Why this matters:**
Without this documentation, users and maintainers must infer memory usage and collision semantics from code inspection. Clear documentation enables better deployment decisions, helps maintainers make informed trade-off choices, and provides a foundation for future optimizations or feature additions.

---

## **IMPROVEMENT 6: Document RunUnchunked() Purpose and Assumptions**

File: `src/models/decoder_only.h` and `.cpp`
Lines: (RunUnchunked method definition)

````markdown
## Clarify RunUnchunked() Purpose and When It Must Be Used

Add documentation explaining why the default `Run()` method is insufficient for speculative verification:

```cpp
/// Run the model without prefill chunking, returning logits for every input token.
///
/// **Purpose:** Speculative decoding's verification step requires logits for
/// every proposed token to compute acceptance probabilities. If prefill chunking
/// is enabled, `Run()` returns logits only for the final chunk, which is
/// insufficient for multi-token verification.
///
/// **When to use:** Called from speculative decoding strategies when verifying
/// multiple proposed tokens in a batched forward pass. Should NOT be called
/// during standard single-token generation, as it bypasses chunking optimizations.
///
/// **Note:** This can be improved in a future PR to integrate with chunking more
/// intelligently (e.g., adaptive chunking boundaries for speculative proposals).
///
/// **Parameters:**
/// - tokens: Full input token sequence (both committed and proposed)
/// - Returns: Full logits matrix (one row per token)
OrtStatusPtr RunUnchunked(
    const std::vector<int32_t>& tokens,
    std::vector<float>& logits);
```

This clarifies assumptions and helps future maintainers understand the trade-off (simplicity now, optimization later).
````

**Why this matters:**
Without documentation, future developers might wonder why `RunUnchunked()` exists or attempt to use it inappropriately (defeating chunking optimizations). The comment explains the motivation, sets expectations for when it's needed, and acknowledges that it's a v0 limitation. This prevents misuse and provides direction for future optimization work.

---

## **IMPROVEMENT 7: Add Defensive RNG State Validation in DrainOne**

File: `src/speculative_decoding_strategy.cpp`
Lines: (in the `DrainOne()` method)

````markdown
## Add Defensive RNG State Validation During Token Emission

Currently, RNG state capture is validated *after* tokens are buffered (lines 742–745):

```cpp
if (!round_.pending_rng_states.empty() &&
    round_.pending_rng_states.size() != round_.pending.size())
  throw std::runtime_error(
      "N-gram decoding captured an inconsistent number of sampling RNG states.");
```

Add an earlier, defensive check in `DrainOne()` when tokens are actually emitted, to catch corruptions sooner:

```cpp
void SpeculativeDecodingStrategy::DrainOne(Generator& g) {
  // ... existing checks ...
  
  if (round_.pending.empty())
    return;

  const int32_t tok = round_.pending.front();
  round_.pending.pop_front();
  
  // NEW: Defensive check - if RNG state capture was enabled for this round,
  // every emitted token must have a corresponding state. This catches
  // asymmetries (e.g., forced tokens without RNG states) early.
  if (!round_.pending_rng_states.empty()) {
    if (round_.pending_rng_states.empty()) {
      throw std::runtime_error(
          "N-gram decoding: RNG state missing for emitted token. "
          "pending_rng_states was non-empty at round start but is now exhausted. "
          "This may indicate asymmetric RNG capture (e.g., forced tokens without states).");
    }
    g.rng_ = round_.pending_rng_states.front();
    round_.pending_rng_states.pop_front();
  }
  
  // ... rest of emission logic ...
}
```

**Benefits:**
- Catches RNG state underruns at emission time (immediate failure)
- Prevents silent RNG divergence if a code path accidentally skips RNG capture
- Acts as an early-warning system for asymmetries like the guidance forced-token issue
````

**Why this matters:**
RNG-related bugs are notoriously hard to debug because they manifest as non-reproducibility, not crashes. A defensive check at emission catches problems immediately, before corrupted RNG state silently breaks multiple downstream sessions. The check is cheap (one condition per token) and greatly improves debuggability.

---

## **IMPROVEMENT 8: Document Configuration Defaults and Tuning Guidance**

File: `src/config.h` and/or `src/config.cpp`
Lines: (speculative config definitions)

````markdown
## Document How Speculative Configuration Defaults Were Chosen

Add comments explaining the rationale behind default values and tuning guidance:

```cpp
struct SpeculativeConfig {
  // N-gram suffix length for n-gram decoding proposals.
  // Valid range: 2–16 (0 = disabled).
  // Lower values (2–4) are more conservative; higher values (8–16) require
  // more repetition in the sequence to find matches.
  // Default 3 balances lookup overhead vs. proposal likelihood across diverse models.
  // TODO: Profile and adjust based on model distribution analysis.
  int ngram_size = 3;

  // Maximum tokens proposed per speculative round.
  // Default 4 is conservative; balances target verification amortization vs.
  // draft work cost. Best value depends on model size, KV cache efficiency,
  // and execution provider (e.g., CUDA vs. CPU). Consider increasing for
  // models with cheap per-token inference; decrease if target verification
  // becomes a bottleneck.
  int max_draft_tokens = 4;

  // Enable adaptive K: dynamically increase proposal length if acceptance
  // rate is high, to exploit repetitive sequences more aggressively.
  // When enabled, effective K can grow beyond max_draft_tokens (up to
  // min_adaptive_k * max_draft_tokens). Increases verification cost but can
  // yield higher throughput on repetitive workloads.
  bool adaptive_k = false;

  // Minimum probe width when adaptive_k is enabled.
  // Adaptive K grows in increments; this controls the initial step size.
  // Default 2 means probe lengths increase by 2x, then 3x, etc.
  // Higher values (e.g., 4) make probes grow faster but less granularly.
  int min_adaptive_k = 2;

  // Enable cooldown: if acceptance rate drops below a threshold,
  // temporarily reduce K to avoid wasted verifications.
  // Helps maintain efficiency during sequence transitions (e.g., from
  // repetitive prompt to unique generation).
  bool cooldown = false;
};
```

These comments help users make informed tuning decisions and provide context for future optimizations.
````

**Why this matters:**
Without tuning guidance, users don't know whether to adjust `ngram_size`, `max_draft_tokens`, `adaptive_k`, etc., and why. Clear documentation of trade-offs enables better deployments and sets expectations for performance. It also signals where future profiling or adaptive tuning could improve results.

---

File: src/cuda/search_cuda.cpp [hunk 16]
```diff
@@ -40,16 +40,15 @@ void GreedySearch_Cuda::EnsureSamplingData() {
...
   // Allocate a single buffer for all sampling data
   size_t sampling_buffer_size = cuda::SamplingData::CalculateTotalSize(params_->search.batch_size, params_->config.model.vocab_size, GetStream());
   sampling_buffer_ = params_->p_device->Allocate<uint8_t>(sampling_buffer_size);
 
   // Create SamplingData with the externally allocated buffer
-  samplingdata_ = std::make_unique<cuda::SamplingData>(random_seed, params_->search.batch_size, params_->config.model.vocab_size, GetStream(),
+  // SamplingData requires initialized cuRAND states even before a stochastic call.
+  // Stochastic sampling replaces them from the caller-owned RNG in SampleTopKTopP;
+  // 0 is therefore only a deterministic initialization value.
+  samplingdata_ = std::make_unique<cuda::SamplingData>(0, params_->search.batch_size, params_->config.model.vocab_size, GetStream(),
```
kunal-vaishnavi:
```suggestion
  sampling_data_ = std::make_unique<cuda::SamplingData>(0, params_->search.batch_size, params_->config.model.vocab_size, GetStream(),
```

src/models/decoder_only.h
```diff
@@ -23,6 +23,9 @@ struct DecoderOnly_State : State {
   void SetExtraInputs(const std::vector<ExtraInput>& extra_inputs) override;
 
   DeviceSpan<float> Run(int total_length, DeviceSpan<int32_t>& next_tokens, DeviceSpan<int32_t> next_indices) override;
+  // Speculative verification needs a logits row for every input token, so it cannot split the run.
+  DeviceSpan<float> RunUnchunked(int total_length, DeviceSpan<int32_t>& next_tokens,
```
kunal-vaishnavi: Can we remove this from the header? The header should only need `Run` as the logic to use chunking or no chunking is within `Run` and scoped to the `decoder_only.cpp` file.

src/base_speculative_strategy.cpp
```diff
@@ -95,13 +97,16 @@ SpeculativeDecodingStrategy::Proposal BaseSpeculativeStrategy::Propose(
...
   }
-  Proposal proposal;
+  Proposal proposal{greedy ? ProposalMode::kGreedyMatch : ProposalMode::kDraftSampling};
   proposal.tokens.resize(K);
   if (!greedy)
-    // greedy-match leaves probs empty
     proposal.probs.resize(K);
 
   SampledCategorical sampled;
+  auto sample_with_round_rng = [&](std::span<const float> logits, int index) {
```
kunal-vaishnavi: Can we put this in a named helper function instead of an anonymous lambda function?

src/speculative_decoding_strategy.cpp
```diff
@@ -638,9 +858,12 @@ void SpeculativeDecodingStrategy::DrainOne(Generator& g) {
   // Last token of the round just went out: re-anchor now. (Deferring it to here means an EOS
   // partway through the round skips the re-anchor and its wasted target pass.)
   if (round_.pending.empty() && round_.IsActive()) {
-    FinishRound();
+    round_.phase = RoundState::Phase::kFinalizing;
```
kunal-vaishnavi: Can we find a way to reduce the lengths of the name aliasing?

kunal-vaishnavi (dismissed):

src/speculative_decoding_strategy.cpp
samsat701: Completed in https://github.com/microsoft/onnxruntime-genai/pull/2314/changes/11df5e5c69d9cf6d070adef90bf48dfb28fafa86.

src/base_speculative_strategy.cpp
samsat701: Fixed in https://github.com/microsoft/onnxruntime-genai/pull/2314/changes/11df5e5c69d9cf6d070adef90bf48dfb28fafa86.

src/models/decoder_only.h
samsat701: Fixed in https://github.com/microsoft/onnxruntime-genai/pull/2314/changes/11df5e5c69d9cf6d070adef90bf48dfb28fafa86.

src/cuda/search_cuda.cpp
samsat701: Fixed in https://github.com/microsoft/onnxruntime-genai/pull/2314/changes/11df5e5c69d9cf6d070adef90bf48dfb28fafa86.

kunal-vaishnavi (dismissed):

baijumeswani (dismissed):

baijumeswani (approved):

PR #2318, opened 2026-07-22
[WebGPU] implements webgpu support in benchmarking script

prathikr:
Planning to leverage this script for WebGPU Performance Dashboard

### DIFF
--- a/benchmark/python/benchmark_e2e.py
+++ b/benchmark/python/benchmark_e2e.py
@@ -422,11 +422,15 @@ def run_benchmark(args, batch_size, prompt_length, generation_length, max_length
     # Calculate prompt processing metrics
     avg_prompt_latency_s = sum(prompt_times) / len(prompt_times)
     avg_prompt_latency_ms = avg_prompt_latency_s * 1000
-    std_prompt_latency_ms = float(np.std(prompt_times)) * 1000
     avg_per_token_prompt_latency_ms = avg_prompt_latency_ms / prompt_length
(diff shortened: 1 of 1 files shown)


baijumeswani (dismissed):

benchmark/python/benchmark_e2e.py
```diff
@@ -496,6 +505,19 @@ def main(args):
         og.register_execution_provider_library(provider_registration_name, args.ep_library_path)
         if args.verbose:
             print(f"Successfully registered {provider_registration_name} from {args.ep_library_path}")
+    elif args.execution_provider == "webgpu":
```
baijumeswani, line 520: We should also add support for the case where the webgpu ep path is passed in in the if block above.

baijumeswani (approved):

kunal-vaishnavi (approved):

benchmark/python/benchmark_e2e.py
```diff
@@ -478,6 +491,14 @@ def main(args):
                 "Please specify an execution provider using -e (e.g., -e cuda or -e NvTensorRtRtx)"
             )
 
+        if args.execution_provider == "webgpu":
```
baijumeswani, line 494: In theory it is possible to just pass in the path to the dll. We don't need install the plug in ep package in python.

prathikr: Will do this in a separate pipeline if needed.

PR #2321, opened 2026-07-24
Support Quantized KV Cache

tianleiwu:
This is to support quantized kv cache. End-to-end tested with gpt-oss-20b int8/fp8 kv cache.


src/python/py/models/kv_cache_calibration.py
```diff
@@ -0,0 +1,684 @@
...
+chain-of-thought workloads.
+"""
+
+from __future__ import annotations
+
+import json
+import logging
+import os
+from pathlib import Path
+from typing import Any
```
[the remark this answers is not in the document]
tianleiwu: Fixed in 2c6d7f2484: removed the unused `typing.Any` import.

src/python/py/models/kv_cache_calibration.py
```diff
@@ -0,0 +1,684 @@
...
+# Diverse, long-form corpus spanning MMLU-ish domains (STEM, humanities, social science,
+# reasoning). Passages are concatenated and sliced to TARGET_SEQ tokens so calibration covers
+# the same post-RoPE position range that the eval context length exercises.
+CORPUS = [
+    "The theory of general relativity, formulated by Albert Einstein in 1915, describes gravity "
+    "not as a force but as the curvature of spacetime caused by mass and energy. Massive objects "
+    "such as stars and planets warp the geometry of spacetime, and this curvature dictates how "
+    "other objects move. The theory predicted phenomena such as the bending of light by gravity, "
+    "the precession of Mercury's orbit, gravitational time dilation, and the existence of black "
+    "holes and gravitational waves, all of which have since been confirmed experimentally.",
```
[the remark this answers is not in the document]
tianleiwu: Fixed in 2c6d7f2484: the calibration corpus is now stored as triple-quoted passages that are whitespace-normalized at import time, so no passage relies on implicit string-literal concatenation inside a list. This clears all 16 alerts of this class in the file; the resulting corpus text is byte-for-byte unchanged.

src/python/py/models/builder.py
```diff
@@ -86,6 +87,140 @@
...
+    """
+    kv_pairs = {}
+
+    if kv_items:
+        for kv_str in kv_items:
+            kv = kv_str.split("=")
+            kv_pairs[kv[0].strip()] = kv[1].strip()
+
+    print(f"Extra options: {kv_pairs}")
+    check_extra_options(kv_pairs, precision, execution_provider)
```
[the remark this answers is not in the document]
tianleiwu: Fixed in da2cbade. The stray 3-argument `check_extra_options`/`parse_extra_options` were leftover duplicates from the merge with `main` (which refactored these helpers to the 7-argument signature). The duplicate definitions have been removed and the `kv_cache_quant_type` validation was ported into the surviving 7-arg `check_extra_options`, so this call site no longer exists.

src/python/py/models/builder.py
```diff
@@ -86,6 +87,140 @@
...
+            )
+        if quant_type != "none" and execution_provider not in {"cpu", "cuda"}:
+            raise ValueError(
+                "Quantized KV cache is only supported for the CPU and CUDA execution providers. "
+                f"Got execution_provider='{execution_provider}'."
+            )
+        kv_pairs["kv_cache_quant_type"] = quant_type
+
+
+def parse_extra_options(kv_items, precision, execution_provider):
```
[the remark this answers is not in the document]
tianleiwu: Fixed in da2cbade. The duplicate `parse_extra_options` (and `check_extra_options`) definitions left over from the merge with `main` have been removed, so the redefinition no longer occurs.

baijumeswani (dismissed):

src/python/py/models/builders/base.py
```diff
@@ -282,6 +282,15 @@ def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
         }
         self.make_attention_init(config)
 
+        kv_cache_quant_type = extra_options.get("kv_cache_quant_type", "none")
+        self.kv_cache_quant_type = str(kv_cache_quant_type).strip().lower()
+        if self.kv_cache_quant_type not in KV_CACHE_QUANT_TYPES:
```
kunal-vaishnavi: Can we do these validation checks inside `builder.py` for this new extra option?

src/python/py/models/builders/base.py
```diff
@@ -555,6 +564,89 @@ def make_attention_init(self, config):
...
+            cache_dtype = ir.DataType.UINT8
+        else:
+            cache_dtype = ir.DataType.INT8
+        self.input_types["past_key_values.key"] = cache_dtype
+        self.input_types["past_key_values.value"] = cache_dtype
+        self.output_types["present.key"] = cache_dtype
+        self.output_types["present.value"] = cache_dtype
+
+        if is_int4:
+            packed_head_size = (self.head_size + 1) // 2
```
kunal-vaishnavi, line 592: [Edge](https://github.com/microsoft/onnxruntime-genai/pull/2084#discussion_r3601023574) wants the last dim to be a string for a quantized KV cache. Will your changes to make it a constant work for them? I would prefer it to be a constant but I understand their reasoning for keeping the dim shape as a string.

src/python/py/models/builders/base.py
```diff
@@ -2744,15 +2836,30 @@ def make_group_query_attention(self, name, **kwargs):
             "",  # attention_bias
             kwargs.get("sinks", ""),
         ]
+        layer_id = kwargs.get("layer_id")
+        if self.kv_cache_quant_type != "none":
+            if layer_id is None:
+                raise ValueError("layer_id is required for quantized KV cache.")
+            inputs.extend(
+                [
```
kunal-vaishnavi, line 2848: Can we change these names to follow the same naming conventions as other initializers?

https://github.com/microsoft/onnxruntime-genai/blob/da2cbadeb14512bd0bf04c243d2335e31022e6dc/src/python/py/models/builders/base.py#L3175-L3182

src/python/py/models/builders/base.py
tianleiwu: Done in 6ae2218. The validation and normalization already live in `check_extra_options` in `builder.py`, and `create_model` requires the `hf_details` that `check_extra_options` populates, so it always runs before the model is constructed. I removed the redundant re-validation (and the now-unused `KV_CACHE_QUANT_TYPES` import) from `base.py` so `builder.py` is the single source of validation.

src/python/py/models/builders/base.py
tianleiwu: Done in 6ae2218. Renamed the per-layer scales to `model.layers.{layer_id}.attn.k_scale` / `model.layers.{layer_id}.attn.v_scale` via a new `get_kv_cache_scale_names()` helper (mirroring `get_qk_norm_weight_names`), used by both the initializer creation and the GQA input wiring so the names cannot drift.

src/python/py/models/builders/base.py
tianleiwu: Good question. In this PR the default last dim stays symbolic (`kv_cache_dim`, see the past_key_values shapes), so a single export still serves both the non-quantized (`head_size`) and the int8/fp8 quantized caches — those paths never touch the dim, so Edge (PR #2084) keeps the string dim they want.

The only place I override it to a constant is the int4 path, where the head dim is bit-packed to `(head_size + 1) // 2` bytes. That factor-of-2 packing is a build-time integer that a symbolic `kv_cache_dim` cannot express (the symbol has no value), so leaving it symbolic would make the exported shape inconsistent with the actual packed tensor. Since head_size is already fixed at export time, emitting the packed dim as a constant is exact and matches your preference.

If Edge specifically needs int4 to remain symbolic too, I can drop the override and leave `kv_cache_dim` for int4 as well — happy to go either way, but the constant is intentional for shape correctness here.

kunal-vaishnavi (approved):

PR #2330, opened 2026-07-28
Add paged attention support to the model builder

baijumeswani:
Adds support for building CUDA models with paged attention for the continuous-batching engine.

- Supports FP16 and BF16 models, including Qwen3.
- Generates the required paged-attention inputs and engine configuration.
- Handles Q/K normalization, sliding-window attention, and paged KV-cache types correctly.
- Validates incompatible or invalid builder options.
- Fixes completed-request handling during dynamic batching.

Example build commands:

```
CACHE=/path/to/hf_cache

Phi / Qwen2.5 — int4 weights, fp16 activations:

# Phi-3.5-mini-instruct
python src/python/py/models/builder.py -m microsoft/Phi-3.5-mini-instruct \
  -o phi-3.5-mini-instruct-paged -p int4 -e cuda -c $CACHE \
  --extra_options use_paged_attention=true

# Phi-3-medium-128k-instruct
python src/python/py/models/builder.py -m microsoft/Phi-3-medium-128k-instruct \
  -o phi-3-medium-128k-instruct-paged -p int4 -e cuda -c $CACHE \
  --extra_options use_paged_attention=true

# Qwen2.5-14B-Instruct
python src/python/py/models/builder.py -m Qwen/Qwen2.5-14B-Instruct \
  -o qwen2.5-14b-instruct-paged -p int4 -e cuda -c $CACHE \
  --extra_options use_paged_attention=true

# Qwen2.5-32B-Instruct
python src/python/py/models/builder.py -m Qwen/Qwen2.5-32B-Instruct \
  -o qwen2.5-32b-instruct-paged -p int4 -e cuda -c $CACHE \
  --extra_options use_paged_attention=true

Qwen3 — bf16 (required; fp16 overflows to NaN, int4 produces incoherent output):

# Qwen3-1.7B
python src/python/py/models/builder.py -m Qwen/Qwen3-1.7B \
  -o qwen3-1.7b-paged -p bf16 -e cuda -c $CACHE \
  --extra_options use_paged_attention=true

# Qwen3-32B
python src/python/py/models/builder.py -m Qwen/Qwen3-32B \
  -o qwen3-32b-paged -p bf16 -e cuda -c $CACHE \
  --extra_options use_paged_attention=true
```


src/engine/engine.cpp
```diff
@@ -21,17 +21,15 @@ void Engine::RemoveRequest(std::shared_ptr<Request> request) {
...
+  while (HasPendingRequests()) {
+    if (!ready_requests_.empty()) {
+      auto request = ready_requests_.front();
+      ready_requests_.pop();
+      return request;
+    }
 
-  if (auto scheduled_requests = scheduler_->Schedule()) {
+    auto scheduled_requests = scheduler_->Schedule();
     model_executor_->Decode(scheduled_requests);
```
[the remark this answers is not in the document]
baijumeswani: CI does not have meaningful paged attention tests. Will leave this one out for now.

src/python/py/models/builders/base.py
```diff
@@ -405,11 +408,29 @@ def make_inputs_init(self):
         else:
             del self.input_names["inputs_embeds"]
 
+        if self.use_paged_attention:
+            self.input_shapes["input_ids"] = ["num_tokens"]
+            self.input_shapes["past_key_values.key"] = ["num_blocks", "block_size", self.num_kv_heads, self.head_size]
+            self.input_shapes["past_key_values.value"] = ["num_blocks", "block_size", self.num_kv_heads, self.head_size]
+            if "attention_mask" in self.input_names:
+                del self.input_names["attention_mask"]
+            for name in ["block_table", "cumulative_sequence_length", "past_seqlens"]:
```
kunal-vaishnavi: Can we add `block_table`, `cumulative_sequence_length`, and `past_seqlens` to the dicts when those dicts are created? The inputs are added to the graph based on whether they exist in `self.input_names`, so it is alright to initialize them and then remove their entries (see `inputs_embeds` earlier as an example).

src/python/py/models/builders/base.py
```diff
@@ -405,11 +408,29 @@ def make_inputs_init(self):
...
+                self.input_names[name] = name
+                self.input_types[name] = ir.DataType.INT32
+            self.input_shapes["block_table"] = ["batch_size", "max_num_blocks"]
+            self.input_shapes["cumulative_sequence_length"] = ["batch_size + 1"]
+            self.input_shapes["past_seqlens"] = ["batch_size"]
```
kunal-vaishnavi: Can we rename this to `past_sequence_lengths` instead of using abbreviations in the I/O names?

src/python/py/models/builders/base.py
```diff
@@ -405,11 +408,29 @@ def make_inputs_init(self):
...
+
     def make_outputs_init(self):
         # Always use float32 logits to improve accuracy in the case of bf16 models.
         if self.io_dtype == ir.DataType.BFLOAT16:
             self.output_types["logits"] = ir.DataType.FLOAT
 
+        if self.use_paged_attention:
```
kunal-vaishnavi, line 443: Same comment on outputs

src/python/py/models/builders/base.py
```diff
@@ -424,6 +445,12 @@ def make_outputs_init(self):
         if self.exclude_lm_head:
             del self.output_names["logits"]
 
+    def hidden_state_shape(self, last_dim, seq_dim="sequence_length"):
+        """Return a standard 3D shape or a packed 2D paged-attention shape."""
+        if getattr(self, "use_paged_attention", False):
```
kunal-vaishnavi: `self.use_paged_attention` should already be defined by this point.

```suggestion
        if self.use_paged_attention:
```

src/python/py/models/builders/base.py
```diff
@@ -526,6 +553,41 @@ def make_attention_init(self, config):
         self.q_size = self.num_attn_heads * self.head_size
         self.kv_size = self.num_kv_heads * self.head_size
 
+        if self.use_paged_attention:
```
kunal-vaishnavi, line 572: Instead of having an `if` case that returns early, can we integrate this with the existing `if-elif` logic below it?

src/python/py/models/builders/base.py
```diff
@@ -526,6 +553,41 @@ def make_attention_init(self, config):
...
+                and not self.attention_attrs["k_norm"]
+                and not self.extra_options.get("disable_qkv_fusion", False)
+            )
+
+            # PagedAttention fuses the rotary embeddings, so `position_ids` is not needed as an input.
+            self.attention_attrs["use_rope_in_attn"] = True
+            if "position_ids" in self.input_names:
+                del self.input_names["position_ids"]
+
+            self.past_present_share_buffer = True
```
kunal-vaishnavi: After the conditional integration, we can update the buffer sharing check below to see if GQA or PA is being used.

src/python/py/models/builders/base.py
```diff
@@ -801,6 +867,16 @@ def make_genai_config(self, model_name_or_path, extra_kwargs, out_dir):
             ep_options = {ep_name: self.ep_attrs[self.ep]}
             genai_config["model"]["decoder"]["session_options"]["provider_options"].append(ep_options)
 
+        if self.use_paged_attention:
+            block_size = int(self.extra_options.get("paged_block_size", 256))
```
kunal-vaishnavi: We should already have the paged block size assigned in the `attention_attrs` dict so that we don't have to re-parse the extra options. I see this line repeated 4 times in this file.

src/python/py/models/builders/base.py
```diff
@@ -424,6 +445,12 @@ def make_outputs_init(self):
         if self.exclude_lm_head:
             del self.output_names["logits"]
 
+    def hidden_state_shape(self, last_dim, seq_dim="sequence_length"):
```
kunal-vaishnavi: Can we reorder the signature here?

```suggestion
    def hidden_state_shape(self, seq_dim="sequence_length", last_dim=self.hidden_size):
```

The param ordering would now follow the standard model's pattern of `(batch_size, sequence_length, hidden_size)`. We can also simplify some of the `hidden_state_shape` calls to `self.get_hidden_state_shape()`.

src/python/py/models/builders/base.py
```diff
@@ -2276,6 +2360,33 @@ def make_rotary_embedding_multi_cache(self, **kwargs):
...
+        #
+        # The result feeds the Greater node that selects the short vs. long rotary cache.
+        basename = "/model/context_length_subgraph"
+        shape_name = f"{basename}/Shape"
+        shape_output = f"{shape_name}/output_0"
+        self.make_shape(shape_name, "block_table", shape=[2])
+
+        gather_name = f"{basename}/Gather"
+        gather_output = f"{gather_name}/output_0"
+        self.make_node(
```
kunal-vaishnavi: Can we use `make_gather` to simplify?

src/python/py/models/builders/base.py
```diff
@@ -2276,6 +2360,33 @@ def make_rotary_embedding_multi_cache(self, **kwargs):
...
+            "Gather", inputs=[shape_output, "/model/constants/INT64/1"], outputs=[gather_output], name=gather_name
+        )
+        self.make_value(gather_output, ir.DataType.INT64, shape=None)
+
+        # max_num_blocks (block_table dim 1) x block_size approximates the max representable
+        # context length. block_size is a build-time constant (matches the engine's block_size).
+        block_size = int(self.extra_options.get("paged_block_size", 256))
+        multiply_name = f"{basename}/Multiply"
+        multiply_inputs = [gather_output, f"/model/constants/INT64/{block_size}"]
+        self.make_node("Mul", inputs=multiply_inputs, outputs=[f"{multiply_name}/output_0"], name=multiply_name)
```
kunal-vaishnavi: Can we use `make_mul` to simplify?

src/python/py/models/builders/base.py
```diff
@@ -2276,6 +2360,33 @@ def make_rotary_embedding_multi_cache(self, **kwargs):
...
+        gather_output = f"{gather_name}/output_0"
+        self.make_node(
+            "Gather", inputs=[shape_output, "/model/constants/INT64/1"], outputs=[gather_output], name=gather_name
+        )
+        self.make_value(gather_output, ir.DataType.INT64, shape=None)
+
+        # max_num_blocks (block_table dim 1) x block_size approximates the max representable
+        # context length. block_size is a build-time constant (matches the engine's block_size).
+        block_size = int(self.extra_options.get("paged_block_size", 256))
+        multiply_name = f"{basename}/Multiply"
```
kunal-vaishnavi: Node names should [end with](https://github.com/microsoft/onnxruntime-genai/blob/main/src/python/py/models/DESIGN.md#namespaces) the node's op type.

```suggestion
        multiply_name = f"{basename}/Mul"
```

src/python/py/models/builders/base.py
```diff
@@ -2668,6 +2781,14 @@ def make_attention_op(self, name, **kwargs):
                 total_seq_len=f"{self.mask_attrs['total_seq_len']}/output_0",
                 **kwargs,
             )
+        elif op_type == "PagedAttention":
+            self.make_paged_attention(
+                name,
+                cumulative_sequence_length="cumulative_sequence_length",
```
kunal-vaishnavi: Can we set default names inside `self.input_names` if using raw model inputs or `self.mask_attrs` if using derived inputs for the op and reference those instead of hardcoding strings here?

src/python/py/models/builder.py
```diff
@@ -176,6 +177,34 @@ def check_extra_options(
...
+            key
+            for key in ("exclude_embeds", "exclude_lm_head", "prune_lm_head")
+            if extra_options.get(key, False)
+        ]
+        if incompatible_options:
+            raise ValueError(
+                "use_paged_attention cannot be combined with " + ", ".join(incompatible_options) + "."
+            )
+
+        for key, default in (("paged_block_size", 256), ("max_batch_size", 100)):
```
kunal-vaishnavi: You already have default values set later on in `base.py`. Here, it should be sufficient to just check for a positive integer.

src/python/py/models/builder.py
```diff
@@ -657,6 +686,20 @@ def get_args():
...
                     In addition to `logits`, you will have `hidden_states` as an output to your ONNX model.
+                use_paged_attention = Build the model with PagedAttention for the continuous-batching engine. Default is false.
+                    Replaces GroupQueryAttention with the PagedAttention contrib op, packs all sequences into a single
+                    flattened token axis (`input_ids` becomes 1D), stores the KV-cache in paged
+                    [num_blocks, block_size, num_kv_heads, head_size] buffers, and removes the `attention_mask` and
+                    `position_ids` inputs in favor of the `block_table`, `cumulative_sequence_length`, and `past_seqlens`
+                    metadata inputs. An `engine` section (block_size, gpu_utilization_factor, max_batch_size) is added to
+                    genai_config.json. Currently only supported for the CUDA execution provider with fp16 or bf16 precision.
+                    Cannot be combined with exclude_embeds, exclude_lm_head, or prune_lm_head.
+                paged_block_size = Paged KV-cache block size used when use_paged_attention is set. Default is 256.
```
kunal-vaishnavi: Can you list valid block sizes similar to how `block_size` and `qmoe_block_size` do?

src/python/py/models/builder.py
```diff
@@ -657,6 +686,20 @@ def get_args():
...
+                    Must be a positive integer. Also written to the `engine.dynamic_batching` section of genai_config.json.
+                gpu_utilization_factor = Fraction of available GPU memory used for the paged KV-cache. Default is 0.6.
```
kunal-vaishnavi, line 723: Is this something that the user should control here? If we set a default of 0.6 in the GenAI config, a user can either modify the JSON file or programmatically change it during inference.

src/python/py/models/builder.py
```diff
@@ -657,6 +686,20 @@ def get_args():
...
+                    Must be greater than 0 and at most 1.
+                max_batch_size = Maximum number of requests in a dynamic batch. Default is 100.
```
kunal-vaishnavi, line 725: Same comment here as [this one](https://github.com/microsoft/onnxruntime-genai/pull/2330/changes#r3668065685)

kunal-vaishnavi, line 725: We have existing logic inside GenAI for restricting the batch size during normal generation because of an IcM issue. Will this conflict?

https://github.com/microsoft/onnxruntime-genai/blob/07e16463efb6b04bded136099202d3ac39370b4a/src/generators.cpp#L384-L396

src/config.cpp
```diff
@@ -1335,13 +1335,25 @@ struct DynamicBatching_Element : JSON::Element {
...
+      v_->block_size = static_cast<size_t>(parsed_value);
     } else if (name == "num_blocks") {
-      v_->num_blocks = static_cast<size_t>(JSON::Get<double>(value));
+      const auto parsed_value = SafeDoubleToInt(JSON::Get<double>(value), name);
+      if (parsed_value <= 0)
+        throw std::out_of_range("num_blocks must be > 0");
+      v_->num_blocks = static_cast<size_t>(parsed_value);
     } else if (name == "gpu_utilization_factor") {
-      v_->gpu_utilization_factor = static_cast<float>(JSON::Get<double>(value));
+      const auto parsed_value = JSON::Get<double>(value);
```
kunal-vaishnavi, line 1348:
```suggestion
      const auto parsed_value = SafeDoubleToInt(JSON::Get<double>(value), name);
```

baijumeswani: `gpu_utilization_factor` should be a double. It is not an int.

src/python/py/models/builders/base.py
baijumeswani: No new outputs were added. Could you elaborate on what you mean by your comment?

src/python/py/models/builders/base.py
baijumeswani: I think the problem is that `=self.hidden_size` cannot be used as a default argument.

To do this, we would need to change the signature to

```py
def hidden_state_shape(self, seq_dim="sequence_length", last_dim=None):
```

And then default to self.hidden_size inside the body. Also, all call-sites would need to invoke the function by named args instead of positional args.

If that seems more reasonable, i can go ahead and do it.

src/python/py/models/builders/base.py
baijumeswani: Added it to the attention_attrs.

src/python/py/models/builder.py
baijumeswani: Yes, user/application should have control over how much of gpu memory they want to dedicate to ort-genai's continuous batching engine.

src/python/py/models/builder.py
baijumeswani: I think it makes sense for this to be in the genai_config.json. max_batch_size bounds how many requests can be in-flight in the paged cache manager at once. The more requests you allow, the fewer tokens you're allowing for generation. So, it makes sense to have this be configurable.

baijumeswani: The Engine does not use the generator at all. It bypasses the generator and so this should not interfere with the engine's mechanics.

Having said that, i am wondering why we have that check in place. It seems restrictive (especially since the batch size is user supplied).

src/python/py/models/builders/base.py
kunal-vaishnavi: Let's do that then. I think having named args is clearer since you're specifying which dim's shape will change.

src/python/py/models/builders/base.py
```diff
@@ -279,6 +291,8 @@ def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
             "mask_filter_value": -10000.0,                   # Masking value to use in attention mask
             "unidirectional": False,                         # Whether every token can only attend to previous tokens
             "use_matmul_in_attn": False,                     # Use MatMuls with attention (instead of separate MatMul ops)
+            # Attributes for PagedAttention op:
+            "block_size": None,                              # KV cache block size (set when use_paged_attention is enabled)
```
kunal-vaishnavi: Can we rename this to `paged_block_size` to be consistent with the extra options flag and not confuse it with other block size variables?

src/python/py/models/builders/base.py
```diff
@@ -147,6 +153,9 @@ def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
             "inputs_embeds": self.io_dtype,                                                                      # For standard models where you want to remove the embedding layer from the model (note that `inputs_embeds` is written this way to match Hugging Face format)
             "past_key_values.key": self.io_dtype,                                                                # For standard models (note that `past_key_values.key` is written this way to match Hugging Face format)
             "past_key_values.value": self.io_dtype,                                                              # For standard models (note that `past_key_values.value` is written this way to match Hugging Face format)
+            "block_table": ir.DataType.INT32,                                                                    # For paged attention models
+            "cumulative_sequence_lengths": ir.DataType.INT32,                                                    # For paged attention models
```
kunal-vaishnavi, line 157: Why int32 instead of int64 to match the input ids and attention mask dtypes?

baijumeswani: I think int32 makes more sense. At the time of implementation, we used int32 in the pagedattention operator. int64 is overkill imo. It is also what other frameworks use if I am not mistaken.

kunal-vaishnavi (dismissed):

tianleiwu (approved):

kunal-vaishnavi (approved):

PR #2333, opened 2026-07-29
PagedAttention: quantized KV cache, attention_metadata, and CUDA graph capture

tianleiwu:
## Summary

Extends the PagedAttention path end to end: the model builder can now export a quantized (int8/fp8) paged KV cache, both the builder and the engine wire up the new `attention_metadata` input that removes a per-node host sync, the continuous-batching engine can capture and replay decode steps as CUDA graphs, and the paged cache manager no longer double-counts prefill tokens when reserving blocks.

Measured on 8×H200 (only one GPU was used in benchmark) with gpt-oss-20b, batch-1 decode on the int8-KV paged model goes from **242.7 → 305.7 tok/s (+26%)** from the two changes in this PR (`attention_metadata` + CUDA graphs), and the paged KV pool now serves twice the context/concurrency it did for the same `num_blocks`.

Requires the matching ONNX Runtime changes (`PagedAttention` `attention_metadata` input at index 16, XQA paged decode kernels) in https://github.com/microsoft/onnxruntime/pull/29912.

### 1. Quantized KV cache for PagedAttention (model builder)

| Change | File |
|---|---|
| Accept `PagedAttention` in `make_quantized_kv_cache_init()` (was GQA-only) | `builders/base.py` |
| Reject `int4_*` on the paged path — `PagedAttention`'s `T_CACHE` is `{float16, bfloat16, int8, float8e4m3fn}`, there is no sub-byte paged cache backend | `builders/base.py` |
| Emit per-channel scales in the canonical `(num_kv_heads, 1, head_size)` shape the paged op validates (GQA only checks the element count) | `builders/base.py` |
| Wire `k_scale`/`v_scale` optional inputs and set `k_quant_type`/`v_quant_type`. `PagedAttention` has no `kv_cache_bit_width` attribute — it derives the element type from the cache tensor | `builders/base.py` |
| Document the paged + quantized-KV combination and its constraints | `models/README.md`, `builder.py` |

Shared logic between `make_group_query_attention` and `make_paged_attention` was factored into `get_qk_norm_weight_inputs()`, `get_kv_cache_scale_inputs()`, `get_attention_op_attributes()` and `extend_with_optional_inputs()`, so the two builders now differ only in their fixed input ordering and GQA's `kv_cache_bit_width`.

### 2. `attention_metadata` input

`PagedAttention::ComputeInternal` previously copied the sequence-length tensors device→host and called `cudaStreamSynchronize` to obtain `max_query_len`/`max_kv_len` — **once per node, per step**, i.e. 24 pipeline stalls per token on a 24-layer model.

The new optional CPU input (`int32[2]`, index 16) carries *upper bounds* for those values, letting the kernel skip the readback. It only ever predicts the backend selection; it never overrides it.

- `builders/base.py` — emits it as a graph input (static shape `[2]`, int32), wires it to input 16, lists it in `genai_config.json`.
- `src/config.{h,cpp}` — `Defaults::AttentionMetadataName` and `Decoder::Inputs::attention_metadata`.
- `varlen_decoder_io.cpp` — `PrepareAttentionMetadata()` fills it each step from sequence lengths the engine already holds on the host. It no-ops when `!session_info_.HasInput(...)`, so **older models still load**.

| batch | without metadata | with metadata | gain |
|---|---|---|---|
| 1 | 242.65 tok/s | 263.75 tok/s | **+8.7%** |
| 8 | 796.09 tok/s | 811.50 tok/s | +1.9% |
| 32 | 2635.67 tok/s | 2681.96 tok/s | +1.8% |

TTFT unchanged. The gain is inversely proportional to batch size, as expected for a fixed per-step host cost.

### 3. CUDA graph capture in the continuous-batching engine

An nsys trace of batch-1 decode showed ~2.4 ms of GPU work against a 3.8 ms wall clock — roughly **37% GPU idle**, spent launching ~170 kernels per step. Change 2 is a prerequisite here: ORT captures with `cudaStreamCaptureModeGlobal`, so no sync API may run during capture.

- **`VarlenGraphBuffers`** (`varlen_decoder_io.{h,cpp}`) — persistent `input_ids`, `cumulative_sequence_lengths`, `past_sequence_lengths` and `logits` tensors sized for `max_batch_size`, owned by `SimpleDecoder` so their addresses outlive the per-step IO. A step views a smaller prefix.
- **Static device block table** (`paged_key_value_cache.{h,cpp}`) — the block table used to be a pageable CPU `OrtValue`, which cannot be H2D-copied during capture. Under capture it becomes a persistent device tensor of `[max_batch_size, max_block_table_columns]`.
- **Shape stability via bucketing** — column counts round *up to a power of two* (min 8) and clamp to the reserved capacity, so a long decode touches only a handful of distinct shapes instead of re-capturing every time a block is appended.
- **Capture policy** (`simple_decoder.cpp`) — only pure-decode steps that fit the buffers are captured; prefill and mixed steps run eagerly with `gpu_graph_id = -1`. `GraphId = batch_size * 64 + column_bucket_exponent + 1`.

Because `ComputeInternal` never re-runs on replay, `attention_metadata` must carry replay-wide bounds; the engine writes `[1, block_table_columns * block_size]`.

| config | graphs OFF | graphs ON | gain |
|---|---|---|---|
| b=1, p=128, n=256 | 263.68 tok/s | **305.68 tok/s** | **+15.9%** |
| b=8, p=128, n=256 | 810.15 tok/s | 870.54 tok/s | +7.5% |
| b=32, p=128, n=256 | 2697.87 tok/s | 2879.20 tok/s | +6.7% |
| b=2, p=4096, n=256 | 390.10 tok/s | 435.02 tok/s | +11.5% |

TTFT unchanged — prefill is never captured. fp8-KV reaches 318.96 tok/s at b=1.

### 4. Fix: the paged cache manager double-counted every prefill

Pinning `num_blocks` to exactly `ceil(max_length / block_size)` made every paged request fail on its first step with `Cannot append tokens to request that is not ready.`

`BlockPool::AllocateBlocks()` constructs blocks with their slots *already marked used*, and `PagedKeyValueCache::Add()` called it with the full prefill. The first `Step()` then ran `AppendTokens()` over those same still-unprocessed tokens and allocated a **second complete set of blocks**. A 4,079-token prefill at `block_size = 256` consumed **32 blocks instead of 16**.

The fix makes `AppendTokens()` the single place slots become used:

- `BlockPool::ReserveBlocks()` (new) allocates blocks with `size_ = 0`; `AllocateBlocks()` keeps its fill-on-allocate behaviour and both share one implementation behind a `mark_slots_used` flag.
- `Add()` reserves capacity without consuming it — still enough to stop a second request admitted in the same batch from stealing the blocks.
- `AppendTokens()` walks *all* of the request's blocks filling empty slots before asking the pool for more, instead of only `blocks.back()`.
- `CanAppendTokens()` counts every empty slot the request owns (new `EmptySlots()` helper) and takes the block size from `BlockPool::BlockSize()` rather than `blocks.back()->Capacity()`, which was also UB for an empty block list.
- `CanAdd()` uses `>=` rather than `>`, so a pool sized exactly for one request admits it.

Block ordering, and therefore the physical KV layout, is unchanged. A given KV pool now supports twice the context/concurrency it did, at no throughput cost:

| batch | before | after |
|---|---|---|
| 1 | 375.60 tok/s | 374.97 |
| 8 | 1679.86 | 1690.98 |
| 32 | 4097.30 | 4127.86 |

## Experiment Results

Measured on 8×H200 141 GB (sm_90), CUDA 13.0, ORT 1.29.0.

### Cumulative decode throughput — gpt-oss-20b, int8 per-channel KV, prompt 128 / new 256

| batch | baseline | + `attention_metadata` | + CUDA graphs | total |
|---|---|---|---|---|
| 1 | 242.7 | 263.8 | **305.7** | **+26.0%** |
| 8 | 796.1 | 811.5 | **870.5** | +9.3% |
| 32 | 2635.7 | 2682.0 | **2879.2** | +9.2% |

Attention is no longer the bottleneck at these sizes — the 24 mxfp4 `QMoE` GEMMs (~1.5 ms/step) and 49 `MatMulNBits` nodes dominate, against ~0.2 ms for all 24 attention nodes.

### GQA vs PagedAttention on a matched RC11f model pair

Two models built from the *identical* recipe (int4 body, int4 QMoE, int8 per-channel KV, same scale file); only the attention operator differs. `model.onnx.data` is byte-for-byte the same size in both.

| | result |
|---|---|
| **Accuracy** | MMLU-Pro-800 0.7200 (GQA) vs 0.7163 (paged); GPQA-diamond 0.6061 (GQA) vs 0.6212 (paged). **Equivalent** — both deltas are 3 questions and they point in opposite directions. |
| **Determinism** | 0/198 discordance on replicates of *both* models; generation is bit-reproducible. |
| **Memory** | Identical to within 24 MiB (0.16%) at matched KV capacity from 16k to 128k `max_length`. Paged costs nothing extra; its advantage is structural (shared pool vs `batch × max_length` reservation). |
| **Decode TPS** | b=1 **+0.5%**, b=8 −6.1%, b=32 **−22.3%** (see below). |

> **Note on the accuracy numbers.** An earlier revision of this description reported GPQA-diamond
> 0.5404 (GQA) vs 0.6212 (paged) and MMLU-Pro-800 0.6175 vs 0.7163. Those GQA figures were
> produced by an ORT bug, not by GQA itself: `FlashAttentionAndQuantizeKV` in
> `contrib_ops/cuda/bert/group_query_attention_impl.cu` — the only GQA prompt path taken when the
> KV cache is quantized — called `flash::mha_fwd()`, which had `constexpr void* head_sink = nullptr`
> hardcoded, so gpt-oss's attention sinks were silently disabled for the entire prompt on all 24
> layers (decode was unaffected). PagedAttention was immune because it applies the sink as an exact
> post-hoc epilogue on the FlashAttention `softmax_lse` rather than passing it into the kernel.
> The numbers above are with that fixed. **This PR is unaffected** — the bug is entirely on the ORT
> side and predates it — but the comparison needed an ORT with the fix to be meaningful.

| config | GQA tok/s | paged tok/s | delta |
|---|---|---|---|
| b=1, p=128, n=256 | 374.4 | **376.3** | **+0.5%** |
| b=8, p=128, n=256 | 1792.7 | 1683.8 | −6.1% |
| b=32, p=128, n=256 | 5285.9 | 4109.5 | **−22.3%** |
| b=2, p=4096, n=256 | 684.3 | **686.6** | +0.3% |

**Known gap:** PagedAttention is throughput-neutral at batch 1 and long context but loses ground as the batch grows. TTFT is 2–4% higher everywhere, which is the block-table setup and `ExpandBlockTableToPages` running per node per step. Hoisting that out of the per-step path is the next profiling target; the b=32 regression should be closed before proposing paged as a drop-in replacement for batch serving.

## Testing Notes

```bash
# Builder unit tests
python3 -m pytest test/python/builder -q          # 299 passed, 3 skipped
lintrunner -a                                     # clean
```

End-to-end validation on gpt-oss-20b (8×H200):

- **Bit-exactness:** decode output is byte-identical with CUDA graphs off and on.
- **Long context:** needle-in-a-haystack passes at 6,093 / 8,093 / 8,153 prompt tokens with graphs on. The 8,093 case crosses the 32→64 column bucket mid-decode, exercising mid-run re-capture.
- **Replay actually happens:** nsys over 94 decode steps shows 1 × `cudaStreamBeginCapture`, 1 × `cudaGraphInstantiate`, **94 × `cudaGraphLaunch`**.
- **Block accounting:** with `block_size = 256`, `num_blocks = 16` now serves a 4,079-token prefill plus 16 generated tokens (4,095 tokens in 4,096 slots — exactly full); `num_blocks = 15` is cleanly rejected.
- **Backward compatibility:** models built before this change have no `attention_metadata` input; `PrepareAttentionMetadata()` no-ops and they load and run unchanged.

To reproduce the builder side:

```bash
python -m models.builder -m openai/gpt-oss-20b -o $OUT -p int4 -e cuda \
  --extra_options use_paged_attention=true paged_block_size=256 enable_cuda_graph=true \
  kv_cache_quant_type=int8_per_channel kv_cache_scale_file=path_to_scales.json
```

### DIFF
--- a/src/python/py/models/builders/base.py
+++ b/src/python/py/models/builders/base.py
@@ -169,7 +169,7 @@ class Model:
             "block_table": ["batch_size", "max_num_blocks"],                                                     # For paged attention models
             "cumulative_sequence_lengths": ["batch_size + 1"],                                                   # For paged attention models
             "past_sequence_lengths": ["batch_size"],                                                             # For paged attention models
-            "attention_metadata": [2],                                                                          # For paged attention models. Static shape: a pair of scalars, not a per-sequence tensor.
+            "attention_metadata": [2],                                                                           # For paged attention models. Static shape: a pair of scalars, not a per-sequence tensor.
         }
         self.make_inputs_init()
 
--- a/src/python/py/models/builders/base.py
+++ b/src/python/py/models/builders/base.py
@@ -145,6 +145,7 @@ class Model:
             "block_table": "block_table",                                                                        # For paged attention models
             "cumulative_sequence_lengths": "cumulative_sequence_lengths",                                        # For paged attention models
             "past_sequence_lengths": "past_sequence_lengths",                                                    # For paged attention models
+            "attention_metadata": "attention_metadata",                                                          # For paged attention models
         }
         self.input_types = {
             "input_ids": ir.DataType.INT64,                                                                      # For standard models
@@ -156,6 +157,7 @@ class Model:
             "block_table": ir.DataType.INT32,                                                                    # For paged attention models
             "cumulative_sequence_lengths": ir.DataType.INT32,                                                    # For paged attention models
             "past_sequence_lengths": ir.DataType.INT32,                                                          # For paged attention models
+            "attention_metadata": ir.DataType.INT32,                                                              # For paged attention models
         }
         self.input_shapes = {
             "input_ids": ["batch_size", "sequence_length"],                                                      # For standard models
@@ -167,6 +169,7 @@ class Model:
             "block_table": ["batch_size", "max_num_blocks"],                                                     # For paged attention models
             "cumulative_sequence_lengths": ["batch_size + 1"],                                                   # For paged attention models
             "past_sequence_lengths": ["batch_size"],                                                             # For paged attention models
+            "attention_metadata": [2],                                                                          # For paged attention models. Static shape: a pair of scalars, not a per-sequence tensor.
         }
         self.make_inputs_init()
 
@@ -432,7 +435,7 @@ class Model:
             if "attention_mask" in self.input_names:
                 del self.input_names["attention_mask"]
         else:
-            for name in ["block_table", "cumulative_sequence_lengths", "past_sequence_lengths"]:
+            for name in ["block_table", "cumulative_sequence_lengths", "past_sequence_lengths", "attention_metadata"]:
                 del self.input_names[name]
 
     def make_outputs_init(self):
@@ -901,6 +904,7 @@ class Model:
             inputs["block_table"] = self.input_names["block_table"]
             inputs["cumulative_sequence_lengths"] = self.input_names["cumulative_sequence_lengths"]
             inputs["past_sequence_lengths"] = self.input_names["past_sequence_lengths"]
+            inputs["attention_metadata"] = self.input_names["attention_metadata"]
         if "past_key_values.key" in self.input_names:
             inputs["past_key_names"] = "past_key_values.%d.key"
         if "past_key_values.value" in self.input_names:
@@ -2902,6 +2906,7 @@ class Model:
                 cumulative_sequence_lengths=self.input_names["cumulative_sequence_lengths"],
                 past_sequence_lengths=self.input_names["past_sequence_lengths"],
                 block_table=self.input_names["block_table"],
+                attention_metadata=self.input_names["attention_metadata"],
                 **kwargs,
             )
         else:
@@ -3148,9 +3153,13 @@ class Model:
             k_scale_name, v_scale_name = self.get_kv_cache_scale_names(layer_id)
 
         # Optional trailing inputs of PagedAttention, in schema order:
-        #   10: slot_mapping, 11: head_sink, 12: q_norm_weight, 13: k_norm_weight, 14: k_scale, 15: v_scale
+        #   10: slot_mapping, 11: head_sink, 12: q_norm_weight, 13: k_norm_weight, 14: k_scale, 15: v_scale,
+        #   16: attention_metadata
         # The scheduler-provided slot_mapping is not used here; slots are derived from
         # past_seqlens / cumulative_sequence_length / block_table by the op.
+        # attention_metadata carries [max_query_len_bound, max_kv_len_bound] in CPU memory so the op
+        # can select a backend and size its launch without a device-to-host readback of the sequence
+        # lengths. Feeding it is what removes the per-node stream synchronization on the decode path.
         optional_inputs = [
             "",  # slot_mapping
             kwargs.get("sinks", ""),  # head_sink
@@ -3158,6 +3167,7 @@ class Model:
             k_norm_weight,
             k_scale_name,
             v_scale_name,
+            kwargs.get("attention_metadata", ""),
         ]
         while optional_inputs and not optional_inputs[-1]:
             optional_inputs.pop()
--- a/src/python/py/models/builders/base.py
+++ b/src/python/py/models/builders/base.py
@@ -2970,6 +2970,52 @@ class Model:
             output, self.io_dtype, shape=["batch_size", "sequence_length", self.head_size * self.num_attn_heads]
         )
 
+    def extend_with_optional_inputs(self, inputs, optional_inputs):
+        # ONNX lets trailing optional inputs be omitted entirely, so drop the unused ones at the
+        # end instead of emitting empty placeholders. Placeholders in the middle are kept because
+        # they reserve the position of the later inputs.
+        while optional_inputs and not optional_inputs[-1]:
+            optional_inputs.pop()
+        inputs.extend(optional_inputs)
+
+    def get_qk_norm_weight_inputs(self, **kwargs):
+        # Shared by GroupQueryAttention and PagedAttention: both fuse the Q/K RMSNorm and both
+        # require the pair to be provided together.
+        q_norm_weight = kwargs.get("q_norm_weight", "")
+        k_norm_weight = kwargs.get("k_norm_weight", "")
+        if bool(q_norm_weight) != bool(k_norm_weight):
+            raise ValueError("q_norm_weight and k_norm_weight must be provided together.")
+        return q_norm_weight, k_norm_weight
+
+    def get_kv_cache_scale_inputs(self, **kwargs):
+        # Shared by GroupQueryAttention and PagedAttention: returns the per-layer k/v scale
+        # initializer names, or empty placeholders when the KV cache is not quantized.
+        if self.kv_cache_quant_type == "none":
+            return "", ""
+        layer_id = kwargs.get("layer_id")
+        if layer_id is None:
+            raise ValueError("layer_id is required for quantized KV cache.")
+        return self.get_kv_cache_scale_names(layer_id)
+
+    def get_attention_op_attributes(self, **kwargs):
+        # Attributes shared by GroupQueryAttention and PagedAttention. Op-specific attributes
+        # (e.g. GQA's `kv_cache_bit_width`) are added by the caller.
+        attributes = {
+            "num_heads": self.num_attn_heads,
+            "kv_num_heads": self.num_kv_heads,
+            "scale": self.attention_attrs["scale"],
+            "local_window_size": self.window_size,
+            "softcap": self.attention_attrs["softcap"],
+            "do_rotary": self.attention_attrs["use_rope_in_attn"],
+            "rotary_interleaved": self.rope_attrs["interleaved"],
+        }
+        if kwargs.get("q_norm_weight", ""):
+            attributes["qk_norm_epsilon"] = kwargs.get("qk_norm_epsilon", self.attention_attrs["qk_norm_epsilon"])
+        if self.kv_cache_quant_type != "none":
+            attributes["k_quant_type"] = self.kv_quant_type
+            attributes["v_quant_type"] = self.kv_quant_type
+        return attributes
+
     def make_group_query_attention(self, name, **kwargs):
         inputs = [
             kwargs["q_path"],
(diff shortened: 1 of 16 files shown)


baijumeswani (dismissed):

src/python/py/models/builders/base.py
```diff
@@ -645,6 +645,14 @@ def make_quantized_kv_cache_init(self):
         self.kv_cache_bit_width = 4 if is_int4 else 8
         self.kv_quant_type = "PER_CHANNEL" if self.kv_cache_quant_type.endswith("per_channel") else "PER_TENSOR"
 
+        if self.use_paged_attention and is_int4:
```
kunal-vaishnavi, line 651: How much work is needed to support int4?

src/python/py/models/builders/base.py
```diff
@@ -3121,21 +3134,64 @@ def make_paged_attention(self, name, **kwargs):
             kwargs.get("cos_cache", ""),
             kwargs.get("sin_cache", ""),
         ]
+
+        q_norm_weight = kwargs.get("q_norm_weight", "")
```
kunal-vaishnavi: Some of these QK norm and KV scale logic checks and name appends are shared across GQA and PA. Can we refactor and share the logic between the two?

src/python/py/models/builders/base.py
tianleiwu: Not much. It can be added if needed.

src/python/py/models/builders/base.py
```diff
@@ -705,6 +713,11 @@ def make_kv_cache_scale_initializers(self):
                 f"got k={len(k_scales_per_layer)} v={len(v_scales_per_layer)}"
             )
 
+        # PagedAttention validates the per-channel scale shape and requires the canonical
+        # (kv_num_heads, 1, head_size) form, while GroupQueryAttention only looks at the element
+        # count. Emit the canonical shape on the paged path and keep the flat vector elsewhere.
+        scale_shape = (self.num_kv_heads, 1, self.head_size) if (per_channel and self.use_paged_attention) else (-1,)
+
```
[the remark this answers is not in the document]
tianleiwu: Added in 7c20fa88. `test_paged_per_channel_scale_initializers_use_canonical_shape` builds with `use_paged_attention=True` + `kv_quant_type="PER_CHANNEL"` and asserts the captured initializers have shape `(num_kv_heads, 1, head_size)`; `test_paged_per_tensor_scale_initializers_stay_scalar` guards that the reshape does not leak onto the per-tensor path, and the existing GQA per-channel test now also asserts the flat `(num_kv_heads * head_size,)` shape so a future change cannot silently swap the two.

src/python/py/models/builders/base.py
tianleiwu: Refactored in 7c20fa88. The pieces that were duplicated between `make_group_query_attention` and `make_paged_attention` are now shared helpers on `Model`:

- `get_qk_norm_weight_inputs(**kwargs)` — the q/k norm pairing check.
- `get_kv_cache_scale_inputs(**kwargs)` — the `layer_id` check plus the per-layer `k_scale`/`v_scale` name lookup (returns `("", "")` when the cache is not quantized).
- `get_attention_op_attributes(**kwargs)` — the common attributes (`num_heads`, `kv_num_heads`, `scale`, `local_window_size`, `softcap`, `do_rotary`, `rotary_interleaved`) plus the conditional `qk_norm_epsilon` and `k_quant_type`/`v_quant_type`.
- `extend_with_optional_inputs(inputs, optional_inputs)` — trims the unused trailing optional inputs instead of emitting empty placeholders.

Only the genuinely op-specific parts stay in each builder: the fixed input list and its ordering (GQA appends `k_scale, v_scale, q_norm, k_norm` at 12-15; PagedAttention appends `slot_mapping, head_sink, q_norm, k_norm, k_scale, v_scale, attention_metadata` at 10-16), and GQA's `kv_cache_bit_width` attribute, which PagedAttention does not have because it derives the cache element type from the tensor.

This also removed the special-cased empty `k_scale`/`v_scale` placeholders in the GQA path in favour of the same trailing-trim used by PagedAttention. Existing GQA input-position tests still pass unchanged, and `test_gqa_attributes.py` now covers the PagedAttention scale inputs, the `layer_id` requirement, and the unpaired q/k norm rejection so both ops exercise the shared helpers.

src/python/py/models/builders/base.py
tianleiwu: To be concrete about the scope, the builder side is the cheap part — dropping the `int4` guard in `make_quantized_kv_cache_init` and reusing the existing `(head_size + 1) // 2` packing is a few lines. The blockers are below it:

1. ORT: `PagedAttention`'s `T_CACHE` constraint is `{float16, bfloat16, int8, float8e4m3fn}` (`bert_defs.cc`), so there is no sub-byte cache type to bind to, and the paged FMHA/paged-decode kernels would need packed-int4 load/dequant paths.
2. GenAI engine: `PagedKeyValueCache` sizes blocks with `Ort::SizeOf(KeyValueCacheType(model))`, which is byte-granular. Sub-byte would need the block/slot arithmetic to become bit-granular (or to carry an explicit packed head size), and `block_size * head_size` would have to stay even so blocks do not straddle a byte.

So the work is mostly in the ORT kernels rather than in this PR. Keeping the explicit rejection here means users get a clear error instead of a type-binding failure at session load, and it can be relaxed as soon as the kernel side lands.

baijumeswani (dismissed):

src/python/py/models/builders/base.py
```diff
@@ -167,6 +169,7 @@ def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
             "block_table": ["batch_size", "max_num_blocks"],                                                     # For paged attention models
             "cumulative_sequence_lengths": ["batch_size + 1"],                                                   # For paged attention models
             "past_sequence_lengths": ["batch_size"],                                                             # For paged attention models
+            "attention_metadata": [2],                                                                          # For paged attention models. Static shape: a pair of scalars, not a per-sequence tensor.
```
kunal-vaishnavi: Looks like there's a slight spacing issue.

```suggestion
            "attention_metadata": [2],                                                                           # For paged attention models. Static shape: a pair of scalars, not a per-sequence tensor.
```

kunal-vaishnavi (dismissed):

src/python/py/models/builders/base.py
```diff
@@ -156,6 +157,7 @@ def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
             "block_table": ir.DataType.INT32,                                                                    # For paged attention models
             "cumulative_sequence_lengths": ir.DataType.INT32,                                                    # For paged attention models
             "past_sequence_lengths": ir.DataType.INT32,                                                          # For paged attention models
+            "attention_metadata": ir.DataType.INT32,                                                              # For paged attention models
```
kunal-vaishnavi, line 160: Another extra space

```suggestion
            "attention_metadata": ir.DataType.INT32,                                                             # For paged attention models
```

kunal-vaishnavi (approved):

src/python/py/models/builders/base.py
tianleiwu: Will fix it later.

PR #2334, opened 2026-07-29
Support windowed KV cache for sliding-window layers on CUDA

tianleiwu:
## Description

Models with alternating attention (for example gpt-oss-20b: 24 layers, sliding-window `W=128` interleaved with full attention) currently get a KV cache of `search.max_length` for **every** layer on CUDA, even though the sliding-window layers can never attend beyond their window. This PR lets the CUDA EP hold a window-sized KV cache for those layers, which cuts KV memory roughly in half on gpt-oss-20b at long context.

The reduced allocation was previously gated to `NvTensorRtRtx`, which evicts inside the EP. CUDA now gets the same treatment by relying on the new `sliding_window_cache` attribute on `com.microsoft::GroupQueryAttention`, which makes the kernel index the past/present buffers in cache-relative coordinates and evict by compaction.

**Depends on** microsoft/onnxruntime#29904 (`[CUDA] GQA support sliding_window_cache`). Without that ORT change the emitted `sliding_window_cache` attribute is unknown to the kernel, so this PR should not be merged before it.

### Runtime: window-sized KV cache on CUDA

| File | Change |
|------|--------|
| `src/models/kv_cache.cpp` | `DefaultKeyValueCache` now sizes sliding-window layers below `max_length` on CUDA as well as `NvTensorRtRtx`. CUDA gets `window_size + slack` (slack is at least `chunk_size - 1` so a whole prefill chunk fits); `NvTensorRtRtx` keeps exactly the window because its EP evicts internally. Restricted to `num_beams == 1` — beam reordering across a compacted cache is correct in principle but untested. |
| `src/models/kv_cache.cpp` | `RewindTo` throws a descriptive error when the requested position has already been evicted from a windowed cache, instead of silently returning misaligned keys. With `past_present_share_buffer_` the rewind is purely logical, while the kernel derives the resident range from `seqlens_k`, so the two only agree while nothing has been evicted. |
| `src/models/kv_cache.h` | Adds `windowed_cache_size_` (0 when no layer is windowed, so the guard is inert for full-size caches) and `current_length_`, tracked from `Update()`. |

### Config: `model.decoder.sliding_window.cache_slack`

| File | Change |
|------|--------|
| `src/config.h` | New `cache_slack` field (default 256) — the KV positions kept beyond `window_size` on EPs that evict by compaction. It amortizes compaction and doubles as the `RewindTo` budget. |
| `src/config.cpp` | Parses `cache_slack` in `SlidingWindow_Element`. |

### Model builder

| File | Change |
|------|--------|
| `src/python/py/models/builders/base.py` | Introduces `eps_with_windowed_kv_cache = {"trt-rtx", "cuda"}` and uses it for the `sliding_window` block in `genai_config.json` (now including `cache_slack`) and for `make_key_value_cache_shape`, so windowed layers get a distinct symbolic `sliding` sequence dim and ONNX shape inference does not unify them with the `max_length` full-attention layers. |
| `src/python/py/models/builders/base.py` | Emits `sliding_window_cache=1` on the GQA node for CUDA sliding-window layers. Subclasses (gemma, gpt-oss, smollm) already set `window_size = -1` around global layers, so the attribute lands only on windowed layers. |

### Tests

| File | Change |
|------|--------|
| `test/python/builder/test_windowed_kv_cache.py` | New, 24 offline tests: the `sliding_window_cache` GQA attribute (EP and per-layer gating), `make_key_value_cache_shape` dim renaming, and the `genai_config.json` `sliding_window` block including `cache_slack`. |
| `test/sliding_window_config_test.cpp` | New, 5 gtest cases for `cache_slack` parsing through `OgaConfig::Create`: accepted, optional, and rejected when misspelled, non-numeric, or out of int range. |

## Testing

- Python builder tests: `python -m pytest test/python/builder -q` → 311 passed, 3 skipped.
- C++ unit tests: `cmake --build <build-dir> --target unit_tests && ./unit_tests --gtest_filter='SlidingWindowConfigTest.*:*Rewind*:*Config*'` → all pass.
- The new builder tests were validated against the pre-change `base.py`, where 4 of them fail, confirming they are regression-catching.
- Backward compatibility: `cache_slack` is optional, so existing `genai_config.json` files keep loading. Non-CUDA/non-trt-rtx EPs, models without alternating attention, and beam search (`num_beams > 1`) all keep the previous full-length allocation.
- End-to-end (gpt-oss-20b, A100-80GB, int8 KV, `max_length=131072`): KV memory 5138 → 3090 MiB (−39.9%), process peak 18636 → 16556 MiB. GPQA-diamond accuracy 0.6010 → 0.5960 over 198 questions (McNemar p = 1.0, accuracy-neutral). Decode throughput regresses ~1–2% from the compaction launches.

## Motivation and Context

`local_window_size` in ORT's CUDA `GroupQueryAttention` narrows the attention *mask* only, never the cache: `CheckInputs` computes `present_sequence_length = max(total_sequence_length, past_sequence_length)`, and K/V are appended at the absolute index `past_seq_lens[b] + s`. Binding a shorter present buffer therefore fails shape validation outright. microsoft/onnxruntime#29904 adds the opt-in `sliding_window_cache` attribute that switches the kernel to cache-relative coordinates with shift compaction; this PR is the genai-side half that allocates the smaller cache and writes the attribute.

Note that windowed caching is not expected to be faster — the baseline already passes `local_window_size` to flash attention, so the sliding layers are already `O(W)` in compute. The feature buys memory, not speed.

## Checklist

- [x] Tests added/updated
- [x] No breaking changes (opt-in via config; existing configs and EPs unaffected)
- [ ] Documentation updated (if applicable)

### DIFF
--- a/src/models/kv_cache.cpp
+++ b/src/models/kv_cache.cpp
@@ -425,11 +425,20 @@ DefaultKeyValueCache::DefaultKeyValueCache(State& state)
     throw std::runtime_error("Graph capture is not supported with past_present_share_buffer set to false.");
   }
 
-  // Set the size after empty_past_ has been created with 0 for this field
-  if (state_.model_.p_device_->GetType() == DeviceType::NvTensorRtRtx && model_.config_->model.decoder.sliding_window.has_value() &&
-      model_.config_->model.decoder.sliding_window->window_size > 0) {
-    const int sliding_window_size = model_.config_->model.decoder.sliding_window->window_size;
-    const int max_length = state_.params_->search.max_length;
+  // Compute the capacity for sliding-window layers and apply it to the cache shapes.
+  // ComputeWindowedKvCacheSize() returns 0 when the config does not call for a windowed cache.
+  const DeviceType device_type = state_.model_.p_device_->GetType();
+  const int max_length = state_.params_->search.max_length;
+  const int windowed_cache_size = ComputeWindowedKvCacheSize(
+      device_type, model_.config_->model.decoder, state_.params_->search, max_length);
+
+  if (windowed_cache_size > 0 &&
+      // Beam reordering across a compacted cache is correct in principle but untested.
+      state_.params_->search.num_beams == 1) {
+    // Only a cache smaller than max_length ever evicts, and only then is RewindTo restricted.
+    if (windowed_cache_size < max_length) {
+      windowed_cache_size_ = windowed_cache_size;
+    }
 
     // Check if we need per-layer allocation for models with alternating attention patterns
     if (!model_.config_->model.decoder.sliding_window->layers.empty()) {
@@ -459,14 +468,14 @@ DefaultKeyValueCache::DefaultKeyValueCache(State& state)
       for (int model_layer_idx : model_.config_->model.decoder.sliding_window->layers) {
         auto it = model_layer_to_cache_slot.find(model_layer_idx);
         if (it != model_layer_to_cache_slot.end()) {
-          layer_shapes_[it->second][2] = std::min(max_length, sliding_window_size);
+          layer_shapes_[it->second][2] = windowed_cache_size;
         }
       }
       // Set shape_[2] to max of all layer shapes for RewindTo bounds checking
       shape_[2] = max_length;
     } else {
       // Uniform sliding window allocation (backward compatibility)
-      shape_[2] = std::min(max_length, sliding_window_size);
+      shape_[2] = windowed_cache_size;
     }
   } else if (past_present_share_buffer_) {
     // For fixed kv-cache models the cache size comes from the model graph,
@@ -548,6 +557,8 @@ void DefaultKeyValueCache::Add() {
 }
 
 void DefaultKeyValueCache::Update(DeviceSpan<int32_t> beam_indices, int total_length) {
+  current_length_ = total_length;
+
   // If we're sharing past & present buffers there is nothing to do here, so early exit
   if (past_present_share_buffer_)
     return;
@@ -593,6 +604,10 @@ void DefaultKeyValueCache::Update(DeviceSpan<int32_t> beam_indices, int total_le
 
 void DefaultKeyValueCache::RewindTo(size_t index) {
   if (past_present_share_buffer_) {
+    // Sliding-window layers hold only the most recent windowed_cache_size_ positions, anchored at
+    // cache index 0. Once anything has been evicted, position i no longer lives at cache offset i,
+    // so a rewind would silently read misaligned keys. Refuse instead of returning wrong logits.
+    CheckWindowedKvCacheRewind(windowed_cache_size_, current_length_, index);
     return;
   } else if (shape_[2] <= static_cast<int>(index)) {
     throw std::runtime_error("Requested length of rewind is greater than the current length.");
--- a/src/python/py/models/builders/base.py
+++ b/src/python/py/models/builders/base.py
@@ -53,6 +53,14 @@ class Model:
             else self.context_length
         )
         self.window_size = config.sliding_window if hasattr(config, "sliding_window") else -1  # default is -1 in GroupQueryAttention kernel
+        # EPs that can hold a KV cache smaller than max_length for sliding-window layers: trt-rtx
+        # evicts inside the EP, cuda and cpu evict inside GroupQueryAttention (sliding_window_cache=1).
+        self.eps_with_windowed_kv_cache = {"trt-rtx", "cuda", "cpu"}
+        # Positions kept beyond the window to cover a prefill chunk and amortize cache compaction.
+        # 0 means "use the EP default at runtime"; set explicitly to override.
+        # CUDA optimal: 0 (launch overhead dominates; attention is O(W) regardless of C).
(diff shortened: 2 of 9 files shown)


src/models/kv_cache.cpp
```diff
@@ -426,11 +426,36 @@ DefaultKeyValueCache::DefaultKeyValueCache(State& state)
...
+      state_.params_->search.num_beams == 1) {
     const int sliding_window_size = model_.config_->model.decoder.sliding_window->window_size;
     const int max_length = state_.params_->search.max_length;
 
+    int windowed_cache_size = std::min(max_length, sliding_window_size);
+    if (needs_cache_slack) {
+      const auto& chunk_size_opt = model_.config_->search.chunk_size;
+      const int chunk_size = chunk_size_opt.has_value() ? static_cast<int>(*chunk_size_opt) : 0;
+      const int slack = std::max(chunk_size - 1, model_.config_->model.decoder.sliding_window->cache_slack);
+      windowed_cache_size = std::min(max_length, sliding_window_size + slack);
```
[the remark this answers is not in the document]
tianleiwu: Good catch — fixed in 50d8881. The slack is now computed entirely in `size_t` (`chunk_size - 1` only when `*chunk_size > 0`, `cache_slack` clamped to `>= 0` with `std::max(0, ...)`), and the result is bounded by `max_length` before being narrowed back to `int`.

src/models/kv_cache.cpp
```diff
@@ -593,6 +620,18 @@ void DefaultKeyValueCache::Update(DeviceSpan<int32_t> beam_indices, int total_le
...
+    // so a rewind would silently read misaligned keys. Refuse instead of returning wrong logits.
+    if (windowed_cache_size_ > 0 && static_cast<int>(index) < current_length_ &&
+        current_length_ > windowed_cache_size_) {
+      throw std::runtime_error(
+          "Cannot rewind to " + std::to_string(index) + ": the sliding-window KV cache holds only the last " +
+          std::to_string(windowed_cache_size_) + " of " + std::to_string(current_length_) +
+          " positions, so the tokens needed at that point have already been evicted. Increase "
+          "model.decoder.sliding_window.cache_slack in genai_config.json to keep more positions "
+          "resident, or rewind before the cache fills.");
+    }
```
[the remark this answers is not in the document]
tianleiwu: The strict guard is intentional; the looser check would return wrong results.

With `past_present_share_buffer_` the rewind is purely logical — this branch just `return`s, there is nothing to copy or truncate. The GQA kernel re-derives the resident range from `seqlens_k` on every step as `[T - min(T, C), T)`, anchored at cache offset 0. After a rewind to `i` the kernel would read `[i - min(i, C), i)` while the buffer still physically holds `[T - min(T, C), T)`. Those two ranges coincide only when nothing has been evicted, i.e. exactly the `T <= C` case the current check already allows — even for an `i` that is still "resident", every row is at the wrong offset.

Making the resident case work requires actually left-shifting each sliding layer's cache by `(i - min(i, C)) - (T - min(T, C))` rows (the same compaction the kernel does), which is not implemented in this PR. Refusing is the safe behavior rather than silently returning misaligned logits.

I've added a comment at 50d8881 explaining this so the guard doesn't look like an off-by-one. If rewind-after-eviction becomes a real requirement, the shift is a follow-up.

src/models/kv_cache.cpp
```diff
@@ -426,11 +426,55 @@ DefaultKeyValueCache::DefaultKeyValueCache(State& state)
   }
 
   // Set the size after empty_past_ has been created with 0 for this field
-  if (state_.model_.p_device_->GetType() == DeviceType::NvTensorRtRtx && model_.config_->model.decoder.sliding_window.has_value() &&
-      model_.config_->model.decoder.sliding_window->window_size > 0) {
+  //
+  // Three execution providers can hold a KV cache smaller than max_length for sliding-window
```
kunal-vaishnavi: Should we move all of this logic into `windowed_kv_cache.cpp` since that class is about sliding window KV caches? As a side note, we need to refactor a lot of logic inside this file into more separated structures.

src/python/py/models/builders/base.py
```diff
@@ -53,6 +53,14 @@ def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
             else self.context_length
         )
         self.window_size = config.sliding_window if hasattr(config, "sliding_window") else -1  # default is -1 in GroupQueryAttention kernel
+        # EPs that can hold a KV cache smaller than max_length for sliding-window layers: trt-rtx
+        # evicts inside the EP, cuda and cpu evict inside GroupQueryAttention (sliding_window_cache=1).
+        self.eps_with_windowed_kv_cache = {"trt-rtx", "cuda", "cpu"}
+        # Positions kept beyond the window to cover a prefill chunk and amortize cache compaction.
+        # 0 means "use the EP default at runtime"; set explicitly to override.
```
kunal-vaishnavi, line 60: Can these be initialized inside `make_attention_init` instead?

tianleiwu: I put it here to close to window_size attribute, which is related to this PR.

For refactoring, we probably need to group them into a class like AttentionConfig. That's future work.

kunal-vaishnavi (approved):

PR #2343, opened 2026-07-30
[CUDA] Optimize Engine decode with scheduler-owned batched sampling

tianleiwu:
## Description

Continuous-batching Engine decode paid a large per-request CUDA synchronization and sampling cost after every model step. This change restructures token generation into deferred completion and scheduler-owned batched sampling, reducing host/device round trips and scaling sampling launches with distinct parameter groups while preserving per-request RNG, sequence, EOS, and logits-processor state.

This PR now includes the changes originally developed in stacked PRs #2345 and #2361.

### Deferred completion and host token mirror

| Area | Change |
|------|--------|
| `src/search.h`, `src/cuda/search_cuda.*` | Split CUDA greedy token selection into launch and completion phases so all requests can queue work before one synchronization. |
| `src/engine/request.*` | Maintain a host-side sequence mirror, enforce one sequence/beam per Engine request, and expose explicit prepare/complete generation phases. |
| `src/engine/decoders/*_decoder_io.cpp` | Build decoder inputs from the host mirror and batch contiguous logits casts. |

### Scheduler-owned batched sampling

| Area | Change |
|------|--------|
| `src/engine/scheduler.*`, `src/engine/scheduled_requests.*` | Own reusable sampling state and planning arrays per scheduler; group active requests by resolved sampling parameters. |
| `src/smartptrs.h`, `src/cuda/interface.cpp` | Add the device-neutral batched sampler interface and CUDA implementation with reusable packed logits, metadata, output, and workspace buffers. |
| `src/cuda/cuda_sampling.*`, `src/cuda/cuda_topk.*` | Support persistent per-request cuRAND states and make the local Top-K algorithm cache batch-size-aware. |

The fast path supports heterogeneous options, pinned seeds, scheduler reorder, and noncontiguous logits. Unsupported providers or search modes retain the existing per-request fallback.

### Tests and documentation

- Add CUDA tests for heterogeneous rows, stable seeded RNG across batch reorder, and Top-K cache reuse across different bucket sizes.
- Add the phased design and performance rationale in `docs/paged_attention_batching.md`.

## Testing

- CUDA 13.0 Release build for `sm_80`: `cmake --build build/cu130/Release --target unit_tests --parallel 12`.
- Full C++ suite: 137 tests run, 115 passed, 22 environment-dependent tests skipped, 0 failed.
- Focused CUDA sampling and continuous-decoding suite: 11 passed, 0 failed.
- Bit-exact comparison against the pre-change GQA Engine path: 628 tokens across seven fast-path and fallback scenarios.
- Ruff and Ruff-format passed; the changed C/CUDA sources were formatted with ClangFormat 20.1.8.
- H200 profiling showed batch-32 throughput improving from 4109 to 4999 tok/s across the combined phases.

## Motivation and Context

Profiling showed that the Engine issued one sampling pass and synchronization chain per request even though requests shared a CUDA stream. Deferring completion removes serialization; scheduler-owned sampling then combines homogeneous and heterogeneous request groups without coupling RNG state to scheduler position. Per-request max length, EOS sets, sequence state, and logits processors remain independent.

## Checklist

- [x] Tests added/updated
- [x] Documentation updated
- [x] No breaking public API changes
- [ ] CI passes


src/engine/request.h
```diff
@@ -83,18 +83,39 @@ struct Request : std::enable_shared_from_this<Request>,
    */
   DeviceSpan<int32_t> UnprocessedTokens();
 
+  /**
+   * @brief Returns the unprocessed tokens from the host-side mirror of the sequence.
+   * @return Span of unprocessed token IDs that is valid until the next call to GenerateNextTokens.
+   *
```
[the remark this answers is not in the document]
tianleiwu: Agreed, the contract was wrong. The span points into `tokens_host_`, so it is invalidated by any append to the sequence — `CompleteGeneration()`, `AddTokens()` or `Assign()` — and a `std::vector` reallocation makes it dangle rather than merely stale.

Fixed in dcd8eb5e; the comment now names those three call sites and tells callers to copy if the span must outlive them. The only in-tree callers (`VarlenDecoderIO::PrepareInputIds`, `StaticBatchDecoderIO`) copy immediately, so there is no live defect, but the documented contract was misleading.

src/engine/request.cpp
```diff
@@ -22,7 +22,11 @@ DeviceSpan<int32_t> AllocateOnDevice(GeneratorParams& params,
 }  // namespace
 
 Request::Request(std::shared_ptr<GeneratorParams> params)
-    : params_{params}, search_{CreateSearch(*params.get())} {}
+    : params_{params}, search_{CreateSearch(*params.get())} {
+  // The engine drives one independent search per request, so completion is batched: see
+  // ScheduledRequests::GenerateNextTokens().
+  search_->DeferCompletion(true);
```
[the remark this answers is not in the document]
tianleiwu: Good catch — both are reachable and both corrupt silently. Fixed in dcd8eb5e by rejecting them at construction.

`batch_size > 1`: `CompleteGeneration()` takes the last `new_token_count` entries of `GetNextTokens().CpuSpan()`, which is the *last* row rather than row 0. That is inconsistent with `UnprocessedTokens()` and `CurrentSequenceLength()`, which both read `GetSequence(0)`, so the mirror would drift to the wrong row.

`num_beams > 1`: worse than a wrong row. `CreateSearch` returns `BeamSearch_Cuda`, which does not override `DeferCompletion`/`CompleteGeneration` — they stay the base-class no-ops. So `next_tokens_buffer_.CopyDeviceToCpu()` never runs and `CpuSpan()` returns whatever was last in host memory.

The single-sequence assumption predates this PR, but the host mirror made it silent instead of merely wrong, so enforcing it is the right fix. `ContinuousDecodingTest::RunEngine` already sets `batch_size` back to 1 before creating requests, which confirms the intended contract; the guard just makes it explicit.

src/cuda/search_cuda.h
```diff
@@ -16,7 +16,14 @@ struct Search_Cuda : Search {
...
   bool IsDone() const {
-    cudaStreamSynchronize(GetStream());
+    // done_cpu_ is pinned host memory written by CheckForEOSAndPad on the device, so it is only
+    // meaningful once the stream has drained. Tracking whether such a launch is outstanding lets a
+    // caller that has already synchronized (see GreedySearch_Cuda::CompleteGeneration) skip a
+    // redundant synchronization here.
+    if (done_pending_) {
+      cudaStreamSynchronize(GetStream());
+      done_pending_ = false;
+    }
```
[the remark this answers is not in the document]
tianleiwu: Fixed in dcd8eb5e — now `CUDA_CHECK(cudaStreamSynchronize(GetStream()))`, matching the convention used elsewhere in this file.

The unchecked call predates this PR, but the point about `done_pending_` is specific to this change: clearing it after a failed sync would make the next `IsDone()` skip the wait entirely and read a stale `done_cpu_`. With `CUDA_CHECK` the throw happens before the assignment, so the flag stays set and the pending state is preserved.

src/cuda/interface.cpp
```diff
@@ -46,34 +47,250 @@ struct GpuMemory final : DeviceBuffer {
...
+    }
+
+    const auto params_less = [&](int lhs, int rhs) {
+      if (params[lhs].k != params[rhs].k)
+        return params[lhs].k < params[rhs].k;
+      if (params[lhs].p != params[rhs].p)
+        return params[lhs].p < params[rhs].p;
+      return params[lhs].temperature < params[rhs].temperature;
+    };
+    std::sort(row_order_.begin(), row_order_.end(), params_less);
```
baijumeswani: Because  std::sort  is unstable, equal sampling parameters do not guarantee identity ordering. With libstdc++, equal elements commonly begin reordering at 17 elements. Consequently:

- Request A can consume request B’s random stream.
- A pinned  random_seed  no longer controls that request’s output.
- Joining, leaving, or resizing the batch can unexpectedly change ongoing requests’ output.
- RNG streams become cross-contaminated between clients.

This affects the most important fast path: a large decode batch where all requests use the same sampling parameters and logits are contiguous.

Recommended fix: Use original-order state indices in the contiguous branch, or require  row_order_[i] == i  before taking that branch.  std::stable_sort  would also preserve identity for the uniform case, but explicitly matching score and state order is safer.

Add a CUDA regression test with:

- At least 17 contiguous rows from one allocation.
- Identical sampling parameters.
- Different fixed seeds.
- Comparison against separately allocated rows or an explicitly reordered batch.

The existing reorder test uses only two separately allocated rows, so it exercises the gather/scatter path and cannot detect this defect.

src/engine/scheduler.cpp
```diff
@@ -5,10 +5,26 @@
...
+  constexpr size_t default_static_batch_size = 4;
+  size_t max_batch_size = default_static_batch_size;
+  const auto& engine_config = model->config_->engine;
+  if (engine_config.dynamic_batching)
+    max_batch_size = std::max(max_batch_size, engine_config.dynamic_batching->max_batch_size);
+  if (engine_config.static_batching)
+    max_batch_size = std::max(max_batch_size, engine_config.static_batching->max_batch_size);
+
+  batched_sampling_plan_.Reserve(max_batch_size);
+  batched_sampler_ = model->p_device_scoring_->CreateBatchedSampler(
```
baijumeswani, line 18: `Scheduler` constructs the CUDA batched sampler using the configured `max_batch_size`, and `CudaBatchedSampler` immediately allocates its full GPU workspace for that capacity:

```cpp
CudaBatchedSampler(int max_batch_size, int vocab_size)
    : state_pool_{std::make_shared<CudaSamplerStatePool>(max_batch_size)} {
  EnsureCapacity(max_batch_size, vocab_size);
}
```

The workspace includes packed logits, multiple Top-K intermediate score/index buffers, and sampling buffers. Its dominant memory usage is approximately:

```text
32 bytes * max_batch_size * vocab_size
```

For a vocabulary of approximately 201K tokens, this is roughly:

| Configured maximum batch size | Approximate sampler workspace |
|---:|---:|
| 16 | 0.10 GiB |
| 32 | 0.19 GiB |
| 100 | 0.60 GiB |
| 256 | 1.53 GiB |

This memory is allocated when the scheduler is created, even if the application only processes one or a few concurrent requests. Because the model builder currently defaults dynamic batching to 100, users may incur hundreds of megabytes of additional GPU memory at model load without explicitly requesting that tradeoff. On memory-constrained GPUs, this could make engine creation fail even when the application's actual batch size would have fit.

### Suggested improvement

Keep `max_batch_size` as the scheduler and KV-cache capacity, but allocate the sampler's GPU workspace lazily:

1. Reserve the inexpensive host-side planning vectors to `max_batch_size`.
2. Initialize the CUDA sampler workspace at zero or a small capacity.
3. Grow it geometrically when an observed batch exceeds the current capacity, without exceeding the configured maximum.

baijumeswani: We can address this in a follow up pr if needed.

src/cuda/interface.cpp
tianleiwu: Good catch. Let me replace it by stable_sort .

tianleiwu: Fixed in 01be680f.

Two changes rather than just `stable_sort`, since you're right that matching score and state order explicitly is the safer contract:

1. `std::sort` -> `std::stable_sort`, so rows with equal sampling parameters keep their original relative order and each request stays bound to its own RNG state.
2. The contiguous branch now additionally requires `row_order_[row] == row`. `stable_sort` already guarantees that for a single bucket, but the fast path samples `scores[0]` in place while uploading state indices in packed order, so the invariant is load-bearing. With the explicit check, any future ordering change degrades to the gather/scatter path instead of silently swapping RNG streams.

Added `SamplingTests.SchedulerOwnedSamplerKeepsRngRowsAlignedForContiguousBatchCuda` with the shape you described: 20 rows (past the 17-element introsort threshold) carved out of a single allocation so the contiguous path is taken, identical sampling parameters on every row, distinct fixed seeds per row, compared against a second sampler fed separately allocated rows with the same seeds in the same order. It also asserts the sampled tokens are not all identical, so the comparison cannot pass trivially.

Verified the test is a real regression test: with `std::sort` and the identity check removed it fails (`RNG stream leaked across rows at step 3, row 19`), and it passes with the fix. Full `SamplingTests` suite is green (25 passed, 7 skipped for NvTensorRtRtx).

src/engine/scheduler.cpp
tianleiwu: Addressed in 01be680f rather than deferring, since the change turned out to be contained.

`CudaBatchedSampler`'s constructor no longer calls `EnsureCapacity(max_batch_size, vocab_size)`. It now only reserves the cheap host-side planning vectors (`row_order_`, `bucket_offsets_`, `bucket_params_`) to `max_batch_size`; the GPU workspace starts empty and `Sample()` grows it on the first observed batch. `EnsureCapacity` already doubled geometrically, so the only other change was clamping the doubling step to `max_batch_size` so growth cannot overshoot the configured capacity:

```cpp
batch_capacity_ = std::max(batch_size, std::min(max_batch_size_, std::max(4, batch_capacity_ * 2)));
```

Net effect for a 201K vocabulary with the model builder's default `max_batch_size` of 100: engine creation no longer pays ~0.6 GiB up front, and a workload that only ever runs 4 concurrent requests settles at ~25 MiB instead. `max_batch_size` still governs the scheduler and KV-cache capacity, which is unchanged.

The `curandState` pool is still sized to `max_batch_size` at construction. That is ~48 bytes per slot (12 KiB at 256), and pre-sizing it avoids a `cudaStreamSynchronize` on the `AddRequest` path when the pool grows, so it did not seem worth making lazy.

`SamplingTests` and the engine tests pass with the workspace now allocated on first use.

baijumeswani (approved):

PR #2344, opened 2026-07-30
decode baseline engine benchmark

prathikr:
### Summary

This benchmark establishes the Qwen 2.5 series baseline across model sizes and concurrency levels.

| Params (B) | Concurrency | TTFT p50 (ms) | TTFT p95 (ms) | ITL p50 (ms) | ITL p95 (ms) | Avg. tokens/s | Peak device memory (GiB) | Peak host memory (GiB) |
|---:|---:|---:|---:|---:|---:|---:|---:|---:|
| 0.5 | 1 | 102.2 | 105.8 | 18.1 | 18.4 | 97.3 | 50.059 | 2.330 |
| 0.5 | 4 | 135.9 | 273.4 | 3.9 | 18.5 | 690.7 | 50.059 | 2.578 |
| 0.5 | 8 | 198.9 | 326.6 | 5.4 | 5.5 | 1220.9 | 50.059 | 2.581 |
| 1.5 | 1 | 84.7 | 84.9 | 5.8 | 5.8 | 165.2 | 50.059 | 2.788 |
| 1.5 | 4 | 262.8 | 363.4 | 6.2 | 6.5 | 524.0 | 50.059 | 2.816 |
| 1.5 | 8 | 772.2 | 1751.0 | 26.9 | 27.6 | 380.5 | 50.059 | 2.825 |
| 7 | 1 | 549.1 | 750.7 | 27.4 | 32.9 | 57.3 | 52.059 | 3.468 |
| 7 | 4 | 984.8 | 1344.4 | 11.1 | 11.8 | 246.0 | 52.059 | 3.468 |
| 7 | 8 | 3321.7 | 6078.1 | 50.1 | 50.9 | 135.1 | 52.059 | 3.468 |

### Configuration

| Component | Value |
|---|---|
| Models | Qwen 2.5 0.5B / 1.5B / 7B Instruct |
| Execution provider | CUDA |
| ONNX Runtime | 1.29.0 |
| CUDA plugin EP | 0.1.0 |
| GenAI | 0.16.0-dev |
| Prompt length | 4K |
| Generation length | 256 tokens/request |
| Measured runs | 20 |
| Warmup runs | 5 |
| GPU | NVIDIA A100 |


benchmark/engine/benchmark-design.md
```diff
@@ -0,0 +1,204 @@
...
+Example prompts.json:
+```json
+{
+  "dataset": "RULER",
+  "version": "subset_v1",
+  "source_branch": "rulerv2-ns",
+  "description": "Single-file curated subset with 5 prompts per prompt-length bucket.",
+  "length_buckets": {
+    "4k": [
+      "[RULER 4K sample 0] Full prompt text goes here.",
```
baijumeswani: Could you confirm if the 4K here is the token length of the length of the prompt?

benchmark/engine/benchmark-design.md
```diff
@@ -0,0 +1,204 @@
...
+      "[RULER 4K sample 1] Full prompt text goes here.",
+      "[RULER 4K sample 2] Full prompt text goes here.",
+      "[RULER 4K sample 3] Full prompt text goes here.",
+      "[RULER 4K sample 4] Full prompt text goes here."
+    ],
+    "32k": [
```
baijumeswani, line 73: Should we also introduce 16K/18K, 48K and 96K as well?

benchmark/engine/benchmark-design.md
baijumeswani, line 1: Overall, looks good. Some feedback:

- Could we make the scenarios build on top of reusable components? Otherwise every scenario will implement its own computation of relevant metrics.
- Could we define what each metric means as part of the benchmark documentation? As in ttft is based on request arrival or start of request execution?
- Could we also add ort, genai versions to the result?

baijumeswani (dismissed):

benchmark/engine/benchmark-design.md
prathikr: it is token length

benchmark/engine/CMakeLists.txt
```diff
@@ -0,0 +1,42 @@
...
+  ${ORT_HEADER_DIR}
+)
+
+target_compile_features(engine_benchmark PRIVATE cxx_std_17)
+target_compile_definitions(engine_benchmark PRIVATE ORT_GENAI_BENCH_GENAI_VERSION="${VERSION_INFO}")
+
+target_link_libraries(engine_benchmark PRIVATE
+  onnxruntime-genai
+  ${ONNXRUNTIME_LIB}
+  nlohmann_json::nlohmann_json
```
tianleiwu: Blocking: configuring the top-level project with `-DENABLE_MODEL_BENCHMARK=ON` fails here because `nlohmann_json::nlohmann_json` is not defined. The repository dependency loader has already populated this FetchContent name, so the nested `FetchContent_MakeAvailable` does not add the upstream target. Please reuse the JSON dependency/include contract already established by the top-level build (or explicitly ensure the target exists), and add an opt-in configure/build smoke check so this path is exercised in CI.

benchmark/engine/scenarios/decode_baseline.cpp
```diff
@@ -0,0 +1,159 @@
...
+
+#include "common/utils.h"
+#include "ort_genai.h"
+
+#ifdef _WIN32
+#include <Windows.h>
+#include <dxgi1_4.h>
+#pragma comment(lib, "dxgi.lib")
+#endif
+  if (!IsAllowedConcurrency(concurrency)) {
```
tianleiwu: Blocking: this `if` is at namespace scope; the file is missing its namespace/function definitions and still contains the old `bool`/JSON-output implementation. A direct syntax-only compile fails here with `expected unqualified-id before 'if'`, and the file never defines the `ValidateConfig` or `Execute` overrides declared in `decode_baseline.h`. Please implement those class methods and return `ScenarioExecutionOutput`.

benchmark/engine/scenarios/decode_baseline.cpp
```diff
@@ -0,0 +1,159 @@
...
+    raw << "[";
+
+    for (int run = 0; run < measured_runs; ++run) {
+      auto generator = OgaGenerator::Create(*model, *params);
+      generator->AppendTokenSequences(*prompt_sequences);
+
+      std::vector<double> first_token_ms(static_cast<size_t>(concurrency), -1.0);
+      std::vector<double> e2e_ms(static_cast<size_t>(concurrency), 0.0);
+
+      const auto run_start = std::chrono::steady_clock::now();
```
tianleiwu: This timer starts after `AppendTokenSequences`, but append performs prompt `ComputeLogits` in ORT GenAI. The reported TTFT therefore excludes the entire prefill and is roughly first-token selection latency; E2E excludes the same dominant long-context work. Please start request timing before append/admission, end TTFT when the first token is observable, and report prefill/decode durations separately if both are useful.

benchmark/engine/scenarios/decode_baseline.cpp
```diff
@@ -0,0 +1,159 @@
...
+
+    auto prompt_sequences = OgaSequences::Create();
+    for (int i = 0; i < concurrency; ++i) {
+      tokenizer->Encode(prompt.c_str(), *prompt_sequences);
+    }
+
+    const size_t prompt_token_count = prompt_sequences->SequenceCount(0);
+    auto params = OgaGeneratorParams::Create(*model);
+    params->SetSearchOption("batch_size", static_cast<double>(concurrency));
+    params->SetSearchOption("max_length", static_cast<double>(prompt_token_count + static_cast<size_t>(generation_tokens)));
```
tianleiwu: `concurrency` is being implemented as one generator's static `batch_size`: every row has the same prompt and later receives the same batch TTFT/E2E. This does not exercise independent Engine requests, arrivals, queueing, per-request completion, or fairness, and duplicating one batch timestamp creates artificial percentile samples. Please submit independent requests to one Engine instance; if static batching is intentional, name this `batch_size` and report batch latency plus aggregate throughput instead of per-request concurrency metrics.

benchmark/engine/scenarios/decode_baseline.cpp
```diff
@@ -0,0 +1,159 @@
...
+        ++emitted_steps;
+      }
+
+      const auto run_end = std::chrono::steady_clock::now();
+      const double run_elapsed_ms = std::chrono::duration<double, std::milli>(run_end - run_start).count();
+
+      for (int i = 0; i < concurrency; ++i) {
+        e2e_ms[static_cast<size_t>(i)] = run_elapsed_ms;
+        const double ttft_ms = std::max(0.0, first_token_ms[static_cast<size_t>(i)]);
+        const double tps = static_cast<double>(generation_tokens) / std::max(0.001, run_elapsed_ms / 1000.0);
```
tianleiwu: This uses the requested token count even when `IsDone()` stops generation early, so EOS inflates throughput. The actual sequence delta is calculated below but not used, and batched rows also need post-EOS padding excluded. Please count actual emitted tokens per request, record the completion reason, and calculate aggregate throughput as total actual output tokens divided by workload makespan.

benchmark/engine/scenarios/scenario_base.cpp
```diff
@@ -0,0 +1,74 @@
...
+    result["error"] = nullptr;
+    result["core_metrics"] = {
+        {"summary",
+         {{"ttft_ms", {{"p50", output.ttft_p50_ms}, {"p95", output.ttft_p95_ms}}},
+          {"inter_token_latency_ms", {{"p50", output.inter_token_latency_p50_ms}, {"p95", output.inter_token_latency_p95_ms}}},
+          {"peak_device_memory_bytes", output.peak_device_memory_bytes},
+          {"steady_state_device_memory_bytes", output.steady_state_device_memory_bytes}}},
+        {"raw_requests", std::move(raw_requests)}};
+    result["scenario_metrics"] = output.scenario_metrics;
+  } catch (const std::exception& ex) {
```
tianleiwu, line 86: Catching the exception here produces a `status: failed` artifact, but `DispatchScenarios` never checks that status and unconditionally returns 0. Invalid configs, provider failures, and generation errors will therefore look successful to CI/regression automation. Please preserve the failed artifact while tracking failures and returning nonzero after dispatch (or rethrow after recording the result).

tianleiwu (changes requested):
Requesting changes because the opt-in benchmark target currently neither configures nor compiles. After those build blockers, the decode prototype also needs to align its timing, concurrency, token-accounting, and failure semantics with the Engine workload described by the requirements. I reproduced the CMake failure with `ENABLE_MODEL_BENCHMARK=ON` and independently reproduced the first source error with a syntax-only C++ compile.

benchmark/engine/scenario_dispatcher.cpp
```diff
@@ -0,0 +1,217 @@
...
+
+  const std::string html =
+      "<!doctype html><html><head><meta charset='utf-8'><title>Benchmark Results</title>"
+      "<style>body{font-family:Segoe UI,Arial,sans-serif;margin:16px}"
+      ".tabs{display:flex;gap:8px;flex-wrap:wrap;margin-bottom:12px}"
+      "button{padding:6px 10px;border:1px solid #999;background:#f4f4f4;cursor:pointer}"
+      "button.active{background:#e1f0ff;border-color:#5da9ff}"
+      "pre{background:#111;color:#ddd;padding:12px;border-radius:6px;overflow:auto}</style>"
+      "</head><body><h2>Scenario Results</h2><div id='tabs' class='tabs'></div><pre id='content'></pre><script>"
+      "const results = " + embedded.dump() + ";"
```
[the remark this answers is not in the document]
prathikr: removed

build.py
```diff
@@ -791,6 +791,10 @@ def build(args: argparse.Namespace, env: dict[str, str]):
         util.run(csharp_build_command, cwd=REPO_ROOT / "src" / "csharp")
         util.run(csharp_build_command, cwd=REPO_ROOT / "test" / "csharp")
 
+    engine_benchmark_dir = args.build_dir / "benchmark" / "engine"
+    if util.is_linux() and engine_benchmark_dir.is_dir():
+        util.setup_engine_benchmark_dependencies(args.build_dir, engine_benchmark_dir)
```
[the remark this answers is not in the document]
prathikr: will put it all behind a --engine-benchmark flag

benchmark/engine/scenarios/decode_baseline.cpp
```diff
@@ -0,0 +1,264 @@
...
+  output.ttft_p50_ms = Percentile(ttft_values, 50.0);
+  output.ttft_p95_ms = Percentile(ttft_values, 95.0);
+  output.inter_token_latency_p50_ms = Percentile(inter_token_latency_values, 50.0);
+  output.inter_token_latency_p95_ms = Percentile(inter_token_latency_values, 95.0);
+
+  memory.Stop();
+  output.peak_device_memory_mb = BytesToMb(memory.PeakDeviceBytes());
+  output.steady_state_device_memory_mb = BytesToMb(memory.SteadyStateDeviceBytes());
+
+  output.scenario_metrics = {
```
tianleiwu, line 250: The benchmark requirements call for fixed-seed captured outputs and reporting output mismatches/incomplete requests, but each run discards `request_tokens` and this result records only run-level E2E/throughput arrays. A request can change output or finish early and still produce `status: success`, and `raw_requests` lacks the required request-level E2E and tokens/sec values. Please persist each request's generated suffix (or a stable token hash plus count), completion state, E2E latency, and throughput, then validate the fixed-seed output across standalone and co-scheduled runs.

tianleiwu (changes requested):
The benchmark scaffolding is useful, but the result artifact still cannot enforce its stated correctness gate because generated outputs and per-request completion/latency evidence are discarded. Please address the inline finding before merge.

benchmark/engine/scenarios/decode_baseline.cpp
prathikr: Hi @tianleiwu, you're right that a request finishing early (or otherwise misbehaving) shouldn't be able to produce status: success, and that was a real gap.

After discussing with @baijumeswani, we're deprioritizing full correctness verification (captured/hashed outputs, cross-run token-stream comparison for fixed seeds, standalone-vs-co-scheduled matching) for this iteration — the benchmark's near-term goal is performance measurement, and building out output capture/comparison infrastructure is a larger effort we want to scope separately rather than bundle into this PR.

That said, we did take the core of your finding: a request can currently under-generate tokens and still be reported as a successful run. I have fixed this portion of the decode baseline benchmark.

baijumeswani (dismissed):

tianleiwu (dismissed):

baijumeswani (dismissed):

baijumeswani (approved):

PR #2347, opened 2026-07-31
Testing foundation for the continuous-batching engine

baijumeswani:
This lays the groundwork for testing the correctness of the continuous-batching Engine. The Engine juggles request scheduling, a paged KV-cache, and model execution, but there's no direct way to check that its internal bookkeeping stays consistent. This PR adds the scaffolding and the correctness contracts to start doing that.

### Changes

- A small production fix: `BlockPool::Free` now validates every block in a batch (non-null, in-range id, actually owned by this pool, not listed twice) before touching any pool state, so a bad request throws up front instead of half-freeing the batch and corrupting the free list.
- A new `engine_unit_tests` binary: a white-box suite that links the genai object library so it can reach internal Engine types, kept separate from the public-API `unit_tests` so test-only helpers don't leak into it.
- State snapshots and a validator: immutable, read-only snapshots of Request and paged-cache state, plus one validator that defines the Engine's invariants in a single place (block accounting, single block ownership, slot capacity, and the sequence-length relationships per request).
- A dependency-injection seam for the engine.

### DIFF
--- a/src/engine/request.cpp
+++ b/src/engine/request.cpp
@@ -82,6 +82,19 @@ int64_t Request::CurrentSequenceLength() const {
   return search_->GetSequenceLength();
 }
 
+RequestStateSnapshot Request::Snapshot() const {
+  const int64_t current = CurrentSequenceLength();
+  RequestStateSnapshot snapshot;
+  snapshot.request_id = this;
+  snapshot.status = status_;
+  snapshot.current_sequence_length = current;
+  snapshot.processed_sequence_length = processed_sequence_length_;
+  snapshot.seen_sequence_length = seen_sequence_length_;
+  snapshot.unprocessed_token_count = current - processed_sequence_length_;
+  snapshot.is_prefill = is_prefill_;
+  return snapshot;
+}
+
 int32_t Request::UnseenToken() {
   auto sequence = search_->GetSequence(0).CopyDeviceToCpu();
   if (static_cast<size_t>(seen_sequence_length_) >= sequence.size())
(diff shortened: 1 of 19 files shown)


src/engine/request.cpp
```diff
@@ -82,6 +82,19 @@ int64_t Request::CurrentSequenceLength() const {
...
 
+RequestStateSnapshot Request::Snapshot() const {
+  const int64_t current = CurrentSequenceLength();
+  RequestStateSnapshot snapshot;
+  snapshot.request_id = this;
+  snapshot.status = status_;
+  snapshot.current_sequence_length = current;
+  snapshot.processed_sequence_length = processed_sequence_length_;
+  snapshot.seen_sequence_length = seen_sequence_length_;
+  snapshot.unprocessed_token_count = current - processed_sequence_length_;
```
tianleiwu: `Snapshot()` computes `unprocessed_token_count` directly as `current - processed_sequence_length_`, and `ValidateRequestInvariants()` later compares it with that exact expression. This makes the check tautological for every production snapshot; only a hand-constructed snapshot like the unit test can violate it. Please either remove the redundant field/check or populate it from an independent source, and add a test that reaches the validator through `Request::Snapshot()` so the contract stays aligned with producible state.

tianleiwu (commented):
The test foundation is well scoped: invalid block frees are now atomic, snapshots keep validation model/device independent, and the dedicated target is wired into the modified CI workflows. One non-blocking validator issue remains: the unprocessed-token relationship is derived rather than independently observed, so its production check cannot currently detect a mismatch.

src/engine/request.cpp
baijumeswani: Thanks, I removed the unnecessary field/check.

tianleiwu (approved):

baijumeswani:
Thank you for the review. :)

PR #2351, opened 2026-07-31
Qwen3.6: export ModelOpt NVFP4/FP8 checkpoints natively

tianleiwu:
### Description

> Stacked on #2350 (`tlwu/20260731/nvfp4_quant_core`). Review only the last commit; the base branch will be retargeted to `main` once #2350 merges.

Wires the NVFP4 / FP8 scheme into the Qwen3.6 MoE builder so a TensorRT Model Optimizer quantized checkpoint is carried into the ONNX model instead of being dequantized to fp16 and re-quantized to int4.

* **NVFP4 routed experts** — `moe_quant_type=nvfp4` emits a native NVFP4 `QMoE` op built straight from the checkpoint's packed E2M1 codes and E4M3 block scales (block size 16).
* **NVFP4 dense modules** — `use_original_nvfp4_weights=true` emits the shared expert and `lm_head` as the weight-only `MatMulBlockQuantizedFp4Weight` contrib op, feeding the checkpoint tensors through unmodified. `nvfp4_dense_exclude_layers` / `nvfp4_lmhead_fp16` keep selected modules at fp16 for A/B.
* **FP8 dense projections** — `use_original_fp8_weights=true` emits the self-attention `q/k/v/o` projections as `MatMulBlockQuantizedFp8Weight`; `fp8_linear_attn` (default on) does the same for the GatedDeltaNet `in_proj_qkv` / `in_proj_z` / `out_proj`. Optional calibrated per-tensor activation scales via `fp8_attn_static_input_scale`, `fp8_linear_attn_static_input_scale` and `share_fp8_attn_qkv_activation`; `fp8_attn_exclude_layers` isolates a layer for error analysis.
* **FP8 KV cache** — `fp8_kv_cache=true` with `kv_cache_scale_file=...` stores the KV cache as E4M3, supporting both calibrated per-layer scales (see #2348) and the legacy shared unit scale.
* **Gate fusion** — `fuse_linear_attn_gates` (default on for CUDA) collapses the float32 gate glue around `LinearAttention` into `com.microsoft::LinearAttentionGate` and `com.microsoft::GatedRMSNorm`, removing ~9 nodes per linear-attention layer and the corresponding CUDA-graph replay overhead.

### Motivation and Context

The reference Qwen3.6 checkpoint is already NVFP4/FP8 quantized. Round-tripping through fp16 loses the calibrated scales and inflates both the export and the runtime weight traffic. These options require an ONNX Runtime build providing the corresponding contrib ops.

Adds unit tests for the FP8 activation-scale and FP8 linear-attention paths, and documents the new extra options in the model builder README.


src/python/py/models/builders/qwen.py
```diff
@@ -698,7 +706,7 @@
         # Fix model_type: HF architecture "Qwen3VLForConditionalGeneration" would produce "qwen3vl"
         # but the C++ runtime expects "qwen3_vl" (with underscore).
         # Intentional override of the superclass attribute (used in genai_config.json).
-        self.model_type = "Qwen3_VLForConditionalGeneration"  # noqa: overrides Model.model_type on purpose
+        self.model_type = "Qwen3_VLForConditionalGeneration"
```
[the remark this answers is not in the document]
tianleiwu: Addressed in 609e98f1. This override is intentional: the Hugging Face architecture name maps to a different C++ runtime model type, as documented immediately above the assignment. The exact CodeQL rule is now suppressed on that line.

src/python/py/models/builders/qwen.py
```diff
@@ -2101,13 +2237,90 @@
...
+        )
+
         # The base builder derives the GenAI model.type by stripping the suffix
         # after "For" and lowercasing, matching Qwen3.5 text-only export.
-        self.model_type = (
-            "Qwen3_5_Moe_textForCausalLM"
-            if self.is_text_only
-            else "Qwen3_5_MoeForConditionalGeneration"
-        )
+        self.model_type = "Qwen3_5_Moe_textForCausalLM" if self.is_text_only else "Qwen3_5_MoeForConditionalGeneration"
```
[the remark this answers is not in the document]
tianleiwu: Addressed in 609e98f1. The MoE subclass must override the inherited model type so generated GenAI configuration identifies Qwen3.5-MoE rather than the dense parent; the nearby comment documents this, and the exact CodeQL rule is now suppressed on the assignment.

src/python/py/models/builders/qwen.py
```diff
@@ -698,7 +706,7 @@
         # Fix model_type: HF architecture "Qwen3VLForConditionalGeneration" would produce "qwen3vl"
         # but the C++ runtime expects "qwen3_vl" (with underscore).
         # Intentional override of the superclass attribute (used in genai_config.json).
-        self.model_type = "Qwen3_VLForConditionalGeneration"  # noqa: overrides Model.model_type on purpose
+        self.model_type = "Qwen3_VLForConditionalGeneration"  # lgtm[py/overwritten-inherited-attribute]
```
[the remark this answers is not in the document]
tianleiwu: Addressed in a18a8708 by removing the suppression comment rather than adding another one.

The previous attempt annotated the assignment with `# lgtm[py/overwritten-inherited-attribute]`. Code scanning does not honor `lgtm[...]` suppression comments, so the alert simply re-fired against the new line content (alert 789 -> alert 1081). Worse, editing the comment pulled this pre-existing line into the PR diff, which is the only reason the bot annotates it here at all — the assignment itself is unchanged from `main` and is not part of this PR.

Both lines (`Qwen3VLTextModel` and `Qwen35MoeTextModel`) are now byte-identical to `main`, so they are out of the diff. The override remains intentional and documented in the comment directly above it: the Hugging Face architecture name does not map to the model type the C++ runtime expects in `genai_config.json`. The same pattern already exists elsewhere in this file on `main` (`VideoChat_Flash_QwenForCausalLM`, `Qwen3_5_textForCausalLM`).

src/python/py/models/builders/base.py
```diff
@@ -3932,6 +3932,39 @@ def make_fp8e4m3_initializer(self, scales_uint8, name):
         value.const_value = ir_tensor
         self.model.graph.register_initializer(value)
 
+    @staticmethod
```
kunal-vaishnavi: Let's remove any `staticmethod` decorators.

src/python/py/models/builders/qwen.py
```diff
@@ -2129,6 +2298,297 @@ def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
...
+            return f"model.language_model.layers.{layer_id}.self_attn.{proj}"
+
+        # GatedDeltaNet: only in_proj_qkv / in_proj_z / out_proj are stored as FP8 in the
+        # ModelOpt checkpoint (they carry .input_scale + .weight_scale). in_proj_a / in_proj_b
+        # are BF16 and only 32 elements wide, so they stay on the float path.
+        if proj not in {"in_proj_qkv", "in_proj_z", "out_proj"}:
+            return None
+        return f"model.language_model.layers.{layer_id}.linear_attn.{proj}"
+
+    def _fp8_attention_input_scale(self, basename):
```
kunal-vaishnavi: A lot of these methods are generic enough for many models, not just Qwen models. Can we generalize these new methods inside the base class?

src/python/py/models/builders/qwen.py
```diff
@@ -2231,40 +2717,210 @@ def make_moe(self, layer_id, mlp, root_input):
         )
         self.layernorm_attrs["skip_input"] = f"{combine_name}/output_0"
 
+    # ------------------------------------------------------------------
+    # NVFP4 (Model Optimizer) pre-quantized expert loading
+    # ------------------------------------------------------------------
+    def _nvfp4_snapshot_dir(self):
```
kunal-vaishnavi: For NVFP4/NVFP8 model loading, can we create a new class inside `quantized_model.py` to load the weights and preprocess them over there first? I'd like to keep this clean.

src/python/py/models/builder.py
```diff
@@ -168,6 +168,10 @@ def check_extra_options(
         "fuse_qk_norm_gqa",
         "prune_lm_head",
         "use_paged_attention",
+        "fp8_kv_cache",
+        "fuse_linear_attn_gates",
+        "use_original_fp8_weights",
```
kunal-vaishnavi: Rather than add new extra options, can we instead assume some defaults?

- For example, linear attention gates and QK norm should be fused for best performance now. I opened a [PR](https://github.com/microsoft/onnxruntime/pull/32106) to add the remaining kernels for our EPs.
- For the original weights, can we create new precisions for the `--precision` flag to support (e.g. `-p fp8`, `-p nvfp4`)?

src/python/py/models/builder.py
```diff
@@ -355,15 +377,7 @@ def parse_extra_options(
             kv_pairs[kv[0].strip()] = kv[1].strip()
 
     print(f"Extra options: {kv_pairs}")
-    check_extra_options(
```
kunal-vaishnavi, line 358: Do we need all of these linter changes? Can we keep the style as is instead?

src/python/py/models/builders/qwen.py
```diff
@@ -1960,6 +2074,27 @@ def _make_gated_rms_norm(self, basename, input_name, gate_name, norm_module, lay
         hv = self.linear_value_head_dim
         nv = self.linear_num_value_heads
 
+        # Norm weight (NO offset — Qwen3_5RMSNormGated uses raw weight, not 1+w)
+        norm_weight = f"model.layers.{layer_id}.linear_attn.norm.weight"
+        self.make_initializer(norm_module.weight, norm_weight, to=self.io_dtype)
+
+        if self.fuse_linear_attn_gates:
```
kunal-vaishnavi, line 2073: I am cleaning up the entire `qwen.py` file in [this PR](https://github.com/microsoft/onnxruntime-genai/pull/2386). A lot of the file's code in the main branch is superfluous. I am also adding all of the fused contrib ops that we can use to improve performance and implementing any missing kernels in ORT PRs such as [here](https://github.com/microsoft/onnxruntime/pull/31728) and [here](https://github.com/microsoft/onnxruntime/pull/32106).

src/python/py/models/builders/base.py
tianleiwu: Addressed in `958ccd5fa1`. I removed the `@staticmethod` decorators; the ModelOpt helpers are now normal instance methods on `Model`.

src/python/py/models/builders/qwen.py
tianleiwu: Addressed in `958ccd5fa1`. The generic scalar/shape validation, activation-scale sharing, native FP8/NVFP4 matmul emission, and NVFP4 repacking helpers now live on `Model` in `builders/base.py`. Qwen keeps only the topology-specific MoE assembly.

src/python/py/models/builders/qwen.py
tianleiwu: Addressed in `958ccd5fa1`. `ModeloptModel` in `quantized_model.py` now loads, materializes, and validates the native FP8/NVFP4 tensors. The Qwen builder consumes the structured modules and no longer performs raw checkpoint file I/O.

src/python/py/models/builder.py
tianleiwu: Addressed in `958ccd5fa1`. I removed the new opt-in flags and made the CUDA fusion/native-format selection automatic from ModelOpt metadata. I kept `--precision` as the surrounding graph and I/O dtype rather than adding `fp8`/`nvfp4` values, because a ModelOpt checkpoint is mixed-format and each quantized tensor retains its checkpoint dtype independently. Tests cover FP16, BF16, and FP32 graph precision with native NVFP4 experts.

src/python/py/models/builder.py
tianleiwu: Addressed in `958ccd5fa1`. I reverted the unrelated `builder.py` formatting churn and kept the pre-existing style while preserving the functional ModelOpt metadata handling. The complete changed-file set passes `lintrunner -a -r HEAD`.

src/python/py/models/builders/qwen.py
tianleiwu: Thanks for the heads-up. The refactor in `958ccd5fa1` minimizes overlap with #2386 by moving generic ModelOpt loading and native quantized graph emission out of `qwen.py`; the Qwen class now retains only Qwen3.5 topology-specific behavior and native QMoE initializer assembly.

src/python/py/models/quantized_model.py
```diff
@@ -1723,10 +1730,10 @@
         try:
             self.layers.extend(self._build_layer(layer_id) for layer_id in range(num_layers))
 
-            # Globals: embeddings + final norm are BF16; lm_head is NVFP4.
+            # Globals: embeddings + final norm are BF16; lm_head retains its native NVFP4 tensors.
             self.embedding.weight = self._get("model.language_model.embed_tokens.weight")
             self.final_norm.weight = self._get("model.language_model.norm.weight")
-            self.lm_head.weight = self._dequant_linear("lm_head")
+            self.lm_head = self._linear_module("lm_head")
```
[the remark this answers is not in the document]
tianleiwu: Addressed in `756c3a184b`. `QuantizedModel` now accepts the LM-head module it should initialize, and `ModeloptModel` supplies a `ModeloptLinearModule`. The loader populates that existing module in place instead of overwriting the inherited `lm_head` attribute; the loader test now also asserts the preserved module type.

baijumeswani (approved):

PR #2352, opened 2026-07-31
Add MTP self-speculative decoding runtime

tianleiwu:
### Description

Adds the runtime side of MTP (multi-token prediction) self-speculative decoding: a draft head shares the target model's hidden states, proposes `N` tokens, and a single batched verify forward accepts a prefix.

**Config / model plumbing**
* `Config::Mtp` describes the auxiliary `mtp.onnx` (filename, its own KV layout, and the main decoder's `hidden_states` output it consumes); parsed from the `model.mtp` section of `genai_config.json`.
* The decoder gains an optional `hidden_states` input/output pair, plumbed through `HiddenStatesInputs` (`src/models/hidden_states_inputs.{h,cpp}`).
* `Model`/`State` gain `GraphIdForLength` with a `graph_ids_` map instead of a single `graph_id_`, so one session can hold CUDA graphs for several token counts (1 for decode, `N+1` for verify). `max_graph_capture_length` bounds what is captured.
* `RecurrentState` supports a window-major state layout, and `State` gains `SnapshotState` / `CanCropRecurrentState` / `CropToAccepted` / `SetHiddenStates` so a partial accept can crop the recurrent state instead of replaying the main model.
* `CreateTensor` takes an optional `static_capacity_bytes` so graph-captured buffers keep a stable address across shrinking shapes.

**Speculative loop**
* `MtpGenerator` (`src/mtp_generator.{h,cpp}`) drives draft/verify/accept and exposes forward / accept / trial counters.
* On the multi-token greedy path the post-verify head refeed is deferred and **fused into the next step's first draft**: the accepted drafts and the newly committed token are consecutive positions that all pair with main-model hidden states, so one `(a+1)`-token head forward both re-materializes the accepted drafts in the head KV and yields the next draft from its last row. This holds the head at exactly `N` forwards per step instead of `N + ~0.9` (measured -4.5% ms/step at N=3 on H200, acceptance rate unchanged).
* The CUDA greedy draft chain keeps each draft ID on device: device-output ArgMax feeds the next head forward directly, and all `N` IDs are copied to the host once before acceptance. Non-CUDA devices retain the host path; `ORT_MTP_DEVICE_DRAFT_CHAIN=0` opts out and `ORT_MTP_VALIDATE_DEVICE_DRAFT_CHAIN=1` compares device and host ArgMax results. At N=3 on H200 this reduced 8.027 to 7.929 ms/step (-1.22%) and stream synchronizations from 11 to 9 per step; 300/300 validation drafts matched.
* `speculative_sampling.h` implements the standard rejection-sampling correction; `sampling_distribution.h` holds the shared distribution helpers.
* `DeviceInterface` gains `ArgMax`, `ArgMaxDevice`, `Top2` and `TopKScores`, appended at the end of the vtable for ABI stability, with CUDA implementations in `src/cuda/interface.cpp`.

**APIs**
* C: `OgaCreateMtpGenerator`, `OgaMtpGenerator_{AppendTokens,GenerateNextToken,IsDone,GetSequenceCount,GetSequenceData,GetForwardCount,GetAcceptCount,GetTrialCount}`, `OgaDestroyMtpGenerator`, plus `OgaGenerator_SnapshotState` / `OgaGenerator_SetHiddenStates`.
* C++ wrappers in `ort_genai.h`; Python `MtpGenerator` (`append_tokens` / `generate_next_token` / `is_done` / `get_sequence` / `get_stats`).

**Example**
* `examples/python/qwen-3.6-mtp.py` and a walkthrough in `examples/python/qwen-3.6-mtp.md`.

### Motivation and Context

Qwen3.6 ships an MTP head; using it turns each decode step into one draft pass plus one batched verify pass, trading a small amount of extra compute for several tokens per step. The corresponding builder support is in the follow-up PR.

### Validation

* GenAI `UnitTests` pass.
* CUDA device/host ArgMax validation matched 300/300 draft IDs over 100 steady N=3 steps.
* H200 final smoke test: 7.945 ms/step with `head_fwd/step=3.00`.

**Notes for reviewers**
* Supersedes the closed #2218.
* Touches `src/smartptrs.h` and `src/cuda/interface.cpp`, which #2345 also modifies (different methods — textual conflict only), and `src/config.{h,cpp}`, which #2334 also modifies.


src/models/hidden_states_inputs.cpp
```diff
@@ -0,0 +1,121 @@
...
+  // The concrete per-length buffer is bound in Update() (called before every Run); start unbound.
+  state_.inputs_.push_back(nullptr);
+  state_.input_names_.push_back(model_.config_->model.decoder.inputs.hidden_states.c_str());
+}
+
+Tensor* HiddenStatesInputs::GetOrCreateBuffer(int sequence_length) {
+  auto it = buffers_by_len_.find(sequence_length);
+  if (it != buffers_by_len_.end())
+    return it->second.get();
+
```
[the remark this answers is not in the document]
tianleiwu: Addressed in 068dc5d1. Hidden-state inputs and outputs now keep dedicated stable buffers only for graph-captured lengths; uncaptured prompt/chunk lengths reuse a single dynamic buffer instead of accumulating per-length allocations.

src/mtp_generator.cpp
```diff
@@ -0,0 +1,929 @@
...
+  if (const char* env = std::getenv("ORT_MTP_PREFILL_CHUNK")) {
+    prefill_chunk_ = std::atoi(env);
+    if (prefill_chunk_ < 0) prefill_chunk_ = 0;
+    prefill_chunk_explicit_ = true;
+  }
+  // Capture the 1-token decode and the verify shapes up to N+1 tokens.
+  auto& main_params = const_cast<GeneratorParams&>(params);
+  main_params.max_graph_capture_length = num_speculative_tokens_ + 1;
+
+  main_ = CreateGenerator(main_model_, params);
```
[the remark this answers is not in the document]
tianleiwu: Addressed in 068dc5d1. MtpGenerator now clones all relevant caller settings into an MTP-owned main_params_ object and sets max_graph_capture_length there, so the const input params are no longer mutated or observable by later generators.

src/models/recurrent_state.cpp
```diff
@@ -145,16 +212,29 @@ void RecurrentState::Update() {
...
+    // The recurrent (conv + linear-attention) state cannot be cropped like the
+    // attention KV cache. A partial rewind is only possible by restoring a
+    // snapshot captured at the target length (used by speculative decoding to
+    // roll back a rejected draft). The caller (e.g. the MTP orchestrator)
+    // guarantees the snapshot was taken at exactly `index`.
+    if (snapshot_valid_) {
+      RestoreSnapshot();
+      return;
+    }
     throw std::runtime_error(
```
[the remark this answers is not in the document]
tianleiwu: Addressed in 068dc5d1. SnapshotState now records the generator sequence length with the recurrent snapshot, and RewindTo rejects a nonzero target that does not match that captured length before restoring state.

src/generators.cpp
```diff
@@ -318,6 +318,17 @@ DeviceInterface* OrtGlobals::LoadCudaInterface(DeviceType type) {
...
+    using GetInterfaceVersionFn = uint32_t (*)();
+    auto get_interface_version = reinterpret_cast<GetInterfaceVersionFn>(cuda_library_->GetSymbol("GetInterfaceVersion"));
+    if (!get_interface_version)
+      throw std::runtime_error(
+          "CUDA add-on library does not export GetInterfaceVersion; install a matching onnxruntime-genai-cuda package");
+    const uint32_t interface_version = get_interface_version();
+    if (interface_version != kDeviceInterfaceVersion)
+      throw std::runtime_error(
+          "CUDA add-on interface version mismatch: expected " + std::to_string(kDeviceInterfaceVersion) +
+          ", got " + std::to_string(interface_version) + "; install a matching onnxruntime-genai-cuda package");
```
baijumeswani, line 330: Could we separate the versioning of the interface library into another PR if possible?

src/mtp_generator.cpp
```diff
@@ -0,0 +1,1065 @@
...
+    } else {
+      const float* rowf = MainLogitsRowsCpu(last, 1);
+      auto processed = ProcessMainLogitsRow(
+          std::span<const float>(rowf, static_cast<size_t>(vocab_size_)), last);
+      ComputeSampledCategorical(processed,
+                                top_k_, top_p_, temperature_, sampled_scratch_);
+      idx = sampled_scratch_.indices;
+      prob = sampled_scratch_.probs;
+    }
+    next_token_ = SampleSparseToken(idx, prob, rng_);
```
baijumeswani, line 598: After prefill, the wrapped  `main_` generator may already be done, but `MtpGenerator` ignores that state and primes `next_token_`. The next call pushes it before checking the limit, producing `max_length + 1`  tokens. The same mismatch applies to terminal prompt/EOS semantics.

src/mtp_generator.cpp
```diff
@@ -0,0 +1,1065 @@
...
+  //     forward validates N+1 tokens). The batched (M=N+1) forward is numerically ~equal but not
+  //     bit-identical to single-token decode (different GEMM tiling for M=1 vs M>1, plus XQA-vs-
+  //     Flash attention), so a greedy argmax can occasionally differ on near-ties. This is the same
+  //     tradeoff the N=1 verify already makes; the reject path below re-runs decode-consistently to
+  //     bound divergence from plain greedy. ---
+  // SnapshotState() copies every linear-attention layer's conv+recurrent state (2*num_layers D2D
+  // copies + launches) on EVERY step. It is only needed by the RewindToLength fallback below. On a
+  // croppable (windowed-state) model that fallback is unreachable (a==N takes the bonus branch,
+  // a<N takes the crop branch), so the snapshot is dead overhead -- skip it.
+  const bool crop_commit = main_->CanCropRecurrentState();
```
baijumeswani: The update changes `crop_commit` from an opt-in to `main_->CanCropRecurrentState()`. Every partial accept on a windowed model now takes `verify_argmax_[a]` from an early row of a wide forward, despite the code explicitly stating that this is not bit-identical to sequential decode. This contradicts the public C/C++ contract that output is “identical” and “lossless.” Either preserve a lossless/default replay path, or explicitly redefine the API contract and add regression evidence covering rejection and near-tie cases.

The following 

```
else if (main_->CanCropRecurrentState() && a >= 1)
```

is now unreachable, while the design document still presents it as an available finalize branch.

src/models/recurrent_state.cpp
```diff
@@ -136,6 +168,73 @@ void RecurrentState::Add() {
...
+        std::to_string(state_window_) + " for a forward of length " + std::to_string(forward_length_) +
+        "; rebuild the model with a larger recurrent_state_window");
+  if (signed_slot >= state_window_)
+    throw std::runtime_error(
+        "RecurrentState::CropToPosition(" + std::to_string(position) + ") is past the last position of a forward of length " +
+        std::to_string(forward_length_));
+  const size_t slot = static_cast<size_t>(signed_slot);
+  if (slot + 1 == static_cast<size_t>(state_window_)) return;  // Already the committed slot.
+
+  auto& device = *model_.p_device_;
```
baijumeswani: Should this be `p_device_kv_cache_`?

src/config.h
```diff
@@ -439,6 +445,40 @@ struct Config {
 
     } decoder;
 
+    // Multi-token-prediction (MTP) self-speculative head (e.g. Qwen3.6). Optional:
+    // when ``filename`` is empty the model has no MTP head and runs plain autoregressive
+    // decode. The head is a single extra decoder layer that, given the main model's last
+    // hidden state and the just-emitted token, predicts the next-next token used to draft
+    // a speculative token (verify-and-accept against the main model).
+    struct Mtp {
```
baijumeswani, line 451: Is this wired in? Does it get used anywhere?

src/mtp_generator.cpp
```diff
@@ -0,0 +1,1065 @@
...
+  if (main_hidden_type != mtp_hidden_type) {
+    throw std::runtime_error("MtpGenerator requires matching hidden-state tensor types");
+  }
+  // Number of speculative draft tokens per step (N), shared with draft-model speculative decoding
+  // via speculative.max_draft_tokens. N=1 is the original single-token fast path; N>1 chains the
+  // single MTP module N times (feeding its own post-norm hidden back), as vLLM's
+  // AutoRegressiveSpeculator does. Chaining needs the head to emit that hidden, so a head exported
+  // without `mtp_emit_hidden=true` can only run N=1.
+  num_speculative_tokens_ = std::max(1, params.speculative.max_draft_tokens);
+  if (num_speculative_tokens_ > 1 && !mtp_model_.session_info_.HasOutput(kHeadHiddenStatesOut)) {
```
baijumeswani: Silently reducing a requested `max_draft_tokens > 1` to 1 makes the effective behavior differ from `GeneratorParams`, and it still requires the literal `hidden_states_out` name. Please make the feedback output configurable and validate it during construction; if `N>1` cannot be honored, return a clear error or expose the effective value rather than silently changing it.

src/generators.cpp
tianleiwu: I considered splitting this out, but the version handshake needs to land atomically with the `DeviceInterface` vtable additions in this PR (`ArgMax`, `TopKScores`, `ArgMaxDevice`, and `CopyStateSlots`). Without the check, a core library can load an older CUDA add-on successfully and then dispatch through incompatible virtual slots, which is undefined behavior. I kept the handshake in this PR for that ABI safety reason.

src/mtp_generator.cpp
tianleiwu: Fixed in `1105b242`. `MtpGenerator::AppendTokens` now marks the wrapper done when the prompt already reaches `max_length`, before priming or queuing another token. Prompt EOS remains consistent with the ordinary `Generator`: user-appended prompt tokens reset search done state, since chat prompts commonly contain EOS delimiters.

src/mtp_generator.cpp
tianleiwu: Fixed in `1105b242`. I removed the always-on direct arena commit, restored the per-round recurrent snapshot, and made partial accepts use the existing crop + `M=1` replay or snapshot rewind + replay paths. The unreachable branch and stale design-table entry are gone, so the default path no longer consumes an early wide-forward argmax as the bonus token.

src/models/recurrent_state.cpp
tianleiwu: Fixed in `1105b242`. Both `CropToPosition` and `TryBatchedSlotPromote` now use `p_device_kvcache_`, matching the allocation and other recurrent-state operations.

src/config.h
tianleiwu: Wired in with `1105b242`. `model.mtp.main_hidden_states` now selects and validates the main-model handoff output, while `model.mtp.outputs.hidden_states` selects the MTP feedback output used for chained drafts. The latter is parsed from JSON, documented, and covered by `MtpConfigTest`. The caller still loads the MTP head as a separate `Model`, so the struct comment now states that ownership explicitly.

src/mtp_generator.cpp
tianleiwu: Fixed in `1105b242`. The feedback output is now configurable through `model.mtp.outputs.hidden_states` (default `hidden_states_out`) and validated during construction. A request for `max_draft_tokens > 1` now fails with a clear error when that output is missing instead of silently reducing the effective draft count to 1.

baijumeswani (approved):

docs/speculative_decoding.md
```diff
@@ -0,0 +1,468 @@
+# Speculative Decoding in ONNX Runtime GenAI
```
kunal-vaishnavi, line 1: Can we merge the two [docs PRs](https://github.com/microsoft/onnxruntime-genai/pull/2414) so that we have all of the info?

src/python/python.cpp
```diff
@@ -545,12 +591,22 @@ PYBIND11_MODULE(onnxruntime_genai, m) {
...
       .def("rewind_to", &PyGenerator::RewindTo)
+      .def("snapshot_state", &PyGenerator::SnapshotState)
+      .def("set_hidden_states", &PyGenerator::SetHiddenStates)
       .def("get_next_tokens", &PyGenerator::GetNextTokens)
       .def("get_sequence", &PyGenerator::GetSequence)
       .def("set_active_adapter", &PyGenerator::SetActiveAdapter)
       .def("set_runtime_option", &PyGenerator::SetRuntimeOption)
       .def("get_speculative_stats", &PyGenerator::GetSpeculativeStats);
 
+  pybind11::class_<PyMtpGenerator>(m, "MtpGenerator")
```
kunal-vaishnavi, line 602: Why do we need a new `MtpGenerator` object instead of the existing `Generator` object?

src/cuda/interface.cpp
```diff
@@ -419,6 +422,133 @@ struct CudaInterfaceImplBase : DeviceInterface {
     return true;
   }
 
+  bool RunArgMax(const void* logits, ONNXTensorElementDataType logits_type, int num_rows, int vocab_size) {
```
kunal-vaishnavi, line 425: Why does argmax need to be reimplemented? Do we not already have argmax calculated on CUDA?

src/models/hidden_states_inputs.cpp
```diff
@@ -0,0 +1,137 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#include "../generators.h"
+#include "model.h"
+#include "hidden_states_inputs.h"
+
+namespace Generators {
+
+HiddenStatesInputs::HiddenStatesInputs(State& state)
```
kunal-vaishnavi, line 10: Can we rename these files to `hidden_states.cpp` and `hidden_states.h` since it deals with hidden states as both inputs and outputs?

docs/speculative_decoding.md
tianleiwu: Yes

src/python/python.cpp
tianleiwu: MtpGenerator has a lot of changes upon Generator, and it is still experimental. This is safe way to land MTP support.

We can refactor them later once MTP is mature.

src/cuda/interface.cpp
tianleiwu: This is a generic interface. I think the one in search_cuda does not expose the interface.
We can refactor them later.

src/models/hidden_states_inputs.cpp
tianleiwu: Good suggestion. I'll do that in next PR.

PR #2353, opened 2026-07-31
Qwen3.6: build the MTP self-speculative head

tianleiwu:
## Summary

Adds Qwen3.6 MoE multi-token-prediction (MTP) head export for the self-speculative decoding runtime merged in #2352. With `--extra_options enable_mtp=true include_hidden_states=true`, the builder writes `mtp.onnx` alongside the main decoder and adds the corresponding `model.mtp` contract to `genai_config.json`.

This completes the builder side of Qwen3.6 MTP while preserving native ModelOpt NVFP4/FP8 tensors by default and allowing the MTP model to use an independent structured quantization configuration.

### MTP head export

- Adds `Qwen35MtpHead`, a single full-attention Qwen3.5 MoE decoder layer that predicts the next-next token from:
  - the main decoder's last hidden state; and
  - the embedding of the just-emitted token.
- Reads `mtp.*` tensors directly from source safetensors because Hugging Face `transformers` does not retain them in the loaded model.
- Exports `hidden_states_out` for recurrent multi-token drafting.
- Adds `model.mtp` filename, dimensions, inputs, outputs, and main hidden-state mapping to `genai_config.json`.
- Validates that MTP export keeps main-model hidden states and the LM head enabled.

### Independent MTP quantization configuration

- By default, the MTP model inherits the main model's settings.
- ModelOpt checkpoints preserve each MTP tensor's native format by default:
  - NVFP4 linears and routed experts remain NVFP4;
  - FP8 projections remain FP8; and
  - unquantized tensors use the graph precision.
- Adds `mtp_quant_config`, which accepts inline JSON or a JSON file using the structured `QuantConfig` schema.
- Resolves MTP I/O dtype, dense MatMul weights, MoE experts, runtime settings, and mixed-precision overrides independently from the main model.
- An explicit MTP configuration dequantizes native ModelOpt MTP tensors before applying the requested formats.
- Supports unquantized or INT4/INT8 dense MTP weights and independently configured INT4/INT8/MXFP4/NVFP4 MoE experts.
- Supports independent calibrated MTP KV-cache scales through the optional `mtp` section of `kv_cache_scale_file`.

### External-data deduplication

- Shares byte-identical embedding and `lm_head.MatMul.*` initializers between the main model and MTP graph.
- Repacks `mtp.onnx.data` after redirected tensors are removed.
- Stages both ONNX metadata and external data before replacing either file, and detects truncated input data so a failed optimization leaves the original export intact.

### Windowed recurrent state

- Adds `state_window=W` to widen Qwen3.6 convolution and recurrent state I/O to `[W, B, ...]`.
- Emits the matching `state_window` attribute on `CausalConvWithState` and `LinearAttention`.
- Allows a multi-token verification forward to crop state to the accepted prefix instead of replaying the decoder.
- Requires `W >= num_speculative_tokens + 1` for MTP verification. The default `0` retains legacy unwindowed I/O.

### Builder robustness

- Guards cleanup of the cache directory shared by the main and MTP exports.
- Validates ModelOpt tensor metadata and explicit quantization selections.
- Resolves Qwen3.5 dense, MoE, and MTP runtime model types during base initialization rather than overwriting inherited attributes.
- Documents the supported MTP export options and windowed-state behavior.

## Usage

```bash
python -m onnxruntime_genai.models.builder \
  -i <qwen3.6-model-directory> \
  -o <output-directory> \
  -p <precision> \
  -e <execution-provider> \
  --extra_options \
    enable_mtp=true \
    include_hidden_states=true \
    state_window=4
```

Optional independent MTP quantization configuration:

```bash
--extra_options \
  enable_mtp=true \
  include_hidden_states=true \
  mtp_quant_config='{"io_dtype":"bf16","weights":{"type":"int4","block_size":64},"moe":{"type":"nvfp4"}}'
```

To keep the complete MTP model in FP16, use:

```text
mtp_quant_config='{"io_dtype":"fp16","weights":{"type":"none"},"moe":{"type":"none"}}'
```

MTP export is currently supported for Qwen3.6 MoE checkpoints with architecture `Qwen3_5MoeForConditionalGeneration` and source safetensors containing `mtp.*` weights.

## Validation

- `PYTHONPATH=src/python/py python -m pytest -q test/python/builder/test_mixed_precision_config.py test/python/builder/test_quant_config.py test/python/builder/test_qwen_mtp_head_quant.py test/python/builder/test_qwen_mtp_export.py test/python/builder/test_precision.py`
  - `124 passed`
- `lintrunner` passed for the touched Python builder and test files.
- `git diff --check` passed.
- Qwen3.6-35B-A3B ModelOpt NVFP4 MTP export completed end to end before the configuration surface was generalized.
- Full PR CI is in progress.

## Related PRs

- #2350: Model builder NVFP4 quantization core, merged.
- #2351: Qwen3.6 native ModelOpt NVFP4/FP8 export, merged.
- #2352: MTP self-speculative decoding runtime, merged.
- Supersedes the closed #2218.


src/python/py/models/builders/qwen.py
```diff
@@ -2385,6 +2468,119 @@ def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
...
+        if self.enable_mtp:
+            # Stash the constructor arguments so the MTP head can be built from a
+            # pristine config after the main model has been generated.
+            self._mtp_config = copy.deepcopy(config)
+            self._mtp_io_dtype = io_dtype
+            self._mtp_onnx_dtype = onnx_dtype
+            self._mtp_ep = ep
+            self._mtp_cache_dir = cache_dir
+            self._mtp_extra_options = copy.deepcopy(extra_options)
+            self._resolve_mtp_head_quantization(extra_options)
```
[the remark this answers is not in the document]
tianleiwu: Fixed in f21dfc4d6d. MTP booleans are now normalized centrally and parsed safely in the Qwen constructor, and enable_mtp enforces include_hidden_states=true with exclude_lm_head disabled. Focused tests cover the false-string and prerequisite cases.

src/python/py/models/README.md
```diff
@@ -365,6 +366,30 @@ The relevant extra options are:
...
+* `mtp_head_quant_type=int4/int8/mxfp4/nvfp4` — quantize the whole head with the given scheme (same style as `moe_quant_type`): the routed experts become a `QMoE` op **and** the head's dense MatMuls (`mtp.fc`, the attention q/k/v/o projections, the shared expert and the draft lm_head) become `MatMulNBits` at the matching bit width. `int4` -> 4-bit dense + INT4 QMoE, `int8` -> 8-bit dense + INT8 QMoE. `mxfp4`/`nvfp4` are microscaling FP4 schemes that only exist for `QMoE`, so with those the dense MatMuls stay int4 and only the experts use FP4. For example, `mtp_head_quant_type=int8` is ~2.6x smaller on disk than the fp16 head with comparable acceptance and throughput, so it is preferred when GPU memory matters. (The legacy boolean `mtp_head_int8=true` is deprecated; use `mtp_head_quant_type=int8`.)
+* `mtp_head_int4_lmhead=true` — keep the head's draft lm_head at int4 even when the rest of the head is int8. The draft logits are corrected by speculative rejection sampling, so this cannot change output accuracy — only draft acceptance.
+
+`mtp_head_fp16` and `mtp_head_quant_type` are mutually exclusive. Both require the main model to be quantized (`-p int4`).
```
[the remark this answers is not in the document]
tianleiwu: Fixed in f21dfc4d6d. The README now matches the implementation: MTP head precision can override a floating-point main model, with FP4 schemes using int4 for the dense MatMuls and FP4 only for QMoE experts.

src/python/py/models/builders/qwen.py
```diff
@@ -2385,6 +2468,119 @@ def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
...
+
+    def _resolve_mtp_head_quantization(self, extra_options):
+        """Resolve the MTP head's weight precision from ``mtp_head_fp16`` / ``mtp_head_quant_type``.
+
+        Updates ``self._mtp_onnx_dtype`` / ``self._mtp_io_dtype`` / ``self._mtp_extra_options``
+        in place. Split out of ``__init__`` so it can be exercised directly by unit tests.
+        """
```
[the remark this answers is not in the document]
tianleiwu: Fixed in f21dfc4d6d. The builder CLI help now documents enable_mtp, all MTP head precision flags, the deprecated int8 alias, recurrent_state_window, and mtp_emit_hidden, including defaults and constraints.

src/python/py/models/builders/qwen.py
```diff
@@ -2633,6 +2830,150 @@ def make_matmul_op(self, matmul, basename, root_input, **kwargs):
...
+
+        def sampled_equal(path_a, off_a, path_b, off_b, length, chunks=8, chunk_size=1 << 20):
+            with open(path_a, "rb") as fa, open(path_b, "rb") as fb:
+                for i in range(chunks):
+                    off = (length // chunks) * i
+                    n = min(chunk_size, length - off)
+                    fa.seek(off_a + off); fb.seek(off_b + off)
+                    if fa.read(n) != fb.read(n):
+                        return False
+            return True
```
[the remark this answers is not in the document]
tianleiwu: Fixed in f21dfc4d6d. Deduplication now performs a deterministic full byte comparison in bounded-memory chunks and verifies each expected chunk length before redirecting external data.

src/python/py/models/builder.py
```diff
@@ -179,6 +180,27 @@ def check_extra_options(
...
+        except (TypeError, ValueError) as e:
+            raise ValueError("state_window must be a non-negative integer.") from e
+        if state_window < 0:
+            raise ValueError("state_window must be a non-negative integer.")
+        extra_options["state_window"] = state_window
+
+    if extra_options.get("enable_mtp", False):
+        if not extra_options.get("include_hidden_states", False):
+            raise ValueError("enable_mtp requires include_hidden_states=true on the main model.")
+        if extra_options.get("exclude_lm_head", False):
```
baijumeswani: should `prune_lm_head=true` be also rejected?

src/python/py/models/builders/qwen.py
```diff
@@ -2299,6 +2371,248 @@ def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
...
+        if self.mtp_head is not None:
+            self.mtp_head.save_model(out_dir)
+            # Deduplicate the embedding + lm_head weights, which the MTP head shares
+            # bit-identically with the main model: redirect mtp.onnx's copies to the
+            # main model's external data file and pack them out of mtp.onnx.data
+            # (~2 GB on disk; the two sessions then mmap the same bytes on the host).
+            self._share_mtp_embedding_lm_head(out_dir)
+
+    @staticmethod
+    def _share_mtp_embedding_lm_head(out_dir, main_file="model.onnx", mtp_file="mtp.onnx"):
```
baijumeswani: can we explicitly request callers to pass in `main_file` without defining a default. Callers can pass in `self.filename`?

src/python/py/models/builders/qwen.py
```diff
@@ -2299,6 +2371,248 @@ def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
...
+                f"Warning: could not share MTP embedding/lm_head weights ({exc}); "
+                f"the duplicated copies remain in {mtp_data_name}."
+            )
+            return
+
+        # Commit only fully staged files. A replacement failure is intentionally
+        # fatal: suppressing it could report success for an inconsistent pair.
+        os.replace(tmp_data, mtp_data)
+        os.replace(tmp_onnx, mtp_onnx)
+        saved_mb = sum(ln for _, ln in redirect.values()) / 1e6
```
baijumeswani: from ai:

> This two-file update is not transactional. If replacing `tmp_data` succeeds but replacing `tmp_onnx` fails, the original ONNX metadata references offsets in the newly repacked data file and the export is corrupted. Please retain backups and restore both originals if either replacement fails.

src/python/py/models/builder.py
tianleiwu: Good catch. enable_mtp now rejects prune_lm_head=true alongside exclude_lm_head=true, with documentation and regression coverage.”

src/python/py/models/builders/qwen.py
tianleiwu: Updated the helper to require main_file explicitly; the production caller now passes self.filename.

src/python/py/models/builders/qwen.py
tianleiwu: Good catch. The commit step now retains backups, restores both original files if either replacement fails, and cleans staged files. An injected second-replacement failure test verifies the rollback.

baijumeswani (approved):

src/python/py/models/builders/base.py
```diff
@@ -37,6 +37,9 @@
 
 
 class Model:
+    def _get_model_type(self, config):
```
kunal-vaishnavi, line 40: Why do we need this wrapper?

src/python/py/models/builders/base.py
```diff
@@ -325,18 +328,21 @@ def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
         # (weights / moe / runtime), then source every quantization knob below from it so there is a
         # single source of truth. `from_extra_options` mirrors the legacy desugaring exactly, so the
         # exported models remain byte-identical to the flat-option path.
-        self.quant_config = QuantConfig.from_extra_options(
-            extra_options,
-            precision=self.onnx_dtype_to_precision(self.onnx_dtype),
-            execution_provider=self.ep,
-        )
+        self.quant_config = extra_options.get("_quant_config")
```
kunal-vaishnavi, line 331: Why is there an underscore prefix?

src/python/py/models/builders/base.py
```diff
@@ -1327,8 +1340,9 @@ def callback(tensor: ir.TensorProtocol, metadata: dict):
                 callback=callback,
             )
 
-        # Delete temporary cache dir if empty
-        if not os.listdir(self.cache_dir):
+        # Delete temporary cache dir if empty. The MTP head shares the main model's
+        # cache dir and saves afterwards, so it may already be gone.
+        if os.path.isdir(self.cache_dir) and not os.listdir(self.cache_dir):
```
kunal-vaishnavi, line 1345: The value stored in `self.cache_dir` should already be a directory so the first predicate seems unnecessary.

src/python/py/models/builders/qwen.py
```diff
@@ -2299,6 +2371,272 @@ def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
...
+            # pristine config after the main model has been generated.
+            self._mtp_config = copy.deepcopy(config)
+            self._mtp_io_dtype = io_dtype
+            self._mtp_onnx_dtype = onnx_dtype
+            self._mtp_ep = ep
+            self._mtp_cache_dir = cache_dir
+            self._mtp_extra_options = copy.deepcopy(extra_options)
+            self._resolve_mtp_model_config(extra_options)
+
+    @staticmethod
```
kunal-vaishnavi, line 2402: Let's remove any `staticmethod` references.

src/python/py/models/builders/qwen.py
```diff
@@ -2553,3 +2891,413 @@ def make_shared_expert(self, layer_id, shared_expert, shared_expert_gate, root_i
...
+
+
+class _RMSNormWeight:
+    """Lightweight stand-in for an RMSNorm module exposing only ``weight``."""
+
+    def __init__(self, weight):
+        self.weight = weight
+
+
+class Qwen35MtpHead(Qwen35MoeTextModel):
```
kunal-vaishnavi, line 2916: Shall we move this class into its own `.py` file?

src/python/py/models/builders/base.py
tianleiwu: This is an intentional polymorphic hook. `Qwen35TextModel` and `Qwen35MoeTextModel` need different model type values, and the base constructor needs to select the value during initialization. This replaces the previous subclass attribute reassignment that CodeQL flagged as overwriting an attribute defined by a superclass, so I would keep the wrapper here.

src/python/py/models/builders/base.py
tianleiwu: The underscore marks this as an internal handoff of the already-parsed `QuantConfig` from the main builder to the MTP head builder. The public user-facing option is `mtp_quant_config`; `_quant_config` is deliberately not part of the supported CLI option surface.

src/python/py/models/builders/base.py
tianleiwu: `self.cache_dir` is intended to name a directory, but it may no longer exist at this point. The main model and MTP head share the cache directory and save sequentially; the first save can remove it when empty, so the second save needs the existence check to avoid `os.listdir` raising `FileNotFoundError`.

src/python/py/models/builders/qwen.py
tianleiwu: Agreed. I will handle this in a cleanup PR immediately after this feature PR: the helper will move with the MTP builder into its dedicated module, which removes the need for this `staticmethod`.

src/python/py/models/builders/qwen.py
tianleiwu: Yes. I will send a cleanup PR immediately after this feature PR to move `Qwen35MtpHead` and its private helper types/functions into a dedicated module. Keeping that mechanical reorganization separate will make this feature diff easier to validate.

PR #2354, opened 2026-07-31
Fix off-by-one past_sequence_lengths in the varlen decoder

tianleiwu:
### Description

The varlen decoder was telling PagedAttention that a decode step's KV started one position farther along than it actually did.

`past_sequence_lengths` is the position where the step starts writing. During prefill it was correctly `0`, but during decode it was set to `search->GetSequence(0).size()`, which already includes the token about to be processed. The attention metadata and paged-cache allocation used the same shifted convention, so the write index, read extent, and allocation remained internally consistent while leaving a hole in the cache:

- Prefill writes positions `0 .. L-1` (`past = 0`).
- The first decode step should write at `L`, but wrote at `L+1`, and later steps remained shifted.
- Position `L` was never written even though `kv_len` covered it on every decode step.

For a 9-token prompt, the decoder therefore reported `past_sequence_lengths` as `0, 10, 11, 12, ...` instead of `0, 9, 10, 11, ...`. The unwritten slot could contain data from a recycled block and was included in attention on every subsequent step. PagedAttention also derives RoPE positions from the same value, so generated tokens were positioned one step too far from the prompt.

This revision makes the existing `Request::processed_sequence_length_` cursor the source of truth for the number of tokens already represented in the KV cache:

- `Request::ProcessedSequenceLength()` exposes that cursor, initializes it to `0`, and `IsPrefill()` derives its result from whether the cursor is still zero.
- `VarlenDecoderIO` uses the cursor directly for `past_sequence_lengths`; its post-step `kv_len` remains the current sequence length.
- `StaticBatchDecoderIO` uses the same cursor as the base position for position IDs.
- `PagedKeyValueCache` reserves the current sequence length after a step and asserts that the slots already in use match the processed cursor before appending.

The current change is limited to five existing engine files. It does not add a shared sequence-position header or standalone arithmetic helpers.

### Motivation and Context

This is a correctness fix for the engine's PagedAttention path (#2330, #2333). The observable symptom is cross-request contamination: because the never-written slot holds whatever a recycled block last contained, a request's output can depend on what ran before it.

Running the same greedy request `X` (a) on its own and (b) immediately after an unrelated request `Y` used 48 generated tokens, gpt-oss-20b, INT8 paged KV, `block_size` 256, and CUDA graphs:

| | `X` alone | `X` after `Y` |
| --- | --- | --- |
| Before fix | `md5 4de28a624d4e` | `md5 400d532cb93b` |
| After fix | `md5 2aef32a73a74` | `md5 2aef32a73a74` |

Each case reproduced twice. Before the fix the two execution orders differed; after the fix they matched.

This is not a general determinism fix. Both builds were already reproducible across three runs in a fresh process (`160129c5c480` before and `5d4d6a701801` after). A cold process receives zeroed allocations, while block reuse inside a live process exposes the stale slot.

It is also not presented as a measurable benchmark-accuracy improvement. On an 800-question MMLU-Pro subset with the same model and ORT build, greedy decoding, and medium reasoning effort, the result was `573/800` before and `570/800` after. The reason to fix the issue is that the cache error is unbounded and non-local, not that this sample showed an aggregate accuracy gain.

### Context-Limit Safety

The shifted write did not go out of bounds at the context limit. The highest position written is `max_length - 1` with the bug and `max_length - 2` after the fix because the final generated token's KV is never consumed. That one step of slack absorbed the shift, and `max_length <= context_length` kept it below the reserved `ceil(context_length / block_size)` columns.

The bug did waste capacity: a sequence whose length lands just past a block boundary could reserve one extra block on every layer for its lifetime.

### Validation

- The cross-request block-reuse reproduction above produces matching output after the fix.
- A run with `context_length = 512` and `block_size = 256` reached the limit without `Block table exceeds the capacity reserved for graph capture.`
- Fresh-process reproducibility and the MMLU-Pro comparison were rechecked as described above.
- No new automated regression test or test source file is included in the current five-file revision.


src/engine/decoders/varlen_decoder_io.cpp
```diff
@@ -104,14 +105,11 @@ void VarlenDecoderIO::PrepareInputIds(std::shared_ptr<DecoderOnly_Model> model,
...
-      // But the kv cache is not updated until the first token is generated.
-      // So we set the past sequence length to current sequence length minus the unprocessed tokens length.
-      sequence_lengths_cpu_span[i] = static_cast<int32_t>(request->CurrentSequenceLength() - input_ids.size());
-    } else {
-      sequence_lengths_cpu_span[i] = static_cast<int32_t>(request->CurrentSequenceLength());
-    }
+    // The operator writes this step's token j of sequence i at absolute position
+    // past_sequence_lengths[i] + j, so this has to be the number of tokens already in the cache.
+    // Deriving it from the sequence length instead is off by one on a decode step, where the
+    // search has already appended the token that is about to be processed.
```
baijumeswani: While this comment is relevant, it only makes sense in the context of this diff. Any person reading this in the source code may question why the comment references the sequence length at all.

src/engine/sequence_positions.h
```diff
@@ -0,0 +1,31 @@
...
+//
+// PagedAttention writes token j of sequence i at absolute position `past_sequence_lengths[i] + j`,
+// so the decoder and the cache have to agree on what that base is. Both derive it from the number
+// of tokens whose keys and values are already in the cache -- never from the sequence length, which
+// is one longer on a decode step because the search has already appended the token that is about to
+// be processed.
+
+// Slots the cache must own once the pending step has run: the step writes at absolute positions
+// [processed_length, processed_length + scheduled_tokens).
+constexpr size_t SlotsAfterStep(int64_t processed_length, size_t scheduled_tokens) {
```
baijumeswani: These two helpers return the same value in every reachable state. Since `UnprocessedTokens()` always spans `[processed, current)`, `SlotsAfterStep(processed, unprocessed) == processed + (current − processed) == current == SlotsForWholeSequence(current)`. I compiled this header and enumerated every `(processed, current)` pair to confirm — they never diverge.

So the file introduces a distinction that doesn't exist, and the commit message asserts they "are no longer the same quantity." That's the opposite of the goal: the next person debugging this area will reason from a false premise.

Suggest deleting this header. If we want to anticipate chunked prefill (which is the case where they *would* genuinely diverge), say so in a comment at the single call site instead of encoding it as two functions.

src/engine/sequence_positions.h
baijumeswani, line 1: I do not think this new header is necessary. In every currently reachable request state:

```cpp
SlotsAfterStep(processed, unprocessed)
    == processed + (current - processed)
    == current
    == SlotsForWholeSequence(current)
```

Therefore, these helpers introduce two names for the same value. This also conflicts with the commit message’s statement that total and pending slots are now different quantities.

Could we delete this header and use `CurrentSequenceLength()` directly for slot requirements? If chunked prefill is expected to make these quantities differ later, that future behavior should be introduced when it exists rather than modeled prematurely.

src/engine/paged_key_value_cache.cpp
```diff
@@ -54,20 +56,21 @@ size_t UsedSlots(const std::vector<std::shared_ptr<Block>>& blocks) {
...
-// Counting appended tokens instead leaves the cache exactly one slot short on every decode step.
-// That stays invisible while the last block still has room, and turns into an out-of-bounds
-// block-table entry the moment a sequence length lands on a block boundary.
-size_t RequiredSlots(const std::shared_ptr<Request>& request) {
-  const size_t sequence_length = request->CurrentSequenceLength();
-  return request->IsPrefill() ? sequence_length : sequence_length + request->UnprocessedTokens().size();
+// This has to match how VarlenDecoderIO fills `past_sequence_lengths`: the decoder writes this
+// step's tokens at absolute positions [processed, processed + pending), so the cache must own
+// `processed + pending` slots.
+size_t PendingSlots(const std::shared_ptr<Request>& request) {
```
baijumeswani: These functions currently return the same value because `UnprocessedTokens()` is the range `[processed, current)`.

I suggest retaining one helper:

```cpp
size_t RequiredSlots(const std::shared_ptr<Request>& request) {
  return static_cast<size_t>(request-CurrentSequenceLength());
 }
```

This works for admission and append: once the current step completes, there is one KV slot for every token currently in the sequence. It also makes the invariant much easier to understand.

src/engine/paged_key_value_cache.cpp
```diff
@@ -160,7 +163,7 @@ void PagedKeyValueCache::AppendTokens(std::shared_ptr<Request> request) {
                                            });
   assert(block_table_it != block_tables_.end());
 
-  const size_t required_slots = RequiredSlots(request);
+  const size_t required_slots = PendingSlots(request);
   const size_t used_slots = UsedSlots(block_table_it->blocks);
```
baijumeswani, line 153: Do you think we should add a check or assertion that the cache and request cursors agree before appending?

```cpp
assert(UsedSlots(block_table_it->blocks) ==
       static_cast<size_t>(request-ProcessedSequenceLength()));
```

`BlockPool` tracks the exact number of used slots, not rounded block capacity, so this equality should hold on every step. It would have failed on the first decode step with the original bug and directly protects the decoder/cache contract this PR is fixing.

src/engine/decoders/varlen_decoder_io.cpp
```diff
@@ -164,11 +162,7 @@ void VarlenDecoderIO::PrepareAttentionMetadata(std::shared_ptr<DecoderOnly_Model
   } else {
     for (auto& request : scheduled_requests) {
       const int32_t query_len = static_cast<int32_t>(request->UnprocessedTokens().size());
-      // CurrentSequenceLength() already counts the unprocessed tokens for a prefill request, and
-      // excludes them for a generation request, mirroring how past_sequence_lengths is filled above.
-      const int32_t kv_len = request->IsPrefill()
-                                 ? static_cast<int32_t>(request->CurrentSequenceLength())
-                                 : static_cast<int32_t>(request->CurrentSequenceLength()) + query_len;
+      const int32_t kv_len = static_cast<int32_t>(SlotsAfterStep(request->ProcessedSequenceLength(), query_len));
```
baijumeswani: `SlotsAfterStep(processed, query_len)` reduces to `CurrentSequenceLength()` because `query_len` is the number of tokens between the processed and current cursors.

Could this be written directly?

```cpp
const int32_t kv_len =
    static_cast<int32_t>(request-CurrentSequenceLength());
```

A short comment explaining `kv_len = past + query = current` would be clearer than depending on the new arithmetic helper.

src/engine/request.h
```diff
@@ -118,6 +118,14 @@ struct Request : std::enable_shared_from_this<Request>,
    */
   int64_t CurrentSequenceLength() const;
 
+  /**
+   * @brief Number of leading tokens of the sequence whose keys and values are already in the cache.
+   *
+   * This is the absolute position the next scheduled token will be written at, which is what the
+   * decoder has to report to the model as the past sequence length.
+   */
+  int64_t ProcessedSequenceLength() const;
```
baijumeswani, line 152: `ProcessedSequenceLength()` is useful, but it also makes `is_prefill_` redundant. Prefill is exactly the state where no tokens have yet been processed:

```cpp
bool Request::IsPrefill() const {
  return processed_sequence_length_ == 0;
}
```

Deriving this value removes mutable state that must otherwise remain synchronized with the processed cursor.

src/engine/request.cpp
```diff
@@ -82,6 +82,10 @@ int64_t Request::CurrentSequenceLength() const {
   return search_->GetSequenceLength();
 }
 
+int64_t Request::ProcessedSequenceLength() const {
```
baijumeswani, line 115: Since `processed_sequence_length_` now means “tokens whose KV entries are already in the cache,” it should be initialized to zero when assigning the request:

```cpp
processed_sequence_length_ = 0;
```

`CurrentSequenceLength()` happens to be zero here today, but using zero directly makes the invariant explicit and avoids depending on the search’s initial state.

src/engine/request.h
baijumeswani, line 152: Not directly tied to your PR. Could we now document what the different length member variables and what they mean?

AI provided:
With `seen_sequence_length_`, `processed_sequence_length_`, and the search’s current length, it would help to document the cursor invariants together:

```text
seen <= current
processed <= current
UnprocessedTokens() == sequence[processed, current)
IsPrefill() == (processed == 0)
```

This would make the request lifecycle easier to reason about and reduce future off-by-one errors.

src/engine/decoders/varlen_decoder_io.cpp
baijumeswani, line 1: Could we also update the static batch decoder io, the existing position-ID calculation is correct only because non-prefill decode steps currently contain one token:

```cpp
request->IsPrefill() ? 1 : request->CurrentSequenceLength()
```

The universal base position is now available directly:

```cpp
const int64_t base = request->ProcessedSequenceLength();
position_id = base + j;
```

This is equivalent for current prefill and one-token decode behavior, but it remains correct for multi-token non-prefill steps such as chunked prefill or continuation.

test/sequence_positions_test.cpp
```diff
@@ -0,0 +1,110 @@
...
+
+#include "engine/sequence_positions.h"
+
+namespace Generators::test {
+namespace {
+
+// Mirrors the two cursors a Request keeps over its own sequence: the search's sequence length,
+// which grows as soon as a token is selected, and the processed length, which only grows once the
+// tokens have been through the model.
+struct SequenceCursor {
```
baijumeswani: These tests recreate request cursor behavior in a test-only `SequenceCursor`, so they do not exercise `Request`, `VarlenDecoderIO`, or `PagedKeyValueCache`. As a result, they would still pass if the original production bug were restored.

Could we build on top of the testing infrastructure in https://github.com/microsoft/onnxruntime-genai/pull/2362 so our tests are more meaningful?

baijumeswani (commented):
Thanks for the bug fix. The underlying fix is correct: decode must report the number of tokens already processed into the KV cache, not the current search sequence length, which already includes the next selected token.

My main concern is that the implementation adds more abstraction than the invariant requires, while the new tests do not exercise the affected production code. `TotalSlots()` and `PendingSlots()` are currently identical because `processed + unprocessed == current`, so `sequence_positions.h` can be removed and slot accounting can continue using a single `RequiredSlots(CurrentSequenceLength())` helper.

I recommend keeping `ProcessedSequenceLength()`, deriving `IsPrefill()` from it, using it for both varlen and static-batch position bases, and adding a runtime/cache test that would fail under the original bug. This should make the PR smaller, clearer, and better protected against future cursor drift.

src/engine/decoders/varlen_decoder_io.cpp
tianleiwu: Addressed in d6bfa752. `kv_len` now uses `CurrentSequenceLength()` directly, with a comment documenting that the post-step KV length is past length plus query length, which equals the current sequence length.

kunal-vaishnavi (approved):
Approving this PR. From Copilot:

Could you update the PR description to reflect the current implementation?

It currently says this revision adds `src/engine/sequence_positions.h`, the `SlotsForWholeSequence`/`SlotsAfterStep` helpers, and `test/sequence_positions_test.cpp`, but those files and helpers are not present in the current five-file diff. It looks like the implementation was simplified after review, so the description should be brought in sync with the actual changes and tests before merge. This will avoid documenting behavior or regression coverage that does not exist in the submitted revision.

tianleiwu:
Thank @kunal-vaishnavi, PR description updated.

PR #2355, opened 2026-07-31
Add chunked prefill to the engine

tianleiwu:
### Description

> Stacked on #2354. Review the top commit only; the base PR is the first commit in the diff.

A prompt currently has to be prefilled in a single step, so a long prompt produces one enormous forward pass: a latency spike for whoever is decoding alongside it, a batch that has to be sized for the worst-case prompt, and a KV allocation that lands all at once.

This lets the scheduler hand a request its prompt a chunk at a time. `Config::Search::chunk_size` already exists and is already parsed; this is what makes it do something.

- `Request::ScheduleTokens(allow_chunking)` decides once per step how many unprocessed tokens the request will offer, and `UnprocessedTokens()` returns exactly that many. Everything downstream — the cache, the decoder inputs, the position arithmetic — already works off `UnprocessedTokens()`, so nothing else needs to know that chunking exists.
- `IsChunkComplete()` says whether the step finished the prompt. `ScheduledRequests::GenerateNextTokens()` samples a token only for the requests where it did; the rest just advance to the next chunk. That is the one place the distinction matters.
- The decision itself is `ScheduledTokenCount(unprocessed, chunk_size, allow_chunking)` in `src/engine/sequence_positions.h`, next to the position arithmetic it has to stay consistent with.

`DynamicBatchScheduler` enables it; `StaticBatchScheduler` does not, since it has no other requests to protect. With no `chunk_size` configured the behaviour is unchanged.

### Motivation and Context

Chunked prefill is the usual answer to prefill/decode interference in a continuous-batching engine. It is also a hard prerequisite for the paged sliding-window ring cache, which can only bound its memory if a step's token count is bounded.

### Tests

`test/prefill_chunking_test.cpp` drives a simulated request through the same arithmetic the engine uses and checks the resulting `(past, query_len)` sequence:

- a 512-token prompt at `chunk_size` 128 produces `{0,128} {128,128} {256,128} {384,128} {512,1} {513,1}` — four prefill steps, then decode,
- a short final chunk, a chunk larger than the prompt, and `chunk_size` 1 all behave,
- chunking off or `chunk_size` unset still prefills in one step,
- every position is written exactly once across the whole run, and the cache is asked for exactly the scheduled tokens at each step, never the whole prompt.

```
8 tests from PrefillChunkingTest — PASSED
4 tests from SequencePositionsTest — PASSED
```

### DIFF
--- a/src/engine/request.cpp
+++ b/src/engine/request.cpp
@@ -4,6 +4,7 @@
 #include "request.h"
 
 #include "engine.h"
+#include "sequence_positions.h"
 #include "../search.h"
 
 namespace Generators {
@@ -120,6 +121,24 @@ int64_t Request::ProcessedSequenceLength() const {
   return processed_sequence_length_;
 }
 
+size_t Request::ScheduledTokenCount() const {
+  const size_t unprocessed = static_cast<size_t>(CurrentSequenceLength() - processed_sequence_length_);
+  return std::min(scheduled_token_count_, unprocessed);
+}
+
+void Request::ScheduleTokens(bool allow_chunking) {
+  const size_t unprocessed = static_cast<size_t>(CurrentSequenceLength() - processed_sequence_length_);
+  scheduled_token_count_ = Generators::ScheduledTokenCount(unprocessed, params_->search.chunk_size, allow_chunking);
+}
+
+bool Request::IsChunkComplete() const {
+  return processed_sequence_length_ + static_cast<int64_t>(ScheduledTokenCount()) >= CurrentSequenceLength();
+}
+
+void Request::AdvanceChunk() {
+  processed_sequence_length_ += static_cast<int64_t>(ScheduledTokenCount());
+}
+
 int32_t Request::UnseenToken() {
   if (static_cast<size_t>(seen_sequence_length_) >= tokens_host_.size())
     throw std::runtime_error("All tokens have been seen.");
@@ -133,13 +152,12 @@ bool Request::HasUnseenTokens() const {
 
 DeviceSpan<int32_t> Request::UnprocessedTokens() {
   auto sequence = search_->GetSequence(0);
-  auto unprocessed_tokens = sequence.subspan(processed_sequence_length_, CurrentSequenceLength() - processed_sequence_length_);
-  return unprocessed_tokens;
+  return sequence.subspan(processed_sequence_length_, ScheduledTokenCount());
 }
 
 std::span<const int32_t> Request::UnprocessedTokensCpu() const {
   const size_t begin = static_cast<size_t>(processed_sequence_length_);
-  const size_t end = static_cast<size_t>(CurrentSequenceLength());
+  const size_t end = begin + ScheduledTokenCount();
   if (end > tokens_host_.size())
     throw std::runtime_error("The host token mirror is out of sync with the search sequence.");
 
@@ -240,7 +258,7 @@ void Request::CommitStep(const RequestStepPlan& plan,
   if (result.token_appended) {
     tokens_host_.push_back(result.token);
   }
-  processed_sequence_length_ = plan.sequence_length_before;
+  processed_sequence_length_ = static_cast<int64_t>(plan.target_cache_slots);
   status_ = result.done ? RequestStatus::Completed : RequestStatus::InProgress;
 }
 
--- a/src/engine/request.cpp
+++ b/src/engine/request.cpp
@@ -4,6 +4,7 @@
 #include "request.h"
 
 #include "engine.h"
+#include "sequence_positions.h"
 #include "../search.h"
 
 namespace Generators {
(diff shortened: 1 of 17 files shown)


src/engine/scheduled_requests.h
```diff
@@ -32,10 +32,14 @@ struct BatchedSamplingPlan {
...
 struct ScheduledRequests {
+  // `allow_chunked_prefill` lets a request spread a prompt longer than its `search.chunk_size` over
+  // several steps. Only the paged cache can serve that, since it is the only one that can hold a
+  // partly written sequence.
   ScheduledRequests(std::vector<std::shared_ptr<Request>> requests,
                     std::shared_ptr<Model> model,
                     BatchedSampler* batched_sampler,
-                    BatchedSamplingPlan* sampling_plan);
+                    BatchedSamplingPlan* sampling_plan,
+                    bool allow_chunked_prefill = false);
```
baijumeswani: Could we assume chunked prefill is always enabled and avoid the input argument?

src/engine/scheduler.cpp
```diff
@@ -184,18 +185,19 @@ StepPlanningResult DynamicBatchScheduler::PlanStep(StepPlan& plan) {
...
 
     RequestStepPlan entry;
     entry.request = request;
     entry.request_id = request.get();
     entry.sequence_length_before = snapshot.current_sequence_length;
-    entry.unprocessed_token_count =
-        static_cast<size_t>(unprocessed_token_count);
+    entry.unprocessed_token_count = ScheduledTokenCount(
+        static_cast<size_t>(remaining_token_count), request->SearchOptions().chunk_size,
+        /*allow_chunking=*/true);
```
baijumeswani: This change would make it such that a request might become serviceable based on the chunk size during prefill. Should/could we keep the admission criteria based on the whole prompt serviceability?

src/engine/request.cpp
```diff
@@ -120,6 +121,24 @@ int64_t Request::ProcessedSequenceLength() const {
...
+void Request::ScheduleTokens(bool allow_chunking) {
+  const size_t unprocessed = static_cast<size_t>(CurrentSequenceLength() - processed_sequence_length_);
+  scheduled_token_count_ = Generators::ScheduledTokenCount(unprocessed, params_->search.chunk_size, allow_chunking);
+}
+
+bool Request::IsChunkComplete() const {
+  return processed_sequence_length_ + static_cast<int64_t>(ScheduledTokenCount()) >= CurrentSequenceLength();
+}
+
+void Request::AdvanceChunk() {
```
baijumeswani, line 140: after advance chunk, the isprefill() becomes false even though we are still in prefill phase.

src/engine/scheduled_requests.h
tianleiwu: Done — the flag is gone. `Request::ScheduleTokens()` and the `ScheduledRequests` constructor no longer take it, and `ScheduledTokenCount(unprocessed, chunk_size)` is now gated only by `search.chunk_size`.

The one caller that relied on the `false` default was the static batch path: `StaticBatchDecoderIO` rebuilds its contiguous cache from the whole sequence on every step (the attention mask is derived from `CurrentSequenceLength()`), so it cannot resume a half written prompt. Rather than keep a parameter for that, `StaticBatchScheduler::AddRequest` now rejects `search.chunk_size` outright, so the combination fails loudly instead of silently producing a wrong cache.

Fixed in 69faf156.

src/engine/scheduler.cpp
tianleiwu: Good catch — you're right, and it was worse than just a policy change: `PagedKeyValueCache::Add()` reserves blocks for the *whole* sequence, so a request green-lit on its first chunk could then fail `CanAdd()` in `Allocate()`, or be admitted into a pool that can never hold the rest of its prompt.

Admission is now decided on the whole prompt again. `RequestStepPlan` carries `whole_sequence_cache_slots` (from the `SlotsForWholeSequence` helper, which had the right comment but no caller), and `PlanStepResources` uses it for both the `permanently_unserviceable` test and the free-block check on newly admitted requests. `target_cache_slots` still carries the chunk, so the chunk continues to size the step and the cache growth — it is only the admission decision that looks at the whole sequence.

Added two tests in `paged_key_value_cache_tests.cpp`: a prompt too large for the pool is unserviceable even when its first chunk fits, and admission defers until enough free blocks exist for the whole prompt.

Fixed in 69faf156.

src/engine/request.cpp
tianleiwu: Right — `processed_sequence_length_ == 0` only held for the first chunk. That mattered: `IsPureDecodeStep` and `StepPlan::graph_capture_eligible` both use `is_prefill` to keep prefill steps out of graph capture, so a trailing one token prefill chunk (e.g. a 5 token prompt at `chunk_size=2`) would have been captured as if it were a decode step.

`Request` now tracks `prompt_sequence_length_` — the sequence length the application's tokens reach up to, set in `Assign()` and refreshed by `AddTokens()` on a continued request — and `IsPrefill()` is `processed_sequence_length_ < prompt_sequence_length_`. It stays true across every chunk and flips to false only once the whole prompt is in the cache.

Updated `RequestLifecycleTest.PartialPrefillAdvancesOnlyAtCommit` accordingly: after committing 2 of the 3 prompt tokens the snapshot now reports `is_prefill == true`.

Fixed in 69faf156.

baijumeswani (dismissed):

baijumeswani (dismissed):

src/engine/scheduler.cpp
tianleiwu: Follow-up: the first fix was incomplete. Admission was approved on `whole_sequence_cache_slots`, but `PagedCacheStepReservation` still built the reservation from `target_cache_slots`, so a newly admitted chunked prefill only held blocks for its first chunk. The rest of its capacity could be admitted to another request on the next step, and once every resident request needs growth with no free blocks left there is no way to make progress.

`PagedCacheReservationRequest` now carries `reserved_slots` next to `target_slots`: blocks are taken for `reserved_slots` (the whole sequence) while `AdvanceCommittedSlots` still moves `committed_slots` only to `target_slots`. `PlanStepResources` sizes its block budget the same way, so the planner and the reservation cannot disagree. Subsequent chunks find the capacity already owned and reserve nothing new.

New test `PagedCacheReservationTest.ChunkedPrefillHoldsWholePromptButCommitsOnlyTheChunk` covers both halves.

Fixed in 613db750.

baijumeswani (approved):

PR #2357, opened 2026-07-31
Serve sliding-window paged layers from a ring of blocks (runtime)

tianleiwu:
### Description

> Draft. Stacked on #2355 (chunked prefill), which is stacked on #2354. Review the top commit only.
> Needs the builder counterpart, #2358, to be useful; on its own it is inert (no model emits `block_table_windowed` yet), so it is safe to merge first — and it has to, because the config parser would otherwise reject models built by #2358.

A PagedAttention sliding-window layer never reads further back than `window_size` positions, yet the cache hands it one block per position exactly like a full-attention layer. Nearly all of that is allocated and never read, and it is taken out of the same budget the full-attention layers are competing for.

This gives those layers their own small pool and their own block table. Each request gets a **ring** of blocks, claimed in full at admission and never grown, and `block_table_windowed` repeats that ring across its columns: column `j` names ring block `j % ring_blocks`. The operator is unchanged — it indexes the table with the token's true position, so position `p` resolves to ring slot `p % (ring_blocks * block_size)` and positions that have fallen out of the window are overwritten in place.

```
ring = [5, 9], block_size = 4   ->   table = [5, 9, 5, 9, 5, 9, ...]
position 0..3  -> block 5     position 8..11 -> block 5 again
position 4..7  -> block 9     position 12..15 -> block 9 again
```

The ring has to span the whole live set of a step. A step covers `[past, past + scheduled)` and every one of those queries reaches back `window_size` positions, so the ring must hold `chunk_size + window_size - 1` positions. That makes chunked prefill a requirement rather than an option: the constructor rejects a model with no `search.chunk_size`, and `AppendTokens` rejects a step carrying more tokens than the ring was sized for.

Block accounting changes shape accordingly. The windowed layers' fixed cost comes off the top of the memory budget and what remains is divided among the layers that keep the whole sequence, which is where the gain shows up.

### Motivation and Context

On gpt-oss-20b at `block_size` 256 and window 128 the ring is 2 blocks. The full-attention layers go from **9910 to 19626 blocks** — 5.02M rather than 2.54M resident KV tokens, **1.98x**, for a 0.03% difference in bytes allocated. That is roughly twice the concurrent sequence length, or twice the batch, in the same memory.

The layout only engages for models built with a `block_table_windowed` input, so existing paged models are unaffected.

### Tests

`test/window_ring_test.cpp` covers the two things this has to get right.

Sizing: `WindowLiveSpan(256, 128) == 383`, `WindowRingBlocks(256, 128, 256) == 2`, exact fits do not round up to a spare block, and smaller chunks buy a smaller ring.

Addressing: the table the runtime fills and the indexing the operator performs must compose. The test asserts `WindowRingColumn(p / block_size) * block_size + p % block_size == p % (ring_blocks * block_size)` over 4000 positions.

The invariant the sizing exists to guarantee: for every prefill step and every decode step, all positions the step is allowed to read land in **distinct** ring slots. There is a companion test showing that a ring sized for `chunk_size` 64 does collide on a 256-token step, which is what the `AppendTokens` check rejects.

```
8 tests from WindowRingTest — PASSED
20 tests total (with #2354 and #2355) — PASSED
```

End-to-end, greedy output was compared token for token against a build with full-length KV on the sliding-window layers, on 6-, 754- and 3000-token prompts at chunk sizes 256, 128, 64 and 1. Identical in every case.


src/engine/paged_key_value_cache.cpp
```diff
@@ -16,29 +18,54 @@ ONNXTensorElementDataType KeyValueCacheType(std::shared_ptr<Model> model) {
...
+  if (!decoder.sliding_window.has_value() ||
+      decoder.inputs.block_table_windowed.empty() ||
+      !model->session_info_.HasInput(decoder.inputs.block_table_windowed)) {
+    return {};
+  }
+
+  std::vector<size_t> layers;
+  for (int layer : decoder.sliding_window->layers) {
+    layers.push_back(static_cast<size_t>(layer));
+  }
```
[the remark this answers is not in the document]
tianleiwu: Addressed in f24daa65e0. WindowedLayers now validates each signed layer index against the decoder layer range before converting it to size_t, and fails with a clear runtime error for invalid values.

src/engine/paged_key_value_cache.cpp
```diff
@@ -78,28 +105,63 @@ size_t PendingSlots(const std::shared_ptr<Request>& request) {
...
-                                                   static_cast<int64_t>(model->config_->model.decoder.head_size)};
+  const auto& decoder = model->config_->model.decoder;
+  const size_t block_size = model->config_->engine.dynamic_batching->block_size;
+  const size_t max_batch_size = model->config_->engine.dynamic_batching->max_batch_size;
+
+  const auto windowed_layers = WindowedLayers(model);
+  const std::set<size_t> windowed{windowed_layers.begin(), windowed_layers.end()};
+  size_t num_window_blocks = 0;
+  if (!windowed.empty()) {
+    window_size_ = static_cast<size_t>(decoder.sliding_window->window_size);
```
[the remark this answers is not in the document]
tianleiwu: Addressed in f24daa65e0. The windowed-ring path now rejects window_size <= 0 before the value is converted to size_t or used in ring arithmetic.

test/window_ring_test.cpp
```diff
@@ -0,0 +1,141 @@
...
+
+#include "engine/window_ring.h"
+
+using Generators::WindowLiveSpan;
+using Generators::WindowRingBlocks;
+using Generators::WindowRingColumn;
+using Generators::WindowRingSlot;
+
+// A sliding-window paged layer is served from a ring of blocks rather than one block per position.
+// These tests pin the two things that has to get right: the ring is big enough that a step never
```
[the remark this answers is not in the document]
tianleiwu: Fixed in f24daa65e0: the comment now reads "the two things this has to get right."

src/engine/paged_key_value_cache.cpp
baijumeswani, line 1: `Snapshot()` reports only global blocks, so leaks, duplicate ownership, malformed ring sizes, or rollback errors in the window pool are invisible to the existing invariant validator. Add per-group snapshots and validate both pools independently.

test/window_ring_test.cpp
baijumeswani, line 1: Tests cover reservation release, but not removal/reuse after commit, mixed existing/new requests, rollback after model-execution failure, or graph-captured state reuse. At minimum add a test proving `Remove()` returns committed window blocks and a later request reuses them; also test both pools after transaction rollback.

baijumeswani (commented):
Can you add the necessary updates to the paged attention doc to highlight what is explicitly supported through this PR?

src/engine/paged_key_value_cache.cpp
tianleiwu: Addressed in 33a4a750. The snapshot now includes independent window-pool total/free accounting, fixed per-request ring width, committed block IDs, and transaction-reserved block IDs. The invariant validator checks window ownership, range, duplicates, ring size, and balanced accounting separately from the global pool.

test/window_ring_test.cpp
tianleiwu: Addressed in 33a4a750 and 4775f062. Added mixed resident/new-request rollback coverage that verifies both global and window pools are restored, plus committed removal/reuse coverage. PagedKeyValueCache::Remove() now delegates to the same tested two-pool block-table removal operation, and a later request is verified to receive the released window block IDs.

baijumeswani:
The chunk size can be dynamic:

> The ring is sized from the model-level `search.chunk_size`, but dynamic scheduling uses the request-level chunk size. Since transactional execution bypasses `AppendTokens()`, its ring-size check is not reached. A request can override `chunk_size` with a larger value or 0 and schedule more tokens than the ring can safely hold, potentially overwriting live KV slots and producing incorrect output.

tianleiwu:
Thanks for raising the dynamic chunk concern. This is addressed in `374516a1`.

Dynamic scheduling now caps each request's query tokens to the window ring's model-level chunk capacity, including absent/zero and oversized request overrides, while preserving smaller positive overrides.

Two regression tests have been added. Additionally, `docs/paged_attention_engine.md` now documents activation requirements, ring sizing/addressing, request cap behavior, independent pool accounting, and graph capture.

Validation: 142 engine tests and 8 WindowRing tests passed.

baijumeswani (approved):

PR #2362, opened 2026-07-31
Add engine scheduler and request unit tests

baijumeswani:
## Summary

This PR builds on the continuous-batching engine testing foundation (#2347) by adding focused unit tests for the engine's scheduling, admission, request-lifecycle, and step-execution behavior. To make those behaviors testable without spinning up a real model on a real device, it also introduces a small number of seams in the production code that let tests substitute lightweight, observable stand-ins for the heavyweight collaborators.

The goal is to lock down the engine's observable contracts — what it calls, in what order, how many times, and how it transitions request state — so that future refactors (including the upcoming transactional-execution work) can be made with confidence.

## Production changes (test seams)

These are deliberately small, behavior-preserving changes whose purpose is to expose an interface the engine already depends on so a test can observe or script it.

- **ModelExecutor becomes an interface.** `ModelExecutor` is now an abstract execution boundary with a virtual `Decode` and a `Create` factory, and the real decoder-backed implementation moves to `DecoderModelExecutor`. Production still assembles the concrete executor via `ModelExecutor::Create`, but tests can now supply an alternative that returns scripted outcomes or injects an execution failure, exercising the engine's execution path without running an ONNX graph. Only `Decode` is part of the boundary, since that is what `Engine::Step` drives; encoding stays on the concrete executor until it becomes part of the engine execution path.
- **Admission arithmetic is extracted into a pure helper.** The KV-slot sizing math (`RequiredSlots`) moves into `src/engine/admission.h`, taking plain counters instead of a `Request`. This lets the prefill-versus-decode slot accounting be unit-tested directly, without a model, a device, or a Search, and documents the off-by-one hazard at block boundaries that motivated the branch. The paged cache manager now calls this helper with request-derived values.
- **Minor supporting adjustments** in `engine.cpp` / `engine.h` and `paged_key_value_cache.cpp` to route through the new seams.

## Test infrastructure

- **Recording / scripted test doubles** (`engine_test_doubles.h`) that stand in for the scheduler, cache, and model executor. They record the calls the engine makes so tests can assert on call order and counts, and they can be scripted to return specific outcomes.
- **Shared test helpers** (`engine_test_helpers.h`) for constructing requests and common fixtures.
- **A tiny dummy decoder fixture model** (`test/models/engine/dummy-decoder/`) plus a generator script (`create_dummy_engine_model.py`) that produces a minimal ONNX decoder and `genai_config.json`. This gives the engine something real enough to load and step through without depending on a large external model.

## Tests added

- **Scheduler contract tests** — assert the engine drives the scheduler through the expected sequence of interactions.
- **Engine::Step call-order tests** — verify the per-step ordering of scheduler, cache, and executor calls across the single-request, batched, backpressure/draining, and no-work cases.
- **Admission arithmetic tests** — pin down `RequiredSlots` across the prefill and decode branches, including the block-boundary edge case.
- **Request lifecycle transition tests** — verify a request moves through its states correctly across steps.


test/engine/engine_step_tests.cpp
```diff
@@ -0,0 +1,142 @@
...
+  for (const auto& request : requests) {
+    EXPECT_TRUE(request->IsDone());
+  }
+  ASSERT_EQ(engine.executor->decoded_batch_sizes.size(), 2u);
+  EXPECT_EQ(engine.executor->decoded_batch_sizes[0], 2u);
+  EXPECT_EQ(engine.executor->decoded_batch_sizes[1], 1u);
+}
+
+// With nothing queued, Step reports no work and returns cleanly.
+TEST_F(EngineStepTest, StepWithNoRequestsReturnsNull) {
```
tianleiwu, line 132: The PR description says these `Engine::Step` tests include behavior when execution reports a failure, but the suite currently covers only successful execution, batching/backpressure, draining, and the no-work case. Since the new `ModelExecutor` seam explicitly supports failure injection, could you either add a throwing-executor test that pins the intended exception/state contract, or remove the failure-coverage claim until the transactional-execution work defines that behavior?

tianleiwu (commented):
The production seams are narrowly scoped and behavior-preserving, and the dedicated engine suite builds and passes all 61 tests locally. I found one non-blocking mismatch between the PR's stated failure-path coverage and the tests currently present.

tianleiwu (approved):

test/engine/engine_step_tests.cpp
baijumeswani: Removed the failure coverage claim from the PR description. thanks for catching this.

PR #2363, opened 2026-07-31
Fix process-exit teardown crashes (telemetry deadlock + ORT env abort)

baijumeswani:
## Summary

A process using ONNX Runtime GenAI can hang or crash during static destruction. Foundry Local exposed two independent failures on macOS-arm64:

1. A deterministic deadlock in the vendored 1DS telemetry library.
2. An abort while destroying the last `OrtEnv`.

The underlying problems are static-destruction-order hazards and are not inherently macOS-specific.

## 1. 1DS telemetry teardown

`OfflineStorageHandler::Flush()` calls `StartActivity()` and later balances it with a raw `EndActivity()` call.

During process exit, `DebugEventSource::DispatchEvent()` can lock a function-local static mutex that has already been finalized. The resulting `std::system_error` escapes `Flush()`, skipping `EndActivity()` and permanently leaving the activity count at one.

The subsequent `FlushAndTeardown()` transitions to `Pausing` and waits indefinitely in `WaitPause()` for that activity count to reach zero.

This was confirmed from a hung process:

- `m_pause_active_count == 1`
- `m_pause_state == Pausing`
- the 1DS worker thread was idle
- the blocked thread was inside `WaitPause()`, not GenAI's telemetry mutex

### Fix

The vendored 1DS source is patched to:

- Balance `StartActivity()` with an RAII guard so exceptions cannot leak the activity count.
- Keep `LogManagerFactory` alive through static destruction so `LogManagerProvider::Release()` can safely locate and destroy the manager.
- Keep `PlatformAbstractionLayer` alive through static destruction so `PAL::shutdown()` operates on live members.

The manager is still genuinely torn down: queued events are flushed or persisted and worker threads are joined. GenAI's telemetry implementation remains unchanged.

The exception from the already-finalized `DebugEventSource` mutex is deliberately tolerated. Making that mutex immortal removed the exception but allowed dispatch to continue into already-destroyed listener containers, producing an intermittent use-after-free.

These patches should be removed after equivalent fixes are available in the pinned `cpp_client_telemetry` release.

## 2. ORT environment teardown

Destroying the final `OrtEnv` from inside `__cxa_finalize` drives ORT's `Environment`/`LoggingManager` destructor to lock a static mutex that may already have been finalized. The exception escapes a destructor and terminates the process.

### Fix

Only the `OrtEnv` is released from its `unique_ptr` during static destruction. `OrtGlobals` is still destroyed normally, preserving:

- graph-session cache cleanup
- device allocator cleanup
- device-interface teardown, including `RyzenAI_Shutdown`
- CUDA add-on unloading

The default GenAI `OrtEnv` does not own global thread pools, so leaking it does not leave threads executing during module unload.

`g_process_exiting` is also set during `dlclose`/`FreeLibrary`. In that case the conservative behavior leaks one `OrtEnv` per module load/unload cycle, but it does not leave background threads running in an unloaded module.

Explicit runtime `OgaShutdown()` remains unchanged and continues to perform complete teardown, including destroying the `OrtEnv` and supporting subsequent reinitialization.

### DIFF
--- a/cmake/patches/cpp_client_telemetry/apply_patch.cmake
+++ b/cmake/patches/cpp_client_telemetry/apply_patch.cmake
@@ -119,6 +119,11 @@ ortgenai_replace_required(
 # exits, versus 85/85 clean with this hunk alone). Tolerating the exception is the correct fix.
 # The guard's destructor locks m_pause_mutex, a member of the still-live heap-allocated
 # LogManagerImpl, so it cannot itself throw while unwinding.
+#
+# There is also a distinct rejected-activity path: a scheduled flush sets m_flushPending and resets
+# m_flushComplete before its task starts. If teardown pauses the LogManager first, StartActivity()
+# returns false; returning without signaling completion leaves WaitForFlush() blocked forever.
+# Cancel the scheduled handle and complete the pending flush under m_flushLock on that path.
 # Reported upstream; drop this hunk once the fix is in a pinned release.
 ortgenai_replace_required(
   "${SOURCE_DIR}/lib/offline/OfflineStorageHandler.cpp"
@@ -126,6 +131,12 @@ ortgenai_replace_required(
             return;
         }]=]
   [=[        if (!m_logManager.StartActivity()) {
+            // Teardown can pause the LogManager after a flush has been scheduled but before its task
+            // starts. Complete the rejected flush so WaitForFlush() cannot wait forever.
+            LOCKGUARD(m_flushLock);
+            m_flushHandle.Cancel();
+            m_flushComplete.post();
+            m_flushPending = false;
             return;
         }
         // Balance the activity count on every exit path, including an exception unwinding out of
--- a/cmake/patches/cpp_client_telemetry/apply_patch.cmake
+++ b/cmake/patches/cpp_client_telemetry/apply_patch.cmake
@@ -101,3 +101,76 @@ ortgenai_replace_required(
         curl_easy_setopt(curl, CURLOPT_HTTP_VERSION, CURL_HTTP_VERSION_2_0);]=]
   [=[        // The embedded Linux curl omits nghttp2 to keep the transport self-contained.
         curl_easy_setopt(curl, CURLOPT_HTTP_VERSION, CURL_HTTP_VERSION_1_1);]=])
+
+# OfflineStorageHandler::Flush() brackets its body with ILogManager::StartActivity() /
+# EndActivity() using raw calls rather than a scope guard, so any exception thrown between them
+# permanently leaks one pause-activity count. That is not hypothetical: during process exit,
+# OnStorageRecordsSaved() -> DebugEventSource::DispatchEvent() locks a std::recursive_mutex whose
+# backing static may already have been finalized, which throws std::system_error
+# ("mutex lock failed: Invalid argument"). The leaked count then makes the *next* FlushAndTeardown()
+# call PauseActivity() (state -> Pausing) and block forever in WaitPause(), which waits for the
+# activity count to drain to zero. That is a deterministic, unrecoverable process-exit deadlock.
+#
+# Make the activity strictly scope-bound so the count is balanced on every exit path, including
+# exceptional ones. Note this deliberately does not try to prevent the throw itself: the throwing
+# lock doubles as a circuit breaker that stops DispatchEvent() from walking listener/cascaded
+# containers that are themselves already destroyed. Making that mutex immortal was measured to
+# remove the throw but introduce an intermittent use-after-free crash (3 failures in 75 process
+# exits, versus 85/85 clean with this hunk alone). Tolerating the exception is the correct fix.
+# The guard's destructor locks m_pause_mutex, a member of the still-live heap-allocated
+# LogManagerImpl, so it cannot itself throw while unwinding.
+# Reported upstream; drop this hunk once the fix is in a pinned release.
+ortgenai_replace_required(
+  "${SOURCE_DIR}/lib/offline/OfflineStorageHandler.cpp"
+  [=[        if (!m_logManager.StartActivity()) {
+            return;
+        }]=]
+  [=[        if (!m_logManager.StartActivity()) {
+            return;
+        }
+        // Balance the activity count on every exit path, including an exception unwinding out of
+        // this function; leaking it would deadlock a later FlushAndTeardown() in WaitPause().
+        struct OrtGenAIActivityGuard final {
+            ILogManager& log_manager;
+            ~OrtGenAIActivityGuard() { log_manager.EndActivity(); }
+        } ortgenai_activity_guard{m_logManager};]=])
+
+ortgenai_replace_required(
+  "${SOURCE_DIR}/lib/offline/OfflineStorageHandler.cpp"
+  [=[        m_flushPending = false;
+        m_logManager.EndActivity();]=]
+  [=[        m_flushPending = false;
+        // EndActivity() now runs from ortgenai_activity_guard above, on every exit path.]=])
+
+# LogManagerProvider::Release() walks LogManagerFactory's registries to find and destroy the manager.
+# The factory is a function-local static, so during static destruction it can already be destroyed by
+# the time a host's exit path calls Release() -- the registries are then destroyed std::maps and
+# walking them dereferences freed nodes (observed: EXC_BAD_ACCESS in LogManagerFactory::release()).
+# Make the factory immortal: it is a small, fixed-size object, and leaking it keeps the registries
+# readable so Release() can still perform the *real* teardown (FlushAndTeardown + worker-thread join)
+# it is being asked to do. Reported upstream; drop this hunk once the fix is in a pinned release.
+ortgenai_replace_required(
+  "${SOURCE_DIR}/lib/api/LogManagerFactory.hpp"
+  [=[        static LogManagerFactory& instance() {
+            static LogManagerFactory impl;
+            return impl;
+        }]=]
+  [=[        static LogManagerFactory& instance() {
+            static LogManagerFactory& impl = *new LogManagerFactory();
+            return impl;
+        }]=])
+
+# GetPAL()'s function-local static is destroyed at a point that is not ordered against the host's
+# teardown call: PAL is constructed lazily on first telemetry use, so whether its static outlives the
+# LogManager teardown depends on runtime timing. When it does not, LogManagerImpl::FlushAndTeardown()
+# -> PAL::shutdown() releases shared_ptr members of an already-destroyed PlatformAbstractionLayer and
+# faults (observed: intermittent EXC_BAD_ACCESS in ~shared_ptr<ISystemInformation>, roughly 3 in 25
+# process exits). Make the PAL immortal so shutdown() always operates on live members; it is a single
+# fixed-size object and PAL::shutdown() already performs the real resource teardown explicitly.
+# Reported upstream; drop this hunk once the fix is in a pinned release.
+ortgenai_replace_required(
+  "${SOURCE_DIR}/lib/pal/PAL.cpp"
+  [=[        static PlatformAbstractionLayer pal;
+        return pal;]=]
+  [=[        static PlatformAbstractionLayer& pal = *new PlatformAbstractionLayer();
+        return pal;]=])
--- a/src/generators.cpp
+++ b/src/generators.cpp
@@ -152,21 +152,30 @@ void Shutdown() {
 
   // Reset g_ort_globals directly (rather than through GetOrtGlobals(), which would lazily construct
   // the globals just to immediately tear them down). If genai was never initialized there is nothing
-  // to do. ~OrtGlobals tears down the device interfaces (including the RyzenAI EP shutdown), unloads
-  // the genai add-on libraries, and finally releases the OrtEnv.
+  // to do. ~OrtGlobals clears the graph session cache and the device allocators, tears down the
+  // device interfaces (including the RyzenAI EP shutdown), unloads the genai add-on libraries, and
+  // finally releases the OrtEnv. All of that runs on both the runtime and the process-exit path; the
+  // only difference at process exit is that the OrtEnv itself is leaked rather than destroyed, for
+  // the reason described below.
   std::scoped_lock lock{g_ort_globals_mutex};
 
-  // At process exit (the EnsureShutdown static destructor sets g_process_exiting under this lock),
-  // running ~OrtGlobals is unsafe: releasing the OrtEnv from within __cxa_finalize drives ORT's
-  // Environment/LoggingManager destructor to lock a mutex whose backing static may already be
-  // finalized, which throws "mutex lock failed: Invalid argument" -- and a throwing destructor at
-  // teardown is uncaught, so the process aborts via std::terminate. Re-init after shutdown is
-  // impossible once the process is exiting, so intentionally leak the globals here and let the OS
-  // reclaim the memory. Runtime OgaShutdown() (g_process_exiting == false) still performs the full
-  // teardown that in-process re-init relies on.
-  if (g_process_exiting) {
-    (void)g_ort_globals.release();
-    return;
+  if (g_ort_globals && g_process_exiting) {
+    // At process exit (the EnsureShutdown static destructor sets g_process_exiting under this lock),
+    // destroying the OrtEnv is unsafe: releasing it from within __cxa_finalize drives ORT's
+    // Environment/LoggingManager destructor to lock a mutex whose backing static may already be
+    // finalized, which throws "mutex lock failed: Invalid argument" -- and a throwing destructor at
+    // teardown is uncaught, so the process aborts via std::terminate. Nothing else ~OrtGlobals does
+    // (clearing the graph session cache, tearing down device interfaces including the RyzenAI EP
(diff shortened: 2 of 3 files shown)


src/generators.cpp
```diff
@@ -152,9 +152,23 @@ void Shutdown() {
...
+  // Environment/LoggingManager destructor to lock a mutex whose backing static may already be
+  // finalized, which throws "mutex lock failed: Invalid argument" -- and a throwing destructor at
+  // teardown is uncaught, so the process aborts via std::terminate. Re-init after shutdown is
+  // impossible once the process is exiting, so intentionally leak the globals here and let the OS
+  // reclaim the memory. Runtime OgaShutdown() (g_process_exiting == false) still performs the full
+  // teardown that in-process re-init relies on.
+  if (g_process_exiting) {
+    (void)g_ort_globals.release();
+    return;
+  }
```
tianleiwu: **The leak is much broader than the diagnosed bug, and skips real teardown work.**

The abort you describe originates only in step 4 of `~OrtGlobals` (`env_.reset()` -> ORT `~Environment`/`~LoggingManager`). Releasing the whole `unique_ptr<OrtGlobals>` also skips steps 1-3, which are not cleanup-for-cleanup's-sake:

- `owned_interfaces_.clear()` -> `RyzenAI::Interface::~Interface()` resolves and calls `RyzenAI_Shutdown` in the resident EP module (Windows). Skipping this leaves the NPU EP un-shut-down for any host that does not call `OgaShutdown()`.
- `cuda_library_.reset()` -> unloads the genai CUDA add-on so *its* static destructors run.
- `graph_session_cache_.sessions_.clear()` and the `device_allocators_` teardown.

All three ran on this same process-exit path before this PR and none were implicated in the macOS abort. The PR body notes "only the process-exit path changes, on all platforms" — that is the concern: a macOS-arm64 fix ends up disabling a Windows NPU shutdown hook for every host that never calls `OgaShutdown()`, which the header documents as SHOULD-not-MUST.

`env_` is a public `std::unique_ptr<OrtEnv>` on `OrtGlobals`, so leaking only the object that is actually unsafe to destroy is a one-line change:

```suggestion
  // At process exit, releasing the OrtEnv from within __cxa_finalize drives ORT's
  // Environment/LoggingManager destructor to lock a mutex whose backing static may already be
  // finalized, which throws "mutex lock failed: Invalid argument" -- and a throwing destructor at
  // teardown is uncaught, so the process aborts via std::terminate. Leak just the env and let the
  // OS reclaim it; everything that depends on it (device interfaces including the RyzenAI EP
  // shutdown, the genai add-on libraries, the session allocators) is still torn down below.
  if (g_process_exiting && g_ort_globals)
    (void)g_ort_globals->env_.release();
```

The ordering rationale in `~OrtGlobals` ("destroy everything that uses env-owned resources BEFORE the env") is not just preserved but strengthened, since the env then outlives all of it unconditionally.

If the broad leak is deliberate — e.g. you have evidence that unloading the CUDA add-on via `FreeLibrary`/`dlclose` during `__cxa_finalize` is *also* unsafe — could you say so in the PR body and in the comment? As written, nothing here signals that RyzenAI/CUDA teardown is being dropped.

src/generators.cpp
```diff
@@ -152,9 +152,23 @@ void Shutdown() {
 
   // Reset g_ort_globals directly (rather than through GetOrtGlobals(), which would lazily construct
   // the globals just to immediately tear them down). If genai was never initialized there is nothing
-  // to do. Delete now because on process exit is too late. ~OrtGlobals tears down the device
-  // interfaces (including the RyzenAI EP shutdown) and unloads the genai add-on libraries.
+  // to do. ~OrtGlobals tears down the device interfaces (including the RyzenAI EP shutdown), unloads
+  // the genai add-on libraries, and finally releases the OrtEnv.
```
tianleiwu: Nit — comment drift. This retained paragraph still asserts that `~OrtGlobals` "tears down the device interfaces (including the RyzenAI EP shutdown), unloads the genai add-on libraries, and finally releases the OrtEnv". With the new early return that is only true on the runtime path. If the targeted `env_`-only leak isn't adopted, please add an explicit "none of this runs at process exit" sentence here so the next reader doesn't assume the RyzenAI shutdown still fires.

src/telemetry/telemetry.cpp
```diff
@@ -283,25 +283,47 @@ void GenAiTelemetry::Shutdown() {
...
+
+  // Process exit (static destructor): deliberately do NOT run the 1DS teardown -- just leak the
+  // manager. FlushAndTeardown() calls an *unbounded* WaitPause() that deadlocks at exit if a logging
+  // activity count was left unbalanced (a thread torn down mid-LogEvent by exit()), which would wedge
+  // the whole process. Leaking is safe here: 1DS does not tear the manager down itself at process
+  // exit (LogManagerFactory keeps managers in a std::set<ILogManager*> of raw pointers and its
+  // destructor is empty), so ~LogManagerImpl never runs and there is no hang; the OS reclaims the
+  // manager and its background threads. Re-init after process exit is impossible, so nothing is lost,
+  // and no teardown work races the surrounding static-destruction sequence.
+  (void)impl.release();
```
tianleiwu: **The exit path still blocks on `mutex_`, so this may not actually fix the reported hang.**

`Shutdown()` unconditionally takes `std::unique_lock<std::shared_mutex> lock(mutex_)` at line 279, before any `s_instance_destroyed` check. The comment above names the root cause as "a logging activity count left unbalanced (a thread torn down mid-`LogEvent` by `exit()`)". A thread that dies inside `LogEvent` is still holding the `std::shared_lock` handed out by `LockForLogging()` — so the exclusive acquisition at line 279 never returns and the process wedges *before* ever reaching this leak. Same symptom, different mutex.

Could you confirm from the captured Foundry-Local hang stack whether the wait was inside 1DS's `WaitPause()` or inside `mutex_`?

If the exit path must not block at all, it may be cleaner to make the destructor non-blocking and drop the `s_instance_destroyed` branch from `Shutdown()` entirely:

```cpp
GenAiTelemetry::~GenAiTelemetry() {
  s_instance_destroyed.store(true);
#if defined(ORTGENAI_ENABLE_TELEMETRY)
  // Never block the exit path: a thread killed mid-LogEvent still holds a shared_lock, so a
  // blocking acquisition here would wedge the process just as FlushAndTeardown() did.
  std::unique_lock<std::shared_mutex> lock(mutex_, std::try_to_lock);
  initialized_.store(false);
  (void)impl_.release();
#endif
}
```

That also removes the "`s_instance_destroyed == false` is exactly *not at process exit*" invariant, which currently depends on a TOCTOU-prone `IsDestroyed()` check in `OgaShutdown()`.

src/telemetry/telemetry.cpp
```diff
@@ -283,25 +283,47 @@ void GenAiTelemetry::Shutdown() {
...
-  impl_.reset();
+
+  // Process exit (static destructor): deliberately do NOT run the 1DS teardown -- just leak the
+  // manager. FlushAndTeardown() calls an *unbounded* WaitPause() that deadlocks at exit if a logging
+  // activity count was left unbalanced (a thread torn down mid-LogEvent by exit()), which would wedge
+  // the whole process. Leaking is safe here: 1DS does not tear the manager down itself at process
+  // exit (LogManagerFactory keeps managers in a std::set<ILogManager*> of raw pointers and its
+  // destructor is empty), so ~LogManagerImpl never runs and there is no hang; the OS reclaims the
+  // manager and its background threads. Re-init after process exit is impossible, so nothing is lost,
+  // and no teardown work races the surrounding static-destruction sequence.
```
tianleiwu: Two things this justification doesn't cover:

**1. The leaked manager keeps background threads running through the rest of static destruction.** 1DS is FetchContent-built and statically linked into the genai module (see `cmake/telemetry.cmake`), so the leaked manager's uploader / offline-storage threads keep executing genai-module code and touching curl/SQLite/CRT statics *while those statics are being finalized around them*. `FlushAndTeardown()` previously joined them. This trades a deterministic hang for a nondeterministic exit crash. Consider a bounded quiesce before leaking — `PauseTransmission()`, or dropping the transmit profile — anything that does not go through the unbounded `WaitPause()`. At minimum, please state the tradeoff here rather than "there is no hang".

**2. "nothing is lost" isn't quite right — RAM-queued events are dropped.** Neither `Flush()` nor `FlushAndTeardown()` runs on this path, so events still in the 512 KiB RAM queue are discarded instead of persisted to the offline cache for the next run. `CFG_INT_MAX_TEARDOWN_TIME = 0` only meant "don't *upload* at exit"; persistence still happened. Since `ort_genai_c.h` documents `OgaShutdown()` as SHOULD-not-MUST, "host never calls `OgaShutdown()`" is the majority path, so expect a measurable drop in end-of-session events. Is 1DS `Flush()` (persist-to-disk only, no teardown) bounded? If so, keeping just that call on the exit path recovers the data without reintroducing the deadlock. Either way this behavior change is worth a line in the PR description.

src/telemetry/telemetry.cpp
```diff
@@ -283,25 +283,47 @@ void GenAiTelemetry::Shutdown() {
...
-      MAT::LogManagerProvider::Release(impl_->config);
-    } catch (...) {
+  if (!impl_) return;
+  auto impl = std::move(impl_);
+
+  // Explicit runtime shutdown (e.g. OgaShutdown()): tear down synchronously. No threads are being
+  // force-killed here, so FlushAndTeardown() completes promptly, and a synchronous teardown keeps a
+  // subsequent Initialize() from racing a half-torn-down manager over the SDK's process-global state.
+  // OgaShutdown() only reaches here while the singleton is alive (it checks IsDestroyed() first), so
+  // s_instance_destroyed == false is exactly "not at process exit".
```
tianleiwu: `s_instance_destroyed` (and `g_process_exiting` on the ORT side) really mean "static destruction has begun", not "the process is exiting". Both also fire when the genai shared library is `dlclose`d / `FreeLibrary`d at runtime — Python extension teardown, host plugin unload. In that case the leaked 1DS threads are executing code in a module that is about to be unmapped, which is a hard crash, and the "the OS reclaims it" justification below does not apply.

At minimum please reword so the invariant isn't overstated. If module unload is a supported scenario for genai, it needs a different answer than leaking.

src/telemetry/telemetry.cpp
```diff
@@ -283,25 +283,47 @@ void GenAiTelemetry::Shutdown() {
...
-    }
-    try {
-      impl_->log_manager->FlushAndTeardown();
-    } catch (...) {
-    }
-    try {
-      MAT::LogManagerProvider::Release(impl_->config);
-    } catch (...) {
+  if (!impl_) return;
+  auto impl = std::move(impl_);
```
tianleiwu: Nice — moving `impl_` out before any teardown is a real improvement over mutating it in place, since `LockForLogging()` tests `impl_` / `impl_->logger` and can no longer observe a half-torn-down `Impl`.

Nit on structure: consider inverting the branch below so the exceptional (process-exit) case is the early return, mirroring the shape you used in `Generators::Shutdown()`:

```cpp
if (s_instance_destroyed.load()) {
  (void)impl.release();  // see comment: never run 1DS teardown at exit
  return;
}
// normal 1DS teardown ...
```

The current form buries a bare `return;` in the middle of the function with the long exit-path rationale trailing after it.

tianleiwu (changes requested):
Thanks for the detailed root-cause write-up in the comments — the split between "runtime `OgaShutdown()` = real teardown" and "process exit = leak" is the right shape for both hazards, and moving `impl_` out before teardown so `LockForLogging()` can never observe a half-torn-down `Impl` is a genuine improvement.

Two things I'd like to resolve before this merges:

1. **The ORT-side leak is broader than the diagnosed bug** and skips teardown work with externally visible effects (RyzenAI EP shutdown, CUDA add-on unload). Inline on `src/generators.cpp`.
2. **The telemetry exit path still blocks on the exclusive `mutex_`** before it can reach the leak, on exactly the scenario the comment names as the root cause. Inline on `src/telemetry/telemetry.cpp`.

### Cross-cutting (files not in this diff)

- **Doc drift in `src/ort_genai_c.h`.** The `OgaShutdown()` warning still reads: *"If it is not called, GenAI's globals are destroyed at static-destruction time in undefined order, which may crash."* After this PR they are deliberately **not** destroyed. Please update it to say the process-exit path intentionally leaks, and — more useful for callers — that EP shutdown hooks (RyzenAI), add-on library unload, and telemetry flush now only run when `OgaShutdown()` is called explicitly. This PR makes calling `OgaShutdown()` *more* important, and the header should say so.

- **Two flags, one concept.** `g_process_exiting` (generators.cpp) and `s_instance_destroyed` (telemetry.cpp) encode the same predicate but are set by two independent static destructors whose relative order is construction-order dependent. A single shared `Generators::IsStaticDestructionInProgress()` helper, set by whichever destructor runs first, would make the two paths impossible to disagree.

- **No test coverage for the changed path.** `test/reinit_tests.cpp` covers the runtime path this PR preserves, but nothing covers the path it changes. Since process exit is only reachable from a real process exit, consider a small helper executable registered with ctest that loads a model, runs one generate, and returns from `main()` *without* calling `OgaShutdown()`, asserting exit code 0 and a bounded wall-clock time. Without it, both the hang and the abort can silently regress.

- **Please link the upstream ORT issue** for the `~Environment`/`~LoggingManager` static-destruction-order bug in the code comment, so the workaround can be removed once ORT fixes it rather than becoming permanent by default.

src/generators.cpp
baijumeswani: Thanks. I made the change so now we only leak/release the `g_ort_globals->env_` during the static destruction process.

src/generators.cpp
baijumeswani: Updated the comment to reflect what the code does.

src/telemetry/telemetry.cpp
baijumeswani: I dig deeper into the 1ds library and found a few bugs in their teardown process that were manifesting 

I confirmed this from the stack from the process when it hangs that the blocked thread was inside 1DS `WaitPause()`.

I moved the fix into a patch for the 1ds library so we can proceed with the release. Will submit an upstream PR shortly.

src/telemetry/telemetry.cpp
baijumeswani: Removed this approach. And implemented a real fix as a patch.

src/telemetry/telemetry.cpp
baijumeswani: After the patch, the manager is no longer leaked, so its worker threads are joined before module unload.

src/telemetry/telemetry.cpp
baijumeswani: The onnxruntime-genai telemetry shutdown changes have been removed entirely now after the patch fix.

bmehta001 (dismissed):
Thanks, tested this last version and everything works. Stale .wal SQLite files are no longer left behind.

cmake/patches/cpp_client_telemetry/apply_patch.cmake
```diff
@@ -101,3 +101,76 @@ ortgenai_replace_required(
...
+# exits, versus 85/85 clean with this hunk alone). Tolerating the exception is the correct fix.
+# The guard's destructor locks m_pause_mutex, a member of the still-live heap-allocated
+# LogManagerImpl, so it cannot itself throw while unwinding.
+# Reported upstream; drop this hunk once the fix is in a pinned release.
+ortgenai_replace_required(
+  "${SOURCE_DIR}/lib/offline/OfflineStorageHandler.cpp"
+  [=[        if (!m_logManager.StartActivity()) {
+            return;
+        }]=]
+  [=[        if (!m_logManager.StartActivity()) {
```
tianleiwu, line 133: **The rejected-activity path can still deadlock `WaitForFlush()`.** A scheduled flush sets `m_flushPending = true` and resets `m_flushComplete` before its task starts. If teardown reaches `PauseActivity()` first, `StartActivity()` returns false here; the retained immediate return then neither posts completion nor clears the pending flag, so `OfflineStorageHandler::Shutdown()` can wait forever. The upstream companion fix in [cpp_client_telemetry#1429](https://github.com/microsoft/cpp_client_telemetry/pull/1429) handles this path in addition to adding the activity guard. Please carry that part of the fix too, under `m_flushLock`:

```cpp
if (!m_logManager.StartActivity()) {
    LOCKGUARD(m_flushLock);
    m_flushHandle.Cancel();
    m_flushComplete.post();
    m_flushPending = false;
    return;
}
```

tianleiwu (changes requested):
Thanks for reworking the latest head. The previous ORT teardown-scope and telemetry-manager lifetime concerns are addressed: process exit now leaks only the unsafe `OrtEnv`, while the 1DS manager still performs real teardown and joins its worker threads. One teardown race remains before this is safe to merge; details inline.

tianleiwu (approved):
The follow-up commit fixes the remaining rejected-flush teardown race: when `StartActivity()` fails after the log manager pauses, the patch now cancels the scheduled handle, signals completion, and clears the pending flag under `m_flushLock`, matching the upstream fix. The previous blocking concerns are addressed. The process-exit regression test remains a useful nonblocking follow-up; CI for this head is still running.

baijumeswani:
Thanks for the review @tianleiwu @bmehta001. :)

kunal-vaishnavi (approved):

PR #2365, opened 2026-08-03
Transactional engine execution

baijumeswani:
## Summary

This PR makes each dynamic paged continuous-batching Engine step transactional.

A batch now either:

- commits all Request, Search/sampler, scheduler-visible, and paged KV-cache progress; or
- rolls back without publishing partial results.

This provides the correctness foundation for long-running local workloads such as coding agents, where multiple requests share one Engine and may carry long contexts. The implementation keeps rollback cost independent of context length: it does not copy KV tensors or token history.

The existing static-batching path is unchanged.

## Why this is needed

Before this change, a dynamic Engine step updated several independently owned pieces of state:

- scheduler admission and batching decisions;
- paged-cache block ownership and occupied slots;
- Search sequence, EOS, completion, and RNG state;
- Request host-visible tokens, progress counters, and status; and
- ready-request publication.

If model execution or post-processing failed after only some of those updates, retrying could observe partially advanced state. That is especially problematic when one model run serves several requests: one request failing must not expose progress from another request in the same batch.

This PR introduces an explicit prepare/commit boundary around the complete shared step.

## Transactional code flow

```mermaid
flowchart TD
    A[Engine::Step] --> B{Ready request already queued?}
    B -- Yes --> C[Return ready request]
    B -- No --> D{Dynamic paged batching?}
    D -- No --> E[Existing static batching path]
    D -- Yes --> F[Build immutable StepPlan]

    F --> G{Executable work?}
    G -- No work --> H[Return null]
    G -- Unserviceable --> I[Return structured EngineStepError]
    G -- Yes --> J[Reserve complete paged-cache proposal]

    J --> K[Validate all requests]
    K --> L[Checkpoint Search state]
    L --> M[Checkpoint scheduler-owned sampler RNG state]
    M --> N[Run model with proposed block table]
    N --> O[Process logits and sample next tokens]
    O --> P[Stage tokens, completion state, and ready requests]

    P --> Q[Commit Search/sampler checkpoint]
    Q --> R[Commit paged-cache ownership and occupied slots]
    R --> S[Publish Request host state]
    S --> T[Publish ready requests]
    T --> C

    K -. validation failure .-> U[Restore Search/sampler state]
    N -. retryable execution failure .-> U
    O -. post-processing failure .-> U
    U --> V[Release reserved cache blocks]
    V --> W[Return retryable batch-abort outcome]

    J -. reservation contract failure .-> X[Mark Engine unhealthy]
    N -. unknown or fatal failure .-> Y[Best-effort rollback]
    Y --> X
    Q -. commit failure .-> X
    X --> Z[Reject subsequent steps with stored fatal error]
```

The successful path is:

```text
plan -> reserve -> checkpoint -> execute -> stage -> commit -> publish
```

The safe failure path is:

```text
failure -> restore Search/sampler -> release reservation -> report retryable abort
```

## Component responsibilities

| Component | Transactional responsibility |
| --- | --- |
| `Engine` | Coordinates planning, reservation, execution, rollback, commit, publication, and Engine health. |
| `DynamicBatchScheduler` | Builds a stable `StepPlan` without mutating live admission state. |
| `PagedKeyValueCache` | Calculates resource needs and creates an all-or-nothing cache reservation. |
| `PagedCacheReservation` | Owns speculative blocks and proposed block tables until commit or release. |
| `ModelExecutor` / decoder | Executes against the proposed cache layout without committing ownership first. |
| `ScheduledRequests` | Coordinates Search checkpoints and the scheduler-owned batched sampler checkpoint. |
| `Search` | Restores logical sequence length, EOS state, done state, and local sampling state when applicable. |
| `BatchedSampler` | Checkpoints and restores persistent per-Request CUDA RNG states as one batch. |
| `Request` | Stages generated tokens and publishes host-visible progress only during commit. |

## Paged-cache transactions

The scheduler first builds a plan containing the requests selected for the next model run and their target cache occupancy.

The cache then reserves the entire proposal atomically:

- growth in an existing request's partially filled tail block;
- additional blocks for active requests;
- blocks for newly admitted requests; and
- the proposed padded block table used by the model.

Reserved blocks are not visible as committed Request ownership until the batch succeeds. Rollback returns them to the block pool. Commit advances logical slot occupancy and publishes new block ownership without copying KV data.

## Search and sampler rollback

The Engine supports both standalone Search sampling and scheduler-owned batched sampling.

For the batched CUDA path:

- each Request owns a stable sampler-state identity;
- the sampler gathers selected persistent CURAND states into one checkpoint buffer;
- rollback scatters those states back in one batch;
- Search separately restores EOS, done, deferred-completion, and logical sequence state; and
- shared next-token storage remains scratch that is overwritten before a retry reads it.

CUDA rollback is deliberately two-phase:

1. drain outstanding sampling/tail work;
2. queue sampler and Search device restores;
3. synchronize once for the batch; and
4. publish restored pinned-host completion state.

This avoids per-Request synchronization and prevents an outstanding EOS kernel from overwriting restored host state.

## Request publication

Generated output is not immediately visible to callers.

During execution, each Request stages:

- whether a token was appended;
- the generated token value;
- completion state; and
- the post-step lifecycle status.

Only after model execution, sampling, and cache commit are known to be successful does the Engine publish:

- the host token mirror;
- processed-length and prefill state;
- Request completion status; and
- ready-request events.

The host token vector reserves `max_length` when the Request is assigned, so commit does not introduce an allocation proportional to context length.

## Failure behavior

| Failure | Result |
| --- | --- |
| Request validation or logits/post-processing failure | Restore Search/sampler state, release cache reservation, return a retryable batch-abort error with the original cause. |
| Explicitly retryable model-execution failure | Roll back and return a retryable batch-abort error. |
| Unknown model, device, or execution failure | Attempt rollback, mark the Engine unhealthy, and reject subsequent steps. |
| Rollback failure | Mark the Engine unhealthy because state can no longer be proven consistent. |
| Commit failure | Mark the Engine unhealthy because publication may be partially complete. |
| Permanently unserviceable request | Return a structured unserviceable-request outcome without blocking smaller fitting requests. |

## Performance characteristics

The transactional path is designed for large batches and long contexts:

- no KV-cache tensor copies during checkpoint or rollback;
- no token-history copies during checkpoint or rollback;
- no context-length-dependent device-to-host copy;
- one shared next-token D2H copy per decode step;
- batch-level RNG gather/scatter;
- no per-Request CUDA synchronization;
- proposed block tables are built from host-owned planning state; and
- static batching stays on its existing path.

The implementation was simplified after review by removing a resource-less transaction phase state machine, no-op commit hooks, write-only plan/context fields, and a synthetic timing test. The final cleanup removed 349 net lines without changing the transaction boundary.

### Correctness

- 99 Engine/Search tests pass.
- 8 CUDA sampling tests pass.
- Coverage includes:
  - initial admission and active-request cache growth;
  - atomic cache reservation, commit, and release;
  - deterministic CPU and CUDA sampling retry;
  - scheduler-owned batched sampler RNG rollback;
  - EOS, done, logical sequence, and deferred-completion restore;
  - staged Request host publication;
  - failure of a later Request restoring earlier samples;
  - retryable versus fatal execution outcomes;
  - Engine unhealthy-state behavior;
  - paged-cache ownership and accounting invariants; and
  - preservation of the static-batching path.

Tests that only exercised implementation scaffolding or synthetic orchestration timing were removed. The retained tests execute production planning, reservation, Request, Search, sampler, rollback, and commit paths.

### Real-model coverage

- Qwen2.5 0.5B:
  - batch sizes 1, 4, and 8;
  - staggered request admission; and
  - an 8,196-token context.
- Qwen2.5 7B on an A100 80 GB:
  - batch sizes 8, 16, and 32;
  - batch 32 sustained approximately 1.19K aggregate tokens/s; and
  - batch 32 delivered approximately 1.95x batch-16 aggregate throughput.


src/engine/paged_key_value_cache.cpp
```diff
@@ -153,46 +138,188 @@ void PagedKeyValueCache::AppendTokens(std::shared_ptr<Request> request) {
...
+    const auto& table = block_tables_[i];
+    if (entry.newly_admitted || entry.request_id != table.request_id) {
+      throw std::runtime_error("Step plan order does not match the committed block table order.");
+    }
+
+    const auto growth = calculate_growth(entry, &table);
+    planned_blocks += growth.new_blocks;
+    max_blocks_per_request =
+        std::max(max_blocks_per_request, growth.proposed_blocks);
+    if (planned_blocks > available_blocks ||
```
tianleiwu: This treats cumulative pressure across all committed requests as permanent unserviceability of `entry`. Admission reserves only immediate cache needs, so two active requests can each need one new block while only one block is free. Either request is executable alone and may finish to release capacity, but this branch rejects the whole step and `Engine::Step()` cannot progress. The candidate loop below already separates permanent unserviceability from temporary deferral; please do the same for committed work and select a fitting active subset (or guarantee each admitted request’s full future growth). A real paged-cache test for this two-active/one-free-block case would cover the boundary.

tianleiwu (changes requested):
The transactional structure and rollback coverage are strong, and the dedicated 99-test engine/Search suite passes. One blocking scheduling issue remains: temporary aggregate cache pressure across already-committed requests is classified as permanent unserviceability, so the engine can stall even when a smaller active subset is executable.

tianleiwu (dismissed):
LGTM

tianleiwu (approved):

baijumeswani:
Thank you for your review @tianleiwu . :)

PR #2368, opened 2026-08-04
Complete Marian batching correctness

mustjab:
## Summary

Completes Marian batching correctness with value-sensitive coverage for variable-length inputs and beam search.

## Changes

- Reorder Marian SSRU recurrent state using selected parent-beam indices.
- Fix `OgaGenerator_GetLogits` sizing for beam search by using `batch_size * num_beams`.
- Add compact deterministic fixtures covering:
  - per-row EOS insertion and attention-mask values,
  - recurrent-state parent-beam reordering, and
  - beam-search logits sizing.
- Add Python and C API coverage for extracting generated output with existing APIs.

## Output extraction

No new generated-sequence API is introduced. Applications can:

1. Record `TokenCount()` after appending the padded batch.
2. Retrieve each final sequence with the existing sequence APIs.
3. Slice each sequence from that recorded offset.
4. Stop at the first EOS or PAD token.

## Motivation

Marian batching support previously left two correctness gaps: attention-mask values were not tested for unequal prompt lengths, and beam search copied recurrent state in physical beam order instead of following selected parent beams.

## Validation

- Clean MinSizeRel native and Python builds.
- Native focused tests: 4 passed.
- Python API module: 18 passed, 45 model/environment-dependent skips.
- Lintrunner: clean.
- Real Marian 97-sentence serial-versus-batch parity: zero token mismatches.
- Reproducible CPU benchmark against commit `08695dbe`:
  - batch 20: 4.36x E2E, 7.02x generation-only,
  - batch 32: 4.73x E2E, 7.42x generation-only.
- Beam reorder test was mutation-checked: restoring direct state copying causes the test to fail.
## Dependency roll

- Update ONNX Runtime Extensions to [microsoft/onnxruntime-extensions#1099](https://github.com/microsoft/onnxruntime-extensions/pull/1099) (60687b12) for locale-independent Marian Unicode case restoration.


src/generators.cpp
```diff
@@ -930,4 +932,23 @@ DeviceSpan<int32_t> Generator::GetSequence(size_t index) const {
...
 }
 
+DeviceSpan<int32_t> Generator::GetGeneratedSequence(size_t index) const {
+  auto sequence = GetSequence(index);
+  if (generated_sequence_start_ > sequence.size())
+    throw std::runtime_error("Generated sequence start is past the current sequence length");
+
+  auto generated = sequence.subspan(generated_sequence_start_, sequence.size() - generated_sequence_start_);
+  auto generated_cpu = generated.CopyDeviceToCpu();
+  const auto& eos_token_ids = model_->config_->model.eos_token_id;
```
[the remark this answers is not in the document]
mustjab: Fixed in df72ad7e. GetGeneratedSequence now performs the device-to-CPU copy once and scans the generated subspan through CpuSpan().

src/ort_genai_c.h
```diff
@@ -661,6 +661,22 @@ OGA_EXPORT OgaResult* OGA_API_CALL OgaGenerator_SetLogits(OgaGenerator* generato
  */
 OGA_EXPORT size_t OGA_API_CALL OgaGenerator_GetSequenceCount(const OgaGenerator* generator, size_t index);
 
+/**
+ * \brief Returns the number of generated tokens after the most recent append and before the first EOS or PAD token.
+ * \param[in] generator The generator to get the generated sequence count from.
+ * \param[in] index The sequence index.
+ * \return The generated token count.
+ */
```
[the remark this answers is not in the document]
mustjab: Fixed in df72ad7e. Both generated-sequence C API declarations now document that RNNT/TDT transducer models return the full emitted token stream.

src/ort_genai_c.cpp
```diff
@@ -644,6 +644,20 @@ size_t OGA_API_CALL OgaGenerator_GetSequenceCount(const OgaGenerator* generator,
...
+  }
+  return generator->GetGeneratedSequence(index).size();
+}
+
+const int32_t* OGA_API_CALL OgaGenerator_GetGeneratedSequenceData(const OgaGenerator* generator, size_t index) {
+  if (auto* transducer = dynamic_cast<Generators::TransducerState*>(generator->state_.get())) {
+    return transducer->GetAllTokens().data();
+  }
+  return generator->GetGeneratedSequence(index).CopyDeviceToCpu().data();
+}
```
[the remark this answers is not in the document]
mustjab: Fixed in df72ad7e. OgaGenerator_GetGeneratedSequenceData now returns the already-copied CPU span instead of issuing a second device-to-CPU copy.

src/python/python.cpp
```diff
@@ -234,6 +234,10 @@ struct PyGenerator {
     return ToPython(generator_->GetSequence(index));
   }
 
+  pybind11::array_t<int32_t> GetGeneratedSequence(int index) {
+    return ToPython(generator_->GetGeneratedSequence(index));
```
kunal-vaishnavi: Copilot seems to think that we should not be using `ToPython` for any of the Python bindings.

From Copilot:

`ToPython` creates a non-owning NumPy view over memory owned by `generator_`. If the `Generator` is destroyed—or if a later call refreshes/reallocates the staging buffer—the returned array may become dangling or change contents unexpectedly. Could we either return an owned copy, attach the generator/buffer as the NumPy base object, or document this lifetime explicitly?

src/ort_genai_c.h
```diff
@@ -661,6 +661,24 @@ OGA_EXPORT OgaResult* OGA_API_CALL OgaGenerator_SetLogits(OgaGenerator* generato
...
+ * \brief Returns the number of generated tokens after the most recent append and before the first EOS or PAD token.
+ *        For transducer models (RNNT and TDT), this returns the full emitted token stream.
+ * \param[in] generator The generator to get the generated sequence count from.
+ * \param[in] index The sequence index.
+ * \return The generated token count.
+ */
+OGA_EXPORT size_t OGA_API_CALL OgaGenerator_GetGeneratedSequenceCount(const OgaGenerator* generator, size_t index);
+
+/**
+ * \brief Returns the generated tokens after the most recent append and before the first EOS or PAD token.
```
kunal-vaishnavi: From Copilot:

Could we make the lifetime of this returned pointer explicit in the API documentation? It is backed by the generator’s CPU staging buffer, and a subsequent generator or sequence-data call may refresh that buffer. Please document exactly when the pointer becomes invalid—for example, whether it remains valid until the next generator call or only until another sequence-data accessor is called.

src/python/python.cpp
```diff
@@ -234,6 +234,10 @@ struct PyGenerator {
     return ToPython(generator_->GetSequence(index));
   }
 
+  pybind11::array_t<int32_t> GetGeneratedSequence(int index) {
```
kunal-vaishnavi: I'm wondering if it makes sense to do this or have an attribute in `GetSequence` for whether the output should contain padding or not. `GetGeneratedSequence` is quite similar in name to `GetSequence` and I don't want to introduce user confusion.

src/python/python.cpp
mustjab: Addressed in 29341465. `ToPython` now allocates an owned NumPy array and copies the span into it, so all sequence/tokenizer arrays remain valid independently of generator lifetime or later staging-buffer updates. Added a regression test that mutates the returned array and verifies generator storage is unchanged.

src/ort_genai_c.h
mustjab: Addressed in 29341465. The C API docs now state that the pointer remains valid until the next generator API call or generator destruction, and instruct callers to copy data when it must outlive either event.

src/python/python.cpp
mustjab: Addressed in 29341465. Python now exposes a single `get_sequence(index, generated_only=False)` method. Callers use `generated_only=True` for prompt-free, EOS/PAD-trimmed output; the separate `get_generated_sequence` Python method was removed.

baijumeswani: Do we need a new api at all? Could the application not get the sequence length before generation begins, and then query the token count after generation ends, and then extract the generated sequence from there?

mustjab: Agreed. Addressed in 08695dbe by removing the generated-sequence API entirely across C, C++, and Python. Applications can record `TokenCount()` after appending the padded batch, retrieve each final sequence with the existing APIs, slice from that offset, and stop at EOS/PAD. The value-sensitive Marian tests now exercise that existing-API flow. The separate owned-copy fix for Python arrays remains.

src/python/python.cpp
```diff
@@ -45,8 +45,10 @@ std::span<T> ToSpan(OgaTensor& v) {
 }
 
 template <typename T>
-pybind11::array_t<T> ToPython(std::span<T> v) {
-  return pybind11::array_t<T>{{v.size()}, {sizeof(T)}, v.data()};
+pybind11::array_t<std::remove_const_t<T>> ToPython(std::span<T> v) {
+  pybind11::array_t<std::remove_const_t<T>> result(v.size());
+  std::copy(v.begin(), v.end(), result.mutable_data());
+  return result;
```
baijumeswani: Why is this change needed?

baijumeswani (dismissed):

baijumeswani (approved):

kunal-vaishnavi (approved):

PR #2370, opened 2026-08-04
webgpu: Shape-free synthetic position_ids for Qwen3.5 VL graph capture

qjia7:
## Summary

Enable WebGPU graph capture for the Qwen3.5 **VL** model. Text-only Qwen3.5 was already handled by #2245 (mRoPE bypass — GQA does fused RoPE). VL cannot use that bypass because the three position axes (T/H/W) carry genuinely different values, so mRoPE must be computed explicitly.

On the VL export path the only remaining Shape ops live inside `Qwen35TextModel._make_synthetic_position_ids` (`qwen.py` lines 1475 and 1502). Their INT64 outputs land on CPU and force `MemcpyToHost`/`MemcpyFromHost` around every downstream consumer, breaking graph capture.

This PR follows the existing `enable_webgpu_graph` expansion pattern (see `WebGPU.make_attention_mask_graph_capture_reformatting_for_gqa`) and swaps in a Shape-free construction of the same `[B, S]` INT64 linear-index tensor (`r*S + c`) when `--extra_options enable_webgpu_graph=true` is set.

## Approach

Replace `Shape → Slice → Reshape → Shape → Gather → Range → Reshape` with a chain that never materializes a scalar length or a shape tensor:

```
position_ids [3, B, S]
  → Gather(idx=[0], axis=0)      → [1, B, S]
  → Squeeze(axis=[0])            → [B, S]           (t_positions; shape only used)
  → Mul(*0) → Add(+1)            → [B, S] all ones  (ones_bs)
    ├─ CumSum(axis=1, exclusive) → [B, S]           (row_idx = 0..S-1 per row)
    └─ ReduceSum(axis=[1], keepdims)
        → [B, 1] = S per row
        → CumSum(axis=0, exclusive)
        → [B, 1] = 0, S, 2S, …    (batch_offset)
  → Add(row_idx, batch_offset)   → [B, S] = r*S + c
```

Same shape, dtype, and values as before. Consumers unchanged.

## Activation

Only active when the user passes `--extra_options enable_webgpu_graph=true`. Default exports are byte-identical to today.

## Changes

- `src/python/py/models/builders/expansions/webgpu.py`
  - New `make_synthetic_position_ids_graph_capture` method that emits the Shape-free subgraph.
- `src/python/py/models/builders/base.py`
  - New `make_cum_sum` helper (mirrors `make_reduce_sum` style).
  - Under `enable_webgpu_graph`, additionally bind `_make_synthetic_position_ids` to the graph-capture version (same descriptor pattern already used for `make_attention_mask_reformatting_for_gqa`).

## Test plan

- [ ] Export Qwen3.5-VL with and without `enable_webgpu_graph=true`; confirm the default graph is byte-identical to pre-PR and the graph-capture graph has zero `Shape` nodes under `/model/attn/synthetic_pos_ids/`.
- [ ] One-step prefill numerical equivalence: logits match within fp16 tolerance between default (CPU EP reference) and graph-capture (WebGPU EP) graphs.
- [ ] End-to-end: run Qwen3.5-VL with `enable_graph_capture=1` on WebGPU EP; confirm no `Memcpy*` nodes at runtime and capture activates.
- [ ] Regression: Qwen3.5 text-only export unchanged (this code path is not reached).

### DIFF
--- a/src/python/py/models/builders/base.py
+++ b/src/python/py/models/builders/base.py
@@ -419,6 +419,14 @@ class Model:
                 self.make_attention_mask_reformatting_for_gqa = (
                     WebGPU.make_attention_mask_graph_capture_reformatting_for_gqa.__get__(self, self.__class__)
                 )
+                # Qwen3.5 VL: replace the Shape-based synthetic position_ids
+                # subgraph with a Shape-free equivalent so no Memcpy nodes are
+                # inserted. Bound as an instance attribute so the subclass's
+                # class method is shadowed at call time. Harmless for models
+                # that never call _make_synthetic_position_ids.
+                self._make_synthetic_position_ids = (
+                    WebGPU.make_synthetic_position_ids_graph_capture.__get__(self, self.__class__)
+                )
 
         else:
             return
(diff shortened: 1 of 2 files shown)


qjia7:
Close it since we have mrope support.

PR #2372, opened 2026-08-05
Reuse the host’s loaded ONNX Runtime on Linux and macOS

baijumeswani:
## Summary

Reuse an ONNX Runtime library already loaded by the host before attempting to load another copy on Linux or macOS.

## Why

Hosts such as Foundry Local load ONNX Runtime before ONNX Runtime GenAI and register execution provider plugins on its environment.

GenAI could subsequently load ONNX Runtime using a different filename:

- `libonnxruntime.so` versus `libonnxruntime.so.1` on Linux
- `libonnxruntime.dylib` versus `libonnxruntime.1.dylib` on macOS

This could load a second ONNX Runtime image with a separate environment. Execution providers registered by the host would then be unavailable to GenAI.

## Changes

- Check for resident versioned and unversioned ONNX Runtime libraries using `RTLD_NOLOAD`.
- Reuse the resident library when found.
- Preserve the existing dynamic-loading fallback when ONNX Runtime has not already been loaded.
- Apply the behavior to Linux and macOS.

### DIFF
--- a/src/models/onnxruntime_api.h
+++ b/src/models/onnxruntime_api.h
@@ -201,7 +201,7 @@ inline void* LoadDynamicLibraryIfExists(const std::string& path) {
   return ort_lib_handle;
 }
 
-#if defined(__linux__)
+#if (defined(__linux__) || defined(MACOS_USE_DLOPEN)) && defined(RTLD_NOLOAD)
 inline void* GetLoadedDynamicLibraryIfExists(const std::string& path) {
   LOG_INFO("Attempting to reuse loaded library %s", path.c_str());
   void* ort_lib_handle = dlopen(path.c_str(), RTLD_NOW | RTLD_LOCAL | RTLD_NOLOAD);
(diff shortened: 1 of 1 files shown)


src/models/onnxruntime_api.h
```diff
@@ -201,6 +201,25 @@ inline void* LoadDynamicLibraryIfExists(const std::string& path) {
   return ort_lib_handle;
 }
 
+#if (defined(__linux__) || defined(MACOS_USE_DLOPEN)) && defined(RTLD_NOLOAD)
```
skottmckay, line 206: What defines RTLD_NOLOAD? 

Should we always try the RTLD_NOLOAD path first and fallback to dlopen without RTLD_NOLOAD if that finds nothing, or would there be a downside to that?

baijumeswani: `RTLD_NOLOAD` is defined by `<dlfcn.h>`. It exists on glibc Linux and macOS, but not on all POSIX toolchains. The `#if defined(RTLD_NOLOAD)` keeps those other POSIX platforms happy. They will continue using the dlopen route.

If available, we check to see if ort was already loaded using `RTLD_NOLOAD`. If we don't find it, or if RTLD_NOLOAD is not available on the platform, we fallback to dlopen.

The env variable `ORT_LIB_PATH` is the only other override available. And this serves the purpose where the user manually injects an env override variable (which should probably take precedence).

Will add a comment in the code.

skottmckay (approved):

PR #2376, opened 2026-08-05
feat: AMD GPU umbrella execution provider integration

zz002:
## Summary

Integrates the AMD GPU umbrella execution provider (`amdgpu-ep`, a plugin EP shipped by the ROCm hip-ep project) into GenAI on top of `rel-0.15.1`, adding `DeviceType::AMDGPU` and a device interface that registers and selects the EP through the plugin EP (V2) API.

## What

`src/amdgpu/` holds the new device interface. It resolves the EP library from `AMDGPU_EP_PATH`, then next to `onnxruntime-genai.dll`, `onnxruntime.dll` and the executable, before falling back to the working directory, and registers it on the interface's `OrtEnv` (tolerating an already-registered library from an earlier init cycle). `SetupProvider` picks the EP devices reported as GPUs and appends the EP with the model's provider options; the same options are mirrored to `ep.amdgpu.*` session config entries, which is how the umbrella EP selects its backend. `ShapeInitSessionProviderOptions` forwards those options to the device-init session so the allocator can be created there, and `GetMemoryInfo()` builds the memory info with `CreateMemoryInfo_V2` using the AMD vendor id and GPU device type the EP factory registers with.

The umbrella EP's allocator is host- and device-accessible on the AMD APU iGPU, so model inputs stay in device memory and the device is on the continuous-decoding list.

## Status

This is a draft opened as a patch source for AMD's downstream CI, which currently builds GenAI from `rel-0.15.1` plus this branch. It overlaps in intent with #2165 (AMD GPU EP support); once that lands and covers the umbrella EP's provider-option plumbing, this PR will be closed and the downstream CI switched over. It supersedes #2194, which targeted v0.14.0.

## Test plan

- [x] Windows build against ORT 1.27 (`--no_telemetry`, Ninja, Release)
- [x] `model_benchmark` on Llama-3.1-8B (AWQ int4) with `"AMDGPU": {"profile": "hip"}` and `session.disable_cpu_ep_fallback=1`: matches the v0.14.0 + #2194 baseline on time-to-first-token and token generation throughput

### DIFF
--- a/cmake/global_variables.cmake
+++ b/cmake/global_variables.cmake
@@ -77,6 +77,8 @@ file(GLOB generator_srcs CONFIGURE_DEPENDS
   "${GENERATORS_ROOT}/openvino/*.cpp"
   "${GENERATORS_ROOT}/ryzenai/*.h"
   "${GENERATORS_ROOT}/ryzenai/*.cpp"
+  "${GENERATORS_ROOT}/amdgpu/*.h"
+  "${GENERATORS_ROOT}/amdgpu/*.cpp"
   "${GENERATORS_ROOT}/cuda/session_options.h"
   "${GENERATORS_ROOT}/cuda/session_options.cpp"
   "${GENERATORS_ROOT}/nvtensorrtrtx/*.h"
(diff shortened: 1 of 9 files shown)


kunal-vaishnavi:
Closing this PR since the [other PR](https://github.com/microsoft/onnxruntime-genai/pull/2165) has been merged

PR #2381, opened 2026-08-05
Add bounds checking to prevent out-of-bounds reads in multi-image vision

apsonawane:
This pull request strengthens input validation for multi-image vision processing by adding defensive bounds checks to prevent out-of-bounds memory access when handling the `image_grid_thw` tensor. It also introduces new unit tests to ensure these checks are effective and to document the expected behavior in various edge cases.

**Input validation improvements:**

* Added a check in `GetImageFeatureBatchSize` (in `multi_modal.cpp`) to ensure the `image_grid_thw` tensor contains at least `num_images * 3` elements, throwing a runtime error if not.
* Added a similar bounds check in `QwenVisionState::Run` to verify that `image_grid_thw` has enough elements for all images before processing, with a descriptive error message if the check fails.

**Testing enhancements:**

* Added a new test file `qwen_multi_image_vision_test.cpp` with unit tests that verify the bounds checking logic, including tests for insufficient elements, improper tensor rank, valid cases, and ensuring all scatter loops are protected.

### DIFF
--- a/src/models/multi_modal.cpp
+++ b/src/models/multi_modal.cpp
@@ -204,7 +204,7 @@ DeviceSpan<float> QwenVisionState::Run(int current_length, DeviceSpan<int32_t>& 
 
   OrtValue* grid_full = inputs_[grid_idx];
   const int64_t* grid_data = grid_full->GetTensorData<int64_t>();
-  
+
   // Defensive bounds check: verify grid_data has enough elements for all images
   // Each image requires 3 elements (t, h, w)
   int64_t grid_elem_count = grid_full->GetTensorTypeAndShapeInfo()->GetElementCount();
--- a/src/models/multi_modal.cpp
+++ b/src/models/multi_modal.cpp
@@ -81,7 +81,16 @@ int64_t GetImageFeatureBatchSize(const std::vector<ExtraInput>& extra_inputs) {
       assert(extra_inputs[i].tensor->ort_tensor_);
       const auto shape = extra_inputs[i].tensor->ort_tensor_->GetTensorTypeAndShapeInfo()->GetShape();
       if (!shape.empty()) {
-        return shape[0];  // num_images
+        int64_t num_images = shape[0];
+        // Validate that the tensor contains at least num_images * 3 elements
+        // (t, h, w for each image)
+        int64_t elem_count = extra_inputs[i].tensor->ort_tensor_->GetTensorTypeAndShapeInfo()->GetElementCount();
+        if (elem_count < num_images * 3) {
+          throw std::runtime_error("image_grid_thw element count (" + std::to_string(elem_count) +
+                                   ") is less than required for " + std::to_string(num_images) +
+                                   " images (need at least " + std::to_string(num_images * 3) + ")");
+        }
+        return num_images;
       }
     }
   }
@@ -195,6 +204,15 @@ DeviceSpan<float> QwenVisionState::Run(int current_length, DeviceSpan<int32_t>& 
 
   OrtValue* grid_full = inputs_[grid_idx];
   const int64_t* grid_data = grid_full->GetTensorData<int64_t>();
+  
+  // Defensive bounds check: verify grid_data has enough elements for all images
+  // Each image requires 3 elements (t, h, w)
+  int64_t grid_elem_count = grid_full->GetTensorTypeAndShapeInfo()->GetElementCount();
+  if (grid_elem_count < num_images_ * 3) {
+    throw std::runtime_error("image_grid_thw has " + std::to_string(grid_elem_count) +
+                             " elements but need at least " + std::to_string(num_images_ * 3) +
+                             " for " + std::to_string(num_images_) + " images");
+  }
 
   // Check if the ONNX model accepts dynamic num_images.
   // A non-positive dim-0 (0 or -1) in the model's input shape = dynamic/symbolic.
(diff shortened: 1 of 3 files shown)


test/qwen_multi_image_vision_test.cpp
```diff
@@ -0,0 +1,132 @@
...
+  auto malformed_tensor = OgaTensor::Create(
+      malformed_grid.data(),
+      std::array<int64_t, 1>{2},
+      OgaElementType_int64);
+  harness.inputs->Set("image_grid_thw", *malformed_tensor);
+
+  EXPECT_THROW(harness.generator->SetInputs(*harness.inputs), std::runtime_error);
+}
+
+TEST(QwenVisionMultiImageTest, ValidGridShapeDoesNotTriggerGridBoundsValidation) {
```
kunal-vaishnavi: From Copilot:

This test can pass even when `SetInputs` throws an unrelated `std::runtime_error`, because it catches the exception and only checks that the error message does not contain the new bounds-check text. Please make the success path assert that no exception is thrown, for example:

```cpp
EXPECT_NO_THROW(harness.generator->SetInputs(*harness.inputs));
```

Alternatively, rethrow unexpected exceptions and assert the exact intended behavior. This ensures the test fails when valid input setup breaks for an unrelated reason.

src/models/multi_modal.cpp
```diff
@@ -81,7 +81,21 @@ int64_t GetImageFeatureBatchSize(const std::vector<ExtraInput>& extra_inputs) {
...
+        const int64_t num_images = shape[0];
+        if (num_images < 0) {
+          throw std::runtime_error("image_grid_thw num_images must be non-negative");
+        }
+
+        // Validate that the tensor contains at least one (t, h, w) triplet per image
+        // without multiplying num_images by 3, which could overflow int64_t.
+        const size_t elem_count = extra_inputs[i].tensor->ort_tensor_->GetTensorTypeAndShapeInfo()->GetElementCount();
+        const size_t expected_image_count = static_cast<size_t>(num_images);
+        if (elem_count / 3 < expected_image_count) {
```
kunal-vaishnavi: From Copilot:

The element-count check protects against out-of-bounds reads, but it does not verify that `image_grid_thw` has the expected rank and layout, such as `[num_images, 3]`. A tensor with a different shape can still contain enough elements to pass this check while being semantically invalid for the downstream indexing.

Could we either validate the rank and relevant dimensions explicitly at both validation sites, or document why the downstream code intentionally requires only the flat element-count invariant?

src/models/multi_modal.cpp
```diff
@@ -196,6 +210,20 @@ DeviceSpan<float> QwenVisionState::Run(int current_length, DeviceSpan<int32_t>&
...
 
+  // Defensive bounds check: verify grid_data has enough elements for all images
+  // Each image requires 3 elements (t, h, w)
+  if (num_images_ < 0) {
+    throw std::runtime_error("image_grid_thw num_images must be non-negative");
+  }
+
+  const size_t grid_elem_count = grid_full->GetTensorTypeAndShapeInfo()->GetElementCount();
+  const size_t expected_image_count = static_cast<size_t>(num_images_);
+  if (grid_elem_count / 3 < expected_image_count) {
```
kunal-vaishnavi: From Copilot:

The element-count check protects against out-of-bounds reads, but it does not verify that `image_grid_thw` has the expected rank and layout, such as `[num_images, 3]`. A tensor with a different shape can still contain enough elements to pass this check while being semantically invalid for the downstream indexing.

Could we either validate the rank and relevant dimensions explicitly at both validation sites, or document why the downstream code intentionally requires only the flat element-count invariant?

test/qwen_multi_image_vision_test.cpp
```diff
@@ -0,0 +1,132 @@
...
+  for (const auto& image_path : GetQwenVisionImagePaths()) {
+    if (!std::filesystem::exists(image_path)) {
+      GTEST_SKIP() << "Qwen vision test image not found at " << image_path;
+    }
+  }
+}
+
+}  // namespace
+
+TEST(QwenVisionMultiImageTest, RejectsInsufficientGridElementsViaGeneratorInputs) {
```
kunal-vaishnavi: From Copilot:

These tests only assert that `SetInputs` throws a `std::runtime_error`, so they could pass because of an unrelated model or input-shape validation failure rather than the new `image_grid_thw` validation. Please assert the expected error message or error category, and ensure the remaining inputs are valid enough for execution to reach the targeted check.

For example, verify that the exception identifies the `image_grid_thw` element-count or rank validation.

test/qwen_multi_image_vision_test.cpp
```diff
@@ -0,0 +1,63 @@
...
+    fn();
+  } catch (const std::exception& e) {
+    return e.what();
+  }
+  return {};
+}
+
+}  // namespace
+
+TEST(QwenVisionMultiImageTest, RejectsInsufficientGridElementsViaGeneratorInputs) {
```
kunal-vaishnavi: From Copilot:

These tests call `ValidateImageGridThwLayoutAndCount` directly, so they don't verify that either production call site actually invokes the validation. They would still pass if the checks were removed from `GetImageFeatureBatchSize` or `QwenVisionState::Run`.

Could you exercise the validation through the public generator/input path, using malformed `image_grid_thw` tensors and asserting the expected `std::runtime_error` (ideally checking the error message as well)? Please also add coverage that reaches the multi-image `QwenVisionState::Run` path, since the PR is intended to protect the scatter-loop reads there.

kunal-vaishnavi (approved):

PR #2383, opened 2026-08-05
Add validation for layer_types size in LFM2Cache constructor

apsonawane:
This pull request addresses a critical validation gap in the `LFM2Cache` constructor to prevent out-of-bounds memory access when parsing model configuration. It adds a runtime check to ensure the `layer_types` array size matches `num_hidden_layers`, and introduces regression tests to document and enforce this security property.

**Security validation and regression tests:**

* Added a runtime validation in `LFM2Cache::LFM2Cache` (in `kv_cache.cpp`) to throw a `std::runtime_error` if the size of `layer_types_` does not match `layer_count_`, preventing out-of-bounds reads and potential vulnerabilities.
* Introduced a new test file `lfm2_cache_layer_types_test.cpp` with regression tests for various configuration scenarios, including undersized, oversized, empty, and valid `layer_types` arrays, to ensure the constructor enforces the validation and documents the security property (CWE-125, CWE-1284).

### DIFF
--- a/test/lfm2_cache_layer_types_test.cpp
+++ b/test/lfm2_cache_layer_types_test.cpp
@@ -4,6 +4,7 @@
 #include <filesystem>
 #include <fstream>
 #include <sstream>
+#include <stdexcept>
 #include <string>
 
 #include <gtest/gtest.h>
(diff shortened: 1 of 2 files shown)


test/lfm2_cache_layer_types_test.cpp
```diff
@@ -0,0 +1,134 @@
...
+  SkipIfModelUnavailable();
+
+  const auto model_dir = WriteModelWithLayerTypes("empty", "[]");
+  const std::string message = CaptureThrowMessage(model_dir);
+
+  EXPECT_FALSE(message.empty());
+  EXPECT_NE(message.find("layer_types"), std::string::npos) << message;
+}
+
+TEST(LFM2CacheLayerTypesValidationTest, ValidMatchingLayers) {
```
kunal-vaishnavi, line 128: From Copilot:

This test does not actually verify the valid path. `CaptureThrowMessage()` returns an empty string both when model/generator creation succeeds and when no exception is captured, so all three assertions can pass without proving that `OgaGenerator::Create` completed successfully.

Could you make the helper return success separately from the exception message, or assert construction directly with `EXPECT_NO_THROW`? For example:

```cpp
EXPECT_NO_THROW({
  auto model = OgaModel::Create(model_dir.string().c_str());
});
```

This ensures the regression test fails if the valid configuration is rejected—or if the test setup fails before reaching `LFM2Cache`.

test/lfm2_cache_layer_types_test.cpp
```diff
@@ -0,0 +1,134 @@
...
+
+  std::string config = ReadFile(dst_dir / "genai_config.json");
+  ReplaceFirst(config,
+               "\"layer_types\": [\"conv\", \"full_attention\", \"conv\", \"full_attention\"]",
+               "\"layer_types\": " + layer_types_json);
+  WriteFile(dst_dir / "genai_config.json", config);
+  return dst_dir;
+}
+
+std::string CaptureThrowMessage(const fs_std::path& model_dir) {
```
kunal-vaishnavi: From Copilot:

Could we preserve and assert the exception type here instead of catching every `std::exception` and returning only `what()`? As written, an unrelated model-loading, filesystem, or setup failure could produce a non-empty message containing one of the checked substrings and make the invalid-input tests pass.

Please assert that the malformed configuration reaches `LFM2Cache` and throws `std::runtime_error`, then verify the complete validation message includes the `LFM2Cache` prefix, the actual `layer_types` size, and `num_hidden_layers`. This will ensure the regression tests specifically cover the intended validation rather than just any failure during model creation.

test/lfm2_cache_layer_types_test.cpp
```diff
@@ -0,0 +1,134 @@
...
+  const auto model_dir = WriteModelWithLayerTypes("undersized", "[\"conv\"]");
+  const std::string message = CaptureThrowMessage(model_dir);
+
+  EXPECT_FALSE(message.empty());
+  EXPECT_NE(message.find("LFM2Cache"), std::string::npos) << message;
+  EXPECT_NE(message.find("layer_types"), std::string::npos) << message;
+  EXPECT_NE(message.find("num_hidden_layers"), std::string::npos) << message;
+}
+
+TEST(LFM2CacheLayerTypesValidationTest, OversizedLayerTypesArray) {
```
kunal-vaishnavi, line 110: From Copilot:

Could these tests assert the full validation context consistently with the undersized-array case? Both the oversized and empty-array tests currently check only for a non-empty message containing `layer_types`, so an unrelated model-loading or setup failure could make them pass.

Please assert that the exception message includes `LFM2Cache`, `layer_types`, and `num_hidden_layers`—and, ideally, the expected array size and layer count. This will confirm that both cases are failing specifically because of the new `LFM2Cache` size validation.

kunal-vaishnavi (approved):

PR #2393, opened 2026-08-10
Optimize paged-attention logits memory usage

GopalakrishnanN:
## Summary
- gather each sequence's final packed hidden state before the LM head for paged-attention models
- change paged logits from `[num_tokens, vocab_size]` to `[batch_size, vocab_size]`
- update `VarlenDecoderIO` to allocate and consume one logits row per request
- retain compatibility with legacy paged models that still return per-token logits
- document the output contract and add an execution-level ragged-sequence test

## Validation
- Visual Studio 18 Release native build succeeded
- `UnitTests` and `EngineUnitTests` passed (61 engine tests)
- Python model-builder suite: 325 passed, 3 skipped
- Full CTest additionally ran `ReInitTests`, which segfaulted while probing unavailable TRT-RTX/CUDA DLLs; the unit and engine targets passed

## Memory impact
CUDA I/O-binding benchmark on an NVIDIA RTX A1000, FP16, batch size 4, vocabulary size 131,072; medians of three isolated runs:

| Packed prefill tokens | Baseline peak | Optimized peak | Reduction |
|---:|---:|---:|---:|
| 512 | 228.7 MiB | 102.7 MiB | 126 MiB (55%) |
| 1,024 | 356.7 MiB | 102.7 MiB | 254 MiB (71%) |
| 2,048 | 612.7 MiB | 100.7 MiB | 512 MiB (84%) |
| 4,096 | 1,124.7 MiB | 100.7 MiB | 1,024 MiB (91%) |

The benchmark isolates the changed LM-head path; model weights and KV-cache memory are unchanged.


GopalakrishnanN:
## End-to-end peak GPU memory benchmark

I also measured the full continuous-batching `Engine` path with a real model, in addition to the isolated LM-head benchmark in the PR description.

### Model

- `Qwen/Qwen2.5-7B-Instruct`
- INT4 weights, CUDA paged attention (`-p int4 -e cuda --extra_options use_paged_attention=true`)
- Vocabulary size: 152,064
- Both variants were regenerated from the same Hugging Face snapshot with identical external weights, tokenizer, and semantic `genai_config.json`.
- Baseline graph output: `logits[num_tokens, 152064]`
- PR graph output: `logits[batch_size, 152064]`, with the final-state `Gather` before the LM head

### Experiment

- GPU: NVIDIA A100-SXM4-80GB; driver 580.105.08
- Build: RelWithDebInfo, CUDA 12.8, ONNX Runtime 1.29.0, SM80
- Baseline: current `main` at `7179de88`
- PR variant: the same baseline plus the #2393 optimization. The measured production/model-builder files are identical to current PR head `8eb838cc`; later differences are workflow/test-only.
- Fixed paged KV cache: 32 blocks x 256 tokens = 448 MiB
- One prefill step per fresh process; `max_scheduled_tokens` was set to the packed-token count
- Per-process GPU memory sampled through NVML every 5 ms
- 10 isolated runs per variant/workload, alternating run order; values below are medians
- Default CUDA EP arena growth strategy was retained

### Results

| Packed prompt tokens | Baseline peak | PR peak | Reduction |
|---:|---:|---:|---:|
| 512 | 7,624 MiB | 7,128 MiB | 496 MiB (6.5%) |
| 1,024 | 7,628 MiB | 7,632 MiB | within allocator noise |
| 2,048 | 8,138 MiB | 7,110 MiB | **1,028 MiB (12.6%)** |
| 2,048, ragged batch 8 (`32,64,96,128,256,384,480,608`) | 8,138 MiB | 7,250 MiB | **888 MiB (10.9%)** |
| 4,096 | 9,155 MiB | 7,490 MiB | **1,665 MiB (18.2%)** |

The 2,048-token case matches the engine's default scheduled-token budget and is the most representative headline result: **about 1.0 GiB lower process peak GPU memory**.

The default ORT CUDA arena grows in power-of-two buckets, which produces discrete approximately 1 GiB reservation jumps and masks the 1,024-token case. As a supplementary check, using `arena_extend_strategy=1` (exact requested growth), five fresh-process runs measured 7,616 MiB baseline versus 7,368 MiB PR at 1,024 tokens, a 248 MiB reduction.

A compatibility control using the same legacy `[num_tokens, vocab]` model with both runtimes measured 8,158 MiB on baseline versus 8,134 MiB on the PR runtime (24 MiB difference). This confirms that the material saving comes from regenerating the model with the new `[batch_size, vocab]` output contract, not from the runtime change alone. Host RSS stayed near 1.5 GiB for both variants.

kunal-vaishnavi (dismissed):

src/python/py/models/builder.py
```diff
@@ -720,9 +720,11 @@ def get_args():
                     flattened token axis (`input_ids` becomes 1D), stores the KV-cache in paged
                     [num_blocks, block_size, num_kv_heads, head_size] buffers, and removes the `attention_mask` and
                     `position_ids` inputs in favor of the `block_table`, `cumulative_sequence_lengths`, and
-                    `past_sequence_lengths` metadata inputs. An `engine` section (block_size, gpu_utilization_factor,
-                    max_batch_size) is added to genai_config.json. Currently only supported for the CUDA execution
-                    provider with fp16 or bf16 precision. Cannot be combined with exclude_embeds, exclude_lm_head, or prune_lm_head.
+                    `past_sequence_lengths` metadata inputs. Before the LM head, selects the final packed hidden state
+                    for each sequence so the model outputs [batch_size, vocab_size] logits instead of projecting
+                    every prefill token. An `engine` section (block_size, gpu_utilization_factor, max_batch_size) is
+                    added to genai_config.json. Currently only supported for the CUDA execution provider with fp16 or
```
baijumeswani: Reference to the engine section's block_size and gpu_utilization_factor do not seem necessary at this point. Was there a reason to add it here?

src/python/py/models/builders/base.py
```diff
@@ -4414,7 +4415,41 @@ def make_lm_head(self, lm_head):
         # Sequence dimension for shape annotations ("sequence_length" normally, 1 when pruned)
         seq_dim = "sequence_length"
 
-        if self.prune_lm_head:
+        if self.use_paged_attention:
```
baijumeswani: While this change makes sense, let's fold this into `self.prune_lm_head` and leave the option to build without pruning the lm head.

We can also enable prune_lm_head by default (https://github.com/microsoft/onnxruntime-genai/blob/f4fc76303b2d776e7c8c54f198af20118413ab71/src/python/py/models/builder.py#L712) in the model builder with these changes.

cc @kunal-vaishnavi

src/engine/decoders/varlen_decoder_io.cpp
```diff
@@ -48,6 +50,14 @@ VarlenDecoderIO::VarlenDecoderIO(std::shared_ptr<DecoderOnly_Model> model,
     : DecoderIO(model, scheduled_requests, cache_manager),
       graph_buffers_{graph_buffers},
       execution_context_{execution_context} {
+  // Older paged-attention models expose one row per packed token. Current models identify their
+  // already-selected per-request rows with the symbolic batch_size dimension.
```
baijumeswani: Avoid references to old/new/current/legacy since what is old/new is relative. Instead, you can document the expectation of the output shape in the code.
Please update other areas where such comments appear as well.

src/python/py/models/builders/base.py
```diff
@@ -4411,10 +4414,44 @@ def make_lm_head(self, lm_head):
         matmul_basename = f"{basename}/MatMul"
         root_input = self.layernorm_attrs["output_0"]
 
-        # Sequence dimension for shape annotations ("sequence_length" normally, 1 when pruned)
+        # Sequence dimension used for LM-head shape annotations.
         seq_dim = "sequence_length"
 
-        if self.prune_lm_head:
+        if self.use_paged_attention and self.prune_lm_head:
```
baijumeswani, line 4420: Just want to confirm that building a paged attention model with `prune_lm_head=False` would also work, right?

baijumeswani (dismissed):

src/python/py/models/builder.py
GopalakrishnanN: Removed the engine-section reference from the paged-attention help and the matching README wording. The text now focuses on the pruning-controlled logits contract.

src/engine/decoders/varlen_decoder_io.cpp
GopalakrishnanN: Replaced relative old/current/legacy wording throughout the changed runtime comments with explicit per-request `[batch_size, vocab_size]` and per-token `[total_num_tokens, vocab_size]` contracts.

src/python/py/models/builders/base.py
GopalakrishnanN: Folded the paged Gather path under `self.prune_lm_head`, enabled the option by default, and retained `prune_lm_head=false` as the `[num_tokens, vocab_size]` path. Added coverage for both option values, output-shape initialization, and graph execution.

src/python/py/models/builders/base.py
GopalakrishnanN: Yes. With `prune_lm_head=false`, the paged path skips the final-state Gather, projects all packed hidden states, and declares logits as `[num_tokens, vocab_size]`; the engine consumes that per-token contract. The added tests exercise both option parsing and graph execution for this path.

baijumeswani:
AI suggested one more change:

> Add an Engine-level test for `[batch_size,vocab_size]`. Detection depends on the exact symbolic name `batch_size`, but the synthetic Engine model still outputs `num_tokens`. Add a ragged-prefill fixture exercising allocation, row mapping, and FP16 conversion for the new path.

Could you please update the synthetic model that has been checked in that runs all tests?

GopalakrishnanN:
> Add an Engine-level test for `[batch_size,vocab_size]`. Detection depends on the exact symbolic name `batch_size`, but the synthetic Engine model still outputs `num_tokens`. Add a ragged-prefill fixture exercising allocation, row mapping, and FP16 conversion for the new path.

Updated in eb03eddd. The checked-in synthetic paged model now emits FP16 `[batch_size, vocab_size]` logits by gathering each request's final packed row. The existing simultaneous test runs ragged prompts of lengths 3 and 4 through `og.Engine`, covering symbolic-shape detection, batch-sized output allocation, request-row mapping, and FP16-to-FP32 conversion; a separate assertion pins the model output type and symbolic shape.

baijumeswani (dismissed):

src/python/py/models/builders/base.py
```diff
@@ -459,31 +459,34 @@ def make_outputs_init(self):
...
-        if self.use_paged_attention:
-            self.output_shapes["present.key"] = ["num_blocks", "block_size", self.num_kv_heads, self.head_size]
-            self.output_shapes["present.value"] = ["num_blocks", "block_size", self.num_kv_heads, self.head_size]
-            self.output_shapes["hidden_states"] = ["num_tokens", self.hidden_size]
-            self.output_shapes["logits"] = ["num_tokens", self.vocab_size]
-
         self.exclude_lm_head = self.extra_options.get("exclude_lm_head", False)
         self.include_hidden_states = self.extra_options.get("include_hidden_states", False)
-        self.prune_lm_head = self.extra_options.get("prune_lm_head", False)
+        self.prune_lm_head = self.extra_options.get("prune_lm_head", True)
```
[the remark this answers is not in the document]
baijumeswani: This is intentional.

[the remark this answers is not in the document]
baijumeswani: Will update the static batching engine to work with pruned logits.

baijumeswani (approved):

PR #2396, opened 2026-08-11
Add continuous batching Engine integration tests

baijumeswani:
- Add deterministic synthetic tests for Engine batching, scheduling, lifecycle, and paged-cache behavior.
- Add Linux CUDA integration tests using a real paged-attention Qwen model.
- Wire the tests into existing CI pipelines and document the Engine testing strategy.
- Remove obsolete C API test coverage.


test/python/integration/test_integration_engine.py
```diff
@@ -0,0 +1,351 @@
...
+        pytest.skip("CUDA execution provider is not available in this build.")
+
+    config = og.Config(str(paged_model_path))
+    config.clear_providers()
+    config.append_provider("cuda")
+
+    model = og.Model(config)
+    tokenizer = og.Tokenizer(model)
+    cfg = json.loads((paged_model_path / "genai_config.json").read_text())
+    return _Bundle(model=model, tokenizer=tokenizer, config=cfg, device=device)
```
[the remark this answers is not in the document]
baijumeswani: these tests are only for cuda ep. try catch is not needed.

.pipelines/stages/integration-stage.yml
```diff
@@ -125,6 +129,19 @@ stages:
             ep: 'cuda'
             model: ${{ model }}
 
+- ${{ if and(eq(parameters.build_linux_x64, true), eq(parameters.linux_x64_engine, true)) }}:
+  - stage: integration_engine_test_linux_x64
```
kunal-vaishnavi, line 133: From Copilot:

The new Engine stage (`integration_engine_test_linux_x64`) is correctly gated by `linux_x64_engine` parameter in `.pipelines/integration-stage.yml` (lines 132–143), and the parameter is properly threaded through `.pipelines/integration-tests.yml` (lines 90–93, 171). 

However, verify the following before merging:

1. **Default enablement**: The `linux_x64_engine` parameter defaults to `true` (line 93). Confirm this is intentional for PR runs, or consider defaulting to `false` if the stage should only run on main merges to avoid CI latency on every PR.

2. **Model pinning verification**: The stage hardcodes `model: 'qwen2.5-0.5b-instruct-paged'` and `storage_model: 'qwen2.5-0.5b-instruct'` (lines 142–143). Ensure:
   - The `qwen2.5-0.5b-instruct-paged` model exists in the `foundrylocalmodels` storage account and that SHA256 identity hashes in `models.py` are pinned to this exact artifact.
   - The `resolver.py` logic can resolve both the `storage_model` (for catalog lookup) and the final `model` parameter correctly without conflicts.

3. **job template alignment**: The new `.pipelines/stages/jobs/integration-engine-test-job.yml` is referenced at line 137 but is not included in this review. Confirm it correctly:
   - Downloads the model via managed identity to the agent's `--model-root`.
   - Runs pytest with `--run-engine-tests` and `--execution-provider cuda` flags.
   - Inherits parameter passing from the stage template.

test/python/integration/test_integration_engine.py
```diff
@@ -0,0 +1,351 @@
...
+        pytest.skip(f"No content identity recorded for '{_MODEL_ID}'.")
+    for filename, expected in identity.items():
+        path = paged_model_path / filename
+        assert path.exists(), f"pinned artifact is missing '{filename}' at {paged_model_path}"
+        digest = hashlib.sha256()
+        with path.open("rb") as model_file:
+            for chunk in iter(lambda: model_file.read(8 * 1024 * 1024), b""):
+                digest.update(chunk)
+        actual = digest.hexdigest()
+        assert actual == expected, (
```
kunal-vaishnavi, line 150: From Copilot:

The error message on hash mismatch is helpful but could provide more actionable debugging info. Consider enhancing it to show the current pinned version and how to locate the expected hashes:

```python
assert actual == expected, (
    f"'{_MODEL_ID}' {filename} sha256 {actual} != pinned {expected}. "
    f"The artifact at the pinned v{models.pinned_version(_MODEL_ID)} differs from the catalog. "
    f"Current PINNED_VERSIONS: {models.PINNED_VERSIONS}. "
    f"If re-exporting, update both PINNED_VERSIONS and PINNED_IDENTITY in test/python/integration/models.py "
    f"together and record the actual sha256 hash above."
)
```

test/python/integration/test_integration_engine.py
```diff
@@ -0,0 +1,351 @@
...
+_MODEL_ID = "qwen2.5-0.5b-instruct-paged"
+
+_PROMPTS = [
+    "The capital of France is",
+    "Water is made of hydrogen and",
+    "The opposite of hot is",
+    "One plus one equals",
+]
+
+_MAX_STEPS = 20_000
```
kunal-vaishnavi, line 31: From Copilot:

The `_MAX_STEPS` limit is 2x the synthetic tests' bound (20,000 vs. 10,000). Add a comment explaining the higher threshold to help future maintainers tune it if needed:

```python
# Higher bound for real-model tests: Qwen 0.5B may take more steps due to
# CUDA kernel dispatch overhead and actual inference latency. Regression in
# model speed will still be caught if most tests exceed this. If tests
# consistently timeout, lower this bound; if timeouts are false positives,
# increase selectively per test via _run(engine, max_steps=N).
_MAX_STEPS = 20_000
```

Alternatively, consider adding per-test configuration if some tests are expected to be significantly slower:

```python
_MAX_STEPS_DEFAULT = 20_000
_MAX_STEPS_LONG_HORIZON = 30_000  # e.g., test_eos_gates_termination uses headroom=200
```

kunal-vaishnavi (dismissed):

.pipelines/stages/integration-stage.yml
baijumeswani: 1. Yes, this is intentional.
2. Yes, model exists in the storage account.
3. Yes, the e2e test runs.

test/python/integration/test_integration_engine.py
baijumeswani: Added a small wording fix.

test/python/integration/test_integration_engine.py
baijumeswani: _MAX_STEPS counts engine.step(). it is not related to latency. It is a large number so that we can simulate the non-termination scenario.

tianleiwu (approved):

PR #2398, opened 2026-08-12
Fix package feed isolation

jiafatom:
## Summary
- remove an unused CMake external project that downloaded a NuGet executable
- pass the configured Python package index into the ARM64 manylinux image build using a BuildKit secret
- keep package credentials out of Docker build arguments and image history
- correct a malformed-grid test fixture so it reaches the intended layout validation

## Testing
- `git diff --check`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -330,8 +330,9 @@ if(USE_DML)
   get_filename_component(PACKAGES_DIR ${CMAKE_CURRENT_BINARY_DIR}/_deps ABSOLUTE)
   set(DXC_PACKAGE_DIR ${PACKAGES_DIR}/Microsoft.Direct3D.DXC.1.7.2308.12)
   set(PACKAGES_CONFIG ${PROJECT_SOURCE_DIR}/packages.config)
+  find_program(NUGET_PATH NAMES nuget nuget.exe REQUIRED)
   set(NUGET_RESTORE_COMMAND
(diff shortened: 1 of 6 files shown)


CMakeLists.txt
```diff
@@ -330,8 +330,9 @@ if(USE_DML)
   get_filename_component(PACKAGES_DIR ${CMAKE_CURRENT_BINARY_DIR}/_deps ABSOLUTE)
   set(DXC_PACKAGE_DIR ${PACKAGES_DIR}/Microsoft.Direct3D.DXC.1.7.2308.12)
   set(PACKAGES_CONFIG ${PROJECT_SOURCE_DIR}/packages.config)
+  find_program(NUGET_PATH NAMES nuget nuget.exe REQUIRED)
   set(NUGET_RESTORE_COMMAND
-    ${CMAKE_CURRENT_BINARY_DIR}/nuget/src/nuget restore
+    ${NUGET_PATH} restore
```
sanaa-hamel-microsoft: nit: always quote non-array expansions

sanaa-hamel-microsoft (changes requested):

CMakeLists.txt
jiafatom: Quoted the NuGet path and the other scalar command arguments in a9fd3825.

sanaa-hamel-microsoft (approved):

baijumeswani (approved):

PR #2400, opened 2026-08-12
Fix malformed multi-image grid test input

apsonawane:
This pull request updates the `QwenVisionMultiImageTest` to improve the test case for rejecting malformed image grids. The main change is to the construction of the malformed grid, making the test more accurate in simulating an invalid input scenario.

**Test improvements:**

* Updated the malformed grid in `TEST(QwenVisionMultiImageTest, RejectsMalformedGridInMultiImageVisionRun)` to use a `[3, 2]` layout with 6 values instead of a `[2, 4]` layout with 8 values, ensuring the test case better reflects an invalid input that should be rejected by `QwenVisionState::Run`.
* Clarified comments to accurately describe the grid layout and the reason for rejection in the test.


apsonawane:
Not needed

PR #2401, opened 2026-08-12
Add decode-first token budgeting to dynamic batching

baijumeswani:
## Summary

Adds global token budgeting and decode-first scheduling to the continuous-batching Engine.

Dynamic batching now:

- limits each model step with `max_scheduled_tokens`;
- uses `2048` as the default token budget;
- schedules one-token decode work before prefills;
- allocates the remaining budget to bounded prefill chunks;
- treats `RequestStepPlan` as the authoritative source of scheduled token counts;
- supports executing a selected subset of paged-cache residents without disturbing omitted requests.

`search.chunk_size` remains the per-request prefill limit, while `max_scheduled_tokens` limits the total work across the batch.

## Scheduling flow

```text
┌──────────────────────────────┐
│ Snapshot runnable requests   │
│                              │
│ • Existing cache residents   │
│ • Newly assigned requests    │
└──────────────┬───────────────┘
               │
               ▼
┌──────────────────────────────┐
│ Build scheduling candidates  │
│                              │
│ • Decode:  1 provisional tok │
│ • Prefill: 1 provisional tok │
└──────────────┬───────────────┘
               │
               ▼
┌──────────────────────────────┐
│ Apply decode-first ordering  │
│                              │
│ 1. Decodes, stable order     │
│ 2. Prefills, stable order    │
└──────────────┬───────────────┘
               │
               ▼
┌──────────────────────────────┐
│ Check paged-cache feasibility│
│                              │
│ • Skip work that cannot fit  │
│ • Continue to later requests │
│ • Preserve omitted residents │
└──────────────┬───────────────┘
               │
               ▼
┌──────────────────────────────┐
│ Allocate the token budget    │
│                              │
│ • 1 token per decode         │
│ • Remaining tokens to        │
│   selected prefills          │
│ • Respect chunk_size         │
│ • Respect max batch size     │
└──────────────┬───────────────┘
               │
               ▼
┌──────────────────────────────┐
│ Finalize RequestStepPlan     │
│                              │
│ • Selected token counts      │
│ • Packed input offsets       │
│ • Cache target slots         │
│ • Graph-capture eligibility  │
└──────────────┬───────────────┘
               │
               ▼
┌──────────────────────────────┐
│ Execute transaction          │
│                              │
│ Reserve → Decode → Sample    │
│          → Commit            │
└──────────────────────────────┘
```

Using one provisional token during cache planning allows an infeasible request to be skipped without consuming the global token budget or hiding later work that can run. Selected prefills are expanded only after cache feasibility is known.

## Configuration

```json
{
  "engine": {
    "dynamic_batching": {
      "max_scheduled_tokens": 2048
    }
  }
}
```

`max_scheduled_tokens` must be a positive integer.

## Cache behavior

- Resident requests may be omitted from an execution step while retaining their committed block tables.
- Omitted residents continue to count toward the configured residency limit.
- Newly admitted prefills reserve blocks for the complete prompt, even when the global token budget processes only part of that prompt.
- Cache planning remains responsible for feasibility rather than scheduling priority.

## Scope

This PR implements stateless decode-first scheduling only. Fairness, decode rotation, and bounded prefill promotion will follow separately because service must be recorded only after a transaction commits successfully.

## Testing

Added coverage for:

- configuration defaults and validation;
- decode-first ordering and exact budget accounting;
- per-request and global prefill limits;
- decode demand exhausting the budget;
- blocked-request backfill;
- selected resident subsets;
- whole-prompt cache reservation;
- graph-capture eligibility;
- plan-owned token counts;
- transactional commit and rollback;
- static batching regression behavior.

All 126 Engine unit tests pass.


src/engine/scheduler.cpp
```diff
@@ -199,47 +176,91 @@ StepPlanningResult DynamicBatchScheduler::PlanStep(StepPlan& plan) {
...
+  plan.scheduled_request_limit = DecodeFirstProvisionalRequestLimit(
+      dynamic_batching.max_scheduled_tokens,
+      dynamic_batching.max_batch_size);
+
+  auto result = cache_manager_->PlanStepResources(plan);
   if (!result.executable) {
     return result;
   }
 
+  std::vector<DecodeFirstBudgetCandidate> selected_candidates;
```
tianleiwu, line 231: `PlanStep()` runs once per generated-token batch, but this change creates fresh `candidates`, `budget_candidates`, ordering, and `selected_candidates` vectors on every call, then performs two `find_if` scans over all candidates for each selected request. That adds avoidable heap churn and O(MN) host work on a latency-sensitive path as `max_batch_size` grows. Could we keep reusable scratch buffers on `DynamicBatchScheduler` (reserved to the configured batch size) and carry the selected candidate metadata/index through cache compaction so finalization stays linear without repeated lookups?

tianleiwu (commented):
No correctness blockers found. The decode-first budgeting, cache-selection contracts, and plan-owned token counts are coherent, and all 49 tests in the five changed Engine suites pass after rebuilding the test target. I left one non-blocking performance suggestion on the per-step scheduler path.

src/engine/scheduler.cpp
baijumeswani: This is a valid concern. But for local on-device scenarios, the candidates are a handful (maybe between 1-32) and as such the complexity is not as bad.
AI seems to think adding extra bookkeeping for the local scenario is not worth it right now, unless profiling results show an opportunity to be better.

tianleiwu (approved):

PR #2407, opened 2026-08-13
Recover from model execution capacity failures

baijumeswani:
Classifies confirmed ONNX Runtime execution-memory exhaustion as a recoverable Engine outcome.

When an ORT BFC arena allocation fails during model execution, the Engine now:

- reports `ExecutionCapacityExceeded`;
- preserves the original operator and allocation error;
- rolls back request and paged-cache transaction state;
- remains healthy so the workload can be retried;
- continues treating unrelated execution failures as fatal.

Concurrent-prefill stress testing reproduced a QMoE workspace allocation failure:

```text
BFCArena::AllocateRawInternal
Failed to allocate memory for requested buffer of size 2443129088
```

Previously, this exception followed the unknown execution-failure path and permanently marked the Engine unhealthy, even though allocation pressure is recoverable after memory becomes available or a smaller plan is selected.

Error flow

```text
 ORT/QMoE allocation failure
             │
             ▼
 DecoderModelExecutor classifies
 confirmed allocator exhaustion
             │
             ▼
 ModelExecutionError
 CapacityExceeded
             │
             ▼
 Engine synchronizes and rolls back
 request state + cache reservation
             │
             ▼
 ExecutionCapacityExceeded
 Engine remains healthy
```

Layer responsibilities

 - QMoE and ORT remain responsible for workspace sizing, allocation cleanup, and propagating the failure.
 -  ModelExecutor  translates the confirmed ORT allocation failure into an Engine execution category.
 -  Engine  owns transactional rollback and health decisions.
 - Scheduler-driven smaller-plan retry is intentionally deferred to a follow-up change.


src/engine/step_plan.h
```diff
@@ -23,12 +23,14 @@ enum class StepOutcomeKind {
   UnserviceableRequest,
   Committed,
   RetryableBatchAbort,
+  ExecutionCapacityExceeded,
```
tianleiwu, line 26: `ExecutionCapacityExceeded` is only visible on the internal `EngineStepError`. It does not survive the public API boundary: `OgaEngineStep` returns through the generic `OGA_CATCH` macro, which builds an `OgaResult` holding only `e.what()`, and `OgaCheckResult` in `ort_genai.h` turns that into a plain `std::runtime_error`. The pybind11 `Engine.step` binding inherits the same behavior.

The net effect is that a caller cannot tell a recoverable capacity failure from a fatal one without string-matching the message. That makes the recovery contract hard to use correctly. A naive retry loop that catches and continues will spin forever on a genuinely fatal error, since `Step()` rethrows the stored fatal error on every subsequent call, while a conservative caller has to treat every failure as fatal and discard a batch that was actually recoverable.

Could the outcome kind be surfaced additively, for example an accessor on `OgaResult` that exposes a stable outcome/error code, and then mapped into the C++ and Python wrappers? I am fine with that being a follow-up if you want to keep this PR focused, but until it exists the "remains healthy so the workload can be retried" contract is not reachable from outside the engine.

src/engine/model_executor.cpp
```diff
@@ -4,12 +4,21 @@
...
 
+#include <string_view>
 #include <typeinfo>
 
 namespace Generators {
 
 namespace {
 
+ExecutionFailureKind ClassifyOrtExecutionFailure(std::string_view message) {
+  if (message.find("Failed to allocate memory for requested buffer") !=
```
tianleiwu, line 16: This substring match is the only thing deciding whether an execution failure is recoverable or fatal, and nothing exercises it. The new engine test injects a `ModelExecutionError{CapacityExceeded, ...}` directly from the test double, so it bypasses this function entirely. If this literal were typo'd, or ORT reworded the message, the classifier would silently degrade to `Unknown` (fatal) and every test added in this PR would still pass.

Two focused cases on the observable behavior would close the gap: one asserting a realistic BFC allocation message is translated to a `ModelExecutionError` with `CapacityExceeded`, and one asserting an unrelated `Ort::Exception` still propagates unchanged. Since the helper is in an anonymous namespace, driving it through `DecoderModelExecutor::Decode` with a throwing decoder double keeps it internal.

Separately, a one-line comment noting that this string is coupled to ORT's BFC arena wording would make the coupling discoverable if it ever breaks.

tianleiwu (commented):
### Summary

The rollback path here looks correct and well scoped. `Engine::StepDynamic()` reuses the existing transaction boundary, so a capacity failure restores request/search state and releases the paged-cache reservation before surfacing the new outcome, and the engine is only marked unhealthy when rollback itself fails. Classifying a known ORT allocation failure separately from an unknown one at the execution boundary is the right place to draw that line, and wrapping it while preserving the original ORT diagnostic keeps the failure debuggable. This is a strict improvement over the previous behavior, where an allocation failure permanently poisoned the engine.

Verified locally at `5f65b88`: `engine_unit_tests` builds clean and all 15 `EngineStepTest.*` cases pass, including the new `ExecutionCapacityFailureRollsBackWithoutPoisoningEngine`.

Two things I would like to see addressed, details inline:

1. The new `ExecutionCapacityExceeded` outcome is not observable by any caller outside the internal C++ engine, so the recovery contract described in the PR body cannot actually be implemented by an API consumer today.
2. The new ORT message classifier, which is the one piece of production logic deciding whether a failure is recoverable, has no direct test coverage.

### Documentation (file not in this diff, so noting it here)

`docs/paged_attention_engine.md` is the living architecture doc for this subsystem and still describes the pre-change contract:

- The rollback section states that after a successful rollback "the caller receives an `EngineStepError` with `RetryableBatchAbort`", which is no longer true for capacity failures.
- The step-outcome table does not list `ExecutionCapacityExceeded`.

Worth updating here so the doc and the enum stay in sync, and clarifying that the engine stays healthy but retrying an unchanged plan is not guaranteed to succeed until memory availability or workload composition changes.

src/engine/step_plan.h
baijumeswani: I'll mark this as a follow-up item. In general the API surface needs to be enhanced for the request and engine contracts. Will get to this as part of that work.

tianleiwu (approved):
LGTM

PR #2410, opened 2026-08-13
Add a native Gemma 3n multimodal processor

NishantN2005:
### Description

Adds `Gemma3nMultiModalProcessor`, a native processor for Gemma 3n multimodal packages, and wires `gemma3n` / `gemma3n_text` into the model-type lists.

Before this, a Gemma 3n package had to declare model type `gemma3` and borrow Gemma 3's processor. That does not actually work. HF's `Gemma3nProcessor` expands `self.image_token` (`<image_soft_token>`) into the soft-token sequence, while `Gemma3Processor` expands `boi_token` (`<start_of_image>`). Gemma 3n ships a chat template that emits the former, so the only way to run one was to hand-patch `chat_template.jinja` to emit Gemma 3's marker instead.

`Gemma3nMultiModalProcessor` expands the placeholder the way HF does:

```
"\n\n" + <start_of_image> + 256x<image_soft_token> + <end_of_image> + "\n\n"
```

and the audio equivalent with 188 soft tokens. Both counts are fixed rather than derived from the input:

* the vision tower always sees one 768x768 image and emits a 16x16 grid;
* the audio encoder pads short clips with the audio padding embedding and truncates long ones, so its output is always `audio_soft_tokens_per_image` rows regardless of clip duration (HF `Gemma3nModel.forward`'s `extra_padding_features` step).

Both spellings of the placeholder are accepted. Gemma 3n's own template emits the soft token, but a Gemma 3 style template or a hand-written prompt emits the begin marker, and both should keep working. A prompt with no placeholder at all gets one prepended per input. `token_type_ids` follows HF: image 1, audio 3, text 0 (Gemma 3 only has 0/1).

`gemma3n` joins the MMM list rather than VLM since it carries both a vision and an audio tower; `gemma3n_text` joins the LLM list for text-only packages.

### The `audio_sizes` fix (second commit)

Worth calling out separately, because it is a contract that is easy to miss when adding a new multimodal processor.

`audio_sizes` is not a session input. It is how `MultiModalPipelineState` learns there is audio at all: `GetNumAudioTokens` sums it, and that sum both gates `SpeechState::Run` and sizes the pre-allocated `audio_features` buffer. If a processor does not emit it, `num_audio_tokens_` stays 0, the speech encoder silently never runs, and the embedding model is handed an empty `audio_features` while the prompt still carries 188 placeholders.

The failure mode depends on the EP, which is what made it hard to spot:

* on the CPU EP it raises `indices element out of data bounds` from the embedding `Gather`;
* on EPs that clamp out-of-range indices instead of raising, it produces plausible-but-wrong text with no error at all.

`num_image_tokens` gets the same one-row-per-input treatment for the same reason: `GetNumImageTokens` sums it to size `image_features`, so a single element under-counts a multi-image prompt by `(num_images - 1) * 256`.

The second commit also defaults `audio_features_type_`, which was left uninitialized for a vision-only package, and rejects unexpected tensor ranks from the image processor and feature extractor rather than reading past the shape array.

### Testing

Built `--config Release` on Windows x64 against `main`; `clang-format` 20.1.0 clean on both new files.

Verified end-to-end against a Gemma 3n E4B ONNX package on the ORT CPU EP:

* **audio** — `jfk.flac` transcribes exactly: *"And so, my fellow Americans. Ask not what your country can do for you. Ask what you can do for your country."* (this is the case that failed before the `audio_sizes` fix)
* **image** — single-image caption: *"A relaxed gray tabby cat is comfortably lounging on its back inside an open cardboard box on a light-colored carpet."*
* **text-only** prompts through the same processor are unaffected (the per-modality tensors are skipped, as in the other Gemma processors).

Happy to add a C++ unit test alongside the existing multimodal processor tests if you'd like one in this PR — pointers to the preferred fixture pattern for a processor that needs a real tokenizer welcome.

### Motivation and Context

Makes Gemma 3n multimodal packages runnable without hand-patching the chat template, and fixes the audio path, which did not run at all.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/models/gemma3n_multimodal_processor.cpp
+++ b/src/models/gemma3n_multimodal_processor.cpp
@@ -1,0 +1,238 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#include "../generators.h"
+#include "model.h"
+
+#include <cstring>
+
+namespace Generators {
+
+namespace {
+
+// Gemma 3n special tokens. Unlike Gemma 3, whose chat template emits the *begin*
+// marker <start_of_image>, Gemma 3n's template emits the soft token itself and the
+// processor is what wraps it in begin/end markers (see HF Gemma3nProcessor:
+// `prompt.replace(self.image_token, self.full_image_sequence)`).
+constexpr char kBoiToken[] = "<start_of_image>";
+constexpr char kImageToken[] = "<image_soft_token>";
+constexpr char kEoiToken[] = "<end_of_image>";
+constexpr char kBoaToken[] = "<start_of_audio>";
+constexpr char kAudioToken[] = "<audio_soft_token>";
+constexpr char kEoaToken[] = "<end_of_audio>";
+
+// token_type_ids values consumed by the embedding model, matching HF:
+// image -> 1, audio -> 3 (0 is text).
+constexpr int32_t kTextTokenType = 0;
+constexpr int32_t kImageTokenType = 1;
+constexpr int32_t kAudioTokenType = 3;
+
+size_t CountOccurrences(const std::string& text, const std::string& token) {
+  size_t count = 0;
+  size_t pos = 0;
+  while ((pos = text.find(token, pos)) != std::string::npos) {
+    ++count;
+    pos += token.size();
+  }
+  return count;
+}
+
+// Replace every occurrence of `from` with `to`, skipping past what was just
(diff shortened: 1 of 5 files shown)


NishantN2005:
@microsoft-github-policy-service agree company="Intel"

Anirudh-Swaminathan:
Latest main had a few folder refactors, so a rebase was necessary.
This PR can be closed, as another PR which rebased these commits on top of latest main exists at #2567

PR #2412, opened 2026-08-13
Add Tool Calling and Constrained Decoding to Engine

sayanshaw24:
## Summary

Adds end-to-end tool-calling support to the ONNX Runtime GenAI `Engine` while preserving the existing `Generator` and C tool-calling paths.

Applications can now:

1. Generate a tool call through `Engine`.
2. Consume and parse the generated tool-call tokens.
3. Execute the tool in the host application.
4. Append only the model-specific tool-response suffix with `Request::Continue()` or Python `request.continue_with()`.
5. Resume the same request with its retained KV cache and receive the final assistant response.

## Dependency

This change depends on Bhagirath's Engine continuation work in [PR #2423](https://github.com/microsoft/onnxruntime-genai/pull/2423), which is now merged. That PR owns the `Unassigned -> Assigned -> Active -> TurnComplete` lifecycle and the incremental `Continue()` API.

This PR consumes that API; it does not introduce a second continuation mechanism. `AddTokens()` remains initial-input-only, and later turns pass only newly appended token IDs to `Continue()`.

## Why

`Generator` already supports multi-turn tool execution. The Engine adds concurrent request scheduling, continuous batching, and paged KV-cache management, but tool-calling workflows also need request-local constrained decoding, model admission checks, and complete host examples built on the continuation lifecycle.

Retaining a request avoids recomputing its existing prompt and conversation KV state. It does not free logical context positions, so large tool catalogs still require host-side routing, bounded results, compaction, and eventual request rollover.

### Engine guidance and request safety

- Rejects requests created for a different model instance when added to an Engine.
- Adds per-request guidance/constrained-logits state to Engine requests.
- Applies guidance before the existing minimum-length, repetition-penalty, and no-repeat-ngram processors.
- Avoids copying the full guidance mask on every logits-processing step.
- Checkpoints and restores guidance with search and RNG state during Engine transactions.
- Resets guidance between generation turns while preserving the no-throw commit boundary.
- Explicitly rejects guidance fast-forward tokens, which are not yet compatible with Engine scheduling and cache accounting.
- Reports incomplete guidance type/data configuration separately from build-time guidance unavailability.

Guidance is optional and is not required for native Qwen tool calling. It is off by default at both levels: `USE_GUIDANCE` defaults to `OFF` at build time, and a guidance-enabled binary creates a processor only for requests whose `GeneratorParams` call `set_guidance()`.

For CUDA requests, llguidance advances the grammar and computes a compact allowed-token bitmask on the CPU. The mask is copied to the GPU and applied directly to GPU-resident logits with `LaunchAddLogitsMask`; logits are not copied back to the CPU. The profiling sample measures the combined cost of grammar advancement, transaction cloning, mask construction and transfer, and the GPU mask kernel.

### Engine tool-calling examples

- `examples/python/engine/tool-calling.py` demonstrates a complete Qwen weather-tool round trip.
- `examples/python/engine/coding-tool-calling.py` demonstrates a bounded coding-agent loop with sandboxed `read_file`, `edit_file`, and `run_tests` tools.
- `examples/python/engine/long-context-tool-calling.py` validates large OpenAI-style tool payloads and optionally runs selected live calls.
- `examples/python/engine/guidance-profiling.py` compares guided and unguided Engine setup, TTFT, inter-token latency, and decode throughput.

The examples use true incremental continuation. After each generated assistant turn, the host constructs and tokenizes only Qwen's newly appended chat-template fragment, then passes those IDs to `continue_with()`.

They intentionally do not rerender and retokenize the complete conversation to derive a suffix. BPE tokenization can merge across an old/new text boundary, so a textual prefix is not guaranteed to remain a token-ID prefix.

### Documentation and tests

- Adds a detailed Engine tool-calling design document.
- Updates paged-attention Engine documentation for per-request guidance behavior.
- Adds model-identity and guidance configuration tests.
- Uses PR #2423's continuation tests for lifecycle, cache residency, unread output, backpressure, and rollback behavior.
- Extends the existing Qwen tool-calling E2E path to run the Engine example without removing Generator or C coverage.

## Example

```text
Tool call:
{"name": "get_weather", "arguments": {"city": "Redmond"}}

Tool result: {"city": "Redmond", "temperature": "52 F", "conditions": "Partly cloudy"}
Final answer: The weather in Redmond is partly cloudy with a temperature of 52 degrees Fahrenheit.
```

## Compatibility

- Existing `Generator` APIs and tool-calling behavior are unchanged.
- Existing C tool-calling coverage remains enabled.
- Requests that do not continue after completion retain their previous behavior.
- Requests without guidance do not create a constrained-logits processor.
- Guidance-enabled builds can run guided and unguided requests in the same Engine.
- Tool parsing and execution remain application responsibilities; no model-specific tool API is added to Engine.

## Validation

- Engine test suite: **170/170 passed** in the guidance-disabled Debug build.
- RelWithDebInfo library and Python `0.16.0.dev0` wheel built successfully against ONNX Runtime `1.26.0`.
- Python syntax, editor diagnostics, and `git diff --check` passed.
- Weather E2E passed through tool call, host execution, incremental continuation on a retained request, and final answer.
- Coding-agent E2E passed across seven turns on one retained request; the file edit and unit test succeeded without request replay.
- Long-context validation passed with 94 unique tools: 29,306 prompt tokens in a 32,768-token context, leaving 3,462 positions.
- H100 guidance profiling completed with 30 measured runs per mode. Guided generation was schema-compliant in 30/30 runs versus 0/30 unguided runs. Guided p50 setup was 990.247 ms versus 0.632 ms, end-to-end TTFT was 993.552 ms versus 34.392 ms, inter-token latency increased 6.22%, and decode throughput decreased 5.85%.
- Nsight measured the CUDA mask kernel at 1.516 microseconds on average and approximately 0.05% of GPU kernel time. CPU-side per-request tokenizer and grammar construction is the primary optimization target.

Model used for local E2E validation:

```text
Microsoft/qwen2.5-0.5b-instruct-generic-cpu-4
```

## Known Limitations

- Guidance fast-forward tokens are not supported by Engine requests.
- Guidance-enabled positive runtime tests require an llguidance-capable build environment; the paired CUDA profile and schema checks were run on an H100 host with Rust 1.98.
- The examples implement Qwen's incremental continuation format; hosts supporting other model families must construct the corresponding model-specific suffix.
- A retained request cannot discard prior tokens. Context compaction requires a new request and prefill.
- GPT-OSS tool calling is outside the scope of this PR.

## Design Details

See `docs/engine_tool_calling_design.md` for the complete ownership model, incremental continuation contract, transaction behavior, long-context strategy, and follow-up work.


src/engine/request.cpp
```diff
@@ -499,8 +550,20 @@ void Request::CompleteGeneration() {
     }
   }
 
+  if (sequence_length > sequence_length_before) {
```
bmehta001: Are these ever different? Should `generated_token_count` perhaps be: `sequence_length - tokens_host_.size()`? CoPilot is suggesting replacing this block of code with:

```
void Request::CompleteGeneration() {
  search_->CompleteGeneration();

  const size_t sequence_length =
      static_cast<size_t>(CurrentSequenceLength());

  if (sequence_length > tokens_host_.size()) {
    const size_t new_token_count =
        sequence_length - tokens_host_.size();
    auto next_tokens = search_->GetNextTokens().CpuSpan();

    if (new_token_count > next_tokens.size()) {
      throw std::runtime_error(
          "The search produced fewer tokens than it appended to the sequence.");
    }

    auto new_tokens = next_tokens.last(new_token_count);

    if (guidance_logits_processor_) {
      guidance_logits_processor_->CommitTokens(new_tokens);
    }

    const size_t first_new_token = tokens_host_.size();
    tokens_host_.insert(
        tokens_host_.end(), new_tokens.begin(), new_tokens.end());

    for (size_t i = first_new_token; i < tokens_host_.size(); ++i) {
      unseen_token_indices_.push_back(i);
    }
  }
```

docs/engine_tool_calling_design.md
baijumeswani, line 1: Let's not check in design files. Any relevant information can be put in the paged_attention_engine.md as you've done. Otherwise, the docs are not easy to trace/update.

examples/python/engine/coding-tool-calling.py
baijumeswani, line 1: Let's also collapse the 4 examples into 1 single usage example. It does not need to be coding specific. It can be a simple general purpose example showcasing guidance and tool calling in the engine.

src/engine/engine.cpp
```diff
@@ -56,6 +56,10 @@ EngineDependencies Engine::CreateDependencies(std::shared_ptr<Model> model) {
 
 void Engine::AddRequest(std::shared_ptr<Request> request) {
   ReclaimAbandonedRequests();
+  const auto request_params = request->Params();
+  if (request_params->model_ && request_params->model_.get() != model_.get()) {
+    throw std::runtime_error("Cannot add a request created for a different model to the engine.");
+  }
```
baijumeswani: Is this relevant to the changes here? Why was this needed? The API redesign will eliminate the need for this. So, let's remove it for now.

src/engine/request.cpp
```diff
@@ -49,12 +50,33 @@ Request::Request(std::shared_ptr<GeneratorParams> params)
...
+  if (has_guidance_type != has_guidance_data) {
+    throw std::runtime_error("Guidance type and data must be provided together.");
+  }
+  const bool guidance_requested = has_guidance_type && has_guidance_data;
+  if (guidance_requested && !params->model_) {
+    throw std::runtime_error("Engine guidance requires request parameters associated with a model.");
+  }
+  if (params->model_) {
+    guidance_logits_processor_ = CreateGuidanceLogitsProcessor(*params->model_, params);
+  }
```
baijumeswani: Can guidance_logits_processor_  only be constructed when guidance is requested?

src/engine/request.cpp
```diff
@@ -49,12 +50,33 @@ Request::Request(std::shared_ptr<GeneratorParams> params)
...
+  if (guidance_requested && !guidance_logits_processor_) {
+    throw std::runtime_error("Engine guidance is unavailable. Build with use_guidance=true.");
+  }
 
   // The engine drives one independent search per request, so completion is batched: see
   // ScheduledRequests::GenerateNextTokens().
   search_->DeferCompletion(true);
 }
 
+Request::~Request() = default;
```
baijumeswani, line 78: This should not be necessary

src/constrained_logits_processor.h
```diff
@@ -47,6 +47,7 @@ struct GuidanceLogitsProcessor : public ConstrainedLogitsProcessor {
   static constexpr const char* kTokenizePrefixStr = "\x02";
 
   GuidanceLogitsProcessor(const State& state);
+  GuidanceLogitsProcessor(const Model& model, std::shared_ptr<const GeneratorParams> params);
```
baijumeswani: Does this need a externalmutable copy of the params? A user supplied params could be mutated between calls, and the engine may commit unexpected tokens causing failures. Maybe we can make a copy of the params before passing them in?

src/engine/request.cpp
baijumeswani: This should only be needed when guidance is enabled. Should this go inside the if (guidance_logits_processor) check?

test/python/test_onnxruntime_genai_e2e.py
```diff
@@ -178,6 +178,23 @@ def run_tool_calling():
             ]
             run_subprocess(command, cwd=cwd, log=log).check_returncode()
 
+            # Run a full Engine conversation: tool call, local tool result, and continuation.
```
baijumeswani: Avoid calling engine tests from this file.

src/engine/request.cpp
sayanshaw24: Fixed. New tokens are now derived from the difference between the completed sequence length and tokens_host_.size(), which handles both CPU and CUDA completion timing. The resulting span updates output bookkeeping and advances guidance when enabled. I added a two-step static CPU guidance test as regression coverage.

docs/engine_tool_calling_design.md
sayanshaw24: Done!

examples/python/engine/coding-tool-calling.py
sayanshaw24: Consolidated the usage samples into tool-calling.py, which demonstrates the complete tool-call round trip and supports optional request-local guidance through --guidance. The coding and long-context samples were removed. The profiler was retained as a benchmark and moved to benchmark_engine_guidance.py.

src/engine/engine.cpp
sayanshaw24: Removed the model-instance admission check and its test. I agree this should be handled by the upcoming API redesign rather than introduced in this PR.

src/engine/request.cpp
sayanshaw24: Updated. The constrained-logits processor is now constructed only when both guidance type and guidance data are present.

src/constrained_logits_processor.h
sayanshaw24: Fixed by snapshotting the guidance-relevant GeneratorParams fields during processor construction. The processor no longer depends on caller mutations. I also extended the test to mutate the original grammar after request creation and verify that the request continues using its snapshot.

src/engine/request.cpp
sayanshaw24: Guidance cursor advancement is now conditional on guidance_logits_processor_. The host-token mirror remains unconditional because Engine output and unseen-token bookkeeping require it for unguided requests too.

test/python/test_onnxruntime_genai_e2e.py
sayanshaw24: Removed the Engine example invocation from the general tool-calling E2E file. Engine behavior remains covered by the Engine lifecycle tests and the standalone usage example.

src/engine/request.cpp
sayanshaw24: Updated. The constrained-logits processor is now constructed only when guidance is requested, meaning both guidance type and guidance data are present.

baijumeswani (approved):

PR #2417, opened 2026-08-14
Retry smaller plans after execution capacity failures

baijumeswani:
## Summary

- Retry one smaller plan after a typed execution-capacity failure.
- Fully roll back before replanning with a new transaction ID.
- Reduce prefill tokens and rows while preserving decode work where possible.
- Surface the failure without poisoning the Engine if no smaller plan succeeds.

## Code flow

```text
Engine::Step()
    |
    v
Plan and reserve work
    |
    v
Execute model
    |
    +---------------- Success ----------------+
    |                                         |
    |                                         v
    |                                Stage and commit results
    |
    +--- ExecutionCapacityExceeded
                  |
                  v
          Roll back transaction
                  |
                  v
          Smaller plan available?
             /           \
           no             yes
           |               |
           v               v
     Surface error    Apply temporary limits
                           |
                           v
                    Plan new transaction
                           |
                           v
                    Execute one retry
                       /         \
                  success         failure
                     |               |
                     v               v
                   Commit       Surface error
```

The retry is limited to one attempt and applies only to typed capacity failures containing prefill work. Decode-only and unrelated failures are not retried.

### DIFF
--- a/docs/paged_attention_engine.md
+++ b/docs/paged_attention_engine.md
@@ -402,7 +402,9 @@ Rollback performs both parts:
 1. Restore request search state and sampler state from their checkpoints.
 2. Release every block held by the paged-cache reservation.
 
-After a successful rollback, committed request state and committed cache state match the state before the step began. The caller receives an `EngineStepError` with either `RetryableBatchAbort` or `ExecutionCapacityExceeded`, and the engine remains healthy. Calling `Step()` again with unchanged memory availability and workload composition may produce the same capacity failure.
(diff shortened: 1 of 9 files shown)


GopalakrishnanN:
Heads-up: #2418 is stacked on this branch and should merge after this PR.

It adds recompute-based request preemption to the dynamic paged engine, based on `baijumeswani/bounded-capacity-retry` rather than `main`, so it contains only the preemption layer on top of the work here. It also depends on this PR's capacity-retry loop directly: victims suspended during a step stay out of admission across the retry planning attempts added here, so a retry cannot hand reclaimed capacity straight back to the request the step just suspended.

The head branch is on a fork, and GitHub's native Stacks API does not support cross-fork stacks, so the two PRs will not show stack navigation or cascading merge. The base ref on #2418 is set correctly, so it merges cleanly once this lands — but it will need a rebase onto `main` if this PR is squashed rather than merged.

baijumeswani:
Closing this to avoid adding complexity to the engine.

PR #2419, opened 2026-08-15
Add prefix-aware paged KV caching to the continuous-batching engine

GopalakrishnanN:
> **No model ran in these measurements.** There is no GPU in the environment this was built and measured in (CPU-only build, `USE_CUDA=OFF`). Token counts, block residency, and hit rates below are exact and hardware-independent, because they are counts rather than timings. **End-to-end TTFT and throughput in seconds are not measured**, and the microsecond figures are Engine host overhead only — noise at this scale. Nothing is extrapolated.

> **Off by default** (`engine.dynamic_batching.prefix_caching = false`). Merged behavior is byte-identical to today.

Based on current `main` (`f6a871db`). Not stacked on anything.

### Description

Serving and agent workloads resend large identical prefixes on every turn: system prompts, tool schemas, repository and RAG context, and prior conversation. Today every one of those turns re-prefills from scratch, because `BlockPool` allocates and frees blocks with no notion of content identity.

This makes the engine detect that a new request's prompt shares a prefix with KV blocks that are already resident — or retained from a request that has finished — and adopt those blocks instead of recomputing them.

**Content-addressed blocks.** A filled block gets an identity that chains the identity of the block before it, so a block only matches when the entire preceding token sequence matches. A hash match is never treated as proof of a match: exact tokens are compared, **and the parent is compared as the identity object it was computed behind, rather than as a hash of it**.

That second point is the subtle one. Verifying only a parent *hash* leaves a real hole: if an ancestor is evicted and its hash is later reused by different content, a follower block can be spliced onto a prefix it was never computed behind — silently serving KV from the wrong context. An adversarial review pass caught this. It is fixed, and regression-tested by **`PrefixCacheTest.ABlockIsNeverSplicedOntoAPrefixItWasNotComputedBehind`**, which overrides the hash so that every first block collides with every other first block and every second block collides with every other second block — only the lineage check can tell them apart.

**Reference-counted sharing.** Several requests, the in-flight reservation, and the index can each hold a reference to one block; it returns to the pool only at zero references. The all-or-nothing reservation contract from #2365 holds — adopted references are released on rollback.

**Copy-on-write.** Only full blocks are shared, because a full block is never written again; the partial tail block stays private to its owner. `MakeTailBlockExclusive()` performs a real per-layer KV copy and table swap, and the reservation throws rather than write into a shared block, so "shared blocks are immutable" is enforced rather than assumed. This path is unreachable under the merged policy and is tested directly.

**Admission-time adoption.** Matching is read-only at admission, so a deferred request is left untouched; the cursor is staged only for selected requests, and only as the last thing planning does. It reuses the existing `processed_sequence_length_` / `UnprocessedTokens()` machinery, so the `past_sequence_lengths` guard added in #2354 validates every step — which matters, because that off-by-one is precisely the failure mode prefix adoption could otherwise reintroduce.

**Retention and LRU eviction.** Retention is bounded and always reclaimable under pressure. Planning charges a request for the retained blocks its adoption claims, which makes **adoption capacity-neutral** — that property is what makes the memory-pressure result below a designed outcome rather than a lucky one. Eviction takes a chain's **tail** first, so a head is never orphaned.

**Invariants** now validate per-block reference accounting (references must equal owners), full-blocks-only sharing, and adopted-prefix / committed-slot alignment.

### Measured results

Windows x64, MSVC (VS 18), Release, ORT 1.26, model `test/models/engine/dummy-decoder`. The benchmark drives the real `PagedCacheManager`, the real dynamic scheduler, and the real `Engine` transaction, stubbing only model execution.

Workload: 512-token shared prefix + 15 varying tokens + 8 generated, 32 sequential turns followed by 8 concurrent.

**A — `num_blocks=512`, `block_size=16`, `max_batch_size=8`**

| | cold | warm |
|---|---:|---:|
| prompt tokens computed | 21080 | **1112** |
| prompt tokens skipped | 0 | **19968** |
| decode tokens | 280 | 280 |
| peak KV blocks | 272 | **77** |
| admission -> first token p50 / p95 (µs, host only) | 36.4 / 55.3 | 44.1 / 59.7 |
| concurrent phase wall (µs, host only) | 1356.6 | 912.6 |

Hit rate **97.5%** (39/40 — the first turn is necessarily cold). **94.7%** of prompt tokens skipped. Peak KV residency **−71.7%**. Evictions 0, retention refusals 0, duplicate registrations 11, hash collisions 0.

**B — `num_blocks=96` (memory pressure)**

94.7% of prompt tokens skipped, 97.5% hit rate, peak 56/96 blocks, **24 evictions, 0 retention refusals**, and all 40 requests completed. Retention gave capacity back under pressure rather than starving anyone — which is the capacity-neutral adoption property doing its job.

### Why default-off

Warm-versus-cold output is bit-identical in the deterministic harness, but that cannot be qualified on CUDA or DML, where reused KV is compared against KV that a different batch composition would recompute with a different reduction order. It also changes KV block lifetime. Flipping the default is a one-line change once real-model qualification exists.

| Key (under `engine.dynamic_batching`) | Default | Meaning |
| --- | --- | --- |
| `prefix_caching` | `false` | Reuse resident blocks whose token prefix matches a new prompt. |
| `prefix_cache_pool_fraction` | `0.5` | Share of the block pool the index may retain. |
| `prefix_cache_max_blocks` | unset | Absolute ceiling on indexed blocks; overrides the fraction. |
| `prefix_cache_min_blocks` | `1` | Shortest match worth adopting, in blocks. |

### Relationship to #2418

#2418 (recompute preemption) reimplements `Remove()` on top of a new `CacheManager::ReclaimRequestCache`. That is the same release path this PR's reference-counted block release hooks into. The two are related by design rather than colliding accidentally, and **whichever merges second will need a rebase**.

### Tests

171 engine tests pass; 167 broader `unit_tests` pass with 22 skipped for absent models.

New: `prefix_cache_tests.cpp`, `prefix_caching_engine_tests.cpp`, `prefix_cache_benchmark_tests.cpp`; the invariant tests are rewritten.

Coverage: exact / partial / no hit; hash-collision safety; ancestor-collision splicing (**`ABlockIsNeverSplicedOntoAPrefixItWasNotComputedBehind`**); copy-on-write divergence; reference-count lifetime under transactional rollback; eviction under pressure; retained-block reuse across sequential requests; cursor continuity after adoption; default-config compatibility; and explicit rejection for static batching and beam search.

The decisive test is `WarmCacheProducesTokenForTokenIdenticalOutput`: warm output is token-for-token identical to cold, with strictly less prefill performed.

### Could not verify / residual risk

1. **No real-model or GPU qualification.** This is the main reason the feature is default-off.
2. `LayerBlockCopier` uses offset `DeviceBuffer::CopyFrom`, which is fine on CPU/CUDA/DML, but WebGPU and AMDGPU support only zero-offset copies. Reachable only from the copy-on-write path that the merged policy never takes, and it fails loudly rather than silently.
3. `PagedCacheReservation::Release()` can still throw from a destructor on allocation failure. The paths added here are non-allocating and `Release()` was converted to a non-allocating loop, but the destructor-calls-`Release()` shape is pre-existing and was not restructured.
4. **Ring / windowed pools (#2357, #2358) are not integrated.** They are documented as ineligible for sharing, since ring blocks are overwritten in place, but there is no test against that unmerged layout.
5. Real multi-threaded server concurrency and production block sizes (256+) are unmeasured.

`docs/paged_attention_engine.md` is updated in the same change.

Co-authored-by: Copilot App <223556219+Copilot@users.noreply.github.com>

### DIFF
--- a/docs/paged_attention_engine.md
+++ b/docs/paged_attention_engine.md
@@ -191,7 +191,7 @@ Completed requests that still own paged-cache blocks are deallocated and removed
 
 The scheduler snapshots all requests that already belong to the paged cache. These requests are expected to be `InProgress`.
 
-It then snapshots waiting requests from the scheduler pool. These requests are expected to be `Assigned` and are marked as newly admitted candidates.
+It then snapshots waiting requests from the scheduler pool. These requests are expected to be `Assigned` and are marked as newly admitted candidates. When prefix caching is enabled, each newly admitted candidate is offered to the prefix cache, which returns the longest run of block-aligned leading prompt tokens that is already resident. The query mutates nothing; the match is carried in the plan and only reaches the request in step 3.
 
 The scheduler orders candidates with decodes first. Order remains stable among
 decodes and among prefills. Each candidate initially contributes one provisional
@@ -204,6 +204,7 @@ For each candidate, the scheduler records a `RequestStepPlan` containing:
 - The sequence length before the transaction.
 - The provisional number of unprocessed tokens.
 - The number of cache slots required after the model run.
+- The resident prefix blocks the request would adopt, if any.
 - Whether the work is prefill.
 - Whether the request is newly admitted.
 
@@ -215,8 +216,9 @@ At this point, the plan contains candidates. It has not yet been reduced to the 
 
 - The configured maximum batch size.
 - The provisional scheduled-row limit.
-- The number of free physical blocks.
+- The number of free physical blocks, plus retained prefix blocks that can be reclaimed on demand.
 - Blocks already committed to active requests.
+- Blocks a newly admitted request would adopt from the prefix cache, which widen its block table without drawing on the free pool.
 - Additional blocks each request needs for this step.
 - The maximum block-table width supported by graph-capture buffers, when graph capture is enabled.
 
@@ -236,6 +238,8 @@ Selected entries are compacted to the beginning of the plan. Deferred entries re
 - An existing request keeps its current block table and stays `InProgress`.
 - A new request stays `Assigned` and owns no cache blocks.
 - Both can be considered again by the next engine step.
+
+Only requests that cache planning actually selected move their processed cursor. A selected request carrying a prefix match calls `Request::StagePrefixAdoption()`, which advances `processed_sequence_length_` to the end of the adopted prefix so the rest of the step -- token budget, packed layout, `past_sequence_lengths`, and cache target -- treats those tokens as already computed. The adoption is staged, not committed: a rolled back step restores the cursor.
 
 The planner distinguishes temporary capacity pressure from a request that can never fit:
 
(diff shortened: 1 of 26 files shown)


GopalakrishnanN:
Superseded by #2420, which carries the identical commits from a branch in this repository rather than from a fork.

Closing because CI cannot pass here. The `Integration Tests` Linux x64 and Windows x64 wheel builds fast-fail on this PR at 25-53s, while the same jobs take ~12 minutes and pass on #2420 with the same head commit (`bbb7da1f`) — they are gated for pull requests whose head is a fork. The macOS leg is unaffected and passed here too, which is why it looked like a code problem at first.

No code changed between the two PRs.

Issue #1102, opened 2024-11-26
`pip install onnxruntime-genai` does not work


xquyvu:
**Describe the bug**

Cannot install the package. Got the following error message:

```text
ERROR: Could not find a version that satisfies the requirement onnxruntime-genai (from versions: none)
ERROR: No matching distribution found for onnxruntime-genai
```

I'm aware of this open issue: https://github.com/microsoft/onnxruntime-genai/issues/559, but it's not fully resolved.

**To Reproduce**

`pip install onnxruntime-genai`
or
`pip install onnxruntime-genai --pre`

**Expected behavior**
The package installs without issue

**Desktop (please complete the following information):**
```python
>>> import platform
>>> print(platform.machine())
aarch64
>>> print(platform.version())
#1 SMP Wed Oct  2 16:38:00 UTC 2024
>>> print(platform.platform())
Linux-6.10.4-linuxkit-aarch64-with-glibc2.35
>>> print(platform.system())
Linux
>>> print(platform.processor())
aarch64
```

Python version: `3.12.7`

**Additional context**

xquyvu:
But works when running on an AML compute instance
```python
>>> import platform
>>> print(platform.machine())
x86_64
>>> print(platform.version())
#82~20.04.1-Ubuntu SMP Tue Sep 3 12:27:43 UTC 2024
>>> print(platform.platform())
Linux-5.15.0-1073-azure-x86_64-with-glibc2.31
>>> print(platform.system())
Linux
>>> print(platform.processor())
x86_64
```

Python version: `3.12.2`

maulberto3:
I believe this is happening starting with version 0.5.2 uploaded to pypi Nov 25. Because I just downloaded it a few days ago and everything went fine.

baijumeswani:
Python whls for linux aarch64 are not published. You would need to build the whl from source for linux aarch64.

We can look into adding this for an upcoming release.

PR #2421, opened 2026-08-17
Add chat template kwargs to tokenizer API

jennyf19:
## Summary
- add backward-compatible C and C++ tokenizer APIs for typed chat-template kwargs
- preserve the existing `ApplyChatTemplate` behavior by delegating with null kwargs
- pass model-specific context such as `enable_thinking` and `reasoning_effort` to ONNX Runtime Extensions
- add focused C API coverage for typed values and legacy equivalence

## Dependency
This PR depends on microsoft/onnxruntime-extensions#1102. The production dependency pin is intentionally unchanged; this PR should remain draft until #1102 merges and ORT GenAI updates its normal Extensions pin.

## Motivation
Foundry Local issue microsoft/foundry-local#808 needs a request path for:

```json
{"chat_template_kwargs":{"enable_thinking":false}}
```

Qwen 3 defaults to reasoning and can exhaust bounded output budgets before producing an answer or tool call. Qwen 2.5 has no equivalent reasoning-template branch.

## Validation
The code was stacked against the tested #1102 head only inside a disposable fork workflow. The native build compiled and `CAPITests.ChatTemplate` passed:
https://github.com/jennyf19/onnxruntime-genai/actions/runs/32071927312

No personal-fork dependency appears in this PR's production diff.


jennyf19:
## Ready for review — Extensions leaf landed

microsoft/onnxruntime-extensions#1102 merged as `bd0e21c11187e0b8a2385d1c61122a4d259a53a0` (merged by @sayanshaw24).

This PR now:
- Keeps the backward-compatible `OgaTokenizerApplyChatTemplateWithOptions` / C++ parallel API
- Pins `cmake/deps.txt` `onnxruntime_extensions` **GIT_TAG** to that merge commit (no personal-fork pin)
- Includes latest `main` merge

### Stack
1. ✅ Extensions #1102 merged
2. 👉 **This PR** (GenAI) — please review / undraft path
3. Foundry Local #1009 stays draft until a GenAI **NuGet** (`Microsoft.ML.OnnxRuntimeGenAI.Foundry`) includes this API (Baiju release train)

### Evidence
- Motivating Foundry issue + Qwen 3 before/after: microsoft/foundry-local#808
- Prior stacked fork validation (Extensions symbols + CAPITests.ChatTemplate): https://github.com/jennyf19/onnxruntime-genai/actions/runs/32071927312
- Official Extensions CI green on the merged tip: https://dev.azure.com/aiinfra/PublicPackages/_build/results?buildId=1373337

### Follow-up (non-blocking, Extensions)
Sayan noted a future direction: fold chat-template kwargs into `OrtxUpdateTokenizerOptions` and eventually remove the per-call `WithOptions` surface. Dependents should stay thin adapters.

cc anyone owning GenAI releases / packaging for Foundry Local.

jennyf19:
Looks like the red **Windows CUDA x64** check is an infra flake, not a code failure:

> The self-hosted runner lost communication with the server…

It died mid `Build with build.py` with no compile error. The twin **windows-cuda-x64-build** job on this same SHA is green, as are **linux-cuda-x64**, Windows CPU, lint, etc.

Please **Re-run failed jobs** when you have a moment — no source change needed from our side.

src/models/model.cpp
```diff
@@ -419,8 +419,15 @@ std::string Tokenizer::Decode(std::span<const int32_t> tokens) const {
 }
 
 std::string Tokenizer::ApplyChatTemplate(const char* template_str, const char* messages, const char* tools, bool add_generation_prompt) const {
+  return ApplyChatTemplateWithOptions(template_str, messages, tools, nullptr, add_generation_prompt);
```
baijumeswani: Why does this call need to change? Can it not continue using the OrtxApplyChatTemplate?

baijumeswani: Also, it would appear that this function `Tokenizer::ApplyChatTemplate` may no longer be called. Is that correct? If it is no longer in use, then we can remove it.

jennyf19: Good catch — fixed in the latest push.

1) **Why did the call change?** It didn’t need to. I had routed the no-kwargs path through `WithOptions(..., nullptr)`, but that’s unnecessary indirection. `Tokenizer::ApplyChatTemplate` again calls `OrtxApplyChatTemplate` directly; only `ApplyChatTemplateWithOptions` uses `OrtxApplyChatTemplateWithOptions`.

2) **Is `ApplyChatTemplate` unused?** No — still the public no-kwargs path (C API `OgaTokenizerApplyChatTemplate`, C++ `OgaTokenizer::ApplyChatTemplate`, Python `apply_chat_template`, examples, and tests). After this fix the C/C++ wrappers call it again instead of going through `WithOptions`, so it stays.

Happy to adjust further if you’d rather collapse the internal duplication another way.

baijumeswani (approved):

PR #2422, opened 2026-08-18
long prefill engine benchmark

prathikr:
### Summary

This benchmark measures single-request prefill (time to first token) across model sizes and long
prompt lengths. Prefill latency equals TTFT because each request generates a single token.

| Params (B) | Prompt length (K) | Prefill p25 (ms) | Prefill p50 (ms) | Prefill p95 (ms) | Avg. prompt tokens/s | Peak device memory (GiB) | Peak host memory (GiB) |
|---:|---:|---:|---:|---:|---:|---:|---:|
| 0.5 | 32 | 619.197 | 620.813 | 624.115 | 52754.023 | 50.434 | 2.339 |
| 0.5 | 64 | 1920.057 | 1926.098 | 1934.887 | 34003.247 | 50.426 | 2.340 |
| 0.5 | 128 | 6665.113 | 6680.199 | 14500.657 | 16258.726 | 11.426 | 1.122 |
| 1.5 | 32 | 2115.934 | 2118.584 | 2568.264 | 14803.690 | 11.457 | 1.120 |
| 1.5 | 64 | 3259.079 | 3260.503 | 3264.136 | 20100.623 | 59.428 | 2.620 |
| 1.5 | 128 | 10472.431 | 10480.336 | 10490.953 | 12506.818 | 58.426 | 2.589 |
| 7 | 32 | 3698.797 | 3699.787 | 7789.669 | 8179.296 | 61.434 | 2.668 |
| 7 | 64 | 9423.713 | 9431.170 | 9438.029 | 6950.679 | 61.457 | 2.672 |
| 7 | 128 | 27812.689 | 27818.570 | 27827.903 | 4711.878 | 62.426 | 2.708 |

### Configuration

| Component | Value |
|---|---|
| Models | Qwen 2.5 0.5B / 1.5B / 7B Instruct |
| Execution provider | CUDA |
| ONNX Runtime | 1.29.0 |
| CUDA plugin EP | 0.1.0 |
| GenAI | 0.16.0-dev |
| Concurrency | 1 |
| Prompt lengths | 32K / 64K / 128K |
| Generation length | 1 token/request |
| Measured runs | 20 |

### Other Improvements
- Added run.py to run each scenario in a fresh process, avoiding CUDA/ORT allocator carryover.
- Added optional multi-GPU scheduling with --cuda_visible_devices; each scenario uses one GPU.
- Added quiet lifecycle logging and optional --verbose child output.
- Updated README.md with the new runner and multi-GPU commands.
- Moved model-path resolution and deterministic RULER prompt loading into shared utils.
- Removed the obsolete synthetic config option and related validation/metadata.
- Made dispatcher provider registration support CUDA and WebGPU and find staged versions dynamically.
- Updated benchmark docs for deterministic RULER prompts.
- Improved NuGet dependency downloads with streaming and timeouts.


benchmark/engine/run.py
```diff
@@ -0,0 +1,88 @@
...
+                shutil.move(result_path, args.out / f"{scenario['scenario']}_results_{index:03d}.json")
+            state = "completed" if completed.returncode == 0 else f"failed ({completed.returncode})"
+            print(f"{state.capitalize()} {label}", flush=True)
+            return completed.returncode != 0
+
+    work = list(enumerate(scenarios, 1))
+    max_workers = len(gpu_ids) if gpu_ids is not None else 1
+    with ThreadPoolExecutor(max_workers=max_workers) as executor:
+        failed = any(executor.map(run_scenario, work))
+
```
tianleiwu, line 79: The executor's workers consume one shared queue, but GPU assignment is derived from the scenario index rather than from a worker or per-GPU queue. With GPUs `0,1`, if scenario 2 finishes before scenario 1, that worker can start scenario 3 (assigned to GPU 0) while scenario 1 is still running there. This contaminates latency/memory measurements and can cause avoidable OOMs. Please partition the work into one serial queue per GPU, or hold a per-GPU lock for the complete subprocess lifetime, so each selected device runs at most one scenario at a time.

benchmark/engine/scenarios/long_prefill.cpp
```diff
@@ -0,0 +1,128 @@
...
+              << " complete: prefill_ms=" << prefill_ms
+              << ", prompt_processing_tps=" << prompt_processing_tps << std::endl;
+
+    if (is_warmup) {
+      continue;
+    }
+
+    ttft_values.push_back(prefill_ms);
+    prefill_ms_values.push_back(prefill_ms);
+    prompt_processing_tps_values.push_back(prompt_processing_tps);
```
tianleiwu, line 110: This records `completed=true` through the default field value even when `generation_tokens > 1`. The scenario calls `Step()` once and consumes one token, so the remaining requested tokens are never generated, yet the benchmark reports success instead of detecting an incomplete request. Since this scenario is prefill-only and the supplied configs use one token, please validate `generation_tokens == 1`; alternatively, drain the request and set `completed` from the observed token count.

benchmark/engine/scenario_dispatcher.cpp
```diff
@@ -27,9 +28,23 @@ namespace {
...
-    throw std::runtime_error("versions.json not found; expected the build to stage it under benchmark/engine.");
+  // Match versions.json staged under any build config: build/Linux/<config>/benchmark/engine.
+  const std::regex versions_pattern(R"(build/Linux/[^/]+/benchmark/engine/versions\.json$)");
+  fs::path versions_path;
+  for (const auto& entry : fs::recursive_directory_iterator("build/Linux")) {
+    if (!entry.is_regular_file() || !std::regex_search(entry.path().generic_string(), versions_pattern)) {
+      continue;
+    }
+    if (!versions_path.empty()) {
+      throw std::runtime_error("Multiple versions.json found under build/Linux; expected exactly one build config.");
```
tianleiwu: This makes every benchmark executable fail when the tree contains more than one staged configuration, such as both Debug and RelWithDebInfo. It also still misses custom build roots and launches outside the repository root. Since `versions.json` is staged next to the executable, please resolve that sibling path from the running executable rather than searching the working tree.

tianleiwu (changes requested):
Requesting changes for benchmark-result reliability. The process isolation is a good foundation, but multi-GPU execution can still overlap scenarios on one device, long-prefill can mark an incomplete request successful, and metadata lookup fails in ordinary multi-configuration build trees.

tianleiwu (approved):

baijumeswani (approved):

PR #2424, opened 2026-08-18
Use cuda plugin ep

baijumeswani:
Migrate cuda CI pipelines from `onnxruntime-gpu` (provider-bridge) packages to `onnxruntime-ep-cuda` (plugin) packages.


src/models/adapters.cpp
```diff
@@ -2,12 +2,14 @@
...
+#include "adapters.h"
 
 namespace Generators {
 
-Adapter::Adapter(const char* adapter_file_path, Ort::Allocator* allocator)
-    : adapter_{OrtLoraAdapter::Create(fs::path(adapter_file_path).c_str(), *allocator)} {}
+Adapter::Adapter(const char* adapter_file_path)
+    // Keep adapter parameters CPU-backed so ORT copies them through the session's
+    // data transfer manager, which supports both plugin and provider-bridge EPs.
+    : adapter_{OrtLoraAdapter::Create(fs::path(adapter_file_path).c_str())} {}
```
tianleiwu, line 12: This changes CUDA LoRA parameters from device-resident values created once at adapter load to CPU-backed session inputs. ORT appends active adapter values to each `session.Run()`, and generation runs the session for every token, so a large adapter can now cross PCIe on every decode step. Could we stage these tensors once through a plugin/session-compatible allocator, or provide a representative LoRA decode benchmark showing that the throughput impact is acceptable?

test/python/_test_utils.py
```diff
@@ -11,8 +11,9 @@
 
 # Execution providers shipped as separate plug-in libraries that must be registered with
 # ONNX Runtime before use. Maps the GenAI provider name to the Python package that exposes
-# the plug-in library path via get_library_path().
+# the plug-in name and library path.
 PLUGIN_EP_PACKAGES = {
+    "cuda": "onnxruntime_ep_cuda",
```
tianleiwu, line 16: Adding CUDA to this optional registry inherits the helper's catch-all registration behavior: registration exceptions return `False`, and the bulk caller ignores that result. In plugin-only CI, please assert that an installed CUDA plugin registered successfully so this migration explicitly proves the V2 plugin path instead of relying on a later failure or legacy fallback to reveal the problem.

tianleiwu (changes requested):
The CPU-runtime/CUDA-plugin package separation is well structured across native, Python, and C# validation, and explicitly removing provider-bridge packages is a strong migration check. I am requesting changes for the unconditional LoRA host-backing behavior because it can move a one-time upload into every decode iteration; I also recommend making CUDA plugin registration fail closed in plugin-only CI so the migration gate explicitly proves the V2 path.

src/models/adapters.cpp
baijumeswani: I'll address this in ort. Right now, ort does not accept the cuda allocator if using the plugin ep. This is a temporary workaround. But I think it might be better to address in ort first.

tianleiwu (dismissed):

tianleiwu (commented):
The CUDA plugin migration looks sound across native, Python, C#, and plugin-only CI staging. I found one non-blocking sample regression: the CI memory workaround lowers Whisper's generation cap for all users; please consider exposing it as a test override while retaining the previous sample default.

examples/python/whisper.py
```diff
@@ -60,7 +62,7 @@ def run(args: argparse.Namespace):
             do_sample=False,
             num_beams=args.num_beams,
             num_return_sequences=args.num_beams,
```
tianleiwu, line 64: Could we keep the sample's existing 448-token behavior and scope the lower cap to this CI case? `whisper.py` is a user-facing example for arbitrary audio, so hard-coding 256 can truncate transcriptions that previously fit Whisper's context. One option is to add `--max_length` with a default of 448 and have `run_whisper()` pass 256 for the constrained GPU job.

kunal-vaishnavi: I agree with this comment.

examples/python/common.py
```diff
@@ -10,6 +10,14 @@
 
 import onnxruntime_genai as og
 
+_PLUGIN_EP_PACKAGES = {
```
kunal-vaishnavi: Instead of using top-level variables, can we put this as a local variable where it's used?

tianleiwu (approved):

PR #2427, opened 2026-08-19
Add decoder state manifest contract for hybrid model support

baijumeswani:
## Summary

Introduce an optional, model-described decoder state manifest that defines which logical layers own paged key/value state or fixed-size state, and maps that state to ONNX graph inputs and outputs.

The continuous-batching Engine currently assumes every decoder layer owns the same growing K/V cache. That assumption does not hold for hybrid decoders that combine attention layers with recurrent or convolutional layers. This PR establishes the configuration and validation contract needed to support those models without adding sparse or fixed-state execution to the Engine yet.

## Configuration contract

`model.decoder.state_groups` is an optional list of state groups. Each group contains:

- `kind`: `paged_kv` or `fixed`
- `layer_ids`: logical decoder layer IDs that own the state
- binding templates containing exactly one `%d`, expanded with the logical layer ID
- `key` and `value` bindings for `paged_kv`, or one `state` binding for `fixed`

Example:

```json
{
  "model": {
    "decoder": {
      "state_groups": [
        {
          "kind": "paged_kv",
          "layer_ids": [3, 7],
          "bindings": {
            "key": {
              "input": "past_key_values.%d.key",
              "output": "present.%d.key"
            },
            "value": {
              "input": "past_key_values.%d.value",
              "output": "present.%d.value"
            }
          }
        },
        {
          "kind": "fixed",
          "layer_ids": [0, 1, 2],
          "bindings": {
            "state": {
              "input": "past_state.%d",
              "output": "present_state.%d"
            }
          }
        }
      ]
    }
  }
}
```

A layer may own both paged and fixed state. Multiple fixed groups may cover the same layer when their expanded tensor bindings are distinct. Layers omitted from a group do not own that kind of state.

## Compatibility and Engine behavior

This change is backward compatible:

- `state_groups` remains absent for existing model configurations and builders.
- Absence preserves the current dense, sequential, all-layer paged-KV contract.
- Existing paged and non-paged models therefore retain their current configuration and runtime behavior.

The current dynamic Engine does not consume sparse paged or fixed state yet. It explicitly fails closed when a manifest describes a contract other than the legacy all-layer paged-KV layout. An explicit legacy-equivalent manifest is accepted only when its bindings exactly match the names used by the existing Engine.

This prevents a hybrid model from loading with a manifest that the current cache manager would silently interpret as dense K/V state. Sparse paged-cache allocation and fixed-state management can build on this contract in subsequent PRs.

## Builder changes

The common builder exposes hooks for emitting decoder state groups but preserves manifest absence by default.

The hybrid Qwen builder resolves layer types from either the root model configuration or nested `text_config`, including reduced-layer exports. It emits:

- one sparse `paged_kv` group for full-attention layers;
- independent `fixed` groups for convolutional and recurrent state.

All-attention exports continue to omit the manifest and use the legacy contract.

### DIFF
--- a/docs/paged_attention_engine.md
+++ b/docs/paged_attention_engine.md
@@ -85,6 +85,21 @@ to 2048. Both limits are positive and independent.
 
 Without dynamic batching, the engine uses the older static batching path. Static batching allocates and advances a batch as a unit. It does not use the transaction flow described below.
 
+## Decoder state manifest
+
+`model.decoder.state_groups` can describe decoder-owned state without identifying a model family. The supported group kinds are `paged_kv` and `fixed`. Each group has logical layer IDs and typed input/output binding templates. Paged KV groups have key and value bindings; fixed groups have one state binding. Binding templates contain one `%d` placeholder for the logical layer ID. A layer may own both paged and fixed state, and multiple fixed groups may cover the same layer when it owns more than one independent state tensor. Layers omitted from a group do not own that kind of state.
+
+The group kind defines the state transition:
+
+- Paged KV grows by appending token slots and commits by advancing logical occupancy.
+- Fixed convolution and recurrent state replace one request-indexed state value and require a staged output to be published at commit.
+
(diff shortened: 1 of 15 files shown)


tianleiwu (approved):
Reviewed at `0061d8ee724bb3b826d11cee35c1c3956776cc85`.

No actionable findings. The decoder-state manifest contract is internally consistent across configuration parsing, session metadata validation, builder output, documentation, and the Engine compatibility boundary. Legacy manifest absence remains backward compatible, while unsupported sparse/fixed dynamic Engine layouts fail closed.

Validation performed:
- Python builder tests: 6 passed
- Focused C++ manifest/config tests: 21 passed
- Incremental `engine_unit_tests` build: succeeded
- Ruff lint and format checks: passed
- `git diff --check`: passed

PR #2428, opened 2026-08-19
Enable sparse paged KV layers in the continuous batching Engine

baijumeswani:
## Summary

This change enables the continuous batching Engine to consume an explicit sparse `paged_kv` state group from the decoder state manifest introduced by #2427.

The Engine previously assumed every decoder layer owned a growing KV cache. It now:

- allocates cache tensors only for the manifest's logical layer IDs;
- uses the manifest's exact key/value input and output bindings;
- sizes automatic cache capacity using only participating layers;
- preserves existing dense behavior when `state_groups` is absent.

This is needed for hybrid decoders where only selected layers use paged attention.

## Behavior

Dynamic batching accepts either:

- a legacy model without `state_groups`; or
- exactly one explicit `paged_kv` group.

Sparse logical layer IDs are preserved without renumbering. Existing models continue to use the legacy dense layer range and decoder-wide binding templates.

Fixed state groups remain rejected. Supporting them safely requires request-owned fixed-state storage and commit/rollback transactions, which will follow separately.

Sparse paged layers also integrate with the existing sliding-window cache:

- configured sliding-window layers must belong to the paged group;
- their fixed ring allocation is reserved first;
- the remaining automatic memory budget is divided across participating
  full-attention layers.

## Tests

The `synthetic-paged` fixture now models six logical layers with paged KV state only at layers `1` and `4`. Its explicit bindings differ from the legacy templates, proving the Engine uses the manifest contract.

Coverage includes sparse allocation and bindings, legacy fallback, capacity accounting, sliding-window interaction, and rejection of malformed, fixed, or multiple paged groups.

## Scope

This PR only enables sparse paged-KV consumption. Fixed-state pools, composite state transactions, packed hybrid operators, and model-family enablement remain follow-up work.


tianleiwu (approved):
LGTM

PR #2430, opened 2026-08-20
Add split-KV lower bound to PagedAttention metadata

baijumeswani:
## Description

This updates the continuous-batching Engine and model builder to use the three-value `PagedAttention` metadata contract:

```text
[max_query_len_bound, max_kv_len_bound, max_kv_len_lower_bound]
```

The new lower bound lets ONNX Runtime enable split-KV FlashAttention for eligible long-context decode steps without reading sequence lengths back from the device.

### What changed

- Emit `attention_metadata` as a CPU `int32[3]` tensor.
- Use exact query and KV bounds for eager steps.
- Use replay-safe bounds for CUDA graph capture:
  - query upper bound: `1`;
  - KV upper bound: `block_table_columns * block_size`;
  - KV lower bound: the preceding block-table bucket boundary plus one token.
- Centralize graph block-table bucketing so cache planning, block-table
  construction, and metadata generation share the same rules.
- Fall back to a conservative KV lower bound of `1` if a future reservation
  policy allocates blocks ahead of the live sequence length.
- Update the model builder to export `attention_metadata` with shape `[3]`.
- Update the synthetic paged Engine model so integration tests exercise the
  real three-value input path.
- Document the updated runtime contract and CUDA graph behavior.

Backward compatibility with two-value metadata is intentionally not retained.
Models using `attention_metadata` must be rebuilt with the three-element input.


tianleiwu (approved):
LGTM

baijumeswani:
Thank you for the review.

PR #2436, opened 2026-08-21
Split paged cache reservation commit into validate and publish phases

baijumeswani:
## Summary

This PR splits `PagedCacheReservation::Commit()` into an explicit non-mutating validation phase and a publication phase:

- `ValidateCommit()` verifies that the reservation can still be published without mutating the committed cache, block pools, reserved blocks, or reservation state.
- `CommitValidated()` publishes a reservation that has already passed validation.
- `Commit()` remains the compatibility wrapper and calls both methods in order.

The split provides the preflight boundary required by a later composite Engine transaction. That transaction must validate paged KV state, fixed decoder state, and other request-owned resources before publishing any of them.

## Motivation

The original `Commit()` validated and published paged-cache changes in one operation. That is sufficient while paged KV is the only transaction-owned decoder state, but it cannot safely participate in a multi-resource commit: if paged KV publishes before another resource fails validation, the request is left partially committed.

This change makes paged-cache validation independently callable and repeatable. A composite transaction can validate every resource first and enter publication only after all preconditions are known to hold.

### Added `ValidateCommit()`

`ValidateCommit()` is `const` and does not publish or release anything. It re-derives the assumptions needed by publication from the current committed cache and rejects the reservation if:

- the reservation is no longer in the `Reserved` state;
- its owning block pool or committed-table collection is missing;
- sliding-window reservation configuration is inconsistent;
- a request identity is null or duplicated;
- request ownership changed after reservation;
- an existing request's committed token boundary changed;
- a delta regresses below its committed boundary;
- reserved block or sliding-window block offsets are not contiguous and complete;
- a delta refers beyond its reserved block spans;
- committed plus reserved blocks cannot reach the target token boundary;
- an existing table no longer has the preallocated capacity needed for publication;
- reserved block accounting, window block accounting, or new-table accounting no longer matches the reservation;
- the committed-table vector no longer has this reservation's required headroom.

Because validation is non-mutating, callers may validate more than once before publication.

### Added `CommitValidated()`

`CommitValidated()` contains the existing publication logic. It moves preallocated block handles and tables into the committed cache and advances committed token boundaries.

The method performs a state-only guard before mutation to reject double publication. It intentionally does not repeat ownership, token-boundary, layout, reachability, or capacity validation; those checks belong to `ValidateCommit()` so a composite transaction can run all fallible preflight checks before publishing any resource.

`CommitValidated()` is intentionally not declared `noexcept`. Although the validated single-reservation path performs no new allocation or device work, the standard-library insertion and bounds-checked operations it uses are not `noexcept`-declared. Keeping the honest potentially-throwing signature avoids converting an unexpected precondition violation into `std::terminate`.

### Preserved the existing API

`Commit()` retains its existing behavior:

```cpp
void PagedCacheReservation::Commit() {
  ValidateCommit();
  CommitValidated();
}
```

Existing callers do not need to change in this PR.

## Transaction flow

```mermaid
flowchart LR
    Reserve["Reserve paged blocks and table capacity"] --> Reserved["Reservation state: Reserved"]
    Reserved --> Validate["ValidateCommit()<br/>const and non-mutating"]
    Validate -- "validation fails" --> Unchanged["Reservation and committed cache remain unchanged"]
    Validate -- "validation succeeds" --> Ready["Validated publication preconditions"]
    Ready --> Publish["CommitValidated()<br/>publish preallocated blocks and tables"]
    Publish --> Committed["Reservation state: Committed"]

    Composite["Later composite transaction"] -. "validate paged KV" .-> Validate
    Composite -. "validate fixed state and sibling resources" .-> Other["Other resource validation"]
    Ready -. "publish only after all validations pass" .-> CompositePublish["Composite publication boundary"]
    Other -.-> CompositePublish
    CompositePublish -.-> Publish
```

### DIFF
--- a/src/engine/block.cpp
+++ b/src/engine/block.cpp
@@ -54,29 +54,31 @@ BlockPool::BlockPool(size_t block_size, size_t num_blocks)
     : block_size_(block_size), capacity_(num_blocks) {}
 
 std::vector<std::shared_ptr<Block>> BlockPool::AllocateBlocks(size_t num_slots, bool mark_slots_used) {
-  const auto allocate_block = [this](size_t slots) {
-    for (size_t i = 0; i < Capacity(); ++i) {
-      if (blocks_[i] == nullptr) {
-        blocks_[i] = std::make_shared<Block>(i, slots, block_size_);
-        return blocks_[i];
-      }
-    }
-    return std::shared_ptr<Block>();
-  };
-
-  if (BlocksNeeded(num_slots) > AvailableBlocks()) {
-    throw std::runtime_error("Requested number of blocks " + std::to_string(BlocksNeeded(num_slots)) +
+  const size_t blocks_needed = BlocksNeeded(num_slots);
+  if (blocks_needed > AvailableBlocks()) {
+    throw std::runtime_error("Requested number of blocks " + std::to_string(blocks_needed) +
                              " for number of slots " + std::to_string(num_slots) +
                              " exceeds available blocks " + std::to_string(AvailableBlocks()) + ".");
   }
 
   std::vector<std::shared_ptr<Block>> allocated_blocks;
-  for (size_t i = 0; i < num_slots; i += block_size_) {
-    auto block = allocate_block(mark_slots_used ? std::min(block_size_, num_slots - i) : 0);
-    if (!block) {
-      throw std::runtime_error("Failed to allocate a block.");
+  allocated_blocks.reserve(blocks_needed);
+  for (size_t id = 0; id < Capacity() && allocated_blocks.size() < blocks_needed; ++id) {
+    if (blocks_[id] == nullptr) {
+      const size_t allocated_slots = allocated_blocks.size() * block_size_;
+      const size_t slots =
+          mark_slots_used ? std::min(block_size_, num_slots - allocated_slots) : 0;
+      allocated_blocks.push_back(std::make_shared<Block>(id, slots, block_size_));
     }
-    allocated_blocks.push_back(block);
+  }
+
+  // Publish only after every allocation succeeds. Until this loop, an exception leaves the pool
+  // unchanged and the local handles clean themselves up.
+  for (const auto& block : allocated_blocks) {
+    blocks_[block->Id()] = block;
+  }
+  if (!allocated_blocks.empty()) {
+    ++mutation_generation_;
   }
   return allocated_blocks;
 }
@@ -124,6 +126,19 @@ void BlockPool::Free(const std::vector<std::shared_ptr<Block>>& blocks) {
 
   for (const auto& block : blocks) {
     blocks_[block->Id()].reset();
+  }
+  if (!blocks.empty()) {
+    ++mutation_generation_;
+  }
+}
+
+void BlockPool::RollbackReservedBlocks(
+    const std::vector<std::shared_ptr<Block>>& blocks) noexcept {
+  for (const auto& block : blocks) {
+    blocks_[block->Id()].reset();
+  }
+  if (!blocks.empty()) {
+    ++mutation_generation_;
   }
 }
 
--- a/src/engine/block.cpp
+++ b/src/engine/block.cpp
@@ -143,6 +143,10 @@ size_t BlockPool::BlockSize() const {
   return block_size_;
 }
 
+bool BlockPool::Owns(const std::shared_ptr<Block>& block) const {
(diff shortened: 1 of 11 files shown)


src/engine/paged_cache_reservation.cpp
```diff
@@ -223,25 +322,207 @@ void PagedCacheReservation::FillWindowBlockTable(
...
+
+  // Re-derive everything CommitValidated depends on straight from the committed cache so that a
+  // change to ownership, token boundaries, delta layout, or preallocated capacity between reserve
+  // and commit is rejected here rather than during publication.
+  size_t new_table_count = 0;
+  size_t assigned_reserved_blocks = 0;
+  size_t assigned_reserved_window_blocks = 0;
+  std::unordered_set<const void*> committed_request_ids;
+  committed_request_ids.reserve(committed_tables_->size());
+  for (const auto& table : *committed_tables_) {
```
tianleiwu, line 424: **High-priority:** This scan checks omitted resident tables only for request-ID uniqueness; their mutation generations are never snapshotted or rechecked. After reserving a block for request A, an omitted resident B can replace its mapping with `reservation.ReservedBlocks()[0]`. That changes B's table generation but neither pool generation, and B has no delta, so both `ValidateCommit()` and `CommitValidated()` succeed and publish the same physical block in A and B. Please snapshot/recheck every resident table's identity and generation (including additions/removals), or add a cache-wide table-mapping generation, before publication.

src/engine/block.cpp
```diff
@@ -54,29 +54,31 @@ BlockPool::BlockPool(size_t block_size, size_t num_blocks)
...
-        blocks_[i] = std::make_shared<Block>(i, slots, block_size_);
-        return blocks_[i];
-      }
-    }
-    return std::shared_ptr<Block>();
-  };
-
-  if (BlocksNeeded(num_slots) > AvailableBlocks()) {
-    throw std::runtime_error("Requested number of blocks " + std::to_string(BlocksNeeded(num_slots)) +
+  const size_t blocks_needed = BlocksNeeded(num_slots);
```
tianleiwu, line 57: **Suggestion:** `BlocksNeeded()` still uses `(num_slots + block_size_ - 1) / block_size_`, so `SIZE_MAX` with `block_size_ == 4` wraps to zero. With the rewritten loop, `allocated_blocks.size() < blocks_needed` is false immediately and `AllocateBlocks()` returns an empty vector successfully; the previous loop eventually rejected the request after exhausting the pool. Please use the overflow-safe ceiling `num_slots / block_size_ + (num_slots % block_size_ != 0)` and add a `std::numeric_limits<size_t>::max()` boundary test.

tianleiwu (changes requested):
Requesting changes. The two-phase structure, repeatable validation, and all-delta publication preflight are solid, and the focused engine suite passes. One stale-state gap can violate physical block single ownership, and the rewritten allocation loop has a separate overflow regression; details are inline.

src/engine/paged_cache_reservation.cpp
baijumeswani: Addressed this by snapshotting every request in both validation phases.

src/engine/paged_cache_reservation.cpp
```diff
@@ -16,13 +18,70 @@ namespace {
 PagedCacheBlockTable* FindTable(std::vector<PagedCacheBlockTable>& tables, const void* request_id) {
   const auto it = std::find_if(tables.begin(), tables.end(),
                                [request_id](const PagedCacheBlockTable& table) {
-                                 return table.request_id == request_id;
+                                 return table.RequestId() == request_id;
                                });
   return it == tables.end() ? nullptr : &*it;
 }
 
+void ValidateEmptyReservedBlocks(
```
tianleiwu, line 27: Suggestion: This new helper uses `std::string_view`, but this translation unit does not include `<string_view>`. It currently compiles through transitive standard-library includes, which are not guaranteed across supported toolchains. Please include `<string_view>` directly.

tianleiwu (dismissed):
The validate/publish split, generation checks, resident-table snapshots, and all-delta publication preflight look sound on the current head. The previous blocking ownership and overflow concerns are addressed. I left one non-blocking portability suggestion inline. Focused validation passed all 78 tests in BlockPoolTest, InvariantValidatorTest, and PagedCacheReservationTest.

tianleiwu (approved):

PR #2437, opened 2026-08-21
Integrate composite decoder state transactions

baijumeswani:
## Summary

This PR coordinates paged KV cache state and fixed per-request decoder state in one Engine transaction.

Hybrid models need both forms of state to advance at the same token boundary. A step now plans and reserves both resources together, stages fixed-state updates without changing committed state, and publishes both only after all validation and device work succeeds.

## What changed

- `PagedCacheManager` creates a `FixedStatePool` when the model manifest declares fixed state groups.
- Step planning checks paged and fixed ownership together and includes fixed-state rows, new slots, and staging memory in the plan.
- Step reservation keeps paged blocks and fixed-state slots in the same scheduled request order.
- `ExecutionContext` exposes fixed-state slots and bindings to the later decoder IO integration.
- Commit is split into clear phases:
  1. Validate paged and fixed reservations.
  2. Copy fixed outputs into inactive banks.
  3. Commit request/search state.
  4. Publish paged ownership and fixed-state bank flips.
  5. Commit request bookkeeping and ready events.
- Failures before publication restore request/search state, release paged reservations, and discard staged fixed state.
- Request removal releases paged KV blocks and fixed-state slots together.
- Composite invariants verify that paged and fixed ownership contain the same requests at the same committed token boundary.
- Fixed-state allocations happen before automatic paged KV sizing, so paged capacity is computed from the remaining device memory.
- Static fixed-state batch dimensions are rejected because continuous batching requires a variable number of rows per step.

## Transaction flow

```text
Plan paged + fixed resources
            |
Reserve paged blocks + fixed slots
            |
Checkpoint request/search state
            |
Execute and stage outputs
            |
Validate both state reservations
            |
Prepare fixed inactive banks
            |
Publish request/search + paged KV + fixed state
            |
Commit request bookkeeping and ready events
```

If execution or post-processing fails before publication, the complete step is rolled back. A failure after the publication boundary is treated as fatal because some state may already be committed.


src/engine/block.h
```diff
@@ -79,6 +81,8 @@ struct BlockPool {
   std::vector<std::shared_ptr<Block>> ReserveBlocks(size_t num_slots);
 
   void Free(const std::vector<std::shared_ptr<Block>>& blocks);
+  void ValidateFree(std::span<const std::shared_ptr<Block>> blocks) const;
+  void FreeValidated(std::span<const std::shared_ptr<Block>> blocks) noexcept;
```
tianleiwu, line 87: `FreeValidated` is public and `noexcept`, and does `blocks_[block->Id()].reset()` with no null or bounds check. Before this PR, `Free()` validated inline so an out-of-range or null block was impossible to reach. Now the precondition ("must be immediately preceded by `ValidateFree` on unchanged state") is implicit and unwritten, and because the function is `noexcept` a violation is an out-of-bounds heap write rather than an exception.

The same pattern applies to `FixedStatePool::ReleaseValidated` (`impl_->slots[handle.slot]`, unchecked) and to the free function `RemoveValidatedPagedCacheBlockTable`.

Worth noting that this PR applies exactly the right hardening one function below, in `RollbackReservedBlocks`:

```cpp
if (block && block->Id() < blocks_.size() && blocks_[block->Id()] == block) {
```

Suggest doing both: document the precondition on each `*Validated` declaration, and apply the same cheap guard so a misuse degrades to a no-op instead of memory corruption.

src/engine/cache_manager.cpp
```diff
@@ -236,14 +379,58 @@ void PagedCacheManager::PrepareStep(
...
+                  allocated_to_remove.end(),
+                  request) != allocated_to_remove.end()) {
+      continue;
+    }
+    allocated_to_remove.push_back(request);
+    if (fixed_state_pool_) {
+      // A committed allocation always owns a committed fixed slot (they are published together), so
+      // this handle resolves; collect them all before releasing so a stale handle cannot leave the
+      // two states partially released.
+      fixed_handles.push_back(fixed_state_pool_->HandleFor(request.get()));
```
tianleiwu, line 422: `HandleFor()` throws when the request owns no committed slot, and `ValidateRelease()` below throws from `EnsureIdle()` when a reservation is live. Either throw escapes `Deallocate()` before a single paged block table is freed, so the whole removal batch leaks.

This directly conflicts with the rationale added to `FixedStatePool::Release()` in this PR:

> Teardown depends on this -- a fatal device error marks both the pool and the Engine unhealthy, and the Engine must still be able to remove/reap its requests (releasing paged blocks and this fixed slot) without the release itself throwing and masking the recorded fatal error.

The non-throwing release was made non-throwing on purpose, but the caller reintroduces the throw.

It is reachable today, not just theoretically: `PagedCacheManager::Allocate()` (still live, used by component tests and the non-transactional `Step()` path) pushes into `cache_allocated_requests_` and calls `key_value_cache_->Add()` without ever reserving a fixed slot. For a composite model the next `Deallocate()` throws from this line.

Suggested fix -- the pool already exposes the non-throwing counterpart this PR adds:

```cpp
if (fixed_state_pool_ && fixed_state_pool_->OwnsCommittedSlot(request.get())) {
  fixed_handles.push_back(fixed_state_pool_->HandleFor(request.get()));
}
```

If you would rather keep the strict invariant, then `Allocate()` should reject composite models explicitly so the two admission paths cannot diverge silently. Either way, a test covering removal of a request with no committed fixed slot (and removal after the pool is unhealthy) would lock the intended behavior down.

src/engine/cache_manager.cpp
```diff
@@ -259,9 +446,74 @@ bool PagedCacheManager::IsResident(const std::shared_ptr<Request>& request) cons
...
+  plan.fixed_state = {};
+  auto result = key_value_cache_->PlanStepResources(plan);
+  if (!fixed_state_pool_) {
+    return result;
+  }
+
+  // Paged and fixed ownership advance together at commit, so their committed counts must agree
+  // between steps. A divergence means a prior commit or removal left them out of step.
+  // Uses cheap pool accessors (no per-step snapshot allocation) since planning runs every step.
+  if (fixed_state_pool_->CommittedSlotCount() != cache_allocated_requests_.size()) {
```
tianleiwu, line 474: This sweep is O(n^2) on the per-step hot path. For every entry in `cache_allocated_requests_` it performs three linear scans -- `OwnsCommittedSlot()` (linear over slots), `PagedKeyValueCache::CommittedSlots()` (linear over block tables), and `HandleFor()` (linear over slots again) -- plus another `OwnsCommittedSlot()` per planned row below.

The comment is right that it avoids a snapshot allocation, but allocation was not the only cost. At a `max_batch_size` in the hundreds this is on the order of 10^5 pointer comparisons every single step, for a check that detects corruption rather than doing functional work.

Two options that keep the guarantee:

- Build one `std::unordered_map<const void*, const Slot*>` (or have the pool expose a committed-owner view) and do a single pass, dropping the repeated `HandleFor` lookup since `CommittedTokens` could take the request id directly.
- Gate the full cross-check behind the existing invariant-validation level, keeping only the cheap `CommittedSlotCount() != cache_allocated_requests_.size()` comparison unconditional.

src/engine/step_plan.h
```diff
@@ -60,12 +64,24 @@ struct RequestStepPlan {
...
 };
 
+// Fixed decoder-state demand for a step, planned atomically with the paged-block demand so the
+// Engine can prove the reservation matches the plan before any state is published. Row order
+// mirrors StepPlan::requests exactly. `required` is false (and every count zero) when the model has
+// no fixed groups, which keeps the dense paged path unchanged.
+struct FixedStateResourcePlan {
+  bool required{};          // The plan needs fixed slots this step (the model has fixed groups).
+  size_t row_count{};       // Fixed rows the reservation must expose, one per scheduled request.
+  size_t new_slot_count{};  // Rows that admit a fresh request and consume a free fixed slot.
```
tianleiwu, line 74: `new_slot_count` is computed in `PlanStepResources` and bounds-checked against `AvailableSlots()`, but unlike `required`, `row_count`, and `staging_bytes` it is never verified against the actual reservation in `Engine::StepDynamic`. Its only other readers are the tests.

That makes it look like it participates in the plan-vs-reservation contract when it does not. Either cross-check it (the reservation knows which rows are provisional, so a provisional-row count comparison would be exact and cheap), or drop the field so the plan does not carry unvalidated data.

src/engine/scheduler.cpp
```diff
@@ -224,9 +227,13 @@ StepPlanningResult DynamicBatchScheduler::PlanStep(StepPlan& plan) {
...
 
   const auto& dynamic_batching = *model_->config_->engine.dynamic_batching;
-  plan.scheduled_request_limit = DecodeFirstProvisionalRequestLimit(
-      dynamic_batching.max_scheduled_tokens,
-      dynamic_batching.max_batch_size);
+  try {
+    plan.scheduled_request_limit = DecodeFirstProvisionalRequestLimit(
+        dynamic_batching.max_scheduled_tokens,
+        dynamic_batching.max_batch_size);
+  } catch (const std::invalid_argument& error) {
```
tianleiwu: `DecodeFirstProvisionalRequestLimit` throws `std::invalid_argument` for exactly one reason:

```cpp
if (max_scheduled_tokens == 0 || max_batch_size == 0)
  throw std::invalid_argument("Decode-first scheduling limits must be positive.");
```

Both inputs come straight from `config.engine.dynamic_batching` and never change during a session. So this is a static configuration error, not state corruption -- yet translating it to `StepPlanningConsistencyError` routes it through `MarkUnhealthyAndThrow`, permanently poisoning the Engine on the very first `Step()` with the message "Dynamic step planning failed a cache/state consistency check". That mis-describes the cause and makes the misconfiguration harder to diagnose.

Suggest validating these two values once at config load or Engine construction, and keeping the `StepPlanningConsistencyError` translation for the genuinely state-dependent throws in `AllocateDecodeFirstTokenBudget` (candidate count exceeding the budget, zero pending tokens), which really are consistency failures.

src/models/model_state_manifest.cpp
```diff
@@ -296,6 +296,10 @@ void ModelStateManifest::ValidateDynamicEngineCompatibility(const Decoder& decod
     throw std::runtime_error(
         "Dynamic batching requires exactly one paged_kv decoder state group");
   }
+  if (has_fixed_group) {
```
tianleiwu, line 300: Nit: moving the fixed-group throw after the paged-group-count check changes error precedence. A model that declares fixed groups **and** zero or multiple `paged_kv` groups now reports "requires exactly one paged_kv decoder state group" where it previously reported "does not yet support fixed decoder state groups".

The reorder looks deliberate (the fixed rejection is now the last gate before the composite path opens), but nothing in the diff says so. A one-line comment on why the fixed check must run last would keep the intent from being lost.

tianleiwu (commented):
### Summary

The transaction design is sound: the three-phase `ValidateCommit` / `PrepareCommit` / `Commit` split, with all fallible device work isolated into `PrepareCommit` and publication kept allocation-free, is the right shape for coordinating two ownership systems at one token boundary. The Engine-side plan-vs-reservation cross-check (required flag, row count, staging bytes, per-row request identity) closes a real class of silent-mismatch bugs, and routing `StepPlanningConsistencyError` through `MarkUnhealthyAndThrow` instead of letting it escape with the Engine still healthy is the right containment decision.

A few things worth calling out as genuinely good:

- Allocating the fixed banks **before** constructing `PagedKeyValueCache` so `ComputeNumBlocks` observes the reduced free memory, instead of duplicating the pool's geometry accounting in the paged sizing path.
- `allocated_requests_.reserve()` in the composite reservation constructor, which is what actually makes the "publication cannot throw" claim hold rather than just asserting it in a comment.
- `BlockPool::RollbackReservedBlocks()` now checking `blocks_[block->Id()] == block` and only bumping `mutation_generation_` when something was really released. The previous unconditional bump invalidated generation checks for no-op rollbacks.
- Rejecting a static (non-symbolic) fixed batch axis at construction with a specific message rather than failing fatally on the first row-count mismatch.
- Test coverage is strong for a change this size: staged-bank invisibility across Engine steps, aggregate three-request admission, completion-removal-then-readmission, and the composite invariant suite.

### Main concerns

Two issues are worth addressing before merge, both left as inline comments:

1. **The new `noexcept` `*Validated` publication helpers index memory with undocumented, unchecked preconditions.** `BlockPool::FreeValidated`, `FixedStatePool::ReleaseValidated`, and `RemoveValidatedPagedCacheBlockTable` are all public and `noexcept`, so a caller that skips the matching `Validate*` gets a heap write rather than an exception. Notably, this PR *did* add exactly the right cheap guards to `RollbackReservedBlocks()` but not to its siblings.

2. **`PagedCacheManager::Deallocate()` can throw on the path that `FixedStatePool::Release()`'s new comment says must not throw.** The comment justifies dropping `EnsureHealthy()` so post-fatal teardown can still reclaim state without masking the recorded error, but `Deallocate()` reaches the pool through `HandleFor()` (throws when no committed slot) and `ValidateRelease()` -> `EnsureIdle()` (throws when a reservation is live), and either aborts the batch before a single paged block table is freed.

### Smaller items

- `plan.fixed_state.new_slot_count` is planned and bounds-checked against `AvailableSlots()`, but unlike `required` / `row_count` / `staging_bytes` it is never verified against the reservation by the Engine. Its only other readers are tests.
- The composite consistency sweep in `PlanStepResources` is O(n^2) per step.
- `scheduler.cpp` now converts pure config errors into permanent Engine unhealthiness with a message that mis-describes the cause.
- `ValidateDynamicEngineCompatibility` reorders error precedence without explanation.

### Test gap

There is no test for `Deallocate()` on a composite manager whose request lacks a committed fixed slot, nor for `Release()` / `Deallocate()` after the pool has been marked unhealthy -- which is precisely the scenario the new `FixedStatePool::Release()` comment claims to support.

src/engine/block.h
baijumeswani: Documented the validate before publish precondition. And added defensive guards.

src/engine/cache_manager.cpp
baijumeswani: The hotpath should not have the O(n^2) sweep. Keeping this sweep in this PR for correctness. Will address this in the next PR in this stack: https://github.com/microsoft/onnxruntime-genai/pull/2438

src/engine/cache_manager.cpp
baijumeswani: Addressed. I kept the strict composite invariant rather than silently skip a missing fixed slot. Transactional admission remains the only valid composite path.

src/engine/step_plan.h
baijumeswani: `FixedStateReservation` now reports its actual provisional/new-slot count through `CacheStepReservation`, and `Engine::StepDynamic()`

src/engine/scheduler.cpp
baijumeswani: Addressed.  I removed the `StepPlanningConsistencyError` translation around `DecodeFirstProvisionalRequestLimit()`

src/models/model_state_manifest.cpp
baijumeswani: Added a comment at the check to highlight the ordering intent.

src/engine/cache_manager.cpp
```diff
@@ -236,14 +391,58 @@ void PagedCacheManager::PrepareStep(
...
+  }
+
+  for (const auto& handle : fixed_handles) {
+    fixed_state_pool_->ReleaseValidated(handle);
+  }
+  for (const auto& request : allocated_to_remove) {
+    key_value_cache_->RemoveValidated(request.get());
   }
 
   cache_allocated_requests_.erase(
```
tianleiwu, line 443: **Suggestion — a tripped defensive guard silently orphans state.**

`ReleaseValidated` and `RemoveValidated` are `noexcept` and return silently when their internal guards reject the handle/table. But this `erase` runs unconditionally afterwards, so if either guard ever trips the result is:

- the paged block table stays in `block_tables_` and its blocks stay allocated, forever;
- the fixed slot stays `Committed` with a stale `request_id`, forever;
- the request is gone from `cache_allocated_requests_`, so the next `PlanStepResources` consistency sweep compares `CommittedSlotCount()` against a now-smaller set and reports a divergence that points at the wrong step.

And none of it is observable: no log, no assert, no health flag.

The guards should be unreachable — `ValidateRelease`/`ValidateRemove` ran immediately above with no intervening mutation — which is exactly why a trip is a programming error worth reporting rather than swallowing. Consider having the validated publication helpers report whether they actually published, and logging an error (or asserting in debug) here when the count of published releases does not match `allocated_to_remove.size()` / `fixed_handles.size()`, before erasing.

src/engine/paged_cache_reservation.cpp
```diff
@@ -97,9 +108,34 @@ void RemovePagedCacheBlockTable(
...
+        return candidate.RequestId() == request_id;
+      });
+  if (table == committed_tables.end()) {
+    return;
+  }
+
+  if (!block_pool.CanFreeValidated(table->Blocks()) ||
+      (window_block_pool &&
+       !window_block_pool->CanFreeValidated(table->WindowBlocks()))) {
+    return;
```
tianleiwu: Same silent-failure shape as the `Deallocate` comment: this `return` leaves the table in `committed_tables` with its blocks still owned, and the caller has no way to learn that removal did not happen. Since reaching here means `ValidateRemovePagedCacheBlockTable` and this function disagreed with no intervening mutation, it is a logic error rather than a recoverable condition.

Suggest returning a `bool` (published / not published) so `PagedCacheManager::Deallocate` can log and skip the corresponding `cache_allocated_requests_` erase, keeping the two ownership sets consistent even in the impossible case.

tianleiwu (commented):
## Review round 2 (head `5b4c9d5a`)

All six threads from the previous round on `34bbb679` were verified against this head and are addressed:

- `block.h` `FreeValidated` now guards through `CanFreeValidated` (null, bounds, ownership, duplicate).
- `Deallocate` now skips requests that own no committed state, so `HandleFor`/`ValidateRelease` cannot throw during removal.
- `new_slot_count` is now cross-checked via `CacheStepReservation::FixedStateNewSlotCount()` in `Engine::StepDynamic()`.
- The `StepPlanningConsistencyError` translation was correctly narrowed to `AllocateDecodeFirstTokenBudget` and dropped from the static-configuration path.
- The state-group error precedence comment was added.
- The O(n^2) planning sweep is acknowledged and deferred to the next PR in the stack.

### Assessment of this head

The validate/prepare/publish protocol is sound. Every fallible operation (device staging, capacity preflight, ownership and generation checks) is completed before the publication boundary, and the publication tail is allocation-free. `PagedCacheReservation::CommitValidated()` re-preflights membership, generations, occupancy, and vector capacity before mutating the first table, so a throw there leaves both systems unpublished — which makes the paged-then-fixed publish order in `CompositeCacheStepReservation::Commit()` correct rather than merely documented. `~PagedCacheReservation()` being `noexcept` with a direct rollback, and `~FixedStateReservation()` discarding uncommitted state, together mean the `throw std::logic_error(...)` inside the Engine's reservation-validation block is fully cleaned up by RAII on the way out. I verified `FixedStatePool::PlannedStagingBytes(row_count)` reproduces the `CheckedMultiply`/`CheckedAdd` accumulation in `Reserve()` exactly, so the Engine's strict staging-byte equality check cannot false-positive.

No correctness blocker. The fixed-state path is still closed by `ValidateDynamicEngineCompatibility`, so nothing here reaches production dynamic decoding yet.

### Cross-cutting item that cannot be anchored to the diff

**Per-step staging allocation and two device barriers** (Suggestion, pre-existing code newly activated by this path). `FixedStatePool::Reserve` creates two `OrtValue` tensors per fixed binding on every scheduled step and then calls `DeviceInterface::Synchronize()` after the gather; `PrepareCommit` synchronizes a second time after staging outputs. Once the compatibility gate opens, every decode iteration pays allocator churn plus two host-visible barriers, which conflicts with the zero-allocation inference-loop goal and will likely dominate short decode steps. Consider preallocating max-batch staging storage with the pool and handing out per-reservation row views, and using stream-ordered events instead of full device synchronization. Not a merge blocker while the gate is closed, but worth resolving before it is removed.

### Test coverage

The 17 new engine step tests cover commit, execution / post-processing / chunk-failure rollback, capacity backpressure, removal isolation, staged-output invisibility, aggregate admission, and all four plan/reservation mismatch shapes, against the real `PagedCacheManager` with only the executor doubled. That is the right level. As the artifact notes, the synthetic composite model never runs `VarlenDecoderIO`, so an end-to-end test that actually binds fixed inputs/outputs is still required when the gate is removed.

Inline comments below: one robustness suggestion and several nitpicks.

src/engine/engine_invariants.cpp
```diff
@@ -254,6 +318,67 @@ std::vector<InvariantViolation> ValidateInvariants(const PagedCacheSnapshot& cac
...
+    if (slot.ownership == FixedStateSlotOwnership::Committed) {
+      fixed_owners.emplace(slot.request_id, &slot);
+    }
+  }
+  std::unordered_map<const void*, const RequestStateSnapshot*> request_states;
+  for (const auto& request : requests) {
+    request_states.emplace(request.request_id, &request);
+  }
+
+  for (const auto& [request_id, paged] : paged_owners) {
```
tianleiwu: Nit: both cross-ownership loops iterate `std::unordered_map`, so the order of the emitted `InvariantViolation` messages varies between runs (and between libstdc++/libc++/MSVC). `ThrowIfCompositeStateInvariantsViolated` concatenates them into one message, so a test that asserts on message content or ordering — or a bug report pasting the message — becomes non-reproducible.

Iterating `cache.requests` and `fixed.slots` directly (looking up in the maps rather than iterating them) keeps the snapshot's deterministic order.

src/engine/cache_manager.h
```diff
@@ -145,15 +165,25 @@ struct PagedCacheManager : CacheManager {
 
   PagedCacheSnapshot Snapshot() const override { return key_value_cache_->Snapshot(); }
 
-  StepPlanningResult PlanStepResources(StepPlan& plan) const override {
-    return key_value_cache_->PlanStepResources(plan);
+  PagedCacheSnapshot Snapshot(const PagedCacheReservation* reservation) const override {
```
tianleiwu, line 168: Nit: this `Snapshot(const PagedCacheReservation*)` overload has no callers — not in `src/`, and not in the new `engine_step_tests.cpp` / `engine_invariants_tests.cpp` either, which all use the no-arg `Snapshot()`. If it is groundwork for the packed-IO follow-up, a one-line note saying so would help; otherwise it is worth dropping until something consumes it.

src/engine/cache_manager.cpp
```diff
@@ -176,8 +297,34 @@ bool StaticCacheManager::IsResident(const std::shared_ptr<Request>& request) con
...
 PagedCacheManager::PagedCacheManager(std::shared_ptr<Model> model)
     : CacheManager(model),
-      params_(std::make_shared<GeneratorParams>(*model_)),
-      key_value_cache_(std::make_unique<PagedKeyValueCache>(model)) {
+      params_(std::make_shared<GeneratorParams>(*model_)) {
+  // The paged cache resolves its own paged_kv group. The fixed pool is created only when the
+  // decoder manifest declares fixed state groups, so a dense paged model owns no pool and
+  // the composite path degrades to paged-only. Its capacity matches the paged batch limit so paged
+  // admission (bounded by max_batch_size) can never outrun fixed slots.
+  ModelStateManifest manifest{model->config_->model.decoder};
```
tianleiwu, line 308: Nit: `ModelStateManifest`'s constructor copies `state_groups`, `inputs`, and `outputs` (all string-heavy structs) and re-runs `ValidateConfig(decoder)`, but the only thing needed here is "does any group have kind `FixedConv` or `FixedRecurrent`?". `HasFixedStateGroups` could take `const Config::Model::Decoder&` (or the `state_groups` optional) directly and skip both the copy and the redundant validation.

One-time cost at model load, so purely cosmetic — but it also means a config-validation failure now surfaces from the cache-manager constructor, which is a slightly surprising place for it.

src/engine/cache_manager.cpp
```diff
@@ -3,6 +3,7 @@
 
 #include "cache_manager.h"
 
+#include <algorithm>
```
tianleiwu, line 6: Nit: this file now uses `std::exception_ptr` / `std::current_exception` / `std::rethrow_exception` (`<exception>`), `std::logic_error` / `std::runtime_error` (`<stdexcept>`), and `std::to_string` (`<string>`), all reaching it transitively. Adding the direct includes alongside `<algorithm>` keeps it robust against future header churn in `cache_manager.h`'s dependency chain.

src/engine/block.h
```diff
@@ -92,7 +99,10 @@ struct BlockPool {
 
   const size_t block_size_;
   const size_t capacity_;
-  std::vector<std::shared_ptr<Block>> blocks_{capacity_};
+  std::vector<std::shared_ptr<Block>> blocks_;
+  // Preallocated scratch for allocation-free duplicate detection in validated publication.
+  mutable std::vector<uint64_t> validation_marks_;
```
tianleiwu, line 106: Nit: `CanFreeValidated()` is declared `const noexcept` but writes `validation_marks_` and `validation_epoch_`. That is fine for the single-threaded Engine, but it breaks the standard convention that `const` member functions are safe for concurrent reads — which the other `const` accessors on this type (`AvailableBlocks()`, `MutationGeneration()`) do honor. Since `ValidateFree()`/`CanFreeValidated()` are public, a short comment stating that these two members make the `const` overloads single-thread-only would prevent someone from calling them from an inspection/telemetry path later.

test/cpp/engine/engine_test_doubles.h
```diff
@@ -367,5 +466,23 @@ inline DoublesEngine MakeDoublesEngine(std::shared_ptr<Model> model, size_t capa
   return DoublesEngine{std::move(engine), cache_observer, executor_observer, std::move(trace)};
 }
 
+inline CompositeDoublesEngine MakeCompositeDoublesEngine(std::shared_ptr<Model> model,
+                                                         int32_t forced_token) {
+  // Bypass the public compatibility gate so this PR can exercise the complete composite resource
+  // manager before the following packed-IO PR enables fixed-state model execution.
+  auto cache = std::make_shared<PagedCacheManager>(model);
+  auto* cache_observer = dynamic_cast<PagedCacheManager*>(cache.get());
```
tianleiwu: Nit: `cache` is already `std::shared_ptr<PagedCacheManager>`, so this `dynamic_cast` is a tautology and the `if (!cache_observer) throw` below it is dead code. `auto* cache_observer = cache.get();` is equivalent.

tianleiwu (approved):

src/engine/cache_manager.cpp
baijumeswani: I changed the validated block, fixed-slot, and paged-table publication guards to fail stop with `std::terminate()` on disagreement, with misuse covered by death tests.

src/engine/paged_cache_reservation.cpp
baijumeswani: Addressed with the composite deallocation fix. An ownership disagreement after successful prevalidation now terminates rather than returning silently, so request membership cannot be erased while blocks remain owned.

src/engine/engine_invariants.cpp
baijumeswani: Cross-ownership diagnostics now iterate `cache.requests` and committed `fixed.slots` in snapshot order, using the maps only for lookup.

src/engine/cache_manager.h
baijumeswani: The Snapshot is in use in the engine_step_tests.cpp. Retaining it.

src/engine/cache_manager.cpp
baijumeswani: Added direct `<exception>`, `<stdexcept>`, and `<string>` includes so this translation unit no longer depends on transitive includes from `cache_manager.h`

test/cpp/engine/engine_test_doubles.h
baijumeswani: Replaced it with `auto* cache_observer = cache.get()` and removed the unreachable null-check branch.

src/engine/block.h
baijumeswani: Addressed this.

src/engine/cache_manager.cpp
baijumeswani: Will address this as it becomes a problem. Keeping it as is for now.

PR #2440, opened 2026-08-21
Fix model builder for paged attention with sliding-window layers

GopalakrishnanN:
## Problem

Exporting a model that combines **paged attention** with **alternating sliding-window / full attention** (e.g. `openai/gpt-oss-20b`, `gpt-oss-120b`) fails in two separate places.

### 1. `AttributeError` in `make_key_value_cache_shape()`

The method assumed the cache sequence dimension is always a symbolic string and called `.replace()` on it. Paged caches use a fixed **integer** block dimension, so sliding-window layers raised:

```
AttributeError: 'int' object has no attribute 'replace'
```

Paged caches use the same physical block shape for every layer — locality is expressed by the `PagedAttention` operator, not by a different cache sequence dimension. The shape is now returned unchanged when paged attention is enabled.

### 2. ORT rejects the exported model

`sliding_window_cache` was emitted for sliding-window layers regardless of attention type, but `PagedAttention` does not accept it:

```
Unrecognized attribute: sliding_window_cache for operator PagedAttention
```

That attribute describes a window-sized past/present buffer, which `PagedAttention` does not have since the engine owns the block-managed cache. It is now emitted only on the non-paged path.

## Repro

```
python src/python/py/models/builder.py -m openai/gpt-oss-20b -o <out> -p int4 -e cuda \
  --extra_options use_paged_attention=true moe_quant_type=mxfp4 is_symmetric=true \
  paged_block_size=256
```

Fails at (1); after fixing (1), fails at (2) when loading the model.

## Tests

Added to the existing builder suites:

- `test_paged_attention_keeps_physical_block_shape_on_sliding_layers` — the paged path returns `["num_blocks", 256, 2, 16]` unchanged
- `test_paged_attention_preserves_sliding_window_size` — asserts no `sliding_window_cache` attribute is emitted when paged attention is enabled

Both fakes now set `use_paged_attention`, so the non-paged behavior stays covered.

```
$ python -m pytest test/python/builder/ -q
325 passed, 3 skipped
```

## Validation

Beyond unit tests, the resulting 24-layer gpt-oss-20b export was verified end to end through the continuous-batching Engine on an A100-80GB: single and 4-way concurrent requests with ragged prompts, **exact output parity** between concurrent and isolated runs (0 mismatches), repeat-stable across runs, and coherent generation.

### DIFF
--- a/src/python/py/models/builders/base.py
+++ b/src/python/py/models/builders/base.py
@@ -1042,7 +1042,13 @@ class Model:
         Modifies KV cache shape dimension names for models with alternating attention patterns.
         Sliding window layers get a distinct symbolic sequence dim so ONNX shape inference does not
         unify them with the full-attention layers, whose cache is allocated at max_length.
+
+        Paged caches use the same physical block shape for every layer; local attention is expressed
+        by the PagedAttention operator instead of a different cache sequence dimension.
         """
+        if self.use_paged_attention:
+            return shape
+
         if (
             self.ep in self.eps_with_windowed_kv_cache
             and hasattr(self, "is_local")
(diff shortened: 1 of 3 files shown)


tianleiwu:
See related PR: 
https://github.com/microsoft/onnxruntime-genai/pull/2358/changes

PR #2442, opened 2026-08-21
Add GatedDeltaNet evaluation export for Qwen hybrid models

baijumeswani:
## Summary

This PR adds a GatedDeltaNet export path for Qwen hybrid recurrent layers.

It replaces `VarlenLinearAttention` with the ONNX Runtime `GatedDeltaNet` contrib operator while preserving the same upstream normalization, Q/K/V layout, causal convolution, effective gates, recurrent-state layout, and model weights. This keeps model-level comparisons focused on the recurrent operator.

## What changed

- Added a builder helper that emits `com.microsoft::GatedDeltaNet`.
- Uses packed THD Q/K/V inputs and cumulative sequence lengths from the existing varlen export.
- Preserves the FP32 V-major recurrent-state contract.
- Uses the reviewed `gated_delta` update rule.
- Passes precomputed effective decay and beta gates with activation attributes set to `none`.
- Casts effective gates to the operator's FP32 gate ABI.
- Disables state-checkpoint outputs because the Engine commits final state transactionally.
- Rejects BF16 model I/O because the current ONNX Runtime kernel supports FP16 and FP32 only.
- Adds builder tests for node inputs, outputs, attributes, gate casts, state geometry, update-rule selection, and supported dtypes.

## Evaluation contract

This path is intentionally conservative. It reuses the same precomputed gates as the varlen graph so the comparison isolates GatedDeltaNet, but casting FP16 gates to FP32 cannot recover precision already lost in the shared FP16 path.

A production exporter should use GatedDeltaNet's raw Qwen decay and sigmoid-beta fusion inputs directly. That would avoid the precision round-trip and remove separate gate kernels.

## Dependency

This branch depends on ONNX Runtime PR #32182 and the reviewed local fixes for:

- scalar-per-head `dt_bias` validation;
- dynamic shared-memory setup;
- forced-plan behavior;
- supported dtype advertising;
- update-rule input validation;
- state alias declaration.

The exporter selects GatedDeltaNet unconditionally, so this should remain an evaluation PR until the ONNX Runtime dependency is available. A production version should add an explicit backend option or use a coordinated ONNX Runtime version boundary.


tianleiwu (commented):
Thanks — I built this stack end-to-end and it works. Everything below is about how
it lands next to #2453, plus one rebase hazard that lives lower in the stack.

**What I validated** (this branch + #2453, on ORT #32182 + #32168; Qwen3.8-27B-NVFP4, H200):

- 64-layer paged export: 16 `PagedAttention`, 0 `GroupQueryAttention`, 48 `GatedDeltaNet`,
  48 `VarlenCausalConvWithState`; `input_ids [num_tokens]`, no `attention_mask`;
  `state_groups` = 1 `paged_kv` (16 layers) + 2 fixed (48 each).
- **Ragged-batch correctness**: a batch of 17 / 512 / 1993-token prompts produced
  bit-identical output to the same prompts run alone. That is the property packed
  conv/recurrent state is most likely to get wrong, and it holds.
- Quality is statistically indistinguishable from the dense export under an identical greedy
  recipe: MMLU-Pro 800 **83.25% vs 83.00%** (McNemar exact p=0.88), GPQA-diamond
  **75.25% vs 76.26%** (p=0.86), 0 unextracted answers.
- Throughput: 5,969 tok/s prefill at 2k (vs 914 dense), decode 343 tok/s at batch 8,
  587 tok/s at batch 64, 24 tok/s at a 256k context.

Three inline comments cover the #2453 reconciliation. Three further notes that do not map to
a line in this PR's diff:

**1. A rebase hazard that belongs to #2427, not here.** `_make_config_model` in
`test/python/builder/test_decoder_state_groups.py:61` builds its model with `__new__` and
never sets `use_windowed_paged_kv_cache`. That attribute did not exist when #2427 was
branched, but #2358 has since landed on `main` and made `make_genai_config` call
`has_windowed_paged_layers()` for every paged build. I rebased this stack onto current
`main` and got four failures:

```
AttributeError: 'Qwen35TextModel' object has no attribute 'use_windowed_paged_kv_cache'
  test_common_paged_builder_preserves_legacy_manifest_absence
  test_common_nonpaged_builder_preserves_manifest_absence
  test_qwen_all_attention_builder_preserves_legacy_manifest_absence
  test_qwen38_official_geometry_emits_exact_sparse_groups
```

Adding `model.use_windowed_paged_kv_cache = False` beside `model.window_size = None` fixes
all four; `False` is right because Qwen3.8 has no sliding window. Flagging it here because
this is the PR whose export it blocks, but the one-line change belongs at the bottom of the
stack.

**2. `_setup_hybrid_cache_io` needs its state-dtype rule consolidated with #2453.** This PR
keys V-major FP32 recurrent state off `use_paged_attention`; #2453 keys it off
`linear_attn_op == "gated_delta_net"`. Merging both naively leaves two competing assignments
to `recurrent_state_dtype` and a stale `state_dtype` reference. The single rule that worked
for me was `v_major_state = self.use_paged_attention or self.linear_attn_op == "gated_delta_net"`,
feeding both the dtype and the trailing-extent swap.

**3. Unrelated but it affects every paged build.** The native FP8/NVFP4 MatMul emitters in
`base.py` hardcode `["batch_size", seq_dim, out_features]` rather than going through
`hidden_state_shape()`. With `use_paged_attention=true prune_lm_head=true` the LM head then
declares `logits` as `[batch_size, batch_size, vocab]`. Already fixed in #2449, so it
resolves itself if that merges first — noting it so it is not diagnosed twice.

I have the full reconciliation applied and tested in a combined tree, and am happy to hand
over the patch or open a PR against this branch.

src/python/py/models/builders/base.py
```diff
@@ -3374,6 +3420,87 @@ def make_linear_attention(self, name, **kwargs):
...
+            update_rule=kwargs.get("update_rule", "gated_delta"),
+            scale=kwargs.get("scale", 1.0),
+            decay_activation=kwargs.get("decay_activation", "none"),
+            beta_activation=kwargs.get("beta_activation", "none"),
+            max_checkpoints=0,
+        )
+        self.make_value(output, self.io_dtype, shape=kwargs["output_shape"])
+        self.make_value(present_recurrent, ir.DataType.FLOAT, shape=kwargs["present_recurrent_shape"])
+
+    def make_gated_delta_net(self, name, **kwargs):
```
tianleiwu: This name collides with #2453, which adds a `make_gated_delta_net` to the same
class with a different contract:

| | this PR | #2453 |
|---|---|---|
| layout | packed THD, `cumulative_sequence_length` required | rank-4 `[B, S, H, D]`, `cu_seqlens` optional |
| gates | FP16 pre-cast to FP32, `gate_activation="none"` | raw Qwen projections, fused `gate_activation="qwen"` / `beta_activation="sigmoid"` |
| state | `past`/`present_recurrent_state`, `state_checkpoints=0` | `initial_state`/`final_state`/`checkpoints`, window up to 8 |

These are two different uses of the operator, not one method with options, so whichever
merges second silently overwrites the first.

Suggest renaming this one `make_varlen_gated_delta_net` — it also matches the
`make_varlen_causal_conv_with_state` / `make_varlen_linear_attention` spelling already in
this file. Both then coexist and `_make_linear_attention` selects by layout. I ran that
combination through a full 64-layer export plus the builder suite; nothing else on this side
needed to change.

src/python/py/models/builders/qwen.py
```diff
@@ -1812,75 +1962,140 @@ def _make_linear_attention(self, layer_id, linear_attn, root_input):
...
+                    ["num_tokens", heads, dim],
+                )
+                return f"{reshape_name}/output_0"
+
+            q_thd = reshape_thd(q_scaled_output, "q", n_k, hk)
+            k_thd = reshape_thd(k_norm_out, "k", n_k, hk)
+            v_thd = reshape_thd(v_out, "v", n_kv, hv)
+            present_recurrent_shape = ["batch_size", n_kv, hv, hk]
+            la_op_name = f"{basename}/GatedDeltaNet"
+            self.make_gated_delta_net(
```
tianleiwu: Once #2453 merges, this branch becomes unreachable for exactly the
configuration it exists to serve.

#2453 adds an early return to `_make_linear_attention` **above** this block:

```python
# Split QKV, L2 norm, gates
if self.linear_attn_op == "gated_delta_net":
    la_output = self._make_gated_delta_net(layer_id, linear_attn, conv_out_t_output, b_name, a_name)
    self._make_linear_attention_output(layer_id, linear_attn, la_output, z_name)
    return
```

A paged build with `linear_attn_op=gated_delta_net` would take the dense rank-4 path and
never reach here. It also breaks outright, because #2453 passes `conv_out_t_output` while
this PR renames that variable to `conv_out`.

Guarding the dense path is enough:

```python
if not packed and self.linear_attn_op == "gated_delta_net":
```

src/python/py/models/builders/base.py
```diff
@@ -3374,6 +3420,87 @@ def make_linear_attention(self, name, **kwargs):
         self.make_value(output, self.io_dtype, shape=kwargs["output_shape"])
         self.make_value(present_recurrent, self.io_dtype, shape=kwargs["present_recurrent_shape"])
 
+    def make_varlen_linear_attention(self, name, **kwargs):
```
tianleiwu: This method has no production call site. The packed path calls
`make_gated_delta_net` instead (`qwen.py:2052`); outside its own docstring the only
reference to `make_varlen_linear_attention` is `test_decoder_state_groups.py:211`.

It emits `com.microsoft.VarlenLinearAttention`, which exists only in ORT #32166 — currently
unmerged and in "request changes". As it stands this adds a builder method for an operator
no exported model reaches and no shipped ORT provides.

Worth either dropping it here or saying in the PR description that it is staged ahead of
#32166, so a reader does not assume the packed path uses it.

src/python/py/models/builders/base.py
baijumeswani: Changed to make_varlen_gated_delta_net

baijumeswani:
Superseded by https://github.com/microsoft/onnxruntime-genai/pull/2453. Closing this PR.

PR #2450, opened 2026-08-22
Optimize greedy multi-token MTP scheduling

tianleiwu:
## Summary

Remove redundant target replay from greedy multi-token MTP and keep CUDA verification inputs on device. The non-windowed recurrent-state fallback and sampling path remain unchanged.

## Key changes

- Commit accepted tokens directly from the wide target verification when recurrent state can be cropped.
- Avoid the unconditional state snapshot and the partial-accept target replay.
- Reuse persistent pinned and device buffers for chained draft tokens and target verification input.
- Assemble verify tokens with stream-ordered device copies and submit target verification before host draft readback.
- Accumulate target-row argmax results on device and copy them to the host once per round.
- Disable implicit EP-end synchronization only for the internal CUDA greedy N>1 states whose input lifetimes are now persistent.

## Performance

On H200 with a Qwen3.8-27B model, a 2K prompt and 2K greedy generated tokens at N=3 improved from 62.41 to 98.24 decode tok/s after removing snapshot/replay. The device-input path also removed pinned-buffer destructor fences; a graph-stable N=2 profile improved from 152.63 to 154.11 tok/s.

## Validation

- Built `python`, `unit_tests`, and `cuda_batched_sampler_tests` with CUDA.
- Focused MTP/speculative suite: 51 passed.
- `ArgMaxTests.DeviceRowResultsAccumulateBeforeOneHostCopyCuda`: passed.
- `lintrunner` and `git diff --check`: clean.
- Graph-stable sampled reset/replay smoke produced converged output hashes.

### DIFF
--- a/docs/speculative_decoding.md
+++ b/docs/speculative_decoding.md
@@ -236,14 +236,14 @@ stream costs one main forward and one head forward per two tokens.
 
 The single head module is chained `N` times, feeding its own post-norm hidden forward
 (`head_out_hidden_`), the way vLLM's `AutoRegressiveSpeculator` does. Then `[t, d_0 .. d_{N-1}]` is
-verified in one `N+1`-wide main forward, `a` = length of the greedy-matching prefix, and one of four
+verified in one `N+1`-wide main forward, `a` = length of the greedy-matching prefix, and one of three
 finalize branches runs:
 
 | Branch | Condition | Cost |
 |---|---|---|
 | All accepted | `a == N` | No extra forward; bonus is verify row `N` |
-| Lossless crop + `M=1` replay | `a >= 1` and windowed recurrent state | Crop to `L+a`, replay the last committed token (1-wide, decode-consistent bonus) |
-| Snapshot rewind + replay | otherwise | `RewindToLength(L)` + replay `[t, d_0 .. d_{a-1}]` |
+| Windowed direct commit | `a < N` and windowed recurrent state | Crop to `L+a+1`; bonus is verify row `a` |
+| Snapshot rewind + replay | `a < N` otherwise | `RewindToLength(L)` + replay `[t, d_0 .. d_{a-1}]` |
 
 Re-materializing the `a` accepted drafts in the head's KV and drafting the next round's first token
 both consume *main-model* hidden states over consecutive positions, so they are **deferred and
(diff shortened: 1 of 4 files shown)


src/mtp_generator.cpp
```diff
@@ -723,22 +769,31 @@ void MtpGenerator::GenerateStepMulti(int32_t t) {
...
     if (pending_refeed_count_ > 0) {
       const int a_prev = pending_refeed_count_;
       mtp_->RewindToLength(pending_refeed_head_len_);  // drop last step's speculative drafts
       head_len_ = pending_refeed_head_len_;
       merged_tokens_[a_prev] = t;  // last row: the token committed at the top of this step
+      auto head_tokens_cpu = head_tokens_device_.CpuSpan();
+      std::copy_n(merged_tokens_.begin(), a_prev + 1, head_tokens_cpu.begin());
+      head_tokens_device_.CopyCpuToDevice();
+      auto head_tokens = head_tokens_device_.subspan(0, static_cast<size_t>(a_prev + 1));
+      current_token_device = head_tokens_device_.subspan(static_cast<size_t>(a_prev), 1);
```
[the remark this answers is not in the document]
tianleiwu: Fixed in 925ae06. Real head token IDs now upload through a separate persistent buffer and are copied D2D into head_tokens_device_, whose CPU mirror remains non-pad metadata for Logits::Update. This covers both fused and singleton paths, and the CUDA regression includes a trailing pad-valued token.

docs/speculative_decoding.md
baijumeswani, line 1: AI suggested 3 documentation updates:

> - docs/speculative_decoding.md:309-310: "`MtpGenerator` snapshots before each wide greedy verify so a partial rejection can always replay" — false for windowed models now.
> - docs/speculative_decoding.md:439-441 : "the greedy path crops to the last accepted state and replays only that token with an `M=1` forward. When no draft is accepted, it restores the snapshot" — describes exactly the removed behavior.
> - line 437 documents `chunk_size` default `256` on windowed models; code sets `1024` (pre-existing).

baijumeswani (approved):

docs/speculative_decoding.md
tianleiwu: Thanks, I'll update later

PR #2453, opened 2026-08-22
Export compact GatedDeltaNet state updates for Qwen3.5/3.8

tianleiwu:
## Summary
- export packed Qwen3.5/3.8 linear-attention layers with `VarlenCausalConvWithState` and `GatedDeltaNet`
- replace dense recurrent checkpoints with compact convolution values and GDN transition capsules
- export raw `A_log`/`dt_bias` for native Qwen gate arithmetic, without `arithmetic_mode`
- keep committed paged recurrent state unwindowed, FP32, and V-major while preserving the dense `LinearAttention` default

The existing `state_groups` schema is unchanged in this PR. Parsing compact update metadata and replaying these transitions is handled by the stacked runtime PR #2454.

## Dependencies
- microsoft/onnxruntime#32282
- microsoft/onnxruntime#32290
- microsoft/onnxruntime#32307

## Testing
- `python3 -m pytest test/python/builder/test_decoder_state_groups.py test/python/builder/test_qwen_vl.py -q` (50 passed)
- `python3 -m pytest test/python/builder -q` (493 passed, 3 skipped)


src/python/py/models/builders/base.py
```diff
@@ -3726,6 +3767,125 @@
         )
         self.make_value(output, self.io_dtype, shape=["batch_size", "sequence_length", self.linear_value_dim])
 
+    def make_gated_delta_net(self, name, **kwargs):
```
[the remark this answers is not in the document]
tianleiwu: Fixed in c5bda4a8 by renaming the Qwen architecture-level builder to `make_qwen_gated_delta_net`, so it no longer overrides `Model.make_gated_delta_net` with an incompatible signature. The related new helpers were also converted to regular, publicly named instance methods.

src/python/py/models/builders/qwen.py
```diff
@@ -227,47 +321,88 @@ def make_attention_output_proj(self, layer_id, attention, root_input, **kwargs):
...
-        z_name, b_name, a_name, qkv_t_output, conv_weight_name = self.make_linear_attention_input_proj(
+        z_name, b_name, a_name, conv_input, conv_weight_name = self.make_linear_attention_input_proj(
             layer_id, linear_attn, root_input
         )
 
-        # --- Fused conv: CausalConvWithState (com.microsoft) ---
         conv_bias_name = f"model.layers.{layer_id}.linear_attn.conv1d.bias"
         self.make_initializer(torch.zeros(self.linear_conv_dim, dtype=torch.float32), conv_bias_name, to=self.io_dtype)
 
+        if self.use_paged_attention:
```
baijumeswani, line 333: This new paged-hybrid code path assumes the complete qwen layer stack is token-major, but full-attention layers still pass through `make_attention_input_proj()` above, whose Q/gate reshape targets are dense-only.
In paged mode the base projection is rank-2 `[num_tokens, 2 * num_heads * head_size]`, so a real tiny hybrid export fails ORT shape validation at that reshape.

src/python/py/models/builders/qwen.py
```diff
@@ -126,6 +163,63 @@ def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
...
+            ]
+            self.input_types["past.recurrent"] = ir.DataType.FLOAT
+            self.input_shapes["past.recurrent"] = recurrent_shape
+            self.output_types["present.recurrent"] = ir.DataType.FLOAT
+            self.output_shapes["present.recurrent"] = recurrent_shape
+
+        if not self.state_update_capacity:
+            return
+
+        self.input_names["state_update.capture_count"] = "state_update_capture_count"
```
baijumeswani, line 181: The PR emits the existing typed `state_groups`, but those groups describe only committed paged/conv/recurrent state. When `state_update_capacity > 0`, the graph gains required compact-update inputs and per-layer outputs that are not represented in `genai_config.json`, so the runtime cannot discover how to bind or replay them.

src/python/py/models/builders/qwen.py
baijumeswani, line 333: This paged branch unconditionally emits `VarlenCausalConvWithState` and `GatedDeltaNet`, but the common paged validation currently permits WebGPU FP16 and both operators are CUDA-only in their ORT PRs. Should we reject webgpu for this code path?

src/python/py/models/builders/qwen.py
```diff
@@ -112,9 +112,46 @@ def load_weights(self, input_path):
 
 
 class Qwen35TextModel(Model):
+    def validate_state_update_options(self, capacity, use_paged_attention, linear_attn_op, state_window):
+        if capacity < 0 or capacity > 8:
+            raise ValueError("state_update_capacity must be between 0 and 8")
+        if capacity and not use_paged_attention:
+            raise ValueError("state_update_capacity requires use_paged_attention=true")
+        if capacity and linear_attn_op != "gated_delta_net":
```
baijumeswani: In paged mode the branch below always selects `GatedDeltaNet` regardless of `linear_attn_op`, and `configure_gated_delta_net_io()` deliberately exports unwindowed fixed state. Neither packed operator receives `state_window`. Requiring both `linear_attn_op=gated_delta_net` and `state_window >= state_update_capacity + 1` therefore forces users to provide two no-op options; the documented command encodes an invariant the graph does not implement.

src/python/py/models/builders/qwen.py
```diff
@@ -126,6 +163,63 @@ def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
...
+        ]
+        if not linear_layers:
+            return
+
+        if self.use_paged_attention:
+            conv_shape = ["batch_size", self.linear_conv_dim, self.linear_conv_kernel_dim - 1]
+            self.input_shapes["past.conv"] = conv_shape
+            self.output_shapes["present.conv"] = conv_shape
+
+        if self.use_paged_attention or self.linear_attn_op == "gated_delta_net":
```
baijumeswani, line 165: > For non-paged `linear_attn_op=gated_delta_net`, this rewrites recurrent state to unwindowed FP32 V-major, while convolution state remains windowed and `CausalConvWithState` still receives `state_window`. The runtime requires convolution and recurrent state to be either both windowed or both unwindowed; this combination produces incompatible logical state boundaries. Please either implement matching GDN window semantics or reject dense GDN with `state_window > 0` at option validation.

src/python/py/models/builders/base.py
```diff
@@ -3731,6 +3772,130 @@ def make_linear_attention(self, name, **kwargs):
...
+            name=name,
+            domain="com.microsoft",
+            **attributes,
+        )
+        self.make_value(output, self.io_dtype, shape=kwargs["output_shape"])
+        self.make_value(final_state, ir.DataType.FLOAT, shape=kwargs["state_shape"])
+
+    def make_varlen_gated_delta_net(self, name, **kwargs):
+        """Emit packed GatedDeltaNet with optional compact transition capsules."""
+        if self.io_dtype == ir.DataType.BFLOAT16:
```
baijumeswani: Rejecting BF16 is correct for the current ORT GDN registration, but doing it inside a per-layer emitter means a 27B export can load weights and partially build the graph before failing. Please move this validation into `Qwen35TextModel.__init__` / `check_extra_options` , before weight loading. The same applies to the dense helper above.

src/python/py/models/builders/base.py
```diff
@@ -3731,6 +3772,130 @@ def make_linear_attention(self, name, **kwargs):
...
+        outputs = [output, present_recurrent]
+        state_update_capsule = kwargs.get("state_update_capsule", "")
+        if state_update_capacity:
+            if not state_update_capsule:
+                raise ValueError("state_update_capsule is required when state_update_capacity is positive")
+            outputs.append(state_update_capsule)
+
+        attributes = {
+            "update_rule": kwargs.get("update_rule", "gated_delta"),
+            "scale": kwargs.get("scale", 1.0),
```
baijumeswani: > ORT defines `scale=0.0` as "use  1/sqrt(head_size_qk)", but this reusable helper defaults omitted `scale` to literal `1.0`. Current Qwen callers explicitly pass `0.0`, so this is not a live failure today, but a future caller can silently lose output scaling. Please default to `0.0` here and in `make_gated_delta_net()`

src/python/py/models/builders/qwen.py
tianleiwu: Fixed in a513fe6c. The full-attention Q/gate path now derives its token shape and reshape constants from use_paged_attention, producing rank-2 [num_tokens, ...] tensors for paged exports while preserving the rank-3 dense layout. The projection tests cover both layouts.

src/python/py/models/builders/qwen.py
tianleiwu: Fixed in a513fe6c. genai_config.json now records state_update_capacity, both compact-update input bindings, and both per-layer output-name templates. The C++ config schema parses those fields, with Python serialization and C++ parser coverage.

src/python/py/models/builders/qwen.py
tianleiwu: Fixed in a513fe6c. Qwen3.5/3.8 validation now rejects every paged GatedDeltaNet export unless the execution provider is CUDA, including WebGPU. A focused validation case covers WebGPU FP16.

src/python/py/models/builders/qwen.py
tianleiwu: Fixed in a513fe6c. Paged exports now select GatedDeltaNet without requiring linear_attn_op, and the documented command no longer supplies that no-op option. Nonzero state_window is rejected for GatedDeltaNet instead of being tied to state_update_capacity.

src/python/py/models/builders/qwen.py
tianleiwu: Fixed in a513fe6c. GatedDeltaNet exports now reject any nonzero state_window, including the dense path, so convolution and recurrent state cannot be emitted with incompatible logical boundaries. The dense validation case is covered by a focused test.

src/python/py/models/builders/base.py
tianleiwu: Fixed in a513fe6c. BF16 validation now runs from Qwen35TextModel initialization through configure_gated_delta_net_io(), before make_model() calls load_weights(). Focused cases cover both paged and dense GatedDeltaNet exports.

src/python/py/models/builders/base.py
tianleiwu: Fixed in a513fe6c. Dense and variable-length GatedDeltaNet now share make_gated_delta_net_attributes(), whose omitted scale defaults to 0.0. The parity test locks that default for both emitters.

src/python/py/models/builders/base.py
```diff
@@ -1142,6 +1152,10 @@ def make_genai_config(self, config, extra_kwargs, out_dir):
             outputs["present_conv_names"] = "present.%d.conv"
         if "present.recurrent" in self.output_names:
             outputs["present_recurrent_names"] = "present.%d.recurrent"
+        if "state_update.conv_value" in self.output_names:
+            outputs["state_update_conv_value_names"] = "state_update.%d.conv_value"
```
kunal-vaishnavi, line 1185: Aren't the state updates already captured in `past/present.conv/recurrent`? Why are these new ones needed?

src/python/py/models/builder.py
```diff
@@ -733,8 +748,15 @@ def get_args():
                     In addition to `logits`, you will have `hidden_states` as an output to your ONNX model.
                 mtp_quant_config = JSON object/file: Configure MTP I/O, dense weights, MoE, and runtime using the
                     structured QuantConfig schema independently from the main model.
-                state_window = Widen Qwen3.6 recurrent/conv state I/O to [W, B, ...]. Default is 0 (disabled).
+                linear_attn_op = linear_attention/gated_delta_net: Select the recurrent operator for non-paged
+                    Qwen3.5/3.8 exports. Default is linear_attention. Paged exports always use GatedDeltaNet and
+                    ignore this option. gated_delta_net requires state_window=0 and does not support bfloat16 I/O.
```
kunal-vaishnavi: Why can't we support bf16?

src/config.h
```diff
@@ -423,6 +426,8 @@ struct Config {
         std::string attention_metadata{Defaults::AttentionMetadataName};
         std::string past_conv_names{Defaults::PastConvName};  // Conv cache input name template (LFM2)
         std::string past_recurrent_names{Defaults::PastRecurrentName};
+        std::string state_update_capture_count;  // Per-sequence count of transitions to capture this forward
```
kunal-vaishnavi: Can we add default names with the `Defaults::` syntax for the new input/output strings?

src/python/py/models/builder.py
tianleiwu: We can support it later.

src/python/py/models/builders/base.py
tianleiwu: See https://github.com/microsoft/onnxruntime/blob/382910c3eaf84e47f10a4c581537c1bfd7a222bd/onnxruntime/core/graph/contrib_ops/bert_defs.cc#L2771

It is a new output for the operator.

src/config.h
tianleiwu: Added

baijumeswani (dismissed):

baijumeswani (approved):

PR #2454, opened 2026-08-23
Replay compact fixed state for hybrid decoding

tianleiwu:
## Summary

Replace dense recurrent checkpoint tensors with compact state-update replay for dynamically batched hybrid models. The engine now owns fixed convolution and GatedDeltaNet state beside paged KV cache, captures packed per-token updates, and replays only an accepted prefix into inactive state banks before atomically publishing both cache types.

## Stack

- Stacked on #2453, which exports compact Qwen3.5/3.8 `value` and `capsule` outputs.
- Depends on microsoft/onnxruntime#32282 and microsoft/onnxruntime#32290.
- MTP scheduling and drafter orchestration remain outside this PR.

## Key Changes

- Parse explicit `paged_kv` and `fixed` decoder state groups, including compact convolution `value` and GDN `capsule` update bindings.
- Add a fixed-state pool with per-request ownership, double-buffered publication, compact prefix replay, and CPU/CUDA provider implementations.
- Add composite paged/fixed reservations with validation and preparation before the allocation-free publication boundary.
- Bind packed variable-length inputs and fixed state together through `HybridDecoderIO`, including optional packed `position_ids`.
- Resolve paged cache tensor names from explicit manifest bindings while retaining legacy name-template synthesis for models without `state_groups`.
- Add synthetic hybrid/composite fixtures and focused manifest, reservation, replay, transaction, and end-to-end Engine coverage.

## Memory Impact

The runtime no longer requires dense per-token recurrent checkpoint outputs. Speculative rollback data is bounded by `state_update_capacity` and represented as compact convolution values or packed GDN transition capsules.

## Testing

- `test/python/builder/test_decoder_state_groups.py`: 24 passed
- CPU Debug `engine_unit_tests`: 256 passed
- CUDA Debug `engine_unit_tests` on H200: 261 passed
- Paged and hybrid Engine Python tests across CPU/CUDA: 32 passed
- CUDA compact convolution/GDN partial-prefix replay: passed

## Checklist

- [x] Tests added and updated
- [x] CPU and CUDA replay paths validated
- [x] No MTP/drafter scheduling included
- [x] Existing paged-only Engine behavior preserved


tianleiwu:
Stack: **1/6** → #2455 → #2456 → #2457 → #2458 → #2459

src/config.h
```diff
@@ -391,13 +391,40 @@ struct Config {
...
         PagedKeyValue,
-        FixedConv,
-        FixedRecurrent,
+        Fixed,
+      };
+
+      struct StateBinding {
+        std::string input;
+        std::string output;
+      };
```
baijumeswani, line 400: This was the original design in https://github.com/microsoft/onnxruntime-genai/pull/2427, but was later updated based on review comments and superseded by https://github.com/microsoft/onnxruntime-genai/pull/2463.

Does this need to be updated to reflect that change?

src/engine/fixed_state_pool.cpp
```diff
@@ -751,25 +957,68 @@ FixedStateReservation FixedStatePool::Reserve(
...
+    const auto allocate_update_output = [&](const Impl::StateUpdateOutputSpec& output) {
+      if (!capture_state_updates || output.name.empty()) {
+        return std::unique_ptr<OrtValue>{};
+      }
+      return OrtValue::CreateTensor(
+          impl_->device->GetAllocator(), StorageShape(batch_rows, output.session_shape),
+          output.data_type);
+    };
+    state_updates.value = allocate_update_output(spec.state_update_value);
+    state_updates.capsule = allocate_update_output(spec.state_update_capsule);
```
baijumeswani, line 970: `FixedStatePool::Reserve()` creates every state-update value/capsule tensor through the device allocator. This contradicts the pool’s invariant that device staging is allocated before paged-KV auto-sizing.

It is dormant in this pull-request because production requests pass `capture_count == 0 ` but #2491 activates it. At that point a scheduled speculative step can fail with a device OOM after planning, and it adds allocator churn to the hot path.

Recommended resolution: preallocate capacity-sized value/capsule backing per `TensorSpec` during pool construction and expose exact-row views, matching gathered/output staging.

baijumeswani (approved):

src/engine/fixed_state_pool.cpp
tianleiwu: will address it later.

src/config.h
tianleiwu: Will apply 2463 later

PR #2459, opened 2026-08-23
Run chained MTP drafting in the batch Engine

tianleiwu:
## Summary
- run the auxiliary MTP decoder transactionally beside the target decoder
- support per-request draft lengths and compact active requests between head stages
- keep chained draft IDs on device and perform one final host materialization
- expose cumulative Engine speculative telemetry through C, C++, and Python
- defer drafts for a selected transaction that mixes prefill and speculative rows, then resume MTP on decode-only transactions

## Stack
PR 6/6. Depends on `tlwu/20260823/gdn-paged-mtp-config`.

## Correctness fix
Packed recurrent operators choose one execution plan for the whole ragged batch. A long prefill can select chunked GatedDeltaNet, which publishes final state but not the intermediate recurrent checkpoints required to reject drafts. The Engine previously treated those checkpoint outputs as valid, so a rejected request could publish an uninitialized checkpoint and diverge on its next target call.

After paged selection, the composite cache planner now clears draft counts only when the selected transaction contains both prefill and draft rows. Decode requests still advance by one token, prefills continue, and K=3 resumes when the selected transaction is decode-only.

## Validation
- `engine_unit_tests`: 323/323 passed
- focused `CompositeDefersDraftsWhilePrefillSharesTheStep` regression passed
- clang-format dry run and `git diff --check` passed
- post-fix layer tracing is exact for 2,048/2,048 rows in transaction 1 and 1,803/1,803 rows in transaction 2
- batch-1 token parity: 438/438; 49.02 to 91.93 tok/s (1.88x)
- sustained concurrency-16 token parity: 2,004/2,004; 288.55 to 316.00 tok/s (+9.5%)
- seven-shard MMLU-Pro: true plain 664/800 (83.00%), post-fix K=3 666/800 (83.25%), zero unextracted
- seven-shard GPQA: true plain 147/198 (74.24%), post-fix K=3 151/198 (76.26%), zero unextracted
- weighted accepted/evaluated draft rate: 84.98% MMLU-Pro and 84.73% GPQA

## Release disposition
K=3 MTP is quality-safe for continuous batching with this safeguard. The prior batch-1-only disposition and numerical-shape explanation are superseded by the recurrent-checkpoint rollback root cause above. Keep the stack in draft pending CI.

PR #2460, opened 2026-08-24
Run a DFlash2 block drafter in the batch Engine

tianleiwu:
### Description

Adds end-to-end support for **DFlash 2**, a *block* drafter, to the model builder and the batch `Engine`.

Unlike an MTP head, which produces one token per forward pass and has to be chained, a DFlash 2 drafter consumes the target's *auxiliary hidden states* and proposes an entire `block_size`-token block in a single pass, then walks a lattice to pick the best path.

Stacked on #2459 (`tlwu/20260823/gdn-paged-mtp-engine`). Review only the top four commits.

| Commit | Area |
|---|---|
| `feat(builder): tap auxiliary hidden states for speculative drafters` | `base.py` |
| `feat(builder): export the DFlash 2 block drafter for Qwen3.8` | `dflash2.py`, `qwen.py` |
| `feat(engine): surface auxiliary hidden states from the packed decoder` | decoder IO chain |
| `feat(engine): run a DFlash 2 block drafter in the batch Engine` | `config.*`, `dflash2_drafter.*`, `engine.*` |

### Design notes

**Auxiliary hidden states.** Entry `i` is the residual stream *entering* decoder layer `i`. That value only exists as a tensor inside layer `i`'s input `SkipLayerNorm`, so the tap reads that node's `output_3`. Taking the layer boundary instead is an off-by-one that produces a drafter whose measured acceptance rate saturates at exactly `1.000` — it looks like a perfect drafter and is in fact reading the wrong tensor.

**The drafter graph.** `dflash2.onnx` is a standalone 5-layer model with no embedding and no LM head; both are shared with the target. Attention is a packed, **non-causal** `PagedAttention` whose `qkv_row_map` gathers K/V from `concat(query, context)`, so the query block attends to itself bidirectionally while the cached context stays strictly to the left. This depends on the `is_causal` attribute added in onnxruntime#32225.

**dtype.** The body runs in bfloat16 — activations reach `1.5e5` and overflow fp16 — while the external interface stays fp16 to match the target.

**Memory.** The drafter is constructed *before* the main cache manager so the paged pool's free-memory measurement already excludes it. Its own pool is a sliding-window ring sized from `max_batch_size` rather than the context length; ingesting unwindowed context costs ~10.6 GB at this geometry.

**Commit ordering.** Feeds are captured *before* `Request::CommitStep` (which clears the accepted-draft counts the row trimming depends on) and published after. `first_position` comes from `ProcessedSequenceLength()`; deriving it from `sequence_length_before` is wrong for a prefill chunk.

### Validation

Target `Qwen3.8-27B-NVFP4` + `Qwen3.8-27B-DFlash2` on H200 (SM90, CUDA 13.0).

**Correctness.** The exported ONNX drafter matches a PyTorch reference at 4.135 vs 4.128 accepted tokens/step, with 135 of 141 lattice walks identical. Losslessness against plain greedy: **2042 tokens, 0 mismatches**, `spec_tokens_per_step = 4.263` (2043 base steps → 479 speculative steps).

**Quality** — no regression, 7 shards, concurrency 8:

| Benchmark | With DFlash 2 | Control (no drafter) |
|---|---:|---:|
| MMLU-Pro (n=800) | 82.75% | 82.63% |
| GPQA-diamond (n=198) | 75.25% | 75.25% |

Zero unextracted and zero degenerate answers on every arm.

**Throughput** — speedup over the same model with speculation disabled:

| Context | bs 1 | bs 4 | bs 16 |
|---|---:|---:|---:|
| 512 | 5.00x | 3.26x | 2.68x |
| 2048 | 5.12x | 2.07x | 2.10x |
| 8192 | 4.86x | 3.82x | 3.68x |

### Notes for reviewers

- A short-prompt losslessness test is **not** a sufficient gate for this path. The class of bug fixed by `c2ee5a37` (already in #2459) only appears in *mixed prefill + decode* batches, which only a slot-refilling harness produces.
- `c2ee5a37` currently defers drafts for the whole batch whenever any request is prefilling. A per-request carve-out would recover that throughput, but it requires the packed GDN operator to accept a per-row execution plan; that is tracked separately and is not part of this PR.


tianleiwu:
Superseded by #2496, which rebuilds the DFlash2 runtime on the compact #2495 stack without dense recurrent checkpoints, duplicated builder changes, or an ONNX Runtime dependency bump.

PR #2448, opened 2026-08-22
Harden Engine request lifetime and sampler acquisition

bmehta001:
## Summary

- harden Engine request lifetime handling so external C-handle release, reacquisition, and owner-thread reclamation cannot prematurely close a live request
- preserve failure-atomic first-turn and continuation admission across the redesigned Engine turn API, including scheduler rollback and poisoned-request handling
- route ownership, turn admission, cancellation, failure, and close through narrow validated Request transitions while retaining targeted friendships where private collaboration is required
- keep final-handle notification inside `ExternalRefCounted` as an optional release signal while synchronized ownership remains the source of truth
- keep CUDA batched-sampler index growth and wrapper allocation failure-atomic in `src/ep/cuda/interface.cpp`
- strengthen paged, full, and windowed-cache invariant diagnostics and rollback/retry coverage

## Scope

This PR changes internal experimental Engine implementation and tests; it does not add or change the public C Engine API.

The final `main...HEAD` diff contains no DML replay implementation. The CUDA sampler helper remains implementation-local and does not add a standalone pool API or test target.

## Validation

- LLVM 20 format dry-run over the complete changed C/C++ set: passed
- whitespace and broad Engine ownership-friendship checks: passed
- Linux/WSL Debug `engine_unit_tests` build: passed
- Linux/WSL Engine tests: **406/406 passed**
- H100/SM90 Engine smoke benchmark: passed in ADO build **1391727** on merged baseline commit `b066c3aa`
- latest commit `ef860a40` keeps Engine lifecycle methods private without class-wide `Request` friendship; CI retriggered and CUDA EOS tests await hardware execution

A full local Windows MSBuild remains blocked before project compilation because the installed Visual Studio environment lacks Spectre-mitigated libraries (`MSB8040`). Security settings were not weakened.

PR #2464, opened 2026-08-24
Redesign experimental Engine turn API

bmehta001:
## Summary

- redesign the experimental Engine API around `Engine -> Request -> Turn`, with Requests created and owned by an Engine
- replace polling and unseen-token state with typed caller-buffered events from one canonical bulk `OgaEngineRun`
- add request-local Turn IDs, named cancellation, per-Turn generation limits, usage and finish metadata, and automatic reclamation when the final Request handle is released
- update the C ABI, C++ RAII wrapper, Python binding, examples, benchmarks, tests, and Engine architecture documentation

Addresses #2448.

## API contract

```c
OgaResult* OgaEngineRun(
    OgaEngine* engine,
    void* event_buffer,
    size_t event_capacity,
    size_t event_stride,
    size_t* out_event_count);
```

- each positive-capacity call drains retained events or executes at most one model transaction, never both
- capacity one preserves one-event-at-a-time pacing; larger buffers can receive one event per affected Request from the committed transaction
- events identify a borrowed `OgaRequest*` and Request-local `turn_id`; flags are a bitmask, so the final token may also carry `TurnFinished`
- `OgaRequestOptions.max_session_tokens == 0` uses the Request's snapshotted `OgaGeneratorParams.search.max_length`
- opaque reusable `OgaTurnParams` provide per-Turn overrides; `max_generated_tokens` is implemented and unsupported setters return explicit errors
- stop sequences use token-ID sequences through `OgaSequences`; generation-time stop handling remains deferred
- `OgaRequestClose` releases Engine state while `OgaDestroyRequest` releases the caller-owned handle
- releasing the final public Request handle marks it abandoned; the Engine reclaims scheduler membership, retained events, and dynamic cache state at the next owner-thread boundary
- Engine, Request, and Turn operations use one host owner thread; model execution may still batch Requests in parallel

## Runtime behavior

- dynamic batching supports resident continuation and prompt chunking
- static batching preserves logical close immediately, but a closed resident row remains physically allocated until the shared batch is recycled
- capacity, retryable, request-terminal, contract, and fatal failures are represented by typed events rather than diagnostic-string parsing
- Engine-level events have no Request; examples and benchmarks classify them before request lookup and use bounded retry handling where appropriate
- `OgaCreateEngine` retains the Model, so the caller may release its Model handle after successful Engine creation

## Breaking changes

This intentionally replaces the previous experimental Engine API. The old create/add/continue/step/remove and unseen-token entry points, Request opaque-data accessors, and compatibility aliases are removed. Generator APIs are unchanged.

## Validation

- Linux/WSL Debug native, Python, and Engine benchmark builds
- 77 focused Engine lifecycle/runtime tests
- focused Engine C API tests
- 29 Python Engine tests
- Python examples and benchmark syntax checks
- C++17 public-wrapper compile smoke
- full material review of runtime, ABI, wrapper, example, benchmark, and documentation changes

Windows-native build validation was unavailable because this environment has no Visual Studio 2022 installation.

## Follow-up

- [ ] Consider adding Engine event-buffer generation tracking (for example, `OgaEngineEventBufferGetGeneration`) so C++ and managed wrappers can reject event/usage views retained across a later validated `OgaEngineRun`, instead of relying only on the documented borrowed-view lifetime.


src/ort_genai_c.h
```diff
@@ -82,6 +82,10 @@ typedef struct OgaEngine OgaEngine;
 typedef struct OgaRequest OgaRequest;
 typedef struct OgaStreamingProcessor OgaStreamingProcessor;
 
+typedef struct OgaTurnOptions {
+  size_t max_generated_tokens;
+} OgaTurnOptions;
```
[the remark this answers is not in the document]
bmehta001: I think this is ok to omit for now and have intentionally skipped it.

benchmark/engine/scenarios/decode_baseline.cpp
```diff
@@ -107,33 +107,31 @@ ScenarioExecutionOutput DecodeBaselineScenario::Execute(const ScenarioConfig& co
...
-          reinterpret_cast<std::uintptr_t>(request_tokens.data() + request_tokens.size());
-
-      if (ptr_addr < base_addr || ptr_addr >= end_addr) {
-        throw std::runtime_error(log_tag + ": opaque data pointer not in request_tokens");
+      const auto now = std::chrono::steady_clock::now();
+      const auto owned_request = std::find_if(
+          requests.begin(), requests.end(),
+          [ready_request](const auto& request) {
+            return request.get() == ready_request;
+          });
```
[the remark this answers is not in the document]
bmehta001: Fixed

src/ort_genai.h
```diff
@@ -972,29 +975,18 @@ struct OgaEngine : OgaAbstract {
...
-   * Repeated calls after the request has already been removed are successful no-ops.
-   */
-  void Remove(OgaRequest& request) {
-    OgaCheckResult(OgaEngineRemoveRequest(this, &request));
+  std::unique_ptr<OgaRequest> CreateRequest(
+      const OgaGeneratorParams& params) {
+    OgaRequest* request;
+    OgaCheckResult(OgaEngineCreateRequest(this, &params, &request));
+    return std::unique_ptr<OgaRequest>(request);
   }
```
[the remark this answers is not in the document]
bmehta001: Fixed

src/ort_genai_c.h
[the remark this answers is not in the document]
bmehta001: Ok, will add

src/ort_genai_c.h
```diff
@@ -80,8 +80,79 @@ typedef struct OgaStringArray OgaStringArray;
...
+} OgaErrorCode;
+
+/**
+ * \brief Public options copied by OgaEngineCreateRequest.
+ *
+ * max_session_tokens == 0 uses the Request's snapshotted OgaGeneratorParams search.max_length.
+ * That search limit normally defaults from the model context length, but the caller may set it
+ * lower before Request creation.
+ */
+typedef struct OgaRequestOptions {
```
baijumeswani: Let's use an opaque type for requestoptions (similar to TurnParams). Also, let's keep the naming consistent (RequestOptions and TurnOptions).

src/ort_genai_c.h
```diff
@@ -80,8 +80,79 @@ typedef struct OgaStringArray OgaStringArray;
...
+
+/**
+ * \brief One caller-buffered Engine event.
+ *
+ * Pump OgaEngineRun while work remains and inspect flags as a bitmask: Token and TurnFinished may be
+ * combined. request is a borrowed identity alias for the caller-owned OgaRequest, and turn_id
+ * identifies the Request-local Turn. Consume token only when Token is set. The borrowed request
+ * remains valid only while the caller retains the owned OgaRequest handle.
+ */
+typedef struct OgaEngineEvent {
```
baijumeswani: Let's also make the OgaEngineEvent an opaque type and introduce a concept of an OgaEngineEventBuffer. The OgaEngineEventBuffer owns all the OgaEngineEvents and offers only a view of the events.

Let's also make the OgaTurnUsage also opaque. With this change, we shouldn't need any non-opaque structs.

```cpp
size_t capacity = 4;
OgaEngineEventBuffer* buffer = nullptr;
OgaCreateEngineEventBuffer(engine, capacity, &buffer);
OgaEngineRun(engine, buffer);
 
size_t count = OgaEngineEventBufferGetCount(buffer);
const OgaEngineEvent* event = OgaEngineEventBufferGet(buffer, index);
 
OgaEngineEventGetFlags(event, &flags);
OgaEngineEventGetRequest(event, &borrowed_request);
OgaEngineEventGetTurnId(event, &turn_id);
OgaEngineEventGetToken(event, &token);
OgaEngineEventGetFinishReason(event, &reason);
OgaEngineEventGetUsage(event, &usage);

OgaDestroyEngineEventBuffer(buffer);
```

The buffer allocates once and is reused. Event pointers are borrowed from it and valid until the next  Run , reset, or buffer destruction. Overflow remains retained by Engine and is drained on later calls, as in the current design.

This gives several advantages:

 - no public event layout, version, stride, alignment, or enum-width contract;
 - stop metadata, capacity details, and future speculative fields can be added through getters;
 - no caller-controlled per-run allocation;
 - C++, Python, and C# wrappers can expose natural value/view objects;
 - the C++ wrapper can return a non-owning Request reference instead of a deletable raw handle.

src/ort_genai_c.h
```diff
@@ -80,8 +80,79 @@ typedef struct OgaStringArray OgaStringArray;
...
+/** \brief Stable behavioral classification for an Engine event. */
+typedef enum OgaErrorCode {
+  OgaErrorCode_None = 0,
+  OgaErrorCode_CapacityDeferred = 1,
+  OgaErrorCode_ExecutionCapacityExceeded = 2,
+  OgaErrorCode_RetryableExecution = 3,
+  OgaErrorCode_RequestUnserviceable = 4,
+  OgaErrorCode_EngineContractFailure = 5,
+  OgaErrorCode_EngineExecutionFailure = 6,
+} OgaErrorCode;
```
baijumeswani: Plain C enums do not guarantee fixed-width adnd can break if new fields are added. Use typedefs to work around this.

```cpp
typedef uint32_t OgaFinishReason;
 
 #define OgaFinishReason_None               ((OgaFinishReason)0u)
 #define OgaFinishReason_Eos                ((OgaFinishReason)1u)
 #define OgaFinishReason_StopString         ((OgaFinishReason)2u)
 #define OgaFinishReason_MaxGeneratedTokens ((OgaFinishReason)3u)
 #define OgaFinishReason_MaxSessionTokens   ((OgaFinishReason)4u)
 #define OgaFinishReason_Cancelled           ((OgaFinishReason)5u)
 #define OgaFinishReason_Failed              ((OgaFinishReason)6u)
```

Then getters can remain ABI-stable:

```cpp
OgaResult* OgaEngineEventGetFinishReason(
    const OgaEngineEvent* event,
    OgaFinishReason* out_reason);
```

Same applied to the other enums in this file.

baijumeswani: Make sure that the new c api honors the C ABI (so it allows for future expandability and does not break for applications).

src/engine/engine.cpp
```diff
@@ -43,8 +89,26 @@ Engine::Engine(std::shared_ptr<Model> model, EngineDependencies dependencies)
...
+    auto request = tracked_requests_.back().lock();
+    tracked_requests_.pop_back();
+    if (!request) {
+      continue;
+    }
+    if (IsClosed(request->status_)) {
+      continue;
+    }
+    try {
+      CloseRequest(request);
```
baijumeswani: > This cannot execute the normal close path during `~Engine`: once destruction begins, `request->engine_.lock()` is empty, so `CloseRequest()` fails its ownership check and every live `Request` falls into the catch-all. `CompleteClose()` changes status but skips scheduler removal, event purge, and cache deallocation. Please add a teardown-specific, no-throw detach path that releases scheduler/cache ownership and Request runtime state without rechecking the expired weak `Engine` pointer. The test should retain the cache-manager double and verify actual resource release, not only `RequestStatus::Closed`.

src/engine/engine.cpp
```diff
@@ -54,45 +118,265 @@ EngineDependencies Engine::CreateDependencies(std::shared_ptr<Model> model) {
...
+  }
+
+  auto request = std::make_shared<Request>(
+      CloneRequestParams(params, *model_), max_total_tokens);
+  request->ValidateEngineCompatibility();
 
-  // Track the request before assignment so every successfully submitted request can later be found
-  // even when the scheduler and cache hold it through implementation-specific containers. The
-  // registry allocation therefore happens before any request lifecycle mutation.
   tracked_requests_.push_back(request);
```
baijumeswani, line 126: > `tracked_requests_` stores only a weak reference, so before the first `BeginTurn` the external handle can be the Request's only strong owner. Releasing that handle on an HTTP/GC thread then destroys Search, guidance, parameters, and any device state on that thread, even though the documented contract says final release only marks abandonment for owner-thread reclamation.
> Please make the Engine strongly retain active Requests until `Close`/reclamation, and release all ORT/device-affine state on the owner thread before dropping that strong reference. A surviving public handle can then retain only a lightweight closed tombstone.

src/engine/engine.cpp
```diff
@@ -184,67 +479,96 @@ void Engine::ValidateRequestCanContinue(const std::shared_ptr<Request>& request)
...
+    RetainEvent(EventFromStepError(error));
+  }
+  return DrainPendingEvents(events);
 }
 
-std::shared_ptr<Request> Engine::StepStatic() {
-  while (scheduler_->HasPendingRequests()) {
+void Engine::RunStatic() {
+  if (scheduler_->HasPendingRequests()) {
     auto scheduled_requests = scheduler_->Schedule();
```
baijumeswani: Schedule() is outside the try-catch. Which means that the engine could be healthy but we return an OgaResult. Can the Schedule be inside the try-catch?

src/engine/engine.cpp
```diff
@@ -54,45 +118,265 @@ EngineDependencies Engine::CreateDependencies(std::shared_ptr<Model> model) {
...
+      [&request, turn_id](const EngineEvent& event) {
+        return event.request == request && event.turn_id == turn_id;
+      });
+  const bool has_existing_event = existing != pending_events_.end();
+  if (!has_existing_event) {
+    pending_events_.reserve(pending_events_.size() + 1);
+  }
+
+  request->status_ = RequestStatus::TurnComplete;
+  request->finish_reason_ = GenerationFinishReason::Canceled;
```
baijumeswani, line 306: > Cancellation completes the Turn here without resetting its guidance cursor. If a guided Turn has already emitted a partial grammar, the next `BeginTurn` inherits that state because guidance is otherwise reset only on normal generation completion. Please establish a fresh guidance cursor transactionally during `BeginTurn` and restore the old cursor if admission fails. Be careful to use the Request-level save/commit/restore consistently; combining `Request::SaveStateForTransaction()` with the current direct `search_->CommitStateForTransaction()` would leave a stale guidance checkpoint that a later rollback could reinstall.

src/ort_genai_c.h
bmehta001: Thanks - fixed

src/ort_genai_c.h
bmehta001: This makes a lot of sense - added

src/ort_genai_c.h
bmehta001: Fixed

src/ort_genai_c.h
```diff
@@ -1185,189 +1227,211 @@ OGA_EXPORT OgaResult* OGA_API_CALL OgaCreateEngine(OgaModel* model, OgaEngine**
...
- *
- * \param[in] request The request to set the input sequences on.
- * \param[in] tokens The input sequences to set on the request.
- * \return OgaResult containing the error message if the setting of the input sequences failed, or nullptr on success.
+ * This operation is currently not implemented. stop_sequences is not dereferenced or retained
+ * before the not-implemented result is returned.
  */
-OGA_EXPORT OgaResult* OGA_API_CALL OgaRequestSetOpaqueData(OgaRequest* request, void* opaque_data);
+OGA_EXPORT OgaResult* OGA_API_CALL OgaTurnOptionsSetStopSequences(
+    OgaTurnOptions* options, const OgaSequences* stop_sequences);
```
baijumeswani: We should have 1 for StopStrings and StopTokenIds. Can we rename this function as OgaTurnOptionsSetStopTokenIds. And add another OgaTurnOptionsSetStopStrings?

examples/python/engine/continuous-batching.py
```diff
@@ -4,40 +4,38 @@
...
+        return event.request
+    if event.flags & og.EngineEventFlags.FAILED:
+        outcome = "failed"
+    elif event.flags & og.EngineEventFlags.CAPACITY_BLOCKED:
+        outcome = "was capacity-blocked"
+    elif event.flags & og.EngineEventFlags.RETRYABLE:
+        outcome = "reported a retryable failure"
+    else:
+        outcome = "returned an invalid request-less event"
+    raise RuntimeError(f"Engine {outcome}; error_code={event.error_code}")
```
baijumeswani, line 22: capacity blocked and retriable are not terminal errors. The engine still remains healthy and can schedule requests in such cases.

src/engine/request.h
```diff
@@ -320,15 +256,21 @@ struct Request : std::enable_shared_from_this<Request>,
...
   // Sequence length the application's tokens reach up to. Everything below it is prompt, so the
   // request is still prefilling while processed_sequence_length_ has not caught up with it.
   int64_t prompt_sequence_length_{};
   size_t scheduled_token_count_{};
+  std::optional<size_t> turn_max_generated_tokens_;
+  size_t turn_prompt_tokens_{};
+  size_t turn_generated_tokens_{};
+  const size_t max_total_tokens_;
+  uint64_t current_turn_id_{};
+  uint64_t next_turn_id_{};
```
baijumeswani: Can we start Turn IDs at 1 and reserve 0 for "no Turn"? Output parameters and Engine-level events already use 0 as the empty value, so this removes ambiguity for callers.

src/engine/engine.cpp
```diff
@@ -470,15 +912,24 @@ std::shared_ptr<Request> Engine::DrainReadyRequest() {
       outcome == StepOutcomeKind::ExecutionContractFailure) {
     ++transaction_metrics_.fatal_execution_failures;
   }
+  for (const auto& tracked : tracked_requests_) {
```
baijumeswani: This marks every active Turn as failed, but only emits one request-less Engine event. Can we either emit a terminal event for each affected Turn or document how callers should complete those pending operations?

src/python/python.cpp
```diff
@@ -730,43 +760,177 @@ PYBIND11_MODULE(onnxruntime_genai, m) {
...
+      .def(
+          "create_event_buffer",
+          &OgaEngine::CreateEventBuffer,
+          pybind11::arg("capacity"))
+      .def(
+          "run",
+          [](OgaEngine& engine, pybind11::object buffer_object) {
+            auto& buffer =
+                buffer_object.cast<OgaEngineEventBuffer&>();
+            engine.Run(buffer);
```
baijumeswani: Run can execute a full model step while holding the Python GIL. Can we release the GIL around this call and reacquire it before returning? This lets other Python threads continue stream or tool work.

src/engine/engine.cpp
```diff
@@ -105,9 +389,10 @@ void Engine::RemoveRequest(std::shared_ptr<Request> request) {
...
 
 void Engine::ReclaimAbandonedRequests() {
-  // ExternalRelease only publishes an atomic abandonment marker. Engine entry points are externally
-  // serialized, so this boundary can safely perform the normal removal sequence: scheduler/cache
-  // release, ready-notification purge, and terminal close.
+  // ExternalRelease only publishes an atomic abandonment marker. The host's owner-thread boundary
+  // can safely perform the normal removal sequence: logical scheduler removal, ready-notification
+  // purge, and terminal close. Dynamic cache ownership is released immediately; a resident static
+  // batch row can remain physically retained until its shared batch is recycled.
   std::vector<std::shared_ptr<Request>> abandoned_requests;
```
baijumeswani: This allocates a temporary vector on every Run, pending check, and Request creation, even when nothing was abandoned. Can we gate this on an 'abandonment pending' flag or reuse Engine-owned scratch storage?

src/engine/engine.cpp
```diff
@@ -438,25 +776,129 @@ std::shared_ptr<Request> Engine::StepDynamic() {
...
+      break;
+    case StepOutcomeKind::ExecutionContractFailure:
+      event.flags = EngineEventFlagFailed;
+      event.error_code = EngineErrorCode::EngineContractFailure;
+      break;
+    case StepOutcomeKind::FatalExecutionFailure:
+      event.flags = EngineEventFlagFailed;
+      event.error_code = EngineErrorCode::EngineExecutionFailure;
+      break;
+    case StepOutcomeKind::UnserviceableRequest:
```
baijumeswani, line 1004: Can this mapping handle every outcome without throwing from inside the catch path? FailUnserviceableRequest() can throw, and the bare throw below depends on caller context. Please return a typed event or mark the Engine failed explicitly for every case.

src/engine/scheduled_requests.cpp
```diff
@@ -192,7 +187,17 @@ bool ScheduledRequests::TryGenerateNextTokensBatched(std::vector<DeviceSpan<floa
   next_tokens.CopyDeviceToCpu();
 
   for (auto* request : sampling_plan_->requests) {
-    request->CompleteGeneration();
+    const auto result = request->CompleteGeneration();
+    if (results) {
+      const auto it = std::find_if(
```
baijumeswani: This searches the scheduled batch once for every sampled Request. Can the sampling plan keep each Request's result index so result placement stays linear and avoids identity lookups in the decode path?

src/ort_genai_c.h
bmehta001: Yes, can add - just wanted to make sure; will we have access to the tokenizer inside the engine, and will we be able to  stop for strings, or just for token ids?

src/engine/request.h
bmehta001: Fixed

src/engine/engine.cpp
bmehta001: Fixed

src/engine/engine.cpp
bmehta001: Created a separate try-catch to address this

src/engine/request.cpp
```diff
@@ -176,78 +120,42 @@ void Request::Remove() {
...
 
-  ValidateAppendLength(*params_, prefill_input_ids_.size(), tokens.size());
-  std::copy(tokens.begin(), tokens.end(), std::back_inserter(prefill_input_ids_));
+void Request::CompleteClose() noexcept {
+  engine_.reset();
+  status_ = RequestStatus::Closed;
+  guidance_transaction_checkpoint_.reset();
+  guidance_logits_processor_.reset();
+  batched_sampler_state_.reset();
+  search_.reset();
```
baijumeswani, line 169: for static batching, the request might still be alive as long as a parallelly batched request is alive. Resetting search and params will lead to a crash. Could we reject close on an in-flight static batched request?

src/engine/request.cpp
```diff
@@ -83,74 +89,15 @@ void Request::OnFirstExternalReference() noexcept {
 
 void Request::OnLastExternalReference() noexcept {
   externally_abandoned_.store(true, std::memory_order_release);
+  if (auto engine = engine_.lock()) {
+    engine->abandonment_pending_.store(true, std::memory_order_release);
+  }
```
baijumeswani: The last Request handle can be released from any thread. Locking the Engine here can race with `CompleteClose()` resetting `engine_`, and it can make this thread the Engine's final owner, causing Engine/runtime cleanup on the wrong thread.

This callback should only publish atomic abandonment state and leave cleanup to the Engine owner thread.

src/ort_genai_c.h
baijumeswani: The engine has access to the Model and should be able to create the tokenizer through the model's interface. I will address this in a follow up PR.

src/engine/engine.cpp
bmehta001: Fixed

src/engine/engine.cpp
bmehta001: Fixed

examples/python/engine/continuous-batching.py
bmehta001: Fixed

src/engine/engine.cpp
bmehta001: Fixed - documented and now emitting a failure event for each request (unless no executable Turns are possible, in which case we have a request-less event)

src/engine/request.cpp
bmehta001: Yes - fixed

src/python/python.cpp
bmehta001: Fixed

src/engine/engine.cpp
bmehta001: Yes, fixed and used your `abandonment_pending_` flag idea

src/engine/scheduled_requests.cpp
bmehta001: Yes, now doing that

src/engine/engine.cpp
bmehta001: Yes, mapping every outcome now w/o throwing and returning an event with the relevant error code and flags

src/engine/request.cpp
bmehta001: Fixed

baijumeswani (approved):

PR #2466, opened 2026-08-25
Optimize Engine guidance for production workloads

baijumeswani:
## Summary

Makes constrained decoding practical for concurrent, long-running Engine workloads by amortizing grammar setup and moving per-token guidance work off the serialized Engine step path.

This change builds on Engine guidance and continuation introduced by microsoft/onnxruntime-genai#2412. It adds:

- model-local immutable tokenizer assets;
- a bounded, single-flight compiled grammar cache;
- request-local mutable grammar cursors;
- low-cost grammar reset for retained-request continuation;
- asynchronous cross-request mask computation using llguidance's parallel API;
- reusable host/device mask storage;
- one batched H2D transfer and CUDA mask launch for contiguous Engine decode rows;

## Motivation

The original guidance path was correct, but each request independently rebuilt tokenizer/trie state and compiled its grammar. It also computed, allocated, copied, and applied masks once per request on every decode step.

On the target Qwen3.8 27B workload, the production impact and measured improvement were:

| Problem | Before | After |
|---|---:|---:|
| Repeated schema request construction | ~2.6 seconds | ~0.5 ms p95 cached |
| Guided continuation | Recompiled/reset grammar state | <0.25 ms p95 |
| B4 guided decode regression | ~12.94% | 0.47% upper 95% bound |
| B8 guided decode regression | ~13.76% | 0.58% upper 95% bound |
| Unique grammar cache growth | Unbounded concern | 64 retained entries after 800 misses |

Coding workloads repeatedly reuse the same tool and response schemas across many requests and continuation turns, so both costs should be amortized at the model and scheduled-batch levels.

### Model-local guidance cache

- Creates the llguidance tokenizer/trie once per loaded model.
- Caches compiled initial constraints by guidance type, exact guidance data, and relevant options.
- Single-flights concurrent identical misses.
- Clones an independent mutable cursor for every request.
- Uses a bounded 64-entry LRU with a 16 MiB key-byte limit.
- Keeps evicted assets alive while active requests or asynchronous work still reference them.
- Removes failed compilations from the cache so retries perform a fresh compile.
- Releases all cached assets with the model.

### Low-cost continuation

`CloneForNewTurn()` now clones from the cached initial constraint instead of cloning tokenizer state and recompiling the grammar. Every continued turn still starts from grammar state zero.

### Overlapped mask computation

- Commits selected tokens on the Engine thread and marks their cursors dirty.
- Collects dirty cursors across the scheduled request batch.
- Calls `llg_par_compute_mask` once for the batch.
- Overlaps CPU grammar work with the next model forward.
- Waits for the result only when logits masking needs it.

### Batched CUDA mask application

For contiguous CUDA Engine decode rows:

- stores mask rows contiguously;
- reuses one scheduler-owned device workspace;
- uses one H2D transfer;
- applies all guided and pass-through rows with one CUDA kernel launch.

CPU, noncontiguous, and unsupported batching cases retain the per-request fallback. A guided request never silently becomes unconstrained when the fast path is unavailable.

### Correctness and ownership

- Uses copy-on-write grammar cursors for transactional and speculative clones.
- Keeps rollback checkpoints isolated from staged attempts.
- Pins grammar and tokenizer assets until asynchronous callbacks finish.
- Destroys constraints before their tokenizer.
- Publishes pending-mask futures only after the complete batch job is prepared.
- Contains all exceptions at the C callback boundary.
- Keeps synchronous and parallel stop-mask semantics identical.
- Retains the corrected padded mask-row stride for non-aligned vocabularies.
- Continues to reject Engine guidance fast-forward tokens, whose scheduler and KV accounting are not implemented.

### Qualification
  
The change was qualified with an external paired guided/unguided harness. No benchmark tooling or provider-loading behavior is added by this PR.

## Performance

Qwen3.8-27B INT4, A100 80 GB, CUDA 12.8, vocabulary 248,320, 4K-token prompt, 128 generated tokens:
  
  | Concurrency | Guided ITL mean change | Mean ITL upper 95% bound | Throughput mean change |
  |---:|---:|---:|---:|
  | B1 | +0.449% | +0.491% | -0.447% |
  | B4 | +0.389% | +0.471% | -0.311% |
  | B8 | +0.408% | +0.578% | -0.313% |


src/engine/scheduled_requests.cpp
```diff
@@ -203,6 +211,79 @@ bool ScheduledRequests::TryGenerateNextTokensBatched(std::vector<DeviceSpan<floa
...
+  float* const first_row = logits.front().Span().data();
+  for (size_t i = 0; i < logits.size(); ++i) {
+    if (logits[i].size() != vocab_size ||
+        logits[i].Span().data() != first_row + i * vocab_size) {
+      return false;
+    }
+  }
+
+  bool has_guidance = false;
+  sampling_plan_->guidance_masks.assign(logits.size() * words_per_row,
```
sayanshaw24: Could we add an end-to-end CUDA Engine test with guided and unguided requests scheduled in the same batch? This would verify row-to-mask mapping and ensure unguided rows remain untouched by scheduler-owned batched mask application.

src/engine/scheduled_requests.cpp
```diff
@@ -203,6 +211,79 @@ bool ScheduledRequests::TryGenerateNextTokensBatched(std::vector<DeviceSpan<floa
...
+                                        std::numeric_limits<uint32_t>::max());
+  for (size_t i = 0; i < requests_.size(); ++i) {
+    const auto& request = requests_[i];
+    if (!request->HasGuidance() || !request->IsChunkComplete()) {
```
sayanshaw24: similar to above comment - could we add a CUDA Engine test where one row is in partial prefill while another row performs guided decode? This is an important production batching combination and would verify that masks are applied only to eligible decode rows.

src/constrained_logits_processor.cpp
```diff
@@ -215,73 +507,178 @@ void GuidanceLogitsProcessor::CommitTokens(std::span<int32_t> tokens) {
...
+      steps.push_back(LlgConstraintStep{
+          job->constraints[output.first_constraint + i].get(),
+          job->masks.data() + output.first_mask_word + i * output.words_per_row,
+          output.words_per_row * sizeof(uint32_t),
+      });
+    }
+  }
+
+  for (size_t i = 0; i < candidates.size(); ++i) {
+    candidates[i]->pending_masks_ = std::move(futures[i]);
```
sayanshaw24, line 734: would it be possible to add a stress or synchronization test that removes or abandons a request while its asynchronous guidance-mask computation is pending? the shared ownership appears correct, but this lifetime contract should be exercised explicitly.

src/constrained_logits_processor.cpp
```diff
@@ -41,150 +387,87 @@ GuidanceLogitsProcessor::GuidanceLogitsProcessor(
...
     }
   }
+  return masks;
+}
+
+void GuidanceLogitsProcessor::EnsureMaskReady() {
+  if (!pending_masks_.valid()) {
+    return;
+  }
+  masks_ = pending_masks_.get();
```
sayanshaw24: maybe we can add fault-injection coverage for an asynchronous mask computation failure during a dynamic Engine transaction, including rollback and retry behavior? existing tests cover cursor rollback and reset failures, but not a mask future that completes exceptionally.

sayanshaw24 (dismissed):

tianleiwu (commented):
## Review: Optimize Engine guidance for production workloads

**Verdict: COMMENT.** The design is sound and unusually well-defended. I did not find a correctness blocker; everything below is hardening around cache accounting, error signalling, and the new public C API shape.

### Verified against the implementation

- **Sync/parallel stop-mask parity is real.** `ComputeMasks()` sets the EOS bit on `LlgMaskResult::is_stop`, and upstream `ffi_par.rs::par_compute_mask_inner` does the same (`add_eos = r.is_stop()` -> `*mask_dest.add(eos/32) |= 1 << (eos%32)`, using `constraint.tok_trie().eos_token()`, which is the `tok_eos` this code passes as `config.model.eos_token_id[0]`). The claim in the description holds.
- **Async lifetime is correct.** `ParallelMaskJob` holds `shared_ptr` copies of both the cursors and the grammar assets (which own the tokenizer), so a request or model destroyed mid-flight cannot free memory a rayon worker is writing. Dropping those references *before* satisfying the promises is a nice touch - the woken request sees unique ownership and skips an unnecessary COW clone.
- **Destruction ordering is correct in all three asset types** (`llg_tokenizer` last in `GuidanceTokenizerAsset`, `initial_constraint` after `tokenizer` in `GuidanceGrammarAsset`, `llg_constraints_` after `grammar_asset_` in the processor), satisfying llguidance's "must not free the tokenizer while constraints exist" rule.
- **The batched CUDA path never silently unconstrains.** `CollectBatchedGuidanceMasks` returns `FallbackRequired` for any guided row it cannot fill and `Ready` only when every eligible guided row was masked; `guidance_applied` is only consulted for rows gated on the same `IsChunkComplete()` predicate. `AddLogitsMask` assigns `float::lowest()`, so re-application after a transaction rollback is idempotent.
- **`mask_dirty_` and `pending_masks_.valid()` are mutually exclusive on every path**, including through `Clone()`, so the `logic_error` guard in `ScheduleGuidanceMaskComputation` is not reachable from speculative clones.
- **`CloneForNewTurn()` reads no mutable source state**, which is exactly what makes the clone-then-swap sequence in `Request::Continue()` retry-safe. Good that there is a test for the failing-reset case.

### Cross-cutting notes (files not in this diff)

**Implicit hard dependency on llguidance's `rayon` feature.** `cmake/external/onnxruntime_external_deps.cmake` calls `corrosion_import_crate()` without `FEATURES`, so this relies on `rayon` staying in llguidance's `default = ["lark", "rayon", "referencing", "ahash"]`. Without it, `llg_par_compute_mask` calls `cc.set_error(...)` on every constraint, which clears the parser (`self.constraint = None`) *permanently* - so every guided request in such a build would fail unrecoverably rather than fall back to `llg_compute_mask`. Now that the Engine decode path depends on the parallel API, pinning `FEATURES rayon` explicitly would make that dependency non-silent.

Related: the "leaves failed cursors dirty so mask construction retries" comments are accurate for injected/transient failures, but a genuine llguidance mask error poisons the cursor for good (`llg_get_error` keeps returning the same pointer until the constraint is freed), so the retry will fail every step. That is arguably the right failure mode - failing loudly beats silently unconstraining - but the comments read as if recovery is expected.

### Positives

Grammar/tokenizer caching, single-flight, COW cursors, and async mask overlap are each individually testable and each individually tested (concurrency, eviction with live processors, failed compilation not retained, pending mask outliving a removed processor, rollback isolation, `Continue()` transactionality). Consolidating validation into `ValidateGuidanceRequest()` removes the duplicated and slightly divergent checks from `Request`'s constructor. The docs accurately match the implemented limits, including the subtlety that `cached_grammars` counts ready entries while `cached_key_bytes` also covers in-flight compilations.

Specific file/line feedback is inline.

src/constrained_logits_processor.cpp
```diff
@@ -33,6 +212,227 @@ std::shared_ptr<const GeneratorParams> SnapshotGuidanceParams(
...
+      while (!cache->lru.empty() &&
+             (cache->grammars.size() >= kMaxCachedGrammars ||
+              !FitsGrammarCacheKeyBytes(
+                  cache->cached_key_bytes, key.size()))) {
+        const std::string evicted_key = cache->lru.back();
+        cache->lru.pop_back();
+        cache->cached_key_bytes -= evicted_key.size();
+        cache->grammars.erase(evicted_key);
+        ++cache->grammar_evictions;
+      }
```
tianleiwu: Two eviction-loop issues worth tightening:

**1. An oversized key flushes the entire shared cache.** `FitsGrammarCacheKeyBytes` returns `false` unconditionally when `key.size() > kMaxCachedGrammarKeyBytes`, so for a >16 MiB `guidance_data` the `!Fits(...)` disjunct is never satisfied and the loop only terminates when `lru` is empty. One such request evicts all 64 entries shared by every other request on the model - and then still does not cache itself. Worth short-circuiting first:

```cpp
const bool cacheable = FitsGrammarCacheKeyBytes(0, key.size());
if (cacheable) {
  while (...) { /* evict */ }
}
```

**2. Pending entries cause over-eviction.** `cache->grammars.size()` counts in-flight single-flight entries, but only *ready* entries are in `lru`. With N concurrent unique misses the loop must evict N extra ready entries just to push `grammars.size()` below `kMaxCachedGrammars`. The "64 retained entries after 800 misses" figure in the PR description only holds for serialized misses; under concurrency this thrashes. Gating capacity on `cache->lru.size()` (ready entries) instead would make the bound behave as documented.

src/constrained_logits_processor.cpp
```diff
@@ -33,6 +212,227 @@ std::shared_ptr<const GeneratorParams> SnapshotGuidanceParams(
...
+    } catch (...) {
+      promise->set_exception(std::current_exception());
+      if (cached_entry) {
+        std::lock_guard lock(cache->mutex);
+        const auto inserted = cache->grammars.find(key);
+        if (inserted != cache->grammars.end()) {
+          cache->cached_key_bytes -= key.size();
+          cache->grammars.erase(inserted);
+        }
```
tianleiwu: This cleanup erases the entry from `grammars` and decrements `cached_key_bytes`, but never removes the key from `cache->lru`.

Today that is benign, because the entry is only pushed onto `lru` on the success path and nothing after `push_front` can throw. But it is an unguarded invariant with a nasty failure mode: if a future edit ever marks the entry ready before a throwing step, the stale `lru` key survives, a later eviction subtracts its bytes a *second* time, and `size_t cached_key_bytes` underflows. After that `FitsGrammarCacheKeyBytes` is permanently `false` and the cache is effectively disabled while evicting on every single miss.

Cheap insurance - erase the LRU node when it is not `end()`, or assert it is not:

```cpp
if (inserted->second.lru_iterator != cache->lru.end())
  cache->lru.erase(inserted->second.lru_iterator);
```

src/constrained_logits_processor.h
```diff
@@ -27,6 +35,10 @@ struct ConstrainedLogitsProcessor {
   // Based on the masks which are derived from constraints, it sets the logits to -inf for invalid tokens
   virtual void ProcessLogits(DeviceSpan<float> logits) = 0;
 
+  // Returns a host-resident mask row when the processor supports scheduler-owned batched
+  // application. An empty span asks the scheduler to use ProcessLogits() instead.
+  virtual std::span<const uint32_t> GetReadyMask() { return {}; }
```
tianleiwu, line 44: The returned span aliases `masks_`, which `EnsureMaskReady()` reassigns wholesale on the next call. The single current caller (`CollectBatchedGuidanceMasks`) copies immediately and is correct, but the contract on this virtual does not say the span is invalidated by the next call on the same processor.

llguidance documents exactly this for `LlgMaskResult::sample_mask` ("valid until any subsequent `llg_*` call on the same constraint"); mirroring that wording here would stop a future caller from caching the span across a step.

src/constrained_logits_processor.h
```diff
@@ -50,63 +70,96 @@ struct GuidanceLogitsProcessor : public ConstrainedLogitsProcessor {
...
-  std::vector<std::unique_ptr<LlgConstraint, LlgConstraintDeleter>> llg_constraints_;
-  std::unique_ptr<LlgTokenizer, LlgTokenizerDeleter> llg_tokenizer_;
-  std::shared_ptr<Tokenizer> tokenizer_;
+  size_t mask_words_per_row_{};
+  std::vector<uint32_t> masks_;
+  std::shared_future<std::vector<uint32_t>> pending_masks_;
+  bool mask_dirty_{};
+  DeviceSpan<uint32_t> device_masks_;
+  std::shared_ptr<const struct GuidanceGrammarAsset> grammar_asset_;
+  std::vector<std::shared_ptr<LlgConstraint>> llg_constraints_;
```
tianleiwu, line 127: This ordering is load-bearing: reverse member destruction frees `llg_constraints_` before `grammar_asset_` releases the tokenizer, which is what satisfies llguidance's "`llg_free_tokenizer` must not be called while there are still constraints using it".

`GuidanceTokenizerAsset` carries an explicit comment for the same reason; this declaration and `GuidanceGrammarAsset` (`tokenizer` before `initial_constraint`) do not. A one-liner here would stop a future reorder from silently introducing a use-after-free that only reproduces at teardown.

src/engine/scheduled_requests.cpp
```diff
@@ -201,6 +209,106 @@ bool ScheduledRequests::TryGenerateNextTokensBatched(std::vector<DeviceSpan<floa
...
+    ScheduleGuidanceMaskComputation(processors);
+  } catch (const std::exception& error) {
+    if (g_log.enabled && g_log.warning) {
+      Log("warning") << "Guidance mask precomputation was deferred: "
+                     << error.what() << std::endl;
+    }
+  } catch (...) {
+    if (g_log.enabled && g_log.warning) {
+      Log("warning",
+          "Guidance mask precomputation was deferred after a non-standard exception.");
```
tianleiwu, line 331: `ScheduleGuidanceMaskComputation()` can throw `std::logic_error("A dirty guidance cursor already has pending mask work.")`, which signals a broken invariant rather than a transient failure. Catching it here alongside genuine transients and logging at `warning` means that, if it ever fires, the whole batch silently loses async precompute on every step - a hard latency cliff that is invisible unless warning logging happens to be enabled.

Suggest splitting it out so an invariant violation is at least diagnosable:

```cpp
} catch (const std::logic_error& error) {
  Log("error") << "Guidance mask precomputation invariant violated: " << error.what() << std::endl;
  assert(false);
} catch (const std::exception& error) {
  ...
```

Keeping the function `noexcept` (so a committed step stays valid) is the right call; this is only about not losing the signal.

src/models/model.h
```diff
@@ -161,6 +161,11 @@ struct Model : std::enable_shared_from_this<Model>, LeakChecked<Model>, External
 
   SessionInfo session_info_;
 
+  // Lazily initialized, model-local constrained-decoding assets. The concrete state lives in
+  // constrained_logits_processor.cpp so builds without llguidance do not depend on its types.
+  mutable std::mutex guidance_cache_mutex_;
```
tianleiwu, line 167: `std::mutex` is used here but `<mutex>` is not included by this header - it currently works only through a transitive include. Please add `#include <mutex>` next to the existing `<map>` / `<optional>` includes so this does not break when an unrelated header trims its includes.

(The `std::shared_ptr<GuidanceCacheState>` with an incomplete type is fine, since `shared_ptr` captures the deleter at construction in `constrained_logits_processor.cpp`.)

src/python/python.cpp
```diff
@@ -595,6 +595,15 @@ PYBIND11_MODULE(onnxruntime_genai, m) {
       .def_property_readonly("type", [](const OgaModel& model) -> std::string { return model.GetType().p_; })
       .def_property_readonly(
           "device_type", [](const OgaModel& model) -> std::string { return model.GetDeviceType().p_; }, "The device type the model is running on")
+      .def("get_guidance_cache_stats", [](const OgaModel& model) {
+        pybind11::dict stats;
+        for (const char* name : {"tokenizer_initializations", "grammar_hits", "grammar_misses",
+                                 "grammar_waits", "grammar_compile_microseconds", "grammar_evictions",
+                                 "cached_grammars", "cached_key_bytes"}) {
+          stats[name] = model.GetGuidanceCacheCount(name);
+        }
```
tianleiwu: This loops over eight names, and each `GetGuidanceCacheCount` call takes the model's guidance-cache mutex and builds a fresh `GuidanceCacheStats` snapshot. So the dict costs eight lock acquisitions and, more importantly, **is not a consistent snapshot** - under concurrent traffic `grammar_hits` and `grammar_misses` in the same returned dict can come from different points in time, which undermines the counters' main use (deriving a hit rate).

A single struct-returning entry point (or exposing `GetGuidanceCacheStats` once and filling the dict from it) would fix both. It would also remove the third copy of the counter-name list - it is currently duplicated here, in `OgaModelGetGuidanceCacheCount`, and in the `ort_genai_c.h` doc comment, with nothing keeping them in sync.

src/ort_genai_c.h
```diff
@@ -424,6 +424,17 @@ OGA_EXPORT OgaResult* OGA_API_CALL OgaCreateModelFromConfig(const OgaConfig* con
...
+/**
+ * \brief Gets a constrained-decoding cache counter for a loaded model.
+ * \param[in] model The model whose cache is queried.
+ * \param[in] name Counter name: tokenizer_initializations, grammar_hits, grammar_misses,
+ * grammar_waits, grammar_compile_microseconds, grammar_evictions,
+ * cached_grammars, or cached_key_bytes.
+ * \param[out] value The counter value.
+ */
+OGA_EXPORT OgaResult* OGA_API_CALL OgaModelGetGuidanceCacheCount(
+    const OgaModel* model, const char* name, uint64_t* value);
```
tianleiwu: This is a new public C API, so the shape is worth settling now: a string-keyed getter turns what could be a compile-time error into a runtime `OgaResult` for every typo, and it commits us to keeping the eight valid names in sync across three places (this doc comment, the `if/else` chain in `ort_genai_c.cpp`, and the Python binding).

An `OgaGuidanceCacheCounter` enum - or a single call filling an `OgaGuidanceCacheStats` struct - would be more conventional for this header, cheaper for callers that want all counters, and would make adding a counter a one-place change.

benchmark/engine/scenario_dispatcher.cpp
```diff
@@ -77,8 +77,9 @@ void RegisterExecutionProviderLibraries(const std::vector<ScenarioConfig>& confi
         config.execution_provider == "cuda" ? "CUDAExecutionProvider" : "WebGpuExecutionProvider";
 
     if (config.execution_provider_library.empty()) {
-      throw std::invalid_argument(
-          "execution_provider_library is required when execution_provider is '" + config.execution_provider + "'");
+      // A source-built ORT may expose its provider through the legacy provider path instead of a
+      // separately registered plugin EP. In that case AppendProvider() below performs the setup.
+      continue;
```
tianleiwu: Understood why this is needed for source-built ORT, but replacing the explicit `invalid_argument` with a silent `continue` also swallows the common misconfiguration it was catching: an omitted or typo'd `execution_provider_library` for a genuine plugin-EP scenario now surfaces as a later, much less obvious `AppendProvider()` failure.

An explicit opt-in in `ScenarioConfig` (e.g. `use_builtin_provider`) - or, minimally, an informational log here noting that the legacy provider path is being assumed - would keep both cases diagnosable.

src/ort_genai_c.h
baijumeswani: We probably don't need a public c api for this. I will remove the new apis added. The telemetry can be internal only.

benchmark/engine/scenario_dispatcher.cpp
baijumeswani: benchmarking changes need not be part of this pr. I'll remove them and address the provider-bridge ep in another pr.

src/python/python.cpp
baijumeswani: Removing public c/c++/python api. It is not necessary to expose these metrics.

src/constrained_logits_processor.cpp
baijumeswani: Oversized keys now compile without touching the shared cache, and pending compilations no longer consume the 64 ready-entry limit.

src/constrained_logits_processor.cpp
baijumeswani: Failure cleanup now removes any LRU entry before erasing the cache entry and updates both total and pending byte accounting.

src/constrained_logits_processor.h
baijumeswani: Added the lifetime contract. Callers must not retain the returned span across another call on the same processor.

src/constrained_logits_processor.h
baijumeswani: Added comments documenting the required declaration order for the tokenizer, grammar asset, and request-local constraints.

src/constrained_logits_processor.cpp
```diff
@@ -216,73 +587,181 @@ void GuidanceLogitsProcessor::CommitTokens(std::span<int32_t> tokens) {
...
+            std::shared_ptr<LlgConstraint>(clone, GuidanceLogitsProcessor::LlgConstraintDeleter{});
+      }
+    }
+    candidates.push_back(guidance);
+  }
+  if (candidates.empty()) {
+    return;
+  }
+
+  auto job = std::make_unique<ParallelMaskJob>();
```
tianleiwu, line 699: **High priority:** This runs once per guided decode token, but it allocates a `ParallelMaskJob`, promise/future state, multiple vectors for constraints/assets/outputs/steps, and the full batched mask buffer. The completion callback then allocates and copies another `std::vector<uint32_t>` for every request before the scheduler copies those masks again into its reusable buffer. At large vocabularies and concurrent batch sizes, that leaves substantial allocator traffic and memory bandwidth in the exact path this PR is optimizing. Please move the job/step/mask storage into scheduler-owned reusable storage or a bounded pool and deliver results into pre-sized processor-owned buffers. If reuse is not practical, add allocation and throughput measurements at the maximum supported batch/vocabulary size that demonstrate this churn is negligible.

tianleiwu (changes requested):
The cache ownership, async lifetime, rollback, and CUDA fallback design is careful and well tested. One performance blocker remains: the guided decode loop still constructs heap-owned job/future/vector state and copies each completed mask on every generated token, which conflicts with the zero-allocation inference-path requirement and this PR's production-latency goal. Separately, the PR description still says cache telemetry is exposed through C, C++, and Python, but those APIs were intentionally removed after the earlier review; please update the description to say the counters are internal only.

src/constrained_logits_processor.cpp
baijumeswani: I validated the impact at the largest production shape supported by this model rather than assuming it was negligible.
 
Test shape: Qwen3.8-27B INT4 on A100 80 GB, vocabulary size 248,320, Engine batch B8 (the model’s configured maximum), 4K-token prompt, 128 generated tokens, with paired guided/unguided runs.
 
Measured guided overhead:
  - mean inter-token latency: +0.408%
  - upper 95% confidence bound for mean ITL: +0.578%
  - mean decode throughput: -0.313%
  - median decode throughput: -0.318%
 
Structurally, the B8 mask is about 248 KB per contiguous mask matrix ( 8 × ceil(248,320 / 32) × 4 ). The current path materializes the job mask and per-request completed masks, then copies into reusable scheduler storage—roughly 0.75 MB of mask writes/copies per decode step, or about 96 MB across 128 tokens. Despite that deliberately worst-case vocabulary and batch size, the paired measurements keep the total observed decode regression below 0.6% at the upper 95% confidence bound.
 
I agree this is not allocation-free. I avoided moving callback output directly into request- or scheduler-owned buffers because mask jobs may outlive request cancellation, turn rollback, or Engine teardown; preserving independent job ownership is what makes those paths safe. Given the measured B8/248K result, I think we could keep the current lifetime-safe implementation and treating pooled reusable job storage as a follow-up optimization if we establish a stricter zero-allocation requirement.

tianleiwu (approved):

PR #2468, opened 2026-08-25
[benchmark] Add prompt token truncation and trim benchmark output

amd-namalr:
## Benchmark (C): prompt token truncation and cleaner verbose output

Improves prompt handling and the verbose print output in the C `model_benchmark`.

### Changes

- **Truncate a text prompt to `-l` tokens.** `-l/--prompt_length` can now be combined with `--prompt` or `--prompt_file`. The prompt is encoded and truncated to the requested number of tokens. If the prompt has fewer tokens than requested, a warning is printed to `stderr` and all available tokens are used. Previously `-l`, `--prompt`, and `--prompt_file` were mutually exclusive (`--prompt` and `--prompt_file` still are, with each other).
- **Trim printed input prompt.** In verbose mode, add new option `--prompt_print_chars N` so that only the last N characters of a large prompt are printed (with a leading `...`), instead of dumping the entire prompt.
- **Print only generated tokens.** The verbose `[OUTPUT ...]` block now decodes only the newly generated tokens instead of the full prompt + output sequence.
- **Preserve exact generated prompt length.** Use prompt token IDs instead of decode/re-encoding them and fail clearly if prompt generation unexpectedly produces too few tokens.

### Implementation notes

- `Options` gains `std::optional<size_t> prompt_truncate_tokens`. Parsing now reads `-l` and the text prompt into separate locals and reconciles them afterward, so a text prompt can carry a truncation length while `-l` alone still selects the synthetic generated prompt.
- Truncation is done at the token level: the encoded prompt's token IDs are sliced to the requested count and fed verbatim via `OgaSequences::Append`, and the displayed prompt is re-derived from those exact tokens.

### Testing

- `--prompt_file big.txt -l 128` -> prompt truncated to 128 tokens.
- `--prompt_file small.txt -l 100000` -> warning printed, all available tokens used.
- Large prompt with `-v --prompt_print_chars 256` -> prompt shows only its last 256 chars with a leading `...`; `[OUTPUT ...]` shows only generated text.

Assisted-by: Claude Opus 4.8 <noreply@anthropic.com>

### DIFF
--- a/benchmark/c/main.cpp
+++ b/benchmark/c/main.cpp
@@ -218,7 +218,11 @@ void RunBenchmark(const benchmark::Options& opts) {
                   << " (-l/--prompt_length); using all " << encoded_count << " tokens.\n";
       }
     }
-    prompt_tokens.assign(encoded_data, encoded_data + effective_count);
+    if (effective_count > 0) {
+      prompt_tokens.assign(encoded_data, encoded_data + effective_count);
+    } else {
+      prompt_tokens.clear();
+    }
     num_prompt_tokens = effective_count;
     // Reflect the tokens actually used (after truncation) in the displayed prompt.
     prompt = std::string{tokenizer->Decode(prompt_tokens.data(), prompt_tokens.size())};
@@ -316,9 +320,12 @@ void RunBenchmark(const benchmark::Options& opts) {
       }
       // Print only the generated tokens, not the prompt echoed back.
       const auto total_len = gen->TokenCount();
-      const auto* output_sequence_data = gen->GetSequenceData(0);
       const size_t gen_len = total_len > num_prompt_tokens ? total_len - num_prompt_tokens : 0;
-      const auto output = tokenizer->Decode(output_sequence_data + num_prompt_tokens, gen_len);
+      std::string output;
+      if (gen_len > 0) {
+        const auto* output_sequence_data = gen->GetSequenceData(0);
+        output = std::string{tokenizer->Decode(output_sequence_data + num_prompt_tokens, gen_len)};
+      }
       std::cout << "[OUTPUT BEGIN]" << output << "[OUTPUT END]\n";
     }
   }
--- a/benchmark/c/main.cpp
+++ b/benchmark/c/main.cpp
@@ -191,6 +191,9 @@ void RunBenchmark(const benchmark::Options& opts) {
   size_t num_prompt_tokens;
   bool need_generate_prompt = false;
   std::string prompt;
+  // Populated when an explicit set of prompt token IDs should be used verbatim
+  // (a text prompt from --prompt/--prompt_file, optionally truncated to -l tokens).
+  std::vector<int32_t> prompt_tokens;
 
   if (opts.use_random_tokens) {
     num_prompt_tokens = std::get<size_t>(opts.prompt_num_tokens_or_content);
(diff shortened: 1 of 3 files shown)


benchmark/c/main.cpp
```diff
@@ -202,7 +205,23 @@ void RunBenchmark(const benchmark::Options& opts) {
...
+      const size_t requested = *opts.prompt_truncate_tokens;
+      if (encoded_count >= requested) {
+        effective_count = requested;
+      } else {
+        std::cerr << "Warning: prompt has " << encoded_count
+                  << " tokens, fewer than the requested " << requested
+                  << " (-l/--prompt_length); using all " << encoded_count << " tokens.\n";
+      }
+    }
+    prompt_tokens.assign(encoded_data, encoded_data + effective_count);
```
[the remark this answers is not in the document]
amd-namalr: Addressed with https://github.com/microsoft/onnxruntime-genai/pull/2468/changes/0e1224287349f76993c92b624afdc150cef5364c

VishalX:
@kunal-vaishnavi can you review this?

benchmark/c/main.cpp
```diff
@@ -202,7 +205,28 @@ void RunBenchmark(const benchmark::Options& opts) {
...
+      const size_t requested = *opts.prompt_truncate_tokens;
+      if (encoded_count >= requested) {
+        effective_count = requested;
+      } else {
+        std::cerr << "Warning: prompt has " << encoded_count
+                  << " tokens, fewer than the requested " << requested
+                  << " (-l/--prompt_length); using all " << encoded_count << " tokens.\n";
+      }
+    }
+    if (effective_count > 0) {
```
kunal-vaishnavi, line 221: From Copilot:

When `effective_count` is 0 (empty prompt), `prompt_tokens.assign(encoded_data, encoded_data + effective_count)` performs pointer arithmetic on a potentially null `encoded_data`, which is undefined behavior even when the offset is 0. The same issue occurs when `tokenizer->Decode(prompt_tokens.data(), prompt_tokens.size())` is called with an empty `prompt_tokens` vector — `std::vector::data()` may return `nullptr`, and passing that to the detokenizer can crash.

Guard these operations to only proceed when the token count is greater than 0:

```cpp
if (effective_count > 0) {
  prompt_tokens.assign(encoded_data, encoded_data + effective_count);
  prompt = std::string{tokenizer->Decode(prompt_tokens.data(), prompt_tokens.size())};
} else {
  prompt_tokens.clear();
  prompt.clear();
}
```

This ensures no pointer arithmetic or decode calls happen on null pointers, even with a 0-token count (e.g., `--prompt ... -l 0`).

benchmark/c/main.cpp
```diff
@@ -236,7 +260,21 @@ void RunBenchmark(const benchmark::Options& opts) {
...
-    prompt = std::string{tokenizer->Decode(output_sequence_data, output_sequence_length)};
+    if (output_sequence_length < num_prompt_tokens) {
+      throw std::runtime_error(
+          "Generated prompt has fewer tokens than requested by -l/--prompt_length.");
+    }
+
+    // Keep the generated token IDs instead of decoding and re-encoding them.
+    // Tokenizer round trips are not token-count preserving, so re-encoding can
+    // produce more than the exact prompt length requested with -l.
+    if (num_prompt_tokens > 0) {
```
kunal-vaishnavi, line 271: Same comment as [this](https://github.com/microsoft/onnxruntime-genai/pull/2468#discussion_r4089145827)

benchmark/c/main.cpp
```diff
@@ -251,6 +289,11 @@ void RunBenchmark(const benchmark::Options& opts) {
       });
       prompt_sequences->Append(random_tokens.data(), random_tokens.size());
     }
+  } else if (!prompt_tokens.empty()) {
```
kunal-vaishnavi, line 292: From Copilot:

This condition treats an empty `prompt_tokens` vector as though no exact token sequence was produced. For a valid zero-token prompt—such as `--prompt ... -l 0`—execution falls through to `tokenizer->Encode(prompt.c_str(), ...)`, potentially reintroducing tokens and violating the requested truncation.

Please make the exact-token path authoritative for all non-random prompts, including an empty sequence, rather than selecting it based on `prompt_tokens.empty()`. This may require tracking whether `prompt_tokens` was populated intentionally, then appending the sequence even when its size is zero.

PR #2469, opened 2026-08-25
Enable MultiModal Rewind

Anirudh-Swaminathan:
Enable MultiModal model Rewind functionality

- Add DecoderState::RewindTo to reset position inputs, KV cache, and recurrent state, mirroring DecoderOnly_State.
- Add MultiModalPipelineState::RewindTo to forward to the decoder state and reset is_prompt_ so continuous decoding re-enters the prompt stage.


src/models/io/multi_modal_features.cpp
```diff
@@ -73,9 +75,17 @@ void MultiModalFeatures::ReuseFeaturesBuffer(MultiModalFeatures& other) {
...
+  // Take ownership of other's computed tensor as this input.
+  shape_ = other.shape_;
   features_ = std::move(other.features_);
-  state_.inputs_[index_] = other.state_.outputs_[other.index_];
+  state_.inputs_[index_] = features_.get();
+
+  // Give other a real copy so its output binding has valid data before its next Run() overwrites it.
+  other.shape_ = other.native_shape_;
+  other.features_ = OrtValue::CreateTensor(other.model_.p_device_->GetAllocator(), other.shape_, other.type_);
+  ByteWrapTensor(*other.model_.p_device_, *other.features_)
```
baijumeswani, line 86: This copies the full image or audio feature tensor after every prompt, including normal runs that never rewind. For large inputs, that can add noticeable memory use and time. Could we create a fresh output tensor without copying old data, since the encoder will overwrite it on the next replay? Another option is to keep the computed features and reuse them instead of running the encoder again.

src/models/multi_modal.cpp
```diff
@@ -927,8 +937,6 @@ DeviceSpan<float> MultiModalPipelineState::Run(int current_length, DeviceSpan<in
                       : decoder_state_->Run(current_length, next_tokens, next_indices);
 
     is_prompt_ = false;
```
baijumeswani, line 939: We now keep the vision and speech states alive for the whole generation. Those states can hold large input and output buffers, so memory may stay much higher after the prompt. Could we rebuild these states only when a full rewind happens, or keep only the small amount of data needed to replay the prompt?

src/models/multi_modal.cpp
```diff
@@ -941,6 +949,13 @@ DeviceSpan<float> MultiModalPipelineState::Run(int current_length, DeviceSpan<in
   return decoder_state_->Run(current_length, next_tokens, next_indices);
 }
 
+void MultiModalPipelineState::RewindTo(size_t index) {
```
baijumeswani, line 952: This enables a new rewind path, but there is no test that runs a multimodal prompt, generates tokens, rewinds to zero, and runs the prompt again. Please add an end-to-end test that compares the replay with a fresh generator. It would be helpful to cover image and audio inputs, repeated rewinds, Qwen position reset, and the stateful OpenVINO or QNN case that motivated the change.

src/generator/generators.cpp
```diff
@@ -962,7 +962,7 @@ SpeculativeStats Generator::GetSpeculativeStats() const {
 }
 
 void Generator::RewindToLength(size_t new_length) {
-  if (model_->config_->model.type == "whisper" || model_->config_->model.type == "phi3v" || model_->config_->model.type == "decoder-pipeline" || model_->config_->model.type == "lfm2")
+  if (model_->config_->model.type == "whisper" || model_->config_->model.type == "decoder-pipeline" || model_->config_->model.type == "lfm2")
```
baijumeswani: This is now one of the main behavior changes in the live diff. The stateful cache factory work described by the PR title and body already landed in #2465, so the current description is misleading. Please rebase and squash the fixup commits, then update the title and body to explain that this PR enables safe multimodal rewind and prompt replay.

baijumeswani (commented):
The rewind state changes look consistent, but I have concerns about added memory/copy cost, retained encoder state, missing end-to-end coverage, and the PR description no longer matching the live diff. I left recommendations inline.

src/generator/generators.cpp
Anirudh-Swaminathan: Rebased on top of latest main now

src/models/io/multi_modal_features.cpp
Anirudh-Swaminathan: Thanks! Updated the logic to avoid the full copy now

src/models/multi_modal.cpp
Anirudh-Swaminathan: This one was a bit trickier. Technically, with Rewind, the features can't be completely reset, but I've added a flag to recompute if rewound to 0, but will continue to hold the features otherwise.

PR #2490, opened 2026-08-28
Export DSpark block drafters for Qwen 3.5 and 3.8

tianleiwu:
## Summary

Add builder support for exporting a DSpark block drafter alongside Qwen 3.5 and 3.8 target models. The exporter uses selected target-model residual streams and emits the DSpark graph and configuration as an alternative block-drafting path to DFlash 2.

## Stack

Depends on #2489.

## Changes

- Add `dspark_path`, `dspark_num_draft_tokens`, and `dspark_top_k` builder options.
- Require paged attention and make DSpark and DFlash 2 export mutually exclusive.
- Validate the DSpark target-layer metadata against the configured auxiliary residual taps.
- Build `dspark.onnx`, export its configuration, and share compatible target-model initializers.
- Suppress MTP-head export when a DSpark drafter is requested.
- Add coverage for validation, option overrides, graph/configuration output, and MTP suppression.

## Testing

```bash
PYTHONPATH=src/python/py python3 -m pytest test/python/builder/test_qwen_dspark_export.py -q
```

15 passed.


src/python/py/models/builders/dspark.py
```diff
@@ -0,0 +1,801 @@
...
+            self.graph.outputs.append(self.values[f"present.{i}.value"])
+
+    def _make_lm_head(self, root):
+        w = self._weights
+        weight, scale = w["lm_head.weight"], w["lm_head.weight_scale"]
+        if weight.dtype != torch.float8_e4m3fn:
+            name = "/lm_head/MatMul"
+            return self.matmul(
+                name, root, weight, self.hidden_size, self.vocab_size, "num_block", weight_name="lm_head.MatMul.weight"
+            )
```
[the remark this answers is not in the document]
tianleiwu: Addressed in ec7f6ef3. The LM-head scale is now loaded only for FP8 weights, and a missing FP8 scale raises a clear ValueError. Added regression coverage for both non-FP8 and missing-scale FP8 heads.

src/python/py/models/builders/dspark.py
```diff
@@ -0,0 +1,801 @@
...
+        return name
+
+    # --------------------------------------------------------------- op sugar
+
+    def _out(self, name):
+        return f"{name}/output_0"
+
+    def unary(self, op, name, x, dtype, shape, domain="", **attrs):
+        out = self._out(name)
+        self.make_node(op, [x], [out], name=name, domain=domain, **attrs)
```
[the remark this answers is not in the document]
tianleiwu: Addressed in ec7f6ef3. Renamed the DSpark helper methods and internal state to remove leading underscores and updated all call sites and tests.

baijumeswani (commented):
This PR adds DSpark export support for Qwen3.5 and Qwen3.8. It builds a separate `dspark.onnx` block drafter from a SpecForge checkpoint, feeds it selected target-model hidden states, uses the target embedding and LM head, and emits the same candidate-and-score lattice consumed by the block-drafter runtime. It also adds builder options for the checkpoint path, draft depth, and lattice width, updates `genai_config.json`, and suppresses the MTP head when DSpark is selected.

The DSpark-specific math looks strong. The YaRN calculation matches the reference formula, the PagedAttention inputs and KV shapes match the runtime contract, the Markov pair scores use the expected axes, and the candidate lattice matches the runtime's greedy walk. Promoting the small Markov dot product to FP32 is sensible. The residual-tap offset is clearly explained and enforced, and the updated FP8/non-FP8 LM-head handling is better than the first version. Normal Qwen, MTP, and DFlash exports remain unchanged when DSpark is not requested.

I still see several compatibility and maintainability issues that can produce an unloadable model, out-of-range token IDs, unnecessarily large packages, or confusing errors. The most important fixes are to align the draft-count limits with the consumer, validate that the target and drafter use one token vocabulary, make initializer sharing match the documented behavior, handle RoPE and FP8 scale formats explicitly, and reduce the large amount of duplicated DFlash2 code. The inline comments below include concrete suggestions.

src/python/py/models/builders/dspark.py
```diff
@@ -0,0 +1,804 @@
...
+        self.is_causal = bool(cfg.get("is_causal", False))
+
+        self.mask_token_id = int(dfl["mask_token_id"])
+        self.target_layer_ids = list(dfl["target_layer_ids"])
+        self.markov_rank = int(cfg.get("markov_rank", dfl.get("markov_rank", 0)))
+        if self.markov_rank <= 0:
+            raise ValueError("A DSpark drafter needs markov_rank > 0.")
+        # One query row per draft; a smaller request just takes a prefix of the same block.
+        try:
+            self.block_size = int(num_draft_tokens) if num_draft_tokens is not None else int(cfg["block_size"])
```
baijumeswani: DSpark uses the draft count as `block_size`. The builder accepts `1`, but the consumer requires a block size greater than 1, so that option produces a package that cannot load. An override larger than the checkpoint's trained block size is also accepted even though only taking a smaller prefix is justified. **Recommendation:** require `2 <= num_draft_tokens <= checkpoint block_size`, update the public option text, and give a clear error for both limits.

src/python/py/models/builders/dspark.py
```diff
@@ -0,0 +1,804 @@
...
+        self.vocab_size = int(cfg["vocab_size"])
+        self.rms_eps = float(cfg["rms_norm_eps"])
+        self.max_position = int(max_position_embeddings or cfg["max_position_embeddings"])
+        self.rope_parameters = dict(cfg["rope_parameters"])
+        self.sliding_window = int(cfg["sliding_window"]) if cfg.get("use_sliding_window") else -1
+        # The dual-source block attends to itself bidirectionally; DFlash's attention module
+        # hard-codes is_causal=False and the published configs do not override it.
+        self.is_causal = bool(cfg.get("is_causal", False))
+
+        self.mask_token_id = int(dfl["mask_token_id"])
```
baijumeswani, line 97: The vocabulary size comes from the drafter config, but the embedding and LM head come from the target, while the Markov tables come from the drafter. There is no check that all of these use the same token IDs. A reduced-vocabulary checkpoint or an out-of-range MASK token can cause a bad `Gather` or return draft IDs the target does not understand. **Recommendation:** validate equal row counts for the target embedding, target LM head, `markov_w1`, `markov_w2`, and the configured vocabulary; require `0 <= mask_token_id < vocab_size`; and reject checkpoints containing a draft-to-target ID map until remapping is implemented.

src/python/py/models/builders/dspark.py
```diff
@@ -0,0 +1,804 @@
...
+        self.graph.outputs.append(self.values["draft_candidate_ids"])
+        self.graph.outputs.append(self.values["draft_scores"])
+        for i in range(self.num_layers):
+            self.graph.outputs.append(self.values[f"present.{i}.key"])
+            self.graph.outputs.append(self.values[f"present.{i}.value"])
+
+    def make_lm_head(self, root):
+        w = self.weights
+        weight = w["lm_head.weight"]
+        if weight.dtype != torch.float8_e4m3fn:
```
baijumeswani: The module says the target LM head can be shared, but this branch transposes it and casts it to the drafter's BF16 type. An FP16 target therefore has a different dtype and cannot share this initializer, which can duplicate more than a gigabyte for a Qwen-sized vocabulary. **Recommendation:** keep the shareable initializer in the target's dtype/layout and add any needed cast or transpose in the graph. If only the FP8 head is intended to share, narrow the documentation instead.

src/python/py/models/builders/dspark.py
```diff
@@ -0,0 +1,804 @@
...
+        if scale is None:
+            raise ValueError("FP8 LM head weight is missing 'lm_head.weight_scale'.")
+        # The quantized head is the target's and only runs in the target's dtype. Its input is the
+        # drafter's final normed hidden state, which is back inside the fp16 range.
+        root = self.unary(
+            "Cast", "/lm_head/Cast", root, self.external_dtype, ["num_block", self.hidden_size], to=self.external_dtype
+        )
+        self.make_initializer(weight.contiguous(), "lm_head.MatMul.fp8_weight")
+        self.make_initializer(
+            scale.reshape(self.vocab_size, 1), "lm_head.MatMul.fp8_weight_scale", to=ir.DataType.FLOAT
```
baijumeswani: This assumes every FP8 scale already has one value per vocabulary row. The loader also maps `weight_global_scale` here, but that can be a scalar and this reshape then fails. Other FP8 dtypes currently fall through the normal-weight path without a clear error. **Recommendation:** explicitly allow only the FP8 type supported by the target builder and reuse `prepare_matmul_block_quantized_scales()` to normalize scalar, row, and transposed scale layouts.

src/python/py/models/builders/dspark.py
```diff
@@ -0,0 +1,804 @@
...
+        interpolation = 1.0 / (factor * pos_freqs)
+        ramp = np.clip((np.arange(dim // 2, dtype=np.float32) - low) / (high - low), 0.0, 1.0)
+        weight = 1.0 - ramp
+        inv_freq = interpolation * (1.0 - weight) + extrapolation * weight
+        return inv_freq.astype(np.float64), float(attention_factor)
+
+    def make_rope_caches(self):
+        # PagedAttention type-constrains cos/sin to the query's element type. YaRN's attention
+        # factor scales cos/sin, exactly as HF's rotary embedding does, so it is baked in here.
+        if str(self.rope_parameters.get("rope_type", "default")).lower() in ("yarn", "longrope"):
```
baijumeswani: LongRoPE and YaRN do not use the same formula. This branch sends `longrope` through YaRN and ignores LongRoPE's short/long per-dimension factors, which can create a wrong cache or a raw missing-key error. **Recommendation:** accept only `yarn` in this implementation and reject unsupported RoPE types with a clear message, or reuse the base builder's LongRoPE cache logic.

src/python/py/models/builders/qwen.py
```diff
@@ -1076,6 +1087,108 @@ def add_dflash2_to_genai_config(self, out_dir):
...
+        from .dspark import DSparkBuilder  # noqa: PLC0415
+
+        print("Building DSpark draft model -> dspark.onnx")
+        target_dir = input_path if input_path and os.path.isdir(input_path) else self.decoder.model_name_or_path
+        self.dspark = DSparkBuilder(
+            self.dspark_path,
+            target_dir,
+            self.dspark_attrs["io_dtype"],
+            self.decoder.attention_attrs["paged_block_size"],
+            self.decoder.original_context_length or self.decoder.context_length,
```
baijumeswani: This prefers `original_context_length`, but DSpark emits one fixed rotary cache and uses absolute target positions. If the original training length is shorter than the target's supported context, long requests can index past the DSpark cache. **Recommendation:** size this cache to the target `context_length` and validate it is at least as long as every supported request position.

src/python/py/models/builders/qwen.py
```diff
@@ -1076,6 +1087,108 @@ def add_dflash2_to_genai_config(self, out_dir):
...
+
+        ``dspark_path`` points at the draft checkpoint. SpecForge taps the *output* of each
+        ``target_layer_ids`` entry (``hidden_states[layer_id + 1]``), which is the residual stream
+        entering layer ``layer_id + 1`` -- the tensor aux_hidden_state_layers names. Getting that
+        off by one leaves acceptance at exactly 1.0.
+        """
+        self.dspark_path = extra_options.get("dspark_path")
+        if not self.dspark_path:
+            return
+        if self.dflash2_path:
```
baijumeswani, line 1167: This mutual-exclusion check runs after DFlash2 initialization. If both paths are supplied, DFlash2 may fail first with an unrelated layer-tap message, so users never see this clearer error. **Recommendation:** reject multiple block-drafter options before either drafter is initialized, using the existing `requested_block_drafter` helper.

src/python/py/models/builders/dspark.py
```diff
@@ -0,0 +1,804 @@
...
+                raise ValueError(f"Could not find '{required}' in the target checkpoint '{self.target_dir}'.")
+        self.weights = weights
+        return weights
+
+    # ------------------------------------------------------------------ save
+
+    def save_model(self, out_dir):
+        out_path = os.path.join(out_dir, self.filename)
+        data_path = out_path + ".data"
+        for path in (out_path, data_path):
```
baijumeswani: Existing output files are deleted before the replacement is fully saved. If saving fails, the output folder is left with neither the old model nor a complete new one. **Recommendation:** write the model and data to temporary files and rename them only after both writes succeed, following the staging approach already used by initializer sharing.

src/python/py/models/builders/dspark.py
```diff
@@ -0,0 +1,804 @@
...
+import os
+
+import numpy as np
+import onnx_ir as ir
+import torch
+from onnx_ir.tensor_adapters import TorchTensor, to_torch_dtype
+from tqdm import tqdm
+
+
+class DSparkBuilder:
```
baijumeswani: Most of this builder repeats DFlash2's graph helpers, PagedAttention construction, MLP, LM head, IO declarations, weight loading, save logic, and config generation. The two copies have already started to drift, and future fixes will need to be made twice. **Recommendation:** extract a shared block-drafter builder for the common graph and file/config logic. Keep only DSpark-specific RoPE, plain layers, and Markov-lattice code in this class.

src/python/py/models/builders/dspark.py
```diff
@@ -0,0 +1,804 @@
...
+    def binary(self, op, name, a, b, dtype, shape):
+        out = self.out(name)
+        self.make_node(op, [a, b], [out], name=name)
+        self.make_value(out, dtype, shape)
+        return out
+
+    def reshape(self, name, x, shape_const, dtype, shape):
+        return self.binary("Reshape", name, x, self.const(shape_const), dtype, shape)
+
+    def matmul(self, name, x, weight_tensor, in_features, out_features, rows, weight_name=None):
```
baijumeswani: `in_features` is passed at every call site but never checked. A transposed or mismatched checkpoint weight can therefore produce a graph with declared shapes that do not match the actual initializer. **Recommendation:** validate each weight has shape `[out_features, in_features]` here and raise a message naming the node and actual shape. Apply similar checks to the Markov tables and `fc.weight`.

src/python/py/models/README.md
```diff
@@ -324,6 +325,20 @@ python -m onnxruntime_genai.models.builder -i path_to_target_model -o path_to_ou
 python builder.py -i path_to_target_model -o path_to_output_folder -p fp16 -e cuda -c cache_dir_for_hf_files --extra_options use_paged_attention=true aux_hidden_state_layers=1,11,21 dflash2_path=path_to_dflash2_checkpoint dflash2_num_draft_tokens=4
 ```
 
+#### Build a DSpark Block Drafter
+
+Set `dspark_path` to a DSpark checkpoint to export an auxiliary `dspark.onnx` block drafter beside a Qwen3.5 or Qwen3.8 target model. The target must use paged attention. SpecForge identifies the target layers whose outputs are tapped, while `aux_hidden_state_layers` identifies residual streams entering layers, so each configured auxiliary layer must be one greater than the corresponding `target_layer_ids` entry in the DSpark checkpoint. The drafter reuses the target's embedding and LM-head initializers. `dspark_path` and `dflash2_path` are mutually exclusive.
```
baijumeswani: DSpark uses full-attention KV by default, so its cache adds bytes for every target cache block and reduces the target's maximum resident context. The drafter body also runs in BF16 even when the target is FP16. These are important deployment costs. **Recommendation:** document the cache-capacity tradeoff, BF16 requirement, default draft-count source, and that `dspark_path` replaces the MTP head. A practical upper bound for `dspark_top_k` would also help because the score tensor and host copy grow with `top_k²`.

src/python/py/models/builders/dspark.py
tianleiwu: Addressed in 35cd70aa. Both the composite and direct builder now require `2 <= num_draft_tokens <= checkpoint block_size`, with boundary tests and updated option documentation.

src/python/py/models/builders/dspark.py
tianleiwu: Addressed in 35cd70aa. The shared block-drafter validation now checks configured vocabulary rows across the target embedding, target LM head, and DSpark Markov tables; validates `mask_token_id`; and rejects draft-to-target ID maps until remapping is implemented.

src/python/py/models/builders/dspark.py
tianleiwu: Addressed in 35cd70aa. Non-FP8 LM heads now retain the target initializer's name, transposed layout, and external dtype, with an activation Cast before MatMul when the drafter body is BF16. The same fix now covers DFlash2 through the shared base.

src/python/py/models/builders/dspark.py
tianleiwu: Addressed in 35cd70aa. The shared LM-head path accepts only `float8_e4m3fn`, requires a scale, and reuses `prepare_matmul_block_quantized_scales()` for scalar, row, and transposed-row layouts. Focused tests cover each supported layout.

src/python/py/models/builders/dspark.py
tianleiwu: Addressed in 35cd70aa. DSpark now supports only `default` and `yarn` in this implementation and reports a clear error for `longrope` or other unsupported RoPE types.

src/python/py/models/builders/qwen.py
tianleiwu: Addressed in 35cd70aa. DSpark's fixed rotary cache is now sized from the target decoder's supported `context_length`, with a regression test that distinguishes it from `original_context_length`.

src/python/py/models/builders/qwen.py
tianleiwu: Addressed in 35cd70aa. `requested_block_drafter()` now detects multiple requested drafters and raises before MTP, DFlash2, or DSpark initialization can produce a less relevant error.

src/python/py/models/builders/dspark.py
tianleiwu: Addressed in 35cd70aa. Block-drafter model and data files are now serialized in a temporary staging directory and installed with backup/rollback handling. Tests cover both serialization failure and partial replacement failure while preserving existing files.

src/python/py/models/builders/dspark.py
tianleiwu: Addressed in 35cd70aa. Common graph/value helpers, checked MatMul emission, LM-head handling, I/O declarations, staged saving, validation, and config generation now live in `BlockDrafterBuilder`; DSpark and DFlash2 retain only their architecture-specific RoPE, layer, and lattice logic.

src/python/py/models/builders/dspark.py
tianleiwu: Addressed in 35cd70aa. The shared MatMul helper validates `[out_features, in_features]` at every call, and DSpark validates `fc.weight`, both Markov tables, the embedding, and LM head against checkpoint dimensions before graph construction.

src/python/py/models/README.md
tianleiwu: Addressed in 35cd70aa. The public DSpark documentation now covers BF16 execution, extra KV-cache capacity and resident-context tradeoffs, checkpoint-derived draft defaults and bounds, MTP replacement, and the quadratic `top_k` score/host-copy cost.

src/python/py/models/builders/dflash2.py
```diff
@@ -39,13 +39,11 @@
...
 import torch
-from onnx_ir.tensor_adapters import TorchTensor, to_torch_dtype
-from tqdm import tqdm
 
-from builders.base import Model
+from .block_drafter import BlockDrafterBuilder
 
 
-class DFlash2Builder:
+class DFlash2Builder(BlockDrafterBuilder):
```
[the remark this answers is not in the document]
tianleiwu: `DFlash2Builder` no longer has a superclass to initialize.

Fixed in e30fb955. Root cause was the same for every alert in this batch: `BlockDrafterBuilder` inherited from `Model` purely to reuse `prepare_matmul_block_quantized_scales`, but a block drafter emits a hand-written graph and shares none of the decoder builder's contract, so `Model.__init__` was never called and every graph method was an incompatible override.

`BlockDrafterBuilder` is now a standalone class. The two helpers it actually needs (`stamp_build_metadata`, `prepare_matmul_block_quantized_scales`) are called explicitly on `Model`, and the latter is now a `@staticmethod` since it never touched `self`.

src/python/py/models/builders/dspark.py
```diff
@@ -0,0 +1,568 @@
...
+import os
+
+import numpy as np
+import onnx_ir as ir
+import torch
+
+from .block_drafter import BlockDrafterBuilder
+
+
+class DSparkBuilder(BlockDrafterBuilder):
```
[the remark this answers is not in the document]
tianleiwu: `DSparkBuilder` no longer has a superclass to initialize.

Fixed in e30fb955. Root cause was the same for every alert in this batch: `BlockDrafterBuilder` inherited from `Model` purely to reuse `prepare_matmul_block_quantized_scales`, but a block drafter emits a hand-written graph and shares none of the decoder builder's contract, so `Model.__init__` was never called and every graph method was an incompatible override.

`BlockDrafterBuilder` is now a standalone class. The two helpers it actually needs (`stamp_build_metadata`, `prepare_matmul_block_quantized_scales`) are called explicitly on `Model`, and the latter is now a `@staticmethod` since it never touched `self`.

src/python/py/models/builders/dflash2.py
```diff
@@ -825,64 +690,9 @@
...
-                        f"past_key_values.{i}.{suffix}",
-                        self.io_dtype,
-                        ["num_blocks", self.paged_block_size, self.num_kv_heads, self.head_size],
-                    )
-                )
-
     # --------------------------------------------------------------- weights
 
-    def _load_weights(self):
+    def load_weights(self):
```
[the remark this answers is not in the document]
tianleiwu: `DFlash2Builder.load_weights` no longer overrides `Model.load_weights`.

Fixed in e30fb955. Root cause was the same for every alert in this batch: `BlockDrafterBuilder` inherited from `Model` purely to reuse `prepare_matmul_block_quantized_scales`, but a block drafter emits a hand-written graph and shares none of the decoder builder's contract, so `Model.__init__` was never called and every graph method was an incompatible override.

`BlockDrafterBuilder` is now a standalone class. The two helpers it actually needs (`stamp_build_metadata`, `prepare_matmul_block_quantized_scales`) are called explicitly on `Model`, and the latter is now a `@staticmethod` since it never touched `self`.

src/python/py/models/builders/dspark.py
```diff
@@ -0,0 +1,568 @@
...
+            attention_factor = 1.0
+        pos = np.arange(self.max_position, dtype=np.float64)[:, None]
+        freqs = pos * inv_freq[None, :]
+        for name, fn in (("dspark.cos_cache", np.cos), ("dspark.sin_cache", np.sin)):
+            values = (fn(freqs) * attention_factor).astype(np.float32)
+            self.make_initializer(torch.from_numpy(values), name, to=self.io_dtype)
+
+    # ------------------------------------------------------------------ graph
+
+    def make_model(self):
```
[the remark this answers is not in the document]
tianleiwu: `DSparkBuilder.make_model` no longer overrides `Model.make_model`.

Fixed in e30fb955. Root cause was the same for every alert in this batch: `BlockDrafterBuilder` inherited from `Model` purely to reuse `prepare_matmul_block_quantized_scales`, but a block drafter emits a hand-written graph and shares none of the decoder builder's contract, so `Model.__init__` was never called and every graph method was an incompatible override.

`BlockDrafterBuilder` is now a standalone class. The two helpers it actually needs (`stamp_build_metadata`, `prepare_matmul_block_quantized_scales`) are called explicitly on `Model`, and the latter is now a `@staticmethod` since it never touched `self`.

src/python/py/models/builders/dspark.py
```diff
@@ -0,0 +1,568 @@
...
+            )
+            hidden = self.make_mlp(i, y, w, rows_q)
+
+        final, _ = self.skip_rms_norm("/dspark/norm", residual, hidden, w["norm.weight"], rows_q, want_sum=False)
+
+        self.make_candidates_and_markov(final, w)
+        self.graph.sort()
+        return self.model
+
+    def make_attention(self, i, x, ctx_kv, rows_q):
```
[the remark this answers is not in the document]
tianleiwu: `DSparkBuilder.make_attention` no longer overrides `Model.make_attention`.

Fixed in e30fb955. Root cause was the same for every alert in this batch: `BlockDrafterBuilder` inherited from `Model` purely to reuse `prepare_matmul_block_quantized_scales`, but a block drafter emits a hand-written graph and shares none of the decoder builder's contract, so `Model.__init__` was never called and every graph method was an incompatible override.

`BlockDrafterBuilder` is now a standalone class. The two helpers it actually needs (`stamp_build_metadata`, `prepare_matmul_block_quantized_scales`) are called explicitly on `Model`, and the latter is now a `@staticmethod` since it never touched `self`.

src/python/py/models/builders/dspark.py
```diff
@@ -0,0 +1,568 @@
...
+        return self.matmul(
+            f"{p}/attn/o_proj/MatMul",
+            block_out,
+            w[f"layers.{i}.self_attn.o_proj.weight"],
+            self.num_heads * self.head_size,
+            self.hidden_size,
+            rows_q,
+        )
+
+    def make_mlp(self, i, x, w, rows_q):
```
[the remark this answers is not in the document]
tianleiwu: `DSparkBuilder.make_mlp` no longer overrides `Model.make_mlp`.

Fixed in e30fb955. Root cause was the same for every alert in this batch: `BlockDrafterBuilder` inherited from `Model` purely to reuse `prepare_matmul_block_quantized_scales`, but a block drafter emits a hand-written graph and shares none of the decoder builder's contract, so `Model.__init__` was never called and every graph method was an incompatible override.

`BlockDrafterBuilder` is now a standalone class. The two helpers it actually needs (`stamp_build_metadata`, `prepare_matmul_block_quantized_scales`) are called explicitly on `Model`, and the latter is now a `@staticmethod` since it never touched `self`.

src/python/py/models/builders/dspark.py
```diff
@@ -0,0 +1,568 @@
...
+
+        self.graph.outputs.append(self.values["draft_candidate_ids"])
+        self.graph.outputs.append(self.values["draft_scores"])
+        for i in range(self.num_layers):
+            self.graph.outputs.append(self.values[f"present.{i}.key"])
+            self.graph.outputs.append(self.values[f"present.{i}.value"])
+
+    # --------------------------------------------------------------- weights
+
+    def load_weights(self):
```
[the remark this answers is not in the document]
tianleiwu: `DSparkBuilder.load_weights` no longer overrides `Model.load_weights`.

Fixed in e30fb955. Root cause was the same for every alert in this batch: `BlockDrafterBuilder` inherited from `Model` purely to reuse `prepare_matmul_block_quantized_scales`, but a block drafter emits a hand-written graph and shares none of the decoder builder's contract, so `Model.__init__` was never called and every graph method was an incompatible override.

`BlockDrafterBuilder` is now a standalone class. The two helpers it actually needs (`stamp_build_metadata`, `prepare_matmul_block_quantized_scales`) are called explicitly on `Model`, and the latter is now a `@staticmethod` since it never touched `self`.

src/python/py/models/builders/block_drafter.py
```diff
@@ -0,0 +1,351 @@
...
+        return output
+
+    def resolve_sliding_window(self, config):
+        """Return the single ``local_window_size`` every layer runs with, or -1 for full attention."""
+        if not config.get("use_sliding_window"):
+            return -1
+        window = int(config["sliding_window"])
+        # PagedAttention takes one window per model, so a checkpoint that windows only a suffix of
+        # its layers (HF's `max_window_layers`, or a mixed `layer_types`) cannot be exported as-is.
+        if int(config.get("max_window_layers", 0)) != 0:
```
baijumeswani: This rejects every non-zero `max_window_layers`, including checkpoints where all layers are windowed. The qualified Qwen3.8 DFlash2 checkpoint has `num_hidden_layers=5`, `max_window_layers=5`, and five `sliding_attention` entries, so this change prevents that model from exporting even though its window is uniform. **Recommendation:** when `layer_types` is present, require its length to match `num_hidden_layers` and require every entry to be `sliding_attention`; then accept a matching `max_window_layers` such as `num_hidden_layers` (and any explicitly supported zero convention). Only reject values that actually describe a mixed window. Please add a regression case with 5 layers, `max_window_layers=5`, and all five layer types set to `sliding_attention`.

baijumeswani (commented):
The latest update addresses the earlier DSpark concerns well, but the new shared window check introduces one blocking DFlash2 regression. The inline comment includes the real checkpoint shape and a suggested validation rule.

src/python/py/models/builders/block_drafter.py
tianleiwu: Addressed in 11d618f7. `resolve_sliding_window` now accepts the two uniform-window conventions (`max_window_layers=0` or `max_window_layers=num_hidden_layers`), validates that an explicit `layer_types` list has one entry per hidden layer, and still requires every entry to be `sliding_attention`. Added the requested regression with 5 layers, `max_window_layers=5`, and five sliding-attention entries, plus negative coverage for partial and mismatched declarations. Both published DSpark and DFlash2 configs now instantiate with the expected attention settings, and the full focused suite passes (79 tests).

src/python/py/models/builders/base.py
```diff
@@ -2234,7 +2234,8 @@ def make_matmul_op(self, matmul, basename, root_input, **kwargs):
         else:
             raise NotImplementedError(f"The {self.onnx_dtype} precision is not currently supported.")
 
-    def prepare_matmul_block_quantized_scales(self, weight_scale, out_features, block_count):
+    @staticmethod
```
baijumeswani, line 2237: model builder coding-style standards suggest not to use staticmethod. non-blocking.

baijumeswani (approved):

src/python/py/models/builders/base.py
tianleiwu: I will update it later.

tianleiwu: Addressed in #2497 by 95fdbcd6. `prepare_matmul_block_quantized_scales` is now a class method, which keeps the shared `Model` helper available to the standalone block-drafter builder without using `staticmethod`.

PR #2491, opened 2026-08-28
Verify speculative drafts in dynamic Engine steps

tianleiwu:
## Summary
- Thread draft proposals through Request/Scheduler/Engine and verify them inside a dynamic step.
- Narrow paged KV and compact fixed-state transactions to the accepted prefix.
- Defer or reject unsafe draft rows based on cache and state-update capacity.
- Use compact state-update replay without restoring dense recurrent checkpoints.

## Testing
- `engine_unit_tests`: 285 passed.

## Dependency
- Stacked on #2454 only.
- No ONNX Runtime dependency or version bump.

### DIFF
--- a/src/engine/cache_manager.cpp
+++ b/src/engine/cache_manager.cpp
@@ -50,12 +50,11 @@ class CompositeCacheStepReservation final : public CacheStepReservation {
       if (fixed_state_pool) {
         // Fixed and paged track the same per-request cache-slot boundary: the fixed target_tokens
         // mirror the paged target so both commit at one token boundary. The pool infers resident
-        // vs. new ownership itself, so no newly_admitted flag is passed. capture_count stays 0
-        // until MTP scheduling requests compact state_update captures, so no step currently
-        // reaches FixedStateReservation::CommitPrefix.
+        // vs. new ownership itself, so no newly_admitted flag is passed.
         fixed_requests.push_back(FixedStateReservationRequest{
             entry.request_id,
             entry.target_cache_slots,
+          entry.draft_token_count,
         });
       }
       if (entry.newly_admitted) {
@@ -75,6 +74,10 @@ class CompositeCacheStepReservation final : public CacheStepReservation {
 
   PagedCacheReservation* PagedReservation() override {
     return &*paged_reservation_;
+  }
+
+  FixedStateReservation* FixedReservation() override {
+    return fixed_reservation_ ? &*fixed_reservation_ : nullptr;
   }
 
   std::span<const FixedStateSlotHandle> FixedStateSlots() const override {
@@ -500,12 +503,19 @@ StepPlanningResult PagedCacheManager::PlanStepResources(StepPlan& plan) const {
     throw StepPlanningConsistencyError(
         "Paged planning selected more admissions than fixed state can reserve.");
   }
-
+  const bool capture_state_updates = std::any_of(
+      plan.requests.begin(), plan.requests.end(),
+      [](const RequestStepPlan& entry) { return entry.draft_token_count != 0; });
+  if (capture_state_updates && !fixed_state_pool_->SupportsStateUpdates()) {
+    throw std::logic_error(
+        "Planned draft verification on a model whose fixed state declares no compact updates.");
+  }
   plan.fixed_state = FixedStateResourcePlan{
       true,
       plan.requests.size(),
       new_slot_count,
-      fixed_state_pool_->PlannedStagingBytes(plan.requests.size()),
+      fixed_state_pool_->PlannedStagingBytes(
+          plan.requests.size(), capture_state_updates),
   };
   return result;
 }
--- a/src/engine/cache_manager.cpp
+++ b/src/engine/cache_manager.cpp
@@ -54,7 +54,7 @@ class CompositeCacheStepReservation final : public CacheStepReservation {
         fixed_requests.push_back(FixedStateReservationRequest{
             entry.request_id,
             entry.target_cache_slots,
-          entry.draft_token_count,
+            entry.draft_token_count,
         });
       }
       if (entry.newly_admitted) {
--- a/src/engine/cache_manager.cpp
+++ b/src/engine/cache_manager.cpp
@@ -98,6 +98,20 @@ class CompositeCacheStepReservation final : public CacheStepReservation {
     return fixed_reservation_ ? fixed_reservation_->NewSlotCount() : 0;
   }
 
+  void CommitPrefix(size_t row, const void* request_id,
+                    size_t step_tokens, size_t kept_tokens) override {
+    if (prepared_ || committed_) {
+      throw std::logic_error(
+          "Composite cache step reservation no longer accepts prefix commits.");
+    }
+    // Fixed first: it is the only one of the two that can reject the request (no checkpoints, or a
+    // step longer than the checkpoint window), so the paged boundary is never lowered on its own.
+    if (fixed_reservation_) {
+      fixed_reservation_->CommitPrefix(row, step_tokens, kept_tokens);
+    }
+    paged_reservation_->CommitPrefix(request_id, step_tokens, kept_tokens);
+  }
+
   void ValidateCommit() const override {
     if (committed_) {
       throw std::logic_error(
@@ -438,6 +452,20 @@ void PagedCacheManager::Deallocate(std::vector<std::shared_ptr<Request>>& reques
 }
 
 bool PagedCacheManager::SupportsDynamicBatching() const { return true; }
+
+size_t PagedCacheManager::MaxDraftTokensPerStep() const {
+  // Recurrent state can only be replayed through the compact transitions the operators captured.
+  // Without fixed state the paged KV boundary alone decides, and any tail slot can be left
+  // uncommitted.
+  size_t limit = kMaxDraftTokensPerStep;
+  if (fixed_state_pool_) {
+    limit = std::min(limit, fixed_state_pool_->StateUpdateCapacity());
+  }
+  if (const size_t query_cap = MaxQueryTokensPerRequest(); query_cap != 0) {
+    limit = std::min(limit, query_cap - 1);
+  }
+  return limit;
+}
 
 std::vector<std::shared_ptr<Request>> PagedCacheManager::AllocatedRequests() const {
   return cache_allocated_requests_;
--- a/src/engine/decoders/varlen_decoder_io.cpp
+++ b/src/engine/decoders/varlen_decoder_io.cpp
@@ -261,7 +261,9 @@ void VarlenDecoderIO::PrepareInputIds(std::shared_ptr<DecoderOnly_Model> model, 
     // The operator writes token j at past_sequence_lengths[i] + j. The processed cursor is the
     // number of tokens already in the cache and therefore the base position for this step.
     const int64_t processed_sequence_length = request->ProcessedSequenceLength();
-    if (entry && (request->CurrentSequenceLength() != entry->sequence_length_before ||
+    if (entry && (request->CurrentSequenceLength() !=
+                      entry->sequence_length_before +
+                          static_cast<int64_t>(entry->draft_token_count) ||
                   SlotsAfterStep(processed_sequence_length, entry->unprocessed_token_count) !=
                       entry->target_cache_slots)) {
       throw std::runtime_error("Step plan processed sequence length does not match the request.");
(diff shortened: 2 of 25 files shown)


src/engine/scheduler.cpp
```diff
@@ -283,7 +296,18 @@ StepPlanningResult DynamicBatchScheduler::PlanStep(StepPlan& plan) {
...
-  for (auto& entry : plan.requests) {
+  for (size_t i = 0; i < plan.requests.size(); ++i) {
+    auto& entry = plan.requests[i];
+    const size_t scheduling_index = entry.scheduling_order;
+    // The budget only distributes committed tokens. A step that stops short of the sequence tail
+    // has no row that predicts the first draft, so it cannot verify one.
+    if (token_counts[scheduling_index] != selected_remaining_lengths[scheduling_index]) {
+      entry.draft_token_count = 0;
+    }
+    entry.unprocessed_token_count = token_counts[scheduling_index] + entry.draft_token_count;
```
baijumeswani: At this point `PlanStepResources()` has already run using the one-token target. Adding drafts here can make the reservation need an extra KV block or a wider block table that was never planned. When a draft crosses a block boundary, the reservation can throw and the Engine is marked unhealthy instead of deferring for capacity. These extra rows also exceed `max_scheduled_tokens` because the budget was assigned before drafts were added.

Recommendation: include the optimistic draft count in cache demand before `PlanStepResources()` and charge drafts against the global token budget. A real `PagedCacheManager` test with a small block size and a draft crossing the boundary should cover both block allocation and block-table width.

src/engine/request.cpp
```diff
@@ -254,8 +254,103 @@ int64_t Request::CurrentSequenceLength() const {
...
+  if (draft_count == 0) {
+    return;
+  }
+  if (staged_draft_count_ != 0 || draft_count > draft_tokens_.size()) {
+    throw std::logic_error("The step staged more draft tokens than the request proposed.");
+  }
+
+  const std::span<const int32_t> drafts{draft_tokens_.data(), draft_count};
+  auto device_tokens = AllocateOnDevice(*params_, drafts);
+  search_->AppendTokens(device_tokens);
```
baijumeswani, line 323: `AppendTokens()` is the prompt/continuation path. It intentionally skips generated-token EOS handling and resets the done state. If the target model accepts an EOS draft, that EOS is therefore treated like input, a bonus token can be produced after it, and the request stays active. This differs from normal greedy generation.

Recommendation: commit accepted drafts with generated-token semantics (the existing `Search::CommitToken()` helper is designed for this), stop at the first accepted EOS, and do not publish a bonus token after the turn has completed. Please add CPU and CUDA tests with EOS as the first and a middle draft token.

src/engine/scheduled_requests.cpp
```diff
@@ -150,14 +160,56 @@ std::vector<DeviceSpan<float>> ScheduledRequests::ProcessLogits() {
...
+      continue;
+    }
+
+    // Row j predicts the token at committed_length + j, which is exactly where draft j sits. Accept
+    // the longest prefix of the proposal the target model would have produced on its own; the row
+    // after it holds the token that replaces the first rejected draft, or the bonus token.
+    const auto drafts = requests_[i]->StagedDraftTokens();
+    size_t accepted_count = 0;
+    while (accepted_count < draft_count) {
+      auto row_values = verify_rows[row + accepted_count].CopyDeviceToCpu();
```
baijumeswani: This host copy is inside the per-draft acceptance loop. A `DeviceSpan` row is only a view; `CopyDeviceToCpu()` copies its whole backing allocation and synchronizes the CUDA stream. In the common casted-logits path, all rows share one `[total_rows, vocab]` FP32 allocation, so this can copy the full logits tensor once for every accepted draft. For a large vocabulary and a multi-request batch, that can mean gigabytes of D2H traffic and dozens of synchronizations in one decode step.

Recommendation: run one batched argmax on the device and copy only token IDs, or at minimum copy the packed logits allocation once and inspect all verification rows from that single host copy. Please add a GPU benchmark or profile that reports verification D2H bytes and synchronization count.

src/engine/decoders/varlen_decoder_io.cpp
```diff
@@ -447,17 +452,29 @@ std::vector<DeviceSpan<float>> VarlenDecoderIO::ProcessLogits() {
...
         running_length += scheduled_requests_[i]->ScheduledTokenCount();
       }
     }
   } else {
+    const auto* plan = plan_;
+    if (plan && std::any_of(plan->requests.begin(), plan->requests.end(),
+                            [](const RequestStepPlan& entry) {
+                              return entry.draft_token_count != 0;
+                            })) {
+      throw std::runtime_error(
```
baijumeswani, line 473: This is a permanent model capability mismatch, but it is thrown during post-processing and becomes a retryable batch abort. Rollback keeps the draft proposal pending, so the next `Step()` plans the same drafts and fails again forever, also preventing other requests in the batch from progressing.

Recommendation: expose the per-token-logits capability through `Engine::MaxDraftTokensPerStep()` (return 0 when unsupported) so `SetDraftTokens()` rejects the proposal up front. Alternatively, classify this as non-retryable and clear the proposal before returning the error.

baijumeswani (dismissed):
## Overall review

This PR has a good overall shape. It keeps speculative verification inside the existing Engine transaction, checkpoints request and sampler state before drafts are staged, and narrows paged KV and fixed state to the same accepted prefix before publishing. The rollback path is thoughtfully designed, the packed logits-row mapping is consistent, and the latest row-order fix keeps request budgets attached correctly after fixed-state slot ordering. The tests cover full acceptance, first-token rejection, partial acceptance, rollback/retry, prefill deferral, guidance rejection, and ready-order behavior.

I found a few production issues that should be addressed before merging:

1. Draft tokens are added after paged-cache resource planning. A draft that crosses a KV block boundary can require blocks and block-table columns that were never planned, causing a fatal Engine failure instead of a normal capacity deferral. Drafts also bypass the configured `max_scheduled_tokens` limit.
2. Accepted draft tokens are appended through the prompt/input-token path, which resets EOS state. If an accepted draft is EOS, the Engine can emit tokens after EOS and continue the turn.
3. GPU verification copies logits to the CPU inside the per-draft loop. Because each row is a view of the shared logits allocation, each call can copy the full allocation and synchronize the stream. On a large-vocabulary model this can make verification much slower than normal decoding.
4. Models that do not return per-token logits fail during post-processing with a retryable error, while the draft proposal remains pending. Repeating `Step()` then fails forever and blocks the batch.

I recommend adding real `PagedCacheManager` coverage for a draft crossing a block boundary, CPU/CUDA tests for EOS inside the accepted prefix, a device-side batched argmax (or at least one host copy per step), and proposal-time rejection for models without per-token logits. It would also be helpful to add an end-to-end test that exercises real paged and fixed-state prefix commits together rather than only recording doubles.

The transaction design is promising, but the cache-planning and EOS issues are correctness blockers, and the logits-copy path needs to be fixed and benchmarked before this can provide a reliable performance benefit.

src/engine/scheduler.cpp
tianleiwu: Addressed in 9938a6bd. Draft rows now contribute to each candidate's target cache slots before PlanStepResources(), and decode extras are charged against max_scheduled_tokens while preserving one base token per selected request. SpeculativePlanReservesThePagedBlockBeforeExecution covers a block-size-4 PagedCacheManager boundary, and SpeculativeRowsStayWithinTheGlobalTokenBudget covers the batch budget.

src/engine/request.cpp
tianleiwu: Addressed in 9938a6bd. The temporary input-style proposal is now rewound and the matched prefix is replayed through Search::CommitToken(). If EOS or max length completes the search, verification keeps only the tokens actually appended and skips bonus sampling. CPU and CUDA engine tests cover EOS as the first and a middle draft token.

src/engine/scheduled_requests.cpp
tianleiwu: Addressed in 9938a6bd. SelectSampledRows() now runs one batched DeviceInterface::ArgMax over contiguous packed verification rows, so CUDA transfers only the resulting token IDs. The host fallback copies a shared logits backing allocation once before scanning its row views, instead of copying it once per draft.

src/engine/decoders/varlen_decoder_io.cpp
tianleiwu: Addressed in 9938a6bd. ModelExecutor now exposes whether the model supports draft verification; the production executor derives that capability from the logits symbolic shape. Engine::MaxDraftTokensPerStep() returns zero for batch-only logits, so SetDraftTokens() rejects the proposal before scheduling. The post-processing check remains as a defensive contract guard.

src/engine/scheduler.cpp
```diff
@@ -182,9 +190,16 @@ StepPlanningResult DynamicBatchScheduler::PlanStep(StepPlan& plan) {
...
     candidate.entry.request_id = request.get();
     candidate.entry.sequence_length_before = snapshot.current_sequence_length;
-    candidate.entry.unprocessed_token_count = 1;
+    // Drafts extend a decode step, which by definition ends at the sequence tail. A prefill chunk
+    // has committed tokens of its own left to push through, so it can never verify one.
+    candidate.entry.draft_token_count =
+        snapshot.is_prefill
+            ? 0
+            : std::min(request->PendingDraftTokenCount(), max_draft_token_count);
+    candidate.entry.unprocessed_token_count = 1 + candidate.entry.draft_token_count;
```
baijumeswani, line 199: This plans cache space for every proposed draft before the scheduler knows how many drafts will actually run. If the normal one-token decode fits but the drafts need another KV block, the cache planner can defer or reject the whole request. The proposal remains pending, so the same decision can repeat on every `Step()` even though ordinary decoding could continue.

**Recommendation:** treat drafts as optional cache demand. If the full draft set does not fit, reduce `draft_token_count`—down to zero if needed—and retry planning with the required one-token decode before deferring the request. Please add a test with no free blocks where the base token fits but the drafts cross a block boundary; the request should fall back to normal decode rather than stall.

src/engine/cache_manager.cpp
```diff
@@ -500,12 +531,19 @@ StepPlanningResult PagedCacheManager::PlanStepResources(StepPlan& plan) const {
     throw StepPlanningConsistencyError(
         "Paged planning selected more admissions than fixed state can reserve.");
   }
-
+  const bool capture_state_updates = std::any_of(
```
baijumeswani, line 534: This capture decision uses the draft counts **before** the global token budget is applied. The later budget can reduce every draft count to zero. On a fixed-state model, the plan then includes compact-capture bytes, but `ReserveStep()` receives zero capture counts and reports fewer bytes. The strict plan/reservation check treats that as a fatal contract mismatch and marks the Engine unhealthy.

**Recommendation:** compute fixed-state capture requirements after the final draft counts are known, or carry the planned capture flag into the reservation instead of deriving it twice. Please add a composite fixed-state test where the selected request count equals `max_scheduled_tokens`, leaving no budget for drafts; the step should complete normally with drafts disabled.

src/engine/scheduled_requests.cpp
```diff
@@ -150,14 +160,96 @@ std::vector<DeviceSpan<float>> ScheduledRequests::ProcessLogits() {
...
 
   return logits;
 }
 
+std::vector<DeviceSpan<float>> ScheduledRequests::SelectSampledRows(
+    std::vector<DeviceSpan<float>>& verify_rows) {
+  if (!sampling_plan_) {
+    throw std::logic_error("Draft verification requires scheduler-owned argmax storage.");
+  }
+  auto& predicted_tokens = sampling_plan_->verification_tokens;
```
baijumeswani, line 184: `SelectSampledRows()` is called on every dynamic Engine step, and this code runs argmax over all logits rows before checking whether any request actually has drafts. Normal non-speculative decoding therefore pays for an extra full-vocabulary argmax plus a CUDA synchronization even though none of these predicted IDs are used. On providers without device argmax, the fallback may also copy logits to the CPU unnecessarily.

**Recommendation:** add an early return when all `draft_token_counts_` are zero. When drafts are present, ideally run argmax only for the rows that verify draft tokens. A regression test or lightweight counter should confirm that ordinary Engine steps do not call the verification argmax path.

baijumeswani (commented):
The latest update fixes the earlier block-width planning, EOS handling, repeated full-logits copies, unsupported-logits retry loop, and global token-budget issues. The new device argmax and EOS-aware accepted-token commit are good improvements. I found three remaining edge cases worth addressing before relying on this in production; I left actionable inline suggestions below. This is non-blocking review feedback.

baijumeswani (approved):
Open comments can be addressed in the next pull-request in the stack

src/engine/scheduled_requests.cpp
tianleiwu: See https://github.com/microsoft/onnxruntime-genai/pull/2493/changes/4e69c0cf2ba8e33cabade41ea5b666050ebd1081

src/engine/scheduler.cpp
tianleiwu: See https://github.com/microsoft/onnxruntime-genai/pull/2493/changes/4e69c0cf2ba8e33cabade41ea5b666050ebd1081

src/engine/cache_manager.cpp
tianleiwu: See SpeculativeDraftFallsBackWhenOnlyBaseDecodeFitsCache test added in https://github.com/microsoft/onnxruntime-genai/pull/2493/changes/4e69c0cf2ba8e33cabade41ea5b666050ebd1081

PR #2493, opened 2026-08-28
Expose speculative draft proposals through runtime APIs

tianleiwu:
## Summary

Expose speculative draft proposals through the runtime APIs and verify them within the dynamic batching engine. The verify path preserves request transaction boundaries and uses one batched device argmax for contiguous logits rows, with the existing host path retained as a compatibility fallback.

## Stack

Depends on #2491.

This runtime change does not depend on an ONNX Runtime version bump.

## Changes

- Add C, C++, and Python APIs for submitting draft token proposals and querying the per-step draft limit.
- Plan paged block-table capacity for target tokens plus the proposed draft block.
- Preserve the pre-sampling sequence boundary when committing accepted draft tokens, including drafted stop tokens.
- Select verify rows with a batched device argmax when their logits are contiguous.
- Fall back to the existing host-copy argmax path when the device or layout cannot use the batched path.
- Add engine coverage for speculative request behavior and drafted stop-token completion.

## Testing

```bash
cmake --build build/Linux/Debug --parallel $(nproc)
./build/Linux/Debug/engine_unit_tests
```

286 tests passed.

The built Python module also exposes `Request.set_draft_tokens` and `Engine.max_draft_tokens_per_step`.


src/ort_genai_c.h
```diff
@@ -1431,6 +1431,32 @@ OGA_EXPORT OgaResult* OGA_API_CALL OgaRequestCancelTurn(
...
+ * (see OgaEngineMaxDraftTokensPerStep).
+ *
+ * \param[in] request The request to propose drafts for.
+ * \param[in] tokens One sequence holding the draft continuation, in order.
+ * \return OgaResult containing the error message if the proposal was rejected, or nullptr on success.
+ */
+OGA_EXPORT OgaResult* OGA_API_CALL OgaRequestSetDraftTokens(OgaRequest* request, const OgaSequences* tokens);
+
+/**
+ * \brief Reports how many speculative draft tokens one request may attach to a single engine step.
```
baijumeswani: The public Engine API no longer has a step operation. Callers submit turns, call Run, and consume events, and a Run may only drain queued events. Could we name this around the proposal instead, for example `OgaEngineMaxDraftTokensPerProposal`, and use the same name in C++ and Python? Internal code can still use step terminology.

baijumeswani (commented):
Overall, the core implementation looks solid. Accepted drafts, replacement or bonus tokens, terminal events, rollback, and event-buffer overflow now fit the Engine event API well. The normal path also avoids the extra planning pass when there are no drafts. My comments are mainly about making the new public API simple and predictable before it becomes part of the ABI: use proposal language instead of the internal step concept, accept a flat token list, keep replacement failures atomic, define when proposals are valid, and cover the public bindings and docs.

src/ort_genai_c.h
```diff
@@ -1431,6 +1431,32 @@ OGA_EXPORT OgaResult* OGA_API_CALL OgaRequestCancelTurn(
...
+ * \brief Proposes speculative draft tokens for the request's next engine step.
+ *
+ * The step then runs one row per draft on top of its own token, verifies each draft against the
+ * model's prediction, and keeps the accepted prefix. The proposal applies to the next step only:
+ * a committed step consumes it whether or not it could verify it. Passing an empty sequence clears
+ * the proposal. Requires a greedy request and an engine whose cache can roll a rejected draft back
+ * (see OgaEngineMaxDraftTokensPerStep).
+ *
+ * \param[in] request The request to propose drafts for.
+ * \param[in] tokens One sequence holding the draft continuation, in order.
```
baijumeswani, line 1443: `OgaSequences` suggests that callers can pass several sequences, but this API accepts exactly one flat token list. Could this take `const int32_t* tokens` and a token count, like `OgaRequestBeginTurn`? A C++ `std::span<const int32_t>` wrapper would also let Python avoid the extra `OgaSequences` allocation and copy.

src/ort_genai_c.h
```diff
@@ -1431,6 +1431,32 @@ OGA_EXPORT OgaResult* OGA_API_CALL OgaRequestCancelTurn(
  */
 OGA_EXPORT OgaResult* OGA_API_CALL OgaRequestClose(OgaRequest* request);
 
+/**
+ * \brief Proposes speculative draft tokens for the request's next engine step.
+ *
+ * The step then runs one row per draft on top of its own token, verifies each draft against the
+ * model's prediction, and keeps the accepted prefix. The proposal applies to the next step only:
+ * a committed step consumes it whether or not it could verify it. Passing an empty sequence clears
```
baijumeswani: Could replacing a proposal be failure-atomic? `Request::SetDraftTokens` currently clears the pending proposal before it validates the replacement. If the new proposal is invalid, the call fails but the previous valid proposal is lost. Please validate first and only replace the stored proposal after all checks pass. A valid -> invalid replacement test would capture the expected behavior.

src/ort_genai_c.h
```diff
@@ -1431,6 +1431,32 @@ OGA_EXPORT OgaResult* OGA_API_CALL OgaRequestCancelTurn(
...
+/**
+ * \brief Reports how many speculative draft tokens one request may attach to a single engine step.
+ *
+ * Zero means this engine cannot roll a rejected draft back, so OgaRequestSetDraftTokens will fail.
+ *
+ * \param[in] engine The engine to query.
+ * \param[out] out The maximum draft token count.
+ * \return OgaResult containing the error message on failure, or nullptr on success.
+ */
+OGA_EXPORT OgaResult* OGA_API_CALL OgaEngineMaxDraftTokensPerStep(const OgaEngine* engine, size_t* out);
```
baijumeswani: The Engine API contract says Engine operations must run on the owner thread, but this new query does not validate that rule. Please either add the same owner-thread check used by the other Engine operations or explicitly document this query as thread-safe.

src/python/python.cpp
```diff
@@ -864,6 +868,11 @@ PYBIND11_MODULE(onnxruntime_genai, m) {
           pybind11::arg("tokens"),
           pybind11::arg("turn_options") = pybind11::none())
       .def("cancel_turn", &OgaRequest::CancelTurn)
+      .def("set_draft_tokens", [](OgaRequest& request, ContiguousArray<int32_t> tokens) {
```
baijumeswani, line 871: Please reject arrays that are not one-dimensional, as `begin_turn` already does. Right now a shape such as `(2, 3)` is silently flattened into one six-token proposal. An `ndim() == 1` check and a small rejection test would make this behavior clear.

docs/SpeculativeDecoding.md
```diff
@@ -88,6 +94,40 @@ speculative_options = {
...
+## Engine caller-supplied proposals
+
+Dynamic Engine clients can attach a proposed continuation to one request:
+
+```python
+max_drafts = engine.max_draft_tokens_per_step()
+if max_drafts:
+    request.set_draft_tokens(np.asarray(draft_ids[:max_drafts], dtype=np.int32))
+
+ready = engine.step()
```
baijumeswani: This example uses the removed `step()` and unseen-token APIs. Could it use a reusable event buffer with `engine.run(buffer)` and iterate over token events instead? It would also be helpful to show that callers must drain retained events before another model execution. The same stale terminology appears in `paged_attention_engine.md`.

docs/SpeculativeDecoding.md
```diff
@@ -88,6 +94,40 @@ speculative_options = {
...
+`OgaRequest::SetDraftTokens`.
+
+This API has a different ownership boundary from the automatic `Generator` strategies:
+
+- The application owns proposal generation; the Engine only verifies the supplied IDs.
+- The request must already be submitted to the Engine, and only dynamic batching is supported.
+- Verification is greedy and accepts the longest prefix matching the target model's argmax rows.
+- Guidance and logits penalties are not supported for a request carrying a proposal.
+- The proposal applies to the next committed step. Budgeting or cache pressure may verify fewer
+  drafts, and prefill verifies none; either case still consumes the proposal. A rolled-back step
```
baijumeswani: Silently consuming a proposal during prefill seems hard for callers to reason about, especially because scheduling decides when prefill ends. Could `set_draft_tokens` accept proposals only when the request is ready to decode? Another valid design is to retain the proposal until the first eligible decode, but consuming it without verification is surprising.

test/cpp/c_api_tests.cpp
```diff
@@ -2460,3 +2460,12 @@ TEST(CAPITests, LoadAudiosFromBuffersRejectsEmptyBuffer) {
   // audios should not have been created
   EXPECT_EQ(audios, nullptr);
 }
+
+TEST(CAPITests, RequestSetDraftTokensRejectsNullArguments) {
```
baijumeswani, line 2464: This only tests null arguments. Could we add one successful public C API flow that queries the limit, submits a proposal, and checks the ordered token events? A Python test for proposal submission would also catch binding issues such as invalid array dimensions.

src/ort_genai_c.h
baijumeswani: I think ogasequences is fine here. Let's not address this in this PR since there are other API as well that use this ogasequences.

docs/SpeculativeDecoding.md
```diff
@@ -88,6 +94,51 @@ speculative_options = {
...
+`OgaRequest::SetDraftTokens`.
+
+This API has a different ownership boundary from the automatic `Generator` strategies:
+
+- The application owns proposal generation; the Engine only verifies the supplied IDs.
+- The request must have completed prefill and be ready to decode; only dynamic batching is
+  supported.
+- Verification is greedy and accepts the longest prefix matching the target model's argmax rows.
+- Guidance and logits penalties are not supported for a request carrying a proposal.
+- The proposal applies to the next committed decode operation. Budgeting or cache pressure may
```
baijumeswani, line 132: A pending proposal should belong only to the current Turn. If drafts are set and the Turn is cancelled before `Run`, cancellation marks the Turn complete but leaves those drafts pending. Could cancellation clear the pending proposal? A cancel-then-continue test would help make sure old drafts do not carry into the next Turn.

baijumeswani (approved):
open comment can be addressed in the next pr

docs/SpeculativeDecoding.md
tianleiwu: Addressed in #2494 by be5399429f. Successful cancellation now clears the current Turn pending proposal, and CancelClearsPendingDraftsBeforeContinuation verifies that drafts do not carry into the continued Turn. Retryable rollback still preserves the proposal within the same Turn.

PR #2494, opened 2026-08-28
Configure MTP models for paged runtime decoding

tianleiwu:
## Summary

Project exported `model.mtp` metadata into the paged decoder configuration used by the dynamic batch engine. Also bind the optional packed `hidden_states` decoder output that chained MTP drafting consumes.

This is a runtime-only change and does not depend on an ONNX Runtime package bump or include model-builder changes.

## Stack

Depends on #2493.

## Changes

- Parse and validate the MTP model configuration for paged decoding.
- Derive the internal MTP decoder configuration from the target decoder settings.
- Bind the optional packed `hidden_states` output in `VarlenDecoderIO`.
- Add focused configuration coverage for valid and invalid MTP projections.

## Testing

- `cmake --build build/Linux/Debug`
- MTP configuration and speculative unit tests: 52 passed.
- Packed decoder and speculative engine tests: 17 passed.
- Full `engine_unit_tests`: 286 passed.


src/config.cpp
```diff
@@ -20,6 +20,91 @@ namespace Generators {
...
+
+  auto projected = std::make_unique<Config>(config);
+  auto& decoder = projected->model.decoder;
+  decoder.filename = mtp.filename;
+  if (mtp.session_options) {
+    decoder.session_options = *mtp.session_options;
+    InheritProviderOptions(config.model.decoder.session_options,
+                           decoder.session_options);
+  }
+  decoder.run_options = mtp.run_options;
```
baijumeswani: Could we make the inheritance behavior consistent here?

When `model.mtp.session_options` is present, assigning it first drops parent settings such as `config_entries`, graph optimization, custom-op libraries, memory settings, and thread settings; only provider settings are merged back. The next assignments also clear parent `run_options` when the MTP field is omitted.

Recommendation: start with the parent decoder settings and overlay only the MTP values that were explicitly provided. Please add a test that keeps a parent `config_entry` and run option while overriding one MTP session/provider option. If an empty `shared_initializers` list intentionally means “do not share,” document and test that separately.

src/config.cpp
```diff
@@ -20,6 +20,91 @@ namespace Generators {
...
+  decoder.outputs.hidden_states = mtp.outputs.hidden_states;
+  decoder.outputs.present_key_names = mtp.outputs.present_key_names;
+  decoder.outputs.present_value_names = mtp.outputs.present_value_names;
+
+  // The MTP graph is one full-attention layer. Paged-attention metadata and block-table names are
+  // deliberately inherited above; fixed recurrent state and a main-model sliding window are not.
+  decoder.layer_types.assign(static_cast<size_t>(mtp.num_hidden_layers), "full_attention");
+  decoder.conv_cache_size = 0;
+  decoder.sliding_window.reset();
+  decoder.state_groups.reset();
```
baijumeswani, line 141: Since this projected config is meant to describe only the one-layer MTP head, could we clear the remaining main-model-only metadata too?

`state_update_capacity`, decoder pipeline entries, and `model.mtp` itself are still carried by the copied config. They are mostly inactive for the intended Qwen LLM path today, but they make the internal contract ambiguous and allow this helper to project an already projected config again.

Recommendation: reset the state-update and pipeline fields, clear `projected->model.mtp`, and reject any source configuration that the MTP runtime does not support. That will make future consumers fail early instead of depending on these fields staying accidentally inactive.

src/engine/decoders/varlen_decoder_io.cpp
```diff
@@ -437,6 +449,35 @@ void VarlenDecoderIO::PrepareLogits(std::shared_ptr<DecoderOnly_Model> model, Sc
...
+  // the hidden state of whichever tokens the verify step accepted. Note this is not necessarily
+  // the logits row count: a pruned LM head emits only one logits row per request.
+  const std::vector<int64_t> hidden_states_shape = {
+      static_cast<int64_t>(TokenCount(scheduled_requests)),
+      static_cast<int64_t>(model->config_->model.decoder.hidden_size)};
+  if (graph_buffers_ != nullptr && graph_buffers_->hidden_states != nullptr) {
+    graph_buffers_->hidden_states->CreateTensor(hidden_states_shape, /*make_static=*/true);
+    active_hidden_states_ = graph_buffers_->hidden_states.get();
+  } else {
+    hidden_states_ = std::make_unique<Tensor>(model->p_device_inputs_,
```
baijumeswani, line 473: Could we add runtime coverage for this output before the MTP consumer is wired in?

Right now, any paged model that exposes this output allocates and asks ORT to produce packed hidden states on every step, even when no MTP proposer is active. The hybrid Qwen path runs eagerly, so it also takes this allocation path every decode step. The current tests do not exercise this code.

Recommendation: add a small paged hidden-state fixture that checks dtype, `[total_tokens, hidden_size]` shape, packed row order, the pruned-logits case, and hybrid forwarding. When the consumer is added, gate this binding on actual MTP demand and reuse an eager buffer instead of allocating it each step.

baijumeswani (commented):
Overall, this is a good foundation for bringing model-provided MTP metadata into the continuous-batching Engine.

The stack does several things well: #2493 adds a clear caller-supplied draft API, treats drafts as optional scheduler work so they do not displace another request’s required decode, and keeps verification, accepted-prefix cache commit, rollback, EOS handling, and multi-token event delivery within the existing transaction contract. This PR then projects `model.mtp` into a one-layer paged decoder config, exposes packed hidden states for the future MTP head, and makes sure cancellation does not carry an old proposal into the next turn.

The main limitation is that this is still scaffolding: the projected config and hidden-state accessor do not yet have a production MTP consumer. I left a few non-blocking inline comments around configuration inheritance, cleaning main-model-only metadata from the projected config, and testing/gating the hidden-state output before the consuming PR lands.

The overall direction looks sound, and I did not find a transaction, rollback, EOS, or request-isolation issue. #2493 has strong API and Engine coverage. For #2494, I recommend addressing the inline items here or making them explicit requirements for the follow-up that wires in the MTP session.

src/config.cpp
tianleiwu: Addressed in 9a0d2cd2. MTP session options now overlay all explicitly provided values on the parent decoder settings, including scalar session fields, config entries, providers/provider options, and run options. Parent values remain when the MTP field is omitted. An empty MTP shared-initializer list remains an intentional no-sharing override and is now documented and covered by MtpDecoderConfigTest.

src/config.cpp
tianleiwu: Addressed in 9a0d2cd2 and 59e7765d. The projection now resets state_update_capacity, clears decoder pipeline entries, and clears projected model.mtp. Existing filename/geometry validation remains in place, and the test now verifies that attempting to project the already projected config fails immediately.

src/engine/decoders/varlen_decoder_io.cpp
tianleiwu: Addressed in 9a0d2cd2. Hidden-state binding is now gated on configured MTP demand, and the synthetic paged/composite fixtures provide real ORT runtime coverage for FP16 dtype, [total_tokens, hidden_size] shape, packed request/token row order, pruned logits, hybrid forwarding, and the no-MTP gating case. The eager-buffer reuse remains part of the consuming MTP-session work, where its lifetime and maximum token capacity can be owned by that consumer.

baijumeswani (approved):
Looks good.

PR #2495, opened 2026-08-28
Run chained MTP drafting in the batch engine

tianleiwu:
## Description

Adds chained multi-token-prediction drafting to the paged batch engine on top of #2494. The runtime keeps compact GDN and convolution state transitions, MTP draft inputs, and accepted-prefix publication coordinated with the target request transaction without restoring dense recurrent checkpoints or changing the ONNX Runtime dependency.

This replaces #2459 with the compact replay and composite-transaction design.

### Chained MTP execution

- Runs configured MTP heads sequentially and verifies variable-length draft prefixes in the target batch.
- Keeps hidden-state and token handoff on device between MTP heads when supported.
- Supports greedy and sampled target requests, including deterministic batched sampling for draft verification.
- Rolls target, auxiliary KV, fixed state, search state, RNG state, and staged tokens back together on failure.

### Scheduling and state management

- Reserves MTP KV capacity alongside target paged KV and compact fixed-state resources.
- Defers drafts when decode and prefill requests share a step, avoiding unsupported mixed speculative rows while still progressing both requests.
- Discards rejected speculative rows so paged cache capacity can be reused.
- Preserves compact fixed-state replay using one convolution value or packed GDN capsule per transition; no intermediate dense checkpoints are captured.

### Telemetry and APIs

- Aggregates speculative step, draft, accepted-token, and accepted-length statistics in the engine.
- Exposes cumulative telemetry through the C++, C, and Python APIs as `Engine::GetSpeculativeStats` / `OgaEngineGetSpeculativeStats` / `Engine.get_speculative_stats`.

## Testing

- `build/Linux/Debug/engine_unit_tests` (295/295 passed)
- `cmake --build build/Linux/Debug` (passed)
- Python import/API smoke test for `Engine.get_speculative_stats` (passed)
- Focused MTP, speculative sampling, telemetry, and mixed-prefill tests (14/14 passed)

### Performance validation

Measured on one NVIDIA H200 with the CUDA Release build from stack head `dabb4103`, ORT CUDA 1.30.0, `qwen38_mtp_gdn_native_compact_final`, `K=7`, 1,024 generated tokens, CUDA graphs off, 2 warmups, and 3 measured runs:

| Prompt tokens | Decode tokens/s | Wall tokens/s | Evaluated-draft acceptance |
|---:|---:|---:|---:|
| 512 | 139.92 | 138.15 | 70.37% |
| 2,048 | 138.59 | 132.99 | 70.45% |
| 8,192 | 111.26 | 97.63 | 67.63% |

All measured runs produced deterministic output hashes and speculative counters. The compact model loaded without dense-checkpoint compatibility warnings.

## Dependencies

- Based on #2494.
- Does not include builder/export changes from Stack A.

## Checklist

- [x] Tests added/updated
- [x] Public C/C++/Python telemetry bindings validated

### DIFF
--- a/docs/paged_attention_engine.md
+++ b/docs/paged_attention_engine.md
@@ -363,12 +363,13 @@ borrowed. A Buffer event holds the Request internally for the lifetime of that v
 returned `const OgaRequest*` is never an independently owned handle and is invalidated with the
 event.
 
-A committed step emits a token event, a terminal event, or one combined
-`Token | TurnFinished` event. EOS can emit terminal completion without a visible token. Event
-delivery does not mutate the Request's retained logical sequence. One successful committed step
-produces at most one combined event per affected Request. If speculative verification accepts a
-visible draft immediately before an accepted EOS ends the Turn, the accepted visible token and
-terminal completion share that combined event; the EOS itself is never emitted as a token.
+A committed step emits zero or more token events and may emit a terminal event. The final visible
+token can carry `Token | TurnFinished`; EOS can emit terminal completion without a visible token.
+Speculative verification emits each accepted draft and its correction or bonus as a separate token
+event in sequence order. If verification accepts a visible draft immediately before an accepted EOS
+ends the Turn, the accepted visible token and terminal completion share that combined event; the EOS
+itself is never emitted as a token. Event delivery does not mutate the Request's retained logical
+sequence.
 
 A chunked partial-prefill transaction can commit cache and Request progress without producing an
 event. `Run()` then returns count zero while `HasPendingRequests()` remains true. It does not loop
@@ -386,12 +387,13 @@ This distinction is important:
 If the engine has previously encountered a fatal transaction or execution failure, `Run()` rethrows
 the stored error instead of attempting more work.
 
-## Caller-supplied speculative drafts
+## Speculative drafts
 
-The dynamic Engine can verify a caller-supplied continuation together with a request's next decode
-token. This is separate from the `Generator` API's draft-model and n-gram strategies: the Engine
-does not create a proposer. The caller proposes tokens for one request before the next step with
-`Request::SetDraftTokens` (`OgaRequestSetDraftTokens` in C or `Request.set_draft_tokens` in Python).
+The dynamic Engine can verify a continuation together with a request's next decode token. A caller
+can propose tokens before the next step with `Request::SetDraftTokens` (`OgaRequestSetDraftTokens`
+in C or `Request.set_draft_tokens` in Python). When `model.mtp` names an auxiliary draft head, the
+Engine also maintains an internal shadow Request and automatically proposes a chained greedy
+continuation after each committed target step.
 
 Query the internal `Engine::MaxDraftTokensPerStep` capability through
 `OgaEngineMaxDraftTokensPerProposal`, `OgaEngine::MaxDraftTokensPerProposal`, or
@@ -401,16 +403,18 @@ The reported value is a capability limit, not a guarantee that every proposed to
 step's global token budget or current cache capacity.
 
 The request must already belong to the Engine, have completed prefill, and be ready to decode.
-Verification currently supports only requests whose resolved search mode is greedy (`do_sample` is
-false, `top_k` is 1, or `temperature` is zero), with no guidance, repetition penalty,
-no-repeat-ngram processing, or active minimum-length processing. Passing an empty sequence clears a
-pending proposal.
+Verification supports greedy target selection and random target sampling with a positive `top_k`;
+proposals remain deterministic. Guidance,
+repetition penalty, no-repeat-ngram processing, and active minimum-length processing are not
+supported. Passing an empty sequence clears a pending proposal.
 
 For a decode with K scheduled drafts, the packed input is the request's one unprocessed token
 followed by the K drafts. The decoder must return K+1 logits rows for that request. Row `i` predicts
 draft `i`; verification accepts the longest prefix whose tokens equal the target argmax. The first
 nonmatching row supplies the replacement token, or row K supplies a bonus token when every draft is
-accepted. Accepted drafts and the replacement or bonus are published together, so one model run can
+accepted. Under random target sampling, each target row is sampled and the deterministic draft is
+accepted when it matches that sample; the first mismatch is the correction token. Accepted drafts
+and the correction or bonus are published together, so one model run can
 publish several ordered token events. If the caller's event buffer cannot hold all of them, `Run()`
 retains the overflow and drains it on subsequent calls before executing the model again. Callers
 must process all returned events before reusing the buffer.
@@ -544,7 +548,7 @@ The fixed sub-reservation, when present, admits the same rows in the same order.
 Physical execution ordering does not change event ordering. Each row retains its logical scheduler
 rank, and the Engine restores that order when it publishes the transaction's events after commit.
 
-The fixed `target_tokens` mirror the paged `target_cache_slots`, so both states commit at one token boundary. `StepPlan::fixed_state` records the fixed row count, new-slot count, and binding footprint; `Engine::StepDynamic()` then proves the reservation matches that plan exactly -- required flag, row count, new-slot count, bytes, and per-row request identity -- and fails fatally on any mismatch. A composite reservation wraps exactly one paged reservation and the Engine holds at most one at a time, so the paged split-commit contract (its constructor reserves its own `committed_tables_` headroom, and the pool permits a single live reservation) holds without any cross-reservation aggregate check.
+The fixed `target_tokens` mirror the paged `target_cache_slots`, so both states commit at one token boundary. `StepPlan::fixed_state` records the fixed row count, new-slot count, and binding footprint; `Engine::RunDynamic()` then proves the reservation matches that plan exactly -- required flag, row count, new-slot count, bytes, and per-row request identity -- and fails fatally on any mismatch. A composite reservation wraps exactly one paged reservation and the Engine holds at most one at a time, so the paged split-commit contract (its constructor reserves its own `committed_tables_` headroom, and the pool permits a single live reservation) holds without any cross-reservation aggregate check.
 
 While the reservation is active, decoder input preparation can view a combined block table containing:
 
--- a/docs/paged_attention_engine.md
+++ b/docs/paged_attention_engine.md
@@ -429,6 +429,46 @@ retryable rollback restores the request and leaves the proposal pending for retr
 the cache reservation is narrowed from the scheduled K+1 slots to the accepted prefix plus the
 request's mandatory token; paged KV and fixed recurrent state publish at that same boundary.
 
+### Engine-hosted MTP head: operational contract
+
+`model.mtp` turns the head on automatically for every request the dynamic Engine decodes. Server
+authors should size capacity and handle failures against the following behaviors.
+
+**Auxiliary memory accounting.** The head is a second paged pool that always holds the same block
+count as the target pool, so both are sized from one budget. With
+`engine.dynamic_batching.gpu_utilization_factor`, the auxiliary bytes per block are folded into the
+capacity computation. With an explicit `engine.dynamic_batching.num_blocks`, that value is the
+combined budget: the target pool is scaled down so target plus head together cost what the
+configured block count would have cost without a head. A `num_blocks` too small to leave at least
+one block for each pool is rejected at Engine construction.
+
+**Failure isolation.** The target decode is mandatory; MTP drafting is optional acceleration. A
+recoverable head failure (cache pressure, shape mismatch, binding or session error) releases only
+MTP state, and the target step still commits — without drafts for that step — and increments
+`standard_fallback_steps` in the speculative statistics. Only a contract violation, or a failure to
+roll MTP state back, marks the Engine unhealthy. A persistent head failure therefore degrades to
+ordinary decoding instead of stalling the request.
+
+**Shadow lifecycle.** The head's shadow Request mirrors only the suffix the target committed during
+the current turn. Beginning a continuation and canceling a turn both drop the shadow and release its
+auxiliary blocks, so the next drafted step rebuilds it from the new turn's suffix. Closing the
+request releases it as well.
+
+**Mixed prefill batches.** Drafts are proposed only for requests that committed a decode token in
+the step. A request that was prefilling, finished its turn, or has a proposal the Engine cannot
+verify is skipped for that step and picks drafting back up on its next decode step.
+
+**Seeded sampling.** `search.random_seed` reproduces output for a given decode path, not across
+decode paths. A verified drafted step draws its target tokens from the Request's own host random
+stream because acceptance is sequential, while an ordinary batched step draws from the device
+sampler state. Whether drafts are admitted depends on batch composition and cache pressure, so a
+seeded run is only bit-reproducible against another run scheduled the same way. Disable `model.mtp`
+when exact cross-scheduling reproducibility is required.
+
+**Telemetry thread ownership.** `Engine::GetSpeculativeStats` (`OgaEngineGetSpeculativeStats`,
+`Engine.get_speculative_stats`) reads counters that `Run()` mutates without synchronization. Like
+every other Engine query, it must be called from the Engine's owner thread and throws otherwise.
+
 ## One dynamic step in detail
 
 `Engine::RunDynamic()` coordinates the complete transaction.
--- a/src/engine/engine.cpp
+++ b/src/engine/engine.cpp
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.
 
 #include "engine.h"
+#include "../search.h"
 
 #include <limits>
 
@@ -40,6 +41,37 @@ std::string AddExceptionCause(std::string message, std::exception_ptr error) {
   return message;
 }
 
+std::vector<int32_t> GreedyTokens(
+    const std::shared_ptr<DecoderOnly_Model>& model,
+    std::vector<DeviceSpan<float>>& logits) {
+  std::vector<int32_t> tokens(logits.size());
+  auto device_tokens = model->p_device_->Allocate<int32_t>(logits.size());
+  (void)device_tokens.CpuSpan();
+  bool device_argmax = true;
+  for (size_t i = 0; i < logits.size(); ++i) {
+    device_argmax &= model->p_device_->ArgMaxDevice(
+        logits[i].Span().data(), Ort::TypeToTensorType<float>, 1,
+        model->config_->model.vocab_size, device_tokens.subspan(i, 1));
+  }
+  if (device_argmax) {
+    device_tokens.CopyDeviceToCpu();
+    std::copy(device_tokens.CpuSpan().begin(), device_tokens.CpuSpan().end(),
+              tokens.begin());
+    return tokens;
+  }
+
+  for (size_t i = 0; i < logits.size(); ++i) {
+    if (!model->p_device_->ArgMax(
+            logits[i].Span().data(), Ort::TypeToTensorType<float>, 1,
+            model->config_->model.vocab_size, &tokens[i])) {
+      const auto values = logits[i].CopyDeviceToCpu();
+      tokens[i] = static_cast<int32_t>(
+          std::max_element(values.begin(), values.end()) - values.begin());
+    }
+  }
+  return tokens;
+}
+
 }  // namespace
 
 Engine::Engine(std::shared_ptr<Model> model)
@@ -49,7 +81,10 @@ Engine::Engine(std::shared_ptr<Model> model, EngineDependencies dependencies)
     : model_{std::move(model)},
       cache_manager_{std::move(dependencies.cache_manager)},
       scheduler_{std::move(dependencies.scheduler)},
-      model_executor_{std::move(dependencies.model_executor)} {
+      model_executor_{std::move(dependencies.model_executor)},
+      mtp_model_{std::move(dependencies.mtp_model)},
+      mtp_cache_manager_{std::move(dependencies.mtp_cache_manager)},
+      mtp_model_executor_{std::move(dependencies.mtp_model_executor)} {
   // Fail fast on a missing collaborator rather than crashing later on first use.
   if (!cache_manager_) {
     throw std::runtime_error("Engine requires a non-null cache manager.");
@@ -74,6 +109,7 @@ Engine::Engine(std::shared_ptr<Model> model, EngineDependencies dependencies)
   staged_event_order_.reserve(max_batch_size);
   pending_events_.reserve(max_step_events);
   staged_events_.reserve(max_step_events);
+  mtp_requests_.reserve(max_batch_size);
 }
 
 Engine::~Engine() {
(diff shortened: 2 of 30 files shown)


src/engine/engine.cpp
```diff
@@ -85,10 +160,577 @@ Engine::~Engine() {
...
+    for (size_t i = 0; i < feeds.size(); ++i) {
+      previous_stage_rows[i] = i;
+      if (feeds[i].max_draft_tokens > 1) {
+        active_feed_indices.push_back(i);
+        feedback_rows.push_back(step->plan.requests[i].logits_row_index);
+      }
+    }
+    std::unique_ptr<Tensor> feedback_hidden;
+    if (!active_feed_indices.empty()) {
+      Tensor* head_hidden = mtp_requests.HiddenStates();
```
baijumeswani, line 498: The production MTP decoder reaches this check without a hidden-state output. `CreateMtpDecoderConfig` copies the head output name and then clears `projected->model.mtp`, while `VarlenDecoderIO::PrepareHiddenStates` binds that output only when `model.mtp.filename` is non-empty. With the default multi-token chain, `active_feed_indices` is non-empty, `HiddenStates()` is null, and every step throws here. The MTP tests do not catch this because the recording executor fabricates hidden states. **Recommendation:** bind the head output based on an explicit output-demand/capability flag (or at least the configured output plus session metadata), and add a test that runs the projected MTP model through the real `ModelExecutor`/`VarlenDecoderIO` path for K>=2.

src/engine/scheduled_requests.cpp
```diff
@@ -236,13 +342,43 @@ std::vector<DeviceSpan<float>> ScheduledRequests::SelectSampledRows(
...
+      if (accepted_count == draft_count &&
+          selected_tokens[i].size() < token_budget) {
+        const auto selection = BuildTargetSelection(
+            row + draft_count, verify_rows[row + draft_count],
+            requests_[i]->SearchOptions(), topk, sampling_scratch);
+        selected_tokens[i].push_back(
+            SampleTargetToken(selection, requests_[i]->rng_));
+      }
+      accepted_draft_counts[i] = accepted_count;
+      sampled_rows.push_back({});
```
baijumeswani, line 376: This empty `DeviceSpan` is later passed to `TryApplyBatchedGuidanceMasks`. If this sampled+drafted request is row 0 and another request in the same batch uses guidance, that function calls `logits.front().Span()` before checking the row size, which dereferences the empty span's null backing buffer. Guidance is forbidden only on the drafted request itself, so this mixed batch is valid and reachable. **Recommendation:** avoid using an empty span as an in-band sentinel, or validate all row sizes before taking any pointer. Please add a regression test with sampled+drafted row 0 and a separate guided row.

src/engine/engine.cpp
```diff
@@ -812,6 +1486,7 @@ void Engine::RunDynamic() {
             entry.unprocessed_token_count - entry.draft_token_count +
                 entry.request->AcceptedDraftTokenCount());
       }
+      mtp_step = PrepareMtpStep(step_plan_, step_results_, scheduled_requests);
```
baijumeswani: MTP is optional optimization work, but it runs inside the mandatory target transaction here. A persistent MTP cache, shape, binding, or session failure falls into the generic post-processing handler, rolls back the successful target pass, and returns `RetryableBatchAbort`; the next `Run()` repeats the same failure with no committed progress. **Recommendation:** isolate auxiliary preparation from target commit so a recoverable MTP failure restores/releases only MTP state and commits the target step without drafts. Validate permanent contract errors at construction, and add a persistent-failure test proving the request still advances instead of livelocking.

src/engine/scheduled_requests.cpp
```diff
@@ -236,13 +342,43 @@ std::vector<DeviceSpan<float>> ScheduledRequests::SelectSampledRows(
...
+      const auto drafts = requests_[i]->StagedDraftTokens();
+      const size_t token_budget = requests_[i]->RemainingTurnTokenBudget();
+      requests_[i]->RewindDraftsForTransaction(0);
+      size_t accepted_count = 0;
+      while (accepted_count < draft_count &&
+             selected_tokens[i].size() < token_budget) {
+        const auto selection = BuildTargetSelection(
+            row + accepted_count, verify_rows[row + accepted_count],
+            requests_[i]->SearchOptions(), topk, sampling_scratch);
+        const int32_t token = SampleTargetToken(selection, requests_[i]->rng_);
```
baijumeswani, line 361: Sampled drafted steps draw from this host `std::mt19937`, while ordinary non-drafted steps for the same request use an independently seeded device `BatchedSamplerState`. Whether drafts are admitted depends on batch composition and cache pressure, so a fixed `random_seed` can produce different output under different scheduling. **Recommendation:** drive speculative target draws through the same transactional sampler state used by normal steps. If that is intentionally unsupported, explicitly narrow the seed contract and add a solo-vs-mixed scheduling test.

src/engine/paged_key_value_cache.cpp
```diff
@@ -122,7 +123,8 @@ size_t ComputeNumBlocks(std::shared_ptr<Model> model,
       model->config_->model.decoder.num_key_value_heads,
       model->config_->model.decoder.head_size,
       full_layer_count,
-      Ort::SizeOf(dtype));
+      Ort::SizeOf(dtype),
```
baijumeswani, line 130: The auxiliary bytes are included only in this auto-sizing path. When `dynamic_batching.num_blocks` is explicit, `ComputeNumBlocks` returns before considering `auxiliary_bytes_per_block`; Engine construction then gives the MTP pool the same block count, silently allocating an additional head KV pool outside the configured sizing contract. **Recommendation:** define whether `num_blocks` is a combined or target-only budget and enforce it. Either partition/validate the combined allocation or reject explicit `num_blocks` with MTP, with a focused capacity test.

src/engine/engine.cpp
```diff
@@ -217,9 +859,8 @@ uint64_t Engine::BeginTurn(const std::shared_ptr<Request>& request,
   const bool turn_id_exhausted_before = request->turn_id_exhausted_;
   bool added_to_scheduler = false;
 
-  request->SaveStateForTransaction();
+  request->SaveStateForNewTurnTransaction();
```
baijumeswani: A continuation starts a fresh target turn here, but an existing MTP shadow is left untouched; cancellation also leaves it resident. The initial suffix-local MTP state is intentional, but across turns the shadow becomes a concatenation of generated tokens from multiple turns while omitting the intervening prompt. Target verification preserves output correctness, but draft acceptance and auxiliary residency degrade silently. **Recommendation:** reset/close the shadow when beginning a continuation and on cancellation, matching fresh-turn semantics, and cover both lifecycle paths in tests.

src/engine/engine.cpp
```diff
@@ -1114,4 +1810,18 @@ size_t Engine::MaxDraftTokensPerStep() const {
              : 0;
 }
 
+SpeculativeStats Engine::GetSpeculativeStats() const noexcept {
```
baijumeswani: This is the only public Engine query that skips `ValidateOwnerThread()`. The counters are mutated by `Run()` and are not atomic, so calling the C/C++/Python telemetry API from a monitoring thread creates a data race; `noexcept` also prevents following the normal Engine error contract. **Recommendation:** remove `noexcept`, validate the owner thread like `MaxDraftTokensPerStep()`, and add an off-thread C API test.

src/engine/engine.cpp
```diff
@@ -85,10 +160,577 @@ Engine::~Engine() {
...
+      step->drafts[i].reserve(feeds[i].max_draft_tokens);
+      if (!device_draft_chain) {
+        step->drafts[i].push_back(first_drafts[i]);
+      }
+      step->plan.requests[i].request->CommitAuxiliaryDecoderStep();
+    }
+
+    const auto copy_hidden_rows = [&](Tensor& source,
+                                      std::span<const size_t> source_rows) {
+      auto destination = std::make_unique<Tensor>(mtp_model_->p_device_inputs_, hidden_type);
```
baijumeswani, line 472: This allocates a new device tensor for feedback rows, and the surrounding path also creates packed hidden-state and per-stage request/tensor objects on every drafted step. For a latency-sensitive decode loop, allocation churn can erase part of the speculative gain even when acceptance is high. **Recommendation:** reuse capacity-sized scratch buffers owned by the Engine/MTP executor, and report allocation-free steady-state measurements across K=1/2/4/7 and B1/B4/B8 before treating the current throughput numbers as representative.

docs/paged_attention_engine.md
```diff
@@ -364,12 +364,13 @@ This distinction is important:
...
+## Speculative drafts
 
-The dynamic Engine can verify a caller-supplied continuation together with a request's next decode
-token. This is separate from the `Generator` API's draft-model and n-gram strategies: the Engine
-does not create a proposer. The caller proposes tokens for one request before the next step with
-`Request::SetDraftTokens` (`OgaRequestSetDraftTokens` in C or `Request.set_draft_tokens` in Python).
+The dynamic Engine can verify a continuation together with a request's next decode token. A caller
+can propose tokens before the next step with `Request::SetDraftTokens` (`OgaRequestSetDraftTokens`
+in C or `Request.set_draft_tokens` in Python). When `model.mtp` names an auxiliary draft head, the
+Engine also maintains an internal shadow Request and automatically proposes a chained greedy
```
baijumeswani, line 395: Automatic MTP activation is introduced here, but the operational contract is not yet documented: target-vs-auxiliary failure behavior, auxiliary memory accounting (especially explicit `num_blocks`), seeded sampling guarantees, shadow reset behavior across continuation/cancel, mixed-prefill draft demotion, and telemetry thread ownership. **Recommendation:** document these behaviors and the intended degradation policy in this section so server authors can configure capacity and handle failures predictably.

baijumeswani (commented):
This PR adds Engine-hosted chained MTP drafting on top of the existing speculative-verification path. It creates a paged auxiliary decoder and per-request shadow state, feeds target hidden states into the MTP head, verifies variable-length proposals in the target batch, commits only accepted prefixes, supports sampled verification, and exposes speculative telemetry. The target/auxiliary transaction structure and accepted-prefix accounting are thoughtful. However, the production K>=2 path is currently blocked by an output-binding mismatch, and the remaining inline findings cover mixed-batch safety, failure/degradation behavior, reproducibility, lifecycle, memory accounting, threading, performance, and documentation. I recommend addressing the correctness and forward-progress items before approval, then qualifying the exact PR head through production decoder IO rather than only test doubles.

src/engine/engine.cpp
tianleiwu: Fixed in a46473e. The gate no longer derives from `model.mtp.filename`. `Config::Engine` now carries an explicit `hidden_states_output_required` capability flag:

- `CreateMtpDecoderConfig` sets it on the projected head **before** clearing `projected->model.mtp`, so the head keeps emitting the hidden states the next chain stage consumes.
- `Engine::CreateDependencies` sets it on the target decoder when a head is configured.
- `VarlenDecoderIO::PrepareHiddenStates` and `VarlenGraphBuffers` both read only that flag, so a model that merely exports hidden states still does not pay for the output.

Covered by `MtpDecoderConfigTest.ProjectsPagedDecoderWithoutMainFixedState`, which now asserts the projected head demands the output. The `ModelExecutorHiddenStatesTest` cases exercise the real `ModelExecutor`/`VarlenDecoderIO` binding path and caught the missing flag while I was wiring this up.

src/engine/scheduled_requests.cpp
tianleiwu: Fixed in a46473e. The row-size check now runs over every row **before** any pointer is taken, extracted into `PackedLogitsRowBase()`:

```cpp
for (const auto& row : logits) {
  if (row.size() != vocab_size) return nullptr;   // empty row has no backing buffer
}
float* const first_row = logits.front().Span().data();
```

A mixed batch therefore falls back to per-request guidance instead of dereferencing the empty row's null `DeviceBuffer`. Regression test: `GuidanceProcessorTest.PackedLogitsRejectDraftVerifiedSampledRow` covers an empty row 0 alongside a valid row, plus the non-contiguous case.

src/engine/engine.cpp
tianleiwu: Fixed in a46473e. `PrepareMtpStep` is now wrapped in its own handler inside the transaction:

```cpp
try {
  mtp_step = PrepareMtpStep(step_plan_, step_results_, scheduled_requests);
} catch (const MtpRollbackError&) {
  throw;                       // MTP state is untrustworthy -> stays fatal
} catch (const std::logic_error&) {
  throw;                       // contract violation -> stays fatal
} catch (...) {
  mtp_step.reset();            // recoverable: commit the target step without drafts
  ++speculative_stats_.standard_fallback_steps;
}
```

`PrepareMtpStep` already restores every checkpointed shadow, releases the reservation, and erases newly created shadows before rethrowing anything other than `MtpRollbackError`, so the target pass commits with no MTP residue. Two tests cover this: `MtpPlanningFailureCommitsTargetStepWithoutDrafts` (one failure, target commits, next step drafts again) and `PersistentMtpFailureKeepsTheRequestAdvancing` (failure that never clears; the sequence length grows every step instead of livelocking).

src/engine/scheduled_requests.cpp
tianleiwu: Narrowed the contract explicitly rather than rewiring the sampler in this PR. Draft acceptance is sequential — row `j` is only drawn when row `j-1`'s draw matched its draft — which the batched device sampler's single fused call cannot express today. Routing it through `BatchedSamplerState` needs a new sequential-draw entry point on `BatchedSampler` and is a larger change than this PR should carry.

What changed in a46473e:

- A comment at this call site states that these draws come from the Request's own host stream (checkpointed with the transaction, so rollback is still exact) rather than its device sampler state.
- `docs/paged_attention_engine.md` gains a **Seeded sampling** paragraph: `search.random_seed` reproduces output for a given decode path, not across decode paths, because draft admission depends on batch composition and cache pressure. It also tells callers to disable `model.mtp` when exact cross-scheduling reproducibility is required.

Happy to file a follow-up for the unified sampler state if you'd prefer that tracked.

src/engine/paged_key_value_cache.cpp
tianleiwu: Fixed in a46473e. `num_blocks` is now defined as the **combined** budget, matching what the auto-sizing path already does (it divides by `primary + auxiliary` and gives both pools the same count). The explicit path goes through a new `ResolveConfiguredPagedBlockCount()`:

- No head: the configured value is used verbatim.
- With a head: the target pool is scaled to `configured * primary / (primary + auxiliary)`, so target plus head cost what the configured count would have cost on its own.
- A count too small to leave a block for each pool, or one that would overflow the byte-count math, is rejected at Engine construction.

Test: `PagedKeyValueCacheManifestTest.ExplicitBlockCountCoversBothPools`. Documented under **Auxiliary memory accounting** in `docs/paged_attention_engine.md`.

src/engine/engine.cpp
tianleiwu: Fixed in a46473e. `Engine::BeginTurn` now calls `CloseMtpRequest(request)` before admitting the turn, and `Engine::CancelRequest` calls it right after `CompleteCancelFromEngine`. Both release the shadow's auxiliary blocks, so the next drafted step rebuilds a suffix-local shadow, matching fresh-turn semantics.

Doing it before admission (rather than after commit) keeps a rolled back turn consistent: the shadow is simply rebuilt on the next drafted step, and no post-commit failure path can leave the Engine half-updated.

Tests: `ContinuationDropsTheMtpShadowFromThePreviousTurn` and `CancelDropsTheMtpShadow`.

src/engine/engine.cpp
tianleiwu: Fixed in a46473e. `Engine::GetSpeculativeStats` no longer declares `noexcept` and now calls `ValidateOwnerThread()` first, exactly like `MaxDraftTokensPerStep()`, so it follows the normal Engine error contract. Test: `SpeculativeStatsRejectOffOwnerThreadReads` calls it from a non-owner thread and expects the throw. The C API wrapper already routes through `OGA_TRY`/`OGA_CATCH`, so the throw surfaces as an `OgaResult` error. Also documented under **Telemetry thread ownership**.

src/engine/engine.cpp
tianleiwu: Agreed on the direction; not addressed in this PR. The device chain already reuses Engine-owned scratch for the parts that scale with K — `mtp_device_drafts_` and `mtp_device_chain_inputs_` are allocated once at capacity and re-subspanned per step. What remains per step is the packed hidden-state tensor, the per-stage feedback slices, and the per-stage `ScheduledRequests`.

Those are `Tensor`/`OrtValue` objects whose shapes change with the active row count each step, so pooling them means either fixed-capacity buffers with per-step views bound to ORT or a shape-keyed cache; both are a meaningful redesign of the chain loop rather than a review fix, and I would rather land it with the allocation-free steady-state measurements across K=1/2/4/7 and B1/B4/B8 that you asked for than bolt it on here. I will open a follow-up issue with that scope and treat the current throughput numbers as preliminary until then.

docs/paged_attention_engine.md
tianleiwu: Fixed in a46473e. `docs/paged_attention_engine.md` gains an **Engine-hosted MTP head: operational contract** section covering each point you listed:

- **Auxiliary memory accounting** — how the head shares one budget with the target under `gpu_utilization_factor`, and that an explicit `num_blocks` is the combined budget (rejected when too small for both pools).
- **Failure isolation** — target decode is mandatory, MTP is optional; a recoverable head failure commits the target step without drafts and increments `standard_fallback_steps`, and only contract violations or failed MTP rollback mark the Engine unhealthy.
- **Shadow lifecycle** — continuation, cancel, and close all drop the shadow and release its auxiliary blocks.
- **Mixed prefill batches** — drafts are proposed only for requests that committed a decode token in that step.
- **Seeded sampling** — `random_seed` reproduces a decode path, not output across decode paths; disable `model.mtp` when exact cross-scheduling reproducibility is required.
- **Telemetry thread ownership** — `GetSpeculativeStats` must be called from the Engine owner thread and throws otherwise.

src/engine/engine.cpp
```diff
@@ -219,6 +866,9 @@ bool Engine::CancelRequest(const std::shared_ptr<Request>& request, uint64_t tur
 
   const auto counters =
       request->CompleteCancelFromEngine(*this, turn_id);
+  // The canceled turn's suffix is gone, so the shadow that mirrored it must not survive into the
+  // next turn.
+  CloseMtpRequest(request);
```
baijumeswani, line 871: The request has already been marked canceled before this cleanup runs. `CloseMtpRequest` can still throw while creating its temporary list or checking/releasing cache state. If that happens, cancellation is only half finished: the request is canceled, but no final cancellation event is added. **Recommendation:** run MTP cleanup before changing the request to canceled, or make this cleanup guaranteed not to throw after the state change. Please add a test that forces cleanup to fail and verifies cancellation stays all-or-nothing.

src/engine/engine.cpp
```diff
@@ -721,6 +1405,21 @@ void Engine::RunDynamic() {
...
+      // committed target pass back, or a persistent failure would repeat forever with no progress.
+      // PrepareMtpStep restores every piece of MTP state before it rethrows, so the target step can
+      // commit without drafts. Contract violations and failed MTP rollback stay fatal below.
+      try {
+        mtp_step = PrepareMtpStep(step_plan_, step_results_, scheduled_requests);
+      } catch (const MtpRollbackError&) {
+        throw;
+      } catch (const std::logic_error&) {
+        throw;
+      } catch (...) {
```
baijumeswani, line 1419: It is good that an MTP failure now falls back to normal decoding, but this catch drops the reason completely. A permanently broken head can fail again on every token, adding cost while the operator sees only a growing fallback counter. **Recommendation:** log the first failure (then rate-limit repeats), expose a specific MTP failure count or last reason, and disable MTP after repeated permanent failures while continuing normal decoding.

src/engine/engine.cpp
```diff
@@ -85,10 +160,576 @@ Engine::~Engine() {
...
+  size_t mtp_bytes_per_block = 0;
+  if (!model->config_->model.mtp.filename.empty()) {
+    if (!model->config_->engine.dynamic_batching) {
+      throw std::runtime_error("An Engine-hosted MTP head requires dynamic batching.");
+    }
+    mtp_model = std::make_shared<DecoderOnly_Model>(
+        CreateMtpDecoderConfig(*model->config_), GetOrtEnv());
+    mtp_bytes_per_block = PagedKeyValueCacheBytesPerBlock(mtp_model);
+    // The head consumes the target decoder's packed hidden states, so the target must emit them.
+    model->config_->engine.hidden_states_output_required = true;
```
baijumeswani, line 173: This requests hidden states from the main model, but Engine creation does not first confirm that the configured output exists and matches `model.mtp.main_hidden_states`. A bad package can therefore load successfully and fail the whole Engine on its first decode. **Recommendation:** during Engine creation, verify the output name exists and has the expected type and width, and reject the model with a clear message before any request runs.

src/engine/engine.cpp
```diff
@@ -85,10 +160,576 @@ Engine::~Engine() {
...
+      step->drafts[i].reserve(feeds[i].max_draft_tokens);
+      if (!device_draft_chain) {
+        step->drafts[i].push_back(first_drafts[i]);
+      }
+      step->plan.requests[i].request->CommitAuxiliaryDecoderStep();
+    }
+
+    const auto copy_hidden_rows = [&](Tensor& source,
+                                      std::span<const size_t> source_rows) {
+      auto destination = std::make_unique<Tensor>(mtp_model_->p_device_inputs_, hidden_type);
```
baijumeswani, line 472: This creates a new device tensor for each feedback stage. The same path also creates packed hidden-state storage and stage-local request objects on every drafted step. Those allocations sit directly in the token loop and may reduce the speedup, especially at batch size 1. **Recommendation:** keep reusable scratch buffers sized for the largest supported batch and draft count, exercise the CUDA chaining test path, and rerun performance on this exact head for K=1/2/4/7 and B1/B4/B8.

docs/paged_attention_engine.md
```diff
@@ -425,6 +429,46 @@ retryable rollback restores the request and leaves the proposal pending for retr
...
+**Shadow lifecycle.** The head's shadow Request mirrors only the suffix the target committed during
+the current turn. Beginning a continuation and canceling a turn both drop the shadow and release its
+auxiliary blocks, so the next drafted step rebuilds it from the new turn's suffix. Closing the
+request releases it as well.
+
+**Mixed prefill batches.** Drafts are proposed only for requests that committed a decode token in
+the step. A request that was prefilling, finished its turn, or has a proposal the Engine cannot
+verify is skipped for that step and picks drafting back up on its next decode step.
+
+**Seeded sampling.** `search.random_seed` reproduces output for a given decode path, not across
```
baijumeswani, line 461: This warning explains the seed behavior for automatic `model.mtp`, but the same behavior also applies when a caller supplies sampled drafts through `Request::SetDraftTokens` / `OgaRequestSetDraftTokens`. Disabling `model.mtp` does not help that case. **Recommendation:** add the same scheduling-dependent seed warning to the public draft-proposal API docs and say that exact repeatability requires the same draft and scheduling path.

baijumeswani (commented):
This PR adds automatic MTP drafting to the batch Engine. After each normal model step, it runs the MTP head to suggest several next tokens, checks those suggestions with the main model, keeps only the accepted prefix, and tracks the result in the existing request and cache transaction. It also adds sampled verification and public statistics.

The latest update addresses the major issues from the previous review: chained hidden-state output is now connected, the mixed guidance crash is fixed, an MTP failure no longer prevents the main request from moving forward, explicit cache sizing includes the MTP head, shadow state is reset between turns, and statistics follow the Engine thread rule. The added tests and documentation are much better.

I still see a small cancellation correctness issue and a few operational and performance gaps in the inline comments below. I recommend fixing the cancellation ordering and validating the hidden-state contract when the Engine is created before approval. The failed CUDA integration jobs should also be rerun or explained. They do not appear to exercise this Engine path, but the PR should be green. Finally, the published speed numbers came from a different stack commit, so an exact-head run with the real MTP model would make the performance case much stronger.

baijumeswani (approved):

src/engine/engine.cpp
tianleiwu: Addressed in #2496 (`53c7a88c`). `CloseMtpRequest` now runs before `CompleteCancelFromEngine`, so a cleanup exception leaves request state unchanged. `EngineRunTest.MtpCleanupFailureLeavesCancellationRetryable` injects a deallocation failure, verifies that no terminal event/state transition occurs, and verifies that a second cancellation succeeds.

src/engine/engine.cpp
tianleiwu: Addressed in #2496 (`53c7a88c`). Recoverable proposal failures now increment dedicated `mtp_failures` telemetry, preserve and log the first failure reason, and disable Engine-hosted MTP after three consecutive failures while ordinary decoding continues. A successful MTP preparation resets the consecutive count. The transient and persistent failure tests verify fallback, continued request progress, telemetry, and that planning stops after the third failure.

src/engine/engine.cpp
tianleiwu: Addressed in #2496 (`53c7a88c`). Engine creation now binds the configured `model.mtp.main_hidden_states` output and validates both target output and MTP-head input names, rank-2 shapes, configured hidden width, and matching element types before serving requests. `MtpDecoderConfigTest.ValidatesMainAndHeadHiddenStateContract` covers the valid contract and each mismatch.

src/engine/engine.cpp
tianleiwu: Tracked as #2517. This needs a distinct allocation/performance change rather than a local correctness patch. The issue scopes reusable packed-hidden/feedback scratch, removal or pooling of avoidable stage-local allocations, CUDA chaining coverage, allocation counts, and the requested K=1/2/4/7 by B1/B4/B8 performance matrix.

docs/paged_attention_engine.md
tianleiwu: Addressed in #2496 (`53c7a88c`). The scheduling-dependent sampling warning is now present on the public C++ `Request::SetDraftTokens` API, `OgaRequestSetDraftTokens`, the Python binding docstring, and the engine guide. The wording states that exact seeded repeatability requires both the same supplied drafts and the same scheduling path.

PR #2496, opened 2026-08-28
Run a DFlash2 block drafter in the batch engine

tianleiwu:
## Description

Adds the DFlash2 block-drafter runtime to the paged batch engine on top of #2495. The target decoder exposes packed auxiliary hidden states, and the drafter ingests committed rows into its own bounded sliding-window KV cache before proposing a coherent token path through the exported candidate lattice.

This replaces #2460 while excluding builder changes already covered by #2487/#2489, diagnostics churn, unsafe mixed-checkpoint experiments, and ONNX Runtime dependency changes.

### Packed target output

- Binds an optional packed `aux_hidden_states` decoder output sized to the active token rows.
- Forwards the output through hybrid decoder IO and scheduled requests without capturing dense recurrent checkpoints.
- Rejects CUDA graph capture for this dynamically sized intermediate output.

### DFlash2 drafter

- Parses runtime-only `model.dflash2` session, geometry, tensor-name, and sliding-window configuration.
- Creates the DFlash2 session before sizing the target cache so its fixed batch-scaled footprint is included in available-memory accounting.
- Maintains a bounded per-request paged KV ring and releases its blocks with request teardown.
- Packs committed context and query-block rows into one drafter execution and walks the candidate/edge-score lattice greedily.

### Engine integration

- Drops hidden-state rows for rejected target drafts before feeding the drafter.
- Proposes only for eligible greedy requests within per-request and maximum-length draft limits.
- Captures feed metadata before request commit clears accepted-prefix counters, then publishes drafts after the target/MTP transaction commits.
- Records DFlash2 forward passes in the shared speculative telemetry.

## Testing

- `build/Linux/Debug/unit_tests` (238 passed, 22 skipped)
- `build/Linux/Debug/engine_unit_tests` (295/295 passed)
- `cmake --build build/Linux/Debug` (passed)
- Added 6 focused tests for DFlash2 geometry validation, config projection, exact KV bytes, bounded ring sizing, and rejection of unbounded cache pools.

### Performance validation

Measured on one NVIDIA H200 with the CUDA Release build from stack head `dabb4103`, ORT CUDA 1.30.0, compact-final native-GDN artifacts, `K=7`, 1,024 generated tokens, CUDA graphs off, 2 warmups, and 3 measured runs:

| Prompt tokens | MTP decode tokens/s | DFlash2 decode tokens/s | DFlash2/MTP |
|---:|---:|---:|---:|
| 512 | 139.92 | 166.40 | 1.189x |
| 2,048 | 138.59 | 155.92 | 1.125x |
| 8,192 | 111.26 | 129.43 | 1.163x |

DFlash2's geometric-mean decode speedup was **1.159x (+15.89%)**. Its evaluated-draft acceptance was 75.13%, 74.60%, and 72.86% respectively. All measured runs produced deterministic output hashes and speculative counters, with no dense-checkpoint compatibility warnings.

## Dependencies

- Based on #2495.
- Uses auxiliary hidden-state exporter support from #2487/#2489 when producing models; builder code is not duplicated here.

## Checklist

- [x] Focused unit coverage added
- [x] Full C++ and engine suites pass
- [x] Real-model DFlash2 proposal and throughput validation

### DIFF
--- a/docs/paged_attention_engine.md
+++ b/docs/paged_attention_engine.md
@@ -399,14 +399,17 @@ The dynamic Engine can verify a continuation together with a request's next deco
 can propose tokens before the next step with `Request::SetDraftTokens` (`OgaRequestSetDraftTokens`
 in C or `Request.set_draft_tokens` in Python). When `model.mtp` names an auxiliary draft head, the
 Engine also maintains an internal shadow Request and automatically proposes a chained greedy
-continuation after each committed target step.
+continuation after each committed target step. Alternatively, `model.dflash2` runs a block drafter
+over the target's packed auxiliary hidden states. A model cannot configure both automatic drafters.
 
 Query the internal `Engine::MaxDraftTokensPerStep` capability through
 `OgaEngineMaxDraftTokensPerProposal`, `OgaEngine::MaxDraftTokensPerProposal`, or
 `Engine.max_draft_tokens_per_proposal` before proposing work. Zero means that the decoder does not
 return one logits row per packed token or that its cache/state cannot commit an accepted prefix.
 The reported value is a capability limit, not a guarantee that every proposed token fits the next
-step's global token budget or current cache capacity.
+step's global token budget or current cache capacity. Automatic drafters also leave room for the
+target's correction token and cap each proposal by the Request's session limit and remaining Turn
+token budget.
 
 The request must already belong to the Engine, have completed prefill, and be ready to decode.
 Verification supports greedy target selection and random target sampling with a positive `top_k`;
@@ -1113,6 +1116,10 @@ variable-size output does not have persistent graph buffers. Engine construction
 rank, element type, and static width against the drafter input before allocating cache resources.
 The drafter run is synchronous because its packed inputs and outputs are owned by one proposal
 call, so `model.dflash2.run_options` cannot disable execution-provider synchronization.
+The direct drafter session also uses graph id `-1`: its variable-shaped proposal tensors are
+allocated per call and therefore cannot be captured safely. If this optional post-commit drafter
+run fails, the Engine discards any partial proposal, releases the drafter, and still publishes the
+already committed target events; subsequent Turns continue without automatic DFlash 2 proposals.
 
 ## Backpressure and fairness
 
--- a/docs/paged_attention_engine.md
+++ b/docs/paged_attention_engine.md
@@ -1123,6 +1123,11 @@ allocated per call and therefore cannot be captured safely. If this optional pos
 run fails, the Engine discards any partial proposal, releases the drafter, and still publishes the
 already committed target events; subsequent Turns continue without automatic DFlash 2 proposals.
 
+The drafter keeps a fixed sliding-window ring per request from that request's first decode step
+until it is closed, so its pool is sized for `max_batch_size` rings. A request that first decodes
+while every ring is taken is skipped for the rest of its life and decodes without DFlash 2 drafts;
+the drafter keeps serving the requests that already hold a ring.
+
 ## Backpressure and fairness
 
 Continuous batching does not mean every pending request runs on every step.
--- a/docs/paged_attention_engine.md
+++ b/docs/paged_attention_engine.md
@@ -1107,6 +1107,13 @@ Graph buffers are allocated once at configured limits and reshaped as static vie
 
 Prefill and mixed-token steps use graph id `-1`, which tells the CUDA execution provider to run eagerly.
 
+An Engine-hosted DFlash 2 drafter also forces eager target execution. Its target decoder output is
+the packed auxiliary hidden-state tensor named by `model.dflash2.main_aux_hidden_states`; that
+variable-size output does not have persistent graph buffers. Engine construction validates its
+rank, element type, and static width against the drafter input before allocating cache resources.
+The drafter run is synchronous because its packed inputs and outputs are owned by one proposal
+call, so `model.dflash2.run_options` cannot disable execution-provider synchronization.
+
 ## Backpressure and fairness
 
 Continuous batching does not mean every pending request runs on every step.
--- a/docs/paged_attention_engine.md
+++ b/docs/paged_attention_engine.md
@@ -454,9 +454,10 @@ one block for each pool is rejected at Engine construction.
 **Failure isolation.** The target decode is mandatory; MTP drafting is optional acceleration. A
 recoverable head failure (cache pressure, shape mismatch, binding or session error) releases only
 MTP state, and the target step still commits — without drafts for that step — and increments
-`standard_fallback_steps` in the speculative statistics. Only a contract violation, or a failure to
-roll MTP state back, marks the Engine unhealthy. A persistent head failure therefore degrades to
-ordinary decoding instead of stalling the request.
+`standard_fallback_steps` and `mtp_failures` in the speculative statistics. The first failure is
+logged, and three consecutive failures disable automatic MTP drafting for that Engine. Only a
+contract violation, or a failure to roll MTP state back, marks the Engine unhealthy. A persistent
+head failure therefore degrades to ordinary decoding without repeatedly running a broken head.
 
 **Shadow lifecycle.** The head's shadow Request mirrors only the suffix the target committed during
 the current turn. Beginning a continuation and canceling a turn both drop the shadow and release its
@@ -471,8 +472,9 @@ verify is skipped for that step and picks drafting back up on its next decode st
 decode paths. A verified drafted step draws its target tokens from the Request's own host random
 stream because acceptance is sequential, while an ordinary batched step draws from the device
 sampler state. Whether drafts are admitted depends on batch composition and cache pressure, so a
-seeded run is only bit-reproducible against another run scheduled the same way. Disable `model.mtp`
-when exact cross-scheduling reproducibility is required.
+seeded run is only bit-reproducible against another run with the same supplied drafts and scheduling
+path. Disabling `model.mtp` does not make caller-provided drafts scheduling-independent; exact
+repeatability requires replaying the same proposals and schedule.
 
 **Telemetry thread ownership.** `Engine::GetSpeculativeStats` (`OgaEngineGetSpeculativeStats`,
 `Engine.get_speculative_stats`) reads counters that `Run()` mutates without synchronization. Like
--- a/docs/paged_attention_engine.md
+++ b/docs/paged_attention_engine.md
@@ -1112,21 +1112,26 @@ Graph buffers are allocated once at configured limits and reshaped as static vie
 
 Prefill and mixed-token steps use graph id `-1`, which tells the CUDA execution provider to run eagerly.
 
-An Engine-hosted DFlash 2 drafter also forces eager target execution. Its target decoder output is
-the packed auxiliary hidden-state tensor named by `model.dflash2.main_aux_hidden_states`; that
-variable-size output does not have persistent graph buffers. Engine construction validates its
-rank, element type, and static width against the drafter input before allocating cache resources.
+An Engine-hosted DFlash 2 drafter consumes the packed auxiliary hidden-state tensor named by
+`model.dflash2.main_aux_hidden_states`. Single-token target decode steps bind that output through
+persistent graph buffers and remain eligible for CUDA graph capture; prefill and draft-verification
+steps remain eager. Engine construction validates the output's rank, element type, and static width
+against the drafter input before allocating cache resources.
 The drafter run is synchronous because its packed inputs and outputs are owned by one proposal
 call, so `model.dflash2.run_options` cannot disable execution-provider synchronization.
 The direct drafter session also uses graph id `-1`: its variable-shaped proposal tensors are
 allocated per call and therefore cannot be captured safely. If this optional post-commit drafter
-run fails, the Engine discards any partial proposal, releases the drafter, and still publishes the
-already committed target events; subsequent Turns continue without automatic DFlash 2 proposals.
+run fails, the Engine discards any partial proposal and still publishes the already committed target
+events. A transient failure is retried on the next step; three consecutive failures disable DFlash 2
+for the Engine. `dflash2_failures` and `dflash2_disables` report those events.
 
 The drafter keeps a fixed sliding-window ring per request from that request's first decode step
 until it is closed, so its pool is sized for `max_batch_size` rings. A request that first decodes
 while every ring is taken is skipped for the rest of its life and decodes without DFlash 2 drafts;
-the drafter keeps serving the requests that already hold a ring.
+the drafter keeps serving the requests that already hold a ring. This makes `max_batch_size` the
+DFlash 2 service-capacity limit across both active and idle long-lived requests, not merely the
+per-step scheduler limit. `dflash2_admission_misses` reports requests denied a ring because that
+capacity was occupied.
 
 ## Backpressure and fairness
 
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -543,6 +543,8 @@ struct DecoderOutputs_Element : JSON::Element {
       v_.state_update_recurrent_capsule_names = JSON::Get<std::string_view>(value);
     } else if (name == "hidden_states") {
       v_.hidden_states = JSON::Get<std::string_view>(value);
+    } else if (name == "aux_hidden_states") {
+      v_.aux_hidden_states = JSON::Get<std::string_view>(value);
     } else if (name == "outputs") {
       v_.outputs = JSON::Get<std::string_view>(value);
     } else if (name == "lstm_hidden_state") {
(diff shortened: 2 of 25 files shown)


src/engine/engine.cpp
```diff
@@ -240,14 +313,90 @@ EngineDependencies Engine::CreateDependencies(std::shared_ptr<Model> model) {
...
+    const size_t remaining_turn_tokens = entry.request->RemainingTurnTokenBudget();
+    const size_t remaining_turn_tokens_after_step =
+        results[i].visible_token_count < remaining_turn_tokens
+            ? remaining_turn_tokens - results[i].visible_token_count
+            : 0;
+    const size_t width = Dflash2DraftWidth(
+        max_drafts, static_cast<size_t>(entry.request->SpeculativeOptions().max_draft_tokens),
+        static_cast<size_t>(length_after_step), sequence_limit,
+        remaining_turn_tokens_after_step);
+    feed.wants_drafts = width > 0 && results[i].token_appended && !results[i].done && greedy &&
```
baijumeswani: This repeats several draft checks, but it misses guidance. A normal guided request can be marked as wanting drafts; `SetDraftTokens` then rejects it, and the error path permanently turns off DFlash2 for every request in the Engine. **Recommendation:** use `Request::DraftTokenValidationError()` here as the shared check, while keeping the DFlash2-specific greedy requirement. Add a test with one guided request and one ordinary request to confirm guidance does not disable drafting for the other request.

src/engine/engine.cpp
```diff
@@ -1595,13 +1766,42 @@ void Engine::RunDynamic() {
         CommitMtpStep(*mtp_step);
       }
       RecordSpeculativeCommit(step_plan_);
+      if (dflash2_drafter_ && MaxDraftTokensPerStep() > 0) {
+        // Reads the accepted-draft counts, which CommitStep clears below.
+        PrepareDflash2Feeds(step_plan_, step_results_);
```
baijumeswani: This runs after search and cache state have started committing, but it can still allocate while growing the feed vectors and can throw. If that happens, optional DFlash2 bookkeeping makes the whole Engine unhealthy in the middle of a completed target step. **Recommendation:** build these feeds before entering the final commit section, preallocate the vectors to `max_batch_size`, or wrap this call in the same safe fallback used for `PublishDflash2Drafts`. Add a fault-injection test for this point.

src/engine/engine.cpp
```diff
@@ -1595,13 +1766,42 @@ void Engine::RunDynamic() {
...
+        try {
+          PublishDflash2Drafts(scheduled_requests);
+        } catch (...) {
+          const auto dflash2_error = std::current_exception();
+          for (const auto& feed : dflash2_feeds_) {
+            if (feed.request) {
+              feed.request->SetDraftTokens({});
+            }
+          }
+          dflash2_drafter_.reset();
```
baijumeswani: One proposal error disables DFlash2 permanently, whether it was a temporary allocation failure or a real model-contract error. The only signal is a warning, so a server cannot reliably detect or count the loss of drafting. **Recommendation:** handle contract errors separately, use a small retry limit for temporary failures, and expose a `drafter_failures` counter plus a disable counter or reason. This can follow the MTP failure policy already used above.

src/dflash2_drafter.cpp
```diff
@@ -0,0 +1,694 @@
...
+        CheckedMetadataValue(layout.q_row_map.size(), "DFlash 2 cumulative sequence length"));
+    layout.past_sequence_lengths.push_back(
+        CheckedMetadataValue(first_position, "DFlash 2 past sequence length"));
+    const int32_t query_length = CheckedMetadataValue(query_len, "DFlash 2 query length");
+    const int32_t kv_length = CheckedMetadataValue(total_positions, "DFlash 2 KV length");
+    layout.max_query_len = std::max(layout.max_query_len, query_length);
+    layout.max_kv_len = std::max(layout.max_kv_len, kv_length);
+    layout.min_kv_len = std::min(layout.min_kv_len, kv_length);
+
+    state.cached_positions = CheckedAdd(
```
baijumeswani: This marks the new context as cached before the tensors are created and before the model run succeeds. Today the caller destroys the whole drafter after an error, which hides the inconsistent state, but a later retry policy could reuse cache state that was never written. **Recommendation:** keep the new cached positions in local temporary storage and apply them only after `session_->Run()` succeeds. Do the same for newly claimed rings, or clearly document that any exception makes the drafter unusable.

src/engine/engine.cpp
```diff
@@ -960,6 +1110,9 @@ void Engine::CloseRequest(const std::shared_ptr<Request>& request) {
       cache_manager_->IsResident(request);
 
   CloseMtpRequest(request);
+  if (dflash2_drafter_) {
+    dflash2_drafter_->Release(request.get());
```
baijumeswani, line 1115: Normal close releases the DFlash2 ring here, but the emergency `DetachRequestForTeardown()` path does not. If normal close fails during continuation recovery, the request can be removed while the drafter still keeps its raw pointer and ring. **Recommendation:** also call `Release` from the no-throw detach path, before the request is fully detached, and add a cleanup-failure test that checks the ring becomes available again.

src/engine/decoders/simple_decoder.cpp
```diff
@@ -38,7 +38,8 @@ SimpleDecoder::SimpleDecoder(std::shared_ptr<DecoderOnly_Model> model,
   if (IsGraphCaptureEnabled(model_->config_->model.decoder.session_options) &&
       cache_manager_->SupportsDynamicBatching() &&
       !has_fixed_state_groups_ &&
-      !has_position_ids) {
+      !has_position_ids &&
+      model_->config_->model.decoder.outputs.aux_hidden_states.empty()) {
```
baijumeswani: A configured auxiliary output disables target CUDA graph capture for the lifetime of the Engine. That cost remains even after the drafter turns itself off, and it also affects a model that exports this output without configuring DFlash2. **Recommendation:** add a runtime `aux_hidden_states_output_required` flag, similar to the MTP hidden-state flag, and use a reusable graph buffer for single-token decode shapes. At minimum, avoid disabling capture when no drafter consumes the output.

src/dflash2_drafter.cpp
```diff
@@ -0,0 +1,694 @@
...
+    std::fill(cpu.begin(), cpu.end(), int32_t{-1});
+    for (size_t row = 0; row < served.size(); ++row) {
+      const auto& blocks = requests_[feeds[served[row]].request].blocks;
+      if (ring_blocks_ == 0) {
+        std::copy(blocks.begin(), blocks.end(), cpu.begin() + row * max_blocks);
+        continue;
+      }
+      // A windowed drafter repeats its ring across every column: column j holds the block that
+      // owns position j * block_size, which is ring[j % ring_blocks].
+      for (size_t column = 0; column < max_blocks; ++column) {
```
baijumeswani, line 618: This loop grows with the request's full context length even though the cache itself is a fixed-size sliding-window ring. At long context, every token rebuilds and copies a larger table containing repeated ring IDs. **Recommendation:** use a window-relative block table and rebased positions so the number of columns is limited by the ring size rather than total context length. Add a long-context test that asserts table width stays bounded.

src/dflash2_drafter.cpp
```diff
@@ -0,0 +1,694 @@
...
+    tensor->CreateTensor(shape);
+    return tensor;
+  };
+  auto fill_int32 = [](Tensor& tensor, const std::vector<int32_t>& values) {
+    auto span = tensor.GetDeviceSpan<int32_t>();
+    std::copy(values.begin(), values.end(), span.CpuSpan().begin());
+    span.CopyCpuToDevice();
+  };
+
+  auto packed_aux = make(aux_type_, {static_cast<int64_t>(num_ctx_rows),
```
baijumeswani, line 562: The proposal path creates this tensor and many other device tensors on every step. The target path also creates a separate auxiliary-output tensor. These allocations are directly in the token loop and are not included in cache sizing. **Recommendation:** allocate maximum-size reusable buffers when the Engine is created, use smaller views per step, and reserve their memory before sizing the target KV cache. Measure B1 as well as larger batches to confirm allocation cost is under control.

docs/paged_attention_engine.md
```diff
@@ -1107,6 +1112,22 @@ Graph buffers are allocated once at configured limits and reshaped as static vie
...
+The drafter run is synchronous because its packed inputs and outputs are owned by one proposal
+call, so `model.dflash2.run_options` cannot disable execution-provider synchronization.
+The direct drafter session also uses graph id `-1`: its variable-shaped proposal tensors are
+allocated per call and therefore cannot be captured safely. If this optional post-commit drafter
+run fails, the Engine discards any partial proposal, releases the drafter, and still publishes the
+already committed target events; subsequent Turns continue without automatic DFlash 2 proposals.
+
+The drafter keeps a fixed sliding-window ring per request from that request's first decode step
+until it is closed, so its pool is sized for `max_batch_size` rings. A request that first decodes
+while every ring is taken is skipped for the rest of its life and decodes without DFlash 2 drafts;
```
baijumeswani, line 1130: This means the first `max_batch_size` long-lived requests can keep every ring, even while their turns are idle, and later requests can never join because their position is no longer zero. That can create a large difference in performance between otherwise identical requests. **Recommendation:** document this as a service-capacity limit and add telemetry for ring admission misses. Longer term, consider a bounded rebuild or reassignment policy so rings can follow active requests.

test/cpp/dflash2_config_test.cpp
```diff
@@ -0,0 +1,269 @@
...
+}
+
+TEST(Dflash2ConfigTest, CapsDraftWidthBySessionAndTurnLimits) {
+  EXPECT_EQ(Dflash2DraftWidth(7, 5, 8, 10, 9), 1u);
+  EXPECT_EQ(Dflash2DraftWidth(7, 5, 8, 20, 3), 2u);
+  EXPECT_EQ(Dflash2DraftWidth(7, 5, 8, 20, 1), 0u);
+  EXPECT_EQ(Dflash2DraftWidth(7, 5, 9, 10, 9), 0u);
+}
+
+}  // namespace Generators::test
```
baijumeswani, line 269: These tests cover configuration and sizing, but none runs `Dflash2Drafter::Propose` or the Engine integration. That leaves guidance eligibility, packed row maps, ring reuse, proposal failure, and draft publication untested. **Recommendation:** add a tiny executable drafter model or a small test interface, then cover a normal proposal, mixed guided/ordinary requests, rejected draft rows, ring exhaustion/reuse, and a proposal failure that still lets the target step finish.

baijumeswani (commented):
This PR adds a DFlash2 draft model to the continuous-batching Engine. The main model returns selected hidden-state rows, the drafter stores a small sliding-window cache for each request, predicts a block of candidate tokens, and sends those candidates through the existing target-model verification path. The implementation also validates the exported graph contract, shares model weights, and keeps a drafter failure from rolling back an already completed target step.

The core design has several strong parts. The sliding-window ring size and wraparound math look correct, rejected draft rows are removed before they reach the drafter, the candidate-lattice walk matches the exporter layout, and size/overflow checks are thorough. The recent fixes for selector validation, output-name handling, request identity, and full-ring admission are also good improvements.

I still see a few correctness and service-reliability issues, plus performance and test gaps. The most important fixes are to use the shared draft eligibility check, keep feed preparation out of the final commit section, release drafter state during emergency cleanup, and add tests that actually execute the drafter path. The performance suggestions focus on keeping CUDA graphs available when possible, reusing buffers, and making long-context work depend on the window size rather than total context length.

src/engine/engine.cpp
tianleiwu: Addressed in 030736ae. DFlash 2 now uses Request::DraftTokenValidationError() as the shared eligibility gate while retaining its greedy-only requirement, so guided requests are skipped without affecting eligible requests.

src/engine/engine.cpp
tianleiwu: Addressed in 030736ae. Feed construction now runs with reservation preparation, before the irreversible target commit boundary, so an allocation or validation failure can still roll the transaction back cleanly.

src/engine/engine.cpp
tianleiwu: Addressed in 030736ae. Contract errors disable DFlash 2 immediately; transient proposal failures retry and disable after three consecutive failures. The speculative stats surface now exposes dflash2_failures and dflash2_disables.

src/dflash2_drafter.cpp
tianleiwu: Addressed in 030736ae. cached_positions is published only after session_->Run() succeeds. A newly admitted ring remains at position zero after a failed first run, allowing a consistent retry, and teardown releases it normally.

src/engine/engine.cpp
tianleiwu: Addressed in 030736ae. DetachRequestForTeardown() now releases the DFlash 2 request state inside its no-throw cleanup path before the request can disappear.

src/engine/decoders/simple_decoder.cpp
tianleiwu: Addressed in 030736ae. Auxiliary output binding is now controlled by a runtime aux_hidden_states_output_required flag, and capturable single-token decode steps use a persistent auxiliary-output graph buffer. Models that merely export the output no longer pay this cost.

src/dflash2_drafter.cpp
tianleiwu: Agreed. A bounded table requires rebasing the absolute position and PagedAttention block-table contract together and validating it with a long-context executable model. That broader optimization is not included in 030736ae; the current patch keeps the existing indexing semantics while addressing the correctness and lifecycle issues.

src/dflash2_drafter.cpp
tianleiwu: Partially addressed in 030736ae: the target auxiliary output now uses a persistent CUDA graph buffer on decode steps. Reusing the direct drafter proposal tensors also requires maximum-shape sizing, cache-budget reservation, and B1/batched measurements, so that larger performance change remains follow-up work.

docs/paged_attention_engine.md
tianleiwu: Addressed in 030736ae. The documentation now calls max_batch_size the DFlash 2 service-capacity limit across active and idle long-lived requests, and dflash2_admission_misses exposes denied ring admissions through speculative stats.

test/cpp/dflash2_config_test.cpp
tianleiwu: Agreed. The repository currently has no executable DFlash 2 ONNX test fixture or injectable drafter session, so 030736ae retains the 20 configuration/contract tests and focused engine coverage but does not pretend to exercise Propose. A model-backed fixture or test interface is still needed for the end-to-end matrix described here.

baijumeswani (commented):
This update addresses many of the concerns from the previous review. DFlash2 now uses the shared request check, so guidance no longer turns the drafter off. Feed preparation moved before the final commit, cached positions update only after a successful model run, emergency request cleanup releases rings, failure and admission counters were added, and auxiliary hidden states now use a runtime demand flag with persistent graph storage. These are meaningful improvements, and the core ring, packed-row, and candidate-selection logic still looks well designed.

I found a few new issues in the updated failure handling, plus some remaining test and performance gaps. The most important concern is that one temporary proposal failure still leaves the request and drafter out of step; the next token then looks like a contract error and disables DFlash2 for the whole Engine. The new warning path can also abort Debug builds when logging is disabled, and optional feed preparation can still mark the whole Engine unhealthy on an allocation failure.

I recommend fixing those failure paths and adding a testable DFlash2 interface with Engine-level tests before approval. The long-context block table, per-step allocations, and ring use by permanently ineligible requests are important performance follow-ups. The current benchmark numbers are from another stack commit, so rerunning the real model on this exact head would also make the approval case much stronger.

src/engine/engine.cpp
```diff
@@ -1602,6 +1785,48 @@ void Engine::RunDynamic() {
...
+          }
+          const bool disable_dflash2 = contract_error ||
+                                       dflash2_consecutive_failures_ >= kDflash2FailureDisableThreshold;
+          dflash2_disabled_ = disable_dflash2;
+          if (disable_dflash2) {
+            ++speculative_stats_.dflash2_disables;
+          }
+          if (dflash2_consecutive_failures_ == 1 || disable_dflash2) {
+            try {
+              Log("warning", AddExceptionCause(
```
baijumeswani, line 1817: `Log()` asserts that logging is enabled. Since logging is off by default, this warning can abort a Debug build, and the surrounding `try/catch` cannot catch an assertion abort. The same issue exists in the MTP warning above. **Recommendation:** call `Log()` only when `g_log.enabled && g_log.warning`, and add a fallback test with logging disabled.

src/dflash2_drafter.cpp
```diff
@@ -0,0 +1,698 @@
...
+    output_names.push_back(cache_output_names_[i].c_str());
+    outputs.push_back(caches_[i]->GetOrtTensor());
+  }
+
+  model_->session_->Run(run_options_.get(), input_names.data(), inputs.data(), input_names.size(),
+                        output_names.data(), outputs.data(), output_names.size());
+
+  for (const size_t i : served) {
+    const auto& feed = feeds[i];
+    requests_[feed.request].cached_positions = CheckedAdd(
```
baijumeswani, line 670: Moving this update after a successful model run was correct. However, after a temporary run failure the target still advances while this value does not. On the next token, the context check fails and DFlash2 is disabled for the whole Engine, so the documented three-failure retry does not really happen. **Recommendation:** after a recoverable proposal failure, forget and release the affected requests' drafter state. Those requests can continue without drafts, while new requests can still use DFlash2. Keep global disable for a real model-wide contract error.

src/engine/engine.cpp
```diff
@@ -1572,6 +1750,11 @@ void Engine::RunDynamic() {
       if (mtp_step) {
         mtp_step->reservation->PrepareCommit();
       }
+      if (dflash2_drafter_ && !dflash2_disabled_ && MaxDraftTokensPerStep() > 0) {
+        // Reads accepted-draft counts that CommitStep clears below. Keep this fallible work on the
+        // rollback side of the target transaction's commit boundary.
+        PrepareDflash2Feeds(step_plan_, step_results_);
```
baijumeswani, line 1756: Moving feed preparation before commit fixes the earlier partial-commit problem. This call can still allocate while growing its vectors, though, and any exception currently rolls back and permanently marks the Engine unhealthy. DFlash2 is optional, so an allocation failure here should not kill normal decoding. **Recommendation:** reserve these vectors to `max_batch_size` in the Engine constructor and give this call its own fallback handler. Count the failure and continue with a target-only commit for recoverable errors.

test/cpp/dflash2_config_test.cpp
baijumeswani, line 269: The new config and sizing tests are useful, but there is still no test that runs `Dflash2Drafter::Propose()` or an Engine step with DFlash2 attached. That leaves the retry/disable logic, row maps, ring reuse, admission, and draft publication untested. **Recommendation:** add a small drafter interface and recording test double, then mirror the MTP failure tests and add one packed-layout/ring test. A tiny executable drafter model would provide even stronger coverage.

src/config.cpp
```diff
@@ -1122,6 +1124,137 @@ struct Mtp_Element : JSON::Element {
...
+      v_.block_size = SafeDoubleToInt(JSON::Get<double>(value), name);
+      if (v_.block_size <= 1) throw std::out_of_range("block_size must be > 1");
+    } else if (name == "num_draft_tokens") {
+      v_.num_draft_tokens = SafeDoubleToInt(JSON::Get<double>(value), name);
+      if (v_.num_draft_tokens <= 0) throw std::out_of_range("num_draft_tokens must be > 0");
+    } else if (name == "selector_top_k") {
+      v_.selector_top_k = SafeDoubleToInt(JSON::Get<double>(value), name);
+      if (v_.selector_top_k <= 0) throw std::out_of_range("selector_top_k must be > 0");
+    } else if (name == "mask_token_id") {
+      v_.mask_token_id = SafeDoubleToInt(JSON::Get<double>(value), name);
```
baijumeswani, line 1208: This accepts any integer. If the field is missing it defaults to token 0; if it is negative or outside the vocabulary it is sent to the drafter embedding lookup. That can produce silent poor drafts or a runtime failure. **Recommendation:** require `0 <= mask_token_id < model.vocab_size` during config validation and add lower/upper boundary tests.

src/dflash2_drafter.cpp
```diff
@@ -0,0 +1,698 @@
...
+    std::fill(cpu.begin(), cpu.end(), int32_t{-1});
+    for (size_t row = 0; row < served.size(); ++row) {
+      const auto& blocks = requests_[feeds[served[row]].request].blocks;
+      if (ring_blocks_ == 0) {
+        std::copy(blocks.begin(), blocks.end(), cpu.begin() + row * max_blocks);
+        continue;
+      }
+      // A windowed drafter repeats its ring across every column: column j holds the block that
+      // owns position j * block_size, which is ring[j % ring_blocks].
+      for (size_t column = 0; column < max_blocks; ++column) {
```
baijumeswani, line 618: This table still grows with the full context length even though the cache is a fixed sliding-window ring. Long contexts therefore rebuild and copy more repeated block IDs on every token. **Recommendation:** use a window-relative table with rebased positions, or cache a persistent repeating table and update it only when the served requests or required width change. Add a 128K-style test that proves the per-step work stays bounded.

src/dflash2_drafter.cpp
```diff
@@ -0,0 +1,698 @@
...
+    tensor->CreateTensor(shape);
+    return tensor;
+  };
+  auto fill_int32 = [](Tensor& tensor, const std::vector<int32_t>& values) {
+    auto span = tensor.GetDeviceSpan<int32_t>();
+    std::copy(values.begin(), values.end(), span.CpuSpan().begin());
+    span.CopyCpuToDevice();
+  };
+
+  auto packed_aux = make(aux_type_, {static_cast<int64_t>(num_ctx_rows),
```
baijumeswani, line 562: The persistent target auxiliary buffer is a good improvement, but the drafter still creates this tensor and many row-map, metadata, block-table, candidate, and score tensors every token. This can add latency and memory spikes. **Recommendation:** allocate bounded reusable buffers when the drafter is created and reshape views for each step. Include this scratch memory when sizing the target cache.

docs/paged_attention_engine.md
```diff
@@ -1107,6 +1112,27 @@ Graph buffers are allocated once at configured limits and reshaped as static vie
...
+call, so `model.dflash2.run_options` cannot disable execution-provider synchronization.
+The direct drafter session also uses graph id `-1`: its variable-shaped proposal tensors are
+allocated per call and therefore cannot be captured safely. If this optional post-commit drafter
+run fails, the Engine discards any partial proposal and still publishes the already committed target
+events. A transient failure is retried on the next step; three consecutive failures disable DFlash 2
+for the Engine. `dflash2_failures` and `dflash2_disables` report those events.
+
+The drafter keeps a fixed sliding-window ring per request from that request's first decode step
+until it is closed, so its pool is sized for `max_batch_size` rings. A request that first decodes
+while every ring is taken is skipped for the rest of its life and decodes without DFlash 2 drafts;
```
baijumeswani, line 1130: Admission misses are now documented and counted, which is helpful. Sampled and guided requests still take rings and run context through the drafter even though they can never receive DFlash2 drafts. Enough such requests can use all rings and block eligible greedy requests. **Recommendation:** avoid admitting requests that are permanently ineligible, or separate context tracking from proposal eligibility. Also document that automatic DFlash2 drafting is currently greedy-only.

baijumeswani (approved):

src/engine/engine.cpp
tianleiwu: Confirmed and fixed in https://github.com/microsoft/onnxruntime-genai/pull/2531.

`Log()` opens with `assert(g_log.enabled)`, and the surrounding `try`/`catch` only catches exceptions, so a Debug build with logging off could abort on the first drafter failure. Both this warning and the MTP one above it are now guarded with `g_log.enabled && g_log.warning` before the call, matching how `scheduled_requests.cpp` and the KV-cache code gate their warnings.

src/dflash2_drafter.cpp
tianleiwu: Confirmed and fixed in https://github.com/microsoft/onnxruntime-genai/pull/2531.

The analysis was right: `cached_positions` only advances after a successful run, so the next step's contiguity check threw `std::logic_error`, which is classified as a contract error and disabled DFlash 2 immediately. The three-failure budget was unreachable.

A recoverable failure now calls a new `Dflash2Drafter::ReleaseAll()`, which returns every ring to the pool and forgets the tracked requests. Since a request can only join the drafter from position zero, those in-flight requests finish without drafts while requests admitted afterwards still get them. Global disable is kept for a real contract error. This runs on both the prepare-side and publish-side failure paths, because either one leaves the same bookkeeping gap.

src/engine/engine.cpp
tianleiwu: Confirmed and fixed in https://github.com/microsoft/onnxruntime-genai/pull/2531.

`PrepareDflash2Feeds()` now sits outside the transaction-preparation `try` and has its own handler: only a `std::logic_error` is fatal, anything else is counted as a DFlash 2 failure and the step commits target-only. `dflash2_feeds_`, `dflash2_draft_widths_` and `dflash2_drafts_` are reserved to `max_batch_size` in the Engine constructor, so a committed step never has to grow them.

test/cpp/dflash2_config_test.cpp
tianleiwu: Partly addressed in https://github.com/microsoft/onnxruntime-genai/pull/2531.

The gap is real. `engine_unit_tests` builds requests from a fixture model that never runs a graph, and `Dflash2Drafter`'s constructor needs a live session for `session_info_` and the cache allocation, so `Propose()`, ring reuse, admission and draft publication cannot be reached without either the drafter interface plus test double you suggested or a tiny executable `dflash2.onnx`. That is left as follow-up work.

What the follow-up does cover is the host-side logic those paths depend on, now that it is reachable without a session: `mask_token_id` bounds, the greedy-only admission rule (`Dflash2CanDraft`), and the proposal-buffer reuse and growth rules (`Dflash2StepTensor`), including the case where a step outgrows the buffer, which would otherwise only surface at run time as `Tensor: Static buffer new_bytes > bytes_`.

src/config.cpp
tianleiwu: Confirmed and fixed in https://github.com/microsoft/onnxruntime-genai/pull/2531.

`CreateDflash2Config()` now rejects a `mask_token_id` outside `[0, model.vocab_size)`, alongside the existing geometry checks, with lower/upper boundary tests plus both valid endpoints in `dflash2_config_test.cpp`. The upper bound is only enforced when `model.vocab_size` is set, so a config that omits it still loads.

src/dflash2_drafter.cpp
tianleiwu: Half of this is fixed in https://github.com/microsoft/onnxruntime-genai/pull/2531; the other half is not implementable as suggested.

The window-relative table with rebased positions cannot work against the current operator. `PagedAttention` resolves a slot as `block_table[b][(past_seqlens[b] + offset) / block_size]` using the token's absolute position, and with `do_rotary=1` it rotates K with that same absolute position *before* the cache write. Rebasing the query positions while cached keys carry their original rotation breaks the relative offsets RoPE depends on. The op spec calls "the block table stays indexed by absolute position" a load-bearing invariant and documents `block_table[b][i] = ring[i % N]` as the supported sliding-window policy, which is what this code and `PagedKeyValueCache::FillBlockTables` both already do. So the width has to keep tracking absolute KV length.

The persistent-table half is done: the block table is now one of eleven proposal tensors reused across steps, so the per-step allocation is gone, and the fill no longer blanks the whole table first since the windowed path writes every column anyway.

src/dflash2_drafter.cpp
tianleiwu: Confirmed and fixed in https://github.com/microsoft/onnxruntime-genai/pull/2531.

All eleven per-step tensors (packed aux, row maps, block-row index, cumulative and past lengths, block table, attention metadata, and the two lattice outputs) now live in a `StepTensors` member and reshape a view over a buffer the drafter keeps, through a new `Dflash2StepTensor()` helper. It replaces the buffer only when a step outgrows the high-water mark or changes element type, so a steady-state decode step allocates none of them.

One deviation from the suggestion: they grow lazily rather than being allocated at construction. `packed_aux` is bounded by `max_scheduled_tokens * aux_hidden_size`, which for this model is hundreds of MB at a large chunk size, so pre-allocating that ceiling would reserve memory a decode-only workload never touches. Including the scratch in the target cache sizing is still open; in practice it changes little, since `p_device_inputs_` is an ORT arena that never returned the old per-step allocations to the driver either.

docs/paged_attention_engine.md
tianleiwu: Confirmed and fixed in https://github.com/microsoft/onnxruntime-genai/pull/2531.

`PrepareDflash2Feeds()` now skips a request whose search options can never take a block, via a named `Dflash2CanDraft()` predicate, so sampled requests no longer claim a ring or run their context through the drafter. Greediness is fixed when the request is created, so a request skipped on its first step is skipped for its whole life and never joins the drafter's contiguity contract.

`docs/paged_attention_engine.md` now states that automatic DFlash 2 drafting is greedy-only, and also documents the per-request forget on a recoverable failure.

PR #2462, opened 2026-08-24
Add DSpark block drafter support

tianleiwu:
## Description

Adds end-to-end DSpark speculative decoding support for Qwen3.8 models. The implementation
reuses the DFlash2 block-drafter lattice runtime introduced by #2460, while adding DSpark
graph export, Markov scoring, shared target weights, and cache accounting for both full and
sliding-window draft models.

This PR is intentionally stacked on #2460 and should be reviewed after that PR.

### DSpark model export

| File | Change |
|------|--------|
| `src/python/py/models/builders/dspark.py` | Builds the BF16 DSpark graph with dual-source non-causal attention, YaRN RoPE, target embedding and LM-head reuse, top-k candidates, and FP32 Markov lattice scores. |
| `src/python/py/models/builders/qwen.py` | Adds DSpark export options, validates target hidden-state taps, saves the auxiliary graph, shares external initializers, and emits `model.dspark` configuration. |

### Runtime integration

| File | Change |
|------|--------|
| `src/config.cpp` | Parses `model.dspark` through the common block-drafter configuration. |
| `src/dflash2_drafter.h` | Exposes cache sizing helpers used by engine pool planning. |
| `src/dflash2_drafter.cpp` | Accepts DSpark's all-predicting block geometry and supports fixed-ring or full-context cache allocation according to `sliding_window`. |
| `src/engine/engine.cpp` | Budgets full-attention drafter KV bytes with the main paged cache and retains fixed per-request rings for windowed drafters. |

## Testing

- `cmake --build build/cuda130/Release --parallel`
- `./build/cuda130/Release/engine_unit_tests`
- `python3 -m py_compile src/python/py/models/builders/dspark.py src/python/py/models/builders/qwen.py`
- Standalone graph validation: 7/7 top-1 candidates matched; 109/112 top-k overlap.
- Greedy losslessness: 0 mismatches across 2,042 generated tokens.
- Eight-prompt acceptance test: 3.586 tokens per verification step over 1,862 steps.
- H200 benchmark matrix: 2.09x to 5.08x decode throughput over the same target graph without a drafter across context lengths 512-8192 and batch sizes 1-16.
- MMLU-Pro 800 samples: 82.62% with DSpark versus 83.00% without drafting (`p=0.508`).
- GPQA Diamond: 76.26% with DSpark versus 72.73% without drafting (`p=0.143`).

## Motivation and Context

DSpark extends the DFlash-style parallel block drafter with a low-rank Markov head. Both
[...]

PR #2497, opened 2026-08-28
Reuse the block drafter runtime for DSpark

tianleiwu:
## Description

Reuses the DFlash 2 block-drafter runtime (merged in #2496 / #2531) for DSpark artifacts. DSpark emits the same candidate-and-edge-score lattice with a Markov head, so the runtime only needs a config alias, support for one prediction per block row, and full-attention KV budgeting against the target paged-cache size.

This replaces #2462 while excluding DSpark builder changes already covered by #2490, historical diagnostics, dense-checkpoint experiments, and ONNX Runtime dependency changes.

### Shared drafter configuration

- Parses `model.dspark` through the existing block-drafter runtime schema, and rejects a configuration that names both `model.dflash2` and `model.dspark`.
- Accepts `num_draft_tokens == block_size` for DSpark while preserving DFlash 2's `block_size - 1` layout.
- Continues to consume the same packed auxiliary hidden-state output and candidate lattice contract.
- Tightens drafter contract validation: unique input/output names, dynamic packed dimensions, the paged page size, the lattice geometry, and the supported cache element types. Validation now returns that element type so cache budgeting no longer assumes bfloat16.

### Full-attention cache budgeting

- Keeps the existing fixed batch-scaled ring for windowed DFlash 2 artifacts.
- Bills full-attention DSpark KV bytes per target paged-cache block before target allocation, plus a fixed reservation for the query-block spill.
- Creates a matching drafter pool after target sizing, with one query-spill block per configured batch row.
- Adds overflow checks to the paged-cache byte and block-count arithmetic that now mixes two pools.

### Runtime bounds

- A full-attention drafter admits at most `max_batch_size` concurrent requests, matching the windowed drafter. Requests keep their drafter blocks while they are resident and idle, so unbounded admission could exhaust the pool mid-proposal and disable the drafter for the whole Engine. Refusals are reported through `dflash2_admission_misses`.
- Proposal buffers double on growth. A full-attention block table gains a column every paged block of committed context, so an exact-fit static buffer reallocated device memory every few decode steps for the life of a request.

### Scope

- Reuses the #2496 engine lifecycle, accepted-prefix filtering, request teardown, and telemetry path.
- Does not duplicate Python builder/export code from #2490.
- Does not add dense recurrent checkpoints or mixed-checkpoint scheduling.

## Documentation

`docs/paged_attention_engine.md` now covers the `model.dspark` alias, the two drafter row layouts, the windowed-versus-full-attention cache budgeting trade (a DSpark drafter with the target's layer geometry roughly halves paged-cache capacity), and the `max_batch_size` drafter service limit.

## Testing

- `cmake --build build/cpu_guidance/Release --target unit_tests engine_unit_tests` (passed)
- `unit_tests` (251 passed, 0 failed)
- `engine_unit_tests` (592 passed, 0 failed), including 46 focused DFlash 2/DSpark config, cache-planning, and proposal-buffer tests

Real-model DSpark proposal correctness and throughput remain follow-up validation for this draft.

## Dependencies

- Rebased onto `main` after #2496 and #2531 merged.
- Uses DSpark exporter support from #2490; builder code is not duplicated here.

## Checklist

- [x] Focused unit coverage added
- [x] Full C++ and engine suites pass
- [ ] Real-model DSpark proposal and throughput validation


docs/paged_attention_engine.md
```diff
@@ -851,12 +851,24 @@ Graph buffers are allocated once at configured limits and reshaped as static vie
...
-The drafter run is synchronous because its packed inputs and outputs are owned by one proposal
-call, so `model.dflash2.run_options` cannot disable execution-provider synchronization.
+An Engine-hosted DFlash 2 or DSpark block drafter also forces eager target execution. `model.dspark`
+is a configuration alias for the shared DFlash 2 runtime, and both sections cannot appear in one
+configuration. DFlash 2 uses `block_size - 1` draft rows because its first row is an anchor; DSpark
+predicts from all `block_size` rows.
+
+The target decoder output is the packed auxiliary hidden-state tensor named by
+`main_aux_hidden_states`; that variable-size output does not have persistent graph buffers. Engine
+construction validates its rank, element type, and static width against the drafter input. It also
```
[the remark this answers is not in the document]
tianleiwu: Addressed in ce48686a by documenting both full configuration paths: model.dflash2.main_aux_hidden_states and model.dspark.main_aux_hidden_states.

src/engine/engine.cpp
```diff
@@ -292,21 +296,47 @@ EngineDependencies Engine::CreateDependencies(std::shared_ptr<Model> model) {
...
-                                   static_cast<size_t>(batching.max_batch_size)));
+    const size_t pool_blocks = Dflash2Drafter::PoolBlocks(
+        *model->config_, paged_block_size, dflash2_max_batch_size);
+    if (pool_blocks != 0) {
+      // Built before the main pool so the pool's free-memory measurement already excludes it. A
+      // windowed drafter's footprint depends on the batch size, not the context length.
+      dflash2_drafter = std::make_unique<Dflash2Drafter>(
+          dflash2_model, paged_block_size, pool_blocks, dflash2_max_batch_size);
+    } else {
+      // A full-attention drafter mirrors the target pool, so it is billed per target block instead
```
baijumeswani, line 313: Could we limit this full-attention branch to `model.dspark`? Right now any DFlash2 config without a positive sliding window reaches this path. That used to fail during setup, but now it creates a second cache that grows with the target context. This can reduce the available context and make drafting slower as the sequence grows, without the user choosing that behavior.

I suggest checking `is_dspark` here (or in `PoolBlocks`) and keeping the old setup error for DFlash2. If full-attention DFlash2 is intended, please make it an explicit option and document and test its memory and performance cost.

src/engine/paged_key_value_cache.cpp
```diff
@@ -108,7 +119,8 @@ size_t ComputeNumBlocks(std::shared_ptr<Model> model,
                         size_t full_layer_count,
                         size_t windowed_bytes,
                         ONNXTensorElementDataType dtype,
-                        size_t auxiliary_bytes_per_block) {
+                        size_t auxiliary_bytes_per_block,
+                        size_t auxiliary_reserved_memory_bytes) {
```
baijumeswani, line 123: The direct `num_blocks` path above does not use this fixed reserved amount. It includes the drafter cost for each target block, but DSpark later allocates extra blocks for the current query. The real allocation can therefore exceed the memory limit represented by `num_blocks`. The automatic sizing path below already subtracts this reserve.

Please apply the same rule to the direct path: start with the byte budget represented by the configured target blocks, subtract the fixed reserved bytes, then divide the remainder by the combined target-and-drafter bytes per block. Please also reject a setting that cannot cover the fixed reserve and add a test with direct `num_blocks` plus a nonzero reserve.

test/cpp/dflash2_config_test.cpp
```diff
@@ -251,6 +321,54 @@ TEST(Dflash2ConfigTest, RequiresMaskTokenInsideVocabulary) {
   EXPECT_NO_THROW(CreateDflash2Config(config));
 }
 
+TEST(Dflash2ConfigTest, AcceptsOneDraftPerDsparkBlockRow) {
+  auto config = MakeDflash2Config();
+  config.model.dflash2.is_dspark = true;
+  config.model.dflash2.num_draft_tokens = config.model.dflash2.block_size;
+  EXPECT_NO_THROW(CreateDflash2Config(config));
```
baijumeswani, line 328: These new tests do a good job checking the DSpark config and the memory calculations, but none of them runs a full-attention drafter step. A row-layout or block-table mistake could still build successfully and only show up as wrong drafts or very low acceptance.

Please add a small executable test that admits at least two requests, grows them across cache blocks, builds the block tables, produces drafts, and then checks release/reuse and `ReleaseAll()` after a failure. A smoke run with an exported DSpark model should also confirm target output stays correct and report draft acceptance at short and long context.

baijumeswani (commented):
This PR adds DSpark support to the continuous-batching Engine by reusing the existing DFlash2 block-drafter runtime. It adds the `model.dspark` config name, checks that the exported model matches the runtime, supports a full-attention drafter cache, includes that cache in memory sizing, reuses step buffers, and documents the new behavior.

The shared design is a good choice because it avoids building a second runtime for nearly the same job. The config and model checks are thorough, the automatic memory calculation is well thought out, overflow is handled in most new calculations, and the request cleanup paths appear complete. The documentation also explains the difference between DSpark and DFlash2 clearly.

I found three items that I think should be addressed before approval:

1. The new full-attention path also changes DFlash2 behavior when no positive window is set. This should be limited to DSpark unless full-attention DFlash2 is an intentional and tested feature.
2. When `num_blocks` is set directly, the calculation does not include the extra blocks reserved for the drafter's current query. The actual allocation can therefore exceed the requested memory limit.
3. The tests check config values and memory math, but they do not run the new full-attention DSpark path. A small executable test and one exported-model smoke run would give confidence that cache growth, block tables, release/reuse, and draft output work together.

The overall approach looks solid. The inline comments below include specific suggestions for each item.

src/engine/engine.cpp
tianleiwu: Addressed in 56dca92b. `CreateDflash2Config` now rejects a non-positive sliding window unless the configured block drafter is DSpark, so DFlash2 retains its previous setup error instead of silently allocating a full-attention cache. `RejectsFullAttentionForDflash2` covers the guard.

src/engine/paged_key_value_cache.cpp
tianleiwu: Addressed in 56dca92b. The explicit `num_blocks` path now treats the configured target allocation as a byte budget, subtracts `FullAttentionReservedBytes`, and divides the remainder by the combined target-plus-drafter bytes per block. It rejects budgets that cannot cover the fixed reserve and checks both multiplication and addition overflow. The focused test covers a nonzero reserve and the exact-reserve rejection boundary.

test/cpp/dflash2_config_test.cpp
tianleiwu: Addressed in 56dca92b with a deterministic executable ONNX fixture that runs the real full-attention `Dflash2Drafter::Propose` path. The test admits two requests, crosses from four to five cache blocks, derives drafts from the actual block table/past lengths/packed row lengths/q-row map, verifies individual release and block reuse, injects a non-contiguous proposal failure, calls `ReleaseAll()`, and verifies two new requests can draft afterward. The full engine suite passes (594 tests). A production-export acceptance/quality run still requires an external DSpark checkpoint and target model; those assets are not checked into this repository, so that remains model-validation rather than a deterministic unit-test dependency.

baijumeswani (approved):

PR #2507, opened 2026-08-31
Harden Engine fatal-event publication and abandonment recovery

baijumeswani:
## Summary
 
This follow-up hardens how the Engine publishes fatal failures and recovers abandoned requests. It builds on the Request lifecycle and ownership behavior already merged into `main`; it does not redefine request closure or static-batch lifetime.
 
The main goal is to ensure that an Engine-wide failure remains observable and ordered even when it occurs during allocation pressure, speculative multi-token delivery, abandonment cleanup, or an API boundary outside `Run()`.
 
 ## What changed
 
 ### Reliable fatal-event publication
 
 - Reserve dedicated fatal-event capacity before a Request becomes visible to the Engine.
 - Preconstruct a fallback exception so fatal publication does not depend on allocating diagnostic state after the failure.
 - Preserve all undrained events when the Engine becomes unhealthy.
 - Preserve speculative token ordering and attach terminal failure to the final retained event for that Turn.
 - Publish a request-less failure event when no caller-owned Request can safely carry the failure.
 - Allow already-published fatal events to drain before subsequent progress calls rethrow the durable Engine error.
 
 ### Robust abandoned-request recovery
 
 - Retry abandonment cleanup when deallocation fails with `std::bad_alloc`.
 - Treat ownership and cache invariant failures during cleanup as Engine-fatal instead of continuing with potentially inconsistent state.
 - Terminalize other executable Turns when cleanup poisons the Engine.
 - Filter abandoned Request identities from host-visible events because their final public handles have already been released.
 
 ### Host handling and documentation
 
 - Update the Python Engine examples and guidance benchmark to reject failed events consistently.
 - Keep request-scoped `REQUEST_UNSERVICEABLE` failures recoverable in the continuous-batching sample while surfacing Engine-wide failures.
 - Document fatal-event draining, request-less failures, speculative multi-token delivery, abandonment retry behavior, and current static-batch recycling semantics.


src/engine/engine.cpp
```diff
@@ -999,17 +1104,36 @@ EngineEvent Engine::EventFromStepError(
...
+    fatal_events_.push_back(std::move(event));
+  }
+
+  fatal_error_ = durable_error;
+  health_ = EngineHealth::Unhealthy;
+  if (outcome == StepOutcomeKind::FatalExecutionFailure ||
+      outcome == StepOutcomeKind::ExecutionContractFailure) {
+    ++transaction_metrics_.fatal_execution_failures;
+  }
+  pending_events_.swap(fatal_events_);
```
tianleiwu, line 1856: **High** — this swap leaks strong `Request` references for the rest of the Engine's lifetime.

After the swap, `fatal_events_` holds the *old* `pending_events_` contents: the already-drained prefix `[0, pending_event_index_)` plus the events for abandoned Requests that the filter above deliberately excluded. Each `EngineEvent` owns a `std::shared_ptr<Request>`.

`MarkUnhealthyAndThrow` rethrows immediately on every subsequent entry (`if (health_ == EngineHealth::Unhealthy) std::rethrow_exception(fatal_error_);`), so `fatal_events_.clear()` at the top is never reached again. `Engine::~Engine()` / `DetachRequestForTeardown` purge `pending_events_` and `staged_events_` but not `fatal_events_`.

Net effect: after any fatal failure, every Request that had a retained event stays alive until the `Engine` object itself is destroyed — even after the host drains all events, destroys the handles, and `CompleteNonresidentClosedRequests()` drops them from `tracked_requests_`. It also quietly defeats the abandoned-identity filtering this PR adds, since the filtered-out events still hold their strong refs.

Fix is one no-throw, non-allocating line:

```cpp
pending_events_.swap(fatal_events_);
fatal_events_.clear();  // Drop the stale strong Request references carried over by the swap.
pending_event_index_ = 0;
```

src/engine/engine.cpp
```diff
@@ -393,22 +459,33 @@ void Engine::ValidateRequestCanContinue(
...
     std::exception_ptr restore_error) {
   request->MarkFailedFromEngine(*this);
-  std::string message = AddExceptionCause(
-      "Continuation append failed and its Search state could not be restored.",
-      append_error);
+  std::string message;
   try {
-    CloseRequest(request);
-  } catch (...) {
     message = AddExceptionCause(
```
tianleiwu, line 1156: **Suggestion** — if this `AddExceptionCause` throws `std::bad_alloc`, control jumps straight to the outer `catch (...)`, which calls `request->CompleteCloseFromEngine(*this)` **without** ever running `CloseRequest(request)`.

That skips `scheduler_->RemoveRequest(request)` and the `pending_events_`/`staged_events_` purge, and leaves the Request in `tracked_requests_`, while its Search/guidance/sampler state is fully released. `MarkUnhealthyAndThrow` then copies those stale events into `fatal_events_` and hands the host events for a Request that no longer has runtime state.

The previous code had the opposite failure mode (a `bad_alloc` here propagated out with the Engine still *healthy* and the Request poisoned), so this is an improvement — but since the point of the PR is hardening allocation-failure paths, it seems worth attempting the removal sequence before building the diagnostic:

```cpp
try {
  CloseRequest(request);
} catch (...) {
  request->CompleteCloseFromEngine(*this);
}
std::string message;
try {
  message = AddExceptionCause(
      "Continuation append failed and its Search state could not be restored.",
      append_error);
} catch (...) {
  MarkUnhealthyAndThrow(
      StepOutcomeKind::FatalExecutionFailure, 0, request.get(),
      "Continuation append or restore failed, and constructing its diagnostic also failed.",
      nullptr);
}
```

src/engine/engine.cpp
```diff
@@ -310,30 +364,42 @@ void Engine::ReclaimAbandonedRequests() {
...
-    if (request->ExternalReferencesAbandoned()) {
-      CloseRequest(request);
+      auto request = *abandoned;
+      // Recheck immediately before removal in case an external owner was reacquired before this
+      // serialized boundary.
+      if (request->ExternalReferencesAbandoned()) {
+        CloseRequest(request);
+      }
     }
+  } catch (const std::bad_alloc&) {
```
tianleiwu, line 1074: **Suggestion** — the retry is correct today, but only because of an invariant nothing enforces.

It works because the sole realistic `bad_alloc` source inside `CloseRequest` is `scheduler_->RemoveRequest()` → `CacheManager::Deallocate()`, and `PagedCacheManager::Deallocate` gathers `allocated_to_remove` / `fixed_handles` before releasing anything — so a throw there leaves committed ownership untouched. Crucially, it runs *before* `request->MarkClosedFromEngine()`.

If a future change introduces an allocating step after `MarkClosedFromEngine()`, the retry scan's `!IsClosed(request->status_)` predicate will skip the Request forever; `CompleteNonresidentClosedRequests()` will not pick it up either because the failed `Deallocate` left it resident; and `CloseRequest`'s own `if (request && IsClosed(request->status_)) return;` early exit makes the state unrecoverable. That is a silent, permanent cache-block leak.

Worth either an explicit comment/assert stating "nothing after `MarkClosedFromEngine` may allocate", or making recovery state-based (also retry `Closed` Requests that are still resident and abandoned) instead of status-based.

src/engine/engine.cpp
```diff
@@ -129,9 +177,17 @@ std::shared_ptr<Request> Engine::CreateRequest(const GeneratorParams& params,
...
 
-  // Fatal handling may need one terminal event for every tracked Request. Grow that storage before
-  // publishing the Request so failure reporting never allocates after the Engine becomes unhealthy.
-  pending_events_.reserve(tracked_requests_.size() + 1);
+  // Fatal handling preserves every token event from the current step, then needs one terminal
+  // event per tracked Request plus a request-less fallback. Grow dedicated storage before
+  // publishing the Request so terminalization cannot allocate.
+  if (max_step_event_count_ >= fatal_events_.max_size() ||
+      tracked_requests_.size() >
+          fatal_events_.max_size() - max_step_event_count_ - 2) {
```
tianleiwu: **Nitpick** — this guard can underflow. The first clause only proves `fatal_events_.max_size() - max_step_event_count_ >= 1`; when that difference is exactly `1`, `- 2` wraps to `SIZE_MAX`, the check passes, and the intended `std::overflow_error` is replaced by a `std::length_error` from `reserve` below.

Practically unreachable, but since the guard is written to be exact, prefer the non-wrapping form:

```cpp
if (max_step_event_count_ + 2 < max_step_event_count_ ||
    fatal_events_.max_size() - (max_step_event_count_ + 2) <
        tracked_requests_.size()) {
```

Related: the constructor uses a named `fatal_event_overhead = 2` while this site open-codes `2` twice. A single shared constant would keep the two capacity proofs from drifting apart.

src/engine/engine.h
```diff
@@ -74,6 +77,7 @@ struct EngineDependencies {
   std::shared_ptr<CacheManager> cache_manager;
   std::unique_ptr<Scheduler> scheduler;
   std::unique_ptr<ModelExecutor> model_executor;
+  EngineStepErrorFactory make_step_error{};
```
tianleiwu, line 85: **Suggestion** — this is a fault-injection seam with no production consumer. `Engine::CreateDependencies` never sets it, and the only non-default value in the tree is `CaptureBadAllocInsteadOfStepError` in `engine_run_tests.cpp`.

Adding an injection point to a shipped dependency struct purely for testing is a reasonable trade-off here (the failure it simulates is otherwise unreachable), but it reads like a customization point. At minimum, a comment such as `// Test-only: overrides durable fatal-error construction to exercise the fallback path.` would prevent it from being treated as public configuration.

docs/engine_c_api_spec.md
```diff
@@ -328,8 +328,9 @@ The Engine records its owner thread when it is created. Engine operations, Reque
 Turn-option setters are not thread-safe and must be called serially from that owner thread.
 Destroying a Request handle only publishes an atomic abandonment marker when it is the final public
 handle; this final release may occur on another thread. The Engine strongly retains the Request and
-performs the actual close, runtime-state release, and cleanup at its next owner-thread boundary.
-This deferred-release behavior does not permit concurrent Request operations.
+performs the logical close and cleanup at its next owner-thread boundary. A resident static row
+retains the runtime state needed by an executable peer until the shared physical batch becomes
+recycled. This deferred-release behavior does not permit concurrent Request operations.
```
tianleiwu, line 333: **Nitpick** — grammar: "until the shared physical batch becomes recycled" → "until the shared physical batch **is** recycled".

docs/engine_c_api_spec.md
```diff
@@ -449,15 +450,18 @@ struct PendingEngineEvent {
 At Engine construction:
 
 ```cpp
-pending_events_.reserve(cache_manager_->MaxBatchSize());
+const size_t max_step_events =
+    cache_manager_->MaxBatchSize() * kMaxGeneratedTokensPerStep;
+pending_events_.reserve(max_step_events);
+staged_events_.reserve(max_step_events);
```
tianleiwu, line 456: **Nitpick** — the snippet still shows only `pending_events_`/`staged_events_`, but the prose immediately below now describes dedicated fatal-event capacity, and the constructor actually does `fatal_events_.reserve(max_step_event_count_ + 1);`. Adding that line to the block would keep the snippet consistent with the implementation it is documenting.

docs/paged_attention_engine.md
```diff
@@ -766,11 +773,15 @@ The engine becomes unhealthy when it cannot prove that all components still agre
 - Any part of the commit boundary fails.
 - Planning detects inconsistent committed paged and fixed state.
 - The scheduler returns an invalid planning outcome.
+- Abandoned-Request cleanup detects inconsistent cache or scheduler ownership.
 
 The engine stores the fatal error and marks every executable Turn complete with reason `Failed`.
 Before rethrowing the stored error on later `Run()` calls, it emits one
 `TurnFinished | Failed` event per affected Turn with the Turn's Request, ID, usage, and Engine
 failure code. A fatal failure with no affected Turn emits one request-less Engine failure event.
+An externally abandoned Request has no caller-owned handle and does not receive a request-bearing
```
tianleiwu: **Nitpick** — this states the abandoned-identity filter as a guarantee, but it is a publication-time snapshot. `ExternalReferencesAbandoned()` reads an atomic the host may flip from any thread, so a Request abandoned *after* `MarkUnhealthyAndThrow` builds `fatal_events_` but *before* the host drains them still yields a request-bearing event.

Suggest softening to something like "is filtered from host-visible events on a best-effort basis at publication time" so hosts do not rely on it as an invariant.

examples/python/engine/model-qa.py
```diff
@@ -11,11 +11,11 @@
 
 
 def require_request_event(event: og.EngineEvent) -> og.Request:
+    if event.flags & og.EngineEventFlags.FAILED:
+        raise RuntimeError(f"Generation failed; error_code={event.error_code}")
```
tianleiwu, line 17: **Suggestion** — this (and the identical change in `tool-calling.py` and `benchmark/python/benchmark_engine_guidance.py`) now raises on **every** `FAILED` event, including the request-scoped, recoverable `REQUEST_UNSERVICEABLE`.

That directly contradicts what `examples/python/engine/continuous-batching.py` and the new `docs/paged_attention_engine.md` sample teach in this same PR: a `FAILED` event with a non-null `request` and `error_code == REQUEST_UNSERVICEABLE` is recoverable and should be handled with `request.close()` while the Engine keeps running.

These are the reference scripts users copy from, so the divergence is likely to read as an API inconsistency. Either apply the same `REQUEST_UNSERVICEABLE` carve-out here, or add a one-line comment explaining why single-conversation samples deliberately treat it as fatal.

tianleiwu (commented):
## Review: Harden Engine fatal-event publication and abandonment recovery

Overall the direction is right. Splitting `fatal_events_` out of `pending_events_`/`staged_events_` fixes a real hole: the old `pending_events_.reserve(tracked + 1)` guarantee in `CreateRequest` was silently destroyed by the `pending_events_.swap(staged_events_)` at every committed step, so "terminalization never allocates" was not actually true. Preserving undrained events instead of erasing them, merging the terminal failure into the *last* matching retained event (`rbegin`/`rend`), publishing `health_`/`fatal_error_` only after `fatal_events_` is fully built, and separating transient `std::bad_alloc` from ownership-invariant violations during abandonment cleanup are all good changes. The new tests are targeted and meaningful.

I worked through the `fatal_events_` capacity proof and it holds: worst case is `max_step_event_count_` retained step events, plus at most one extra retained event per canceled tracked Request (`CancelRequest` push_back), plus one terminal event per executable tracked Request. Canceled and executable sets are disjoint, so the bound is `max_step_event_count_ + tracked + 1` against the reserved `max_step_event_count_ + tracked + 2`.

### Main items

1. **`fatal_events_` pins every pre-fatal Request for the Engine's lifetime** (inline on `engine.cpp`). After `pending_events_.swap(fatal_events_)`, `fatal_events_` holds the old pending contents — including the drained prefix and the events deliberately filtered out for abandoned Requests — and is never cleared again because `MarkUnhealthyAndThrow` rethrows at the top on every later entry. A one-line `fatal_events_.clear()` after the swap fixes it.
2. **The new outer catch in `HandleContinuationRestoreFailure` can skip the removal sequence** (inline). If building the diagnostic string throws, `CloseRequest` never runs but `CompleteCloseFromEngine` does, leaving a scheduler entry and undrained events pointing at a Request whose runtime state was already released.
3. **The `bad_alloc` retry rests on an unenforced ordering invariant** (inline): it only works because the sole realistic allocating step in `CloseRequest` runs strictly before `MarkClosedFromEngine()`. If that ever changes, the Request becomes permanently unreclaimable.

### Smaller items

- `EngineStepErrorFactory` in `EngineDependencies` is a fault-injection seam used only by tests; worth marking as such.
- Overflow guard in `CreateRequest` can underflow in a corner case.
- `model-qa.py`, `tool-calling.py`, and `benchmark_engine_guidance.py` now reject **every** `FAILED` event, including the recoverable request-scoped `REQUEST_UNSERVICEABLE`, while `continuous-batching.py` and the new doc sample teach the opposite. These are the scripts users copy from.
- Doc nits inline.

### Test gap (not anchorable to the diff)

No test asserts that Request objects are actually *released* after a fatal failure is fully drained — which is precisely why item 1 is invisible today. A `std::weak_ptr` expiry check after draining all fatal events and destroying the handles would catch it and would also guard the abandoned-identity filter.

Also worth noting: `CreateRequest` and `BeginTurn` call `ReclaimAbandonedRequests()` *without* the new try/catch that `Run()` and `HasPendingRequests()` use, so a fatal there surfaces as `EngineStepError` from those APIs while the terminal events stay retained for the next `Run()`. That looks intentional, but it is neither documented nor tested.

src/engine/engine.cpp
baijumeswani: Addressed. `MarkUnhealthyAndThrow` now clears `fatal_events_` immediately after swapping it into `pending_events_`, so the old pending contents and their strong Request references are released before the Engine becomes permanently unhealthy. The focused regression `FatalScratchReleasesDrainedRequestEvents` covers this ownership case.

src/engine/engine.cpp
baijumeswani: Addressed. Continuation restore failure now marks the Request failed and runs `CloseRequest` before any allocation-prone diagnostic construction. If close itself fails, the no-throw teardown path detaches scheduler/cache/event ownership and removes the Request from tracking before fatal publication.

src/engine/engine.cpp
baijumeswani: Addressed. The cleanup sequence keeps every potentially throwing removal operation before `MarkClosedFromEngine`, and an explicit comment at that transition documents that everything afterward must remain allocation-free and non-throwing. The abandonment retry therefore cannot observe a logically closed but still-resident Request after an allocation failure.

src/engine/engine.cpp
baijumeswani: Addressed. The capacity proof now uses subtraction-based, non-wrapping checks guarded by `fatal_events_.max_size() < kFatalEventOverhead`. Commit `c63c38b87b` also replaces the last literal `2` in the reserve expression with the shared `kFatalEventOverhead` constant.

src/engine/engine.h
baijumeswani: Addressed. `EngineDependencies::make_step_error` now has an explicit comment identifying it as test-only fault injection for allocation-sensitive durable error construction.

docs/engine_c_api_spec.md
baijumeswani: Fixed. The lifecycle text now says the static runtime state is retained until the shared physical batch “is recycled.”

docs/engine_c_api_spec.md
baijumeswani: Addressed. The construction snippet now includes `fatal_events_.reserve(max_step_events + 1);`, matching the dedicated fatal-event capacity described below it and the implementation.

docs/paged_attention_engine.md
baijumeswani: Addressed. The documentation now states that externally abandoned Requests are omitted on a best-effort basis at publication time and explicitly notes that final-handle release can race with that snapshot.

examples/python/engine/model-qa.py
baijumeswani: Addressed by documenting the intentional host-policy distinction in all three single-workload drivers. `model-qa.py` and `tool-calling.py` cannot continue their sole conversation after any failed turn, and the guidance benchmark treats any failed workload Request as an invalid run, even when `REQUEST_UNSERVICEABLE` leaves the shared Engine healthy. The multi-request `continuous-batching.py` sample retains the recoverable close-one-request-and-continue behavior.

tianleiwu (approved):

PR #2509, opened 2026-08-31
Add model-independent decoded stop-string matching

baijumeswani:
## Summary

- Add an incremental exact-byte matcher for decoded UTF-8 stop strings.
- Preserve deterministic caller indices across overlapping and duplicate stop strings.
- Separate immediately publishable output from a bounded suffix that could still complete a match.
- Validate UTF-8 and enforce bounded stop-string configuration.
- Cover chunking, overlap, reset, UTF-8, limits, and randomized oracle parity without requiring a model.

## Motivation

Decoded stop strings can span token boundaries or end inside a token's decoded bytes. Generator and Engine integrations therefore need a shared matcher that operates on the incremental decoded byte stream instead of matching each token's text independently.

This change intentionally adds only model-independent internal infrastructure. It does not change the public API or enable the existing Engine stop-option stubs. Generator and transactional Engine integration will follow in separate PRs.

## Behavior

- Matching is exact UTF-8 byte matching with no normalization, trimming, or case folding.
- Arbitrary input chunking does not affect the result.
- Match selection is deterministic: earliest ending byte, then earliest start/longest match, then lowest caller index.
- Duplicate stop strings remain distinct configuration entries.
- A completed match remains sticky until the matcher is reset.
- Bytes that cannot participate in a future match are released as safe output.
- The retained suffix is bounded by the longest configured stop string minus one byte between calls.
- Empty or malformed UTF-8 strings and configurations above 16 strings or 16 KiB total are rejected.

## Testing

- `StopStringMatcherTest.*`: 27 passed.
- Complete CPU unit suite: 263 passed, 22 skipped.
- Lintrunner: clean.
- Independent randomized audit: more than 27,000 configurations and streams with no oracle divergence.
- Three independent code reviews completed with no findings.

### DIFF
--- a/src/stop_string_matcher.cpp
+++ b/src/stop_string_matcher.cpp
@@ -1,0 +1,159 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#include "stop_string_matcher.h"
+
+#include <algorithm>
+#include <stdexcept>
+
+namespace Generators {
+
+namespace {
+
+size_t Utf8SequenceLength(unsigned char lead) {
+  if (lead < 0x80) return 1;
+  if (lead >= 0xC2 && lead <= 0xDF) return 2;
+  if (lead >= 0xE0 && lead <= 0xEF) return 3;
+  if (lead >= 0xF0 && lead <= 0xF4) return 4;
+  return 0;  // Continuation byte, overlong two-byte lead (0xC0/0xC1), or out of range (>= 0xF5).
+}
+
+}  // namespace
+
+bool IsValidUtf8(std::string_view bytes) {
+  size_t i = 0;
+  while (i < bytes.size()) {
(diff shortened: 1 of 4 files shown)


src/stop_string_matcher.cpp
```diff
@@ -0,0 +1,159 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
kunal-vaishnavi, line 1: Couple of questions:

- Can we explain this file's purpose at the top of this file similar to how there's a long description in front of `StopStringMatcher` in `src/stop_string_matcher.h`?
- Can we also update the PR description to explain the usage difference between this and grammars? 
- How will a user activate this?

baijumeswani:
Superseded by https://github.com/microsoft/onnxruntime-genai/pull/2527.
Closing this,

Issue #1989, opened 2026-02-24
[copilot] Qwen3-VL Support


justinchuby:
[copilot] Feature Request: Native Qwen3-VL Support in onnxruntime-genai
================================================================

Summary
-------
Add native support for Qwen3-VL (Qwen/Qwen3-VL-2B-Instruct, Qwen/Qwen3-VL-8B-Instruct)
as a vision-language model in onnxruntime-genai. Currently, Qwen3-VL can run using the
"qwen2_5_vl" model type as a workaround (the 3-model I/O contract is identical for
text-only generation), but native support would enable proper image processing and
model-type-specific optimizations.

Motivation
----------
Qwen3-VL is the latest vision-language model from the Qwen series. We have verified that
the 3-model split (decoder + vision encoder + embedding) works end-to-end with
onnxruntime-genai using the "qwen2_5_vl" model type, but this workaround has limitations:

1. The image processor (QwenImageProcessor) may not handle Qwen3-VL's different
   patch size (16x16 vs 14x14) and image preprocessing correctly.
2. The processor_config.json format may differ between Qwen2.5-VL and Qwen3-VL.
3. Model-type-specific optimizations and configuration cannot be applied.

Working Prototype
-----------------
We have a working prototype in the onnx-genai-models package
(https://github.com/justinchuby/onnx-genai-models) that exports Qwen3-VL
as a 3-model package compatible with onnxruntime-genai:

  - **model.onnx** (decoder): inputs_embeds, attention_mask, position_ids[3,B,S],
    past_kv → logits, present_kv
  - **vision.onnx**: pixel_values[N, C*T_p*P*P], cu_seqlens, rotary_pos_ids,
    grid_thw → image_features
  - **embedding.onnx**: input_ids, image_features → inputs_embeds

Text-only generation works by loading with model.type = "qwen2_5_vl" in
genai_config.json. The model generates correct, coherent text.

Architectural Differences from Qwen2.5-VL
------------------------------------------
| Feature              | Qwen2.5-VL          | Qwen3-VL              |
|----------------------|----------------------|------------------------|
| Patch size           | 14×14                | 16×16                  |
| Vision attention     | Windowed + full-att  | Packed (block-diag)    |
| Vision MLP           | Gated (SiLU)         | Standard (GELU)        |
| Vision normalization | RMSNorm              | LayerNorm              |
| Feature fusion       | Single MLP merger    | DeepStack (multi-layer)|
[...]

PR #2513, opened 2026-09-01
Support DeepStack (Qwen3-VL) in the vision->embedding->decoder pipeline

SanjayAMD:
## Summary

Adds **DeepStack** support for **Qwen3-VL** to the multi-model (vision → embedding → decoder) pipeline.

Qwen3-VL does not rely solely on a single merged vision embedding. Its vision encoder also emits a small number of **intermediate hidden-state feature maps** ("DeepStack" features, one per entry in `deepstack_visual_indexes` — 3 for the 4B model, from ViT layers `[5, 11, 17]`). These are scattered back onto the image-token positions and **added into the decoder residual stream after the first few decoder layers** (feature `i` → after layer `i`), matching Hugging Face's `Qwen3VLTextModel._deepstack_process`.

Without this, the model keeps only the final merged vision embedding and loses the fine-grained visual detail DeepStack carries, which significantly degrades OCR/grounding accuracy. This PR wires DeepStack end to end through config parsing, the C++ runtime, and the ONNX model builder.

## Motivation

DeepStack is required for Qwen3-VL to reach its expected accuracy on detail-sensitive tasks (OCR, document/diagram understanding, grounding). The existing Qwen3-VL path builds and runs, but silently drops the DeepStack features, so accuracy is well below the reference PyTorch model. This change closes that gap.

Design reference: microsoft/onnxruntime-genai#1989.

## How it works

The three sub-models exchange DeepStack tensors as follows:

```
vision.onnx      ->  deepstack_features_{i}   (per-token, image-token length)
embedding.onnx   ->  scatters them to full (batch, seq, hidden) -> deepstack_{i}
decoder.onnx     ->  adds deepstack_{i} into the residual after layer i
```

- The **scatter** (place per-token features at `input_ids == image_token` positions) happens inside `embedding.onnx`, so the runtime only has to move buffers, not compute indices.
- During **generation** there are no image tokens, so `embedding.onnx` emits all-zeros for `deepstack_{i}` and the decoder's per-layer `Add`s become no-ops. No special-casing in the runtime.

## Changes

**`src/config.h` / `src/config.cpp`** — parse the DeepStack tensor-name arrays from `genai_config.json`:
- `model.vision.outputs.deepstack_features`
- `model.embedding.inputs.deepstack_features`
- `model.embedding.outputs.deepstack`
- `model.decoder.inputs.deepstack`

Each is parsed via an `OnArray` hook backed by `StringArray_Element`. `StringArray_Element` is moved earlier in the file so `DecoderInputs_Element` (declared before its previous definition) can use it.

**`src/models/multi_modal.{h,cpp}`** — runtime wiring:
- `VisionState`: allocate the extra DeepStack outputs (`outputs_[1..N]`) and, in the `QwenVisionState` per-image loop, slice them at the same offsets as `image_features` and restore them after the loop.
- `EmbeddingState`: allocate the per-token DeepStack inputs (from vision) and the full-length DeepStack outputs (to decoder); update them each step.
- `DecoderState`: allocate the full-length DeepStack inputs (owning buffers) and size them alongside `inputs_embeds` / `per_layer_inputs`, including the **prefill-chunking** path (`PrepareEmbeddingsForPrefill` / `RunPrefillWithChunking`).
- `MultiModalPipelineState::Run`: bind vision → embedding (`ReuseFeaturesBuffer`) and embedding → decoder (`ReuseEmbeddingsBuffer`) DeepStack buffers in both the prompt and generation paths.

**`src/python/py/models/builders/qwen.py`** — `Qwen3VLTextModel`:
- Declares `deepstack_{i}` decoder graph inputs of shape `[batch, seq, hidden]` (count from `vision_config.deepstack_visual_indexes`, defaulting to 3).
- Overrides `make_layer` to add `deepstack_{layer_id}` into the layer's residual (`skip_input`) for `layer_id < num_deepstack`, so the next layer's `SkipLayerNorm` folds the feature into the residual — exactly HF's "add visual features after layer i".

## Validation

Validated on **Qwen3-VL-4B** (INT4 AWQ vision-fp32 / text) exported as the 3-model split:

- **End-to-end runtime smoke test**: the full vision → embedding → decoder DeepStack path runs and produces correct, grounded output (e.g. an image of a cat yields *"A fluffy Ragdoll cat with striking blue eyes and a white and brown coat rests peacefully, gazing directly at the camera…"*).
- **Accuracy (AI2D, 100 samples)**: **81%** (ONNX INT4) vs **83%** (PyTorch FP32) — a −2 pp gap that is within run-to-run noise, confirming no accuracy regression from the ONNX/runtime path.

## Build / test

- Builds cleanly (DirectML config): C++ compiles and links, Python bindings build.
- `src/python/py/models/builders/qwen.py` passes byte-compilation and existing lint.

## Scope / notes

- Feature is fully **opt-in**: with no `deepstack*` entries in `genai_config.json`, all the new loops iterate zero times and behavior is unchanged for every other model.
- The genai_config `deepstack*` arrays are produced by the export/optimize tooling; this PR is the runtime + builder side that consumes and emits them.
- DeepStack is also wired through the prefill-chunking path so it stays correct when `search.chunk_size` is set.

### DIFF
--- a/src/models/multi_modal.cpp
+++ b/src/models/multi_modal.cpp
@@ -953,6 +953,12 @@ DeviceSpan<float> MultiModalPipelineState::Run(int current_length, DeviceSpan<in
     if (vision_state_) {
       embedding_state_->image_features_->ReuseFeaturesBuffer(*vision_state_->image_features_);
       // Qwen3-VL DeepStack: hand each vision per-token output to the matching embedding input.
+      if (embedding_state_->deepstack_features_in_.size() != vision_state_->deepstack_features_.size()) {
+        throw std::runtime_error("DeepStack config mismatch: embedding.inputs.deepstack_features count (" +
+                                 std::to_string(embedding_state_->deepstack_features_in_.size()) +
+                                 ") must match vision.outputs.deepstack_features count (" +
+                                 std::to_string(vision_state_->deepstack_features_.size()) + ").");
+      }
       for (size_t k = 0; k < embedding_state_->deepstack_features_in_.size(); ++k) {
         embedding_state_->deepstack_features_in_[k]->ReuseFeaturesBuffer(*vision_state_->deepstack_features_[k]);
       }
@@ -975,6 +981,12 @@ DeviceSpan<float> MultiModalPipelineState::Run(int current_length, DeviceSpan<in
       embedding_state_->per_layer_inputs_->ReuseEmbeddingsBuffer(*decoder_state_->per_layer_inputs_);
     }
     // Qwen3-VL DeepStack: bind embedding's full-length scattered outputs to the decoder's inputs.
+    if (embedding_state_->deepstack_out_.size() != decoder_state_->deepstack_in_.size()) {
+      throw std::runtime_error("DeepStack config mismatch: embedding.outputs.deepstack count (" +
+                               std::to_string(embedding_state_->deepstack_out_.size()) +
+                               ") must match decoder.inputs.deepstack count (" +
+                               std::to_string(decoder_state_->deepstack_in_.size()) + ").");
+    }
     for (size_t k = 0; k < embedding_state_->deepstack_out_.size(); ++k) {
       embedding_state_->deepstack_out_[k]->ReuseEmbeddingsBuffer(*decoder_state_->deepstack_in_[k]);
     }
@@ -997,6 +1009,12 @@ DeviceSpan<float> MultiModalPipelineState::Run(int current_length, DeviceSpan<in
   }
   // Qwen3-VL DeepStack: rebind embedding's full-length outputs (all zeros in generation, no image
   // tokens) to the decoder's inputs so the per-layer Adds become no-ops.
+  if (embedding_state_->deepstack_out_.size() != decoder_state_->deepstack_in_.size()) {
+    throw std::runtime_error("DeepStack config mismatch: embedding.outputs.deepstack count (" +
+                             std::to_string(embedding_state_->deepstack_out_.size()) +
+                             ") must match decoder.inputs.deepstack count (" +
+                             std::to_string(decoder_state_->deepstack_in_.size()) + ").");
+  }
   for (size_t k = 0; k < embedding_state_->deepstack_out_.size(); ++k) {
     embedding_state_->deepstack_out_[k]->ReuseEmbeddingsBuffer(*decoder_state_->deepstack_in_[k]);
   }
--- a/src/models/multi_modal.cpp
+++ b/src/models/multi_modal.cpp
@@ -142,6 +142,14 @@ void VisionState::SetExtraInputs(const std::vector<ExtraInput>& extra_inputs, co
                                                          model_.config_->model.vision.outputs.image_features,
                                                          num_images_, num_image_tokens_);
   image_features_->Add();
+  // Qwen3-VL DeepStack: additional per-token vision outputs (outputs_[1..N]); the multi-image
+  // slicing loop in QwenVisionState::Run relies on these following image_features (outputs_[0]).
+  for (const auto& ds_name : model_.config_->model.vision.outputs.deepstack_features) {
+    auto ds = std::make_unique<MultiModalFeatures>(*this, MultiModalFeatures::Mode::Output,
+                                                   ds_name, num_images_, num_image_tokens_);
+    ds->Add();
+    deepstack_features_.push_back(std::move(ds));
+  }
   extra_inputs_.Add(extra_inputs, model_.vision_session_->GetInputNames());
 }
 
@@ -246,6 +254,16 @@ DeviceSpan<float> QwenVisionState::Run(int current_length, DeviceSpan<int32_t>& 
   OrtValue* pv_full = inputs_[pv_idx];
   OrtValue* feat_full = outputs_[0];  // pre-allocated image_features output
 
+  // Qwen3-VL DeepStack outputs (outputs_[1..N]) share image_features' shape/type and are sliced
+  // identically per image. Capture their full buffers so we can restore them after the loop.
+  const size_t num_deepstack = deepstack_features_.size();
+  std::vector<OrtValue*> ds_full(num_deepstack);
+  std::vector<void*> ds_raw(num_deepstack);
+  for (size_t k = 0; k < num_deepstack; ++k) {
+    ds_full[k] = outputs_[1 + k];
+    ds_raw[k] = ds_full[k]->GetTensorMutableRawData();
+  }
+
   // Shapes: pixel_values[total_patches, patch_dim], image_features[total_logical_patches, hidden_size]
   auto pv_info = pv_full->GetTensorTypeAndShapeInfo();
   auto feat_info = feat_full->GetTensorTypeAndShapeInfo();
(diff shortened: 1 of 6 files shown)


src/python/py/models/builders/qwen.py
```diff
@@ -83,6 +83,43 @@ def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
...
+        # DeepStack: the vision encoder emits len(deepstack_visual_indexes) extra feature
+        # tensors (3 for the 4B model, from vision layers [5, 11, 17]). These are scattered to
+        # full length (batch, seq, hidden) by embedding.onnx and injected into the residual
+        # stream after decoder layers 0/1/2 (feature i -> layer i), matching HF's
+        # Qwen3VLTextModel._deepstack_process. We declare them as extra decoder inputs and add
+        # them in make_layer. During generation there are no image tokens, so embedding.onnx
+        # emits all-zeros and the Adds become no-ops.
+        vision_config = getattr(config, "vision_config", None)
+        ds_indexes = getattr(vision_config, "deepstack_visual_indexes", None) if vision_config else None
+        self.num_deepstack = len(ds_indexes) if ds_indexes else 3
```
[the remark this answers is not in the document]
SanjayAMD: Good catch — fixed. num_deepstack now defaults to 0 when vision_config.deepstack_visual_indexes is missing or empty, so DeepStack is opt-in from the model config and a config/tooling that doesn't declare DeepStack won't get mandatory extra decoder inputs:

vision_config = getattr(config, "vision_config", None)
ds_indexes = getattr(vision_config, "deepstack_visual_indexes", None) if vision_config else None
self.num_deepstack = len(ds_indexes) if ds_indexes is not None else 0
With num_deepstack == 0, both make_inputs_and_outputs and make_layer no-op, producing the plain Qwen3-VL decoder graph. Added a builder test (test_qwen3_vl_deepstack_count_follows_vision_config) covering missing/None/empty → 0 and [5, 11, 17] → 3

src/models/multi_modal.cpp
```diff
@@ -892,6 +952,10 @@ DeviceSpan<float> MultiModalPipelineState::Run(int current_length, DeviceSpan<in
     }
     if (vision_state_) {
       embedding_state_->image_features_->ReuseFeaturesBuffer(*vision_state_->image_features_);
+      // Qwen3-VL DeepStack: hand each vision per-token output to the matching embedding input.
+      for (size_t k = 0; k < embedding_state_->deepstack_features_in_.size(); ++k) {
+        embedding_state_->deepstack_features_in_[k]->ReuseFeaturesBuffer(*vision_state_->deepstack_features_[k]);
+      }
```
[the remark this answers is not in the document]
SanjayAMD: Agreed — added explicit fail-fast checks before all three DeepStack binding loops so an inconsistent config errors clearly instead of indexing out of bounds. For the vision→embedding binding:

if (embedding_state_->deepstack_features_in_.size() != vision_state_->deepstack_features_.size()) {
  throw std::runtime_error("DeepStack config mismatch: embedding.inputs.deepstack_features count (" +
                           std::to_string(embedding_state_->deepstack_features_in_.size()) +
                           ") must match vision.outputs.deepstack_features count (" +
                           std::to_string(vision_state_->deepstack_features_.size()) + ").");
}
The two embedding→decoder bindings (prompt path and generation path) get the equivalent guard comparing embedding.outputs.deepstack against decoder.inputs.deepstack. Error messages include the actual counts to make misconfiguration easy to diagnose.

VishalX:
@kunal-vaishnavi / @baijumeswani pls review this.

VishalX:
olive-recipe update: https://github.com/microsoft/olive-recipes/pull/601

src/models/multi_modal.cpp
```diff
@@ -142,6 +142,14 @@ void VisionState::SetExtraInputs(const std::vector<ExtraInput>& extra_inputs, co
                                                          model_.config_->model.vision.outputs.image_features,
                                                          num_images_, num_image_tokens_);
   image_features_->Add();
+  // Qwen3-VL DeepStack: additional per-token vision outputs (outputs_[1..N]); the multi-image
```
kunal-vaishnavi, line 145: Can the DeepStack features be captured using `extra_outputs_` instead of being explicitly mentioned here? They are applicable for other models besides Qwen-3 VL as well. We can update `extra_outputs_` to be the counterpart to `extra_inputs_` for enabling this.

src/models/multi_modal.cpp
```diff
@@ -630,6 +664,19 @@ void EmbeddingState::SetExtraInputs(const int64_t num_images, const int64_t num_
                                                            num_images, num_image_tokens_);
     image_features_->Add();
   }
+  // Qwen3-VL DeepStack: per-token feature inputs (from vision) and full-length scattered outputs
+  // (to decoder). Both are declared per deepstack_visual_index in genai_config.
+  for (const auto& ds_name : model_.config_->model.embedding.inputs.deepstack_features) {
```
kunal-vaishnavi, line 678: With [this change](https://github.com/microsoft/onnxruntime-genai/pull/2513/#discussion_r3927208174), we can let `extra_inputs_` be used here instead.

src/models/multi_modal.cpp
```diff
@@ -668,6 +716,14 @@ DecoderState::DecoderState(const MultiModalLanguageModel& model, DeviceSpan<int3
       recurrent_state_{CreateRecurrentState(*this)} {
   inputs_embeds_.Add();
 
+  // Qwen3-VL DeepStack: full-length feature inputs injected after decoder layers 0/1/2. Allocated
+  // here (owning buffer); the embedding model's matching outputs reuse these buffers each step.
+  for (const auto& ds_name : model_.config_->model.decoder.inputs.deepstack) {
```
kunal-vaishnavi: Same comment as [this one](https://github.com/microsoft/onnxruntime-genai/pull/2513/#discussion_r3927567454)

src/models/multi_modal.cpp
```diff
@@ -927,6 +1007,17 @@ DeviceSpan<float> MultiModalPipelineState::Run(int current_length, DeviceSpan<in
   if (embedding_state_->per_layer_inputs_ && decoder_state_->per_layer_inputs_) {
     embedding_state_->per_layer_inputs_->ReuseEmbeddingsBuffer(*decoder_state_->per_layer_inputs_);
   }
+  // Qwen3-VL DeepStack: rebind embedding's full-length outputs (all zeros in generation, no image
```
kunal-vaishnavi: For any inputs that aren't specified during decoding, can we set their values to zeros by default and let `extra_inputs_` handle the input bindings?

src/models/multi_modal.h
```diff
@@ -168,6 +171,10 @@ struct EmbeddingState : State {
   Embeddings inputs_embeds_{*this, Embeddings::Mode::Output,  // Model output
                             model_.config_->model.embedding.outputs.embeddings};
   std::unique_ptr<Embeddings> per_layer_inputs_;  // Optional model output (Gemma4)
```
kunal-vaishnavi: Actually looking at this some more, there already exists `per_layer_inputs` for inputs to a text decoder. Is it possible to re-use this or re-architect its usage so it works for both Gemma-4 and Qwen-3 VL?

src/models/multi_modal.cpp
SanjayAMD: Hey @kunal-vaishnavi 
extra_outputs_ doesn't do that buffer reuse today, so moving them there would mean an extra copy
so,  kept vision DeepStack outputs on the MultiModalFeatures path (same as image_features) rather than extra_outputs_

Currently, image_feature + deepStack outputs are handed to the embedding model via zero‑copy ReuseFeaturesBuffer

src/models/multi_modal.cpp
SanjayAMD: the embedding's DeepStack inputs are the per‑token MultiModalFeatures handed over from vision via ReuseFeaturesBuffer
So I think its better to Keep them on that path preserves the zero‑copy hand‑off

src/models/multi_modal.cpp
SanjayAMD: Hey @kunal-vaishnavi 

That already happens for free with the unified path (below per_layer_inputs way you suggested). During decode there are no image tokens, so the embedding model outputs zeros for the deepstack tensors and the decoder's Adds are no-ops — same as what per_layer_inputs does.

src/models/multi_modal.h
SanjayAMD: Done. DeepStack is structurally the plural of per_layer_inputs, so I unified them: EmbeddingState and DecoderState each hold one ordered Embeddings collection (aux_feature_outputs_ / aux_feature_inputs_) carrying Gemma‑4's per_layer_inputs (1 entry) and Qwen3‑VL DeepStack (one per deepstack_visual_index),

src/models/multi_modal.cpp
SanjayAMD: These come from the vision model via a zero-copy ReuseFeaturesBuffer, same as image_features above — extra_inputs_ doesn't do that buffer hand-off (it binds a provided tensor or zeros), so moving them there would force a copy. Kept them on MultiModalFeatures for that reason. The decoder-side deepstack inputs I did fold into the shared per_layer_inputs path per your other comment.

PR #2514, opened 2026-09-01
stamps onnx model files with genai commit

prathikr:
GenAI commit id used to build onnx models via model builder can now be surfaced by inspecting the onnx model:

```python
import onnx

model = onnx.load(r'C:\path\to\model.onnx', load_external_data=False)

print('producer:', model.producer_name)
print('GenAI version:', model.producer_version)
```

Expected Output:
```
producer: onnxruntime-genai
GenAI version: 0.16.0-dev+3ca2c97
```

### DIFF
--- a/src/python/py/models/builders/base.py
+++ b/src/python/py/models/builders/base.py
@@ -1639,19 +1639,23 @@ class Model:
     @classmethod
     def get_genai_version(cls) -> str | None:
         """Return the GenAI package version when source metadata is available."""
-        try:
-            # Attempt to read the version information from the VERSION_INFO file in the repository root.
-            repo_root = os.path.abspath(os.path.join(os.path.dirname(__file__), "..", "..", "..", "..", ".."))
-            with open(os.path.join(repo_root, "VERSION_INFO"), encoding="utf-8") as version_info:
-                return version_info.read().strip()
-        except OSError:
+        # Attempt to read the version information from the VERSION_INFO file in the repository root.
+        repo_root = os.path.abspath(os.path.join(os.path.dirname(__file__), "..", "..", "..", "..", ".."))
+        version_info_path = os.path.join(repo_root, "VERSION_INFO")
+        if os.path.exists(version_info_path):
             try:
-                # If VERSION_INFO file is missing, we are installing from whl package. Use the version from the package metadata.
-                from onnxruntime_genai import __version__
+                with open(version_info_path, encoding="utf-8") as version_info:
+                    return version_info.read().strip()
+            except OSError:
+                pass
 
-                return __version__ or None
-            except ImportError:
-                return None
+        try:
+            # If VERSION_INFO file is missing, we are installing from whl package. Use the version from the package metadata.
+            from onnxruntime_genai import __version__
+
+            return __version__ or None
+        except ImportError:
+            return None
 
     @classmethod
     def get_genai_commit(cls) -> str | None:
--- a/src/python/py/models/builders/base.py
+++ b/src/python/py/models/builders/base.py
@@ -11,7 +11,6 @@ from __future__ import annotations
 import ast
 import json
 import os
-from pathlib import Path
 import subprocess
 from collections.abc import Sequence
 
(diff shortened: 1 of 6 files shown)


src/python/py/models/builders/base.py
```diff
@@ -40,6 +43,29 @@
 from quantization import CudaQuantizer, QuantConfig, resolve_dtype
 
 
+@functools.cache
+def get_genai_commit() -> str | None:
```
kunal-vaishnavi: Can we add these methods as part of the `Model` class rather than standalone methods?

src/python/py/models/builders/base.py
```diff
@@ -40,6 +43,29 @@
...
+    """Return the current GenAI source revision when Git metadata is available."""
+    repo_root = Path(__file__).resolve().parents[5]
+    try:
+        result = subprocess.run(
+            ["git", "rev-parse", "HEAD"],
+            cwd=repo_root,
+            check=True,
+            capture_output=True,
+            text=True,
+        )
```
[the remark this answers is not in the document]
kunal-vaishnavi: This is the same issue that we were discussing offline.

src/python/py/models/builders/base.py
```diff
@@ -40,6 +43,29 @@
...
+        return result.stdout.strip()
+    except (OSError, subprocess.CalledProcessError):
+        return None
+
+
+def stamp_build_metadata(model: ir.Model) -> None:
+    """Add the GenAI source revision to an exported ONNX graph."""
+    if commit := get_genai_commit():
+        model.producer_version = commit
```
edgchen1: it seems like a field named "producer_version" should contain a version string, i.e., the onnxruntime-genai version. for a dev version, we could add the commit as part of the version string (e.g., semver build metadata).

or we could name the field something like "producer_commit".

src/python/py/models/builders/base.py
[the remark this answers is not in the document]
prathikr: none for .whl based builds is ok

edgchen1: would this only work when running builder.py from a git repo? could we also populate this info when running from a wheel?

test/python/builder/test_precision.py
```diff
@@ -464,6 +465,70 @@ def save_processing(self, *args):
     assert captured["processing"][0] == "fake-model"
 
 
+def _make_empty_onnx_model():
+    graph = ir.Graph(inputs=(), outputs=(), nodes=(), opset_imports={"": 22})
+    return ir.Model(graph, ir_version=10, producer_name="onnxruntime-genai")
+
+
+def test_stamp_build_metadata_sets_genai_version_and_commit(monkeypatch):
```
edgchen1: nit: is there a better location for these tests? maybe a new test file?

this test file's comment seems to indicate it is more specialized:
https://github.com/microsoft/onnxruntime-genai/blob/b19542e1aa3a90299cfb820fc156ab8d9d2bf8b7/test/python/builder/test_precision.py#L4-L9

edgchen1 (dismissed):

src/python/py/models/builders/base.py
```diff
@@ -41,6 +43,49 @@
 
 
 class Model:
+    def get_genai_version(self) -> str | None:
+        """Return the GenAI package version when source metadata is available."""
+        try:
+            # Attempt to read the version information from the VERSION_INFO file in the repository root.
+            return (Path(__file__).resolve().parents[5] / "VERSION_INFO").read_text(encoding="utf-8").strip()
```
kunal-vaishnavi: Can we use `os.path` methods to stay consistent with the existing codebase and reduce the number of new imported packages we are adding?

src/python/py/models/builders/base.py
```diff
@@ -1670,6 +1715,7 @@ def callback(tensor: ir.TensorProtocol, metadata: dict):
                 pbar.update()
                 pbar.set_description(f"Saving {tensor.name} ({tensor.dtype.short_name()}, {tensor.shape})")
 
+            self.stamp_build_metadata(model)
```
kunal-vaishnavi: Can we move these new methods to be here instead before `save_model` since they are related to when the model is saved? The model builder codebase is quite large so it would be good to keep methods that are calling each other to be close to one another.

edgchen1 (dismissed):

kunal-vaishnavi (dismissed):

kunal-vaishnavi (approved):

PR #2515, opened 2026-09-01
Restore legacy recurrent state binding fallback

jiafatom:
## Summary
- restore the legacy Qwen-style convolution and recurrent state tensor names as the decoder defaults
- preserve explicit state bindings emitted by current model builders
- add direct config coverage and a separate legacy CUDA fixture that executes a first decode

## Background
PR #2386 introduced configurable fixed-state bindings with generic `past.%d.*` defaults. Published Qwen 3.5 packages created before that change omit these fields because the earlier runtime derived `conv_state` and `recurrent_state` names from the KV-cache templates. Those packages therefore stopped binding their recurrent state inputs.

The current builders already emit explicit fixed-state mappings for models that use different names, so restoring the legacy Qwen-compatible defaults fixes old packages without changing modern generated configs.

## Validation
- CUDA 12.8 Release build succeeded
- `DecoderStateGroupsConfigTest.UsesLegacyRecurrentStateBindingDefaults` passed
- `ModelTests.LegacyRecurrentStateBindingDefaultsRunFirstDecode` passed
- `ModelTests.RecurrentStateCudaGraphCaptureMatchesEager` passed with the restored explicit-binding fixture
- modern explicit-binding Qwen 3.5 CUDA tests passed (2/2)
- real legacy Qwen 3.5 package smoke test passed model creation, token append, and first generation


src/models/io/recurrent_state.cpp
```diff
@@ -23,23 +23,51 @@ std::string ComposeCacheName(const std::string& template_string, int index) {
...
+    }
+    return indices;
+  };
+
+  layer_indices_ = discover_layers(past_recurrent_template);
+  const bool uses_default_state_templates =
+      past_conv_template == Config::Defaults::PastConvName &&
+      past_recurrent_template == Config::Defaults::PastRecurrentName &&
+      present_conv_template == Config::Defaults::PresentConvName &&
+      present_recurrent_template == Config::Defaults::PresentRecurrentName;
```
baijumeswani: The current builder explicitly writes these same default values, so comparing values cannot tell whether the fields were omitted. Normal new models still work because discovery succeeds first, but a broken or renamed model could fall back even though it provided explicit bindings. Consider recording whether these fields were present while parsing the config, and only use the legacy fallback when they were truly omitted.

src/models/io/recurrent_state.cpp
```diff
@@ -23,23 +23,51 @@ std::string ComposeCacheName(const std::string& template_string, int index) {
...
+    auto derived_past_recurrent =
+        derive_template(model_.config_->model.decoder.inputs.past_key_names, "recurrent_state");
+    auto derived_indices = discover_layers(derived_past_recurrent);
+    if (!derived_indices.empty()) {
+      layer_indices_ = std::move(derived_indices);
+      past_conv_template = derive_template(model_.config_->model.decoder.inputs.past_key_names, "conv_state");
+      past_recurrent_template = std::move(derived_past_recurrent);
+      present_conv_template = derive_template(model_.config_->model.decoder.outputs.present_key_names, "conv_state");
+      present_recurrent_template =
+          derive_template(model_.config_->model.decoder.outputs.present_key_names, "recurrent_state");
```
baijumeswani: This accepts the fallback after finding only the recurrent input names, then assumes the matching conv inputs and both outputs exist. Consider checking all four derived bindings here and reporting the missing tensor directly. That would give users a much clearer error for partially compatible legacy models.

test/models/hf-internal-testing/tiny-qwen35-cuda/genai_config.json
```diff
@@ -28,16 +28,12 @@
                 "attention_mask": "attention_mask",
                 "position_ids": "position_ids",
                 "past_key_names": "past_key_values.%d.key",
-                "past_value_names": "past_key_values.%d.value",
-                "past_conv_names": "past_key_values.%d.conv_state",
-                "past_recurrent_names": "past_key_values.%d.recurrent_state"
+                "past_value_names": "past_key_values.%d.value"
```
baijumeswani: Making this shared fixture legacy means all recurrent-state CUDA tests now use the fallback, while `create_gqa_model.py` still regenerates it with explicit bindings. Consider keeping a separate legacy config or adding a generator option so both the legacy and modern explicit-binding paths stay reproducible and covered.

test/cpp/model_tests.cpp
```diff
@@ -269,6 +269,15 @@ std::vector<float> RunFirstDecodeWithCudaGraph(bool enable_cuda_graph) {
   return {logits, logits + 1000};
 }
 
+TEST(ModelTests, LegacyRecurrentStateBindingsAreDerivedFromKvNames) {
+  if (!std::filesystem::exists(c_cuda_graph_model_path)) {
+    GTEST_SKIP() << "Legacy recurrent-state model not available at: " << c_cuda_graph_model_path;
+  }
+
+  const auto logits = RunFirstDecodeWithCudaGraph(false);
+  EXPECT_EQ(logits.size(), 1000u);
```
baijumeswani, line 279: `RunFirstDecodeWithCudaGraph` always returns a 1,000-element vector, so this assertion mainly checks that the call did not throw. A stronger check would compare the legacy config with the same model using explicit bindings, or verify that recurrent state changes the logits on a later decode step.

baijumeswani (commented):
This fixes the legacy Generator regression, and the Engine path remains unchanged because it reads fixed-state bindings through the state manifest and FixedStatePool. The suggestions below are non-blocking, but they would make the compatibility behavior and tests clearer and safer.

test/models/hf-internal-testing/tiny-qwen35-cuda/genai_config.json
```diff
@@ -28,16 +28,12 @@
                 "attention_mask": "attention_mask",
                 "position_ids": "position_ids",
                 "past_key_names": "past_key_values.%d.key",
-                "past_value_names": "past_key_values.%d.value",
-                "past_conv_names": "past_key_values.%d.conv_state",
-                "past_recurrent_names": "past_key_values.%d.recurrent_state"
```
kunal-vaishnavi, line 33: Why are we removing this? It is intended to need to specify the past/present conv and recurrent names in the GenAI config and use those templates during processing.

src/models/io/recurrent_state.cpp
```diff
@@ -23,23 +23,51 @@ std::string ComposeCacheName(const std::string& template_string, int index) {
 
 RecurrentState::RecurrentState(State& state)
     : state_{state} {
-  // Discover recurrent layer indices by scanning all session input names
```
kunal-vaishnavi, line 26: Instead of all these changes, can we change the default for the past/present conv and recurrent names to be the default names that Qwen-3.5 uses? Ex:

https://github.com/microsoft/onnxruntime-genai/blob/d8ab08ee3f251c87172a656b095f7d2903bee792/src/config.h#L27-L28

That may fix the issue because the GenAI config for the published model doesn't have this entry so the default value inside GenAI should get used instead.

jiafatom: Implemented as suggested. The four decoder defaults now use the published Qwen 3.5 names, and current builders remain unaffected because they emit explicit mappings.

src/models/io/recurrent_state.cpp
jiafatom: Addressed by replacing value-based fallback detection with the suggested default-name fix. Explicit bindings remain authoritative because the runtime no longer performs fallback discovery.

test/models/hf-internal-testing/tiny-qwen35-cuda/genai_config.json
jiafatom: Restored. The shared tiny-qwen35-cuda fixture again specifies all four past/present conv and recurrent templates. Legacy behavior is covered by a separate fixture.

test/cpp/model_tests.cpp
jiafatom: Addressed. The dedicated legacy test now uses a separate fixture and executes the first decode. The existing explicit-binding fixture continues to run the eager-vs-CUDA-graph logits comparison.

test/models/hf-internal-testing/tiny-qwen35-cuda/genai_config.json
jiafatom: Addressed. The existing tiny-qwen35-cuda fixture is restored with explicit mappings. A separate tiny-qwen35-legacy-cuda fixture now covers omitted mappings.

src/models/io/recurrent_state.cpp
jiafatom: Addressed by removing the runtime fallback entirely. RecurrentState now uses the configured/default templates as before, so partial derived-binding validation is no longer needed.

baijumeswani (approved):

kunal-vaishnavi (approved):

PR #2518, opened 2026-09-02
Enable safe Engine request rewind

bmehta001:
## Description

Adds safe rewind support for completed Engine requests. A request can retain a token prefix, release its resident model state, and replay that prefix with the next turn without changing request identity or turn ordering.

## Changes

- Add `OgaRequestRewindToStartOfTurn`, C++ `OgaRequest::RewindToStartOfTurn`, and Python `Request.rewind_to_start_of_turn`.
- Rewind paged KV, fixed recurrent/convolution state, MTP shadow state, and sole-row static cache allocations safely.
- Restore search, guidance, prompt/generated accounting, CPU RNG, and CUDA batched-sampler boundaries for deterministic continuation.
- Integrate rewind checkpoints with chained MTP and sampled speculative verification transactions.
- Reject active, failed, closed, out-of-range, undrained-event, and static multi-row rewinds without mutation.
- Document the lifecycle contract and add native, C API, CUDA sampler, and Python coverage.

## Validation

- `lintrunner`
- Windows and WSL Engine unit suites
- Windows and WSL C API Engine lifecycle test
- Windows and WSL Python Engine suite

CUDA sampler restoration has dedicated tests and structural coverage, but the CUDA target was not run locally because no CUDA toolkit or GPU was available.


src/engine/engine.cpp
```diff
@@ -1257,6 +1262,62 @@ bool Engine::CancelRequest(const std::shared_ptr<Request>& request, uint64_t tur
...
+
+  // Every fallible preparation completes before committed cache ownership is released. The cache
+  // managers validate their complete release up front; their publication paths are no-throw.
+  cache_manager_->ValidateRewind(request);
+  auto rewind_state =
+      request->PrepareRewindToStartOfTurn(turn_id);
+  // The auxiliary decoder mirrors a generated suffix that is no longer authoritative after
+  // rewind. Release it before the target cache; its manager validates the complete deallocation
+  // before publishing, while the already-validated target release below is allocation-free.
+  CloseMtpRequest(request);
```
baijumeswani, line 1316: Rewind drops the MTP state here, but it leaves the DFlash2/DSpark state attached to this request. That state still points to the old sequence position. When replay starts again at position 0, the drafter reports a position mismatch, clears every request from the drafter, and disables drafting for the rest of the Engine lifetime. Generation still works, so this slowdown can be easy to miss.

Recommendation: call `dflash2_drafter_->Release(request.get())` here before releasing the target cache. Please also add a DFlash2/DSpark rewind test that confirms replay does not increase the disable count and that another request keeps its drafter state.

src/engine/engine.cpp
```diff
@@ -1151,8 +1151,13 @@ uint64_t Engine::BeginTurn(const std::shared_ptr<Request>& request,
     const bool restartable_canceled_turn =
         request->IsRestartableCanceledTurn() &&
         !cache_manager_->IsResident(request);
-    ValidateRequestCanContinue(request, restartable_canceled_turn);
-    if (!restartable_canceled_turn) {
+    const bool replay_rewound_request =
+        request->NeedsReplayAfterRewind() &&
```
baijumeswani, line 1155: This flag can remain set if replay is admitted but then fails because the full replay no longer fits. That failure removes the request from the scheduler. A later `BeginTurn()` still passes this check, but because it is not the first turn, the request is not added back to the scheduler. The call succeeds, yet the turn never runs or produces an event.

Recommendation: clear `needs_replay_after_rewind_` when a request fails or closes, and make this check reject failed requests. A test covering rewind, an unserviceable replay, and a second `BeginTurn()` would protect this path.

test/python/test_onnxruntime_genai_engine.py
```diff
@@ -311,6 +311,44 @@ def test_deterministic_tokens(model):
     assert len(tokens) == max_new
 
 
+def test_request_rewind_replays_retained_prefix(model):
+    engine = og.Engine(model)
+    params = og.GeneratorParams(model)
+    params.set_search_options(do_sample=False, max_length=24)
+    request = engine.create_request(params)
```
baijumeswani: This is the cause of the current Linux and Windows CUDA build failures. `create_request` accepts a keyword-only `RequestOptions`, not a positional `GeneratorParams`, so this raises `TypeError` before the rewind behavior is tested.

Recommendation: create `og.RequestOptions()`, set `max_session_tokens` to 24, and call `engine.create_request(options=request_options)`. Greedy behavior can remain in the turn options or use the model default.

src/engine/request.cpp
```diff
@@ -219,6 +224,7 @@ void Request::PrepareTurnAdmission(
   admission.prompt_sequence_length = prompt_sequence_length_;
   admission.processed_sequence_length = processed_sequence_length_;
 
+  turn_boundaries_.reserve(turn_boundaries_.size() + 1);
```
baijumeswani: Reserving exactly one more slot commonly reallocates and copies the full boundary list on every turn. Over a long conversation, that makes boundary bookkeeping grow quadratically even though each turn adds only one small entry.

Recommendation: only reserve when size reaches capacity, then grow capacity geometrically, for example by doubling it with a small initial capacity. This keeps the later commit safe without repeated full copies.

src/engine/engine.cpp
```diff
@@ -1257,6 +1262,62 @@ bool Engine::CancelRequest(const std::shared_ptr<Request>& request, uint64_t tur
...
+  }
+
+  // Every fallible preparation completes before committed cache ownership is released. The cache
+  // managers validate their complete release up front; their publication paths are no-throw.
+  cache_manager_->ValidateRewind(request);
+  auto rewind_state =
+      request->PrepareRewindToStartOfTurn(turn_id);
+  // The auxiliary decoder mirrors a generated suffix that is no longer authoritative after
+  // rewind. Release it before the target cache; its manager validates the complete deallocation
+  // before publishing, while the already-validated target release below is allocation-free.
```
baijumeswani: The target release is not fully allocation-free. The paged path creates and reserves temporary vectors, and the static path builds temporary sets before publishing the release. An allocation failure is still safe because it happens before target cache ownership changes; only the rebuildable MTP shadow has been dropped.

Recommendation: keep this ordering, but update the comment to say that validation and temporary allocation happen before the no-throw publication step. That describes the actual safety guarantee.

src/engine/cache_manager.cpp
```diff
@@ -307,6 +307,28 @@ void StaticCacheManager::Deallocate(std::vector<std::shared_ptr<Request>>& reque
   cache_allocated_requests_.clear();
 }
 
+void StaticCacheManager::ValidateRewind(
+    const std::shared_ptr<Request>& request) const {
+  if (!IsResident(request)) {
+    return;
+  }
+  if (cache_allocated_requests_.size() != 1) {
```
baijumeswani, line 315: This also counts closed rows whose static cache storage is being retained. As a result, a completed request cannot rewind if another row has already been closed but the old batch has not yet been recycled; it may stay blocked until unrelated new work arrives.

Recommendation: either explicitly document this as a static-path limitation, or allow rewind when every other resident row is closed, release the full batch, and finish those closed rows after the release.

baijumeswani (commented):
This PR adds a public way to rewind a completed Engine request to the start of an earlier turn. It keeps the request handle and retained token prefix, removes the selected turn and later turns, releases model state, and rebuilds that state when the next turn begins. The C, C++, and Python APIs follow the same model.

The overall structure is strong. The lifecycle checks are clear, pending events are handled safely, turn IDs are not reused, sampling state is intentionally preserved, and the paged-cache, fixed-state, static-cache, and MTP paths have good focused coverage. Rebuilding Search state instead of trying to trim device-specific Search objects is also the safer design.

I found two runtime issues that should be fixed before approval, plus the Python test failure now reported by Linux and Windows CI. I also left recommendations for boundary-history growth, comment accuracy, and a static-cache limitation. I agree with the existing review comments about the stale `OnNextTokensSampled` parameter text and the old API names in the PR description, so I did not duplicate those inline comments.

src/engine/request.h
```diff
@@ -489,6 +509,9 @@ struct Request : std::enable_shared_from_this<Request>,
 
   /**
    * @brief Launches the per-sequence tail after a batched sampler has filled the bound slot.
+   * @param sampler_draw_consumed True when the scheduler-owned sampler consumed one draw. Sampled
+   *        speculative verification binds host-selected tokens through the same Search hook without
+   *        advancing the batched-sampler stream.
    */
   void OnNextTokensSampled();
```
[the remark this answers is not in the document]
bmehta001: Fixed in 434cc0f0 by removing the stale nonexistent parameter documentation from `OnNextTokensSampled()`.

src/ort_genai_c.h
```diff
@@ -1514,6 +1514,23 @@ OGA_EXPORT OgaResult* OGA_API_CALL OgaRequestBeginTurn(
...
+ * later Turns are discarded. Turn IDs are never reused, and sampling continues from its current
+ * random stream. The operation releases all resident model state. The next OgaRequestBeginTurn
+ * creates a new Turn and replays the retained prefix together with its new input before generation.
+ *
+ * Dynamic paged Requests release their paged blocks and any fixed recurrent/convolution slot
+ * atomically. Static Requests are rewindable only when they are the sole resident row, because one
+ * row cannot be removed from a shared contiguous allocation. No event is emitted by rewind.
+ */
+OGA_EXPORT OgaResult* OGA_API_CALL OgaRequestRewindToStartOfTurn(
+    OgaRequest* request, uint64_t turn_id);
```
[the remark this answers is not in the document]
bmehta001: Fixed in PR metadata for 434cc0f0: the description now names `OgaRequestRewindToStartOfTurn` and `OgaRequest::RewindToStartOfTurn`.

src/python/python.cpp
```diff
@@ -906,6 +906,9 @@ PYBIND11_MODULE(onnxruntime_genai, m) {
           pybind11::arg("tokens"),
           pybind11::arg("turn_options") = pybind11::none())
       .def("cancel_turn", &OgaRequest::CancelTurn)
+      .def("rewind_to_start_of_turn", &OgaRequest::RewindToStartOfTurn,
+           pybind11::arg("turn_id"),
+           "Rewind a completed Request to the sequence boundary before a Turn began.")
```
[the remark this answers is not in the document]
bmehta001: Fixed in PR metadata for 434cc0f0: the description now names Python `Request.rewind_to_start_of_turn`.

src/engine/engine.cpp
bmehta001: Fixed in 434cc0f0. Rewind now calls `dflash2_drafter_->Release(request.get())` before target-cache release. The DSpark lifecycle test now verifies the rewound request can replay from position zero while a peer continues from its retained drafter position without an admission miss.

src/engine/engine.cpp
bmehta001: Fixed in 434cc0f0. Failed and closed terminal paths clear `needs_replay_after_rewind_`, continuation validation explicitly rejects a failed turn, and `FailedReplayCannotBeginAnotherTurn` covers an unserviceable replay followed by a second `BeginTurn()`.

test/python/test_onnxruntime_genai_engine.py
bmehta001: Fixed in 434cc0f0. The Python test now creates `RequestOptions`, sets `max_session_tokens`, and passes it through the keyword-only `options=` argument.

src/engine/request.cpp
bmehta001: Fixed in 434cc0f0. Turn-boundary capacity now grows geometrically (with a small initial allocation and overflow guard) while preserving the pre-transaction allocation guarantee.

src/engine/engine.cpp
bmehta001: Fixed in 434cc0f0. The comment now states the actual guarantee: validation and temporary allocation precede target-cache ownership mutation, followed by no-throw publication.

src/engine/cache_manager.cpp
bmehta001: Documented in 434cc0f0. Both Engine specifications now state that logically closed static peers remain resident until batch recycling and can temporarily block another completed request from rewinding.

test/cpp/dflash2_config_test.cpp
```diff
@@ -545,22 +543,27 @@ TEST(Dflash2ConfigTest, RunsFullAttentionDsparkAcrossRequestLifecycles) {
...
+  const std::array reused_feeds{
+      Dflash2Drafter::Feed{.request = request_a, .aux_row_begin = 0, .aux_row_count = 9, .first_position = 0, .anchor_token = 15, .draft_eligible = true, .wants_drafts = true},
+      Dflash2Drafter::Feed{.request = request_b, .aux_row_begin = 9, .aux_row_count = 1, .first_position = 13, .anchor_token = 16, .draft_eligible = true, .wants_drafts = true},
   };
-  drafter.Propose(reused_aux, reused_feed, drafts);
-  ASSERT_EQ(drafts.size(), 1u);
-  EXPECT_EQ(drafts[0], (std::vector<int32_t>{14, 0, 13, 6}));
+  drafter.Propose(reused_aux, reused_feeds, drafts);
+  ASSERT_EQ(drafts.size(), 2u);
+  EXPECT_FALSE(drafts[0].empty());
```
baijumeswani, line 558: These checks only show that both requests produced some drafts. This test used to check the exact tokens, so it could now pass even if replay produces wrong drafts or affects the other request. Could we check the expected tokens for both requests here and keep an exact check for a fresh request? That would catch incorrect drafter state, not just missing drafts.

src/engine/engine.cpp
```diff
@@ -1327,6 +1332,65 @@ bool Engine::CancelRequest(const std::shared_ptr<Request>& request, uint64_t tur
...
+  cache_manager_->ValidateRewind(request);
+  auto rewind_state =
+      request->PrepareRewindToStartOfTurn(turn_id);
+  // Auxiliary decoders mirror a generated suffix that is no longer authoritative after rewind.
+  // Release them before the target cache so any fallible cleanup leaves target ownership intact.
+  CloseMtpRequest(request);
+  if (dflash2_drafter_) {
+    dflash2_drafter_->Release(request.get());
+  }
+  cache_manager_->ReleaseForRewind(request);
```
baijumeswani, line 1390: Rewind releases the paged blocks and fixed-state slot here, so another request can reuse them before this request resumes. The next decode might then replay a previously captured CUDA graph. I do not see a CUDA test for that sequence, so this is a coverage concern rather than a known bug. Could we capture a decode graph, rewind request A, let request B reuse the released state, then resume A and compare its tokens with an uncaptured run? That would check that graph replay reads the updated block table and fixed-state binding.

baijumeswani (commented):
Two focused follow-ups for rewind coverage and regression protection.

test/cpp/dflash2_config_test.cpp
bmehta001: Addressed in 04412fba. The DSpark lifecycle test now checks the full expected token vectors for both the rewound request and its peer (and retains the full-vector checks for fresh requests). This catches wrong drafts as well as missing drafts. The expected values follow the synthetic model outputs and block assignment in `src/dflash2_drafter.cpp:611-620,860-881`.

src/engine/engine.cpp
bmehta001: Addressed in 04412fba. `test/python/test_onnxruntime_genai_engine_cuda_graph.py` now compares eager versus captured tokens after A decodes and rewinds, B reuses the released paged state (and fixed-state slot if the supplied model declares one), then A replays its retained prefix. It also compares A's replay against its original turn and checks B's tokens. Like the existing CUDA graph tests, it needs `ORTGENAI_PAGED_CUDA_GRAPH_MODEL` pointing to a capture-capable model; no such CUDA model/GPU is available locally, so execution remains pending in an appropriately provisioned environment.

baijumeswani:
Two non-blocking documentation comments:

1. The code allows another rewind after a request has already been rewound and is no longer resident. The docs currently say that a canceled, never-admitted first turn is the only supported nonresident case. Could we either document and test repeated rewind, or reject it in the code?

2. Rewind releases DFlash/DSpark state, so the request may be considered for drafter admission again during replay. The docs currently say that sampled or admission-denied requests stay ineligible for the rest of the request's life. Could we document the reset and re-admission behavior, and include DFlash/DSpark in the rewind resource list?

baijumeswani (dismissed):

bmehta001:
@baijumeswani Addressed both follow-ups in 8d786054. Consecutive nonresident rewinds to an earlier retained turn remain supported and now have a lifecycle test that checks history pruning, no model residency between rewinds, and monotonic turn IDs. The Engine/C API docs explicitly describe this case. They also include DFlash/DSpark in rewind cleanup and clarify that releasing drafter tracking permits a new position-zero admission attempt on replay (subject to turn eligibility, capacity, and whether prefix adoption skips position zero). Updated the PR description as well.

baijumeswani (approved):

src/engine/engine.cpp
```diff
@@ -1394,6 +1399,65 @@ bool Engine::CancelRequest(const std::shared_ptr<Request>& request, uint64_t tur
...
+  request->CommitRewind(std::move(rewind_state));
```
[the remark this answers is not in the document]
bmehta001: Thanks for flagging this. I verified the static path against scheduler.cpp:91-115: on first admission, allocated_requests is empty, so the loop that erases old entries from requests_pool_ has nothing to remove. An allocated request is removed from the pool only when a replacement batch is allocated; at that point it is no longer resident and rewind rejects it (engine.cpp:1438-1443). For the supported resident single-row rewind, its existing pool entry survives cache release; after BeginTurn sets it queued, StaticBatchScheduler::HasPendingRequests and Schedule pick it up without AddRequest or resetting its sampler state. I ran EngineRunTest.StaticSingleRowRewindReleasesAndReplaysContiguousCache from the current built source; it passes and verifies replay produces a token and reacquires static residency. No change needed here.

PR #2521, opened 2026-09-02
Fix feature leakage across Qwen multi-image inputs

apsonawane:
## Summary

Fix Qwen multi-image inference when images produce different patch-grid sizes.

ORT Extensions pads each image to a common patch stride before flattening the batch. The Qwen vision loop previously treated this tensor as tightly packed, causing later images to read patches from earlier images.

## Changes

- Detect per-image padded patch layout.
- Use the padded stride when calculating image offsets.
- Continue passing only each image's logical patches to the vision model.
- Reject unsupported patch layouts instead of silently mixing features.

## Validation

- Verified inference with 2, 3, 4, and 8 images.
- Confirmed later images no longer contain features from earlier images.
- Passed `QwenVisionMultiImageTest.*`.
- Built and linked the native runtime successfully.

### DIFF
--- a/src/models/multi_modal.h
+++ b/src/models/multi_modal.h
@@ -59,20 +59,6 @@ struct VisionState : State {
   std::unique_ptr<MultiModalFeatures> image_features_;
 };
 
-// QwenVisionState: per-image slicing loop for Qwen2.5-VL / Qwen3-VL.
-//
-// vision.onnx is exported for exactly one image (Dynamo unrolls Python
-// for-loops at trace time, so an N-image dummy produces a graph that only
-// works for that exact N).  This subclass iterates over images in C++,
-// creating zero-copy sub-tensor views of pixel_values / image_grid_thw and
-// writing each result into the correct offset of the pre-allocated
-// image_features output buffer.
-struct QwenVisionState : VisionState {
-  using VisionState::VisionState;  // inherit constructor
-
-  DeviceSpan<float> Run(int current_length, DeviceSpan<int32_t>& next_tokens, DeviceSpan<int32_t> next_indices = {}) override;
-};
-
 // PixtralVisionState: per-image vision loop for Pixtral / Mistral3.
 //
 // Each image is independently smart_resize'd to a different resolution.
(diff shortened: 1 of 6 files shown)


titaiwangms:
## Review findings

**Verdict: request changes.**

### Major — equal-stride detection can shadow the existing temporal-padding layout

`src/models/multi_modal.cpp:297-300`, `src/models/multi_modal.h:128-138`

The new equal-stride predicate runs before temporal-padding detection, but the two layouts are not distinguishable from these aggregate values alone. A previously supported temporal-padded input can therefore be silently reinterpreted as fixed-stride.

Concrete example with two images:

- Grids: `(t,h,w) = (1,2,2)` and `(1,2,4)`
- Logical grid tokens: `4` and `8`; `total_grid_tokens = 12`
- `total_hw = 12`, `total_patches = 24`
- Existing temporal multiplier: `24 / 12 = 2`

The previous temporal path consumes:

- Image 0: offset `0`, count `8`
- Image 1: offset `8`, count `16`

The new code instead infers `candidate_stride = 24 / 2 = 12`, which is at least `max_grid_tokens = 8`. It disables `temporal_padded` and consumes:

- Image 0: offset `0`, count `4`
- Image 1: offset `12`, count `8`

All bounds checks pass, so this produces incorrect image embeddings rather than rejecting the layout.

The fixed-stride interpretation matches the current ORT Extensions pad-to-max producer, but aggregate dimensions cannot prove whether a mismatched tensor is fixed-stride or temporal-padded. Please use explicit producer/layout metadata, or detect the ambiguous case rather than unconditionally prioritizing stride.

The tests should also include a temporal/stride collision case and exercise the offset selection used by `QwenVisionState::Run`; the current additions only verify helper arithmetic and would not catch this regression.

src/models/multi_modal.h
```diff
@@ -125,6 +125,48 @@ inline void ValidateImageGridThwLayoutAndCount(const std::vector<int64_t>& shape
   }
 }
 
+struct QwenPatchLayout {
```
kunal-vaishnavi: Can we migrate all of the Qwen-specific logic into its own files? The base `multi_modal.cpp`/`multi_modal.h` files are getting quite polluted with substantial model-specific logic. Separating it would help keep the shared multimodal implementation focused and easier to maintain.

apsonawane: Updated

titaiwangms:
## Follow-up review

**Verdict: changes are still required.** The earlier temporal-layout issue is now handled by rejecting cases where both predicates match, but that rejection also rejects some valid ORT Extensions fixed-stride inputs.

### Major — valid pad-to-max layouts can be rejected as ambiguous

`src/models/qwen_vl_state.h:43-50`

`ResolveQwenPatchLayout()` treats divisibility by `total_hw` as evidence of temporal padding and throws whenever that predicate overlaps the fixed-stride predicate. The predicates can overlap for legitimate ORT Extensions output.

For example, consider three still images with logical patch counts `256`, `64`, and `64`. ORT Extensions pads each image to the largest patch count before stacking, so flattening `[3, 256, patch_dim]` produces:

- `total_patches = 3 * 256 = 768`
- `total_grid_tokens = total_hw = 384`
- `max_grid_tokens = 256`
- `candidate_stride = 768 / 3 = 256`

Both predicates are then true:

- `stride_padded`: `candidate_stride >= max_grid_tokens`
- `temporal_padded`: `768 % 384 == 0`

The resolver throws `pixel_values patch layout is ambiguous...`, even though this is a normal fixed-stride pad-to-max tensor from the current producer. Multi-image inference therefore fails for input-dependent combinations of otherwise valid image sizes.

The current tests do not cover this producer-valid overlap. `UsesPaddedStrideForDifferentImageGridSizes` uses values where `total_patches` is not divisible by `total_hw`, while `RejectsAmbiguousStrideAndTemporalPadding` codifies the rejection rather than testing a legitimate pad-to-max layout.

Please use the producer's stronger invariant for fixed-stride tensors—for example, `total_patches == num_images * max_grid_tokens`—together with the per-image temporal dimensions to distinguish temporal padding. Alternatively, preserve explicit stride/layout metadata before flattening instead of inferring the layout from aggregate counts. Add a regression test for a valid fixed-stride case where `total_patches % total_hw == 0`.

The new module's CMake, factory, include, and Qwen model-family wiring otherwise look consistent.

titaiwangms:
## Follow-up review

**Verdict: changes are still required.**

### Major — fixed-stride precedence can still misclassify temporal padding

`src/models/qwen_vl_state.h:43-52`

`ResolveQwenPatchLayout()` now recognizes the current ORT Extensions pad-to-max layout more precisely, but a valid temporal-padded layout can still satisfy the fixed-stride predicate. Because fixed stride is checked first, the input is silently sliced using the wrong offsets and counts.

For example, with two grid entries:

- `(t, h, w) = (1, 2, 2)` → `grid_tokens = 4`, `h*w = 4`
- `(t, h, w) = (2, 2, 2)` → `grid_tokens = 8`, `h*w = 4`
- temporal multiplier `= 2`

The aggregate values are:

- `total_grid_tokens = 12`
- `total_hw = 8`
- `total_patches = 16`
- `candidate_stride = 16 / 2 = 8`
- `max_grid_tokens = 8`

Both predicates therefore match, but the resolver selects fixed stride. The loop consumes `(offset=0, count=4)` for the first entry instead of the temporal layout's `(offset=0, count=8)`. The existing bounds checks all pass, so this produces incorrect vision features rather than an error.

The built-in image producer currently emits `t=1`, but public/custom `NamedTensors` can provide mixed-`t` grids, and this code explicitly retains temporal-layout support. Aggregate dimensions alone cannot distinguish the two layouts safely.

Please either preserve explicit producer/layout metadata, restrict fixed-stride detection to a producer invariant that rules out temporal layouts (for example, all `t == 1`), or reject genuinely ambiguous inputs instead of silently choosing fixed stride. Add a regression test for the values above and exercise the slicing decision used by `QwenVisionState::Run`, not only the helper arithmetic.

src/models/qwen_vl_state.h
```diff
@@ -0,0 +1,60 @@
...
+inline QwenPatchLayout ResolveQwenPatchLayout(int64_t total_patches,
+                                              int64_t total_grid_tokens,
+                                              int64_t total_hw,
+                                              int64_t max_grid_tokens,
+                                              int64_t num_images) {
+  if (total_patches == total_grid_tokens) {
+    return {};
+  }
+
+  const bool temporal_padded = total_patches > 0 && total_hw > 0 && total_patches % total_hw == 0;
```
kunal-vaishnavi, line 43: From Copilot:

This still ambiguously classifies valid temporal-padded inputs as fixed-stride. For grids `(1,2,2)` and `(2,2,2)` with `total_patches = 16`, both predicates match: the temporal multiplier is `2`, while the candidate stride equals `max_grid_tokens` at `8`. Because fixed-stride wins, image 0 receives 4 patches instead of 8, silently producing incorrect features.

Could we restrict fixed-stride detection to a producer invariant that excludes temporal layouts, such as requiring all `t == 1`, or preserve explicit layout metadata? Please also add this collision as a regression test.

src/models/qwen_vl_state.cpp
```diff
@@ -0,0 +1,249 @@
...
+  auto pv_info = pv_full->GetTensorTypeAndShapeInfo();
+  auto feat_info = feat_full->GetTensorTypeAndShapeInfo();
+  auto pv_shape = pv_info->GetShape();
+  auto feat_shape = feat_info->GetShape();
+  auto pv_type = pv_info->GetElementType();
+  auto feat_type = feat_info->GetElementType();
+  int64_t patch_dim = pv_shape[1];
+  int64_t hidden_size = feat_shape[1];
+
+  auto element_size = [](ONNXTensorElementDataType type) -> size_t {
```
kunal-vaishnavi, line 141: Can we use `Ort::SizeOf` here instead of this?

PR #2524, opened 2026-09-02
Support multi-image inference for Gemma 4

apsonawane:
## Summary

- preserve per-image vision token counts during Gemma 4 preprocessing
- run static batch-size-1 vision models once per image
- concatenate image features in prompt image order
- validate tensor shapes, token counts, and output capacity
- add tests for per-image metadata and prompt-feature alignment

## Motivation

Mobius-exported Gemma 4 vision models have a static image batch dimension of
one. Passing multiple images directly caused `pixel_position_ids` dimension
validation failures.


src/models/multi_modal.cpp
```diff
@@ -368,6 +375,144 @@ DeviceSpan<float> QwenVisionState::Run(int current_length, DeviceSpan<int32_t>&
...
+  const OrtValuePointerRestore restore_pixel_values{inputs_[pixel_values_index_], pixel_values};
+  const OrtValuePointerRestore restore_position_ids{inputs_[position_ids_index_], position_ids};
+  const OrtValuePointerRestore restore_image_features{outputs_[0], image_features};
+  const auto pixel_info = pixel_values->GetTensorTypeAndShapeInfo();
+  const auto position_info = position_ids->GetTensorTypeAndShapeInfo();
+  const auto feature_info = image_features->GetTensorTypeAndShapeInfo();
+  const auto pixel_shape = pixel_info->GetShape();
+  const auto position_shape = position_info->GetShape();
+  const auto feature_shape = feature_info->GetShape();
+  if (pixel_shape.size() != 3 || position_shape.size() != 3 ||
```
kunal-vaishnavi: From Copilot:

Could we also validate that `position_shape[1] == pixel_shape[1]` before creating the per-image views? The offset and byte length for `position_data` below are calculated using `num_patches` from `pixel_shape[1]`. If the tensors disagree on that dimension, the resulting view could extend beyond the position tensor’s allocation rather than producing a targeted shape error.

test/python/models/test_gemma4_models.py
```diff
@@ -138,12 +138,47 @@ def test_gemma4_vision_multiple_images(test_data_path, relative_image_paths):
...
+    assert np.all(valid_patch_counts % 9 == 0)
+    expected_token_counts = valid_patch_counts // 9
+    np.testing.assert_array_equal(image_token_counts, expected_token_counts)
+
+
+@pytest.mark.parametrize(
+    "relative_image_paths",
+    [[Path("images") / "australia.jpg", Path("images") / "landscape.jpg"]],
+)
+def test_gemma4_multiple_image_tokens_align_with_feature_rows(test_data_path, relative_image_paths):
```
kunal-vaishnavi: From Copilot:

Could this test exercise the new vision execution path with two images that produce different token counts? Despite the test name, it currently only compares `token_type_ids` with preprocessing metadata, so it would still pass if `Gemma4VisionState::Run` sliced the inputs incorrectly, failed to concatenate features in image order, or was never invoked. An end-to-end generation test, or a test that directly runs the static-B=1 vision model and verifies the resulting feature rows, would cover the behavior introduced by this PR.

src/models/gemma4_vision_state.cpp
```diff
@@ -0,0 +1,158 @@
...
+  if (pixel_shape.size() != 3 || position_shape.size() != 3 ||
+      pixel_shape[0] != num_images_ || position_shape[0] != num_images_ ||
+      pixel_shape[1] != position_shape[1]) {
+    throw std::runtime_error("Gemma 4 multi-image vision inputs must have matching [num_images, num_patches, ...] shapes");
+  }
+  if (feature_shape.size() != 2) {
+    throw std::runtime_error("Gemma 4 image_features must have rank 2 [num_image_tokens, hidden_size]");
+  }
+
+  auto element_size = [](ONNXTensorElementDataType type) -> size_t {
```
kunal-vaishnavi: You can use `Ort::SizeOf` instead.

src/models/multi_modal.cpp
```diff
@@ -573,6 +574,9 @@ std::unique_ptr<VisionState> CreateVisionState(const MultiModalLanguageModel& mo
   if (ModelType::IsQwenVLFamily(model.config_->model.type)) {
     return std::make_unique<QwenVisionState>(model, params);
```
kunal-vaishnavi, line 377: Can we also move the Qwen and Pixtral specific code out of the base file as well? Ex:

https://github.com/microsoft/onnxruntime-genai/blob/e3fb734925b7dba6ed026bbcd2ddb05c24010d1c/src/models/multi_modal.cpp#L158-L567

The `CreateVisionState` method can stay in this file.

test/python/models/test_gemma4_models.py
```diff
@@ -59,6 +61,134 @@ def _to_numpy(tensor):
...
+        onnx.numpy_helper.from_array(np.array(0, dtype=np.float32), "zero"),
+    ]
+    graph = helper.make_graph(
+        nodes, "gemma4_static_batch_vision", [pixel_values, position_ids], [image_features], initializers
+    )
+    model = helper.make_model(graph, opset_imports=[helper.make_opsetid("", 14)], ir_version=7)
+    onnx.save(model, output_path)
+
+
+def _create_dynamic_embedding_model(onnx, output_path):
```
kunal-vaishnavi, line 104: From Copilot:

Could we make this fixture observably depend on `image_features` and assert the resulting values or row order? The embedding graph declares `image_features` as an input but never consumes it, so the test only verifies that `generator.set_inputs()` completes. It would still pass if the vision loop produced incorrect values, wrote features at the wrong offsets, or reversed the image order.

Since feature concatenation in prompt-image order is a core behavior of this change, please use distinguishable per-image data and propagate the vision output into an observable downstream output, or directly inspect and assert the combined feature rows.

src/models/gemma4_vision_state.cpp
```diff
@@ -0,0 +1,142 @@
...
+#include "gemma4_vision_state.h"
+
+#include <array>
+#include <numeric>
+
+namespace Generators {
+
+namespace {
+
+struct OrtValuePointerRestore {
```
kunal-vaishnavi, line 14: Other per-image vision states do not introduce a separate pointer-restoration helper. Why is `OrtValuePointerRestore` needed specifically for Gemma-4? Could this implementation follow the same pattern as the existing states, or otherwise be restructured so that temporarily rebinding `inputs_` and `outputs_` does not require a file-local guard?

PR #2527, opened 2026-09-03
Add transactional stop-string support to the dynamic Engine

baijumeswani:
## Summary

Add decoded UTF-8 stop-string support to ordinary dynamic Engine generation while preserving the Engine's transactional execution and rollback guarantees.

Stop strings are matched incrementally against text decoded from generated tokens for the current turn. The token that completes a match remains part of the committed output, and the completion event reports `StopString` together with the matching stop-string index so the host can apply its own output holdback and suppression policy.

## Changes

- Add C, C++, and Python APIs for configuring per-turn stop strings.
- Add a model-independent incremental byte matcher with deterministic overlap resolution and per-pattern KMP state.
- Add a request-local stop controller that bridges tokenizer streaming to the matcher and rebuilds its stream by replaying committed current-turn tokens after rollback.
- Integrate stop matching with dynamic Engine admission, staging, commit, direct rollback, queued rollback, cancellation, fatal replacement, continuation, guidance, and event publication.
- Validate stop strings as nonempty UTF-8, copy caller-owned configuration at admission, and support up to 16 strings with a 16 KiB aggregate limit.
- Rename finish reason value `2` from the unused `StopToken` placeholder to `StopString` and expose the matched caller index in Engine events.
- Add Engine, API, tokenizer-stream, rollback, lifecycle, allocation, and performance qualification coverage.

## Semantics

- Matching applies only to generated output from the current turn; prompts, continuation input, and prior turns do not seed matcher or tokenizer-stream state.
- Matches use exact incremental decoded UTF-8 bytes.
- When multiple strings match at the same ending byte, selection prefers the longest match and then the lowest caller-provided index.
- The token completing the match is retained, emitted, cached, and counted.
- The Engine reports the match but does not suppress matched or trailing decoded bytes; hosts remain responsible for output holdback and presentation.


src/ort_genai_c.h
```diff
@@ -93,7 +93,7 @@ typedef struct OgaStreamingProcessor OgaStreamingProcessor;
 typedef uint32_t OgaFinishReason;
 #define OgaFinishReason_None ((OgaFinishReason)0)
 #define OgaFinishReason_Eos ((OgaFinishReason)1)
-#define OgaFinishReason_StopSequence ((OgaFinishReason)2)
+#define OgaFinishReason_StopString ((OgaFinishReason)2)
 #define OgaFinishReason_MaxGeneratedTokens ((OgaFinishReason)3)
```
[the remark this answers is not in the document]
baijumeswani: This is intentional. There was no release with the previous version. So this is ok for this PR.

src/ort_genai_c.h
```diff
@@ -1393,18 +1401,19 @@ OGA_EXPORT OgaResult* OGA_API_CALL OgaTurnOptionsSetTopK(
...
+ * no normalization, trimming, or case folding, and only text this Engine Request generates during
+ * the turn (never prompt or earlier-turn tokens) is considered.
  *
- * This operation is currently not implemented. stop_strings is not dereferenced or retained
- * before the not-implemented result is returned.
+ * A subsequent OgaRequestBeginTurn snapshots these options, so reusing or mutating this
+ * OgaTurnOptions afterward cannot alter an already-active turn.
  */
 OGA_EXPORT OgaResult* OGA_API_CALL OgaTurnOptionsSetStopStrings(
     OgaTurnOptions* options, const OgaStringArray* stop_strings);
```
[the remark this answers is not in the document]
baijumeswani: This is intentional.

src/python/python.cpp
```diff
@@ -778,7 +778,7 @@ PYBIND11_MODULE(onnxruntime_genai, m) {
   pybind11::enum_<PyFinishReason>(m, "FinishReason")
       .value("NONE", PyFinishReason::None)
       .value("EOS", PyFinishReason::Eos)
-      .value("STOP_SEQUENCE", PyFinishReason::StopSequence)
+      .value("STOP_STRING", PyFinishReason::StopString)
       .value("MAX_GENERATED_TOKENS", PyFinishReason::MaxGeneratedTokens)
       .value("MAX_SESSION_TOKENS", PyFinishReason::MaxSessionTokens)
```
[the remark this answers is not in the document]
baijumeswani: This is intentional

src/engine/request.cpp
```diff
@@ -186,11 +191,16 @@ void Request::PrepareTurnAdmission(
...
+    const TurnOptions& options,
     RequestTurnAdmission& admission) {
   assert(admission.transaction_started);
   CommitStateForTransaction();
-  turn_max_generated_tokens_ = max_generated_tokens;
+  // The caller built this turn's complete stop controller (or null, for a no-stop turn) before
+  // the admission attempt began and it has been untouched ever since, so installing it here is a
+  // plain, infallible move: whatever the Request's stop_controller_ was before this call (a prior
+  // turn's controller, or null) is simply discarded now that the new turn is durable.
+  stop_controller_ = std::move(admission.pending_stop_controller);
```
tianleiwu, line 202: `stop_controller_transaction_checkpoint_` is a *count* that is only meaningful against whichever controller is currently installed. `SaveStateForNewTurnTransaction()` sets it from the **previous** turn's controller, and this line then swaps in a fresh controller with zero tokens without resetting it.

Today nothing between this move and the next `SaveState*ForTransaction()` can throw, so `RollbackTurnAdmission()` -> `RestoreStateForTransaction()` is never reached with the mismatched pair. But if any statement added after the move here can throw, the rollback would call `RollbackTo(previous_turn_count)` against a zero-length history and raise `std::logic_error("Stop-string rollback checkpoint exceeds staged token history.")` from inside a rollback path.

Suggest making the invariant local and self-evident:

```cpp
stop_controller_ = std::move(admission.pending_stop_controller);
stop_controller_transaction_checkpoint_ = 0;
```

src/engine/request.cpp
```diff
@@ -896,20 +963,39 @@ RequestStepResult Request::StageGeneration(int64_t sequence_length_before) {
...
+  // Stop strings take precedence over the turn/context limit for the same generated token: a
+  // token that would otherwise end the turn on one of those limits still reports StopString when
+  // it also completes a match. Only a token this step actually appended can be observed here --
+  // prompt/continuation tokens never reach this point, stop_controller_ is null on the no-stop
+  // fast path (no tokenizer, no decode work), and GreedySearch_Cpu never appends a single-sequence
+  // Request's sampled EOS token because its sampling entry points skip appending once SetNextToken
+  // marks the Search done. A real EOS token's bytes therefore never reach the matcher here --
+  // consistent with GenAI tokenizers registering EOS as a special, zero-byte-decoding token by
+  // default.
+  if (token_appended && stop_controller_) {
```
tianleiwu, line 979: Two things here.

**1. `accepted_draft_count_ != 0` is silently ignored.** This feeds only the freshly sampled `token` to the controller, while `StageVisibleTokens(result, accepted_draft_count_, ...)` a few lines below may publish accepted draft tokens that were never matched against. It is unreachable today, but the guarantee lives entirely in `DraftTokenValidationError()`, three components away. A defensive check inside this branch turns a future regression into a loud contract failure instead of silently missed matches:

```cpp
if (token_appended && stop_controller_) {
  if (accepted_draft_count_ != 0) {
    throw std::logic_error(
        "Stop-string matching cannot observe accepted draft tokens.");
  }
  ...
```

The same coupling means an intermediate stage of the sampled-draft loop in `ScheduledRequests::GenerateNextTokensForTransaction` would hit `"An accepted sampled draft unexpectedly ended the request."` if a match ever landed mid-stage.

**2. Comment understates the precedence rule.** It says stop strings take precedence over "the turn/context limit", but the check is also ahead of the EOS branch. `docs/engine_c_api_spec.md` explains why that is unreachable in practice (EOS is never appended and decodes to zero bytes); one line saying the same here would keep the ordering from reading as an oversight.

tianleiwu (commented):
Reviewed the matcher, the controller, and the Engine/Request transaction wiring. No correctness defects found; the findings below are hardening, latency, and hygiene.

**What I verified by reading the code rather than the comments**

- **Matcher offset math is sound.** In `Consume()`, `match_->start_offset - pending_start_offset_` cannot underflow: after a no-match call the live pending suffix equals the longest active KMP prefix, and a pattern reaching length `L` at chunk byte `k` must have had prefix length `>= L - (k+1)` at chunk start, since KMP advances by at most 1 per byte. `pending_.size() - pending_begin_ - longest_active_prefix` is safe for the same reason.
- **`pattern[prefix_length]` is never read at `prefix_length == pattern.size()`**, because the match branch returns inside the same byte iteration and the matcher is sticky afterwards.
- **The amortized compaction bound holds**: after each call either `pending_begin_ < live` or the buffer is compacted, so logical size stays `< 2 * (longest_stop_string - 1)`.
- **`IsValidUtf8` is correct** for overlongs (`0xC0/0xC1`, `E0 < A0`, `F0 < 90`), surrogates (`ED > 9F`), out-of-range (`F4 > 8F`, leads `>= 0xF5`), and stray/truncated continuations.
- **Accepted-draft tokens cannot bypass the matcher today**, because `Request::DraftTokenValidationError()` returns non-null whenever `stop_controller_` is set and all three draft producers (`SetDraftTokens`, `PrepareMtpStep`, `PrepareDflash2Feeds`) consult it.
- **Event and invariant plumbing is consistent**: `matched_stop_string_index` is written only in `CommitStep` (post-commit, so never rolled back) and reset on new-turn admission, cancellation, and failure; `ValidateRequestInvariants` enforces the biconditional with `StopString`.
- **The removed `OgaTurnOptionsSetStopTokenIds` and the renamed `OgaFinishReason_StopSequence` are safe to drop**: neither symbol exists in `v0.9.2`, and no C#, Java, or Objective-C binding references them. (This matches the already-resolved automated threads on those lines.)

**Cross-cutting notes not attached to a specific line**

1. `stop_controller_` stays installed after the turn completes, holding a live `TokenizerStream` and the full turn token history until the next `BeginTurn` or `Close`. For a server keeping many idle multi-turn Requests around, releasing it in `CommitStep` when `result.done` would free that per-Request state immediately. `MatchedStopStringIndex()` is stored separately, so nothing after the turn reads the controller.
2. `EngineTurnOptionsStopStrings` covers the empty entry and the 16-entry bound but not the 16 KiB aggregate bound, and it does not assert the documented "prior configuration is left unchanged" behavior after a rejected `SetStopStrings`.

**Positives**

- `RequestTurnAdmission` owning the prebuilt controller, with installation confined to `CommitTurnAdmission()`, is the right way to make admission stop-string-transactional in both directions (enable and clear) without needing an undo path.
- Rejecting stop strings on a static-batching Engine before any Request mutation, rather than letting them run without the rollback guarantee the finish reason promises, is the correct trade.
- Draining `TakeSafeOutput()` in `ObserveToken()` prevents the matcher from retaining an entire turn's decoded bytes that nothing would ever read.
- Rejecting embedded NULs in the Python binding closes a real silent-truncation hole in the `OgaStringArray` C surface.
- The randomized differential test against an independent whole-stream `ReferenceMatch`, and the real-GPT-2-tokenizer test that documents the detokenizer's own UTF-8 buffering instead of assuming it, are both exactly the right shape for this component.
- The doc section on the detokenizer-context boundary (why a host's own full-stream decode can differ from the Engine's fresh-stream decode for the first generated token) is genuinely useful and easy to get wrong.

src/engine/engine.cpp
```diff
@@ -1020,6 +1029,18 @@ uint64_t Engine::BeginTurn(const std::shared_ptr<Request>& request,
...
+  // Build this turn's complete stop controller (or leave admission.pending_stop_controller null
+  // for a no-stop turn) before any Request or Engine mutation: tokenizer/stream construction can
+  // throw, and at this point nothing has been touched yet, so a failure here leaves the Request,
+  // its MTP shadow, and Engine health entirely unchanged for a caller to retry. Ownership passes
+  // into admission; Request::CommitTurnAdmission() installs it as the live stop_controller_, and
+  // Request::RollbackTurnAdmission() (or simply this function returning through the catch below)
+  // discards it without ever having touched the Request's actual stop_controller_.
+  if (!options.stop_strings.empty()) {
+    admission.pending_stop_controller = std::make_unique<StopStringController>(
+        GetOrCreateStopTokenizer(), options.stop_strings);
```
tianleiwu, line 1041: `GetOrCreateStopTokenizer()` lazily calls `Model::CreateTokenizer()`, which does disk I/O and ORT Extensions tokenizer initialization. In a continuous-batching server `BeginTurn` runs on the same owner thread as `Run()`, so the first stop-enabled admission adds a one-time head-of-line stall for every request already decoding.

Consider constructing the tokenizer eagerly at Engine construction when the model has one, or at minimum documenting the one-time admission cost in `docs/engine_c_api_spec.md` alongside the other stop-string admission semantics.

src/stop_string_controller.cpp
```diff
@@ -0,0 +1,62 @@
...
+  if (committed_token_count > turn_tokens_.size())
+    throw std::logic_error("Stop-string rollback checkpoint exceeds staged token history.");
+
+  // Nothing was staged since the checkpoint (e.g. a failed new-turn admission that never observed a
+  // token through this controller): the stream and matcher are already at the checkpoint, so skip
+  // rebuilding the stream and replaying tokens that are already reflected in the current state.
+  if (committed_token_count == turn_tokens_.size())
+    return;
+
+  turn_tokens_.resize(committed_token_count);
```
tianleiwu, line 44: Every rollback recreates the `TokenizerStream` and re-decodes the entire retained turn history. Since the unit of rollback is a step, a turn that hits repeated retryable aborts (for example `ExecutionCapacityExceeded` under paged-cache pressure, which the host is expected to retry) does O(n^2) detokenization on the Engine owner thread, multiplied by every stop-enabled request in the batch.

Given a non-cloneable detokenizer this is the only correctness-preserving design, so not a blocker. Worth stating the cost in `docs/engine_c_api_spec.md` next to the existing rollback discussion; a periodic replay checkpoint would be a reasonable follow-up if aborts turn out to be non-rare.

src/ort_genai_c.cpp
```diff
@@ -1924,8 +1926,16 @@ OgaResult* OgaTurnOptionsSetStopStrings(
...
   options->ValidateOwnerThread();
-  throw std::runtime_error(
-      "OgaTurnOptionsSetStopStrings is not implemented.");
+  // An empty array is a valid configuration (clears/disables stop strings); a nonempty array must
+  // have every entry nonempty (an empty member string is invalid even though an empty array is
+  // fine), validated against the same bounds and UTF-8 contract StopStringMatcher enforces at
+  // Request admission time, so a caller learns about a bad configuration immediately rather than at
+  // the next BeginTurn. The matcher itself is discarded; only the validated strings are retained.
+  if (!stop_strings->empty()) {
+    static_cast<void>(Generators::StopStringMatcher(*stop_strings));
```
tianleiwu: This copies every string and builds a full KMP failure table per entry only to discard all of it, and the same tables are rebuilt at `BeginTurn`.

A `ValidateStopStrings(const std::vector<std::string>&)` free function in `stop_string_matcher.h`, called by both this setter and `StopStringMatcher`'s constructor, would express the intent directly and avoid the duplicate table build. It would also make the shared contract between the two call sites explicit rather than implied by the comment.

test/CMakeLists.txt
```diff
@@ -17,6 +17,13 @@ list(REMOVE_ITEM test_srcs "${CMAKE_CURRENT_SOURCE_DIR}/cpp/dflash2_config_test.
 list(REMOVE_ITEM test_srcs "${CMAKE_CURRENT_SOURCE_DIR}/cpp/reinit_tests.cpp")
 list(REMOVE_ITEM test_srcs "${CMAKE_CURRENT_SOURCE_DIR}/cpp/search_checkpoint_tests.cpp")
 
+# The shared library exports only the public C API on Windows. The stop string matcher is a
+# self-contained internal component (no ORT or GenAI dependencies), so compile its translation unit
+# directly into the test binary there rather than relying on unexported symbols.
+if(WIN32)
+  list(APPEND test_srcs "${CMAKE_SOURCE_DIR}/src/stop_string_matcher.cpp")
```
tianleiwu: Two issues in this block.

**1. CRLF line endings.** All 8 added/modified lines in this file use CRLF while the rest of the file is LF. That is also why `target_compile_definitions(unit_tests PRIVATE ORTGENAI_CUDA_ADDON_COMPILATION)` shows up as modified further down -- it gained a bare `\r` and is otherwise unchanged. No other file in the PR is affected; please re-save with LF so the diff contains only the intended block.

**2. Duplicate translation unit.** Compiling this TU into `unit_tests` while it also lives in the DLL puts two independent copies of `Generators::StopStringMatcher` in one process, since `c_api_tests.cpp` exercises the DLL-side copy in the same binary. It works because no matcher object crosses the boundary, but `engine_unit_tests` already links `onnxruntime-genai-obj` and has full internal-symbol access. Moving `test/cpp/stop_string_matcher_tests.cpp` into `test/cpp/engine/` would remove both the platform branch and the duplication.

src/engine/request.cpp
baijumeswani: The new controller now resets `stop_controller_transaction_checkpoint_` to zero when it is installed. Terminal paths also release the previous controller, and new-turn checkpointing asserts that no prior controller remains.

src/engine/request.cpp
baijumeswani: The ordinary stop path now throws locally if `accepted_draft_count_` is nonzero, so a future draft-gating regression cannot silently skip matching. The comment also states that stop classification is formally ordered before EOS and explains why an appended EOS cannot reach this branch. Speculative verification observes accepted drafts in order and therefore does not use this guard.

src/engine/engine.cpp
baijumeswani: Tokenizer creation remains lazy so Engines that never use stop strings keep the existing fast path. The C API specification now documents that the first stop-enabled admission loads tokenizer assets and initializes ORT Extensions synchronously on the Engine owner thread, which can briefly delay other requests.

src/stop_string_controller.cpp
baijumeswani: The C API specification now states that one rollback is linear in each affected request’s committed turn length and that repeated retryable aborts can accumulate quadratic detokenization work across a growing turn. It also calls out the owner-thread and multi-request batch impact.

src/ort_genai_c.cpp
baijumeswani: Validation now uses a shared `ValidateStopStrings` function from both the setter and matcher constructor. The setter no longer copies into a temporary matcher or builds discarded KMP tables, and validation still happens before assignment so a rejected update preserves the previous configuration.

test/CMakeLists.txt
baijumeswani: The pure matcher tests now live under `test/cpp/engine/`, where `engine_unit_tests` links the production object library directly. The Windows-only duplicate matcher translation unit is gone. `test/CMakeLists.txt` was restored byte-for-byte to the merge-base version, removing the line-ending-only diff without normalizing unrelated existing CRLF content.

tianleiwu (approved):
Re-reviewed the current head after the follow-up fixes. The earlier lifecycle, draft-token, validation-cost, test-layout, and documentation threads are addressed and resolved.

Verified the incremental matcher and UTF-8/chunk-boundary behavior, transactional checkpoint/rollback/commit ordering, request terminal cleanup and invariants, speculative-path exclusion, and the C/C++/Python option and event contracts. I found no remaining correctness, performance, security, or released-compatibility issue.

Validation on freshly rebuilt current-head targets:
- 66/66 focused Engine stop-string, lifecycle, invariant, MTP-exclusion, and real-tokenizer tests passed.
- `CAPITests.EngineTurnOptionsStopStrings` passed.
- All 3 focused Python stop-string integration and validation tests passed.

Approved.

PR #2528, opened 2026-09-03
Support stop strings during speculative decoding

baijumeswani:
## Summary

Extend Engine stop-string handling through speculative target verification so MTP, DFlash2/DSpark, manually supplied drafts, and future draft producers share the same decoded-text termination behavior.

- Observe target-verified tokens in order and stop at the first decoded stop-string match.
- Retain, emit, cache, and count the token that completes the match while discarding the remaining speculative suffix.
- Keep Search state, host token history, paged and fixed-state cache commits, generated-token usage, and speculative telemetry aligned with the retained prefix.
- Restore sampled host RNG state to the retained boundary when candidates after a match are discarded.
- Prevent automatic drafters from proposing another block after the target request has completed.

The stop logic remains in the generic target-verification path rather than in individual draft producers. Matching semantics and public Engine behavior are unchanged from #2527.

## Correctness

Greedy and sampled verification now track accepted and evaluated draft counts explicitly. This distinguishes retained accepted tokens from proposals that were evaluated but rejected or discarded after a terminal boundary, ensuring cache accounting and telemetry describe the transaction that was actually committed.

Sampled verification records lightweight RNG checkpoints only for stop-enabled requests with active drafts. If a match truncates the speculative round, the request restores the checkpoint immediately after the retained matching token, so discarded candidates do not alter future sampling.

### DIFF
--- a/src/engine/engine.cpp
+++ b/src/engine/engine.cpp
@@ -917,8 +917,7 @@ void Engine::PublishMtpDrafts(MtpStep& step) {
 }
 
 void Engine::RecordSpeculativeCommit(const StepPlan& plan) noexcept {
-  for (size_t i = 0; i < plan.requests.size(); ++i) {
-    const auto& entry = plan.requests[i];
+  for (const auto& entry : plan.requests) {
     if (entry.draft_token_count == 0) {
       continue;
     }
--- a/src/engine/engine.cpp
+++ b/src/engine/engine.cpp
@@ -489,6 +489,14 @@ std::unique_ptr<Engine::MtpStep> Engine::PrepareMtpStep(
                                                 remaining_turn_tokens_after_step > 1
                                                     ? remaining_turn_tokens_after_step - 1
                                                     : size_t{0}});
+      // A stop-enabled request is intentionally not excluded here: generic target verification
+      // (Request::CommitAcceptedDraftsForTransaction for greedy, the per-stage loop in
+      // ScheduledRequests::GenerateNextTokensForTransaction for sampled/batched) truncates it
+      // transactionally through the same StopStringController the ordinary path uses, so it can
+      // draft and verify normally. result.done still does the work of preventing a redraft after a
+      // match: it is true for any committed StopString result exactly like any other finish
+      // reason, so this check below already skips proposing a new draft block for it with no
+      // stop-specific condition needed.
       if (!result.token_appended || result.done ||
           entry.request->DraftTokenValidationError() ||
           max_draft_tokens == 0) {
@@ -909,17 +917,28 @@ void Engine::PublishMtpDrafts(MtpStep& step) {
 }
 
 void Engine::RecordSpeculativeCommit(const StepPlan& plan) noexcept {
-  for (const auto& entry : plan.requests) {
+  for (size_t i = 0; i < plan.requests.size(); ++i) {
+    const auto& entry = plan.requests[i];
     if (entry.draft_token_count == 0) {
       continue;
     }
 
     const size_t accepted = entry.request->AcceptedDraftTokenCount();
+    // EvaluatedDraftTokenCount() is tracked directly at the actual verification source (the
+    // greedy loop in Request::CommitAcceptedDraftsForTransaction, and the sampled/batched stage
+    // loop's AppendAcceptedSampledToken/PromoteFinalStageAsAcceptedDraft/
+    // MarkFinalStageAsEvaluatedNonAcceptedDraft in
+    // ScheduledRequests::GenerateNextTokensForTransaction),
+    // not inferred here from accepted/finish_reason/token_appended after the fact. It counts draft
+    // positions logically resolved before the committed terminal boundary; greedy comparisons may
+    // have been precomputed for later positions. The old inference overcounted a limit-truncated
+    // round -- e.g. 5 proposed, budget for 2, both confirmed: the logical evaluated prefix is 2,
+    // not accepted + 1 = 3.
+    const size_t evaluated = entry.request->EvaluatedDraftTokenCount();
     ++speculative_stats_.rounds;
     ++speculative_stats_.completed_rounds;
     speculative_stats_.draft_tokens_proposed += entry.draft_token_count;
-    speculative_stats_.draft_tokens_evaluated +=
-        std::min(accepted + 1, entry.draft_token_count);
+    speculative_stats_.draft_tokens_evaluated += evaluated;
     speculative_stats_.draft_tokens_accepted += accepted;
     ++speculative_stats_.acceptance_length_histogram[accepted];
     if (accepted == 0) {
--- a/src/engine/request.cpp
+++ b/src/engine/request.cpp
@@ -277,7 +277,9 @@ void Request::MarkClosedFromEngine(const Engine& engine) noexcept {
   std::vector<int32_t>{}.swap(draft_tokens_);
   staged_draft_count_ = 0;
   accepted_draft_count_ = 0;
+  evaluated_draft_count_ = 0;
   draft_verification_completed_generation_ = false;
+  draft_verification_stop_match_index_ = -1;
 }
 
 void Request::CompleteCloseFromEngine(const Engine& engine) noexcept {
@@ -350,7 +352,9 @@ void Request::CompleteClose() noexcept {
   std::vector<int32_t>{}.swap(draft_tokens_);
   staged_draft_count_ = 0;
   accepted_draft_count_ = 0;
+  evaluated_draft_count_ = 0;
   draft_verification_completed_generation_ = false;
+  draft_verification_stop_match_index_ = -1;
   std::vector<int32_t>{}.swap(tokens_host_);
 }
 
@@ -385,15 +389,10 @@ int64_t Request::CommittedSequenceLength() const {
 }
 
 const char* Request::DraftTokenValidationError() const noexcept {
-  if (stop_controller_) {
-    // The shared boundary every draft producer converges on: manual Request::SetDraftTokens
-    // callers get this as a thrown error, and Engine::PrepareMtpStep's identical check silently
-    // excludes the request from automatic MTP drafting instead (see its use there). Either way, a
-    // stop-enabled turn always runs the plain one-token-per-step path, request-locally, without
-    // disabling draft verification for any other request on this Engine.
-    return "Speculative draft tokens are not supported while stop strings are enabled for the "
-           "active turn.";
-  }
+  // A stop-enabled turn is not excluded here: draft verification observes target-accepted tokens
+  // through stop_controller_ in exactly the same committed order the ordinary one-token path uses
+  // (CommitAcceptedDraftsForTransaction for greedy verification, StageGenerationForTransaction for
+  // sampled/batched verification), so it drafts and verifies normally.
   if (guidance_logits_processor_) {
     return "Speculative draft tokens are not supported with guidance.";
   }
@@ -470,7 +469,9 @@ void Request::AppendDraftsForTransaction(size_t draft_count) {
   tokens_host_.insert(tokens_host_.end(), drafts.begin(), drafts.end());
   staged_draft_count_ = draft_count;
   accepted_draft_count_ = 0;
+  evaluated_draft_count_ = 0;
   draft_verification_completed_generation_ = false;
+  draft_verification_stop_match_index_ = -1;
 }
 
 void Request::CommitAcceptedDraftsForTransaction(size_t accepted_count) {
@@ -484,9 +485,16 @@ void Request::CommitAcceptedDraftsForTransaction(size_t accepted_count) {
   search_->RewindTo(committed_length);
   staged_draft_count_ = 0;
   accepted_draft_count_ = 0;
+  evaluated_draft_count_ = 0;
   draft_verification_completed_generation_ = false;
+  draft_verification_stop_match_index_ = -1;
 
   for (size_t offset = 0; offset < accepted_count; ++offset) {
+    // Every iteration examines exactly one proposed draft position's target acceptance, whether or
+    // not it turns out to append/match/end the round below -- this is what lets
+    // Engine::RecordSpeculativeCommit read an exact "rows examined" count instead of inferring one
+    // from token_appended/finish_reason after the fact.
+    ++evaluated_draft_count_;
     const int32_t token = draft_tokens_[offset];
     const int64_t sequence_length_before = CurrentSequenceLength();
     search_->CommitToken(token);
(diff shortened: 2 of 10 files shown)


src/engine/scheduled_requests.cpp
```diff
@@ -760,39 +804,95 @@ void ScheduledRequests::GenerateNextTokensForTransaction(
...
+            // This stage is a genuinely target-confirmed draft (not a trailing replacement/bonus
+            // token) whenever it falls within the request's own confirmed prefix -- fold it into
+            // accepted_draft_count_ so CommitStep, cache/fixed-state CommitPrefix, and speculative
+            // telemetry all count it as accepted, exactly like an earlier accepted stage. When it
+            // instead falls at or beyond the confirmed prefix but still within the proposed draft
+            // range, it is a logically resolved but non-accepted proposal (either rejected, with
+            // its replacement in this result, or EOS without an append) -- count it as evaluated
+            // but not accepted. A stage beyond the whole proposed range (a trailing bonus token
+            // after full acceptance) is neither: it was never one of the proposed drafts.
+            if (stage < confirmed_draft_counts[request_index]) {
```
tianleiwu: **High / maintainability.** This ~35-line finalization block (promote / mark-evaluated / assign result / truncate `selected_tokens` / restore `rng_`) is duplicated verbatim in the `can_batch_commits == false` fallback below (lines ~872-891). `docs/paged_attention_engine.md` in this same PR states that no test in this repository exercises the `can_batch_commits == true` branch, because the CPU doubles cannot produce a `Search` that supports batched commit.

Two hand-synchronized copies of correctness-critical draft accounting, where the copy that ships on CUDA is the untested one, is a real divergence risk: a future fix applied to the tested branch will pass CI while silently leaving the CUDA branch wrong.

Suggest hoisting the shared tail into a lambda declared just before the stage loop, so both branches share one implementation:

```cpp
const auto finalize_stage = [&](size_t request_index, size_t stage,
                               const RequestStepResult& stage_result) {
  auto& request = requests_[request_index];
  const bool is_natural_last_stage = stage + 1 == selected_tokens[request_index].size();
  const bool stopped = stage_result.finish_reason == GenerationFinishReason::StopString;
  if (!stopped && !is_natural_last_stage) {
    if (!stage_result.token_appended || stage_result.done)
      throw std::logic_error("An accepted sampled draft unexpectedly ended the request.");
    request->AppendAcceptedSampledToken(stage_result.token);
    return;
  }
  // ... promote / mark / truncate / restore rng_, once ...
};
```

That also makes the "structural equivalence is not evidence of testing" caveat in the design doc much cheaper to keep true.

src/engine/scheduled_requests.cpp
```diff
@@ -760,39 +804,95 @@ void ScheduledRequests::GenerateNextTokensForTransaction(
...
+              // those now-invisible tokens by restoring the state captured right after this
+              // stage's own draw (a no-op for a request with no active stop controller, which
+              // never populated a checkpoint here).
+              selected_tokens[request_index].resize(stage + 1);
+              // rng_checkpoints itself is only ever sized when at least one request in this step
+              // needed it (see SelectSampledRows); a request whose own stop match reaches here
+              // always was one of them, but this stays a bounds check rather than an implicit
+              // invariant.
+              if (request_index < rng_checkpoints.size() &&
+                  stage < rng_checkpoints[request_index].size()) {
```
tianleiwu: **Suggestion / diagnosability.** The comment says this stays a bounds check rather than an implicit invariant, but the failure mode is a silent no-op: if the checkpoint is ever missing, `rng_` is simply not restored and the request's seeded stream diverges permanently, with no exception, log, or test signal. Silently non-reproducible sampling is one of the hardest defects to trace back here.

I verified the invariant does hold today (the pre-scan predicate in `SelectSampledRows` is identical to `checkpoint_rng`, and only random-sampled drafted requests populate `selected_tokens`), which is exactly why a violation would mean a programming error rather than a recoverable condition. The surrounding code already throws `std::logic_error` on impossible states, e.g. `"An accepted sampled draft unexpectedly ended the request."` a few lines below - please do the same here instead of falling through:

```cpp
if (request_index >= rng_checkpoints.size() ||
    stage >= rng_checkpoints[request_index].size()) {
  throw std::logic_error(
      "A stop-truncated sampled round is missing its RNG checkpoint.");
}
request->rng_ = rng_checkpoints[request_index][stage];
```

Same applies to the identical block in the non-batched branch below.

src/engine/scheduled_requests.cpp
```diff
@@ -353,6 +365,24 @@ std::vector<DeviceSpan<float>> ScheduledRequests::SelectSampledRows(
...
+      // later stop match truncates this round before its natural end, the caller restores rng_ to
+      // the checkpoint captured after the retained match, undoing every draw made for a now-
+      // discarded, never-visible token so those candidates do not advance the host verification
+      // RNG beyond the retained prefix. A
+      // request with no active stop controller never checkpoints (no reservation, no copies, no
+      // vector growth) and a step with no such request at all never even allocates the outer
+      // vector, so this adds no cost to the existing no-stop/greedy/non-drafted paths.
+      const bool checkpoint_rng = requests_[i]->stop_controller_ != nullptr;
+      if (checkpoint_rng) {
+        rng_checkpoints[i].reserve(draft_count + 1);
```
tianleiwu, line 384: **Suggestion / performance.** `std::mt19937` is ~2.5 KB of state, so this reserves and then fills `(draft_count + 1) * ~2.5 KB` per stop-enabled speculative request (≈15 KB at `draft_count = 5`), plus a full generator copy after **every** draw, and `rng_checkpoints` itself is a fresh `std::vector<std::vector<...>>` constructed per step inside `GenerateNextTokensForTransaction`. That is a per-step heap allocation and a multi-KB memcpy per token in the decode hot path.

The comment correctly argues the no-stop/greedy/non-drafted paths pay nothing, which is good, but the stop-enabled speculative path is exactly the one this PR is meant to make usable. Two cheaper shapes:

1. Move `rng_checkpoints` to a `ScheduledRequests` member scratch buffer (like `sampling_scratch`) and `clear()` rather than reconstruct, so both the outer and inner vector capacities survive across steps. Small change, keeps the current logic.
2. Snapshot `rng_` **once** before the draw loop and, on the rare truncation, restore that snapshot and re-run `stage + 1` draws (the selections are already recorded in `selected_tokens`). O(1) memory, and the extra work is paid only when a match actually truncates.

Option 1 alone would address the allocation churn; option 2 also removes the per-draw copies.

src/engine/scheduled_requests.h
```diff
@@ -116,11 +116,15 @@ struct ScheduledRequests {
...
+  // plus the target-distributed correction or bonus token to commit; confirmed_draft_counts[i] is
+  // that same request's confirmed prefix length alone (excluding the trailing correction/bonus),
+  // used by the caller to tell a confirmed final draft apart from a replacement/bonus token when a
+  // stop match or the turn/context limit ends verification on the request's last staged token.
   std::vector<DeviceSpan<float>> SelectSampledRows(
       std::vector<DeviceSpan<float>>& verify_rows,
       std::vector<std::vector<int32_t>>& selected_tokens,
-      std::vector<size_t>& accepted_draft_counts);
+      std::vector<size_t>& confirmed_draft_counts,
+      std::vector<std::vector<std::mt19937>>& rng_checkpoints);
```
tianleiwu, line 129: Nit: this header now uses `std::mt19937` in a public declaration but does not include `<random>` directly - it arrives transitively through `request.h` -> `generator/generators.h`. Please add the direct include so the header stays self-contained if that chain changes.

src/engine/engine.cpp
```diff
@@ -909,17 +917,28 @@ void Engine::PublishMtpDrafts(MtpStep& step) {
 }
 
 void Engine::RecordSpeculativeCommit(const StepPlan& plan) noexcept {
-  for (const auto& entry : plan.requests) {
+  for (size_t i = 0; i < plan.requests.size(); ++i) {
```
tianleiwu: Nit: the loop was converted from a range-`for` to an index loop, but `i` is only used to re-derive `entry`. Unless a follow-up needs the index, revert to `for (const auto& entry : plan.requests)` to keep the diff to the telemetry change.

src/engine/request.cpp
```diff
@@ -526,14 +556,32 @@ void Request::RewindDraftsForTransaction(size_t accepted_count) {
   search_->RewindTo(static_cast<size_t>(CurrentSequenceLength()) - rejected_count);
 }
 
-void Request::RecordSampledDraftAcceptance(size_t accepted_count) {
-  if (staged_draft_count_ != 0 || accepted_count > draft_tokens_.size()) {
-    throw std::logic_error("Sampled verification recorded an invalid accepted draft prefix.");
+void Request::AppendAcceptedSampledToken(int32_t token) {
+  if (accepted_draft_count_ >= draft_tokens_.size()) {
```
tianleiwu, line 560: Nit: this guards against `draft_tokens_.size()`, i.e. whatever the producer proposed via `SetDraftTokens`, not the count the scheduler actually staged for this step. `Scheduler` clamps `draft_token_count` to `min(PendingDraftTokenCount(), max_draft_token_count)`, so `draft_tokens_` can be strictly larger and this check would not catch over-acceptance past the staged prefix.

Unreachable today (the stage loop is bounded by `selected_tokens`, which is bounded by the staged `draft_count`), but the error message claims a stronger invariant than the check enforces. Comparing against the step's staged draft count would make the guard match its message.

tianleiwu (commented):
### Review summary

No correctness defect found. The design choice to keep stop-string truncation in the generic target-verification path (rather than in each draft producer) is the right one, and the test matrix is unusually thorough. Findings below are one maintainability/coverage risk, two perf/diagnosability suggestions, and a few nitpicks.

#### What I verified against the sources (not just the prose)

| Path | Result |
|---|---|
| Greedy `CommitAcceptedDraftsForTransaction` stop-match break | Search is rewound to `committed_length` and re-committed one token at a time, so no stale draft KV or host token survives the break. `staged_`/`accepted_`/`evaluated_draft_count_` all land on `k+1`. OK |
| `CommitStep`'s `target_cache_slots - (draft_token_count - accepted)` | Checked against `RequiredSlots(processed, 1 + D)`. Holds for full-accept+bonus, reject+replacement, greedy `StageDraftCompletion` (no append), and the **new** promotion case (`accepted = j+1`, `token_appended = false` => `processed = L+j+1`, exactly the KV positions the model actually computed). OK |
| Newly accurate `accepted_draft_count_` at intermediate sampled stages | This makes `StageGeneration`'s `generated_tokens_after_step` correct mid-round, which could in principle trip `TurnLimit` at a non-final stage and fall into `throw std::logic_error("An accepted sampled draft unexpectedly ended the request.")`. It cannot: `SelectSampledRows` caps `selected_tokens[i].size()` at `RemainingTurnTokenBudget()`, so the limit is only reachable on the last stage. OK |
| `rng_checkpoints` sizing invariant | The pre-scan predicate is identical to the per-request `checkpoint_rng` predicate, and only random-sampled drafted requests ever populate `selected_tokens`, so a stop match always finds its checkpoint. `reserve(draft_count + 1)` is the exact upper bound of the draw loop. OK |
| `active.empty()` early break | Newly required, since truncation can shrink the request that originally set `max_selected_tokens`. Correctly placed after `active` is built and does not starve other still-active requests. OK |
| Redraft prevention after a match | `Engine::PrepareMtpStep` (`!token_appended \|\| done`), `PublishDflash2Drafts` (`token_appended && !done`), and `Request::SetDraftTokens` (`IsExecuting`) all gate correctly for both the greedy result (`token_appended == false`) and the sampled-promotion result (`token_appended` cleared). OK |
| Stale "incompatible with drafts" wording | Grepped `src/`, `docs/`, and the public headers; only the two updated docs mention it, no stale public-header text. OK |

#### Cross-cutting concern: public telemetry semantics

`draft_tokens_evaluated` is public surface (C API `OgaGeneratorGetSpeculativeStat`, the Python `stats` dict, and `acceptance_rate = accepted / evaluated` at `src/engine/engine.cpp:2185`). Its Engine-path meaning changed here, but:

- `docs/SpeculativeDecoding.md:319` still describes it as "Proposed tokens examined before acceptance stopped" (that file is not in this diff, hence this is summary-only feedback);
- the same field name produced by `src/decoding/speculative_decoding_strategy.cpp` and `src/generator/mtp_generator.cpp` keeps the old definition.

Net effect: `acceptance_rate` can now be exactly `1.0` on a limit-truncated Engine round, which was previously impossible, and two producers of the same metric name now disagree. Worth updating the table and calling out the divergence so dashboards are not silently reinterpreted.

#### Positives

- The greedy matcher call is correctly nested inside the `sequence_length_after == sequence_length_before + 1` branch, so an accepted-but-unappended EOS draft never reaches the controller. That mirrors `StageGeneration`'s `token_appended` gate exactly, and it makes the removal of the old `accepted_draft_count_ != 0` guard consistent rather than a loosening.
- `PromoteFinalStageAsAcceptedDraft` asserting `result.token_appended` before mutating the host mirror is exactly the right place to enforce the `confirmed_draft_counts`/`token_appended` invariant, and it has a dedicated negative test.
- Moving `draft_tokens_evaluated` from post-hoc inference to source tracking is a genuine fix; `min(accepted + 1, draft_token_count)` really did over-count a budget-truncated round.
- Test coverage spans match at first/middle/last draft position, bonus-token match, rejected-draft replacement match, EOS non-append on both greedy and sampled paths, per-boundary telemetry, budget exhaustion, RNG-sequence preservation after truncation, mixed batches, and both direct and queued rollback during verification.
- `ScriptedDecoderIO`'s new mutual-exclusion check between `row_tokens` and `sampling_candidate_tokens` prevents a silently ambiguous fixture.

#### Minor

Several 8-20 line block comments were added inside hot functions (`PrepareMtpStep`, `RecordSpeculativeCommit`, `SelectSampledRows`) that largely restate what `docs/paged_attention_engine.md` now covers at length. Consider trimming those to a one-line pointer at the doc section.

src/engine/scheduled_requests.cpp
baijumeswani: The common finalization work now lives in one `commit_sampled_stage` lambda used by both commit paths. Draft counters, final-stage promotion, stop truncation, RNG restoration, and result publication therefore use the same code. The CUDA qualification note now describes only the device commit mechanism that remains different.

src/engine/scheduled_requests.cpp
baijumeswani: A missing RNG checkpoint now throws `std::logic_error` before the finalization bookkeeping is changed. The checkpoint should always exist for a stop-enabled sampled request; if that invariant is ever broken, the transaction fails and rolls back instead of silently changing the future sample sequence.

src/engine/scheduled_requests.cpp
baijumeswani: I kept the checkpoints local for now. This allocation happens only for stop-enabled, randomly sampled draft verification, and we do not have measurements showing that it affects decode performance. Reusing persistent scratch is the safe optimization if profiling shows a meaningful cost. Re-running random draws is not safe because it would depend on implementation details of the distribution.

src/engine/scheduled_requests.h
baijumeswani: `scheduled_requests.h` now includes `<random>` directly, so its `std::mt19937` declaration no longer depends on another header including it first.

src/engine/engine.cpp
baijumeswani: The loop is range-based again. The index was only used to retrieve the same entry and was not needed for the telemetry change.

src/engine/request.cpp
baijumeswani: I left this guard unchanged because the current stage loop is already bounded by the staged draft count, so this path cannot accept more staged drafts than the request verified. Comparing with the staged count would require keeping or passing extra per-step state after drafts are rewound. It is worth tightening if this method gains another caller or the stage-loop invariant changes, but it does not correct a reachable failure today.

tianleiwu (approved):
Re-reviewed the current head and the follow-up changes addressing the previous threads. The shared sampled-stage finalization now keeps the batched and fallback paths aligned, missing RNG checkpoints fail transactionally instead of silently changing sampling state, the header/include and documentation issues are resolved, and the remaining scoped explanations are reasonable. I found no new correctness issues. Rebuilt `engine_unit_tests`; all 55 `StopStringEngineTest.*` cases pass.

PR #2532, opened 2026-09-04
Enable per-turn generation options in the Engine

baijumeswani:
## Summary

This change completes the public C and C++ Engine API wiring for generation options.

Engine Request creation no longer accepts `OgaGeneratorParams`. A Request now owns only session-level policy, while each Turn carries its own sampling, token-limit, seed, stop-string, and guidance policy.

This makes the API ownership match the Engine lifecycle:

```mermaid
flowchart LR
    M["Model / Engine configuration<br/>scheduler, cache, provider, chunking,<br/>speculative strategy, model defaults"]
    R["RequestOptions<br/>max_session_tokens"]
    T["TurnOptions<br/>sampling, token limits, seed,<br/>stop strings, guidance"]
    E["Engine"]
    Q["Request<br/>resident conversation state"]
    U["Turn<br/>one generation operation"]

    M --> E
    E --> Q
    R --> Q
    Q --> U
    T --> U
```

## Why

The previous Engine Request API required an `OgaGeneratorParams` object even though most of that object belongs to the classic Generator flow or to one generation turn.

The Engine also exposed Turn option setters for temperature, top-p, top-k, seed, and guidance, but those setters were not implemented.

This change removes that mismatch. The classic Generator API remains unchanged.

## Public API change

Request creation changes from:

```c
OgaResult* OgaEngineCreateRequest(
    OgaEngine* engine,
    const OgaGeneratorParams* params,
    const OgaRequestOptions* options,
    OgaRequest** out);
```

to:

```c
OgaResult* OgaEngineCreateRequest(
    OgaEngine* engine,
    const OgaRequestOptions* options,
    OgaRequest** out);
```

`OgaRequestOptions` contains resident Request policy:

```c
OgaRequestOptionsSetMaxSessionTokens(...)
```

`OgaTurnOptions` contains generation policy:

```c
OgaTurnOptionsSetMaxGeneratedTokens(...)
OgaTurnOptionsSetMinGeneratedTokens(...)
OgaTurnOptionsSetDoSample(...)
OgaTurnOptionsSetTemperature(...)
OgaTurnOptionsSetTopP(...)
OgaTurnOptionsSetTopK(...)
OgaTurnOptionsSetRepetitionPenalty(...)
OgaTurnOptionsSetNoRepeatNgramSize(...)
OgaTurnOptionsSetSeed(...)
OgaTurnOptionsClearSeed(...)
OgaTurnOptionsSetStopStrings(...)
OgaTurnOptionsSetGuidance(...)
OgaTurnOptionsClearGuidance(...)
OgaTurnOptionsReset(...)
```

The C++ wrappers and Python bindings mirror this API.

## Request and Turn lifecycle

A Request remains resident across turns so it can reuse its committed sequence and cache. Generation policy is resolved independently for every new Turn.

```mermaid
stateDiagram-v2
    [*] --> Unassigned: Create Request
    Unassigned --> Assigned: Begin first Turn
    Assigned --> Active: Scheduled
    Active --> Active: Commit generation step
    Active --> TurnComplete: EOS, stop string, limit, cancel, or failure
    TurnComplete --> Assigned: Begin continuation Turn
    Unassigned --> Closed: Close
    Assigned --> Closed: Close
    Active --> Closed: Close
    TurnComplete --> Closed: Close
    Closed --> [*]: Destroy handle
```

An omitted scalar Turn option uses the model-configured default. It never inherits the value used by the previous Turn.

Stop strings and guidance are disabled when omitted.

Seed is intentionally different: an omitted seed continues the Request's existing random stream, while an explicit seed reseeds that Turn.

## Turn admission is transactional

The complete Turn policy is resolved and validated before the Request is changed.

Grammar construction and stop-string controller construction also happen before mutation. A bad grammar, invalid sampling combination, unsupported option, or impossible token limit leaves the Request reusable.

```mermaid
flowchart TD
    A["BeginTurn(input, options)"]
    B["Resolve model defaults + explicit Turn overrides"]
    C["Validate sampling, limits, device support, and static-path support"]
    D["Build stop-string controller"]
    E["Compile or acquire guidance grammar"]
    F["Checkpoint Request admission state"]
    G["Append prompt and reserve scheduler/cache state"]
    H["Commit Turn policy and resources"]
    X["Return error<br/>Request remains unchanged"]
    Y["Turn admitted"]

    A --> B --> C
    C -->|invalid| X
    C --> D
    D -->|failure| X
    D --> E
    E -->|failure| X
    E --> F --> G
    G -->|failure| X
    G --> H --> Y
```

## Per-turn sampling policy

`EffectiveTurnPolicy` is the single resolved policy used by ordinary decoding, batched sampling, speculative verification, MTP eligibility, and DFlash2 eligibility.

Greedy selection is classified consistently as:

```cpp
!do_sample || top_k == 1 || temperature == 0
```

Invalid values and contradictory options are rejected at admission instead of being accepted and ignored.

Examples:

- `do_sample=false` with `temperature=0.7` is rejected because the temperature would not affect greedy output.
- `top_k=1` and `temperature=0` are accepted ways to request greedy output.
- `do_sample=true` is rejected if a model default still forces greedy output and the Turn did not override that default.
- A sampled Turn requires a positive top-k or top-p bound.
- No-repeat n-gram is rejected before execution on scoring devices whose Search implementation cannot apply it.

## Session and Turn token limits

`max_session_tokens` is the one Request-level sequence limit. Search storage, static cache sizing, speculative bounds, and the `MaxSessionTokens` finish reason use the same value.

An explicit `max_session_tokens` may lower the model-configured `search.max_length`, but it may not exceed it.

`max_generated_tokens` limits only the current Turn.

`min_generated_tokens` masks EOS until the current Turn has generated the requested number of tokens. It does not override stop strings, cancellation, failure, guidance termination, or maximum limits.

The Engine rejects model configurations that it would otherwise have to reinterpret or silently override:

- Nonzero legacy `search.min_length`; use per-turn `min_generated_tokens`.
- `search.num_beams` other than 1; the Engine batches Requests rather than beam rows.
- Nonpositive `search.max_length`.
- Nonzero `search.chunk_size` with static batching.

The error messages describe the corresponding config overlay when one is available.

## Transactional seed handling

The Request owns a durable 64-bit seed basis.

An explicit Turn seed is applied only inside a rollback-capable generation transaction, after Request and sampler state have been checkpointed and before any host or device RNG consumer runs.

```mermaid
flowchart TD
    A["Turn admits explicit seed"]
    B["Store pending reseed"]
    C["Checkpoint Search, host RNG, and device sampler state"]
    D["Apply pending seed to host and device streams"]
    E["Sample or verify drafts"]
    F{"Step result"}
    G["Commit tokens and sampler state"]
    H["Promote seed to durable Request basis"]
    I["Restore tokens, Search, host RNG, and device RNG"]
    J["Keep pending seed for identical retry"]

    A --> B --> C --> D --> E --> F
    F -->|commit| G --> H
    F -->|rollback| I --> J
    J --> C
```

Seed zero is valid.

Seeds that fit in 32 bits preserve the existing host `std::mt19937` sequence. Wider seeds use both 32-bit halves, and CUDA receives the full 64-bit value.

The CUDA sampler state is reseeded in place so pooled sampler slots are not released and reacquired.

Because the `BatchedSampler` interface crosses the separately loaded CUDA add-on boundary, the internal device interface version is increased from 4 to 5. Core and CUDA add-on artifacts must come from matching builds.

## Turn-scoped guidance

Guidance is compiled or acquired before admission mutation and installed only when the Turn commits.

The grammar cursor participates in step rollback.

Every terminal path releases the Turn's guidance state.

This supports a guided tool-call Turn followed by an unguided tool-result Turn on the same Request:

```mermaid
sequenceDiagram
    participant App
    participant Request
    participant Engine

    App->>Request: BeginTurn(prompt, guidance)
    Request->>Engine: Admit guided Turn
    Engine-->>App: Tool-call tokens and terminal event
    App->>Request: BeginTurn(tool result, no guidance)
    Request->>Engine: Admit unguided continuation
    Engine-->>App: Normal generated tokens
```

Guided turns do not use speculative drafts because draft rows cannot reproduce the grammar processor. A later unguided Turn becomes draft-eligible again.

## Stop strings and speculative decoding

The decoded UTF-8 stop-string behavior already present in the Engine is preserved.

Stop strings remain Turn-scoped and transactional.

Speculative verification stops at the first matching accepted prefix while keeping token history, cache state, usage, finish reason, and matched stop-string index consistent.

Draft proposals are cleared at every terminal Turn boundary so a proposal validated under one Turn's policy cannot cross into a later Turn with different guidance, minimum-token, repetition, or n-gram settings.

## DFlash2 interaction

DFlash2 drafting remains greedy-only, but greediness now comes from the resolved Turn policy.

An untracked Request can join the DFlash2 ring pool only on a draft-eligible position-zero step.

Once a Request owns a ring, sampled turns continue feeding committed context without requesting drafts. This preserves cache continuity and allows a later greedy Turn to resume drafting.

DFlash2 forward-pass statistics increment only when the drafter session actually runs.

## Static batching

Static batching keeps its existing non-transactional execution path.

Options that require transactional state are rejected before Request mutation on this path:

- Per-turn seed.
- Stop strings.

Static Engine construction also rejects model-configured prefill chunking.

Other supported Turn sampling controls use the same resolved policy as dynamic batching.

## Classic Generator impact

The classic Generator public API and generation behavior are unchanged.

`OgaGeneratorParams`, `OgaCreateGenerator`, `OgaGenerator`, and MTP Generator continue to work as before.

The only shared operational impact is the CUDA add-on interface version change. A mismatched older CUDA add-on is rejected instead of being loaded with an incompatible sampler vtable.

## Caller migration

Before:

```cpp
params->SetSearchOption("do_sample", false);
params->SetSearchOption("max_length", 4096);
auto request = engine->CreateRequest(*params);
request->BeginTurn(tokens);
```

After:

```cpp
auto request_options = OgaRequestOptions::Create();
request_options->SetMaxSessionTokens(4096);
auto request = engine->CreateRequest(request_options.get());

auto turn_options = request->CreateTurnOptions();
turn_options->SetDoSample(false);
request->BeginTurn(tokens, turn_options.get());
```

Python `Engine.create_request` now accepts only the optional keyword argument `options`. Passing the old positional `GeneratorParams` fails clearly instead of being reinterpreted.

Examples, existing benchmark scenarios, C API tests, C++ tests, and Python tests are migrated to the new ownership model.

The benchmark changes are API migrations only. This PR does not add benchmark scenarios or metrics.

### DIFF
--- a/benchmark/engine/scenarios/decode_baseline.cpp
+++ b/benchmark/engine/scenarios/decode_baseline.cpp
@@ -87,13 +87,15 @@ ScenarioExecutionOutput DecodeBaselineScenario::Execute(const ScenarioConfig& co
   for (int run = 0; run < total_runs; ++run) {
     const bool is_warmup = run < config.warmup_runs;
 
-    std::vector<std::unique_ptr<OgaGeneratorParams>> params;
+    std::vector<std::unique_ptr<OgaRequestOptions>> request_options;
+    std::vector<std::unique_ptr<OgaTurnOptions>> turn_options;
     std::vector<RequestRunState> request_states(static_cast<size_t>(config.concurrency));
     std::vector<std::unique_ptr<OgaRequest>> requests;
     std::unordered_map<const OgaRequest*, RequestRunState*>
         request_states_by_request;
 
-    params.reserve(static_cast<size_t>(config.concurrency));
+    request_options.reserve(static_cast<size_t>(config.concurrency));
+    turn_options.reserve(static_cast<size_t>(config.concurrency));
     requests.reserve(static_cast<size_t>(config.concurrency));
 
     // Start timing before submission so TTFT captures arrival-to-first-token, including admission.
@@ -101,21 +103,26 @@ ScenarioExecutionOutput DecodeBaselineScenario::Execute(const ScenarioConfig& co
     size_t generated_tokens = 0;
 
     for (int i = 0; i < config.concurrency; ++i) {
-      params.emplace_back(OgaGeneratorParams::Create(*engineResources.model));
-      const size_t max_length = prompt_token_count + static_cast<size_t>(config.generation_tokens);
-      params.back()->SetSearchOption(
-          "max_length", static_cast<double>(max_length));
-      params.back()->SetSearchOption("random_seed", kRandomSeed);
+      request_options.emplace_back(OgaRequestOptions::Create());
+      const size_t max_session_tokens =
+          prompt_token_count + static_cast<size_t>(config.generation_tokens);
+      request_options.back()->SetMaxSessionTokens(max_session_tokens);
 
       auto& request_state = request_states[static_cast<size_t>(i)];
       request_state.tokens.assign(
           prompt_tokens->SequenceData(0), prompt_tokens->SequenceData(0) + prompt_token_count);
 
-      requests.emplace_back(engineResources.engine->CreateRequest(*params.back()));
+      requests.emplace_back(
+          engineResources.engine->CreateRequest(request_options.back().get()));
       request_states_by_request.emplace(
           requests.back().get(), &request_state);
+      turn_options.push_back(requests.back()->CreateTurnOptions());
+      if (engineResources.uses_dynamic_batching) {
+        turn_options.back()->SetSeed(kRandomSeed);
+      }
       requests.back()->BeginTurn(
-          prompt_tokens->SequenceData(0), prompt_token_count);
+          prompt_tokens->SequenceData(0), prompt_token_count,
+          turn_options.back().get());
     }
 
     auto event_buffer = engineResources.engine->CreateEventBuffer(
--- a/src/dflash2_drafter.cpp
+++ b/src/dflash2_drafter.cpp
@@ -461,8 +461,8 @@ Tensor& Dflash2StepTensor(std::unique_ptr<Tensor>& slot, DeviceInterface* device
   return *slot;
 }
 
-bool Dflash2CanDraft(const Config::Search& search) {
-  return !search.do_sample || search.top_k == 1 || search.temperature == 0;
+bool Dflash2CanJoin(bool draft_eligible, size_t first_position) noexcept {
+  return draft_eligible && first_position == 0;
 }
 
 void Dflash2Drafter::AllocateCache() {
@@ -490,12 +490,11 @@ bool Dflash2Drafter::Admit(const Feed& feed) {
   if (requests_.find(feed.request) != requests_.end()) {
     return true;
   }
-  // The drafter cannot backfill K/V for context whose auxiliary hidden states have already been
-  // consumed, so a request can only join from the start of its sequence. Requests that arrive when
-  // the ring pool is full decode without DFlash 2 drafts instead of taking the whole drafter down.
-  if (feed.first_position != 0) {
+  if (!Dflash2CanJoin(feed.draft_eligible, feed.first_position)) {
     return false;
   }
+  // Requests that arrive when the ring pool is full decode without DFlash 2 drafts instead of
+  // taking the whole drafter down.
   if (ring_blocks_ != 0) {
     if (free_blocks_.size() < ring_blocks_) {
       ++admission_misses_;
@@ -545,11 +544,11 @@ void Dflash2Drafter::ReleaseAll() {
   requests_.clear();
 }
 
-void Dflash2Drafter::Propose(Tensor& aux_hidden_states, std::span<const Feed> feeds,
+bool Dflash2Drafter::Propose(Tensor& aux_hidden_states, std::span<const Feed> feeds,
                              std::vector<std::vector<int32_t>>& drafts) {
   drafts.assign(feeds.size(), {});
   if (feeds.empty()) {
-    return;
+    return false;
   }
 
   const size_t block_size = static_cast<size_t>(config_.block_size);
@@ -565,7 +564,7 @@ void Dflash2Drafter::Propose(Tensor& aux_hidden_states, std::span<const Feed> fe
     }
   }
   if (served.empty()) {
-    return;
+    return false;
   }
 
   // Batch layout. Every served feed contributes its context rows so the drafter cache never
@@ -821,7 +820,7 @@ void Dflash2Drafter::Propose(Tensor& aux_hidden_states, std::span<const Feed> fe
   }
 
   if (!drafts_wanted) {
-    return;
+    return true;
   }
 
   // The spans own the host mirrors these point into, so they must outlive the reads below.
@@ -842,6 +841,7 @@ void Dflash2Drafter::Propose(Tensor& aux_hidden_states, std::span<const Feed> fe
       previous = best;
     }
   }
+  return true;
 }
 
 }  // namespace Generators
--- a/src/engine/engine.cpp
+++ b/src/engine/engine.cpp
@@ -4,6 +4,7 @@
 #include "engine.h"
 #include "../logging.h"
 #include "../search.h"
+#include "../constrained_logits_processor.h"
 #include "../models/preprocessing/genai_tokenizer.h"
 #include "../stop_string_controller.h"
 
@@ -22,23 +23,6 @@ constexpr size_t kDflash2FailureDisableThreshold = 3;
 struct MtpRollbackError : std::runtime_error {
   using std::runtime_error::runtime_error;
 };
-
-std::shared_ptr<GeneratorParams> CloneRequestParams(
-    const GeneratorParams& source,
-    const Model& model) {
-  auto copy = std::make_shared<GeneratorParams>(model);
-  copy->search = source.search;
-  copy->speculative = source.speculative;
-  copy->max_batch_size = source.max_batch_size;
-  copy->use_graph_capture = source.use_graph_capture;
-  copy->max_graph_capture_length = source.max_graph_capture_length;
-  copy->use_multi_profile = source.use_multi_profile;
-  copy->p_device = source.p_device;
-  copy->guidance_type = source.guidance_type;
-  copy->guidance_data = source.guidance_data;
-  copy->guidance_ff_tokens_enabled = source.guidance_ff_tokens_enabled;
-  return copy;
-}
 
 std::string AddExceptionCause(std::string message, std::exception_ptr error) {
   if (!error) {
@@ -252,6 +236,15 @@ void ValidateMtpModelCompatibility(const Config& config,
 }
 
 EngineDependencies Engine::CreateDependencies(std::shared_ptr<Model> model) {
+  // The static batch decoder rebuilds its contiguous cache from the whole sequence every step, so
+  // it cannot resume a half-written prompt. Reject a model that configures chunking here rather
+  // than surfacing it once a Request is already being admitted.
+  if (!model->config_->engine.dynamic_batching &&
+      model->config_->search.chunk_size.value_or(0) != 0) {
+    throw std::runtime_error(
+        "search.chunk_size requires dynamic batching; the static batch scheduler cannot chunk a "
+        "prefill.");
+  }
   std::shared_ptr<DecoderOnly_Model> mtp_model;
   size_t mtp_bytes_per_block = 0;
   if (!model->config_->model.mtp.filename.empty()) {
@@ -367,13 +360,7 @@ void Engine::PrepareDflash2Feeds(const StepPlan& plan,
   dflash2_draft_widths_.reserve(plan.requests.size());
   for (size_t i = 0; i < plan.requests.size(); ++i) {
     const auto& entry = plan.requests[i];
-    const auto& search = entry.request->SearchOptions();
-    // Feeding a request that can never take a block would only burn a ring an eligible request
-    // could use, and a request skipped here is skipped for its whole life, so it never joins the
-    // drafter's contiguity contract.
-    if (!Dflash2CanDraft(search)) {
-      continue;
-    }
+    const bool greedy = entry.request->TurnPolicy().IsGreedy();
     const size_t accepted = entry.request->AcceptedDraftTokenCount();
     if (accepted > entry.draft_token_count) {
       throw std::logic_error("DFlash 2 observed more accepted drafts than the target planned.");
@@ -391,12 +378,12 @@ void Engine::PrepareDflash2Feeds(const StepPlan& plan,
     feed.aux_row_begin = entry.packed_token_offset;
     feed.aux_row_count = valid_rows;
     feed.first_position = first_position;
+    feed.draft_eligible = greedy;
 
     // The committed length this step ends at: the accepted prefix plus the token just sampled.
     const int64_t length_after_step = static_cast<int64_t>(first_position + valid_rows) +
                                       (results[i].token_appended ? 1 : 0);
-    const size_t sequence_limit =
-        std::min(static_cast<size_t>(search.max_length), entry.request->MaxTotalTokens());
+    const size_t sequence_limit = entry.request->MaxSessionTokens();
     const size_t remaining_turn_tokens = entry.request->RemainingTurnTokenBudget();
     const size_t remaining_turn_tokens_after_step =
         results[i].visible_token_count < remaining_turn_tokens
@@ -406,7 +393,7 @@ void Engine::PrepareDflash2Feeds(const StepPlan& plan,
         max_drafts, static_cast<size_t>(entry.request->SpeculativeOptions().max_draft_tokens),
         static_cast<size_t>(length_after_step), sequence_limit,
         remaining_turn_tokens_after_step);
-    feed.wants_drafts = width > 0 && results[i].token_appended && !results[i].done &&
+    feed.wants_drafts = greedy && width > 0 && results[i].token_appended && !results[i].done &&
                         !entry.request->DraftTokenValidationError();
     feed.anchor_token = results[i].token;
     dflash2_feeds_.push_back(feed);
@@ -423,8 +410,9 @@ void Engine::PublishDflash2Drafts(ScheduledRequests& scheduled_requests) {
     throw std::logic_error("The main decoder did not expose auxiliary hidden states for DFlash 2.");
   }
 
-  dflash2_drafter_->Propose(*aux_hidden_states, dflash2_feeds_, dflash2_drafts_);
-  ++speculative_stats_.draft_forward_passes;
+  if (dflash2_drafter_->Propose(*aux_hidden_states, dflash2_feeds_, dflash2_drafts_)) {
+    ++speculative_stats_.draft_forward_passes;
+  }
   for (size_t i = 0; i < dflash2_feeds_.size(); ++i) {
     auto& drafts = dflash2_drafts_[i];
     if (drafts.empty()) {
@@ -553,12 +541,9 @@ std::unique_ptr<Engine::MtpStep> Engine::PrepareMtpStep(
     for (size_t i = 0; i < target_plan.requests.size(); ++i) {
       const auto& entry = target_plan.requests[i];
       const auto& result = target_results[i];
-      const auto& search = entry.request->SearchOptions();
       const int64_t committed_length_after_step =
           entry.request->CurrentSequenceLength();
-      const size_t sequence_limit =
-          std::min(static_cast<size_t>(search.max_length),
-                   entry.request->MaxTotalTokens());
+      const size_t sequence_limit = entry.request->MaxSessionTokens();
       const size_t remaining_turn_tokens =
           entry.request->RemainingTurnTokenBudget();
       const size_t remaining_turn_tokens_after_step =
@@ -610,10 +595,8 @@ std::unique_ptr<Engine::MtpStep> Engine::PrepareMtpStep(
         checkpointed_shadows.push_back(feed.shadow);
         feed.shadow->AppendTokensForAuxiliaryDecoder(feed.tokens);
       } else {
-        auto params = CreateGeneratorParams(*mtp_model_);
-        params->search = search;
         feed.shadow = Request::CreateAuxiliaryDecoderRequest(
-            std::move(params), entry.request->MaxTotalTokens(),
+            *mtp_model_, entry.request->MaxSessionTokens(),
             abandonment_pending_, shared_from_this(), feed.tokens);
         feed.newly_created = true;
       }
@@ -1041,35 +1024,59 @@ void Engine::ValidateOwnerThread() const {
   }
 }
 
-std::shared_ptr<Request> Engine::CreateRequest(const GeneratorParams& params,
-                                               size_t max_total_tokens) {
+std::shared_ptr<Request> Engine::CreateRequest(const RequestOptions& options) {
   ValidateOwnerThread();
   CompleteNonresidentClosedRequests();
   ReclaimAbandonedRequests();
   if (health_ == EngineHealth::Unhealthy) {
     std::rethrow_exception(fatal_error_);
   }
-  if (params.model_.get() != model_.get()) {
+  const auto& model_search = model_->config_->search;
+  if (model_search.max_length <= 0) {
     throw std::runtime_error(
-        "Engine request parameters must belong to the Engine's model.");
+        "The model's search.max_length must be greater than zero; actual value is " +
+        std::to_string(model_search.max_length) + ".");
   }
-  if (params.search.max_length <= 0) {
+  // The Engine expresses a per-turn floor through TurnOptions::min_generated_tokens. A model that
+  // still carries the legacy session-absolute floor would silently mean something else here, so it
+  // is rejected rather than reinterpreted -- but the caller can clear it without editing the model
+  // directory, through the config overlay that OgaCreateConfig/OgaConfigOverlay already support.
+  if (model_search.min_length != 0) {
     throw std::runtime_error(
-        "max_length must be greater than zero; actual value is " +
-        std::to_string(params.search.max_length) + ".");
+        "The model configures a nonzero search.min_length (" +
+        std::to_string(model_search.min_length) +
+        "), which is a session-absolute floor the Engine does not support; its minimum is per "
+        "turn. Clear it on the Config before creating the model -- "
+        R"(OgaConfigOverlay(config, "{\"search\":{\"min_length\":0}}"), or in Python )"
+        R"(config.overlay('{"search": {"min_length": 0}}') -- and use )"
+        "OgaTurnOptionsSetMinGeneratedTokens for a per-turn minimum instead.");
   }
-  if (max_total_tokens == 0 ||
-      max_total_tokens > static_cast<size_t>(params.search.max_length)) {
+  // Beam search does not implement the Engine's deferred completion contract -- its next tokens are
+  // never copied back -- and a Request is one sequence, so the Engine forces num_beams to 1 in the
+  // private search parameters it derives. Silently forcing a model that asked for beams would
+  // decode something the caller never requested, so a model-configured value other than 1 is
+  // rejected here instead. The overlay clears it without editing the model directory.
+  if (model_search.num_beams != 1) {
     throw std::runtime_error(
-        "max_total_tokens (" + std::to_string(max_total_tokens) +
-        ") must be greater than zero and no greater than max_length (" +
-        std::to_string(params.search.max_length) + ").");
+        "The model configures search.num_beams = " +
+        std::to_string(model_search.num_beams) +
+        ", and beam search is not supported by the Engine; every Engine Request decodes one "
+        "sequence with one beam. Clear it on the Config before creating the model -- "
+        R"(OgaConfigOverlay(config, "{\"search\":{\"num_beams\":1}}"), or in Python )"
+        R"(config.overlay('{"search": {"num_beams": 1}}') -- )"
+        "and batch across Requests instead.");
   }
-
+  const size_t model_ceiling = static_cast<size_t>(model_search.max_length);
+  const size_t max_session_tokens =
+      options.max_session_tokens.value_or(model_ceiling);
+  if (max_session_tokens == 0 || max_session_tokens > model_ceiling) {
+    throw std::runtime_error(
+        "max_session_tokens (" + std::to_string(max_session_tokens) +
+        ") must be greater than zero and no greater than the model-configured search.max_length (" +
+        std::to_string(model_ceiling) + ").");
+  }
   auto request = std::make_shared<Request>(
-      CloneRequestParams(params, *model_), max_total_tokens,
-      abandonment_pending_);
-  request->ValidateEngineCompatibility();
+      *model_, max_session_tokens, abandonment_pending_);
   auto engine = shared_from_this();
 
   // Fatal handling preserves every token event from the current step, then needs one terminal
@@ -1105,20 +1112,39 @@ uint64_t Engine::BeginTurn(const std::shared_ptr<Request>& request,
         "Cannot begin a turn for a request that does not belong to this engine.");
   }
   request->ValidateTurnAdmission(tokens, options);
+
+  // Resolve and validate this turn's complete policy before any Request mutation, so an
+  // unsatisfiable combination leaves the previous completed turn entirely reusable.
+  const bool dynamic_batching = cache_manager_->SupportsDynamicBatching();
+  const auto policy = ResolveTurnPolicy(model_->config_->search, options);
+  const size_t turn_prompt_length =
+      static_cast<size_t>(request->IsAwaitingFirstTurn()
+                              ? 0
+                              : request->CurrentSequenceLength()) +
+      tokens.size();
+  ValidateTurnPolicy(policy, options, model_->p_device_scoring_->GetType(),
+                     turn_prompt_length, request->MaxSessionTokens());
   // Static batching completes generation through a non-transactional path (RunStatic) that cannot
-  // stage, roll back, or replay a stop match. Reject before any Request mutation rather than
-  // letting a stop-enabled turn run without the guarantee its finish reason promises.
-  if (!options.stop_strings.empty() && !cache_manager_->SupportsDynamicBatching()) {
+  // stage, roll back, or replay a stop match, and its sampler RNG state is created once when the
+  // Request joins the batch rather than inside a rollback-capable step. Reject both before any
+  // Request mutation rather than letting a turn run without the guarantee its options promise.
+  if (!dynamic_batching) {
+    if (!options.stop_strings.empty()) {
+      throw std::runtime_error(
+          "Stop strings require an Engine configured for dynamic batching.");
+    }
+    if (options.seed) {
+      throw std::runtime_error(
+          "A per-turn seed requires an Engine configured for dynamic batching.");
+    }
+  }
+  if (options.seed && !scheduler_->SupportsTransactionalSamplerState()) {
     throw std::runtime_error(
-        "Stop strings require an Engine configured for dynamic batching.");
+        "A per-turn seed requires a batched sampler that supports transaction rollback.");
   }
 
   const bool first_turn = request->IsAwaitingFirstTurn();
-  if (first_turn) {
-    if (cache_manager_->SupportsDynamicBatching()) {
-      request->ValidateEngineCompatibility();
-    }
-  } else {
+  if (!first_turn) {
     const bool restartable_canceled_turn =
         request->IsRestartableCanceledTurn() &&
         !cache_manager_->IsResident(request);
@@ -1129,19 +1155,21 @@ uint64_t Engine::BeginTurn(const std::shared_ptr<Request>& request,
   }
 
   RequestTurnAdmission admission;
+  admission.policy = policy;
   bool added_to_scheduler = false;
 
-  // Build this turn's complete stop controller (or leave admission.pending_stop_controller null
-  // for a no-stop turn) before any Request or Engine mutation: tokenizer/stream construction can
-  // throw, and at this point nothing has been touched yet, so a failure here leaves the Request,
-  // its MTP shadow, and Engine health entirely unchanged for a caller to retry. Ownership passes
-  // into admission; Request::CommitTurnAdmission() installs it as the live stop_controller_, and
-  // Request::RollbackTurnAdmission() (or simply this function returning through the catch below)
-  // discards it without ever having touched the Request's actual stop_controller_.
+  // Build this turn's complete stop controller and guidance processor (or leave either null for a
+  // turn that does not use them) before any Request or Engine mutation: tokenizer/stream
+  // construction and grammar compilation can throw, and at this point nothing has been touched yet,
+  // so a failure here leaves the Request, its MTP shadow, and Engine health entirely unchanged for
+  // a caller to retry. Ownership passes into admission; Request::CommitTurnAdmission() installs
+  // them, and Request::RollbackTurnAdmission() (or simply this function returning through the catch
+  // below) discards them without ever having touched the Request's live state.
   if (!options.stop_strings.empty()) {
     admission.pending_stop_controller = std::make_unique<StopStringController>(
         GetOrCreateStopTokenizer(), options.stop_strings);
   }
+  admission.pending_guidance = CreateTurnGuidance(options);
 
   // A continuation appends a prompt the MTP shadow never sees, so keeping the shadow would leave it
   // a concatenation of generated tokens across turns with the intervening prompt missing. Drop it
@@ -1716,25 +1744,6 @@ void Engine::RunDynamic() {
     };
 
     try {
-      for (const auto& entry : step_plan_.requests) {
-        entry.request->ValidateEngineCompatibility();
-      }
-    } catch (...) {
-      const auto validation_error = std::current_exception();
-      rollback_transaction();
-      ++transaction_metrics_.post_processing_aborts;
-      ++transaction_metrics_.retryable_aborts;
-      throw EngineStepError{
-          {StepOutcomeKind::RetryableBatchAbort,
-           step_plan_.transaction_id,
-           nullptr},
-          AddExceptionCause(
-              "Request validation failed; the batch was rolled back.",
-              validation_error),
-      };
-    }
-
-    try {
       // Sampling mutates each request's Search state. Checkpoint it before the model run so failures
       // in execution or post-processing can discard the whole batch rather than partially advancing
       // whichever requests happened to finish first.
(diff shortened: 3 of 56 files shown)


src/engine/engine.cpp
```diff
@@ -1075,20 +1082,35 @@ uint64_t Engine::BeginTurn(const std::shared_ptr<Request>& request,
...
+  // Request joins the batch rather than inside a rollback-capable step. Reject both before any
+  // Request mutation rather than letting a turn run without the guarantee its options promise.
+  if (!dynamic_batching) {
+    if (!options.stop_strings.empty()) {
+      throw std::runtime_error(
+          "Stop strings require an Engine configured for dynamic batching.");
+    }
+    if (options.seed) {
+      throw std::runtime_error(
+          "A per-turn seed requires an Engine configured for dynamic batching.");
```
tianleiwu, line 1141: A per-turn seed is gated on dynamic batching, but not on the sampler's transaction support.

`ScheduledRequests::BeginTransaction()` only builds `checkpointed_sampler_states_` inside `if (batched_sampler_ && batched_sampler_->SupportsTransactions())`, and `SupportsTransactions()` defaults to `false` on the base class. A dynamic-batching device that ever supplies a `BatchedSampler` without transaction support would skip the reseed checkpoint entirely, and the first step for a request that already owns a device state would throw `std::logic_error` from `Request::ApplyPendingSeedForTransaction()` — a step failure rather than an admission-time rejection.

Only CUDA implements `BatchedSampler` today, so this is unreachable right now. But the comment in `ApplyPendingSeedForTransaction()` asserts unreachability ("an internal contract violation, not a reachable configuration") without naming the coupling that actually provides it. Consider gating here on the sampler as well, e.g. reject when `scheduler_->GetBatchedSampler()` is non-null and `!SupportsTransactions()`.

src/engine/scheduled_requests.cpp
```diff
@@ -691,6 +704,24 @@ void ScheduledRequests::GenerateNextTokensForTransaction(
...
 
+  // Strictly after every Request and sampler checkpoint taken in BeginTransaction() and strictly
+  // before the first host or device RNG consumer below, so a rolled back step restores both streams
+  // and leaves the reseed pending for the retry. A device state is reseeded only when this step
+  // checkpointed it; BeginTransaction() checkpoints every state a pending reseed targets.
+  for (const auto& request : requests_) {
+    if (!request->IsChunkComplete()) {
+      continue;
+    }
+    const bool device_state_checkpointed =
```
tianleiwu, line 715: Hot-path suggestion: this loop runs for every chunk-complete request on every transactional step, and evaluates `ExistingSamplingState()` (which does a `dynamic_cast` inside `OwnsState`) plus a linear `std::find` over `checkpointed_sampler_states_` *before* calling `ApplyPendingSeedForTransaction()`, which then returns immediately when `pending_reseed_` is unset — the overwhelmingly common case.

Hoisting the cheap check to the top removes both per request per step:

```cpp
for (const auto& request : requests_) {
  if (!request->HasPendingTurnSeed() || !request->IsChunkComplete()) {
    continue;
  }
  ...
}
```

This also makes the predicate identical to the one `BeginTransaction()` uses to build `checkpointed_sampler_states_`, which is the invariant the two loops depend on.

src/dflash2_drafter.cpp
```diff
@@ -390,6 +390,9 @@ bool Dflash2Drafter::Admit(const Feed& feed) {
   if (requests_.find(feed.request) != requests_.end()) {
     return true;
   }
+  if (!Dflash2CanJoin(feed.draft_eligible, feed.first_position)) {
```
tianleiwu, line 493: Dead code just below: this new `Dflash2CanJoin()` check already requires `first_position == 0`, so the `if (feed.first_position != 0) { return false; }` immediately below it can never fire.

Suggest deleting it and folding the explanatory comment above it into `Dflash2CanJoin`'s declaration in the header, which is now the single place the join rule lives.

src/engine/request.cpp
```diff
@@ -13,6 +11,52 @@
...
+  }
+  std::seed_seq seed_sequence{static_cast<uint32_t>(seed & 0xffffffffu),
+                              static_cast<uint32_t>(seed >> 32)};
+  return std::mt19937{seed_sequence};
+}
+
+// The model-configured seed initializes the Request's durable basis exactly once. A negative
+// configured seed means "pick one", and the drawn value becomes the basis so later turns that omit
+// a seed continue that same stream instead of redrawing.
+uint64_t InitialSeedBasis(int configured_seed) {
```
tianleiwu, line 32: `InitialSeedBasis()` default-constructs a `std::random_device` on every Request whose model leaves `search.random_seed` unset — which is the default — including every MTP auxiliary shadow, which never samples. On libstdc++ that can open `/dev/urandom` per call. Request creation runs on the Engine owner thread and is per-request in a server loop, so this is a per-admission syscall for a value most Requests never consume differently.

A function-local `static thread_local std::random_device` preserves the behaviour at no per-call cost.

Minor, same function: the two `random_device()` calls in `(static_cast<uint64_t>(random_device()) << 32) | random_device()` are unsequenced relative to each other (`|` does not sequence its operands). Harmless for a seed, but naming the two halves in separate statements removes the question for the next reader.

src/engine/request.cpp
```diff
@@ -13,6 +11,52 @@
...
+  // A Request is one sequence and the Engine batches Requests, not rows within a Request. These are
+  // internal invariants every Request holds, including the scheduler-private MTP shadows built from
+  // a second model: several places here read row 0 only, and beam search never overrides the
+  // deferred-completion contract. Engine::CreateRequest() separately rejects a public Request whose
+  // model configures num_beams != 1, so forcing it here never silently changes what a caller asked
+  // for.
+  params->search.batch_size = 1;
+  params->search.num_beams = 1;
+  // The Request's session limit is the only length limit the Search may complete on.
+  params->search.max_length = static_cast<int>(max_session_tokens);
```
tianleiwu, line 57: `CreateRequestParams()` neutralizes `batch_size`, `num_beams`, and `max_length`, but leaves `do_sample`, `temperature`, `top_p`, `top_k`, `repetition_penalty`, `no_repeat_ngram_size`, `min_length`, and `random_seed` at their model-configured values.

Nothing reads them any more — I checked, `Search` only reads `batch_size`, `num_beams`, `max_length`, and `num_return_sequences` — but they remain reachable through `params_`, and a future change that reintroduces `params_->search.top_k` would silently bypass `TurnPolicy()` with no test failing. Since this function is already the one place that states the Engine's derived-parameter invariants, zeroing/neutralizing the now-dead policy fields here would make "generation policy lives only in `EffectiveTurnPolicy`" structural rather than conventional.

src/ort_genai_c.cpp
```diff
@@ -1891,30 +1885,111 @@ OgaResult* OgaTurnOptionsSetMaxGeneratedTokens(
...
-#define OGA_UNIMPLEMENTED_TURN_SETTER(name, signature)       \
-  OgaResult* name signature {                                \
+OgaResult* OgaTurnOptionsSetMinGeneratedTokens(
+    OgaTurnOptions* options, uint64_t min_generated_tokens) {
+  OGA_TRY
+  if (!options) {
+    throw std::runtime_error("options must not be null.");
+  }
+  options->ValidateOwnerThread();
+  // No uint64-to-size_t range check: size_t is 64 bits on every platform this library supports, so
```
tianleiwu: This deliberately omits the `uint64_t` → `size_t` range check that `OgaRequestOptionsSetMaxSessionTokens` and `OgaTurnOptionsSetMaxGeneratedTokens` both perform in this same file.

The reasoning is correct on a 64-bit target, but three sibling setters now disagree on whether the guard is required, which is the kind of asymmetry that gets "fixed" in the wrong direction later. Either apply the same guard for symmetry, or move this justification to a comment covering all three so the omission reads as a group decision rather than an oversight.

benchmark/engine/scenarios/decode_baseline.cpp
```diff
@@ -87,35 +87,40 @@ ScenarioExecutionOutput DecodeBaselineScenario::Execute(const ScenarioConfig& co
...
       request_state.tokens.assign(
           prompt_tokens->SequenceData(0), prompt_tokens->SequenceData(0) + prompt_token_count);
 
-      requests.emplace_back(engineResources.engine->CreateRequest(*params.back()));
+      requests.emplace_back(
+          engineResources.engine->CreateRequest(request_options.back().get()));
       request_states_by_request.emplace(
           requests.back().get(), &request_state);
+      turn_options.push_back(requests.back()->CreateTurnOptions());
+      turn_options.back()->SetSeed(kRandomSeed);
```
tianleiwu: This (and the equivalent change in `capacity_pressure.cpp`, `long_prefill.cpp`, and `mixed_workload.cpp`) replaces `SetSearchOption("random_seed", ...)`, which was accepted on both scheduling paths, with `OgaTurnOptions::SetSeed()`, which `Engine::BeginTurn` rejects unless the model configures `engine.dynamic_batching`.

None of these scenarios sets `do_sample`, so the seed does not change what they measure — it only adds a new way for the benchmark suite to abort against a model without dynamic batching. Consider dropping the seed from these scenarios, or setting it only when the Engine is dynamically batched.

tianleiwu (commented):
### Review summary

Verdict: **Comment**. This is high-quality work and I did not find a blocking correctness defect. Moving generation policy out of `GeneratorParams` and into a per-turn `EffectiveTurnPolicy` is the right ownership split, and the transactional discipline around the reseed (applied strictly between the checkpoint and the first RNG consumer, promoted only at commit, discarded at every terminal boundary) is the strongest part of the change. Unifying four open-coded greedy tests behind `EffectiveTurnPolicy::IsGreedy()` removes a real class of divergence between the ordinary, batched, draft-verification, and drafting-eligibility paths.

I specifically verified the pieces most likely to break, and they hold:

- `CreateSearch()` branches only on `num_beams`, so flipping `do_sample` per turn cannot select a different `Search` subclass.
- `ScheduledRequests::CommitStateForTransaction()` (which promotes the pending reseed) runs before `CommitStep()` (which calls `ReleaseTurnResources()` and discards it) in `Engine::RunDynamic()`, so the ordering claim in the `ReleaseTurnResources()` comment is accurate.
- `SupportsNoRepeatNgram(DeviceType::CPU)` matches reality — only `Search_Cpu` overrides `ApplyNoRepeatNgram`, and `ApplyMinLength`/`ApplyRepetitionPenalty` are implemented on both CPU and CUDA, so no equivalent capability gate is needed for the new `min_generated_tokens`/`repetition_penalty` options.
- `ValidateTurnPolicy` covers every combination the removed validation used to reject, so dropping `std::optional` from `ResolveSampleArgs` is safe.
- `TurnEosFloor()` is session-absolute and bounded by admission against `max_session_tokens`, so the `int` narrowing is provably safe.
- The new `assert(!guidance_logits_processor_)` in `SaveStateForNewTurnTransaction()` holds: every terminal path routes through `ReleaseTurnResources()`.
- `BatchedSampler` has exactly one implementation, and the new pure virtual plus the `int` → `uint64_t` seed widening are covered by the `kDeviceInterfaceVersion` 4 → 5 bump.
- No C#/Java/Objective-C binding references `OgaEngineCreateRequest`, so the signature change is contained to C, C++, and Python.

Seven inline comments below (all suggestions/nitpicks). Two more observations that have no single line to anchor to:

**DFlash 2 ring join is now one-shot against a per-turn property.** `feed.draft_eligible` comes from the *current* turn's policy, but a request can only join the drafter at `first_position == 0`. A request whose first turn samples therefore never claims a ring, even if every later turn is greedy. Eligibility used to be a fixed per-request property, so "decided at position zero" was equivalent; with per-turn `do_sample` it no longer is. The behaviour is documented in `docs/paged_attention_engine.md`, so this is a design note rather than a defect — just confirming it is intentional rather than incidental.

**CI.** Two `Integration Tests` jobs (Linux x64 CUDA, `qwen3-0.6b` and `ministral-3-3b-Instruct-2512`) are red on this head. The annotations report bash exit code 127, which reads as infrastructure rather than a test assertion, but `test/python/integration/test_integration_engine.py` is in the diff, so it is worth confirming green (or explaining) before merge.

src/engine/engine.cpp
baijumeswani: Addressed. `BeginTurn` now checks whether the scheduler can checkpoint sampler state before admitting a seeded turn. Host-only sampling remains supported, and a device sampler without transaction support is rejected before the Request is mutated. The requirement is also documented in the C header and Engine docs.

src/engine/request.cpp
baijumeswani: Addressed. `std::random_device` is now `static thread_local`, and the high and low halves are drawn in separate statements before they are combined.

benchmark/engine/scenarios/decode_baseline.cpp
baijumeswani: Addressed across all five Engine benchmark scenarios, including continuation. They now apply a per-turn seed only when the model config uses dynamic batching. Dynamic benchmarks keep deterministic behavior, while static models no longer fail admission because of an otherwise unused seed.

src/engine/request.cpp
baijumeswani: Addressed the ownership concern without rewriting the stored model values. `CreateRequestParams` now explicitly states that generation policy must come from `Request::TurnPolicy()`, while `search.chunk_size` remains live model-owned scheduler policy. Zeroing the fields would leave different misleading values and would not prevent a future accidental read.

src/dflash2_drafter.cpp
baijumeswani: Addressed. I removed the redundant `first_position != 0` check. The no-backfill reason now lives with the `Dflash2CanJoin` declaration, which is the single source of the join rule.

src/ort_genai_c.cpp
baijumeswani: Addressed. `OgaTurnOptionsSetMinGeneratedTokens` now has the same `uint64_t` to `size_t` overflow check and error shape as the max-session and max-generated-token setters. This also handles supported 32-bit targets correctly.

src/engine/scheduled_requests.cpp
baijumeswani: Addressed. The loop now checks `HasPendingTurnSeed()` first, before looking up sampler state or scanning checkpointed states. This matches the predicate used when creating the sampler checkpoints.

baijumeswani:
Thanks for the detailed review. The DFlash2 join behavior is intentional: the drafter cannot backfill earlier auxiliary state, so eligibility is evaluated on the position-zero step. If that step runs under sampling, the Request does not join later; Requests that already joined continue ingesting sampled turns so later greedy turns can resume drafting. I made this lifetime behavior explicit in the Engine documentation.

src/engine/scheduled_requests.cpp
```diff
@@ -17,22 +17,18 @@ namespace Generators {
...
-// Collapses Request::GenerateNextTokens' dispatch into the single (k, p, temperature) triple that
+// Collapses Request::SelectNextToken's dispatch into the single (k, p, temperature) triple that
 // each branch ends up handing to the sampler on CUDA, where SelectTop() is SampleTopKTopP(1, 0, 1),
 // SampleTopK(k, t) is SampleTopKTopP(k, 1, t) and SampleTopP(p, t) is SampleTopKTopP(-1, p, t).
-// Returns nothing for options the per-request path rejects, so that it keeps raising the error.
-std::optional<BatchedSamplingParams> ResolveSampleArgs(const Config::Search& search) {
-  if (!search.do_sample || search.top_k == 1 || search.temperature == 0)
+BatchedSamplingParams ResolveSampleArgs(const EffectiveTurnPolicy& policy) {
+  if (policy.IsGreedy())
     return BatchedSamplingParams{1, 0.0f, 1.0f};
```
tianleiwu, line 25: This maps every greedy turn to the batched CUDA sampler with `k=1`, but `FusedSamplingKernel` unconditionally calls `curand_uniform` even when `k == 1`. That advances this Request's persistent device RNG for a deterministic token, whereas the CPU `SelectTop()` path consumes no RNG. As a result, adding or lengthening a greedy turn changes a later unseeded sampled continuation only on CUDA. Please route greedy rows through device argmax, or special-case `k == 1` without drawing from `curandState`, and add an Engine CUDA test for a greedy-to-sampled continuation.

src/engine/turn_policy.cpp
```diff
@@ -0,0 +1,195 @@
...
+    throw std::runtime_error(
+        "temperature (" + std::to_string(policy.temperature) +
+        ") must be finite and 0 or greater.");
+  }
+  if (!std::isfinite(policy.top_p) || policy.top_p < 0.0f || policy.top_p > 1.0f) {
+    throw std::runtime_error(
+        "top_p (" + std::to_string(policy.top_p) +
+        ") must be finite and between 0.0 and 1.0.");
+  }
+  if (policy.top_k < 0) {
```
tianleiwu, line 95: The Engine validates only the lower bound here. The classic Generator also rejects `top_k > vocab_size`, but this path accepts it; CPU then clamps it and CUDA converts it to full-vocabulary sampling. That makes the same public generation policy behave differently depending on whether it is used through Generator or Engine. Please pass the model vocabulary size into turn-policy validation and preserve the existing upper-bound contract, with a boundary test for the Engine path.

test/cpp/c_api_tests.cpp
```diff
@@ -1464,7 +1640,7 @@ struct Phi2Test {
       tokenizer_->Encode(input_strings[i], *input_sequence);
       auto input_tokens = std::span<const int32_t>{
           input_sequence->SequenceData(0), input_sequence->SequenceCount(0)};
-      requests.push_back({engine->CreateRequest(*params_),
+      requests.push_back({engine->CreateRequest(),
```
tianleiwu: `RunEngine()` no longer consumes `params_`, but `TopKCAPI`, `TopPCAPI`, and `TopKTopPCAPI` still configure only that object before calling this helper. These Engine branches therefore all run model-default generation, so regressions in the new per-turn sampling setters can pass unnoticed (and their configured `max_length` is ignored too). Please translate those configured values into `RequestOptions`/`TurnOptions`, or give each Engine case an explicit turn policy that preserves what the parameterized test claims to exercise.

tianleiwu (changes requested):
### Review summary

**Verdict: Request changes.** The per-turn policy ownership split is coherent, and the transactional handling of request state, guidance, and reseeding is carefully designed. I found two correctness issues to address before merge: deterministic CUDA turns currently advance persistent RNG state, and Engine validation accepts `top_k` values that the classic Generator rejects. The migrated parameterized C API cases also need to apply their configured per-turn policies so these paths remain covered.

All 65 reported checks are complete with no failures or pending jobs. I also rechecked the seven threads from the previous review round; they are resolved on this head.

test/cpp/c_api_tests.cpp
baijumeswani: Addressed in `503b29a6a6`. `Phi2Test::RunEngine` now creates Requests with `RequestOptions.max_session_tokens = 40`, preserving the former session-absolute `max_length`, and accepts a TurnOptions configurator. The Top-K, Top-P, and combined Top-K/Top-P Engine cases now explicitly set `do_sample`, temperature, and their intended distribution bounds through the public per-turn setters, so those parameterized branches exercise the migrated Engine API rather than model defaults.

src/engine/scheduled_requests.cpp
baijumeswani: Addressed in `503b29a6a6`. `FusedSamplingKernel` now returns the sorted top token immediately for `k == 1`, before any `curand_uniform` call, so greedy CUDA rows no longer advance the persistent device RNG stream. I added both an end-to-end CUDA Engine regression comparing a greedy-to-unseeded-sampled continuation with an equivalent fresh-context request and a lower-level same-seed sampler-state regression. Both tests fail without the early return and pass with it. The Engine RNG contract is also documented explicitly.

src/engine/turn_policy.cpp
baijumeswani: Addressed in `503b29a6a6`. `Engine::BeginTurn` now passes `model.vocab_size` into turn-policy validation, and sampled policies reject `top_k > vocab_size` before any Request mutation, matching the classic Generator contract. The validation remains sampled-only, like `Generator::InitializeSamplingMethod`; if the oversized value came from the model default, the error names `search.top_k` and tells the caller to override it. Tests cover explicit rejection, unchanged Request state, `top_k == vocab_size`, and overriding an invalid model default at the boundary.

src/engine/request.cpp
```diff
@@ -300,8 +345,36 @@ void Request::CompleteFailedTurnFromEngine(const Engine& engine) noexcept {
   status_ = RequestStatus::TurnComplete;
   finish_reason_ = GenerationFinishReason::Failed;
   matched_stop_string_index_ = -1;
+  ReleaseTurnResources();
+}
+
+// Everything scoped to one turn goes away together at every terminal boundary, so the next turn
+// starts from the documented defaults: no stop strings, no guidance, and no draft proposal unless
+// it asks for them.
+void Request::ReleaseTurnResources() noexcept {
```
tianleiwu, line 354: `ReleaseTurnResources()` releases the guidance cursor but not its transaction checkpoint, while `MarkClosedFromEngine()` and `CompleteClose()` both clear `guidance_transaction_checkpoint_` and `guidance_logits_processor_` together.

Unreachable today: the comment above is right that cancellation and unserviceable failure happen between steps, and `CommitStep()` only runs after the step committed, so no live checkpoint can outlive this call. But the invariant that used to be established actively is now only asserted (see the `SaveStateForNewTurnTransaction()` comment), so if a terminal boundary is ever reached with a checkpoint still set, `RestoreStateForTransaction()` would move the dead turn's grammar cursor back into `guidance_logits_processor_` and the next turn would inherit it in a release build.

Adding `guidance_transaction_checkpoint_.reset();` here makes the "everything scoped to one turn goes away together" contract cover the checkpoint too, and matches the two close paths.

src/engine/request.cpp
```diff
@@ -793,19 +826,18 @@ void Request::SaveStateForTransaction() {
...
-                               ? guidance_logits_processor_->CloneForNewTurn()
-                               : nullptr;
   search_->SaveStateForTransaction();
-  guidance_transaction_checkpoint_ =
-      std::exchange(guidance_logits_processor_, std::move(new_turn_guidance));
+  // Every terminal path releases the previous turn's guidance cursor and stop controller, and this
+  // turn's replacements are installed only by CommitTurnAdmission(), so a rolled back admission has
+  // neither to restore. The same paths discard any reseed the previous turn left pending, so this
+  // admission cannot promote a stale seed when it commits.
+  assert(!guidance_logits_processor_);
```
tianleiwu: These three invariants used to be established, not asserted: the old code did `guidance_transaction_checkpoint_ = std::exchange(guidance_logits_processor_, CloneForNewTurn())`, so a leftover cursor was actively replaced. Now a violation is a no-op under `NDEBUG`, and the failure mode is silent rather than loud: the new turn decodes under the previous turn's grammar cursor, or `CommitStateForTransaction()` promotes a stale seed to `current_seed_basis_` so every later unseeded turn continues the wrong stream.

`ApplyPendingSeedForTransaction()` in this same file already uses the stronger shape for exactly this class of contract violation:

```cpp
assert(false && "...");
throw std::logic_error("...");
```

Suggest the same here (assert plus throw) so a release build fails the admission rather than silently inheriting turn-scoped state. These run once per `BeginTurn`, so the cost is irrelevant.

src/ep/cuda/cuda_sampling.cu
```diff
@@ -95,6 +95,13 @@ __global__ void FusedSamplingKernel(int32_t* next_token_out, const float* scores
   const float* batch_scores = scores + batch_idx * stride;
   const int* batch_indices = indices + batch_idx * stride;
 
+  if (k == 1) {
```
tianleiwu, line 101: This fast path deserves a comment. It carries two facts a reader cannot see locally:

1. It is only correct because `indices` is sorted descending by score, which the kernel does not state until ~40 lines below ("For sorted input, the max score is always the first element").
2. It is the fix for the greedy-row RNG advance: a `k == 1` row no longer draws from `curandState`, so a greedy turn no longer shifts a later sampled continuation's stream on CUDA. That is a deliberate behavior change, not just a shortcut past dead math.

Something like:

```cpp
// Greedy rows resolve to (k=1, p=0, t=1). The candidates are sorted descending, so the top
// logit is indices[0]; returning here also keeps a greedy row from consuming a curand draw,
// so it cannot shift a later sampled turn's stream.
```

src/engine/engine.cpp
```diff
@@ -1011,35 +994,59 @@ void Engine::ValidateOwnerThread() const {
...
-  }
-  if (params.search.max_length <= 0) {
+        "The model's search.max_length must be greater than zero; actual value is " +
+        std::to_string(model_search.max_length) + ".");
+  }
+  // The Engine expresses a per-turn floor through TurnOptions::min_generated_tokens. A model that
+  // still carries the legacy session-absolute floor would silently mean something else here, so it
+  // is rejected rather than reinterpreted -- but the caller can clear it without editing the model
+  // directory, through the config overlay that OgaCreateConfig/OgaConfigOverlay already support.
+  if (model_search.min_length != 0) {
```
tianleiwu, line 1046: Model-fixed validation is now split across two places. This PR moved the `search.chunk_size` check up to `Engine::CreateDependencies()` (line 240) precisely because it is a property of the model rather than of a request, but `search.max_length <= 0`, `search.min_length != 0`, and `search.num_beams != 1` are all equally model-fixed and are still re-evaluated on every `CreateRequest()`.

Moving these three next to the chunk_size check would fail fast at `OgaCreateEngine` instead of on the first request, and would leave `CreateRequest()` validating only the one thing that actually varies per request (`max_session_tokens` against the ceiling).

One caveat if you take this: `Engine` can be constructed with injected dependencies that bypass `CreateDependencies()` (which is why `StaticBatchScheduler::AddRequest` keeps its own chunk_size guard). If that path matters for these three as well, keeping them here is defensible — but then the split is worth a one-line comment saying why chunk_size moved and these did not.

src/engine/turn_policy.cpp
```diff
@@ -0,0 +1,210 @@
...
+  }
+  message +=
+      ". Set that field explicitly on the turn -- top_k above 1, or a nonzero temperature -- to "
+      "request sampling, or drop do_sample to accept top-logit selection.";
+  throw std::runtime_error(std::move(message));
+}
+
+}  // namespace
+
+bool SupportsNoRepeatNgram(DeviceType scoring_device_type) noexcept {
```
tianleiwu, line 52: Nitpick on the comment, not the code. This is correct today, but only because of a mapping in `src/models/model.cpp` that is not named here:

```cpp
```

So `p_device_scoring_->GetType()` is only ever `CPU`, `CUDA`, or `NvTensorRtRtx`. Every other EP (`DML`, `WEBGPU`, `QnnHtp`, `QnnGpu`, `OpenVINO`, `RyzenAI`, `AMDGPU`) reaches here as `CPU` and builds `GreedySearch_Cpu`, which does implement `ApplyNoRepeatNgram` — the `return type == CPU` answer is right for them for a reason the comment does not give.

The risk is a future EP that starts scoring on-device: it would flip to `false` here and start rejecting a turn its `GreedySearch_Cpu` could still serve. Worth naming the `model.cpp` invariant in this comment so the coupling is visible from the function that depends on it.

tianleiwu (commented):
Re-reviewed at `503b29a6`. Both threads I left open in the previous round are addressed on this head, and I have resolved them:

- The greedy-row `curandState` advance is gone: `FusedSamplingKernel` now short-circuits `k == 1` before `curand_uniform`, and `TurnPolicyTest.CudaGreedyTurnDoesNotAdvanceLaterSampledContinuationRng` pins the greedy-to-sampled continuation on CUDA.
- `ValidateTurnPolicy()` now takes `vocab_size` and rejects `top_k > vocab_size`, with a distinct message when the value came from the model default, plus boundary tests for both cases.

**Verdict: no blocking defects.** The remaining comments are hardening and consistency suggestions.

### What works well

- Turn admission really is transactional. Policy resolution, validation, stop-controller construction and grammar compilation all happen before any Request or Engine mutation, and `ReleaseTurnResources()` gives every terminal boundary one release path. The claim that a bad grammar or an impossible limit leaves the Request reusable is structural rather than conventional.
- The reseed design is careful in the places that are easy to get wrong: pending survives rollback, `applied` does not, and promotion to the durable basis happens only in `CommitStateForTransaction()`. `BeginTransaction()` checkpointing every device state a pending reseed targets, including states created by a sampling plan that was later abandoned, closes the one hole where a reseed would have been an unrecoverable mutation inside a transaction.
- `EffectiveTurnPolicy` deliberately having no member initializers, so `ResolveTurnPolicy()` is the only source of policy, is the right call and is enforced by `IsGreedy()` being the single classification every execution path shares.
- The API break is made loud everywhere it can be: the C++ overload no longer compiles with a `GeneratorParams`, `pybind11::kw_only()` rejects the old positional Python call, and `kDeviceInterfaceVersion` is bumped for the `BatchedSampler` vtable change (`CreateState` signature plus the new pure-virtual `ReseedState`).
- Docs, all three Python engine examples, and all five benchmark scenarios were migrated together, and the benchmark seeds are now applied only under dynamic batching.

### Checks I ran that came back clean

- No stale references remain to `MaxTotalTokens`, `SearchOptions()`, `ValidateEngineCompatibility`, `CloneForNewTurn`, `Dflash2CanDraft`, or `max_total_tokens`.
- `CudaBatchedSampler` is the only in-tree `BatchedSampler`, so the new pure virtual breaks nothing internally.
- MTP shadow requests select tokens through `GreedyTokens`/device argmax and never through `SelectNextToken()`, so deriving their params from `mtp_model_` instead of copying the target's `search` does not change what they decode.
- Member-initializer order in the new `Request` constructor matches declaration order (`turn_policy_` -> `max_session_tokens_` -> `params_` -> `current_seed_basis_` -> `rng_` -> `search_`).
- `guidance_transaction_checkpoint_` is still written by `SaveStateForTransaction()` and `SaveStateForExternalSamplingTransaction()`, so the restore branch is not dead.

### Compatibility note (not a change request)

A model directory whose `genai_config.json` carries a nonzero `search.min_length`, or `search.num_beams != 1`, now fails `OgaEngineCreateRequest` rather than being reinterpreted. That is the right call and the error names the config-overlay escape hatch, but it is a behavior change for existing model directories that previously ran on the Engine. Worth calling out in release notes.

src/engine/request.cpp
```diff
@@ -997,24 +1043,22 @@ void Request::ApplyLogitsProcessors(DeviceSpan<float> logits,
   if (guidance_logits_processor_ && !guidance_applied) {
     guidance_logits_processor_->ProcessLogits(logits);
   }
-  auto& search_params = search_->params_->search;
-  search_->ApplyMinLength(search_params.min_length);
-  search_->ApplyRepetitionPenalty(search_params.repetition_penalty);
-  search_->ApplyNoRepeatNgram(search_params.no_repeat_ngram_size);
+  search_->ApplyMinLength(TurnEosFloor());
```
bmehta001: Is this risky if it terminates guidance too early and EOS is omitted?

tianleiwu (commented):
Re-reviewed at `b9dc4240` (`Harden Engine turn resource invariants`). All five items from my previous round are addressed; I have resolved those threads.

| Item | Resolution |
|---|---|
| `ReleaseTurnResources()` left `guidance_transaction_checkpoint_` set | `guidance_transaction_checkpoint_.reset()` added, so the turn-scoped release now covers the pair the two close paths already cleared together. |
| `SaveStateForNewTurnTransaction()` invariants were `assert`-only | Now assert + `throw std::logic_error`. |
| `k == 1` CUDA fast path was uncommented | Comment names both the sorted-candidates precondition and the RNG-stream reason. |
| Model-fixed validation split between `CreateDependencies()` and `CreateRequest()` | Comment states the reason (injected-dependency Engines bypass `CreateDependencies()`), which is the option I flagged as the caveat. |
| `SupportsNoRepeatNgram()` coupling was implicit | Comment now names the `Model` CPU-scoring mapping it depends on. |

One thing worth calling out as better than what I asked for: the guard in `SaveStateForNewTurnTransaction()` was moved *above* `search_->SaveStateForTransaction()`. That matters — throwing after the search checkpoint would have left a checkpoint that `RollbackTurnAdmission()` never restores, because `admission.transaction_started` is still `false` at that point. As written, a violation throws with nothing mutated: `PrepareTurnAdmission()` has only computed a local `device_tokens` and filled in `admission` bookkeeping fields, and `Engine::BeginTurn()`'s catch calls `RollbackTurnAdmission()`, which correctly no-ops. That keeps the new hard failure consistent with the PR's own "a rejected admission leaves the Request reusable" contract.

I also confirmed the new `throw` is not reachable from a `noexcept` context (`PrepareTurnAdmission()` at `request.cpp:222` is the only caller, inside `BeginTurn`'s try block), and that the extra `guidance_transaction_checkpoint_.reset()` in `ReleaseTurnResources()` is redundant-but-harmless on the commit path, since `ScheduledRequests::CommitStateForTransaction()` (engine.cpp:1912) already runs before `CommitStep()` (engine.cpp:1920).

**Verdict: LGTM.** No remaining findings from my side. @bmehta001's question on `request.cpp:1046` is still open and is not covered by this pass.

tianleiwu (dismissed):

src/engine/request.cpp
baijumeswani: This concern was valid. A terminal guidance state can permit only EOS; applying the per-turn minimum afterward masked that final legal token and left the row with no valid choice. The fix in 5fae2bcb3a now applies the minimum while guidance has any non-EOS continuation, but lets guidance terminate when its ready mask permits only configured EOS tokens. The check uses the model’s full EOS set, ignores mask padding, and validates mask dimensions. Regression coverage includes both sides of the rule: a fixed `"!"` grammar terminates below the minimum, while extendable `[0-9]*` guidance continues until the minimum is reached. Direct mask tests also cover multiple EOS IDs, padding bits, empty masks, and malformed dimensions. The full guidance-enabled Engine suites pass on CPU (602 tests) and CUDA (611 tests), and the revised change received independent sign-off.

tianleiwu (commented):
Re-reviewed at `f070e286`. Two new commits since my last pass, plus a rebase onto the DSpark work that landed on main.

**Rebase integrity checked.** `git range-diff af166b19..b9dc4240 649248d5..e8417de9` reports commits 2 and 3 as identical (`=`); commit 1 differs only in `docs/paged_attention_engine.md` and `dflash2_drafter.h` comments, reconciling upstream's DFlash 2 -> block-drafter/DSpark terminology. No source semantics moved during the replay.

### `22be2c09` Fix terminal guidance under minimum length

This is the right fix for the empty-legal-token-set problem, and the details are careful:

- The new `eos_floor > CurrentSequenceLength()` outer guard exactly reproduces the early-return in **both** `Search_Cpu::ApplyMinLength` and `Search_Cuda::ApplyMinLength`, and `Request::CurrentSequenceLength()` is `search_->GetSequenceLength()`, so it is behavior-preserving and only avoids doing the `AllowsOnlyTokens()` work when the floor is already passed.
- `MaskAllowsOnlyTokens()` masks the tail padding bits (`vocab_size % 32`) rather than trusting the allocator to have zeroed them, and requires at least one supplied token to be permitted — so an all-zero mask returns `false` and the minimum still applies rather than silently unmasking EOS. Both are covered by `IgnoresPaddingBitsAndRequiresAnAllowedToken`.
- It short-circuits on the first word carrying a non-EOS allowed bit, so a permissive grammar (the common case) exits after roughly one word instead of scanning the whole mask every step.
- It keys off `config.model.eos_token_id` rather than the processor's single `eos_token_`, so a model with several EOS ids is handled — `RecognizesAnyConfiguredEosAndRejectsOtherTokens` pins that.
- The fix sits in `ApplyLogitsProcessors()`, which both the per-request path and the batched-sampler path reach (`PrepareGenerationForTransaction()` at `scheduled_requests.cpp:761`), so guided turns under a minimum behave the same on either path.
- Both `ConstrainedLogitsProcessor` test doubles were updated for the new pure virtual; those are the only two subclasses besides `GuidanceLogitsProcessor`.

One divergence flagged inline on `request.cpp`.

### `f070e286` Integrate per-turn drafting with DSpark

Test-only plus one doc sentence, and the conversion to designated initializers is the right call rather than a style change: `draft_eligible` was inserted *before* `wants_drafts` in `Dflash2Drafter::Feed`, so the existing positional `{request_a, 0, 9, 0, 11, true}` would have silently become `draft_eligible = true, wants_drafts = false` and quietly stopped exercising drafting in the DSpark lifecycle test. Designated initializers also make the next field insertion a compile-time concern instead of a silent one. Field order matches declaration order, as C++20 requires.

`TrackedDsparkIngestsSampledTurnsAndResumesDrafting` covers exactly the transition the per-turn eligibility change introduces — tracked request ingests during a sampled turn, then resumes drafting — and the `EXPECT_EQ(drafter.AdmissionMisses(), 0u)` on the ineligible untracked feed is a good distinction: an eligibility rejection is not a capacity miss, which matters because that counter drives the `max_batch_size` sizing guidance in the docs.

### Status

@bmehta001's question on `request.cpp:1046` is addressed by `22be2c09`. No unresolved threads remain from my earlier rounds.

**Verdict: LGTM**, with one non-blocking follow-up noted inline.

src/engine/request.cpp
```diff
@@ -997,24 +1043,31 @@ void Request::ApplyLogitsProcessors(DeviceSpan<float> logits,
...
   }
-  auto& search_params = search_->params_->search;
-  search_->ApplyMinLength(search_params.min_length);
-  search_->ApplyRepetitionPenalty(search_params.repetition_penalty);
-  search_->ApplyNoRepeatNgram(search_params.no_repeat_ngram_size);
+  const int eos_floor = TurnEosFloor();
+  // An extendable accepting grammar still honors the minimum by suppressing its optional EOS.
+  // Once EOS is the grammar's only legal token, guidance termination takes precedence over the
+  // floor; masking it as well would leave no valid token to select.
+  if (eos_floor > CurrentSequenceLength() &&
```
tianleiwu, line 1050: This fix is Engine-only, and the classic Generator composes the same two things in the same order without the guard:

```cpp
// src/decoding/standard_decoding_strategy.cpp
if (g.guidance_logits_processor_) {
  auto logits = g.GetLogits();
  g.guidance_logits_processor_->ProcessLogits(logits);
}
...
g.search_->ApplyMinLength(search.min_length);
```

So the exact scenario this commit fixes — a grammar that reaches a terminal state where only EOS is legal, before `min_length` — still masks itself into an empty legal-token set on the Generator path. Same grammar, same minimum, two different outcomes depending on which API the caller used.

I don't think it blocks this PR: it is a pre-existing Generator bug, and the PR is explicit that the classic Generator surface is unchanged. But it is worth noting that the divergence is now user-visible rather than internal, and that `AllowsOnlyTokens()` was deliberately added to the shared `ConstrainedLogitsProcessor` base rather than to `GuidanceLogitsProcessor` alone, so the classic path can reuse it as-is:

```cpp
if (search.min_length > g.search_->GetSequenceLength() &&
    !(g.guidance_logits_processor_ &&
      g.guidance_logits_processor_->AllowsOnlyTokens(
          0, g.search_->params_->config.model.eos_token_id))) {
  g.search_->ApplyMinLength(search.min_length);
}
```

(One difference to keep in mind if you do follow up: the Generator's `min_length` is session-absolute and applies across the whole sequence, whereas `TurnEosFloor()` is derived per turn — so the guard condition is the same shape but the floor is not.)

A follow-up issue is fine if you'd rather keep this PR scoped.

baijumeswani: Yes, I will address this in the Generator as a follow up.

tianleiwu (approved):

PR #2535, opened 2026-09-04
[WebGPU] Support paged attention models

tianleiwu:
## Summary

Enable model-builder and runtime support for running paged-attention models with the WebGPU execution provider. WebGPU cannot query available device memory and its PagedAttention kernel does not consume the CPU-only `attention_metadata` input, so these models need an explicit cache capacity and a provider-specific input contract.

## Key changes

- Add and validate the `num_blocks` model-builder option, emitting it instead of `gpu_utilization_factor` for fixed-capacity paged KV caches.
- Omit `attention_metadata` from WebGPU paged models while preserving the remaining paged inputs.
- Fail with an actionable message if runtime memory-based cache sizing is attempted on WebGPU.
- Normalize full WebGPU provider names and forward `adapterIndex` to the initialization session.
- Document the WebGPU export command and add focused C++ and Python regression coverage.

## Testing

- `python -m pytest -q test/python/builder/test_paged_windowed_kv_cache.py test/python/builder/test_precision.py` (`108 passed`)
- `clang-format --dry-run --Werror src/config.cpp src/ep/webgpu/interface.cpp test/cpp/model_tests.cpp`

C++ build/tests were not run locally because CMake Tools has no configured build targets in this workspace and could not configure the project.

### DIFF
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -2188,11 +2188,11 @@ void SetProviderOption(Config& config, std::string_view provider_name, std::stri
   }
 
   // JSON-escape all caller-supplied string fragments before concatenating them into the
-  // JSON document. Without escaping, quote/backslash characters in provider_name,
+  // JSON document. Without escaping, quote/backslash characters in the provider name,
   // option_name, or option_value would let a caller inject arbitrary JSON structure
   // (sibling keys, new provider entries, etc.) into the parsed configuration.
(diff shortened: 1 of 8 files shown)


prathikr:
Continued in https://github.com/microsoft/onnxruntime-genai/pull/2570

PR #2537, opened 2026-09-05
Fix tool calling: preserve tool schemas and stop at end of turn

tianleiwu:
## Description

Tool calling could produce arguments that violate the caller's schema and continue generating beyond the assistant turn. These are configuration failures rather than model-quality failures: lossy tool-schema normalization changed the prompt seen by Qwen-style templates, while an incomplete `eos_token_id` allowed generation to continue into fabricated tool-result turns.

This PR consumes the upstream schema-preservation fix from `onnxruntime-extensions`, makes exported GenAI configurations include the tokenizer's end-of-turn token, and adds a deterministic two-turn benchmark that detects schema, stopping, and tool-result grounding regressions.

### Tool schemas were rewritten before reaching the model

`onnxruntime-extensions` normalized OpenAI tool definitions into the flat Phi-4/Minja shape. For templates that serialize tools directly with `tool | tojson`, that conversion dropped constraints such as `required`, `enum`, nested properties, and array `items`, so the model saw a different schema from the caller.

The fix was upstreamed in `microsoft/onnxruntime-extensions#1117`. This PR pins extensions commit `543e968c4fedd2a71a6cca2537d873f77ed60f78`, so no local patch is needed.

### The end-of-turn token was not always a stop ID

Chat models end an assistant turn with the tokenizer's `eos_token` (for example, `<|im_end|>` for Qwen), but `config.json` can list only `<|endoftext|>`. When a checkpoint has no usable `generation_config.json`, exported models retain the wrong stop set and may continue generating subsequent assistant/tool turns themselves.

### Runtime and dependency fixes

| File | Change |
|------|--------|
| `cmake/deps.txt` | Pins `onnxruntime-extensions` to upstream commit `543e968`, which preserves raw schemas for templates that serialize OpenAI tools directly. |
| `src/python/py/models/builders/base.py` | Warns when `generation_config.json` cannot be loaded and unions the tokenizer's end-of-turn token into `eos_token_id` without duplicating existing IDs. |

### Tool-calling benchmark

| File | Change |
|------|--------|
| `benchmark/python/toolcalling/benchmark_toolcalling.py` | Adds Generator and Engine runners, optional Hugging Face prompt-fidelity comparison, configurable concurrency, two-turn evaluation, JSON reporting, and a `--fail_under` regression gate. |
| `benchmark/python/toolcalling/toolcall_parsing.py` | Parses XML, tagged JSON, wrapped JSON, and bare JSON tool-call formats and scores schema fidelity, arguments, stopping, and grounded follow-up answers. |
| `benchmark/python/toolcalling/toolcall_cases.json` | Adds 29 deterministic prompts, expected calls, canned tool results, and expected answer content. |
| `test/tool-definitions/toolcalling.json` | Stores the eight reusable OpenAI tool schemas alongside the existing tool-calling CI fixtures. The benchmark loads this file by default through `--tools`. |
| `benchmark/python/toolcalling/README.md` | Documents benchmark modes, metrics, diagnostics, shared fixtures, and usage. |

Each end-to-end case runs two turns. The first grades function selection, required parameters, enum validity, unknown parameters, expected argument values, and clean stopping. The second replays the emitted call, supplies a canned tool result, and checks that the model answers from that result without repeating the call or continuing past the turn boundary. No external tools are executed, so runs remain deterministic and offline.

## Results

Qwen3.8-27B INT4 with DFlash2, 29 cases, greedy decoding, CUDA:

| Metric | Before | After |
|--------|--------|-------|
| `correct` | 27/29 (93.1%) | **29/29 (100%)** |
| `enum_valid` | 27/29 | **29/29** |
| `clean_stop` | 3/29 (10.3%) | **29/29 (100%)** |
| `end_to_end_correct` | Not measured | **29/29 (100%)** |
| Prompt mismatches vs. Hugging Face | 29/29 | **0/29** |

Quality control for the stop-token change: MMLU-Pro shard 0 scored 85/100 versus 86/100 before, within binomial noise for 100 samples.

## Testing

- Reconfigured the existing CPU build from a clean extensions checkout and confirmed CMake fetched exactly `543e968c4fedd2a71a6cca2537d873f77ed60f78` without a patch step.
- Built the `extensions_shared` target successfully, including `shared/api/chat_template.cc`.
- Ran `python3 -m py_compile` on the benchmark runner and parser.
- Verified all eight moved schemas and all 29 cases are semantically identical to their pre-move content, then loaded and expanded every case through the new shared-fixture path.
- Ran `git diff --check` and JSON structure checks for the case and tool-definition files.

To run the benchmark against a GenAI chat model:

```bash
python benchmark/python/toolcalling/benchmark_toolcalling.py \
    -m {model folder} \
    -e cuda \
    --check_prompt_fidelity \
    --fail_under 100 \
    -o toolcalling.json
```

`--check_prompt_fidelity` requires `transformers` and `jinja2`. Existing exported models still need their `genai_config.json` corrected; the builder change applies to newly exported models.

## Checklist

- [x] Deterministic benchmark coverage added
- [x] No breaking API changes
- [x] Documentation updated
- [x] Upstream extensions fix consumed; local patch removed


benchmark/python/toolcalling/benchmark_toolcalling.py
```diff
@@ -0,0 +1,258 @@
...
+
+
+def run_with_engine(model, tokenizer, prompts, max_new_tokens, concurrency):
+    """Batch prompts through the Engine in waves of `concurrency`."""
+    engine = og.Engine(model)
+    event_buffer = engine.create_event_buffer(max(16, concurrency * 4))
+    outputs = {}
+
+    for start in range(0, len(prompts), concurrency):
+        wave = list(enumerate(prompts))[start : start + concurrency]
```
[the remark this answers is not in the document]
tianleiwu: Good catch, fixed in 5a79332b44.

The wave is now built by slicing first and enumerating with an offset, so each iteration only touches its own wave instead of rebuilding the full enumerated list:

```python
wave = enumerate(prompts[start : start + concurrency], start=start)
```

`start=start` keeps the global case indices the loop body relies on for `live` / `by_request`. Verified the index/prompt pairs are identical to the previous expression for wave sizes 1, 8, 29 and 40 (including the final partial wave and a concurrency larger than the case count).

cmake/patches/onnxruntime_extensions/apply_patch.cmake
```diff
@@ -0,0 +1,106 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
```
kunal-vaishnavi: Can we upstream this as a PR into ORT Extensions?

benchmark/python/toolcalling/toolcall_cases.json
```diff
@@ -0,0 +1,793 @@
+{
```
kunal-vaishnavi, line 1: Cases already exist in [this folder](https://github.com/microsoft/onnxruntime-genai/tree/main/test/tool-definitions). Can we add these there instead? The test cases in that folder are also used in the CI tests for tool calling.

cmake/patches/onnxruntime_extensions/apply_patch.cmake
tianleiwu: Done. The fix was upstreamed in microsoft/onnxruntime-extensions#1117 and merged as `543e968c4fedd2a71a6cca2537d873f77ed60f78`. Commit `497f94b3` pins GenAI to that revision and removes both the local `PATCH_COMMAND` and patch file. I also verified a clean CMake dependency reconfigure and built the `extensions_shared` target successfully.

benchmark/python/toolcalling/toolcall_cases.json
tianleiwu: Addressed in `d1b16f3a`. The reusable OpenAI schemas now live in `test/tool-definitions/toolcalling.json`, using the same array format as the existing CI fixtures. The benchmark file retains only prompts and expected results, and the runner loads the shared fixture by default through `--tools`. I verified that all eight schemas and all 29 cases are semantically identical after the move and that every case expands successfully.

baijumeswani (approved):

PR #2538, opened 2026-09-05
Fix DFlash target hidden-state taps

bmehta001:
## Summary
- translate SpecForge DFlash `target_layer_ids` to the following `aux_hidden_state_layers` entry
- reject unshifted tap configurations that would feed the drafter the preceding layer outputs
- update model-builder help, documentation, examples, and exporter coverage

## Why
SpecForge capture layer `L` is the output of target layer `L` (`hidden_states[L + 1]`). The model builder names auxiliary taps by the residual stream entering a layer, so the equivalent tap is `L + 1`. Matching the IDs directly supplied the previous layer's output and could materially reduce DFlash acceptance.

## Validation
- `python -m pytest -q test\python\builder\test_qwen_dflash2_export.py`
- repository-configured `lintrunner`
- targeted Ruff lint and format checks for the changed linted Python files

### DIFF
--- a/src/python/py/models/builders/qwen.py
+++ b/src/python/py/models/builders/qwen.py
@@ -1062,6 +1062,26 @@ class Qwen35MoEModel(MTPModel):
     def save_processing(self, model_name_or_path, extra_kwargs, out_dir):
         self.decoder.save_processing(model_name_or_path, extra_kwargs, out_dir)
 
+    def require_specforge_aux_taps(self, target_layer_ids, drafter_name):
+        if not target_layer_ids:
+            raise ValueError(f"The {drafter_name} checkpoint must define at least one target_layer_ids entry.")
+
+        aux_layers = [layer_id + 1 for layer_id in target_layer_ids]
+        untappable = [layer_id - 1 for layer_id in aux_layers if not 1 <= layer_id < self.decoder.num_layers]
+        if untappable:
+            raise ValueError(
+                f"The {drafter_name} checkpoint targets decoder layers {untappable}, whose outputs the exporter "
+                f"cannot expose; target_layer_ids must lie in [0, {self.decoder.num_layers - 1})."
+            )
+
+        expected = ",".join(str(layer_id) for layer_id in aux_layers)
+        actual = ",".join(str(layer_id) for layer_id in self.decoder.aux_hidden_state_layers)
+        if actual != expected:
+            raise ValueError(
+                f"The {drafter_name} drafter needs aux_hidden_state_layers={expected} on the main model, "
+                f"got '{actual}'."
+            )
+
     def make_dflash2_init(self, io_dtype, extra_options):
         """DFlash 2 block drafter, exported as an auxiliary ``dflash2.onnx``.
 
@@ -1099,12 +1119,7 @@ class Qwen35MoEModel(MTPModel):
                 f"dflash2_num_draft_tokens must not exceed the drafter checkpoint limit ({checkpoint_draft_limit})."
             )
         target_layer_ids = dflash_config["target_layer_ids"]
-        expected = ",".join(str(i + 1) for i in target_layer_ids)
-        actual = ",".join(str(i) for i in self.decoder.aux_hidden_state_layers)
-        if actual != expected:
-            raise ValueError(
-                f"The DFlash 2 drafter needs aux_hidden_state_layers={expected} on the main model, got '{actual}'."
-            )
+        self.require_specforge_aux_taps(target_layer_ids, "DFlash 2")
 
     def make_dflash2_model(self, input_path):
         if not self.dflash2_path:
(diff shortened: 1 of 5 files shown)


src/python/py/models/builders/qwen.py
```diff
@@ -1097,7 +1099,7 @@ def make_dflash2_init(self, io_dtype, extra_options):
                 f"dflash2_num_draft_tokens must not exceed the drafter checkpoint limit ({checkpoint_draft_limit})."
             )
         target_layer_ids = dflash_config["target_layer_ids"]
```
tianleiwu, line 1121: This validation block is now byte-identical to the one in `make_dspark_init` (lines 1206-1212):

```python
target_layer_ids = ...["target_layer_ids"]
expected = ",".join(str(i + 1) for i in target_layer_ids)
actual = ",".join(str(i) for i in self.decoder.aux_hidden_state_layers)
if actual != expected:
    raise ValueError(...)
```

The model-builder guidelines ask to reduce duplication by reusing existing functionality. A single helper on the composite model (e.g. `require_aux_taps(target_layer_ids, drafter_name)`) would keep the off-by-one rule in exactly one place, so the next drafter cannot repeat the mistake this PR is fixing.

It would also be the natural home for a degenerate case both paths currently accept: an empty `target_layer_ids` compares `"" == ""` and passes validation, producing a drafter with `aux_hidden_size == 0`.

src/python/py/models/builders/qwen.py
```diff
@@ -1097,7 +1099,7 @@ def make_dflash2_init(self, io_dtype, extra_options):
...
-        expected = ",".join(str(i) for i in target_layer_ids)
+        expected = ",".join(str(i + 1) for i in target_layer_ids)
```
tianleiwu: Boundary case worth a targeted check: with the `+ 1` shift, the largest representable `target_layer_ids` entry is `num_hidden_layers - 2`.

`base.py` restricts `aux_hidden_state_layers` to `[1, num_hidden_layers)` because the tap is only emitted while `not layernorm_attrs["last_layernorm"]`. So a SpecForge checkpoint whose `target_layer_ids` includes `num_hidden_layers - 1` makes this validator instruct the user to set `aux_hidden_state_layers=<num_hidden_layers>`, which `base.py` then rejects with an unrelated "outside [1, N)" message.

The failure is loud rather than silent, so this is not a blocker, but a direct check would save a confusing round-trip:

```python
max_tap = max(target_layer_ids, default=-1) + 1
if max_tap >= self.decoder.num_layers:
    raise ValueError(
        f"The DFlash 2 checkpoint taps target layer {max_tap - 1}, whose output the exporter "
        f"cannot expose (taps must lie in [1, {self.decoder.num_layers}))."
    )
```

tianleiwu (dismissed):
Verified the direction of the shift rather than taking it on trust, and it is right:

- `base.py` emits the tap from `output_3` of the **input** SkipLayerNorm of layer `i` (`root_input + skip_input`), i.e. the residual stream entering layer `i`, which is the **output of layer `i - 1`**.
- SpecForge `target_layer_ids` names the layer whose *output* is captured (`hidden_states[L + 1]` in HF terms).
- So the tap index is `L + 1`. ✅

This also removes an internal inconsistency: #2490 (DSpark, `0b40fbdb`) already applied the `+ 1` shift in `make_dspark_init`, while #2489 (DFlash 2, `78074c51`) passed the IDs through unchanged. The two block-drafter paths now agree, and the error message reports the shifted list so a user hitting the guard is told exactly what to pass.

Ran the exporter suites locally: `PYTHONPATH=src/python/py python3 -m pytest -q test/python/builder/test_qwen_dflash2_export.py test/python/builder/test_qwen_dspark_export.py` → **80 passed**. Adding `TARGET_LAYER_IDS` itself to the rejected parametrization is a good regression guard — reverting the shift now fails both the accept and the reject test.

Approving. Two non-blocking suggestions inline, plus two notes below.

**Backward compatibility (worth a line in the PR description / release notes).** DFlash 2 models exported before this commit record the unshifted IDs in `model.dflash2.aux_hidden_state_layers` and were built with taps one layer too early. Nothing at load time can detect that — the runtime only consumes the concatenated `aux_hidden_states` tensor and never cross-checks the layer list against the draft checkpoint — so existing DFlash 2 exports have to be regenerated.

**Nit (README).** The DFlash 2 example moves to `aux_hidden_state_layers=2,12,22` while the DSpark example immediately below still shows `1,11,21`. Both are self-consistent since the checkpoints are hypothetical, but a reader comparing the two adjacent sections may read the difference as meaningful.

src/python/py/models/builders/qwen.py
bmehta001: Addressed in 7da24420. `require_specforge_aux_taps` now owns the target-output to auxiliary-input mapping for both DFlash 2 and DSpark, and it rejects an empty `target_layer_ids` list before comparison. Both drafter test suites cover the empty-checkpoint case.

src/python/py/models/builders/qwen.py
bmehta001: Addressed in 7da24420. The shared validator now checks shifted taps against the exporter range before comparing `aux_hidden_state_layers`, and reports the valid checkpoint range directly. Both DFlash 2 and DSpark tests cover a checkpoint targeting the unexposable final decoder layer.

baijumeswani (approved):

tianleiwu (approved):
Reviewed the follow-up commit on the current head. The shared validator now owns the SpecForge layer-output to residual-input mapping for both DFlash 2 and DSpark, while preserving checkpoint order for the concatenated feature layout. It also rejects empty tap lists and unexposable final-layer targets with focused coverage, so both prior threads are fully addressed.

Validation: `PYTHONPATH=src/python/py python3 -m pytest -q test/python/builder/test_qwen_dflash2_export.py test/python/builder/test_qwen_dspark_export.py` (`84 passed`). Approving.

PR #2543, opened 2026-09-10
Expose terminal EOS tokens to Engine continuation hosts

bmehta001:
## Summary

- add a distinct `TerminalToken` Engine event flag for the EOS token selected to finish a turn
- keep EOS separate from visible `Token` events, retained sequence length, and generated-token usage
- expose flag-gated C++ and Python accessors and document how continuation hosts should consume the boundary
- cover dynamic, speculative, static-batching, capacity-one delivery, C ABI, and Python behavior

## Motivation

ONNX Runtime GenAI selects EOS to end an assistant turn but deliberately does not append that token to the retained request sequence or publish it as a visible token. A continuation host therefore cannot know the exact model-selected assistant boundary. Foundry Local currently reconstructs that boundary with a synthetic `__foundry_engine_assistant_boundary_` marker rendered through the model chat template.

Publishing the unappended EOS as `TerminalToken | TurnFinished` lets a host prepend the exact boundary token to the next `BeginTurn` input while suppressing it from assistant text. A separate flag avoids changing the established meaning of `Token`, and a separate event preserves both accepted speculative draft tokens and EOS when they occur in one step.

This does not introduce a default per-turn output limit. Hosts can continue leaving `max_generated_tokens` unset unless the user explicitly supplies a maximum; the request remains bounded by its configured session/context limit.

## Validation

- repository lint (`lintrunner -a`)
- full Engine C++ suite on Windows/VS 2026 and WSL/Linux: 606 tests per platform
- full Python Engine suite on Windows and WSL/Linux: 40 tests per platform
- public Engine C/C++ API tests on Windows and WSL/Linux: 8 tests per platform

### DIFF
--- a/docs/engine_c_api_spec.md
+++ b/docs/engine_c_api_spec.md
@@ -346,10 +346,13 @@ typedef uint32_t OgaEngineEventFlags;
 #define OgaEngineEventFlag_CapacityBlocked ((OgaEngineEventFlags)(1u << 2))
 #define OgaEngineEventFlag_Failed ((OgaEngineEventFlags)(1u << 3))
 #define OgaEngineEventFlag_Retryable ((OgaEngineEventFlags)(1u << 4))
+#define OgaEngineEventFlag_TerminalToken ((OgaEngineEventFlags)(1u << 5))
 ```
 
 `OgaEngineEventFlags` is a bitmask, not a mutually exclusive type. In particular, a final model step
-may emit one event with both `Token` and `TurnFinished`.
+may emit one event with both `Token` and `TurnFinished`. EOS completion instead emits a distinct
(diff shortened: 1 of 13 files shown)


bmehta001:
Closing this approach because a sampled terminal EOS is not a general chat-template boundary: non-EOS finishes have no sampled token, EOS may differ from EOT, and templates may require multi-token delimiters or role prefixes. Foundry Local is instead rendering and tokenizing the complete structured transcript, reusing the resident Engine request only when its exact retained token sequence is a prefix, appending the unmatched suffix, and rebuilding on mismatch or invalidation. Terminal-token observability can be proposed separately if Engine consumers need it independently of chat continuation.

PR #2544, opened 2026-09-10
Add INT4 paged KV and DFlash2 builder support

tianleiwu:
## Summary

Adds end-to-end builder and runtime support for INT4 paged KV caches and INT4 DFlash2 export. A Qwen3.8-27B model can now be exported as a directly servable INT4 target with per-channel INT4 KV and a quantized DFlash2 drafter, without editing `genai_config.json` or post-processing the ONNX graphs.

### INT4 paged KV cache

- Adds `int4_per_tensor` and `int4_per_channel` static-scale schemes, plus CUDA `int4_per_token` dynamic scaling.
- Stores two INT4 values per byte and carries bit width, scale granularity, and quantization metadata through builder output and runtime cache allocation.
- Supports dense and sparse hybrid layer IDs, scalar or per-channel calibrated scales, and per-token FP16 scale caches.
- Extends synthetic model generation plus C++ and Python coverage for INT4 paged attention.

### Directly servable exports

- Emits paged-engine settings including `max_scheduled_tokens`, `num_blocks`, and `gpu_utilization_factor` directly into `genai_config.json`.
- Supports `exclude_mtp` and `use_device_allocator_for_initializers` and emits `ep.cuda.fpa_intb_gemm=1` for prepacked exports.
- Rejects incompatible or no-op prepack configurations instead of advertising an optimization that was not applied.
- Keeps Qwen GatedDeltaNet gate projections out of INT4 quantization and fixes sparse KV scale-layer mapping.

### INT4 DFlash2

- Adds `dflash2_precision=int4` to quantize the drafter body and LM head.
- Keeps the BF16 drafter body in plain blockwise INT4 because the fpA-intB backend requires FP16 activations; the FP16 drafter LM head remains prepack eligible.
- Preserves target/drafter shared-initializer metadata and configures block drafting for greedy generation by default.

## End-to-end validation

Built Qwen3.8-27B from its Hugging Face target and DFlash2 checkpoints with an INT4 target, per-channel INT4 paged KV, and a seven-token INT4 drafter. The Olive-produced target and drafter payloads are byte-identical to a direct GenAI builder reference, and the final materialized artifact reloads both ONNX graphs with external data.

| Prompt tokens | Batch | Drafts | TTFT (ms) | Target forward (ms) | Throughput (tok/s) | Peak GPU (MiB) | Acceptance |
|---:|---:|---:|---:|---:|---:|---:|---:|
| 512 | 1 | 7 | 368.66 | 25.594 | 325.66 | 26,540.875 | 99.72% |
| 8,192 | 4 | 7 | 2,360.90 | 42.806 | 378.28 | 28,278.875 | 94.52% |
| 32,768 | 4 | 7 | 8,812.43 | 123.894 | 209.97 | 29,302.875 | 100.00% |
| 32,768 | 4 | 0 | 8,520.84 | 39.095 | 107.00 | 25,516.938 | N/A |

At 32K prompt and batch 4, DFlash2 provides 1.962x throughput for an additional 3,785.94 MiB of peak GPU memory.

| Evaluation | Direct reference | This export | Changed generations | McNemar p-value |
|---|---:|---:|---:|---:|
| MMLU-Pro | 663/800 (82.875%) | 663/800 (82.875%) | 0/800 | 1.0 |
| GPQA Diamond | 146/198 (73.737%) | 146/198 (73.737%) | 0/198 | 1.0 |

Both evaluations used fixed four-shard routing and same-process direct-reference repeats. DFlash2 recorded zero failures, disables, admission misses, and standard fallback steps.

## Tests

- `test/python/builder`: 667 passed, 4 skipped, 790 deselected
- Focused INT4 KV and DFlash2 suite: 122 passed
- Quantized KV suite after formatting: 75 passed
- clang-format 20 dry run passes for all affected C++ files
- Exact `og.Engine` load and generation validated with ORT 1.30.0 and the CUDA INT4 KV runtime enabled

## Downstream validation

- microsoft/Olive#2663 adds modern GenAI compatibility and preserves multi-file external-data artifacts during materialization.
- microsoft/olive-recipes#615 publishes the single verified optimized recipe and its latency, memory, MMLU-Pro, and GPQA results.


src/engine/paged_key_value_cache.cpp
```diff
@@ -286,29 +380,68 @@ PagedKeyValueCache::PagedKeyValueCache(std::shared_ptr<Model> model,
...
         ComposeKeyValueName(decoder.outputs.present_value_names, layer_id)});
+    const std::pair<std::string, std::string> scale_names[] = {
+        {decoder.inputs.past_key_scale_names, decoder.outputs.present_key_scale_names},
+        {decoder.inputs.past_value_scale_names, decoder.outputs.present_value_scale_names}};
+    for (const auto& [input_template, output_template] : scale_names) {
+      if (input_template.empty()) {
+        continue;
+      }
+      const auto input_name = ComposeKeyValueName(input_template, layer_id);
+      const auto output_name = output_template.empty() ? std::string{} : ComposeKeyValueName(output_template, layer_id);
```
baijumeswani: Could we require a matching output name whenever a per-token scale input is configured? These scales are produced by PagedAttention and must be saved for the next decode step. If the output template is missing, the Engine feeds the persistent buffer but does not bind it as an output, so later steps can read stale or zero scales.

Please require the past and present scale templates as a pair and always bind the same persistent buffer on both sides. If the operator is expected to update an input-only buffer in place, please document that contract and add a two-step test that proves the updated scale is read on the next step.

src/python/py/models/builders/qwen.py
```diff
@@ -1082,6 +1111,45 @@ def require_specforge_aux_taps(self, target_layer_ids, drafter_name):
...
+        if precision == "bf16":
+            return None
+        bits = 4 if precision == "int4" else 8
+        block_size = int(self.decoder.quant_attrs["matmul_block_size"])
+        prepack = int(self.decoder.matmul_attrs["weights_prepacked"])
+        quant = {"bits": bits, "block_size": block_size, "prepack": prepack, "lm_head": None}
+
+        if self.decoder.exclude_lm_head:
+            return quant
+        head_bits, weight_name, scales_name, zero_point_name = self.decoder.make_tied_quantized_embedding_input_names()
```
baijumeswani, line 1147: This helper can return a four-bit name even when the target LM head is dense or uses a different bit width. As a result, a dense FP16/BF16 target can get a separate INT4 drafter head, and a normal INT8 target can also be treated as Q4. That does not follow the documented rule that the drafter head uses the target's format.

Please first check whether the emitted target LM head is actually quantized, then use its real bit width, block size, names, zero-point mode, and prepack value. Keep the drafter head dense when the target head is dense. Tests for dense FP16, default INT4, default INT8, excluded, and mixed-precision target heads would cover the important cases.

src/engine/paged_key_value_cache.cpp
```diff
@@ -286,29 +380,68 @@ PagedKeyValueCache::PagedKeyValueCache(std::shared_ptr<Model> model,
...
+                                                   static_cast<int64_t>(decoder.num_key_value_heads),
+                                                   static_cast<int64_t>(CacheHeadSize(model, key_name))};
+    auto value_shape_per_layer = key_shape_per_layer;
+    value_shape_per_layer.back() = static_cast<int64_t>(CacheHeadSize(model, value_name));
     cache_.push_back(LayerCache{
-        OrtValue::CreateTensor(model->p_device_kvcache_->GetAllocator(), cache_shape_per_layer, dtype),  // Key cache
-        OrtValue::CreateTensor(model->p_device_kvcache_->GetAllocator(), cache_shape_per_layer, dtype),  // Value cache
-        ComposeKeyValueName(decoder.inputs.past_key_names, layer_id),
-        ComposeKeyValueName(decoder.inputs.past_value_names, layer_id),
+        OrtValue::CreateTensor(model->p_device_kvcache_->GetAllocator(), key_shape_per_layer, dtype),
```
baijumeswani, line 474: The memory calculation above reads each layer's cache type, but allocation here uses the type taken from the first paged layer for every key and value cache. If a graph has different key/value types or a later layer has a different type, the calculated memory can differ from what is allocated, or the graph binding can fail.

Please either require every key/value past/present tensor in the paged group to use one matching type, or allocate each tensor with its declared type and use that same type in the byte calculation. It would also help to validate the past/present shapes and types together during setup.

src/python/py/models/builders/qwen.py
```diff
@@ -542,10 +542,14 @@ def make_linear_attention_input_proj(self, layer_id, attention, root_input):
         z_name = f"{basename}/z_proj/MatMul"
         self.make_matmul(attention.in_proj_z, z_name, root_input)
 
+        # The decay and beta gates drive the GatedDeltaNet recurrence, and their weights are
+        # ~0.1% of the model, so they stay in fp16 regardless of which loader supplied them.
         b_name = f"{basename}/b_proj/MatMul"
+        self.exclude_node_from_quantization(b_name)
```
baijumeswani, line 549: Adding the name to `nodes_to_exclude` protects the later graph-quantization pass, but it does not protect a checkpoint loader that already supplies this weight as quantized. In that case `make_matmul()` takes the quantized branch immediately, even though the comment says the gate always stays FP16.

Please make the loader expose these gate weights as dense, or reject/warn when an already-quantized gate cannot be restored safely. A test using a pre-quantized loader would make the intended behavior clear.

test/cpp/engine/paged_key_value_cache_tests.cpp
```diff
@@ -365,6 +365,98 @@ TEST(PagedKeyValueCacheManifestTest, CacheManagerConstructsFromSparseManifest) {
...
+    const auto* scale_bytes = static_cast<const uint8_t*>(state.inputs_[index]->GetTensorRawData());
+    for (size_t offset = 0; offset < 128 * 4 * sizeof(Ort::Float16_t); ++offset) {
+      EXPECT_EQ(scale_bytes[offset], 0u);
+    }
+  }
+  EXPECT_STREQ(state.input_names_[4], "past_key_values.1.key_scale");
+  EXPECT_STREQ(state.output_names_[7], "present.4.value_scale");
+  auto* first_scale = state.inputs_[4];
+  cache->UpdateState(state, {});
+  EXPECT_EQ(state.inputs_[4], first_scale);
```
baijumeswani, line 418: These tests verify names, shapes, zero initialization, and pointer reuse, but they do not run PagedAttention or show that a nonzero scale written in one step is consumed in the next. That leaves the most important state-update behavior untested.

Please add a small executable two-step test with a real packed UINT8 INT4 cache. Have step one write nonzero scales and step two consume them. If practical, include sparse layer IDs and one windowed layer, plus an INT8 or unquantized case to protect existing behavior.

src/python/py/models/builder.py
```diff
@@ -759,6 +785,16 @@ def get_args():
                 dflash2_num_draft_tokens = Override the number of draft tokens the DFlash 2 block
                     drafter proposes per step. Must be positive and no greater than the draft checkpoint's
                     block size minus its anchor token. That checkpoint limit is the default.
+                dflash2_precision = Weight precision for the DFlash 2 drafter body: bf16 (default),
```
baijumeswani, line 788: This new option is explained in CLI help, but it is missing from the DFlash2 section of the model-builder README. The builder contribution rules require new extra options in both places.

Please add the supported values, default, BF16 activation requirement, body-versus-head behavior, prepack limitation, and a short INT4 example to the README.

src/python/py/models/README.md
```diff
@@ -301,6 +301,8 @@ This scenario is for when you want to build a model that uses the `PagedAttentio
...
 
+`max_scheduled_tokens` and `num_blocks` are the two remaining `engine.dynamic_batching` knobs and are written only when passed. `max_scheduled_tokens` caps the tokens in one dynamically batched forward pass and therefore caps the peak prefill activation, which is the largest transient in a long-context deployment. `num_blocks` pins the cache to an exact block count, so the reachable context is `num_blocks * paged_block_size` tokens; it is mutually exclusive with `gpu_utilization_factor`, which is omitted from the config when `num_blocks` is set.
```
baijumeswani: The reachable-context statement is too broad. `num_blocks * paged_block_size` is only the single-request upper bound when the target owns all configured blocks. MTP or a full-attention auxiliary cache can reduce the target pool, and concurrent requests share it.

Please describe `num_blocks` as the total pre-auxiliary block budget and note that auxiliary caches and other resident requests reduce the context available to one request.

src/python/py/models/README.md
```diff
@@ -730,9 +752,25 @@ python builder.py -i path_to_local_folder_on_disk -o path_to_output_folder -p pr
...
 attributes.
 
-Only `int8_*` and `fp8_*` are supported on the paged path; `int4_*` is rejected because `PagedAttention` has no
-sub-byte cache backend. Per-channel scales are emitted with the `(num_kv_heads, 1, head_size)` shape that
-`PagedAttention` requires.
+The CUDA paged path supports INT4, INT8, and FP8. INT4 requires an ORT build with
+`onnxruntime_USE_INT4_KV_CACHE=ON` and stores two signed values per UINT8 byte, with physical
+cache width `(head_size + 1) // 2`. Per-channel scales retain shape `(num_kv_heads, 1, head_size)`.
+The `int4_per_token` and `int8_per_token` schemes use separate FP16 scale caches of shape
+`[num_blocks, block_size, num_kv_heads]`, with input/output aliases emitted in `genai_config.json`.
```
baijumeswani, line 779: Could we also document this new Engine-owned state in `docs/paged_attention_engine.md`? The runtime now owns persistent scale buffers, initializes them, maps them across sparse and windowed layers, includes them in memory sizing, and aliases them from input to output on every step. Those rules are important for anyone producing a model package without this builder.

baijumeswani (commented):
This PR adds end-to-end support for INT4 paged KV caches and quantized DFlash2 exports. It carries the cache bit width and scale layout from the builder into the Engine, adds static and per-token scale modes, accounts for packed cache memory, emits the Engine settings needed for a directly usable package, and can quantize the DFlash2 body and LM head.

There is a lot that looks good here. The INT4 cache uses the graph's physical packed width, scale-cache memory is included in capacity calculations, sparse layer IDs are handled, invalid calibration metadata is rejected, and the builder now writes settings that previously required manual JSON changes. The DFlash2 path also avoids unsupported prepacking for its BF16 body. The reported long-context, memory, acceptance, and accuracy results are strong evidence for the main Qwen3.8 setup.

I found a few cases outside that main setup that should be tightened before approval. The most important ones are requiring dynamic scale outputs, selecting the drafter LM-head format from the actual target head, and making cache allocation use the same data types as memory accounting. I also left recommendations for the already-quantized gate path, executable cache coverage, and documentation.

Overall, the approach is strong, but I do not think the PR is ready for approval until the main correctness comments below are addressed.

src/python/py/models/README.md
```diff
@@ -568,6 +572,8 @@ python builder.py -m model_name -o path_to_output_folder -p int4 -e cuda --extra
 
 This scenario is for when you want to control the CUDA MatMulNBits (int4/int8) weight layout. The default value is `0`, which exports raw blockwise weights. Use `1` to export the SM80/Ampere `fpA_intB` prepacked layout, or `2` to export the SM90/Hopper `fpA_intB` prepacked layout. This only applies to the CUDA EP, and an offline-prepacked model must be run with `ORT_FPA_INTB_GEMM` enabling the relevant nbits.
 
+A prepacked export therefore also writes `ep.cuda.fpa_intb_gemm=1` into the decoder's session options, which is the per-session equivalent of that environment variable, so the model is self-describing. A prepacked node takes the `fpA_intB` path regardless of the flag; the flag matters for the nodes the prepack pass skipped because their `N`, `K`, or `block_size` is not supported by the layout, keeping the whole model on one kernel family.
```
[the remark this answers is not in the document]
tianleiwu: Addressed in 5ad6be04. The README, CLI help, builder docstring, and quantization guide now state that eligible prepacked nodes select fpA-intB automatically. The emitted session option is optional for those nodes and only keeps ineligible/raw-layout nodes on the same kernel family.

src/engine/paged_key_value_cache.cpp
tianleiwu: Addressed in 5ad6be04. Past/present scale templates are now required as a pair for each side, and every scale buffer is always bound as the same persistent OrtValue on input and output. Tests cover all unpaired configurations, pointer aliasing, persistence across rebinds, and executable prefill/decode Engine steps.

src/python/py/models/builders/qwen.py
tianleiwu: Addressed in 5ad6be04. The drafter first checks whether the emitted target LM head is actually quantized, derives the real target bit width (including global INT8 and mixed last-matmul INT8), and only quantizes/shares layouts it can reproduce safely. Dense, excluded, asymmetric, or unsupported target-head layouts keep the drafter head dense. Regression tests cover dense FP16, default INT4, default INT8, excluded, unshareable, and mixed-precision heads.

src/engine/paged_key_value_cache.cpp
tianleiwu: Addressed in 5ad6be04. Cache setup now validates matching past/present key/value types within every layer and a uniform type across participating layers. Both allocation and byte accounting use that validated type. A regression test compares every allocated binding type and total allocated bytes against PagedKeyValueCacheBytesPerBlock for plain and quantized fixtures.

src/python/py/models/builders/qwen.py
tianleiwu: Addressed in 5ad6be04. The ModelOpt/compressed-tensors loader now dequantizes pre-quantized GatedDeltaNet decay/beta gate projections before exposing them to the builder. The builder also rejects any pre-quantized gate from another loader that was not restored to dense form, rather than silently emitting a quantized gate. Both paths have regression coverage.

test/cpp/engine/paged_key_value_cache_tests.cpp
tianleiwu: Addressed in 5ad6be04. Added an executable Engine test that runs prefill and decode steps with the synthetic packed-UINT8 scale-cache model, plus a two-bind persistence test that writes a sentinel through the output alias and verifies the same device pointer and value on the next input bind. The fixture is structural rather than a CUDA numerical kernel test, so it proves Engine pairing/aliasing/persistence and real ORT-run acceptance; end-to-end numerical INT4 scale correctness remains covered by the reported GPU validation.

src/python/py/models/builder.py
tianleiwu: Addressed in 5ad6be04. The DFlash2 README section now documents bf16/int4/int8, the BF16 activation requirement, dense-versus-quantized body behavior, the fpA-intB prepack limitation, target-head sharing/fallback behavior, and an INT4 wheel/source example.

src/python/py/models/README.md
tianleiwu: Addressed in 5ad6be04. num_blocks is now documented as the total pre-auxiliary block budget. The text notes that MTP/full-attention drafter reservations reduce the target pool, resident requests share it, and num_blocks * paged_block_size is only a single-request upper bound when the target owns every configured block.

src/python/py/models/README.md
tianleiwu: Addressed in 5ad6be04. docs/paged_attention_engine.md now has a dedicated per-token scale-cache section covering configuration and pairing, sparse/windowed addressing, shapes/types, ownership and zero initialization, input/output aliasing, binding layout, memory sizing, MTP/DFlash2 limitations, and test coverage.

baijumeswani (approved):

PR #2551, opened 2026-09-12
Fix accepted draft EOS classification after sampler slot reuse

baijumeswani:
Fix Engine completion classification when an accepted speculative draft is EOS and batched-sampler compaction reuses its next-token slot.
 
 ## Problem
 
 Accepted draft processing previously checked EOS through `Search::GetNextTokens()` during completion staging. That storage is shared by the batched sampler and can be rebound to another request after slot compaction. As a result, an accepted EOS could be overwritten before the request classified its completion, causing the request to miss EOS termination.
 
 ## Changes
 
 - Capture whether an accepted draft is EOS while committing the draft, before sampler-slot reuse can occur.
 - Store that result in request-owned transactional state and use it during completion staging.
 - Reset the state across commit, rollback, discard, retry, and turn lifecycle boundaries.
 - Add a regression test covering shared-slot overwrite, repeated completion staging, immediate and queued rollback, deterministic retry, and a subsequent non-EOS turn.

### DIFF
--- a/src/engine/request.cpp
+++ b/src/engine/request.cpp
@@ -324,6 +324,7 @@ void Request::MarkClosedFromEngine(const Engine& engine) noexcept {
   accepted_draft_count_ = 0;
   evaluated_draft_count_ = 0;
   draft_verification_completed_generation_ = false;
+  draft_verification_eos_ = false;
   draft_verification_stop_match_index_ = -1;
 }
 
@@ -369,6 +370,7 @@ void Request::ReleaseTurnResources() noexcept {
   accepted_draft_count_ = 0;
   evaluated_draft_count_ = 0;
   draft_verification_completed_generation_ = false;
+  draft_verification_eos_ = false;
   draft_verification_stop_match_index_ = -1;
   // The turn's reseed dies with the turn. A reseed that a sampling step already committed was
   // promoted to current_seed_basis_ by CommitStateForTransaction(), which runs before any terminal
@@ -428,6 +430,7 @@ void Request::CompleteClose() noexcept {
   accepted_draft_count_ = 0;
   evaluated_draft_count_ = 0;
   draft_verification_completed_generation_ = false;
+  draft_verification_eos_ = false;
   draft_verification_stop_match_index_ = -1;
   std::vector<int32_t>{}.swap(tokens_host_);
 }
@@ -552,6 +555,7 @@ void Request::AppendDraftsForTransaction(size_t draft_count) {
   accepted_draft_count_ = 0;
   evaluated_draft_count_ = 0;
   draft_verification_completed_generation_ = false;
+  draft_verification_eos_ = false;
   draft_verification_stop_match_index_ = -1;
 }
 
@@ -568,6 +572,7 @@ void Request::CommitAcceptedDraftsForTransaction(size_t accepted_count) {
   accepted_draft_count_ = 0;
   evaluated_draft_count_ = 0;
   draft_verification_completed_generation_ = false;
+  draft_verification_eos_ = false;
   draft_verification_stop_match_index_ = -1;
 
   for (size_t offset = 0; offset < accepted_count; ++offset) {
@@ -606,6 +611,8 @@ void Request::CommitAcceptedDraftsForTransaction(size_t accepted_count) {
             *turn_policy_.max_generated_tokens;
     if (search_->IsDone() || turn_limit_reached) {
       draft_verification_completed_generation_ = true;
+      draft_verification_eos_ =
+          search_->IsDone() && contains(params_->config.model.eos_token_id, token);
       break;
     }
   }
@@ -673,6 +680,7 @@ void Request::DiscardStagedDrafts() noexcept {
   accepted_draft_count_ = 0;
   evaluated_draft_count_ = 0;
   draft_verification_completed_generation_ = false;
+  draft_verification_eos_ = false;
   draft_verification_stop_match_index_ = -1;
 }
 
@@ -901,12 +909,8 @@ RequestStepResult Request::StageDraftCompletionForTransaction() {
             *turn_policy_.max_generated_tokens;
     if (turn_limit_reached) {
       finish_reason = GenerationFinishReason::TurnLimit;
-    } else {
-      const auto next_tokens = search_->GetNextTokens().CpuSpan();
-      if (search_->IsDone() && !next_tokens.empty() &&
-          contains(params_->config.model.eos_token_id, next_tokens.back())) {
-        finish_reason = GenerationFinishReason::EosToken;
-      }
+    } else if (draft_verification_eos_) {
+      finish_reason = GenerationFinishReason::EosToken;
     }
   }
   RequestStepResult result{
@@ -930,6 +934,7 @@ void Request::RestoreStateForTransaction() {
   evaluated_draft_count_ = 0;
   scheduled_token_count_ = 0;
   draft_verification_completed_generation_ = false;
+  draft_verification_eos_ = false;
   draft_verification_stop_match_index_ = -1;
   // The pending reseed itself is deliberately kept so the retry reseeds identically; only its
   // "already applied to the live streams" marker is undone, because rng_ was just rolled back.
@@ -955,6 +960,7 @@ void Request::QueueStateRestoreForTransaction() {
   evaluated_draft_count_ = 0;
   scheduled_token_count_ = 0;
   draft_verification_completed_generation_ = false;
+  draft_verification_eos_ = false;
   draft_verification_stop_match_index_ = -1;
   pending_reseed_applied_ = false;
   // Deliberately does not replay the stop controller yet: like the guidance checkpoint swap below,
@@ -1013,6 +1019,7 @@ void Request::CommitStep(const RequestStepPlan& plan,
   accepted_draft_count_ = 0;
   evaluated_draft_count_ = 0;
   draft_verification_completed_generation_ = false;
+  draft_verification_eos_ = false;
   draft_verification_stop_match_index_ = -1;
 }
 
--- a/src/engine/request.cpp
+++ b/src/engine/request.cpp
@@ -323,9 +323,7 @@ void Request::MarkClosedFromEngine(const Engine& engine) noexcept {
   staged_draft_count_ = 0;
   accepted_draft_count_ = 0;
   evaluated_draft_count_ = 0;
-  draft_verification_completed_generation_ = false;
-  draft_verification_eos_ = false;
-  draft_verification_stop_match_index_ = -1;
+  draft_verification_.Reset();
 }
 
 void Request::CompleteCloseFromEngine(const Engine& engine) noexcept {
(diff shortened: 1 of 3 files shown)


tianleiwu (commented):
Reviewed the accepted-draft EOS classification fix. The approach is correct and minimal: replacing a read of shared, rebindable sampler storage (`Search::GetNextTokens()`) with a request-owned boolean captured at commit time is the right shape of fix for this bug class.

**Verified independently while reviewing:**

- **Reset-site parity is complete.** All 9 sites that reset `draft_verification_stop_match_index_` also reset `draft_verification_eos_` (`request.cpp` 327/373/433/558/575/683/937/963/1022). No omission.
- **Repeated staging is now genuinely deterministic.** `ScheduledRequests::GenerateNextTokensForTransaction()` calls `StageDraftCompletionForTransaction()` twice for the same request (`scheduled_requests.cpp:749` and `:893`). Both calls now depend only on request-owned state, and `StageVisibleTokens()` is a pure function of `tokens_host_`, so both invocations produce identical results.
- **No sibling instance of the same bug.** The only other `GetNextTokens()` reads in the engine are `StageGeneration()` (`request.cpp:1085`) and `CompleteGeneration()` (`request.cpp:1226`). Both run immediately after each request binds its own distinct `next_tokens.subspan(i, 1)` row, with no compaction in between (`scheduled_requests.cpp:459-470`, `:776-785`), so neither is exposed to slot reuse.
- **Using the committed draft `token` is strictly more robust** than `next_tokens.back()`: it is the value the request actually committed, independent of any device-side rewriting of the shared next-token buffer (e.g. CUDA's `Launch_CheckForEOSAndPad` pad substitution).
- **The new test is a genuine regression test.** It asserts `slots.CpuSpan()[0] == 12` (the survivor's token overwriting the slot) *before* asserting `EosToken`, so the pre-fix code would classify `ContextLimit` and fail. Coverage of EOS at offsets 0 and 1, immediate vs. queued rollback, deterministic retry, and a following non-EOS turn is appropriate. The `SharedSlotDevice` partial-forwarding decorator matches the established convention in `request_lifecycle_tests.cpp` and `engine_run_tests.cpp`.

Only non-blocking suggestions and nitpicks below. I did not build or run the branch; the claims above come from source reading.

src/engine/request.h
```diff
@@ -608,15 +608,18 @@ struct Request : std::enable_shared_from_this<Request>,
   // advances it.
   size_t evaluated_draft_count_{};
   bool draft_verification_completed_generation_{};
+  // Capture terminal EOS before another request can reuse the batched sampler's next-token slot.
+  // Like the stop match below, this is immutable until the transaction is committed or reset.
+  bool draft_verification_eos_{};
```
tianleiwu: Suggestion (non-blocking, maintainability): `draft_verification_completed_generation_`, `draft_verification_eos_`, and `draft_verification_stop_match_index_` now form one logical unit whose three members must be reset in lockstep at **9** separate sites. This PR adds the third copy-pasted assignment to every one of them.

I checked and the parity is currently complete, but given that the bug being fixed is itself a state-lifetime bug, the copy-paste pattern is exactly the hazard that would reintroduce it. Consider grouping the fields so a future fourth member cannot be silently missed at one boundary:

```cpp
struct DraftVerificationState {
  bool completed_generation{};
  bool eos{};
  int32_t stop_match_index{-1};
  void Reset() noexcept { *this = DraftVerificationState{}; }
};
```

Each of the 9 sites then becomes a single `draft_verification_.Reset();`.

src/engine/request.h
```diff
@@ -608,15 +608,18 @@ struct Request : std::enable_shared_from_this<Request>,
...
   // verification early, giving StopString precedence over the turn/context limit for the same
   // completing token. -1 when verification completed generation for any other reason (or has not
   // completed yet). The sampled/batched path never sets this: it observes stop matches through
   // the same StageGenerationForTransaction()/StageGeneration() path the ordinary one-token step
   // uses, one stage at a time, so it needs no separate bookkeeping here.
   //
-  // StageDraftCompletionForTransaction() only ever reads this field (and accepted_draft_count_,
-  // and Search's own state) to compute its result; it never resets or otherwise mutates it. Only
+  // StageDraftCompletionForTransaction() only ever reads these fields (and accepted_draft_count_)
+  // to compute its result; it never resets or otherwise mutates them. Only
```
tianleiwu: Nitpick (comment drift): the block above was pluralized to cover both fields, but two details did not follow.

1. `"... reset it, at their own well-defined transaction boundaries"` should be `"reset them"` to match the pluralization on the preceding lines.
2. The enumerated list is presented as exhaustive ("Only ... reset it") but omits four other reset sites this PR touches: `MarkClosedFromEngine()`, `ReleaseTurnResources()`, `CompleteClose()`, and `CommitAcceptedDraftsForTransaction()`'s own prologue. Since the PR is already editing this comment and adding resets in those functions, worth either completing the list or softening it to "only at transaction/turn-lifecycle boundaries".

src/engine/request.cpp
```diff
@@ -901,12 +909,8 @@ RequestStepResult Request::StageDraftCompletionForTransaction() {
             *turn_policy_.max_generated_tokens;
     if (turn_limit_reached) {
       finish_reason = GenerationFinishReason::TurnLimit;
-    } else {
-      const auto next_tokens = search_->GetNextTokens().CpuSpan();
-      if (search_->IsDone() && !next_tokens.empty() &&
-          contains(params_->config.model.eos_token_id, next_tokens.back())) {
-        finish_reason = GenerationFinishReason::EosToken;
-      }
+    } else if (draft_verification_eos_) {
```
tianleiwu: Nitpick: this branch orders `turn_limit_reached` **before** EOS, whereas the ordinary one-token path in `StageGeneration()` gives EOS precedence over the turn limit. The ordering is pre-existing, and the two conditions do appear mutually exclusive in practice — an accepted EOS draft never appends, so it cannot advance `accepted_draft_count_` past the limit at the same offset, and an earlier offset that hit the limit would have already broken out of the loop in `CommitAcceptedDraftsForTransaction()` with `draft_verification_eos_` left `false`. The only residual window is a turn entered already at `max_generated_tokens`.

Now that the new field makes this branch explicit, a one-line comment noting the divergence from `StageGeneration()` is deliberate (and why it is unreachable) would save the next reader the same analysis.

src/engine/request.cpp
```diff
@@ -606,6 +611,8 @@ void Request::CommitAcceptedDraftsForTransaction(size_t accepted_count) {
             *turn_policy_.max_generated_tokens;
     if (search_->IsDone() || turn_limit_reached) {
       draft_verification_completed_generation_ = true;
+      draft_verification_eos_ =
```
tianleiwu: Worth calling out as a genuine improvement beyond the stated slot-reuse fix: keying off the committed draft `token` rather than `next_tokens.back()` also removes a dependency on the device not rewriting the shared next-token buffer. On CUDA, `Launch_CheckForEOSAndPad` substitutes `pad_token_id` into `next_tokens_` once EOS has been seen, so the old read was fragile for more than one reason.

test/cpp/engine/stop_string_engine_tests.cpp
```diff
@@ -133,6 +133,143 @@ class StopStringEngineTest : public ::testing::Test {
   std::shared_ptr<Model> model_;
 };
 
+// Exercise CUDA's shared next-token-slot lifetime with real CPU EOS/search semantics.
+class SharedSlotSearch final : public GreedySearch_Cpu {
+ public:
+  using GreedySearch_Cpu::GreedySearch_Cpu;
+  bool BindNextTokensSlot(DeviceSpan<int32_t> slot) override {
```
tianleiwu, line 144: Nitpick (placement): this test drives `Request`/`Search` transactional APIs directly, which is what the `// --- Request-level transactional behavior ---` section immediately below this insertion point is for. Moving it below that banner would keep the file's section organization intact.

Also minor test-fidelity: `SharedSlotSearch::BindNextTokensSlot` validates only `slot.size() != 1`, while the real `GreedySearch_Cuda::BindNextTokensSlot` it is standing in for also rejects `params_->BatchBeamSize() != 1`. Mirroring that guard would keep the double honest if the batch-beam assumption ever changes.

src/engine/request.h
```diff
@@ -607,25 +607,29 @@ struct Request : std::enable_shared_from_this<Request>,
...
+  // StageGenerationForTransaction()/StageGeneration() path the ordinary one-token step uses, one
+  // stage at a time, so it needs no separate bookkeeping here.
+  struct DraftVerificationState {
+    bool completed_generation{};
+    bool eos{};
+    int32_t stop_match_index{-1};
+
+    void Reset() noexcept { *this = DraftVerificationState{}; }
+  };
   //
```
tianleiwu, line 623: Nitpick (cosmetic, non-blocking): inserting the struct definition into the middle of what was one continuous doc comment leaves this bare `//` dangling between the closing `};` and the second paragraph, which reads as leftover. Moving the `DraftVerificationState` definition above the whole comment block — or below it, immediately before the member — would keep the documentation contiguous.

Also minor: `DraftVerificationState draft_verification_;` on line 632 omits the `{}` that every sibling member in this section uses (`size_t evaluated_draft_count_{};` etc.). Harmless since the default member initializers cover it, but inconsistent with the local style.

tianleiwu (dismissed):
Re-reviewed at `ad4c487a`. All five round-1 comments are addressed, and the follow-up commit is a clean, behavior-preserving refactor.

**Verification performed on this head:**

- **No stale identifiers.** `grep` for `draft_verification_completed_generation_`, `draft_verification_eos_`, and `draft_verification_stop_match_index_` across `src/` and `test/` returns zero hits — the rename is complete.
- **Reset-site count preserved.** Exactly 9 `draft_verification_.Reset();` call sites in `request.cpp`, matching the 9 three-line blocks they replaced. No boundary lost a reset in the conversion.
- **Compiles clean.** Syntax-checked both `src/engine/request.cpp` and `test/cpp/engine/stop_string_engine_tests.cpp` at this head using the project's own `CXX_DEFINES`/`CXX_INCLUDES`/`CXX_FLAGS` from `build/cu130/Release/CMakeFiles/onnxruntime-genai-obj.dir/flags.make`. No diagnostics. This also confirms `Search::params_` is reachable from the test double, which the newly added `params_->BatchBeamSize()` guard depends on.
- **`Reset()` semantics are correct.** `*this = DraftVerificationState{};` restores `stop_match_index` to `-1`, not `0`, so the pre-refactor sentinel is preserved. `noexcept` is appropriate for a trivially copyable aggregate, which matters because several call sites are inside `noexcept` functions (`MarkClosedFromEngine`, `ReleaseTurnResources`, `CompleteClose`, `DiscardStagedDrafts`).
- **Behavior unchanged.** The only semantic line, `draft_verification_.eos = search_->IsDone() && contains(...)`, is logically identical to the previous revision.

One cosmetic nitpick inline; not blocking. Note that I syntax-checked but did not link or run the tests.

tianleiwu (approved):

PR #2555, opened 2026-09-12
Fix get_logits after generator rewind

Pdevadiga45:
## Description
After `Generator.rewind_to(N)`, search keeps the boundary token queued so generation can replay it. Standard `get_logits()` skipped that replay and called the model with inconsistent sequence and attention-mask state, causing continuous decoding to fail in `GptAttention_0`.

Replay the queued token before computing logits after a nonzero rewind. For `rewind_to(0)`, reject `get_logits()` before mutation so the generator remains reusable. Fresh issue and PR searches found no matching work.

## Tests
- CPU `RelWithDebInfo` build with `--parallel 2`
- `test_onnxruntime_genai_api.py::test_rewind` — 1 passed
- `test_onnxruntime_genai_api.py::test_rewind_cuda` — CPU fixture passed locally; existing fp32/fp16 CUDA fixtures run when CUDA is available
- Runnable API module — 11 passed, 45 skipped, 8 unavailable-fixture cases deselected
- clang-format 20.1.8 and Ruff 0.12.12 checks

### DIFF
--- a/src/generator/generators.cpp
+++ b/src/generator/generators.cpp
@@ -994,7 +994,15 @@ DeviceSpan<float> Generator::GetLogits() {
     return strategy_logits;
   }
   if (!computed_logits_) {
-    ComputeLogits(search_->GetNextTokens());
+    auto next_tokens = search_->GetNextTokens();
+    if (last_action_ == Action::rewound) {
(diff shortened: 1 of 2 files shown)


Pdevadiga45:

PR #2556, opened 2026-09-13
Capture CUDA graphs for paged decode steps

tianleiwu:
## Description

The paged Engine advertises CUDA graphs through `enable_cuda_graph`, but on a hybrid speculative model it never captured anything: an nsys trace of Qwen3.8-27B with a DFlash2 drafter showed **zero `cudaGraphLaunch` calls**, 1707 kernel launches per decode round, and 19.8% of each round spent GPU-idle in launch gaps.

Three independent gates each individually blocked capture, and there was no diagnostic for any of them:

- **Packed `position_ids`** — `SimpleDecoder` only allocated graph buffers when the model had no `position_ids`, and `PreparePositionIds` threw for captured steps while allocating a fresh `[3, num_tokens]` tensor every step. An MRoPE model can never satisfy this.
- **One token per request** — `graph_capture_eligible` required `unprocessed_token_count == 1`. With a block drafter every decode step verifies a whole block, so no step ever qualified.
- **Fixed decoder state** — hybrid models were excluded outright, which covers every Qwen3.5/3.8 export.

## Summary of Changes

- Give `position_ids` a persistent buffer in `VarlenGraphBuffers` and borrow it on captured steps instead of throwing.
- Allow multi-token steps. A step qualifies when it is not prefill and **every request contributes the same token count**; that uniformity is what lets one captured shape describe the whole step. Mixed steps stay eager.
- Support hybrid models. Fixed decoder state lives in two persistent banks whose active bank flips on every commit, and bindings are direct views into them, so the bound device pointer alternates between two addresses. `FixedStateReservation::BindingLayoutKey()` reports that layout and it becomes part of the graph key.
- Key every graph on `(batch, tokens per request, block-table width bucket, state binding layout)` via a shared `GraphAnnotationIds` allocator, bounded to 64 shapes. Anything past the budget runs eagerly rather than growing capture memory without bound.
- Size the buffers by packed rows rather than batch, and derive the captured attention bounds from the step instead of hardcoding a query length of 1.

Buffers fail safe: a step that does not fit simply runs eagerly.

### Why a shape is captured on first sight

The second commit removes a "withhold the id until the shape recurs" rule that the first commit shipped. It was meant to stop a churning workload from spending the whole 64-shape budget on graphs it never replays, but measured on H200 it only moves the capture stalls out of warmup and into service. Every steady-state case got worse:

| case | recur-first | first sight |
|---|---|---|
| ctx512 b1 k7 | 1.129x | **1.165x** |
| ctx2048 | 1.112x | **1.151x** |
| concurrent4 | 0.848x | **1.058x** |

Ratios are graph-on over the eager control in the same build. The budget alone already bounds capture memory, and the decoder's key is small and recurrent by construction.

## Testing

`lintrunner` clean; **642/642** engine unit tests pass, including coverage for shape separation, id stability, block-table bucketing, and budget exhaustion.

End-to-end on H200 with Qwen3.8-27B + DFlash2, A/B inside one build toggling only `enable_cuda_graph`:

- **Correctness:** 10 cases × 4 samples × 2 arms, **19,060 generated tokens per arm, token sequences byte-identical, zero divergences**, deterministic across processes. Covers long generations, paged block-boundary crossing, and concurrent requests. This is the gate that matters, because a wrong graph key would replay against the wrong state bank and silently corrupt recurrent state rather than fail loudly.
- **Capture:** 1.00 `cudaGraphLaunch` per decode round versus 0 before; eager launches 1720 → 374 (−78%). Graph ids strictly alternate in pairs, exactly the two-graphs-per-steady-shape the bank flip predicts.
- **Performance:** ctx512 batch1 k7 **199.06 → 225.20 tok/s (+13.1%)**, step p50 20.04 → 17.73 ms, arm ranges non-overlapping. GPU busy is unchanged (+0.6%) while idle drops 2345 µs/step against a 2310 µs measured gain, so the win is removed launch gaps rather than less work.
- **Memory:** peak 41030 → 42136 MiB (+2.7%).

## Known Limitations

- First sight of a shape pays a capture stall, up to ~105 ms p95 and 143 ms worst case. On a shape-diverse cold start this can outweigh the steady-state gain: four-way concurrency measures 0.77x over its first encounter before recovering to 1.06x.
- The control arm also reverts block-table allocation, so it is not a pure capture-only isolation.
- Measured on one GPU, one model, greedy decoding only. No soak test.

### DIFF
--- a/src/engine/decoders/simple_decoder.cpp
+++ b/src/engine/decoders/simple_decoder.cpp
@@ -36,38 +36,60 @@ SimpleDecoder::SimpleDecoder(std::shared_ptr<DecoderOnly_Model> model,
     position_planes_ = position_shape.size() == 2 ? 3 : 1;
   }
   if (IsGraphCaptureEnabled(model_->config_->model.decoder.session_options) &&
-      cache_manager_->SupportsDynamicBatching() &&
-      !has_fixed_state_groups_ &&
-      !has_position_ids) {
-    graph_buffers_ = std::make_unique<VarlenGraphBuffers>(*model_);
+      cache_manager_->SupportsDynamicBatching()) {
+    graph_buffers_ = std::make_unique<VarlenGraphBuffers>(*model_, position_planes_);
   }
 }
 
 namespace {
 
-// A step can be captured only if it looks like every other step the graph will serve: one new token
-// per sequence. Prefill and chunked-prefill steps carry a variable number of tokens and are shaped
-// differently from run to run, so they always execute eagerly.
-bool IsPureDecodeStep(ScheduledRequests& scheduled_requests) {
-  if (scheduled_requests.size() == 0) {
-    return false;
-  }
+// A step can be captured only if it looks like every other step the graph will serve: the same
+// number of new tokens for every sequence. Prefill and chunked-prefill steps carry a variable number
+// of tokens and are shaped differently from run to run, so they always execute eagerly. Returns the
+// shared per-request token count, or zero when the step cannot be captured.
+size_t UniformDecodeTokenCount(ScheduledRequests& scheduled_requests) {
+  size_t tokens_per_request = 0;
   for (auto& request : scheduled_requests) {
-    if (request->IsPrefill() || request->ScheduledTokenCount() != 1) {
-      return false;
+    const size_t token_count = request->ScheduledTokenCount();
+    if (request->IsPrefill() || token_count == 0) {
+      return 0;
+    }
+    if (tokens_per_request == 0) {
+      tokens_per_request = token_count;
+    } else if (tokens_per_request != token_count) {
+      return 0;
     }
   }
-  return true;
+  return tokens_per_request;
+}
+
+// The scheduler already proved a capture-eligible plan is uniform, so any row carries the count.
+size_t PlanDecodeTokenCount(const StepPlan& plan) {
+  if (!plan.graph_capture_eligible || plan.requests.empty()) {
+    return 0;
+  }
+  return plan.requests.front().unprocessed_token_count;
 }
 
 }  // namespace
 
 void SimpleDecoder::Decode(ScheduledRequests& scheduled_requests,
                            ExecutionContext& context) {
-  const bool capture = graph_buffers_ != nullptr &&
-                       graph_buffers_->Fits(scheduled_requests.size()) &&
-                       (context.plan ? context.plan->graph_capture_eligible
-                                     : IsPureDecodeStep(scheduled_requests));
+  const size_t tokens_per_request = context.plan
+                                        ? PlanDecodeTokenCount(*context.plan)
+                                        : UniformDecodeTokenCount(scheduled_requests);
+  // -1 tells the CUDA EP to run eagerly. Otherwise every distinct decode shape gets its own
+  // annotation id: the EP runs that id once eagerly, captures it on the next occurrence, and
+  // replays from then on. The id also covers the fixed-state binding layout, whose device
+  // addresses move when the persistent bank flips.
+  int annotation_id = -1;
+  if (graph_buffers_ != nullptr && tokens_per_request != 0 &&
+      graph_buffers_->Fits(scheduled_requests.size(), tokens_per_request)) {
+    annotation_id = graph_buffers_->GraphId(scheduled_requests.size(), tokens_per_request,
+                                            context.block_table_columns,
+                                            context.fixed_state_binding_key);
+  }
+  const bool capture = annotation_id > 0;
 
   std::unique_ptr<DecoderIO> decoder_state;
   if (!cache_manager_->SupportsDynamicBatching()) {
@@ -76,7 +98,7 @@ void SimpleDecoder::Decode(ScheduledRequests& scheduled_requests,
   } else if (has_fixed_state_groups_) {
     decoder_state = std::make_unique<HybridDecoderIO>(
         model_, scheduled_requests, cache_manager_, context,
-        position_planes_);
+        capture ? graph_buffers_.get() : nullptr, position_planes_);
   } else {
     decoder_state = std::make_unique<VarlenDecoderIO>(
         model_, scheduled_requests, cache_manager_, &context,
@@ -84,19 +106,8 @@ void SimpleDecoder::Decode(ScheduledRequests& scheduled_requests,
   }
 
   if (IsGraphCaptureEnabled(model_->config_->model.decoder.session_options) &&
-      cache_manager_->SupportsDynamicBatching() &&
-      graph_buffers_ == nullptr) {
-    // Inputs without persistent graph buffers have per-step addresses and must stay eager.
-    context.run_options->AddConfigEntry("gpu_graph_id", "-1");
-  } else if (graph_buffers_ != nullptr) {
-    // -1 tells the CUDA EP to run eagerly. Otherwise every distinct decode shape gets its own
-    // annotation id: the EP runs that id once eagerly, captures it on the next occurrence, and
-    // replays from then on.
-    const std::string graph_id =
-        capture ? std::to_string(VarlenGraphBuffers::GraphId(scheduled_requests.size(),
-                                                             context.block_table_columns))
-                : std::string("-1");
-    context.run_options->AddConfigEntry("gpu_graph_id", graph_id.c_str());
+      cache_manager_->SupportsDynamicBatching()) {
+    context.run_options->AddConfigEntry("gpu_graph_id", std::to_string(annotation_id).c_str());
   }
 
   decoder_state->DumpInputs();
--- a/src/engine/decoders/varlen_decoder_io.cpp
+++ b/src/engine/decoders/varlen_decoder_io.cpp
@@ -444,8 +444,6 @@ void VarlenDecoderIO::PreparePositionIds(
                  [](size_t sum, const std::shared_ptr<Request>& request) {
                    return sum + request->ScheduledTokenCount();
                  });
-  auto position_ids = std::make_unique<Tensor>(
-      model->p_device_inputs_, Ort::TypeToTensorType<int64_t>);
   const std::vector<int64_t> position_shape =
       position_planes_ == 1
           ? std::vector<int64_t>{static_cast<int64_t>(num_tokens)}
--- a/src/engine/decoders/varlen_decoder_io.cpp
+++ b/src/engine/decoders/varlen_decoder_io.cpp
@@ -76,9 +76,13 @@ AttentionMetadataValues GetAttentionMetadataForPlan(const StepPlan& plan) {
   return metadata;
 }
 
-AttentionMetadataValues GetAttentionMetadataForGraph(size_t block_table_columns, size_t block_size) {
+AttentionMetadataValues GetAttentionMetadataForGraph(size_t max_query_len, size_t block_table_columns,
+                                                     size_t block_size) {
   if (block_table_columns == 0 || block_size == 0) {
     throw std::runtime_error("Captured attention metadata requires non-zero block-table columns and block size.");
+  }
+  if (max_query_len == 0) {
+    throw std::runtime_error("Captured attention metadata requires a non-zero query length.");
   }
 
   size_t max_kv_len_lower_bound = 1;
@@ -97,7 +101,7 @@ AttentionMetadataValues GetAttentionMetadataForGraph(size_t block_table_columns,
   }
 
   return {
-      1,
+      CheckedMetadataLength(max_query_len, "Query length"),
       CheckedMetadataLength(block_table_columns, block_size, "KV length"),
       CheckedMetadataLength(max_kv_len_lower_bound, "KV lower bound"),
   };
@@ -105,8 +109,11 @@ AttentionMetadataValues GetAttentionMetadataForGraph(size_t block_table_columns,
 
 AttentionMetadataValues GetAttentionMetadataForGraphStep(
     const StepPlan& plan, size_t block_table_columns, size_t block_size) {
-  auto metadata = GetAttentionMetadataForGraph(block_table_columns, block_size);
   const auto exact_metadata = GetAttentionMetadataForPlan(plan);
+  // Every request in a capturable step carries the same token count, so the step's own query bound
+  // is exactly the bound for every step this graph will serve.
+  auto metadata = GetAttentionMetadataForGraph(
+      static_cast<size_t>(exact_metadata.max_query_len_bound), block_table_columns, block_size);
   if (exact_metadata.max_query_len_bound > metadata.max_query_len_bound ||
       exact_metadata.max_kv_len_bound > metadata.max_kv_len_bound) {
     throw std::runtime_error("Captured attention metadata upper bounds do not cover the current step.");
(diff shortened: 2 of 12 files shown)


src/engine/graph_annotation_ids.h
```diff
@@ -0,0 +1,74 @@
...
+      if (++shape.sightings < kSightingsBeforeCapture || assigned_ >= kMaxCapturedShapes) {
+        return -1;
+      }
+      shape.id = static_cast<int>(++assigned_);
+      return shape.id;
+    }
+    if (shapes_.size() >= kMaxTrackedShapes) {
+      // Shapes that never proved they recur are dropped so a steady shape arriving later can still
+      // be tracked. Captured shapes keep their ids.
+      std::erase_if(shapes_, [](const auto& entry) { return entry.second.id == 0; });
```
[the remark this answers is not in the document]
tianleiwu: Addressed in 7d6be6c731: the sightings/eviction implementation was removed, so this code no longer uses std::erase_if and remains C++17-compatible.

src/engine/graph_annotation_ids.h
```diff
@@ -0,0 +1,50 @@
...
+  // "do not capture or replay".
+  int Id(const Key& key) {
+    const auto existing = ids_.find(key);
+    if (existing != ids_.end()) {
+      return existing->second;
+    }
+    if (ids_.size() >= kMaxCapturedShapes) {
+      return -1;
+    }
+    return ids_.emplace(key, static_cast<int>(ids_.size() + 1)).first->second;
```
baijumeswani, line 41: These IDs start at 1 for every `SimpleDecoder`, but CUDA graphs are stored in the shared ORT session. The public API allows two Engines to use the same Model, so both decoders can assign ID 1 to different buffers. The second Engine may then replay the first Engine's graph and use the wrong device addresses. The same problem can happen after the first Engine is destroyed, because its graph is never released and its buffers are gone.

Please allocate IDs from a session-wide collision-free owner and release every ID when `SimpleDecoder` is destroyed, before its persistent buffers are freed. `State::~State()` already shows the `ReleaseCapturedGraph()` cleanup pattern. Please add tests with two live Engines sharing one Model and with destroy-then-recreate on the same Model.

src/engine/decoders/varlen_decoder_io.cpp
```diff
@@ -128,25 +135,45 @@ std::array<int32_t, kAttentionMetadataElementCount> PackAttentionMetadata(
   };
 }
 
-VarlenGraphBuffers::VarlenGraphBuffers(DecoderOnly_Model& model) {
+VarlenGraphBuffers::VarlenGraphBuffers(DecoderOnly_Model& model, size_t position_planes) {
   max_batch_size = model.config_->engine.dynamic_batching->max_batch_size;
+  // A speculative step submits the drafted tokens plus the token the target always advances, so the
+  // packed row count is a multiple of the batch rather than equal to it.
+  max_query_tokens = static_cast<size_t>(std::max(model.config_->speculative.max_draft_tokens, 0)) + 1;
```
baijumeswani, line 142: These persistent buffers are allocated after the paged cache has already chosen its block count, so their memory is outside the cache sizing calculation. The logits buffer can be hundreds of MiB for a large vocabulary and batch, and this PR reports about 1.1 GiB of extra peak memory for the tested model. Also, config allows 16 drafts while the Engine can verify at most 7, so part of this allocation can never be used.

Please clamp the capture width to the Engine's real maximum and account for these fixed buffers before automatic KV-cache sizing, either by allocating them before the free-memory probe or subtracting their bytes from the cache budget. Tests for the supported maximum, an oversized configured width, and a large batch/vocabulary would help protect this contract.

test/cpp/engine/varlen_decoder_io_tests.cpp
```diff
@@ -142,6 +142,101 @@ TEST(VarlenDecoderIOTest, GraphStepRejectsInsufficientUpperBounds) {
...
+  plan.requests = {first, second};
+
+  const auto metadata =
+      GetAttentionMetadataForGraphStep(plan, /*block_table_columns=*/16, /*block_size=*/128);
+
+  EXPECT_EQ(metadata.max_query_len_bound, 4);
+  EXPECT_EQ(metadata.max_kv_len_bound, 2048);
+}
+
+TEST(VarlenDecoderIOTest, GraphIdSeparatesEveryCapturedShape) {
```
baijumeswani, line 170: The key and metadata tests are useful, but they do not run CUDA graph capture or replay. They therefore cannot catch reused session IDs, stale buffer pointers, position-ID pointer mistakes, or a wrong fixed-state bank replay.

Please add CUDA-gated execution tests that compare eager and captured output for repeated decode, uniform multi-token verification, packed `[num_tokens]` and `[3, num_tokens]` position IDs, alternating fixed-state banks, and staged fixed-state bindings. The most important lifecycle cases are two Engines sharing one Model and destroying one Engine before creating another.

src/engine/decoders/simple_decoder.cpp
```diff
@@ -36,38 +36,60 @@ SimpleDecoder::SimpleDecoder(std::shared_ptr<DecoderOnly_Model> model,
...
 namespace {
 
-// A step can be captured only if it looks like every other step the graph will serve: one new token
-// per sequence. Prefill and chunked-prefill steps carry a variable number of tokens and are shaped
-// differently from run to run, so they always execute eagerly.
-bool IsPureDecodeStep(ScheduledRequests& scheduled_requests) {
-  if (scheduled_requests.size() == 0) {
-    return false;
-  }
+// A step can be captured only if it looks like every other step the graph will serve: the same
```
baijumeswani, line 46: This intentionally changes the public graph-capture contract, but `docs/paged_attention_engine.md` still says only one-token steps are capturable and that packed position IDs, fixed state, and draft verification stay eager.

Please update the document to describe uniform multi-token eligibility, persistent position buffers, the full graph key, fixed-state bank/layout handling, the 64-shape budget and eager fallback, graph lifetime, and the first-use stall and memory tradeoff.

src/engine/decoders/simple_decoder.cpp
```diff
@@ -36,38 +36,60 @@ SimpleDecoder::SimpleDecoder(std::shared_ptr<DecoderOnly_Model> model,
...
+                                        ? PlanDecodeTokenCount(*context.plan)
+                                        : UniformDecodeTokenCount(scheduled_requests);
+  // -1 tells the CUDA EP to run eagerly. Otherwise every distinct decode shape gets its own
+  // annotation id: the EP runs that id once eagerly, captures it on the next occurrence, and
+  // replays from then on. The id also covers the fixed-state binding layout, whose device
+  // addresses move when the persistent bank flips.
+  int annotation_id = -1;
+  if (graph_buffers_ != nullptr && tokens_per_request != 0 &&
+      graph_buffers_->Fits(scheduled_requests.size(), tokens_per_request)) {
+    annotation_id = graph_buffers_->GraphId(scheduled_requests.size(), tokens_per_request,
```
baijumeswani, line 88: Could we add lightweight diagnostics for this decision? A main reason for this change is that graph capture was silently disabled, but a shape that does not fit or arrives after the 64-ID budget is still silently sent to eager execution.

A debug log or counters for assigned IDs, fallback reasons, budget exhaustion, and graph-release failures would make production performance problems much easier to diagnose without adding work to the normal hot path.

baijumeswani (commented):
This PR enables CUDA graph capture for paged decode cases that were always eager before: uniform multi-token draft verification, packed text/MRoPE position IDs, and hybrid models with fixed decoder state. It adds persistent buffers, includes the per-request token width and fixed-state binding layout in the graph key, buckets block-table widths, and limits each decoder to 64 captured shapes.

The core design is thoughtful. Prefill and nonuniform batches still fall back to eager execution, the key covers the main shapes and fixed-state pointers that replay depends on, fixed-state bank changes get separate keys, and unsupported or excess shapes fail safely to eager mode. The reported H200 validation is also strong: identical token output, one graph launch per decode round, far fewer eager launches, and a meaningful throughput gain with the cold-start and memory costs clearly reported.

I found one correctness issue that should be fixed before approval: graph IDs are allocated per decoder even though ORT stores them in the shared session, and the graphs are not released when the decoder is destroyed. I also left recommendations for memory budgeting, automated replay coverage, documentation, and diagnostics.

Overall, this is a valuable improvement, but I recommend waiting for the graph-ID ownership fix before approval.

src/engine/graph_annotation_ids.h
tianleiwu: Fixed in c02bb321.

`GraphAnnotationIds` now draws from a process-wide `std::atomic<int64_t>` counter and never recycles an id, so no two allocators — concurrent, or one created after another is destroyed — can name the same graph. `AssignedIds()` exposes what was handed out, and `~SimpleDecoder()` calls `ReleaseCapturedGraph` on each of them before the persistent buffers are freed. It is guarded by `ORT_API_VERSION >= 27` and non-throwing, matching the `State::~State()` pattern you pointed at.

Verified on a CUDA build: three successive Engines over one Model were handed ids 1, 2 and 3, and each destroyed cleanly.

Unit coverage in `varlen_decoder_io_tests.cpp`: `GraphIdsNeverCollideAcrossLiveAllocators`, `GraphIdsAreNotReusedAfterAnAllocatorIsDestroyed`, and `GraphIdsReportEveryAssignedIdForRelease` (a missing id there would leave the session replaying against freed memory).

The end-to-end two-Engine and destroy-then-recreate cases are in the new `test/python/test_onnxruntime_genai_engine_cuda_graph.py`. See my reply on the test thread for why that file cannot run against the checked-in synthetic models.

src/engine/decoders/varlen_decoder_io.cpp
tianleiwu: Fixed in c02bb321.

**Width.** `SimpleDecoder` now sizes the buffers from `cache_manager_->MaxDraftTokensPerStep() + 1` instead of `config.speculative.max_draft_tokens`. That is the same ceiling the scheduler clamps every request to, so it already folds in `kMaxDraftTokensPerStep = 7`, the fixed-state checkpoint window, and the sliding-window query cap. A model whose logits carry one row per request cannot verify drafts at all, so its width is clamped to 1.

Worth noting the old expression was wrong in both directions. `config.speculative.max_draft_tokens` is a per-request *default*, not a bound: a request that raised it through `set_speculative_options`, or drafts attached through `Engine::max_draft_tokens_per_proposal()`, produced a step wider than the buffers, `Fits()` returned false, and capture was silently lost — the exact failure mode this PR exists to remove, just moved to a different trigger.

**Memory.** Added `VarlenGraphBufferBytes()`, and `Engine::CreateDependencies` subtracts it — target plus MTP head — from the reserved bytes before `CacheManager::Create` runs the free-memory probe. To stop the estimate drifting from what the constructor actually allocates, both now walk one shared buffer plan (a list of type/shape entries) rather than duplicating the sizing arithmetic.

One deliberate deviation from your wording: the deduction applies only when `engine.dynamic_batching.num_blocks` is absent, i.e. to the automatic sizing path you named. `ResolveConfiguredPagedBlockCount` also subtracts reserved bytes from an explicit `num_blocks`, and doing that here shrinks a pinned pool below its configured value — which breaks every checked-in synthetic model, whose KV tensors have a static block count. My reading is that pinning the pool means owning that budget. Say so if you would rather the pinned case cover these buffers too.

**Tests.** `GraphBufferBytesIgnoreAWidthTheModelCannotUse` (a model that cannot verify drafts must not grow when asked for a wider window), `GraphBufferBytesGrowWithTheVerifiedBlock`, and `GraphBuffersRejectStepsWiderThanTheyWereSizedFor` (the supported maximum on both axes, and one step past each).

test/cpp/engine/varlen_decoder_io_tests.cpp
tianleiwu: Agreed, and I wrote that file: `test/python/test_onnxruntime_genai_engine_cuda_graph.py` in c02bb321 covers repeated decode, batched decode, uniform multi-token verification, two Engines sharing one Model, and destroy-then-recreate, each as eager-versus-captured token parity against the same model directory with only `enable_cuda_graph` toggled.

It cannot run against the checked-in models, and I would rather flag that than hide it. All three synthetic paged models fault with an illegal memory access on the **first replay**:

```
[E:onnxruntime:, cuda_call.cc:148 CudaCall] CUDA failure 700: an illegal memory access was encountered ;
  file=.../core/providers/cuda/cuda_graph.cc ; line=85 ; expr=cudaStreamSynchronize(stream_)
```

I reproduced that at f0d2cd71 with every source change from this review round stashed, so it predates this work. The cause is the models themselves rather than the engine: their decoders are `Shape`/`Range`/`Slice`/`Gather`/`GatherND` subgraphs that ORT partitions across CPU and CUDA, so there is nothing capture-safe to record. It applies equally to `synthetic-paged`, `synthetic-paged-per-token`, and `synthetic-composite`.

So the file takes its model directory from `ORTGENAI_PAGED_CUDA_GRAPH_MODEL` and skips otherwise, matching the `GTEST_SKIP` convention `ModelTests.RecurrentStateCudaGraphCaptureMatchesEager` already uses for `tiny-qwen35-cuda`. It skips on this machine because the synthetic models are the only ones here; I have not yet run it green against a real export.

What I could verify locally on CUDA, in place of the execution tests:

- three successive Engines over one Model were handed annotation ids 1, 2, 3, and each released cleanly on destruction;
- the new `graph_capture` log confirms the intended decisions per step (prefill eager, then `tokens_per_request=1 block_table_columns=8 state_binding=0 -> graph id 1`);
- unit coverage for id collision across live allocators, no reuse after destruction, and full id enumeration for release.

Leaving this thread open on purpose: closing the gap needs a capture-safe synthetic paged decoder, which I think is better as its own change than bolted onto this PR. Tell me if you would rather it land here and I will do it.

src/engine/decoders/simple_decoder.cpp
tianleiwu: Done in c02bb321. Rewrote the "CUDA graph capture" section of `docs/paged_attention_engine.md` to cover:

- uniform multi-token eligibility, replacing the one-token rule;
- the full graph key as a table — batch size, tokens per request, block-table width bucket, fixed-state binding layout — with what each one changes in the recorded launches, plus the invariant that makes the width bucket an injective name;
- fixed-state bank and binding-set handling through `BindingLayoutKey()`;
- process-wide id allocation, why it cannot be per decoder, and release on decoder destruction;
- the 64-shape budget, the eager fallback, and the separate buffer-capacity fallback;
- the memory reservation against the automatic cache budget, and why an explicit `num_blocks` is left alone;
- persistent buffers and which tensors they cover, including packed `position_ids` in both geometries;
- the first-use capture stall and the fixed memory cost;
- the new `graph_capture` log category.

I also removed three stale claims elsewhere in the same document: the fixed-state pool section said fixed-state models execute eagerly, the mixed-batch section referred to a "pure-decode" graph path, and the block-drafter section said draft verification stays eager. The step-8 packed-inputs section no longer says packed `position_ids` force eager execution, and the step-9 section now derives the query bound from the step instead of hardcoding `1`.

src/engine/decoders/simple_decoder.cpp
tianleiwu: Added in c02bb321: a `graph_capture` log category, via `SetLogBool("graph_capture", true)` or `og.set_log_options(graph_capture=True)`.

Each line carries the step shape — batch, tokens per request, block-table columns, fixed-state binding key — and then either the assigned graph id or the reason it went eager: capture not enabled for this decoder, prefill or mixed token counts, step wider than the persistent buffers (printing the three limits it exceeded), shape budget exhausted, or no annotation key for the shape. Every line ends with the running `assigned/64` shape count, and release failures are logged from the destructor.

It stays off the hot path by construction: a line is emitted only when a shape is newly assigned an id, or when the fallback reason differs from the previous step. Steady state repeats neither, so a healthy run goes quiet after warmup, and with the flag off the cost is one bool test per step.

From a real run on this branch:

```
graph_capture   decode batch=1 tokens_per_request=0 block_table_columns=8 state_binding=0 -> eager (prefill, or requests contribute different token counts) [0/64 shapes assigned]
graph_capture   decode batch=1 tokens_per_request=1 block_table_columns=8 state_binding=0 -> graph id 1 [1/64 shapes assigned]
```

I stopped short of exported counters. `SpeculativeStats` is the wrong home for them, and a new stats struct means new C, Python, C# and Java surface. Happy to add one if you want these numbers queryable rather than logged — I would rather agree the shape of that API first than guess at it here.

test/cpp/engine/varlen_decoder_io_tests.cpp
baijumeswani: can be a separate pr.

src/engine/graph_annotation_ids.h
baijumeswani: non blocking: should we care about engine vs generator? an engine and generator can both exist in the same process.

baijumeswani (approved):

src/engine/graph_annotation_ids.h
tianleiwu: Good point. That's low priority for now. We can address it if needed.

PR #2557, opened 2026-09-13
Capture CUDA graphs for DFlash2 drafter proposals

tianleiwu:
## Description

Stacked on #2556 (paged decode capture).

Once the target decode step is captured, the DFlash2 drafter is what is left running eagerly: it pinned `gpu_graph_id` to `-1` on the grounds that its proposal tensors have stable addresses but unstable shapes. Measurements after the decoder change showed **374 eager launches per round surviving**, about 2.5 ms/step of GPU idle.

The shapes are unstable but they are enumerable. They vary with how many context rows the drafter ingests (which tracks the accepted-token count), the block-table width, and whether a reused proposal tensor has outgrown its buffer and moved.

## Summary of Changes

- Key each proposal graph on the packed row counts, the drafting request count, the block-table width bucket, the max query length, and a **buffer generation counter**. `Dflash2StepTensor` now reports when it replaces a buffer, and the counter increments on every move so a graph captured against the old address is never replayed after it.
- On captured steps, bucket the block-table width to a power of two so a growing context keeps reusing one graph instead of retiring one per block boundary. Columns past the live KV length are never read.
- On captured steps, feed `attention_metadata` **bucket-wide bounds** instead of exact per-step values. A captured step bakes those bounds into its launch dimensions and never re-runs the host code that read them; the kernel re-reads exact per-sequence lengths from device memory, which is what keeps replay correct as sequences grow.
- **Capture only when every served request ingests the same number of context rows.** Only the total reaches the captured shape, so a step whose requests ingest differing counts shares a shape with every other split that sums the same way. Under four-way concurrency those sums walk over a range far wider than the budget.
- Capture only when the drafter session actually enables CUDA graphs, so sessions without graphs keep byte-identical inputs and behaviour.

### Why the uniform-ingest gate is load-bearing

The first version of this change keyed on the summed context-row count. A full-process nsys trace of a four-way concurrent workload showed **47 drafter graphs for 92 replays — 36 of them replayed exactly once**. Each of those captures costs a stalled step (p95 102 ms against a 37 ms median) and buys one replay, and the result was a measured regression rather than a gain:

| case | decode capture only | + drafter, summed key | + drafter, uniform gate |
|---|---|---|---|
| ctx512 b1 k7 | 1.131x | 1.129x | **1.165x** |
| ctx2048 | 1.120x | 1.112x | **1.151x** |
| concurrent4 | 1.056x | **0.848x** | **1.058x** |

Ratios are graph-on over the eager control in the same build. Restricting capture to uniform steps keeps the shape ladder proportional to the per-request row count instead of the sum over requests, which removes the concurrency regression and leaves a real incremental gain over decode capture alone (+3.4 pp at ctx512, +3.2 pp at ctx2048).

## Testing

`lintrunner` clean; **642/642** engine unit tests pass, including coverage that `Dflash2StepTensor` reports buffer moves and not reuse.

End-to-end on H200, A/B toggling only `enable_cuda_graph`:

- **Correctness:** tokens **bit-identical** across arms on 10 cases × 2 inner × 2 reps, including a 1024-token generation, two paged-block-boundary cases, and 2- and 4-way concurrency. Round structure identical, generation deterministic within each arm, and acceptance matches **exactly** — the signal that would have moved first had the wider block tables or looser bounds been wrong. A dedicated buffer-growth stress escalating ingest width 512→4096 and revisiting earlier widths over 3 cycles hashes identically in both arms with no illegal-address fault.
- **Capture:** graph launches/round 1.00 → 2.03; 313 of the 374 previously-eager launches (83.7%) move inside a graph, residual 60.9.
- **Performance (steady state, 5 reps × 3 inner, arm ranges non-overlapping):**

| case | eager | graphs | ratio | step p50 |
|---|---|---|---|---|
| ctx512 b1 k7 | 199.10 tok/s | **232.00 tok/s** | 1.165x | 20.05 → 17.25 ms |
| ctx2048 | 148.21 tok/s | **170.65 tok/s** | 1.151x | 21.27 → 18.52 ms |
| concurrent4 | 442.24 tok/s | **468.01 tok/s** | 1.058x | 38.20 → 36.60 ms |

- **Memory:** peak 41012 → 42169 MiB (+2.8%) for both stages combined.

## Known Limitations

- Shape-diverse cold start still pays the capture cost: over the first encounter of a workload the same three cases measure 0.99x, 0.94x and 0.72x. The drafter's shape still tracks the accepted-token count, so removing that from the key (by padding ingest to a fixed width) is the remaining lever and is not attempted here.
- Generation-counter coverage was exercised only for the buffers the stress test moved.
- Single model, single GPU, one draft width, greedy only.

### DIFF
--- a/src/dflash2_drafter.cpp
+++ b/src/dflash2_drafter.cpp
@@ -666,6 +666,18 @@ bool Dflash2Drafter::Propose(Tensor& aux_hidden_states, std::span<const Feed> fe
     ingest_count[i] = feed.aux_row_count - dropped;
     num_ctx_rows = CheckedAdd(num_ctx_rows, ingest_count[i], "DFlash 2 context rows");
   }
+  // Only the total row count reaches the captured shape, so a step whose requests ingest differing
+  // counts shares a shape with every other split that sums the same way. Under concurrency those
+  // sums walk over a range far wider than the graph budget and almost none of them recur, so each
+  // capture costs more than it ever returns. Restricting capture to uniform steps keeps the shape
+  // ladder proportional to the per-request row count instead of the sum over requests.
+  bool uniform_ingest = true;
+  for (const size_t i : served) {
+    if (ingest_count[i] != ingest_count[served.front()]) {
+      uniform_ingest = false;
+      break;
+    }
+  }
   CheckedMetadataValue(
       CheckedAdd(num_ctx_rows, num_block_rows, "DFlash 2 packed rows"),
       "DFlash 2 packed rows");
@@ -730,9 +742,10 @@ bool Dflash2Drafter::Propose(Tensor& aux_hidden_states, std::span<const Feed> fe
   // Bucketing the width lets a growing context keep replaying one captured graph instead of
   // retiring one at every block boundary; columns past the live KV length are never read. Steps
   // that cannot be captured keep the exact width so their bound inputs are unchanged.
+  const bool graph_eligible = graph_capture_enabled_ && uniform_ingest;
   const size_t block_table_columns =
-      graph_capture_enabled_ ? GetGraphBlockTableColumns(max_blocks, max_block_table_columns_)
-                             : max_blocks;
+      graph_eligible ? GetGraphBlockTableColumns(max_blocks, max_block_table_columns_)
+                     : max_blocks;
   // Any proposal tensor that outgrows its buffer moves, which retires every graph captured against
   // its old addresses.
   bool buffers_moved = false;
@@ -839,7 +852,7 @@ bool Dflash2Drafter::Propose(Tensor& aux_hidden_states, std::span<const Feed> fe
   // Everything the captured launches bake in: the packed row counts, the block-table width, and the
   // generation of the buffers those launches recorded addresses for.
   const int annotation_id =
-      graph_capture_enabled_
+      graph_eligible
           ? graph_ids_.Id(GraphAnnotationIds::Key{
                 served.size(), batch, num_ctx_rows, block_table_columns,
                 static_cast<size_t>(layout.max_query_len), buffer_generation_})
--- a/src/dflash2_drafter.cpp
+++ b/src/dflash2_drafter.cpp
@@ -4,6 +4,8 @@
 #include "generator/generators.h"
 #include "dflash2_drafter.h"
 #include "engine/request.h"
+#include "engine/decoders/varlen_decoder_io.h"
+#include "engine/paged_key_value_cache.h"
 #include "models/io/kv_cache.h"
 
 #include <algorithm>
@@ -425,6 +427,13 @@ Dflash2Drafter::Dflash2Drafter(std::shared_ptr<Dflash2Model> model, size_t paged
     throw std::runtime_error("The DFlash 2 drafter cache exceeds the int32 block-id range.");
   }
   query_spill_blocks_ = (static_cast<size_t>(config_.block_size) - 1) / paged_block_size_ + 1;
+  // Widths are bucketed up to the longest sequence the model can hold, so the bucket ladder is
+  // logarithmic in the context length rather than one width per block boundary.
+  const size_t context_length =
+      static_cast<size_t>(std::max(model_->config_->model.context_length, 1));
+  max_block_table_columns_ = (context_length - 1) / paged_block_size_ + 1;
+  graph_capture_enabled_ =
+      IsGraphCaptureEnabled(model_->config_->model.decoder.session_options);
   if (config_.sliding_window > 0) {
     // Positions older than this are masked out of every query row, so they are never ingested and
     // the ring may alias them.
@@ -457,12 +466,13 @@ Dflash2Drafter::Dflash2Drafter(std::shared_ptr<Dflash2Model> model, size_t paged
       }
     }
   }
-  // Proposal tensors keep a stable address but not a stable shape, so they are not capturable.
-  run_options_->AddConfigEntry("gpu_graph_id", "-1");
+  // The annotation id is chosen per step: proposal tensors keep stable addresses between growth
+  // events, but their shapes track the accepted-token count and the context length.
 }
 
 Tensor& Dflash2StepTensor(std::unique_ptr<Tensor>& slot, DeviceInterface* device,
-                          ONNXTensorElementDataType type, const std::vector<int64_t>& shape) {
+                          ONNXTensorElementDataType type, const std::vector<int64_t>& shape,
+                          bool* reallocated) {
   size_t elements = 1;
   for (const int64_t dimension : shape) {
     elements = CheckedMultiply(elements, static_cast<size_t>(dimension), "DFlash 2 proposal tensor");
@@ -481,6 +491,9 @@ Tensor& Dflash2StepTensor(std::unique_ptr<Tensor>& slot, DeviceInterface* device
       capacity = std::max(bytes, previous_bytes * 2);
     }
     slot = std::make_unique<Tensor>(device, type);
+    if (reallocated) {
+      *reallocated = true;
+    }
   }
   slot->CreateTensor(shape, /*make_static=*/true, std::max<size_t>(capacity, 1));
   return *slot;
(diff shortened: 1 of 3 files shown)


baijumeswani (commented):
Thanks for pushing this — the overall direction here looks good. This PR teaches the DFlash2 drafter to participate in CUDA graph capture by giving proposal steps stable graph identities, bucketing block-table widths so growing contexts can reuse captures, and narrowing capture to uniform-ingest steps to avoid the summed-shape explosion under concurrency. That all makes sense, and the `Dflash2StepTensor(..., reallocated)` plumbing plus the uniform-ingest gate are both solid building blocks.

I do think a few lifecycle and edge-case details still need tightening before this is ready. I left inline comments with concrete suggestions below.

src/dflash2_drafter.h
```diff
@@ -186,6 +189,14 @@ struct Dflash2Drafter {
     std::unique_ptr<Tensor> candidate_ids;
     std::unique_ptr<Tensor> scores;
   } step_tensors_;
+  // Block-table widths are bucketed to powers of two up to this cap so that a growing context
+  // reuses a captured graph instead of retiring one per block boundary.
+  size_t max_block_table_columns_{};
+  bool graph_capture_enabled_{};
+  GraphAnnotationIds graph_ids_;
```
baijumeswani, line 204: Now that the drafter owns captured graph IDs, I think it also needs a destructor that releases them, similar to `SimpleDecoder::~SimpleDecoder()`. Right now the proposal/cache tensors are freed before the session goes away, but the session can still be holding captured graphs that were recorded against those addresses. I’d suggest adding a best-effort non-throwing cleanup path that iterates `graph_ids_.AssignedIds()` and calls `ReleaseCapturedGraph()` while the session is still alive.

src/dflash2_drafter.cpp
```diff
@@ -756,63 +792,91 @@ bool Dflash2Drafter::Propose(Tensor& aux_hidden_states, std::span<const Feed> fe
...
+                                   &buffers_moved);
+
+  if (buffers_moved) {
+    ++buffer_generation_;
+  }
+  // Everything the captured launches bake in: the packed row counts, the block-table width, and the
+  // generation of the buffers those launches recorded addresses for.
+  const int annotation_id =
+      graph_eligible
+          ? graph_ids_.Id(GraphAnnotationIds::Key{
```
baijumeswani, line 899: Using `buffer_generation_` in the key avoids replaying graphs from an old buffer epoch, which is great. One thing still missing is retirement of the old IDs: once `buffer_generation_` bumps, every older key becomes permanently unreachable, but those graphs still consume session memory and count against the 64-shape budget. Could we release the old generation’s captured graphs and reset/clear the ID table when the generation changes? Otherwise a long-lived service can burn through the budget with dead graphs and fall back to eager earlier than expected.

src/dflash2_drafter.cpp
```diff
@@ -756,63 +792,91 @@ bool Dflash2Drafter::Propose(Tensor& aux_hidden_states, std::span<const Feed> fe
...
+  if (graph_capture_enabled_) {
+    run_options_->AddConfigEntry("gpu_graph_id", std::to_string(annotation_id).c_str());
+  }
+
+  {
+    // A captured step bakes these bounds into its launch dimensions and never re-runs the host code
+    // that read them, so they must bound the whole bucket rather than this step. The kernel re-reads
+    // the exact per-sequence lengths from device memory, which is what keeps replay correct as the
+    // sequences grow.
+    const AttentionMetadataValues values =
```
baijumeswani, line 915: I think this path needs the same lower-bound safety guard that `GetAttentionMetadataForGraphStep()` has on the main decoder side. `uniform_ingest` makes the newly ingested rows uniform, but it doesn’t make total KV lengths uniform across requests. That means the bucket-derived `max_kv_len_lower_bound` can still be too high for the shortest request in a mixed-length batch. Could we clamp the lower bound back to `1` when `layout.min_kv_len` is smaller than the bucket-derived value, or reuse a helper that already applies that check?

src/dflash2_drafter.cpp
```diff
@@ -756,63 +792,91 @@ bool Dflash2Drafter::Propose(Tensor& aux_hidden_states, std::span<const Feed> fe
...
-                                           static_cast<int64_t>(top_k)});
+                                           static_cast<int64_t>(top_k)},
+                                          &buffers_moved);
   auto& scores = Dflash2StepTensor(step_tensors_.scores, device, Ort::TypeToTensorType<float>,
                                    {static_cast<int64_t>(batch), static_cast<int64_t>(num_spec),
-                                    static_cast<int64_t>(top_k), static_cast<int64_t>(top_k)});
+                                    static_cast<int64_t>(top_k), static_cast<int64_t>(top_k)},
+                                   &buffers_moved);
+
+  if (buffers_moved) {
```
baijumeswani: This generation bump looks slightly non-exception-safe to me. A proposal tensor can move earlier in the step, freeing the old captured address, and then a later allocation/copy can throw before we reach this increment. On the next smaller step we could still look up the old graph ID even though one of its recorded addresses is gone. I’d consider either (1) staging replacements in temporaries and swapping only after the whole step is safe, or (2) advancing the generation immediately when a successful buffer move invalidates an old captured address.

src/dflash2_drafter.cpp
```diff
@@ -425,6 +427,13 @@ Dflash2Drafter::Dflash2Drafter(std::shared_ptr<Dflash2Model> model, size_t paged
     throw std::runtime_error("The DFlash 2 drafter cache exceeds the int32 block-id range.");
   }
   query_spill_blocks_ = (static_cast<size_t>(config_.block_size) - 1) / paged_block_size_ + 1;
+  // Widths are bucketed up to the longest sequence the model can hold, so the bucket ladder is
+  // logarithmic in the context length rather than one width per block boundary.
+  const size_t context_length =
+      static_cast<size_t>(std::max(model_->config_->model.context_length, 1));
```
baijumeswani, line 445: I’m a little worried that this cap is based only on `context_length`, while the drafter still materializes a full query block even near the session limit. Close to the end of a request, `total_positions` can spill into one more logical block than `ceil(context_length / paged_block_size)`, and the captured block table would then be capped/truncated instead of falling back to eager. Would it be safer to include worst-case query spill in this cap, or explicitly disable capture whenever `max_blocks` grows past the capturable width?

test/cpp/dflash2_config_test.cpp
```diff
@@ -742,6 +742,29 @@ TEST(Dflash2ConfigTest, ReusesProposalBufferUntilAStepOutgrowsIt) {
   EXPECT_EQ(slot->buffer_, grown);
 }
 
+TEST(Dflash2ConfigTest, ReportsWhenAProposalBufferMoves) {
```
baijumeswani, line 756: Nice start on the buffer-move coverage. I think the risky cases here still need a few more focused tests: replay after a buffer-growth epoch, a failure after one tensor moves but before the generation bump, a mixed-KV-length captured batch (to validate the lower-bound metadata), context-boundary spill near `MaxSessionTokens()`, and cleanup/retirement of captured graphs. Those are the cases most likely to hide a correctness regression here.

src/dflash2_drafter.h
tianleiwu: Addressed in 7ec51ded. Dflash2Drafter now releases every assigned graph ID from a non-throwing destructor while both the session and proposal tensors are still alive. The ID iteration path is allocation-free, and individual release failures are logged without escaping the destructor.

src/dflash2_drafter.cpp
tianleiwu: Addressed in 7ec51ded. A successful proposal-buffer move now immediately releases the active graph epoch and clears its ID table before the displaced buffer is destroyed. This restores the per-owner shape budget while process-wide annotation IDs remain non-recycled.

src/dflash2_drafter.cpp
tianleiwu: Addressed in 7ec51ded. The drafter now routes captured metadata through GetAttentionMetadataForGraphStep() with its exact minimum KV length in the lower-bound slot, so a mixed-length batch clamps the bucket lower bound to 1 when needed. Added a focused mixed-KV regression test.

src/dflash2_drafter.cpp
tianleiwu: Addressed in 7ec51ded. Replacement tensors are now fully constructed before swapping into the live slot, so a construction failure preserves the old captured address. After a successful swap, the displaced buffer remains alive until its graph IDs are released and the generation advances. Added failure-path and displaced-buffer tests.

src/dflash2_drafter.cpp
tianleiwu: Addressed in 7ec51ded. The capture cap now includes the worst-case query-block spill beyond the configured context, and capture eligibility also requires max_blocks to fit that cap. Any wider step therefore keeps its exact block-table width and runs eagerly instead of truncating. Added boundary-spill coverage.

test/cpp/dflash2_config_test.cpp
tianleiwu: Added focused coverage in 7ec51ded for displaced-buffer lifetime, failed replacement rollback, mixed-KV lower-bound fallback, context-boundary query spill, and graph-ID iteration/retirement with budget restoration. The focused suite passes 80 tests, and the complete engine_unit_tests binary passes all 653 tests.

baijumeswani (approved):

PR #2559, opened 2026-09-13
Let Model Builder set the speculative draft width

tianleiwu:
## Description

`speculative.max_draft_tokens` decides how many drafted tokens the engine verifies each step, and it defaults to 4. Nothing in the export path could set it, so every speculative model built with Model Builder shipped with a proposal width that no one had chosen or measured.

`dflash2_num_draft_tokens` looks like the knob for this but is not. It fixes the drafter's exported block (`block_size = n + 1`), so changing it changes the drafter itself and requires a re-export. This PR adds a `max_draft_tokens` extra option that writes the `speculative` section of `genai_config.json` instead. It is drafter-agnostic, costs nothing to retune on an already-exported model, and is free to raise because the drafter's block runs at the same price no matter how many of its tokens the engine verifies.

### Model Builder

| File | Change |
|------|--------|
| `src/python/py/models/builders/base.py` | `make_genai_config` emits `genai_config["speculative"]` when `max_draft_tokens` is set. Range-checked to 1–16 to match `Speculative_Element::kMinK`/`kMaxK` in `src/config.cpp`, so a bad value fails at export instead of at load. |

### Documentation

| File | Change |
|------|--------|
| `src/python/py/models/builder.py` | Extra-option help text, including how it differs from `dflash2_num_draft_tokens`. |
| `src/python/py/models/README.md` | Paragraph in the DFlash 2 section explaining why raising the width is nearly free and why the best value must be measured. |

### Tests

| File | Change |
|------|--------|
| `test/python/builder/test_max_draft_tokens.py` | New. Covers the default (no `speculative` section), accepted values, string input as Olive and the command line pass it, and out-of-range rejection. |

## Testing

- `pytest test/python/builder/` — 708 passed, 3 skipped.
- `lintrunner -a` — clean.
- End-to-end on one H200 with Qwen3.8-27B and an INT4 DFlash 2 drafter, driven through the `Engine`/`Request` API. The exported `speculative` section was verified to take effect from disk, with no runtime overlay: the INT4 package proposed 6.98 tokens per round at 7 and the INT8 package proposed 6.00 at 6.

  Decode throughput (tokens/s, six samples per point, min/max ranges non-overlapping around each winner), for a 512-token single-stream prompt and for four concurrent streams:

  | KV cache | k=4 (default) | k=5 | k=6 | k=7 |
  |---|---|---|---|---|
  | INT8, single stream | 210.8 | 233.8 | **241.7** | 231.4 |
  | INT8, four streams | 467.5 | 464.7 | **479.6** | 468.6 |
  | INT4, single stream | 167.5 | 167.9 | 178.3 | **187.5** |
  | INT4, four streams | 409.3 | 418.4 | **421.6** | 417.9 |

  Moving from the default to the measured optimum is worth 15% single-stream decode throughput with an INT8 KV cache and 12% with an INT4 one, with no change to the exported model.

- The generated token sequences were identical across every width above, which is expected: speculative decoding verifies each drafted token against the target, so the width changes how tokens are produced, not which.

## Motivation and Context

The optimum is workload-specific, which is why this is an option rather than a new default. KV-cache precision alone moves it: the INT8 cache peaks at 6 while the INT4 cache peaks at 7. The lossier cache accepts fewer drafted tokens per round at the same width (2.34 vs 2.48 accepted at k=4) and has a more expensive verification step, so it has to propose wider to break even. A single baked-in value would be wrong for one of them.

Behavior is unchanged when the option is unset: no `speculative` section is written and the runtime default still applies.

## Checklist

- [x] Tests added/updated
- [x] No breaking changes
- [x] Documentation updated


src/python/py/models/builders/base.py
```diff
@@ -1378,6 +1378,15 @@ def make_genai_config(self, config, extra_kwargs, out_dir):
                 dynamic_batching["max_scheduled_tokens"] = int(self.extra_options["max_scheduled_tokens"])
             genai_config["engine"] = {"dynamic_batching": dynamic_batching}
 
+        if "max_draft_tokens" in self.extra_options:
+            # Caps how many drafted tokens the engine verifies per step. This is independent of
+            # the drafter's exported geometry, which costs the same no matter how many of its
+            # tokens are used, so the best value is workload-specific and must be measured.
+            max_draft_tokens = int(self.extra_options["max_draft_tokens"])
+            if not 1 <= max_draft_tokens <= 16:
+                raise ValueError(f"max_draft_tokens must be between 1 and 16. Got: {max_draft_tokens}.")
```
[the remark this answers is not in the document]
tianleiwu: Fixed in 93d1d41d. The value is now parsed from its text form, so `6.5` (and `"6.5"`, `"1e1"`, `""`) raises instead of truncating to 6, and `test_genai_config_rejects_non_integer_max_draft_tokens` covers it.

src/python/py/models/README.md
```diff
@@ -325,6 +325,8 @@ Set `dflash2_path` to a DFlash 2 checkpoint to export an auxiliary `dflash2.onnx
 
 `dflash2_num_draft_tokens` optionally overrides how many tokens the drafter proposes per step. It must be a positive integer no greater than the draft checkpoint's block size minus the anchor token; that checkpoint limit is also the default.
 
+`max_draft_tokens` writes `speculative.max_draft_tokens` into `genai_config.json`, capping how many drafted tokens the engine verifies each step. It must be between 1 and 16, and defaults to unset, which leaves the runtime default of 4 in effect. This differs from `dflash2_num_draft_tokens`: the drafter's exported block costs the same to run no matter how many of its tokens are verified, so raising this value buys extra accepted tokens for free until the wider verification step costs more than it saves. The best value is workload-specific and must be measured; it can be retuned on an already-exported model by editing the config.
```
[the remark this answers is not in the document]
tianleiwu: Fixed in 93d1d41d. `max_draft_tokens` is now in all three invocation examples in that section.

tianleiwu:
@copilot resolve the merge conflicts in this pull request

baijumeswani (commented):
I found three issues that should be addressed. I added details inline.

src/config.cpp
```diff
@@ -2537,6 +2542,47 @@ void ValidateModelPaths(const Config& config) {
...
+// that at load time, where the value can still be edited.
+void WarnOnClampedDraftWidth(const Config& config) {
+  if (!g_log.enabled || !g_log.warning) {
+    return;
+  }
+  // DFlash 2 and its DSpark alias are the only drafters whose ceiling is known from the config
+  // alone. MTP's depends on an ONNX output that no session has loaded yet, and MtpGenerator runs
+  // outside the engine, so neither of the bounds below applies to it.
+  const auto& dflash2 = config.model.dflash2;
+  if (dflash2.filename.empty()) {
```
baijumeswani: This returns for every MTP-only config, but Engine-hosted MTP is also clamped by `MaxDraftTokensPerStep()`. That includes the state-update capacity and the 7-token Engine limit. For example, an MTP config can request 16 and silently run with 3, but this code will not warn. Please keep the DFlash geometry check specific to DFlash, and move the common Engine limit warning to Engine setup, where we know which path is being used. Please also add tests for Engine-hosted MTP.

src/python/py/models/builders/base.py
```diff
@@ -1387,6 +1387,23 @@ def make_genai_config(self, config, extra_kwargs, out_dir):
                 dynamic_batching["max_scheduled_tokens"] = int(self.extra_options["max_scheduled_tokens"])
             genai_config["engine"] = {"dynamic_batching": dynamic_batching}
 
+        if "max_draft_tokens" in self.extra_options:
```
baijumeswani, line 1390: Please validate and normalize this option in `check_extra_options` instead. The normal builder flow creates and saves the ONNX model before `make_genai_config` runs, so a bad value such as `max_draft_tokens=17` fails only after a potentially long export and can leave partial output. `make_genai_config` should only consume the already validated integer.

src/config.cpp
```diff
@@ -2537,6 +2542,47 @@ void ValidateModelPaths(const Config& config) {
...
+    return;
+  }
+
+  const int requested = config.speculative.max_draft_tokens;
+  const std::string alias = dflash2.is_dspark ? "dspark" : "dflash2";
+  const auto warn = [&](const std::string& bound_name, int bound) {
+    try {
+      Log("warning", "speculative.max_draft_tokens is " + std::to_string(requested) + " but " +
+                         bound_name + " is " + std::to_string(bound) +
+                         ", so each step will draft only " + std::to_string(bound) + " tokens.");
```
baijumeswani: This message can report the wrong effective width because one warning is emitted for each limit. For example, limits of 6, 3, and 7 produce three messages saying the step will draft 6, 3, and 7 tokens, while the real width is 3. Please emit one warning using the smallest limit, or say that each named limit caps the width without calling it the final width.

src/config.cpp
tianleiwu: Fixed in ea17b106.

`WarnOnClampedDraftWidth(const Config&)` is now DFlash-only and checks just `model.dflash2.num_draft_tokens`, the one bound knowable from the config alone. The hosting bounds moved to `Engine::WarnOnClampedDraftWidth()`, which runs at construction for any hosted speculative path (DFlash 2 **or** MTP) and compares the request against `MaxDraftTokensPerStep()` — so a 16-token MTP config on a 3-token engine now warns.

Added `EngineRunTest.EngineHostedMtpWarnsWhenTheConfiguredDraftWidthIsClamped` and `EngineRunTest.EngineHostedMtpClampsTheConfiguredDraftWidth` for the Engine-hosted MTP path.

src/config.cpp
tianleiwu: Fixed in ea17b106.

There is now at most one warning per source. The Engine-side warning uses the single smallest limit: `MaxDraftTokensPerStep()` already folds the state-update capacity, the paged query limit, and `kMaxDraftTokensPerStep` together, and the DFlash 2 drafter geometry is min-ed in on top, so the number it prints is the real width. The config-time DFlash warning no longer claims a final width — it says the geometry "caps each step to at most N drafted tokens. The engine may cap it further."

`EngineHostedMtpWarnsWhenTheConfiguredDraftWidthIsClamped` asserts the 16-vs-3 case logs `each step will draft only 3 tokens` and never mentions 16 as a width.

src/python/py/models/builders/base.py
tianleiwu: Fixed in ea17b106.

Validation and normalization moved to `check_extra_options` in `builder.py`, next to the `state_update_capacity` check, so `max_draft_tokens=17` now fails during option parsing before any export work. `make_genai_config` just writes `self.extra_options["max_draft_tokens"]` as-is.

The tests moved with it: `test_check_extra_options_rejects_out_of_range_max_draft_tokens` / `..._non_integer_...` exercise the new location (including the `6.5` / `"1e1"` cases), and `test_check_extra_options_normalizes_max_draft_tokens_to_an_int` covers the string form Olive and the CLI pass through.

baijumeswani (approved):

PR #2566, opened 2026-09-15
Honor paged_chunk_size for every paged model

tianleiwu:
### Description

`paged_chunk_size` is validated like any other paged option, but it was only written into `genai_config.json` when `has_windowed_paged_layers()` was true, i.e. when the model's sliding-window layers are served from a ring of blocks. For every other paged model the option was accepted and then silently dropped, so `search.chunk_size` never appeared and the value had no effect.

This writes it whenever it is passed for a paged model, and corrects the option's documentation.

### Motivation and Context

`search.chunk_size` and `engine.dynamic_batching.max_scheduled_tokens` are easy to confuse, and the old wording made it worse by describing the option purely as a ring requirement. They are different limits:

* `max_scheduled_tokens` is the token budget for the **whole step**, shared by every request in it.
* `search.chunk_size` caps what a **single request** contributes to that step (`ScheduledTokenCount` in `sequence_positions.h`, applied as `prefill_token_cap` in `decode_first_scheduler_policy.cpp`).

Without the per-request cap, one prefill takes as much of the step budget as it can and the other requests wait, so concurrent prefills run close to serially. Measured on Qwen3.8-27B (H200, `max_scheduled_tokens = 1024`, `paged_block_size = 256`), 8 concurrent requests, by reading back the chunk the scheduler actually used:

| `chunk_size` | 8 x 2048 tokens | 8 x 8192 tokens |
| --- | --- | --- |
| 256 | 5.08 s, 100.8 tok/s | 16.43 s, 31.2 tok/s |
| 512 | 5.82 s, 87.9 tok/s | 18.48 s, 27.7 tok/s |
| 1024 | 6.41 s, 79.8 tok/s | 22.04 s, 23.2 tok/s |
| unset | 6.38 s, 80.3 tok/s | 22.06 s, 23.2 tok/s |

256 is 25-34% faster than leaving it unset, and cuts worst-case time-to-first-token by about 25%.

Two details worth noting, both of which the new documentation states:

* A value at or above `max_scheduled_tokens` does nothing. `1024` and unset produce byte-identical output and the same effective chunk, because the shared budget binds first.
* A single request sees the opposite trade. One 32k prompt reaches its first token in 10.7 s with the option unset and 13.7 s at 256, since there is nobody to interleave with.

### Testing

`test/python/builder/test_paged_chunk_size.py` covers a paged model with no ring at several values, the string form that Olive and the command line pass, and the non-paged case. The three cases that exercise the fix fail without it.

Full builder suite: 706 passed, 3 skipped.

Verified end to end by building Qwen3.8-27B with `paged_chunk_size=256` and confirming `"search": {"chunk_size": 256}` in the exported `genai_config.json`, then measuring the numbers above through the engine.


baijumeswani (dismissed):

tianleiwu:
Addressed the suppressed nit in 6cc28554: `paged_chunk_size` is now described as applying only when `use_paged_attention` is set and being ignored otherwise, instead of "Requires", which matches `check_extra_options` and the test that expects it to be ignored.

baijumeswani (approved):

PR #2568, opened 2026-09-15
Accept any power-of-two paged_block_size of at least 16

tianleiwu:
### Description

The builder required `paged_block_size` to be a multiple of 256, documented as "required by the ONNX Runtime PagedAttention CUDA kernel". That is not the op's contract, and the rule is wrong in both directions.

`CheckInputs` in [`paged_attention_helper.h`](https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/contrib_ops/cpu/bert/paged_attention_helper.h) requires a **power of two of at least 16**. The multiple-of-256 figure comes from one FlashAttention tile size; `paged_attention.cc` treats that as a backend-selection input, not a validity check, and falls back to PagedDecode or MemoryEfficientAttention when the page is smaller. A quantized KV cache is exempt from the alignment check alone (`flash_block_size_ok = kIsQuantizedCache || block_size % tile == 0`); it still reaches FlashAttention, through a dense dequantized path that has no page to align.

So the old rule:

* **rejected valid values.** 16, 32, 64 and 128 all build and run.
* **accepted an invalid one.** 768 passes `% 256 == 0`, exports without complaint, and then fails on the first forward pass: `block_size must be a power of two and at least 16. Got block_size == 768`. (`CheckInputs` runs from `ComputeInternal`, not at load.)

This replaces it with the op's actual contract and corrects the documentation.

### Motivation and Context

Smaller pages cut the internal fragmentation of the KV pool, which matters most for short sequences and high batch counts, and a quantized cache pays nothing for them. They were simply unavailable.

### Testing

New `test/python/builder/test_paged_block_size.py`: 16 through 2048, the string form Olive and the CLI pass, sub-16, non-powers-of-two (including 768), and non-positive values. 17 cases.

`test_precision.py` asserted that 128 was rejected; it now asserts that 768 is, which is the case the old rule got wrong. Full builder suite: 717 passed, 3 skipped.

Measured end to end on Qwen3.8-27B (H200, bf16, unquantized KV, `head_size` 256, DFlash2 drafter), 200 MMLU-Pro questions at concurrency 8, with `num_blocks` scaled to hold the KV pool at 262144 tokens:

| `paged_block_size` | tok/s | drafts/round | output vs 256 |
| --- | --- | --- | --- |
| 64 | 161.0 | 0.00 | 84/200 identical |
| 128 | 374.6 | 7.00 | 200/200 identical |
| 256 (default) | 381.9 | 7.00 | reference |
| 512 | 369.8 | 7.00 | 200/200 identical |

128, 256 and 512 are byte-identical and their throughput spread is inside this harness's noise, so the default stays the right choice for this model.

### A note on the 64 column, and why it is not in the code

The 64 row is not a block-size effect. On today's `main`, PagedAttention serves `is_causal=0` from FlashAttention only, and the DFlash2 drafter has `head_size` 128, so a 64-token page makes the drafter's node fail and the engine falls back to target-only decoding for the rest of the run:

```
PagedAttention: is_causal=0 requires the FlashAttention backend
(sm>=80, fp16/bf16, head_size 128, block_size 64).
```

microsoft/onnxruntime#32619 removes exactly that restriction: it gives `is_causal=0` non-causal MEA, paged decode, XQA and latent backends, and adds head-size-128 regression cases at 16/32/64-token pages. Once it lands, a small page stops being fatal for a block drafter and becomes what it already is for the target, a backend-selection input.

Because of that, this PR deliberately keeps the runtime detail **out** of the source. The comments in `builder.py`, `block_drafter.py` and the README state only what stays true either way: the tile is a per-head-size throughput choice, not a validity rule, and a drafter usually has the smaller head size and therefore the larger tile, so the block that is fastest for the target is not necessarily the one that is fastest for the drafter. The `TODO` in `block_drafter.py` is scoped to that, giving the drafter its own block size instead of inheriting the target's, which is what a vLLM-style 16-token target page would need.


src/python/py/models/README.md
```diff
@@ -299,7 +299,7 @@ python builder.py -i path_to_local_folder_on_disk -o path_to_output_folder -p pr
...
... bounds without downloading device sequence lengths in every attention layer. Paged attention cannot be combined with `exclude_embeds` or `exclude_lm_head`. `paged_block_size` defaults to `256` and must be a power of two and at least `16`, matching what the ONNX Runtime PagedAttention op accepts; for models with short and long rotary caches, it must also evenly divide `original_max_position_embeddings`. The vendored FlashAttention paged kernel needs the block to be a multiple of its tile as well (256 for `head_size <= 64`, 128 for `head_size <= 128`, otherwise 64), so a smaller block stays valid but makes ORT fall back to a slower attention backend; a quantized KV cache never reaches that kernel and is exempt. `gpu_utilization_factor` defaults to `0.6` and must be greater than `0` and at most `1`. `max_batch_size` defaults to `100` and must be a positive integer no greater than `256`. `paged_chunk_size` defaults to `paged_block_size`, must be a positive integer, and is written to `search.chunk_size`; it applies only to models whose sliding-window layers are served from a ring of blocks, which hold `paged_chunk_size + window_size - 1` positions and therefore require chunked prefill.
```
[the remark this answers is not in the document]
tianleiwu: Fixed in 1502c9b6. The README now says the fallback applies to a causal model and that PagedAttention rejects a non-causal query at run time when FlashAttention is not eligible (`paged_attention.cc:516`), and it calls out that a block drafter shares the target block size and therefore needs a tile-compatible value.

src/python/py/models/builder.py
```diff
@@ -220,8 +220,11 @@ def check_extra_options(
                 raise ValueError(f"{key} must be a positive integer.")
             extra_options[key] = value
 
-        if "paged_block_size" in extra_options and extra_options["paged_block_size"] % 256 != 0:
-            raise ValueError("paged_block_size must be a multiple of 256.")
+        # This mirrors CheckInputs in onnxruntime paged_attention_helper.h. A page smaller than
+        # the FlashAttention tile is still valid; ORT falls back to another backend for it.
```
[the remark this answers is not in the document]
tianleiwu: Fixed in 1502c9b6. The implementation comment now describes the fallback as available to causal models only, and notes that a non-causal query has no fallback at all.

tianleiwu:
Addressed the suppressed review comments as well, in 1502c9b6:

- **Quantized KV cache wording (README.md:302, builder.py:848).** Correct — `flash_block_size_ok = kIsQuantizedCache || (block_size % flash_min_block_size) == 0`, so a quantized cache still reaches FlashAttention and is exempt only from the page-alignment check. Both places now say exactly that: exempt from the tile requirement alone, because that path dequantizes into a dense buffer and uses the non-paged varlen entry point, which has no page alignment to satisfy.
- **`test_paged_block_size.py` module docstring.** Qualified: a sub-tile block makes ORT pick a different backend for a causal model, while a non-causal query has no such alternative and is rejected at run time.
- **`test_precision.py:731` rationale.** Correct — `CheckInputs` is called from `PagedAttention::ComputeInternal`, not at load. The comment now says the built model failed its first forward pass.

Also added in 0ef13f85, following on from the DFlash 2 finding in the PR description: a `TODO` in `block_drafter.py` to give the drafter its own block size. It currently inherits the target's, which is what forces the whole export up to the drafter tile and blocks a vLLM-style 16-token page on the target.

src/python/py/models/README.md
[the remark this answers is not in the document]
tianleiwu: Follow-up in 053bf45e: microsoft/onnxruntime#32619 gives `is_causal=0` non-causal MEA, paged decode, XQA and latent backends, so the "non-causal PagedAttention requires FlashAttention" restriction is going away. Rather than document a restriction that is about to be removed, the README now states only the part that survives it: a sub-tile block is valid and makes ORT fall back to another backend, and a block drafter usually has the smaller head size and therefore the larger tile. The DFlash2 observation moved to the PR description.

src/python/py/models/builder.py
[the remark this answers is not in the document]
tianleiwu: Follow-up in 053bf45e: same reasoning as the README thread. With microsoft/onnxruntime#32619 the non-causal path stops being FlashAttention-only, so the implementation comment no longer describes the fallback as conditional on causality. It now says the tile is a per-head-size throughput choice rather than a validity rule, which is true before and after that PR.

tianleiwu:
@copilot resolve the merge conflicts in this pull request

src/python/py/models/README.md
```diff
@@ -299,7 +299,7 @@ python builder.py -i path_to_local_folder_on_disk -o path_to_output_folder -p pr
...
...nother attention backend. A quantized KV cache is exempt from the tile requirement alone: FlashAttention still serves it, through a dense dequantized path with no page alignment to satisfy. A block drafter (`dflash2_path`/`dspark_path`) shares the target's block size and usually has the smaller head size, so it reaches its tile at a larger block than the target does. `gpu_utilization_factor` defaults to `0.6` and must be greater than `0` and at most `1`. `max_batch_size` defaults to `100` and must be a positive integer no greater than `256`. `paged_chunk_size` must be a positive integer and is written to `search.chunk_size`. It caps the prompt tokens a single request contributes to one step, whereas `max_scheduled_tokens` caps the step as a whole; a value at or above `max_scheduled_tokens` therefore has no effect, and a smaller one lets concurrent prefills interleave rather than letting one request consume the step budget on its own. Models whose sliding-window layers are served from a ring of blocks hold only `paged_chunk_size + window_size - 1` positions, so they require chunked prefill and default to `paged_block_size`. For every other paged model it is written only when passed.
```
baijumeswani, line 302: Non-blocking documentation suggestion: until microsoft/onnxruntime#32619 is merged and included in the consumed ORT package, consider qualifying the fallback sentence as follows:

> For causal attention, a smaller block remains valid and makes ORT fall back to another attention backend. Current ORT releases still require FlashAttention for non-causal attention, so a block drafter may be disabled when the shared block size is smaller than its FlashAttention tile.

src/python/py/models/builder.py
```diff
@@ -220,8 +220,17 @@ def check_extra_options(
...
             extra_options[key] = value
 
-        if "paged_block_size" in extra_options and extra_options["paged_block_size"] % 256 != 0:
-            raise ValueError("paged_block_size must be a multiple of 256.")
+        # Mirrors CheckInputs in onnxruntime paged_attention_helper.h, which is the only hard
+        # bound. ORT's FlashAttention path wants block_size % tile == 0 on top of it, where tile
+        # is 256 for head_size <= 64, 128 for head_size <= 128 and 64 above that, and falls back
+        # to another backend otherwise. That is a throughput choice per head size, not a validity
+        # rule, and the head sizes involved are not known here, so only the op-level rule is
+        # enforced.
```
baijumeswani, line 228: Non-blocking: this keeps the implementation comment accurate for both the currently consumed ORT and the behavior expected after microsoft/onnxruntime#32619.

```suggestion
        # Mirrors CheckInputs in onnxruntime paged_attention_helper.h, which is the only hard
        # bound. ORT's FlashAttention path also requires block_size % tile == 0, where tile
        # depends on head size. Causal attention can fall back to another backend when that
        # condition is not met. Current ORT releases still require FlashAttention for non-causal
        # attention, so a sub-tile block can disable a block drafter until support for the other
        # non-causal backends is available. Only the op-level contract is enforced here.
```

test/python/builder/test_paged_block_size.py
```diff
@@ -0,0 +1,90 @@
...
+
+@pytest.mark.parametrize("block_size", [8, 15])
+def test_block_sizes_below_the_minimum_are_rejected(block_size):
+    with pytest.raises(ValueError, match="paged_block_size must be a power of two and at least 16"):
+        _check_extra_options({"use_paged_attention": "true", "paged_block_size": block_size})
+
+
+@pytest.mark.parametrize("block_size", [48, 96, 192, 768])
+def test_non_powers_of_two_are_rejected(block_size):
+    # 768 is a multiple of 256, so the old rule let it through and the op rejected it at load.
```
baijumeswani, line 82: Non-blocking wording correction: `CheckInputs` runs during execution rather than model load.

```suggestion
    # 768 is a multiple of 256, so the old rule let it through and the op rejected it on the first forward pass.
```

baijumeswani (approved):
The core validation change correctly matches the ONNX Runtime operator contract. Approved. The inline documentation suggestions are non-blocking and can be addressed later if needed.

src/python/py/models/README.md
tianleiwu: Assume that PR will be merged soon, and we do not want another PR to update this comment again.

src/python/py/models/builder.py
tianleiwu: Assume that PR will be merged soon, and we do not want another PR to update this comment again.

test/python/builder/test_paged_block_size.py
tianleiwu: Good catch. Let me fix it later.

PR #2570, opened 2026-09-15
[WebGPU] Support paged attention models

prathikr:
## Summary

Enable model-builder and runtime support for running paged-attention models with the WebGPU execution provider. WebGPU cannot query available device memory and its PagedAttention kernel does not consume the CPU-only `attention_metadata` input, so these models need an explicit cache capacity and a provider-specific input contract.

## Key changes

- Add and validate the `num_blocks` model-builder option, emitting it instead of `gpu_utilization_factor` for fixed-capacity paged KV caches.
- Omit `attention_metadata` from WebGPU paged models while preserving the remaining paged inputs.
- Fail with an actionable message if runtime memory-based cache sizing is attempted on WebGPU.
- Normalize full WebGPU provider names and forward `adapterIndex` to the initialization session.
- Document the WebGPU export command and add focused C++ and Python regression coverage.

## Testing

- `python -m pytest -q test/python/builder/test_paged_windowed_kv_cache.py test/python/builder/test_precision.py` (`108 passed`)
- `clang-format --dry-run --Werror src/config.cpp src/ep/webgpu/interface.cpp test/cpp/model_tests.cpp`

C++ build/tests were not run locally because CMake Tools has no configured build targets in this workspace and could not configure the project.

## Note

Adapted from https://github.com/microsoft/onnxruntime-genai/pull/2535


src/python/py/models/README.md
```diff
@@ -319,6 +319,16 @@ python -m onnxruntime_genai.models.builder -i path_to_local_folder_on_disk -o pa
 python builder.py -i path_to_local_folder_on_disk -o path_to_output_folder -p fp16 -e cuda -c cache_dir_to_store_temp_files --extra_options use_paged_attention=true prune_lm_head=true
 ```
 
+WebGPU requires a fixed cache capacity:
+
+```bash
+# From wheel:
+python -m onnxruntime_genai.models.builder -i path_to_local_folder_on_disk -o path_to_output_folder -p fp16 -e webgpu -c cache_dir_to_store_temp_files --extra_options use_paged_attention=true num_blocks=1024
```
baijumeswani, line 327: `num_blocks` is required for WebGPU, but the builder still allows this option to be omitted. In that case it writes `gpu_utilization_factor`, and engine startup later fails because WebGPU cannot query free device memory. Please validate this during export: when `use_paged_attention=true` and the EP is WebGPU, require a positive `num_blocks` and add a test that verifies the missing option is rejected.

src/python/py/models/builder.py
```diff
@@ -834,8 +834,8 @@ def get_args():
                     cumulative_sequence_lengths, and past_sequence_lengths metadata inputs are added. With
                     prune_lm_head=true, selects the final packed hidden state for each sequence so the model outputs
                     [batch_size, vocab_size] logits. By default, the model outputs [num_tokens, vocab_size] logits.
-                    Currently only supported for the CUDA execution provider with fp16 or bf16 precision. Cannot be
-                    combined with exclude_embeds or exclude_lm_head.
+                    Supports CUDA with fp16 or bf16 precision and WebGPU with fp16 precision. Cannot be combined with
```
baijumeswani: This says WebGPU FP16 is generally supported, but the WebGPU PagedAttention kernel rejects some model features at runtime, including non-zero attention softcap. Gemma2 is one concrete example: it can be exported here but will fail when PagedAttention runs. Please add WebGPU-specific capability checks that match the ORT kernel restrictions, or narrow this statement to the model configurations that are actually supported. A test for a rejected Gemma2/softcap configuration and a supported Llama or Qwen configuration would make the boundary clear.

src/python/py/models/builders/base.py
```diff
@@ -593,6 +593,8 @@ def make_inputs_init(self):
                 del self.input_names["attention_mask"]
             if not self.has_windowed_paged_layers():
                 del self.input_names["block_table_windowed"]
+            if self.ep == "webgpu":
+                del self.input_names["attention_metadata"]
```
baijumeswani: Please keep `attention_metadata` for WebGPU. Current ORT WebGPU PagedAttention accepts this CPU input and uses it to avoid reading sequence lengths back from the GPU. Removing it causes a device-to-host readback and synchronization in every attention layer, and prevents the intended graph-capture-friendly path. The engine already prepares this metadata, so the exporter should preserve the input and config binding.

test/python/builder/test_paged_windowed_kv_cache.py
```diff
@@ -472,6 +486,14 @@ def test_genai_config_honours_paged_chunk_size(monkeypatch, tmp_path):
     assert config["search"]["chunk_size"] == 64
```
baijumeswani, line 485: This verifies only the generated JSON, not that a WebGPU model can load and run. Please add an end-to-end test that exports a small supported model with `num_blocks`, creates a WebGPU session, confirms PagedAttention is assigned to WebGPU, and runs at least one prefill and decode step while comparing outputs/cache updates with a reference. This is needed to support the claim that the model can really be exported for WebGPU.

baijumeswani (commented):
Thanks for adding the WebGPU paged-attention path. I found a few cases where export succeeds but the resulting model may fail to initialize or run. Please see the inline comments.

test/python/builder/test_webgpu_paged_export.py
```diff
@@ -0,0 +1,131 @@
...
+    assert next_caches, "Exported model did not return any paged KV caches"
+    return output_by_name["logits"], next_caches
+
+
+def _assert_close(webgpu_value, cpu_value, label):
+    np.testing.assert_allclose(webgpu_value, cpu_value, rtol=2e-2, atol=2e-2, err_msg=label)
+
+
+def test_webgpu_paged_export_runs_prefill_and_decode_with_cpu_reference(tmp_path):
+    if "WebGPUExecutionProvider" not in ort.get_available_providers():
```
baijumeswani: This test is skipped in the Windows WebGPU CI job, even though the WebGPU plugin package is installed. The test checks the provider list before registering the plugin, so it never tests the new export path. Please register the WebGPU plugin first, using the existing helper in `test/python/_test_utils.py`, and only skip when the plugin package is not installed. If the package is installed but registration or session creation fails, the test should fail. Please confirm this test shows `PASSED` in the WebGPU CI log.

baijumeswani (commented):
Thanks for the updates. The `num_blocks` check and `attention_metadata` fix look good. I found two remaining issues below.

src/python/py/models/builder.py
```diff
@@ -834,8 +836,10 @@ def get_args():
                     cumulative_sequence_lengths, and past_sequence_lengths metadata inputs are added. With
                     prune_lm_head=true, selects the final packed hidden state for each sequence so the model outputs
                     [batch_size, vocab_size] logits. By default, the model outputs [num_tokens, vocab_size] logits.
-                    Currently only supported for the CUDA execution provider with fp16 or bf16 precision. Cannot be
-                    combined with exclude_embeds or exclude_lm_head.
+                    Supports CUDA with fp16 or bf16 precision. WebGPU supports fp16 only for causal, full-context
+                    attention with zero softcap, FP16 KV caches, and no Q/K normalization inputs; for example,
+                    Gemma2's non-zero attention softcap is unsupported. Cannot be combined with exclude_embeds or
```
baijumeswani, line 859: The help text now lists the WebGPU limits, but the builder still exports models that do not meet them. For example, Gemma2 can be exported even though its non-zero softcap will fail when the model runs. Please reject unsupported WebGPU settings during export, at least non-zero softcap, and add a test that Gemma2 is rejected with a clear error. This avoids creating a model that looks valid but cannot run.

baijumeswani (commented):
I found one correctness regression and three WebGPU performance/coverage gaps. Please address these before approval. Also, the current `linux_cpu_x64` check is crashing in `Dflash2ConfigTest.RunsFullAttentionDsparkAcrossRequestLifecycles`; that failure should be diagnosed or shown to be unrelated before merging.

src/dflash2_drafter.cpp
```diff
@@ -1010,13 +1012,23 @@ bool Dflash2Drafter::Propose(Tensor& aux_hidden_states, std::span<const Feed> fe
...
-      config_.inputs.past_sequence_lengths.c_str(), config_.inputs.block_table.c_str(),
-      config_.inputs.attention_metadata.c_str()};
+      config_.inputs.past_sequence_lengths.c_str(), config_.inputs.block_table.c_str()};
   std::vector<OrtValue*> inputs{packed_aux.GetOrtTensor(), input_ids.GetOrtTensor(),
                                 q_row_map.GetOrtTensor(), qkv_row_map.GetOrtTensor(),
                                 block_row_index.GetOrtTensor(), cumulative.GetOrtTensor(),
                                 past_lengths.GetOrtTensor(), block_table.GetOrtTensor(),
                                 metadata.GetOrtTensor()};
+  if (!config_.inputs.attention_metadata.empty()) {
+    auto& attention_metadata = Dflash2StepTensor(
```
baijumeswani: This recreates the same `step_tensors_.attention_metadata` tensor that was filled above with graph-bucket-safe bounds, then overwrites it with the exact values for this step. The earlier `inputs` entry points to this same tensor, while the duplicate entry appended here is not consumed because there is only one new input name. As a result, a captured CUDA graph can use bounds that are too small when a later step grows within the same bucket. Could we keep the existing `metadata` contents and only append the input name here?

src/python/py/models/builders/qwen.py
```diff
@@ -1273,6 +1273,7 @@ def make_dflash2_model(self, input_path):
             lm_head_quant=self.block_drafter_lm_head_quant(),
             embed_quant=self.block_drafter_embed_quant(),
             fuse_gate_up=self.dflash2_attrs["fuse_gate_up"],
+            include_attention_metadata=self.decoder.ep != "webgpu",
```
baijumeswani: The pinned WebGPU PagedAttention implementation now accepts `attention_metadata` and uses it to avoid reading sequence lengths back from the GPU for every layer. Omitting it from DFlash2 and DSpark adds avoidable synchronization and makes later graph capture harder. Could we keep this input enabled for WebGPU too, and update the PR description which still says WebGPU does not consume it?

test/python/builder/test_webgpu_paged_export.py
```diff
@@ -0,0 +1,151 @@
...
+        check=True,
+    )
+
+    config = json.loads((output_dir / "genai_config.json").read_text(encoding="utf-8"))
+    assert config["engine"]["dynamic_batching"]["num_blocks"] == _NUM_BLOCKS
+    assert config["model"]["decoder"]["inputs"]["attention_metadata"] == "attention_metadata"
+
+    model_path = output_dir / config["model"]["decoder"]["filename"]
+    session_options = ort.SessionOptions()
+    session_options.graph_optimization_level = ort.GraphOptimizationLevel.ORT_DISABLE_ALL
```
baijumeswani: Disabling all graph optimizations means this test does not exercise the session settings used in normal GenAI execution. It also cannot catch optimizer-inserted type changes such as an FP32 cast before the FP16-only WebGPU PagedAttention input. Please make the main run use the default production optimization level. A separate optimization-disabled run can remain as a diagnostic if needed.

test/python/builder/test_webgpu_paged_export.py
```diff
@@ -0,0 +1,151 @@
...
+    webgpu_session = ort.InferenceSession(str(model_path), sess_options=session_options)
+
+    cache_inputs = {
+        node_arg.name: np.zeros(_cache_shape(node_arg), dtype=np.float16)
+        for node_arg in webgpu_session.get_inputs()
+        if node_arg.name.startswith("past_key_values.")
+    }
+    assert cache_inputs, "Exported model has no paged KV-cache inputs"
+
+    webgpu_prefill, webgpu_caches = _run_step(webgpu_session, _PREFILL_TOKENS, 0, cache_inputs)
```
baijumeswani: This checks only a batch-1 direct ORT call and verifies that values are finite. Incorrect logits or unequal-length packed batching can still pass. Please add an Engine test with two different prompt lengths and compare each request's greedy output with its isolated run across several decode steps. A CPU/reference logits comparison here would also make this operator smoke test much stronger.

PR #2572, opened 2026-09-16
Fix GraphExecutor legacy EP fallback

mustjab:
## Problem

GraphExecutor creates helper sessions for operations such as WebGPU Cast. When no matching EP device is registered, `AppendExecutionProviderV2` returns false and GraphExecutor falls back to the legacy V1 API.

The fallback currently receives a default-constructed `ProviderOptions`, so its provider name is empty. ORT then rejects the helper session with:

```
Unknown provider name ''
```

This was observed during prompt append on a machine where the WebGPU V2 device lookup returned no match.

## Fix

Create `ProviderOptions` with the requested `ep_name` and reuse it for both V2 and V1 registration paths.

## Validation

- Windows Release build completed successfully, including `graph_executor.cpp` and `onnxruntime-genai.dll`
- CTest: 3/3 tests passed
- Unit tests: 626 passed


baijumeswani (approved):

kunal-vaishnavi (approved):

mustjab:
Replaced by #2576 using the identical commit on an in-repository branch so integration jobs can access required resources. Closing this fork-based PR.

PR #2573, opened 2026-09-16
Add prefix caching to the dynamic engine

bmehta001:
## Summary

- enable cross-request prefix caching by default for supported dynamic-Engine paged KV layouts, with `engine.dynamic_batching.prefix_caching: false` as an explicit opt-out
- retain content-addressed complete blocks with exact token and parent-lineage verification, bounded retention, reclamation, and transactional adoption/rollback
- atomically cache hybrid targets by pairing paged blocks with immutable fixed convolution/recurrent checkpoints at aligned prefill boundaries
- preserve existing configurations that omit the new setting: unsupported sliding-window or mirrored auxiliary layouts continue with caching disabled, while explicit unsupported `true` is rejected
- support fixed-pool DFlash2 through target-first reuse; a warm target hit runs target-only when the drafter cannot join at the adopted nonzero position
- report committed adopted tokens through turn usage and owner-thread prefix-metric snapshots
- leave speculative drafter sidecar caching as future work

This is a clean implementation on current `main`, informed by #2420 and adapted to the current Engine transaction, hybrid state, and request lifecycle.

## Supported layouts

| Layout | Behavior |
| --- | --- |
| Dense full-attention paged KV | Prefix caching enabled by default |
| Hybrid paged KV + fixed convolution/recurrent state | Enabled with aligned fixed-state checkpoints |
| DFlash2 with positive sliding window / fixed auxiliary pool | Target prefix caching enabled; warm adoption may fall back to target-only |
| Target sliding-window paged KV ring | Disabled when omitted or explicitly `false`; explicit `true` rejected |
| MTP, DSpark, or full-attention DFlash2 mirrored auxiliary cache | Disabled when omitted or explicitly `false`; explicit `true` rejected |

Builders emit explicit opt-outs for unsupported generated layouts.

## Validation

- Linux/WSL Engine suite: **681/681 passed**
- Model-builder suite: **703 passed**, 4 unrelated capability skips (QMoE pack/CUDA pybind and MRotaryEmbedding unavailable in the installed runtime)
- Repository `lintrunner -a`: passed
- Focused ownership, rollback, lineage, hybrid, compatibility, threading, MTP, and DFlash tests: passed
- Qwen3.8 27B DFlash2 direct continuation, 9,728 cached tokens:
  - enabled warm TTFT: **111.161 ms**
  - disabled TTFT: **3,406.641 ms**
  - **30.6x faster**, identical output
- Foundry Local Qwen3.8 27B DFlash2 continuation:
  - 5,000-word case, 5,376 cached tokens: **151.904 ms vs 1,914.396 ms** (**12.6x**)
  - 10,000-word case, 11,008 cached tokens: **107.417 ms vs 3,826.831 ms** (**35.6x**)
- Direct DFlash statistics confirmed target-only warm fallback: the cold request produced 41 draft forward passes; the warm request reused 9,728 target tokens without increasing that count

## Validation limitation

Windows configuration with Visual Studio 2026 succeeds, but compilation remains blocked on the test machine by missing Spectre-mitigated libraries (`MSB8040`). Security flags were not weakened.


bmehta001:
A clean Foundry Local `main` integration test now passes on an A100 using Qwen3 1.7B paged attention and this PR at `15807278b53a3a39eeb30f3bec83c314d736ee36`.

- Foundry Local: `e06e92c6376737151c6eda36ad63a2e751f40213`
- Common rendered prompt: 9,907 tokens
- Prefix-enabled cold request: `cached_prompt_tokens=0`, TTFT 791.4 ms
- Prefix-enabled second distinct request: `cached_prompt_tokens=9728`, TTFT 15.2 ms
- No-cache control: both requests reported `cached_prompt_tokens=0`; second-request TTFT 290.6 ms
- Enabled/control outputs matched token-for-token for both suffixes
- Qwen output began with `<think>`, confirming the reasoning path
- Engine remained alive across conversation closure; GPU returned to 0 MiB / 0% after the run

This test exposed that prefix adoption was not being copied into terminal `TurnUsage`, even though reuse was occurring. Commit `15807278` fixes that accounting and adds coverage ensuring later continuous-decoding turns reset the cached-token count. The full CUDA Engine suite passes 677/677 and repository lintrunner reports no issues.

bmehta001:
Hybrid Qwen3.8 target-state support is now in `23a22e90`.

The cache now treats one canonical target prefix as a composite snapshot: paged KV plus every fixed convolution/recurrent tensor at the same committed block boundary. Fixed snapshots are preallocated and independently reclaimable, and a valid target hit remains usable when the fixed-size DFlash2 sidecar cannot join; that request falls back to target-only execution without shortening the target boundary.

### Validation

- Linux `engine_unit_tests`: **673/673 passed**
- Repository `lintrunner -a`: **passed**
- CUDA 12.8 / ORT 1.30 core, Python module, `unit_tests`, and `engine_unit_tests`: **built successfully**
- Windows VS 2026: still blocked before project compilation by the existing missing Spectre libraries (`MSB8040`)

Real A100 validation used `paged-attention/qwen3.8-27b-int4-int8-kv-dflash2` with Foundry Local `e06e92c6376737151c6eda36ad63a2e751f40213`:

| Scenario | Common prefix | Cached | Enabled warm TTFT | Disabled warm TTFT | Speedup | Output |
|---|---:|---:|---:|---:|---:|---|
| Direct cross-request | 9,949 tokens | 9,728 | 111.161 ms | 3,406.641 ms | 30.6x | identical |
| Foundry continuation, 5,000 words | 5,555 tokens | 5,376 | 151.904 ms | 1,914.396 ms | 12.6x | identical |
| Foundry continuation, 10,000 words | 11,055 tokens | 11,008 | 107.417 ms | 3,826.831 ms | 35.6x | identical |

The continuation runs generated real, closed Qwen reasoning on turn 1, then rebuilt turn 2 from the canonical visible transcript (`P+V+B`), omitting hidden reasoning (`R`). Cache accounting matched the deepest safe block boundary exactly.

DFlash2 target-first fallback was also observed directly: the cold request performed 41 drafter forward passes; the warm request adopted 9,728 target tokens and left the cumulative drafter-pass count at 41, proving that it ran target-only rather than rebuilding or shortening the valid target hit.

One Foundry packaging/integration gap surfaced: this Qwen3.8 artifact's tokenizer does not publish BOR/EOR marker metadata even though its chat template explicitly uses `<think>` / `</think>`. The benchmark therefore seeded Foundry's token-aware splitter with those exact template markers and the prompt-open reasoning state. This does not affect target cache correctness, but the model metadata or Foundry fallback should be fixed separately for normal high-level reasoning streaming.

src/engine/block.h
```diff
@@ -40,17 +49,31 @@ struct Block {
 
   std::vector<size_t> SlotIds() const;
 
+  size_t RefCount() const { return ref_count_; }
+  bool IsShared() const { return ref_count_ > 1; }
+  bool IsShareable() const { return IsFull() && HasIdentity(); }
```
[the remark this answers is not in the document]
bmehta001: Fixed in e3a6a08f. `PagedCacheBlockTable` is now move-only, so copying `shared_ptr<Block>` handles can no longer duplicate table ownership without the corresponding `BlockPool::AddRef`. Compile-time assertions cover both deleted copy operations, and test fixtures now construct tables by move.

src/engine/prefix_cache.cpp
```diff
@@ -0,0 +1,261 @@
...
+      // the caller stops here rather than indexing entries no lookup can ever verify.
+      ++metrics_.hash_collisions;
+      return nullptr;
+    }
+    // Two sequences computed the same prefix before either was indexed. The first physical copy
+    // serves every lookup, so this one stays private; the chain still continues through it because
+    // the indexed copy holds exactly the same tokens behind the same parent.
+    ++metrics_.duplicate_registrations;
+    Reorder(existing->second, parent);
+    return existing->second.identity;
```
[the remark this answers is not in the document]
bmehta001: Fixed in e3a6a08f. Registration now distinguishes indexed, duplicate, collision, and transient capacity-refusal outcomes. A duplicate returns no parent identity and permanently stops that request from sealing its suffix, so descendants cannot depend on a canonical block the request does not retain.

src/engine/prefix_cache.cpp
```diff
@@ -0,0 +1,261 @@
...
+  // the head of the chain ends up the most recently used. Evicting the head would orphan every
+  // block behind it -- their identities chain from it, so no lookup can reach them again -- while
+  // evicting the tail just shortens the prefix that stays reusable.
+  for (const auto& it : hits) {
+    Reorder(it->second, it->second.identity->parent);
+  }
+  match.token_count = hits.size() * block_size;
+
+  ++metrics_.hits;
+  metrics_.matched_tokens += match.token_count;
```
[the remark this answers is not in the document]
bmehta001: Fixed in e3a6a08f. `Match()` now records lookup observations only. Hit count and matched-token metrics are published from the committed reservation deltas after both paged and fixed-state cache publication, so deferred or rolled-back candidates are not reported as adopted.

src/config.h
```diff
@@ -655,6 +655,10 @@ struct Config {
       std::optional<float> gpu_utilization_factor;  // Fraction of free GPU memory to use for key-value cache.
       size_t max_batch_size{16};                    // Maximum batch size for dynamically batching requests.
       size_t max_scheduled_tokens{2048};            // Maximum tokens in one dynamically batched model run.
+      bool prefix_caching{true};
```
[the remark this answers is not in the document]
bmehta001: Fixed in 9611edb1. The JSON parser now records whether `prefix_caching` was explicitly present, and one shared effective policy is resolved before hybrid checkpoint and paged-cache allocation. Omitted legacy configs disable caching for unsupported layouts; explicit unsupported `true` still fails clearly. Zero configured capacity disables the policy before layout rejection, and positive fractions retain at least one block.

src/engine/cache_manager.cpp
```diff
@@ -340,18 +346,77 @@ PagedCacheManager::PagedCacheManager(std::shared_ptr<Model> model,
...
+std::shared_ptr<const PrefixCacheMatch> PagedCacheManager::MatchPrefix(
+    const Request& request) {
+  if (!key_value_cache_->PrefixCachingEnabled()) {
+    return nullptr;
+  }
+  const auto tokens = request.TokensCpu();
+  if (tokens.size() <= 1) {
+    return nullptr;
+  }
+  auto match = key_value_cache_->MatchPrefix(tokens, tokens.size() - 1);
```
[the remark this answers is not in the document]
bmehta001: No code change needed here. The Engine request surface is token-only: `Request::BeginTurn` and `Engine::BeginTurn` accept only `span<const int32_t>` and expose no `SetInputs`/NamedTensors path for image, audio, grid, or position tensors (`src/engine/request.h:156-160`, `src/engine/engine.h:200-202`). The Engine simple decoder also explicitly rejects multimodal coordinates (`src/engine/decoders/simple_decoder.cpp:34`). Therefore a multimodal request with media-conditioned inputs cannot reach `PagedCacheManager::MatchPrefix`; multimodal generation remains outside this Engine API.

src/engine/engine.h
```diff
@@ -188,6 +188,9 @@ struct Engine : std::enable_shared_from_this<Engine>,
    * synchronization.
    */
   SpeculativeStats GetSpeculativeStats() const;
+  const PrefixCacheMetrics* PrefixCacheStats() const {
+    return cache_manager_->PrefixMetrics();
+  }
```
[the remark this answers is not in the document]
bmehta001: Fixed in 9611edb1. `Engine::PrefixCacheStats()` now validates the Engine owner thread and returns an `optional<PrefixCacheMetrics>` snapshot by value, matching the synchronization contract of `GetSpeculativeStats()`. Tests cover off-owner-thread rejection and reacquire snapshots after subsequent runs.

src/python/py/models/builders/qwen.py
```diff
@@ -1091,6 +1091,7 @@ def add_mtp_to_genai_config(self, out_dir):
                 "present_value_names": "present.%d.value",
             },
         }
+        genai_config["engine"]["dynamic_batching"]["prefix_caching"] = False
```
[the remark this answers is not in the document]
bmehta001: Fixed in 9611edb1. The MTP builder now updates `prefix_caching` only when `engine.dynamic_batching` exists, preserving static/non-paged exports. A regression test writes an MTP section into a config with no `engine` object and verifies it remains valid.

src/engine/scheduler.cpp
```diff
@@ -208,6 +215,7 @@ StepPlanningResult DynamicBatchScheduler::PlanStep(StepPlan& plan) {
     candidate.entry.request = request;
     candidate.entry.request_id = request.get();
     candidate.entry.sequence_length_before = snapshot.current_sequence_length;
+    candidate.entry.prefix_match = std::move(prefix_match);
```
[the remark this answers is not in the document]
bmehta001: No change needed here. The assignment at lines 226-228 only initializes the candidate before ordering. Before the first `PlanStepResources` call, every candidate copied into `plan.requests` has `target_cache_slots` recomputed from `candidates[candidate_index].processed_sequence_length` at lines 269-275; that field is the effective prefix-aware length set at lines 203-205. The cache planner therefore never sees the raw-snapshot value. `DensePagedPrefixCacheSkipsCommittedFullBlocks` and the hybrid warm-hit test exercise successful 8-token adoption through this exact first planning pass.

src/engine/cache_manager.cpp
```diff
@@ -155,6 +159,7 @@ class CompositeCacheStepReservation final : public CacheStepReservation {
     if (fixed_reservation_) {
       fixed_reservation_->PublishCommit();
     }
+    cache_.RecordPrefixAdoptions(*paged_reservation_);
```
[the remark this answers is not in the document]
bmehta001: Fixed in b70f8a19. `CommitValidated()` now preserves `adopted_blocks_` through the composite reservation post-commit publication step. The pool-local references still transfer to the committed tables, while the retained `shared_ptr` handles keep the exact block span valid for `RecordPrefixAdoptions`; a reservation regression test verifies the handles remain available after commit.

src/engine/prefix_cache.h
```diff
@@ -0,0 +1,211 @@
...
+
+  const PrefixCacheOptions& Options() const { return options_; }
+
+  /**
+   * @brief Longest run of block-aligned leading tokens of `tokens` that is already resident.
+   * @param tokens The whole sequence the request wants computed.
+   * @param max_adoptable_tokens Upper bound on tokens the caller may skip. A request must always
+   *        compute at least its last token, so the caller passes one less than the sequence length.
+   *
+   * Marks every block it hands out as recently used, so adoption feeds the eviction order.
```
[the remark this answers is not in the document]
bmehta001: Fixed in b70f8a19. The `Match` contract now states that lookup is eviction-order-neutral and requires the caller to invoke `RecordAdoption` only after adoption commits, matching the implementation and LRU tests.

src/engine/paged_cache_reservation.cpp
```diff
@@ -80,6 +71,9 @@ PagedCacheBlockTable& PagedCacheBlockTable::operator=(
     committed_slots_ = other.committed_slots_;
     blocks_ = std::move(other.blocks_);
     window_blocks_ = std::move(other.window_blocks_);
+    sealed_blocks_ = other.sealed_blocks_;
```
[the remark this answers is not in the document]
bmehta001: Fixed in 57fe5d2c. Move assignment now terminates unless the destination has no target or window block handles, preventing it from silently discarding pool-local references. The production erase path releases both pools and clears those handles before vector compaction invokes move assignment. A death test covers populated-destination rejection, and all 683 Engine tests pass.

src/engine/paged_cache_reservation.cpp
```diff
@@ -300,24 +329,37 @@ PagedCacheReservation::PagedCacheReservation(
...
       table.request_id_ = request.request_id;
-      table.blocks_.reserve(new_blocks);
+      table.committed_slots_ = committed_slots;
+      table.blocks_.reserve(CheckedAdd(
+          adopted_block_count, new_blocks, "new table block capacity"));
+      if (match) {
+        table.blocks_ = match->blocks;
+        table.sealed_blocks_ = adopted_block_count;
+        table.sealed_identity_ = match->blocks.back()->IdentityPtr();
+      }
```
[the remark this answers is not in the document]
bmehta001: Fixed in 98f1f0d8. New-table construction now copies the adopted handles first and reserves the final `adopted_block_count + new_blocks` capacity afterward, so `CommitValidated()` retains the preallocated no-throw insertion capacity.

test/cpp/engine/prefix_cache_tests.cpp
```diff
@@ -0,0 +1,530 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+// Unit tests for the content-addressed block index, the reference counting that makes block sharing
+// safe, and the copy-on-write guard that keeps a shared block from being written. These exercise the
+// production policy types directly and need no model, device, or engine.
+
+#include <array>
+#include <memory>
+#include <span>
```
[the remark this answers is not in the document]
bmehta001: Fixed in 98f1f0d8. The test now includes the repository `span.h` compatibility shim rather than the C++20 standard header, preserving GCC <10 and Android C++17 builds.

src/engine/paged_cache_reservation.cpp
```diff
@@ -63,23 +63,20 @@ size_t CheckedAdd(
...
-    *this = std::move(copy);
-  }
-  return *this;
-}
-
 PagedCacheBlockTable& PagedCacheBlockTable::operator=(
     PagedCacheBlockTable&& other) noexcept {
   if (this != &other) {
+    if (!blocks_.empty() || !window_blocks_.empty()) {
+      std::terminate();
```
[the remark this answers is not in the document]
bmehta001: Fixed in 7357fbc6. Removal now releases and clears the selected table, moves only the last resident table into that single empty destination, and pops the moved-from tail; `PagedKeyValueCache` rebuilds the request index immediately after removal. A three-resident-table regression covers removing the first entry, and all 685 Engine tests pass.

baijumeswani (commented):
Thanks for adding this. I built the exact PR head with ORT 1.30.0 and CUDA 12.8, ran the Engine tests, and exercised the Qwen3.8-27B hybrid model through Foundry Local and GitHub Copilot CLI BYOK.

The warm-cache result is excellent: an 88K-token request dropped from about 68 seconds cold to about 0.6 seconds TTFT when repeated. However, I found several issues that I think should be addressed before prefix caching is enabled by default for hybrid models.

The biggest concern is cold-prefill performance. This implementation reduces hybrid prefill to one 256-token cache block per model run. In a controlled test with prefix caching disabled, three 96K-token prefills averaged 46.38 seconds with a 1,024-token chunk and 76.05 seconds with a 256-token chunk. The 1,024-token chunk was 1.64x faster. With prefix caching enabled, the same block-sized behavior also adds a fixed-state copy and GPU synchronization at every boundary.

I also found a staged-state correctness path, a post-commit failure risk, an O(cache-size) planning scan, a checkpoint-lifetime edge case, and metrics that mix scheduler retries with actual cache use. I left inline comments with concrete suggestions and focused tests for each.

Validation completed:
- exact head `d60fb363dedeb91fd82dc5f212977bbbbd706f79`;
- PR-built GenAI core/CUDA plugin with ORT 1.30.0 and the Foundry CUDA EP;
- 747/747 Engine tests passed;
- 261 general unit tests passed, with 22 model-dependent skips;
- real Qwen cold, warm, changed-suffix, concurrent warm, and resumed Copilot BYOK scenarios produced correct output;
- service shutdown released GPU memory cleanly.

My suggested release posture is to keep hybrid prefix caching opt-in until the cold-prefill regression and correctness issues are resolved. Dense paged-only models do not have the fixed-state checkpoint cost and can be considered separately.

src/engine/engine.cpp
```diff
@@ -1702,6 +1727,7 @@ void Engine::RunDynamic() {
           "Dynamic scheduler returned no executable work while requests remain pending.",
           nullptr);
     }
+    PrefixAdoptionGuard prefix_adoption_guard{step_plan_};
```
baijumeswani: Please create this guard immediately after `PlanStep()` returns, before handling `unserviceable_request_id`.

Planning can select request A and stage its prefix adoption while also reporting unrelated request B as permanently unserviceable. The early return above fails B but skips this guard, so A remains staged and can fail when planning runs again.

Recommended fix: construct the guard directly after planning, or explicitly roll back all staged adoptions before every early return. Please add a test with one executable prefix-hit request plus one unrelated unserviceable request, then verify the prefix-hit request succeeds on the next `Run()`.

src/engine/cache_manager.cpp
```diff
@@ -490,6 +558,18 @@ void PagedCacheManager::DetachRequestForTeardown(
 
 bool PagedCacheManager::SupportsDynamicBatching() const { return true; }
 
+size_t PagedCacheManager::MaxQueryTokensPerRequest() const {
+  size_t limit = key_value_cache_->MaxQueryTokensPerRequest();
+  if (fixed_state_pool_ &&
+      key_value_cache_->PrefixCachingEnabled()) {
+    const size_t checkpoint_interval = key_value_cache_->BlockSize();
```
baijumeswani: This makes every hybrid prefill chunk no larger than one KV block. For the Qwen3.8 package, that changes the effective chunk from 1,024 tokens to 256 tokens.

A controlled prefix-cache-disabled test with three identical 96K-token prompts averaged 46.38s TTFT at 1,024 tokens versus 76.05s at 256 tokens. The larger chunk was 1.64x faster. Enabling caching also copies fixed state and synchronizes the GPU at each of these smaller boundaries.

Recommended fix: keep the normal larger prefill chunk when it ends on a block boundary, checkpoint fixed state only at that final boundary, and allow adoption only at boundaries that have both KV and fixed-state data. At minimum, make hybrid prefix caching opt-in until this cold-prefill cost is removed.

src/engine/engine.cpp
```diff
@@ -2023,6 +2050,7 @@ void Engine::RunDynamic() {
         step_plan_.requests[i].request->CommitStep(
             step_plan_.requests[i], step_results_[i]);
       }
+      cache_manager_->SealCommittedBlocks(step_plan_);
```
baijumeswani: `SealCommittedBlocks()` performs optional cache publication after the request, paged cache, and fixed state have already committed. Prefix registration and checkpoint creation can still allocate memory or copy device state. Any exception here enters the catch below and permanently marks the Engine unhealthy, even though generation itself committed successfully.

Recommended fix: allocate everything needed for publication before crossing the commit boundary, or make host-side cache publication best-effort and record a refusal metric instead of failing the Engine. Device copy failures can remain fatal if recovery is unsafe, but ordinary metadata allocation failure should only skip caching this boundary. Please add an allocation-failure test.

src/engine/paged_key_value_cache.cpp
```diff
@@ -723,7 +910,8 @@ StepPlanningResult PagedKeyValueCache::PlanStepResources(StepPlan& plan) const {
         "Step plan request limit exceeds the configured batch size.");
   }
 
-  const size_t available_blocks = block_pool_->AvailableBlocks();
+  const size_t available_blocks =
+      block_pool_->AvailableBlocks() + prefix_cache_->ReclaimableBlocks();
```
baijumeswani, line 918: `ReclaimableBlocks()` scans every cached prefix entry, and this call runs during planning for every step, including one-token decode steps. As the cache grows to tens of thousands of blocks, decode planning time will grow with total cached context.

Recommended fix: maintain an exact reclaimable-block counter as references change, and keep immediately reclaimable entries in a dedicated LRU/list so planning and eviction do not scan the full index. Please add a benchmark that fills a large prefix cache and checks decode planning time and inter-token latency.

src/engine/cache_manager.cpp
```diff
@@ -340,18 +346,80 @@ PagedCacheManager::PagedCacheManager(std::shared_ptr<Model> model,
...
+      continue;
+    }
+    if (fixed_state_pool_->AvailablePrefixCheckpoints() == 0) {
+      key_value_cache_->ReclaimPrefixCheckpoints(1);
+    }
+    if (fixed_state_pool_->AvailablePrefixCheckpoints() == 0) {
+      continue;
+    }
+    auto checkpoint =
+        fixed_state_pool_->CapturePrefixCheckpoint(entry.request_id);
```
baijumeswani, line 420: A request that adopted a hybrid prefix still has the matched checkpoint referenced by its `StepPlan` while this code tries to capture the next checkpoint. When checkpoint capacity equals `max_batch_size` (especially 1), that old reference can prevent reclaiming the only slot, so the newly completed boundary is not checkpointed.

Recommended fix: release the consumed `prefix_match` checkpoint reference after restore/commit and before sealing new boundaries. Please add a three-request test: request 1 creates checkpoint N, request 2 adopts N and processes exactly one full block plus a tail, and request 3 must adopt N + block size.

src/engine/prefix_cache.cpp
```diff
@@ -0,0 +1,368 @@
...
+  }
+
+  const size_t block_size = block_pool_.BlockSize();
+  if (block_size == 0) {
+    return match;
+  }
+
+  const size_t adoptable =
+      std::min(max_adoptable_tokens, tokens.size()) / block_size * block_size;
+  ++metrics_.lookups;
```
baijumeswani, line 88: This counts every scheduler probe as a lookup, including the same queued request being reconsidered many times. Hits and matched tokens are recorded only after an adoption commits. Under load, the reported hit rate can therefore look poor simply because a matched request was deferred repeatedly.

Recommended fix: report separate counters for planning probes, candidate matches, deferred matches, and committed adoptions. If `lookups` is intended to be the hit-rate denominator, increment it only when the request is selected for an executable step. Add a test where a matching request is deferred several times and verify the committed hit rate is unchanged.

src/engine/engine.cpp
bmehta001: Fixed in f03ada4b. `PrefixAdoptionGuard` is now constructed immediately after `PlanStep()` returns (`src/engine/engine.cpp:1700`), before the unserviceable and other early-return paths. `UnserviceableRequestRollsBackStagedPrefixAdoption` covers an executable warm hit planned alongside an unrelated unserviceable request and verifies the warm request adopts successfully on the next run.

src/engine/cache_manager.cpp
bmehta001: The benchmark result is useful, but the one-block mechanism in this comment was already removed by 8e5e95af before this follow-up. `PagedCacheManager::MaxQueryTokensPerRequest()` now delegates only to the paged-cache safety limit (`src/engine/cache_manager.cpp:561-566`); fixed state does not cap prefill chunks. `HybridPrefixCacheResumesPartwayThroughConfiguredChunk` verifies an 8-token configured chunk resumes from a 4-token hybrid checkpoint without shrinking to the 4-token block size. No additional change is needed for this mechanism; the cold benchmark can still guide separate performance investigation if it reproduces on the current head.

src/engine/engine.cpp
bmehta001: Fixed in f03ada4b. A `std::bad_alloc` from optional post-commit prefix publication now records `publication_refusals` and preserves the committed generation step (`src/engine/engine.cpp:2057-2061`), while invariant failures remain fatal. Fixed-state device copy/synchronization failures are translated to a non-allocating typed exception after marking the pool unhealthy, so they cannot be mistaken for recoverable metadata pressure. Allocation and invariant-failure regressions cover both outcomes.

src/engine/paged_key_value_cache.cpp
bmehta001: No change needed here: the current implementation does not scan cached entries. `PrefixCache::ReclaimableBlocks()` is exactly `return reclaimable_entries_.size();` (`src/engine/prefix_cache.cpp:369-371`), and block-reference observer callbacks maintain that list incrementally. `ReclaimableCountTracksRequestReferencesExactly` covers the maintained count across release, reacquire, and eviction.

src/engine/cache_manager.cpp
bmehta001: Fixed in f03ada4b. After adoption and request commit consume the match, each `StepPlan` entry now releases `prefix_match` before `SealCommittedBlocks()` (`src/engine/engine.cpp:2054-2058`). `HybridPrefixAdoptionRecyclesSingleCheckpointSlotInSameStep` uses checkpoint capacity 1 and verifies a third request adopts the deeper boundary created by the second request.

src/engine/prefix_cache.cpp
bmehta001: Fixed in f03ada4b. Metrics now separate planning `lookups`, successful candidate `matches`, `deferred_matches`, and committed-adoption `hits` (`src/engine/prefix_cache.h:84-88`). The scheduler records matching candidates omitted by capacity while excluding the explicitly unserviceable request. Unit coverage distinguishes a candidate match from committed adoption, and `DeferredPrefixMatchesAreSeparateFromCommittedAdoptions` verifies repeated deferrals leave one committed hit.

src/engine/paged_key_value_cache.cpp
bmehta001: (That's bc I changed it last night since your review - it was correct when you commented this)

src/engine/cache_manager.cpp
bmehta001: (I fixed it last night after we discussed this; CoPilot is replying late)

src/engine/cache_manager.cpp
```diff
@@ -340,18 +346,80 @@ PagedCacheManager::PagedCacheManager(std::shared_ptr<Model> model,
...
+    }
+    key_value_cache_->SealCommittedBlocks(
+        entry.request_id, entry.request->TokensCpu());
+    if (!fixed_state_pool_ ||
+        !key_value_cache_->CanAttachPrefixCheckpoint(
+            entry.request_id, entry.target_cache_slots)) {
+      continue;
+    }
+    if (fixed_state_pool_->AvailablePrefixCheckpoints() == 0) {
+      key_value_cache_->ReclaimPrefixCheckpoints(1);
```
baijumeswani, line 414: This can remove the only checkpoint that a repeated block-aligned prompt can use. The production Qwen model has 256-token blocks and `max_batch_size=1`, so it has one fixed-state checkpoint slot. For a 96,000-token prompt, the earlier usable checkpoint is reclaimed here and replaced by a checkpoint at token 96,000. A new request must still process its final token, so it cannot adopt that full-boundary checkpoint. In the real model test, an identical replay then reported 0 cached tokens and took 46.02 seconds. Changing only the length to 96,001 kept the 95,232-token checkpoint and reduced first-token time to 0.54 seconds. Please keep the newest earlier checkpoint that is still adoptable, or provide another checkpoint slot for this case. Please also add a hybrid test where the prompt length is exactly block-aligned and assert that replay adopts every block except the final one.

baijumeswani (commented):
Thanks for addressing the earlier concerns. I rebuilt this exact head with ORT 1.30 and CUDA 12.8, then tested the Qwen3.8-27B hybrid model directly through the Engine.

The prefill chunk regression is fixed. A 96K cold request took 46.55 seconds with prefix caching enabled and 46.51 seconds with it disabled when the configured chunk size was 1,024. The 256-token configuration still took 76.05 seconds, so prefix caching is no longer forcing the slower 256-token path.

Normal warm reuse also works well. For a 96,001-token prompt, the replay reported 95,232 cached tokens and first-token time dropped from 46.60 seconds to 0.54 seconds. A changed-suffix request had the same cache reuse, and its output matched a separate uncached run. Cached-token usage is now reported correctly.

I found one remaining issue with prompts that end exactly on a cache block boundary. With 256-token blocks, a 96,000-token prompt gets no reuse: the identical replay reports 0 cached tokens and still takes about 46 seconds. The inline comment explains why. This is important for coding workloads because fixed context sizes such as 32K, 64K, 96K, and 128K are all block-aligned.

All 747 Engine unit tests and the new cache microbenchmarks passed. I recommend fixing the block-aligned hybrid checkpoint case and adding a focused test before merging.

baijumeswani (approved):

src/engine/cache_manager.cpp
bmehta001: Fixed in 686633ef. When the checkpoint pool is full, an exact prompt-boundary checkpoint no longer reclaims the newest earlier checkpoint, because the replay must execute its final prompt token and cannot adopt the terminal boundary. Added `HybridBlockAlignedPromptRetainsLastAdoptableCheckpoint`, which uses a single checkpoint slot and verifies that an exactly block-aligned replay adopts every block except the final one.

Validation:
- 719/719 Engine tests pass.
- Qwen3.8-27B on A100, 256-token blocks, `max_batch_size=1`, 1,024-token prefill chunks, exact 8,192-token prompt: replay reports 7,168 cached prompt tokens.
- TTFT p50 improved from 2,897 ms (cache disabled) to 361 ms (cache enabled); total latency p50 improved from 3,042 ms to 505 ms. Outputs remained identical.

PR #2575, opened 2026-09-16
feat: add LFM2-MoE support

ykhrustalev:
**Problem**
LFM2-MoE models (LFM2.5-8B-A1B, LFM2-24B-A2B) cannot be exported. They reuse the LFM2 conv/attention stack but route the later layers through sparse experts, and their router selects experts on `sigmoid(logits) + expert_bias` while mixing them with the unbiased `sigmoid(logits)` — routing the fused `MoE`/`QMoE` op cannot express, since it only does softmax over the top-k.

**Solution**
- Adds `LFM2MoEModel` for `Lfm2MoeForCausalLM`, exported as model type `lfm2_moe`: dense MLP for the first `num_dense_layers`, fused `MoE`/`QMoE` with interleaved SwiGLU experts after that
- Performs expert selection in the graph and feeds the op `log(sigmoid(logits))` at the selected experts and a large negative sentinel elsewhere, so the op's softmax reproduces the model's normalized routing weights exactly on every EP
- Keeps the router `MatMul` out of int4 quantization so rounding cannot flip an expert choice
- Registers `lfm2_moe` in the runtime through `ModelType::IsLFM2`, reusing the LFM2 conv-state cache and its rewind restriction
- Fixes `make_moe_init` writing `moe_attrs` keys nothing reads, which made every int4 MoE export emit the float `MoE` op instead of `QMoE`
- Switches non-CUDA block-wise `QMoE` weights to the signed-scale MatMulNBits encoding already used for CUDA, so no block extreme is clipped

**Testing**
- LFM2.5-8B-A1B verified against `transformers`: fp32 matches to 5e-5 (KL 0.0000) on CPU; int4 KL 0.096 / 89% top-1; CUDA fp16 KL 0.006 and int4 KL 0.10 on an A10
- New `test/python/builder/test_lfm2_moe.py` pins the router graph, the dense/MoE layer split, expert repacking, and the `QMoE` op selection; new C++ case covers the `lfm2_moe` runtime path

Depends on microsoft/onnxruntime#32644 for usable CPU decode speed on these models.


ykhrustalev:
Thanks — all five addressed in `fix: address LFM2-MoE review findings`.

- **`layer.feed_forward` vs `layer.mlp`**: the feed-forward module is now resolved through `get_feed_forward_module`, which accepts either name, and the `w1/w3/w2` aliasing only runs when those attributes exist (the quantized IR already uses `gate_proj`/`up_proj`/`down_proj`). Note the quantized-checkpoint IR still has no `operator_norm`, `conv` or `ffn_norm`, so pre-quantized checkpoints remain unsupported for the whole LFM2 family, as on `main` — this change just removes the LFM2-MoE-specific obstacle.
- **Packed expert biases**: `make_moe_expert_initializers` never writes biases, so the zeros were the only source rather than an overwrite. Packed biases are now forwarded when a checkpoint supplies them, with zeros as the fallback (LFM2-MoE experts have no bias).
- **`ROUTER_SENTINEL`**: moved into `moe_attrs["router_sentinel"]`.
- **Router method size**: split into `make_moe_router_scores`, `make_moe_router_selection` and `make_moe_router_mask`, with `make_moe_router` composing them.

Re-verified after the refactor: exported fp32 graph still matches `transformers` to 3e-7 on the dummy model and the int4 graph still emits `QMoE`; 730 builder tests pass, including new cases for the loader-independent accessor and the bias passthrough.

src/python/py/models/builders/base.py
```diff
@@ -1031,11 +1031,11 @@ def make_moe_init(self):
 
         # MXFP4 and NVFP4 both resolve to the "mx" kind; the QMoE op tells them apart by dtype name
         # ("mxfp4" -> op "fp4", "nvfp4" -> op "nvfp4"). Integer dtypes use the plain "int" QMoE path.
-        self.moe_attrs["moe_op_type"] = "QMoE" if moe_descriptor.is_quantized else "MoE"
+        self.moe_attrs["op_type"] = "QMoE" if moe_descriptor.is_quantized else "MoE"
```
tianleiwu, line 1044: This is a real and valuable regression fix, and I confirmed the root cause: #2386 renamed the **writer** to `moe_op_type` / `qmoe_quant_type` while every **reader** still uses `op_type` / `quant_type` (`base.py:4944`, `base.py:5131`, `qwen.py:841`, `gptoss.py:519`, `gptoss.py:569`). Because `moe_attrs` seeds `"op_type": "MoE"` and `"quant_type": "int"` at `base.py:414`/`base.py:426`, the wrong keys were dead writes that never raised `KeyError` — the failure was completely silent.

Two things worth calling out:

**The blast radius is larger than the PR description states.** Besides int4/int8 emitting float `MoE`, `quant_type` stuck at `"int"` means `make_qmoe_op` computed `is_fp4 = False`, so **MXFP4/NVFP4 exports also dropped the `quant_type` attribute and the fc1/fc2 global scales at input positions 15/16**. That affects GPT-OSS specifically.

**This does not belong in a new-model PR.** It changes the exported graph for GPT-OSS, Qwen3-MoE, Qwen3.5-MoE, PhiMoE and Granite-MoE-Hybrid. Reviewers looking at LFM2 are not the people who will catch a GPT-OSS MXFP4 regression. Could this be split into a standalone fix PR? If not, please record in the description that GPT-OSS (MXFP4), Qwen3-MoE (int4) and PhiMoE exports were regenerated and validated on this branch — the added `test_make_moe_init_selects_qmoe_for_quantized_experts` unit test pins the attribute values but not the resulting graphs.

src/python/py/models/builders/base.py
```diff
@@ -5237,9 +5236,12 @@ def make_qmoe_weights(self, weights):
         use_blockwise_quant = self.ep in supported_blockwise_eps and self.quant_attrs["qmoe_block_size"] > 0
 
         if use_blockwise_quant:
+            # Non-CUDA QMoE ships raw [N, K/pack] weights on ONNX Runtime's MatMulNBits grid (signed block
+            # scales, so no extreme is clipped). The CPU kernel's MLAS Q4 fast path (ORT_USE_MLAS_Q4_GEMM_MOE=1)
+            # re-quantizes to this same grid, which makes it lossless.
             block_size = self.quant_attrs["qmoe_block_size"]
             try:
-                qweight, scales = self._symmetric_blockwise_quantize(weights, block_size)
+                qweight, scales = self._matmulnbits_blockwise_quantize(weights)
```
tianleiwu: Switching non-CUDA blockwise QMoE from `_symmetric_blockwise_quantize` (`signed_scale=False`) to `_matmulnbits_blockwise_quantize` (`signed_scale=True`) is the right call, and I verified both consumers independently:

- **CPU**: `contrib_ops/cpu/moe/moe_quantization_cpu.cc` dequantizes as `(q - 8) * scale` (`default_zp_4bit = 8.0f`), so a negative block scale is handled correctly and the block extreme maps exactly to `qmin` instead of being clipped to 7/8 of its magnitude.
- **WebGPU**: `contrib_ops/webgpu/moe/qmoe.cc` routes both FC1 and FC2 through `ApplyMatMulNBits`, i.e. it literally wants the MatMulNBits layout this now produces. Strict improvement there.

Two concerns:

**TRT-RTX is the one unverified consumer, and the riskiest.** The README change honestly says it has not been re-measured. WebGPU is provably fine (above), but TRT-RTX is the closed kernel where an implicit `scale > 0` assumption would not show up in code review — and this changes the bytes of *every* existing non-CUDA MoE export, not just LFM2-MoE. A single TRT-RTX smoke export plus a logits comparison before merge would de-risk this a lot. If that is not feasible, consider keeping the old encoding for `ep == "trt-rtx"` until it can be measured.

**New hard failure on DML.** Now that `op_type` correctly resolves to `QMoE` (see my comment on `make_moe_init`), an int4 MoE export with `--execution_provider dml` reaches here with the default `qmoe_block_size=32`, misses `supported_blockwise_eps = ["cpu", "cuda", "webgpu", "trt-rtx"]`, and hits the `RuntimeError` a few lines below. The same command previously produced a (numerically wrong, but loadable) float `MoE` model. `dml` is one of only five `builder.py` EP choices, so this is a user-visible break. Either add DML to the supported list, or make the error message say explicitly that int4 MoE is unsupported on DML and suggest `--precision fp16`.

src/python/py/models/builders/lfm2.py
```diff
@@ -19,7 +20,7 @@ def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
 
         # Calculate the dynamic intermediate_size for the MLP.
         intermediate_size = config.intermediate_size
-        if config.block_auto_adjust_ff_dim:
+        if getattr(config, "block_auto_adjust_ff_dim", False):
```
tianleiwu: `transformers` declares `block_auto_adjust_ff_dim: bool = True` in `Lfm2Config` (`configuration_lfm2.py:77`). The `getattr` default of `False` here exists only because `Lfm2MoeConfig` omits the field — but it now also applies to **dense LFM2**, so any Lfm2 config that lacks the key (custom or `trust_remote_code` configs) silently gets the wrong FFN width with no error and no warning.

I would rather see the model class decide than invert an upstream default. For example a `make_intermediate_size_init(config)` hook on `LFM2Model` that `LFM2MoEModel` overrides to `self.intermediate_size = config.intermediate_size`. That also lines up with the model-builder guideline about avoiding customized logic inside constructors and preferring `make_*_init` hooks.

src/python/py/models/builders/lfm2.py
```diff
@@ -160,19 +161,269 @@ def make_layer(self, layer_id, layer):
...
+        self.use_expert_bias = config.use_expert_bias
+        self.routed_scaling_factor = config.routed_scaling_factor
+
+        # Router score for the experts that were not selected. Any finite value far below every
+        # plausible log-sigmoid works: it must fit in fp16 and exp(sentinel - max) must underflow to 0.
+        self.moe_attrs["router_sentinel"] = -10000.0
+        self.moe_attrs["activation_type"] = "swiglu"
+        self.moe_attrs["swiglu_fusion"] = 1
+        self.moe_attrs["normalize_routing_weights"] = True
+        if self.moe_attrs.get("swiglu_limit") is None and self.ep == "trt-rtx":
```
tianleiwu: If TRT-RTX genuinely requires `swiglu_limit` to be present on `QMoE`, then every MoE model targeting TRT-RTX needs this, not just LFM2-MoE — the fix belongs in `make_moe_init` or the `moe_attrs` defaults at `base.py:413`. As written it is an EP quirk hidden inside one model class, so the next MoE model added will hit the same TRT-RTX failure.

If it really is LFM2-specific, could the comment say why?

src/python/py/models/builders/lfm2.py
```diff
@@ -160,19 +161,269 @@ def make_layer(self, layer_id, layer):
...
+            # use +inf to preserve the "no clamp" behavior when the model omits it.
+            self.moe_attrs["swiglu_limit"] = float("inf")
+
+    def make_feed_forward(self, layer_id, layer, root_input):
+        if layer_id < self.num_dense_layers:
+            super().make_feed_forward(layer_id, layer, root_input)
+        else:
+            self.make_moe(layer_id, self.get_feed_forward_module(layer), root_input)
+
+    def make_moe_preprocessing(self, layer_id, moe, root_input):
```
tianleiwu, line 248: This is a near-copy of `Qwen3MoEModel.make_moe_preprocessing` (`qwen.py:804-822`): the same `[E, 2I, H]` -> interleaved `[g0, u0, g1, u1, ...]` repack, the same `make_moe_expert_initializers` call, the same zero gate_up/down biases. The only real differences are the pre-quantized guard (`quant_type is not None` here vs `gate_up_proj is None` there — functionally the same condition) and the `gate_up_bias`/`down_bias` passthrough.

The model-builder guidelines are explicit about reusing existing functionality and leveraging shared base-class code. This looks like a good candidate for a `Model.make_interleaved_swiglu_moe_preprocessing(layer_id, moe)` that both builders call.

Separately: the `quant_type is not None` branch plus the `gate_up_bias`/`down_bias` fallbacks handle packed-QMoE LFM2-MoE checkpoints that **no loader currently produces** — `QuantizedExperts` in `loaders/base.py` is populated only by the Quark and ModelOpt loaders, neither of which supports `Lfm2MoeForCausalLM`. Given the guidance that most generated model-builder code is unnecessary if designed correctly, consider dropping this until a real checkpoint exists (the unit test covering it is testing a path that cannot be reached end to end).

src/python/py/models/builders/lfm2.py
```diff
@@ -160,19 +161,269 @@ def make_layer(self, layer_id, layer):
...
+        indices_name = self.make_moe_router_selection(layer_id, moe, scores_name)
+        self.moe_attrs["router_probs"] = self.make_moe_router_mask(layer_id, scores_name, indices_name)
+
+    def make_moe_router_scores(self, layer_id, moe, root_input):
+        """Emit the per-expert router scores (`sigmoid(logits)`) in fp32 and return their name."""
+        basename = f"/model/layers.{layer_id}/moe/router"
+        logits_shape = self.make_moe_router_shape()
+
+        # Keep the router in floating point so int4 rounding cannot flip an expert choice.
+        moe.gate.exclude_from_quantization = True
```
tianleiwu: Excluding the router `MatMul` from int4 quantization is exactly the right decision — an expert-selection flip is a discrete, unbounded error rather than a bounded rounding error.

The placement is the issue: this mutates the loader's checkpoint IR from inside a method whose own docstring says it "emit[s] the per-expert router scores". Every other producer of this flag is a loader (`loaders/modelopt.py:266-300`). Moving it to `make_moe_preprocessing` (or to the loader) keeps graph emission free of side effects on the checkpoint IR, which matters if a layer is ever emitted more than once or out of order.

src/python/py/models/builders/lfm2.py
```diff
@@ -160,19 +161,269 @@ def make_layer(self, layer_id, layer):
...
+
+        The MoE/QMoE op takes the softmax of the top-k of this tensor, which then equals the
+        model's `sigmoid_i / sum_selected(sigmoid_j)` mixing weights.
+        """
+        basename = f"/model/layers.{layer_id}/moe/router"
+        logits_shape = self.make_moe_router_shape()
+        selected_shape = self.make_moe_router_shape(last_dim=self.moe_attrs["top_k"])
+
+        log_name = f"{basename}/Log"
+        self.make_node("Log", inputs=[scores_name], outputs=[f"{log_name}/output_0"], name=log_name)
```
tianleiwu: Three points on this block.

**1. `Log(Sigmoid(x))` can underflow below the sentinel and silently change routing.** In fp32, `sigmoid(x)` flushes to exactly `0` for `x` below roughly `-104`, so `Log` emits `-inf`. That selected expert then scores *below* the `-10000` sentinel, and the op's own internal `TopK` picks a **different** expert than the graph selected. That is a routing divergence, not a rounding error, and it is silent. Router logits that extreme are implausible, but a one-node guard removes the whole class of bug:

```python
# Clamp before Log so a flushed-to-zero sigmoid can never fall below the sentinel.
clip_name = f"{basename}/Clip"
self.make_clip(
    clip_name,
    [scores_name, f"/model/constants/FLOAT/{1e-30}", ""],
    dtype=ir.DataType.FLOAT,
    shape=logits_shape,
)
```

`make_clip` already exists at `base.py:2105`, and `log(1e-30)` is about `-69`, comfortably above the sentinel.

**2. Raw `make_node` + `make_value` bypasses the wrapper convention.** `Log` here, plus `TopK`, `GatherElements` and `ScatterElements` in the neighbouring methods, are emitted by hand while `base.py` already provides roughly 40 `make_<op>` wrappers (`make_gather`, `make_where`, `make_clip`, `make_softplus`, ...). The guidelines ask to prefer `self.make_<op>` wrappers and to add new methods inside the `Model` class — please add `make_topk`, `make_log`, `make_gather_elements` and `make_scatter_elements` to `Model`.

**3. `Log` runs over the full expert axis.** It is applied to `[rows, num_experts]` and then gathered down to `[rows, top_k]`. Swapping the order (`GatherElements` then `Log`) is arithmetically identical and `num_experts / top_k` cheaper — 8x for LFM2.5-8B-A1B with 32 experts and top-4, per MoE layer, per token.

src/python/py/models/builders/lfm2.py
```diff
@@ -160,19 +161,269 @@ def make_layer(self, layer_id, layer):
...
+        if self.io_dtype != ir.DataType.FLOAT:
+            cast_name = f"{basename}/Cast_1"
+            self.make_cast(cast_name, router_probs_name, self.io_dtype, shape=logits_shape)
+            router_probs_name = f"{cast_name}/output_0"
+        return router_probs_name
+
+    def make_moe_router_shape(self, last_dim=None):
+        return ["batch_size * sequence_length", self.moe_attrs["num_experts"] if last_dim is None else last_dim]
+
+    def make_moe_subgraph(self, layer_id, moe, root_input):
```
tianleiwu: This duplicates `Qwen3MoEModel.make_moe_subgraph` (`qwen.py:839-866`) except for the router-probs source and Qwen's shared-expert combine. The initializer-name construction (`gate_up_proj` / `down_proj` / `.bias` / `.scales` / `weight_type`) is identical, and it is already duplicated a third time in `Model.make_moe_expert_initializers` (`base.py:4943-4949`).

A single `Model.make_moe_expert_names(layer_id)` helper returning that name tuple would collapse all three call sites and remove the risk of the three copies drifting apart when a new QMoE input is added.

src/python/py/models/README.md
```diff
@@ -591,6 +593,8 @@ python -m onnxruntime_genai.models.builder -m model_name -o path_to_output_folde
 python builder.py -m model_name -o path_to_output_folder -p int4 -e execution_provider --extra_options qmoe_block_size=128
 ```
 
+On CPU, WebGPU, and TRT-RTX the block-wise expert weights use ONNX Runtime's MatMulNBits encoding with signed block scales (each block's max-magnitude element maps exactly to the lowest quantized value, so no extreme is clipped). This encoding has been validated end to end against the HF reference on CPU; WebGPU and TRT-RTX consume the same MatMulNBits layout but have not been re-measured with it. ONNX Runtime builds that include [microsoft/onnxruntime#32644](https://github.com/microsoft/onnxruntime/pull/32644) run these experts on the MLAS QNBit GEMM (`MatMulNBits`) kernels directly, which is what makes CPU decode of MoE models such as LFM2.5-8B-A1B usable (roughly 20x faster per token than dequantize-then-GEMM). On older builds the CPU `QMoE` kernel dequantizes to fp32 on every call unless the environment variable `ORT_USE_MLAS_Q4_GEMM_MOE=1` is set at runtime to enable its AVX-512 MLAS Q4 fast path; that fast path re-quantizes to the same grid as this encoding, so enabling it only changes results at fp32 rounding level.
```
tianleiwu: Being explicit about which EPs the new encoding was actually measured on is the right kind of honesty for a doc change — thanks for that.

The paragraph is hard to read though: it packs the encoding contract, an unmerged-PR dependency, a "roughly 20x faster per token" benchmark claim, and an otherwise-undocumented ORT environment variable into one ~8-line sentence chain. Suggest splitting into three: (a) the encoding contract, (b) a note on `ORT_USE_MLAS_Q4_GEMM_MOE`, (c) the microsoft/onnxruntime#32644 dependency. The 20x figure is also unverifiable until #32644 merges — worth attributing it to specific hardware/model, or softening it.

tianleiwu (commented):
## Review Summary

The LFM2-MoE feature itself is solid. The `log(sigmoid) + sentinel` trick to express LFM2's sigmoid router through the softmax-only fused `MoE`/`QMoE` op is mathematically exact, and I verified it line by line against `transformers`' `Lfm2MoeSparseMoeBlock.route_tokens_to_experts`: selection on `sigmoid + expert_bias`, mixing on the unbiased `sigmoid`, renormalized over the selected experts. Moving `routed_scaling_factor` from the routing weights to a `Mul` on the op output is algebraically equivalent (expert mixing is linear in the weights) and cheaper. Splitting `make_moe_router` into four small emitters addresses the earlier "one large method" feedback well, and the new 358-line test file pins the exact node sequence, wiring, attributes and sentinel representability rather than just asserting "it builds".

My concerns are almost entirely about the **two cross-cutting MoE changes bundled with the new model**, not about LFM2 itself.

### Verification performed on `5ee16167`

| Check | Result |
|---|---|
| `pytest test/python/builder` | 729 passed, 4 skipped |
| Router math vs `transformers` `Lfm2MoeSparseMoeBlock` | matches |
| `gate_up_proj` interleave vs `Qwen2MoeExperts` (`chunk(2, dim=-1)`) | correct |
| `Lfm2MoeConfig` attribute availability | all builder-read fields present |
| `moe_attrs` readers/writers after rename | consistent, no stale keys in `src/` |
| CPU QMoE dequant (`contrib_ops/cpu/moe/moe_quantization_cpu.cc`) | `(q - 8) * scale` — signed scale is safe |
| WebGPU QMoE (`contrib_ops/webgpu/moe/qmoe.cc`) | routes through `ApplyMatMulNBits` — MatMulNBits layout is exactly what it wants |
| Origin of the `moe_attrs` key bug | introduced by #2386 (`git log -L 1030,1040:src/python/py/models/builders/base.py`) |

### Top concerns (details inline)

1. **The `make_moe_init` key fix is broader than the description says** and should probably be its own PR — see the comment on `base.py`.
2. **The signed-scale QMoE switch changes exported bytes for every non-CUDA MoE model**, and TRT-RTX is unverified.
3. **int4 MoE export on the DML EP now raises** where it previously (incorrectly) succeeded.

### Findings outside the diff

- `CudaQuantizer.symmetric_blockwise_quantize` (`src/python/py/models/quantization/cuda_quantizer.py:623`) now has **zero callers**. The only surviving references are the two `loaders/quark.py` docstrings this PR rewrites to point at it. Please delete it, or add a note on why it is retained.
- The non-CUDA blockwise branch does not validate `block_size`, while the CUDA branch rejects anything outside `(32, 64, 128)`. Now that both paths share `CudaQuantizer.matmulnbits_blockwise_quantize`, the check could be hoisted.

### Minor notes

- The `LFM2MoEModel` docstring names the router `Lfm2MoeTopKRouter`; upstream the class is `Lfm2MoeSparseMoeBlock` (with `route_tokens_to_experts`). Worth correcting so the reference is greppable.
- HF normalizes with `routing_weights / (sum + 1e-6)`; the exported graph normalizes without the epsilon. The delta is negligible, but a one-line comment would document the known parity gap.
- `moe_attrs["router_sentinel"]` and `moe_attrs["router_probs"]` are not documented in the `moe_attrs` declaration block (`base.py:413`), which the model-builder guidelines ask for. `router_probs` is also not a *setting* — it is a per-layer graph value name threaded from `make_moe_router` to `make_moe_subgraph`. Consider returning it like `Qwen3MoEModel.make_moe_subgraph` does with its combine output.

### Stale threads

The three open threads from the automated reviewer were filed against `e720558c` and look fully addressed by `de40a22f` / `5ee16167` (`get_feed_forward_module`, `moe_attrs["router_sentinel"]`, and the four-way router split).

src/python/py/models/builders/base.py
ykhrustalev: Agreed, this should not ride along with a new model. Split it out into #2581, which carries the `op_type` / `quant_type` key fix, the MXFP4/NVFP4 global-scale restoration, and the unit tests pinning each scheme. That PR also has before/after GPT-OSS int4 and mxfp4 export results in the description. Once #2581 lands I will rebase this branch onto it and drop the `base.py` change from here.

src/python/py/models/builders/base.py
ykhrustalev: The DML part is handled in #2581 alongside the key fix: the unsupported-EP error in `make_qmoe_weights` now names the EP and suggests `--precision fp16`, with a unit test for the `dml` path. TRT-RTX verification is still outstanding here.

src/python/py/models/builders/lfm2.py
ykhrustalev: Done in 7ae43a09: `LFM2Model.make_intermediate_size_init` keeps the strict `config.block_auto_adjust_ff_dim` read, and `LFM2MoEModel` overrides it to use `intermediate_size` as is. Tests cover both, including the AttributeError when the key is missing.

src/python/py/models/builders/lfm2.py
ykhrustalev: Hoisted into `Model.make_moe_init` in f763e123, so every MoE model gets it on TRT-RTX. Removed the copies from the LFM2-MoE and Qwen3.5-MoE constructors and added a base-level test.

src/python/py/models/builders/lfm2.py
ykhrustalev: Done in f763e123: added `Model.make_interleaved_swiglu_moe_preprocessing`, and both LFM2-MoE and Qwen3.5-MoE now call it. Dropped the LFM2-specific packed-bias passthrough and its tests until a loader produces packed LFM2-MoE experts.

src/python/py/models/builders/lfm2.py
ykhrustalev: Moved to `make_moe_preprocessing` in f763e123. The score emitter no longer touches the checkpoint IR.

src/python/py/models/builders/lfm2.py
ykhrustalev: All three done in f763e123: GatherElements now runs first, then Clip(1e-30) before Log, so a flushed-to-zero sigmoid cannot fall below the sentinel and Log only touches `top_k` entries. Added `make_topk`, `make_log`, `make_gather_elements` and `make_scatter_elements` to `Model` and use them here. Verified a tiny random LFM2-MoE fp32 export against transformers 5.17 in ORT 1.30: max logit diff 2e-7.

src/python/py/models/builders/lfm2.py
ykhrustalev: Added `Model.make_moe_expert_names` in f763e123. `make_moe_expert_initializers`, the Qwen3.5-MoE subgraph and the LFM2-MoE subgraph all use it now. Also `make_moe_router` returns the masked scores and `make_moe` threads them into `make_moe_subgraph` instead of going through `moe_attrs`.

src/python/py/models/README.md
ykhrustalev: Split into three paragraphs in f763e123: encoding contract, the `ORT_USE_MLAS_Q4_GEMM_MOE` note, and the onnxruntime#32644 dependency. Replaced the 20x figure with the measured per-token decode times for LFM2.5-8B-A1B int4 on a 30-core x86 host, and the TRT-RTX sentence now reflects that it keeps the original encoding.

src/python/py/models/builders/base.py
ykhrustalev: TRT-RTX now keeps the original positive-scale encoding in f763e123 until someone can run the parity check on that EP; CPU and WebGPU stay on the signed-scale grid. That also gives `symmetric_blockwise_quantize` its one caller back, and the 32/64/128 block-size check is hoisted so CPU and WebGPU validate it too.

ykhrustalev:
Addressed the summary-level items in f763e123 as well: `symmetric_blockwise_quantize` is no longer dead (TRT-RTX uses it), block-size validation is shared, the docstring names `Lfm2MoeSparseMoeBlock.route_tokens_to_experts`, the missing 1e-6 normalization epsilon is noted in `make_moe_router_mask`, and `router_sentinel` is documented in the `moe_attrs` block while `router_probs` is returned instead of stored. The `make_moe_init` key fix lives in #2581; this branch keeps its copy until that merges.

src/python/py/models/builders/lfm2.py
```diff
@@ -160,19 +163,251 @@
...
+        if self.io_dtype != ir.DataType.FLOAT:
+            cast_name = f"{basename}/Cast_1"
+            self.make_cast(cast_name, router_probs_name, self.io_dtype, shape=logits_shape)
+            router_probs_name = f"{cast_name}/output_0"
+        return router_probs_name
+
+    def make_moe_router_shape(self, last_dim=None):
+        return ["batch_size * sequence_length", self.moe_attrs["num_experts"] if last_dim is None else last_dim]
+
+    def make_moe_subgraph(self, layer_id, moe, root_input, router_probs):
```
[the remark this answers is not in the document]
ykhrustalev: Fixed in 161f6032: `router_probs` is now an optional parameter on both `Model.make_moe_subgraph` (documented there) and the LFM2-MoE override, which raises when it is missing. The other overrides keep their four-argument form.

src/python/py/models/builders/lfm2.py
tianleiwu: The Gather-before-Log change, wrappers, and clamp address the three original implementation points. However, executing the current graph shows that the normalization gap mentioned in the earlier review summary is not always negligible, so I am keeping this routing discussion open for the remaining correctness issue.

At `161f6032`, `make_moe_router_mask` says dropping HF's `sum + 1e-6` is below dtype rounding. I exported the actual FP32 MoE subgraph using the builder's real emitters, loaded the ONNX file in CPU ORT 1.28.0, and compared with a seeded tiny `Lfm2MoeForCausalLM` in Transformers 5.15.0. Give the first four experts logits `[-16, -16, -16, -16]` and expert bias `+2`, with the other logits `[-10, -11, -12, -13]` and bias zero. Both implementations select the same four experts, with nonzero sigmoid values and no clipping-floor involvement. HF's routing weights sum to **0.310411692**, but the fused op forces a sum of **1**: output norms are **5.92604e-5** versus **1.90909e-4**, approximately **3.22x** too large. Multiplying the exported output by the omitted factor restores parity in this case. The ordinary-score control agrees to **1.31e-10 max absolute error**.

For selected logits `[-110, -111, -112, -113]`, HF returns exactly zero, while the clamp turns the export into a nonzero uniform expert mixture. The clamp protects expert selection, but does not preserve the selected scores' total mass.

Please retain `sum_selected = ReduceSum(selected_scores)` **before** the clamp and apply the per-token factor `sum_selected / (sum_selected + 1e-6)` to the MoE result, composed with `routed_scaling_factor`, with a zero-sum case covered. Add numerical coverage that executes the emitted graph; the recording tests cannot detect this. These are controlled edge cases, not a claim about their frequency on the released checkpoints.

src/python/py/models/builders/base.py
tianleiwu: Accepting the split into #2581 and resolving this scope discussion. I verified that #2581 is open and that the key fix still appears in this head, so the planned rebase after #2581 lands remains necessary. The improved unsupported-DML diagnostic can travel with that shared fix as discussed; this follow-up review does not independently validate #2581.

src/python/py/models/builders/lfm2.py
```diff
@@ -160,19 +163,253 @@ def make_layer(self, layer_id, layer):
...
+    def __init__(self, config, io_dtype, onnx_dtype, ep, cache_dir, extra_options):
+        super().__init__(config, io_dtype, onnx_dtype, ep, cache_dir, extra_options)
+
+        if not config.norm_topk_prob:
+            raise NotImplementedError(
+                "LFM2-MoE export requires norm_topk_prob=True: the fused MoE op always normalizes the "
+                "routing weights over the selected experts."
+            )
+
+        self.num_dense_layers = config.num_dense_layers
```
tianleiwu: Nonblocking builder-convention suggestion: the new `num_dense_layers`, `use_expert_bias`, and `routed_scaling_factor` are scenario settings, but they are stored as separate instance fields here. The applicable model-builder instructions require groups of new settings in documented `_attrs` entries and favor `make_*_init` hooks for initialization. Could these settings live in `moe_attrs` with their purpose/usage documented alongside `router_sentinel`? Existing dimension fields such as `moe_intermediate_size` can keep their established convention.

src/python/py/models/builders/lfm2.py
```diff
@@ -160,19 +163,253 @@ def make_layer(self, layer_id, layer):
...
+        )
+        router_probs_name = f"{scatter_name}/output_0"
+
+        if self.io_dtype != ir.DataType.FLOAT:
+            cast_name = f"{basename}/Cast_1"
+            self.make_cast(cast_name, router_probs_name, self.io_dtype, shape=logits_shape)
+            router_probs_name = f"{cast_name}/output_0"
+        return router_probs_name
+
+    def make_moe_router_shape(self, last_dim=None):
```
tianleiwu: Nonblocking builder-convention suggestion: this is a new generic router-shape helper, but it exists only on `LFM2MoEModel`. The repository instructions explicitly ask for new shared helpers on `Model`, with architecture classes overriding behavior as needed. Please move this helper to the base class and check the other new non-override helpers against that rule; the shared expert-preprocessing and operator wrappers added in this revision already follow that pattern.

tianleiwu (changes requested):
Reviewed current head `161f6032`. Most earlier concerns are addressed. The complete Python builder suite passes: **738 passed, 4 skipped**; the focused LFM2-MoE/QMoE subset passes **87 tests, 3 skipped**. `git diff --check` also passes.

One correctness issue remains: the omitted HF routing-normalization epsilon is not uniformly a rounding-level difference. Executing a real builder-generated FP32 MoE graph against a tiny Transformers LFM2-MoE model demonstrates a material output-magnitude error for small selected scores and a nonzero output when the reference returns zero. Numerical evidence and the proposed correction are on the existing router thread, rather than duplicated as a new inline thread. Please preserve that normalization factor and cover it with an executed-graph parity test. The earlier review summary's characterization of this gap as negligible was too broad.

The two new inline notes are nonblocking model-builder convention suggestions. I accept separating the shared MoE key fix into #2581; that PR remains open, so this branch still needs the planned rebase after it lands. The prior fixed/accepted discussions have been resolved.

Local execution used CPU ORT 1.28.0 and Transformers 5.15.0 with controlled inputs, not full pretrained-checkpoint evaluation. Native tests were not rebuilt at this head, and CUDA/WebGPU/TRT-RTX execution was not revalidated.

src/python/py/models/builders/lfm2.py
ykhrustalev: You are right, and thanks for executing the graph. Fixed in 6d87ef39 the way you proposed: `make_moe_router_selected_scores` gathers the unbiased selected scores, `make_moe_router_output_scale` takes `ReduceSum` over them **before** the clamp and emits `sum / (sum + 1e-6)` with `routed_scaling_factor` folded in, reshaped to the MoE output's leading dims, and `make_moe_subgraph` multiplies it into the op output. A flushed-to-zero token therefore yields a zero factor and a zero output, matching HF.

New executed-graph test `test_lfm2_moe_executed_graph_matches_hf_routing` builds the layer with the real emitters, runs it in ORT and compares with HF's routing math for your three cases: ordinary scores, four selected logits at -16 with bias +2 (HF mass 0.31), and -110..-113 (HF returns zero), with and without `routed_scaling_factor`.

One more finding from executing it: ONNX Runtime's CPU `Sigmoid` kernel carries an absolute error of about 6e-8. It returns 5.96e-8 for x = -16 (exact 1.125e-7) and 0 for x <= -17, so the mass factor came out as 0.19 instead of 0.31. Since the scores are logged and summed, the router now emits the sigmoid as `Reciprocal(1 + Exp(Neg(x)))`, which matches `torch.sigmoid` to fp32 rounding on [-4, -30] and flushes to zero at the same point torch does. The tiny-model export still matches transformers 5.17 to 1.8e-7.

src/python/py/models/builders/lfm2.py
ykhrustalev: Done in 6d87ef39: `num_dense_layers`, `use_expert_bias` and `routed_scaling_factor` now live in `moe_attrs` and are documented in the declaration block in `base.py` next to `router_sentinel`. `moe_intermediate_size` stays a dimension field.

src/python/py/models/builders/lfm2.py
ykhrustalev: Moved `make_moe_router_shape` onto `Model` in 6d87ef39, with a test. I checked the remaining new methods on `LFM2MoEModel`: the router emitters (`make_moe_router_scores`, `_selection`, `_selected_scores`, `_output_scale`, `_mask`) encode LFM2's sigmoid-plus-bias routing rather than generic behavior, so they stay as the model's override of `make_moe_router`. The new generic pieces (`make_exp`, `make_neg`, `make_reciprocal`) went onto `Model`.

src/python/py/models/builders/lfm2.py
ykhrustalev: Real-checkpoint check of 6d87ef39 on LFM2.5-8B-A1B (transformers 5.17, ORT 1.31, chat-formatted 81-token prompt, KL(hf||onnx) over the vocab):

| Export | EP | top-1 agree | KL mean / max | ppl hf / onnx |
|---|---|---|---|---|
| fp32 | CPU | 1.000 | 0.0000 / 0.0000 | 20.523 / 20.523 |
| int4 (signed-scale QMoE, default kernel) | CPU | 0.889 | 0.0991 / 1.10 | 20.523 / 21.596 |
| int4, `ORT_USE_MLAS_Q4_GEMM_MOE=1` | CPU | 0.889 | 0.0991 / 1.10 | 20.523 / 21.596 |

fp32 is exact, and the int4 numbers are unchanged from the previous head (0.0982 mean) since quantization noise dominates there; the MLAS fast path being bit-identical to the default kernel confirms the signed-scale grid is lossless for it. On the CUDA EP a tiny random Lfm2Moe (fp16 MoE and int4 QMoE) runs the new router graph with KL 0.000000 and 0.0011 respectively. Native `unit_tests` (LFM2 filter) pass at this head.

src/python/py/models/builders/base.py
```diff
@@ -5204,50 +5299,61 @@ def make_qmoe_weights(self, weights):
...
-
-        raise RuntimeError(f"Please use a supported EP ({', '.join(supported_blockwise_eps)}) "
-                           "for QMoE expert weights quantization. "
-                           f"Got qmoe_block_size={self.quant_attrs['qmoe_block_size']} and ep={self.ep}.")
+        # CPU and WebGPU QMoE ship raw [N, K/pack] weights on ONNX Runtime's MatMulNBits grid (signed block
+        # scales, so no extreme is clipped). The CPU kernel's MLAS Q4 fast path (ORT_USE_MLAS_Q4_GEMM_MOE=1)
+        # re-quantizes to this same grid, which makes it lossless, and the WebGPU kernel consumes the
+        # MatMulNBits layout directly.
+        try:
+            qweight, scales = self._matmulnbits_blockwise_quantize(weights)
```
tianleiwu, line 5517: **The CPU/WebGPU qweight last dim changes when `K % block_size != 0`, and ORT will reject it.**

`_matmulnbits_blockwise_quantize` pads `K` up to `num_blocks * block_size` and returns storage of shape `[N, num_blocks * ceil(block_size/pack)]` (`cuda_quantizer.py`: `padded_k = num_blocks * block_size`, `blob_size = (block_size + pack - 1) // pack`, then `reshape(n, -1)`).

The path this replaces, `_symmetric_blockwise_quantize`, trimmed the padding back off and returned `[N, ceil(K/pack)]`:

```python
quantized_flat = quantized_flat[..., :-pad_size]
packed_shape[-1] = (original_shape[-1] + 1) // 2
```

ORT's QMoE shape validation wants the unpadded form (`contrib_ops/cpu/moe/moe_helper.h`):

```cpp
CHECK_SHAPE(fc1_experts_weights_shape, "fc1_experts_weights", num_experts, fc1_inter_size, hidden_size / pack_size);
CHECK_SHAPE(fc2_experts_weights_shape, "fc2_experts_weights", num_experts, hidden_size, inter_size / pack_size);
```

So a CPU/WebGPU int4 MoE export whose `hidden_size` or `moe_intermediate_size` is not a multiple of `qmoe_block_size` now fails at session creation with a shape mismatch, where before it loaded. (The scales are unaffected — the kernel computes `blocks_per_row` with `ceil`.) LFM2.5-8B-A1B happens to be divisible, so the parity runs would not surface this.

Suggested fix, either:
- trim the trailing padding here so the CPU/WebGPU storage keeps the `[N, ceil(K/pack)]` contract, or
- raise an explicit builder-time error naming the offending dimension and the block size, instead of letting it fail at load.

The same padding behavior already applies to the CUDA `weights_prepacked=0` branch above, so whichever guard you add is worth sharing between them.

src/python/py/models/builders/base.py
```diff
@@ -5204,50 +5299,61 @@ def make_qmoe_weights(self, weights):
...
+            # grid used below has not been measured against the TRT-RTX kernel yet, so the EP stays on
+            # the encoding it shipped with until that parity run exists.
+            try:
+                qweight, scales = self._symmetric_blockwise_quantize(weights, block_size)
+                self.moe_attrs["block_size"] = block_size
+                return qweight, scales.to(torch.float16)
+            except Exception as e:
+                raise RuntimeError(f"Block-wise quantization failed with block_size={block_size}: {e}") from e
+
+        if block_size not in (32, 64, 128):
```
tianleiwu, line 5486: Hoisting this check out of the CUDA branch means CPU and WebGPU now reject `qmoe_block_size=16` and `qmoe_block_size=256`, which they previously accepted through `_symmetric_blockwise_quantize` (it takes any positive block size and pads).

That is a defensible tightening, but it is a user-visible break on two documented values and neither doc was updated:

- `src/python/py/models/README.md:586` — "CUDA block-wise QMoE supports only `32`, `64`, or `128`"
- `src/python/py/models/builder.py:702` — "`qmoe_block_size = <=0/16/32/64/128/256` ... CUDA block-wise QMoE supports 32/64/128 only."

The model-builder instructions require `extra_options` constraints to be documented consistently in both files. Please update both to say the 32/64/128 restriction applies to CPU, CUDA and WebGPU, and that TRT-RTX is the only EP still accepting other block sizes.

src/python/py/models/builders/base.py
```diff
@@ -5007,7 +5096,13 @@ def make_moe_expert_initializers(self, layer_id, experts, gate_up_weight=None, d
     def make_moe_router(self, layer_id, moe, root_input):
         raise NotImplementedError("MoE router construction must be implemented by the model class.")
 
-    def make_moe_subgraph(self, layer_id, moe, root_input):
+    def make_moe_router_shape(self, last_dim=None):
+        """Shape of a per-token router tensor: one row per token, `num_experts` (or `last_dim`) columns."""
+        return ["batch_size * sequence_length", self.moe_attrs["num_experts"] if last_dim is None else last_dim]
```
tianleiwu: Nonblocking follow-up now that this helper lives on `Model`: the row dim is hardcoded to `"batch_size * sequence_length"`, which is wrong for a paged-attention export, where `make_hidden_state_shape` produces `["num_tokens", hidden_size]`. The router tensor would then declare a different symbolic first dim than the MoE op's input.

Deriving it keeps the two consistent:

```python
def make_moe_router_shape(self, last_dim=None):
    rows = "num_tokens" if self.use_paged_attention else "batch_size * sequence_length"
    return [rows, self.moe_attrs["num_experts"] if last_dim is None else last_dim]
```

Worth doing here rather than later since `Qwen3MoEModel.make_moe_router` still spells the literal out inline and should be switched over to this helper too.

src/python/py/models/builders/lfm2.py
```diff
@@ -160,19 +163,350 @@ def make_layer(self, layer_id, layer):
...
+
+        if self.io_dtype != ir.DataType.FLOAT:
+            cast_name = f"{basename}/Cast_1"
+            self.make_cast(cast_name, router_probs_name, self.io_dtype, shape=logits_shape)
+            router_probs_name = f"{cast_name}/output_0"
+        return router_probs_name
+
+    def make_moe_subgraph(self, layer_id, moe, root_input, router_probs=None, output_scale=None):
+        if router_probs is None or output_scale is None:
+            raise ValueError("LFM2-MoE needs the masked router scores and output scale returned by make_moe_router.")
```
tianleiwu: Nitpick: this guard covers a state that cannot occur. `LFM2MoEModel.make_moe` is the only caller and always passes both values straight from `make_moe_router`, which cannot return `None`. The defaults exist only to keep the base-class signature CodeQL-compatible.

Suggest dropping the check (and `test_lfm2_moe_subgraph_requires_router_outputs`, which exists solely to exercise it) — that removes two arguments' worth of defensive plumbing without losing any coverage.

tianleiwu (commented):
Re-reviewed at `519a6ad6`.

The blocking issue from my previous round is resolved: `make_moe_router_selected_scores` / `make_moe_router_output_scale` now apply `sum_selected / (sum_selected + 1e-6)` with `routed_scaling_factor` folded in, the sum is taken before the clamp so a flushed-to-zero token yields exactly zero, and `test_lfm2_moe_executed_graph_matches_hf_routing` executes the emitted graph in ORT against the HF routing math for the ordinary, tiny-mass, fully-flushed, partly-flushed and exactly-epsilon cases. The spelled-out `Reciprocal(1 + Exp(Neg(x)))` sigmoid is the right call given the CPU `Sigmoid` kernel's absolute-error profile. I have resolved that thread and the two convention threads.

What is left is all in the CPU/WebGPU quantization switch, which is the part of this PR with the widest blast radius (it changes the bytes of every non-CUDA MoE export, not just LFM2-MoE). Two of them I would fix before merge:

1. **qweight last dim now keeps the block padding.** `_matmulnbits_blockwise_quantize` returns `[N, num_blocks * ceil(block_size/pack)]`; `_symmetric_blockwise_quantize` trimmed it to `[N, ceil(K/pack)]`. ORT's `moe_helper.h` requires the unpadded shape, so any MoE model whose `hidden_size` or `moe_intermediate_size` is not a multiple of the block size now produces a model the QMoE op rejects at session creation. Inline comment with details.
2. **`qmoe_block_size` 16 and 256 now fail on CPU/WebGPU**, but `README.md:586` and `builder.py:702` still document the 32/64/128 restriction as CUDA-only. The builder instructions require option constraints to be consistent across both files.

One observation worth capturing rather than fixing: the switch also flips the **INT8** byte encoding from two's complement to `q + 128`. ORT CPU uses `default_zp_8bit = 128.0f`, so the new encoding is the correct one and the previous INT8 CPU/WebGPU export was off by 128 on every weight — a second real bug fix hiding in this PR. Only INT4 is measured and tested today; please add INT8 coverage and mention it in the description so it is not lost.

Two nitpicks are inline (an unreachable `ValueError` guard, and `make_moe_router_shape` hardcoding the non-paged row dim). Also, `test_lfm2_moe.py` re-implements the builder-module import shim already in `test_qmoe_weights.py`, and `_moe_attrs()` hardcodes a snapshot of the `moe_attrs` defaults that will drift from `base.py` — worth deriving from the real class.

Nothing else blocking. The dedup into `make_interleaved_swiglu_moe_preprocessing` / `make_moe_expert_names`, holding TRT-RTX on the old encoding, and excluding the router `MatMul` from int4 all look right, and the `op_type` regression fix is valuable well beyond this model.

src/python/py/models/builders/lfm2.py
tianleiwu: Confirmed fixed at `519a6ad6`. `make_moe_router_selected_scores` gathers the unbiased selected scores, `make_moe_router_output_scale` takes the `ReduceSum` before the clamp and emits `sum / (sum + 1e-6)` with `routed_scaling_factor` folded in, and `make_moe_subgraph` multiplies it into the op output, so a fully flushed token yields exactly zero like HF. `test_lfm2_moe_executed_graph_matches_hf_routing` executes the emitted graph in ORT and pins the 3.22x and 2.0x mass ratios plus the zero-output case, which is the coverage the recording tests could not provide. The spelled-out `Reciprocal(1 + Exp(Neg(x)))` sigmoid is the right response to the CPU kernel error you measured. Resolving.

src/python/py/models/builders/base.py
tianleiwu: Confirmed fixed at `93aa6da0`. CPU/raw storage is trimmed to K/pack, scales retain ceil(K/block_size), and odd INT4 K is rejected. My earlier suggestion to trim WebGPU storage alone was incomplete: its MatMulNBits kernel indexes whole-block strides. The latest commit correctly rejects partial blocks on WebGPU instead. The new executed CPU partial-block tests pass for INT4 and INT8 in both projections. Resolving.

src/python/py/models/builders/base.py
tianleiwu: Confirmed at `93aa6da0`: README and CLI help both document the CPU/CUDA/WebGPU 32/64/128 restriction, TRT-RTX 16/256 support, and the new dimension constraints. Resolving.

src/python/py/models/builders/base.py
tianleiwu: Confirmed fixed at `93aa6da0`: `make_moe_router_shape` selects `num_tokens` for paged attention, Qwen3.5 now uses the helper, and tests cover both token layouts. Resolving.

src/python/py/models/builders/lfm2.py
tianleiwu: Confirmed fixed at `93aa6da0`: the unreachable guard and its dedicated test are removed; the normal router-to-subgraph flow remains covered. Resolving.

tianleiwu (dismissed):
Re-reviewed at `93aa6da0`. The four open concerns from the previous round are addressed, and I have resolved those threads. No remaining actionable findings.

The final padding fix retains unpadded CPU/raw QMoE weights, rejects odd raw INT4 dimensions, and explicitly rejects partial blocks on WebGPU, whose MatMulNBits kernels require whole-block strides. The new executed CPU tests cover partial blocks in both projections for INT4 and INT8. README and CLI help now agree on these constraints. The paged router shape and unreachable guard comments are also addressed.

Validation on this head: **768 Python builder tests passed, 4 skipped**, including the new executed partial-block cases; `git diff --check` passed. An additional CPU probe of the retained trimming implementation with ORT 1.30.0 agreed with an independently dequantized expert reference to **3.73e-8 max absolute error** for aligned and partial-block INT4/INT8 cases. Native tests were not rebuilt, and CUDA, WebGPU, TRT-RTX, and full pretrained-checkpoint inference were not run in this round.

The accepted split/rebase arrangement for the shared MoE key fix in #2581 still applies; that PR remains open. This approval supersedes my earlier changes-requested review.

tianleiwu:
@ykhrustalev, please resolve conflicts

tianleiwu (approved):

PR #2580, opened 2026-09-16
Make FixedStatePool capabilities explicit for generic devices

aciddelgado:
## Summary
 
 Replace FixedStatePool's broad CPU/CUDA assumptions with the narrow device capabilities its execution paths actually require.
 
 FixedStatePool can bind contiguous resident state directly by creating tensor views at nonzero offsets into persistent storage. This works for CPU and CUDA, where storage is represented by addressable pointers, but not for devices such as WebGPU that use opaque buffer handles.
 
 This PR adds an explicit offset-view capability and uses FixedStatePool's existing gather/staging path when direct views are unavailable. It also permits generic devices to use ordinary, non-speculative fixed-state execution while retaining the CPU/CUDA restriction for compact partial-accept replay.
 
 This does not enable WebGPU FixedStatePool or continuous batching yet. It removes two provider-specific assumptions needed by a later enablement change.
 
 ## Changes
 
 - Add `DeviceInterface::SupportsOffsetTensorViews()` with a conservative default of `false`.
 - Explicitly enable offset tensor views for CPU and CUDA.
 - Explicitly disable offset tensor views for WebGPU.
 - Gate FixedStatePool's direct persistent-bank binding optimization on that capability.
 - Use the existing gathered-input/output-staging path when offset views are unsupported.
 - Replace the broad CPU/CUDA construction whitelist with a replay-specific restriction:
   - ordinary fixed-state execution is allowed for generic devices;
   - configurations that enable compact state-update replay remain restricted to CPU/CUDA.
 - Add CPU-only test devices and coverage for direct bindings, generic staging, persistence, and replay rejection.
 
 ## Why the ordinary path is device-generic
 
 When direct offset views are unavailable and compact state updates are disabled, FixedStatePool requires only existing `DeviceBuffer`/`DeviceInterface` operations:
 
 - allocator-backed persistent banks, zero rows, and staging tensors;
 - zero initialization through `DeviceBuffer::Zero()`;
 - byte-range copies through `DeviceBuffer::CopyFrom()` for gather and scatter;
 - zero-offset model input/output views over staging storage;
 - `DeviceInterface::Synchronize()` before publishing bank changes or releasing temporary storage.
 
 No provider-specific kernels or pointer arithmetic are required on this path.
 
 ## Compact replay protection
 
 Compact partial-accept replay still constructs raw row addresses into persistent banks and calls `ReplayStateUpdates`. Because there is not yet a provider-neutral capability for that operation, a non-CPU/CUDA device is rejected when the model enables compact state updates with nonzero capacity.
 
 Models without compact state-update support can construct and use FixedStatePool through the generic staging path.
 
 ## ABI impact
 
 This appends a virtual method to `DeviceInterface`, which is part of the dynamically loaded device add-on boundary. `kDeviceInterfaceVersion` is incremented from 5 to 6 so mismatched host and CUDA add-on binaries fail version validation instead of calling through incompatible vtables.
 
 ## Behavior
 
 - CPU and CUDA behavior is unchanged: eligible resident intervals still use direct persistent-bank bindings.
 - Devices without offset-view support use gathered inputs and staged outputs.
 - Generic devices can use ordinary fixed-state execution when compact state updates are unavailable or disabled.
 - Replay-capable fixed-state configurations remain restricted to CPU/CUDA.
 
 ## Out of scope
 
 - Enabling WebGPU FixedStatePool or continuous batching end to end
 - Implementing WebGPU-specific fixed-state operations
 - Generalizing compact partial-accept replay
 - DFlash2 or MTP
 - PagedAttention
 - Graph capture
 - Model-builder changes
 
 ## Testing
 
 - Built `engine_unit_tests` with MSVC.
 - Ran `FixedStatePoolTest.*`: 39/39 tests passed.
 - Verified direct and staged paths preserve identical logical state.
 - Verified a CPU-backed device reporting a non-CPU/CUDA type can reserve, stage, commit, and reload ordinary fixed state.
 - Verified the same generic device is rejected for a compact-replay-capable configuration.
 - Verified `git diff --check` passes.


baijumeswani:
Recommendation: avoid replacing the current CPU/CUDA restriction with unrestricted generic-device admission. Instead, add a conservative capability such as `SupportsTransactionalFixedState()` (default `false`) and require it before constructing `FixedStatePool` on a device. Enable it per EP only after its ranged copies, zero/fill operations, and `Synchronize()` semantics are qualified to guarantee that all fixed-state device work is complete before the Engine publishes a bank flip. This preserves provider-neutral admission without allowing an EP with asynchronous copies and a no-op synchronization implementation to silently consume or publish stale state. The model manifest can establish that the graph supports Engine fixed state, while the device capability independently establishes that the runtime can execute the transactional state contract.

test/python/test_onnxruntime_genai_hybrid_engine.py
baijumeswani, line 1: It seems like this file is not wired into the webgpu ci pipeline. So it does not run in the ci. I think we can merge this in, and address this in a follow up.

baijumeswani (approved):

PR #2581, opened 2026-09-16
fix: restore the moe_attrs keys the MoE emitters read

ykhrustalev:
**Problem**

`make_moe_init` stored the resolved MoE quantization scheme under `moe_op_type` / `qmoe_quant_type`, but every reader uses `op_type` / `quant_type`. Because `moe_attrs` already seeds those keys with `"MoE"` and `"int"`, the mismatch never raised — the scheme was silently discarded on every quantized MoE export since #2386. All MoE models are affected: GPT-OSS, Qwen3-MoE, Qwen3.5-MoE, PhiMoE, Granite-MoE-Hybrid.

**Solution**

- Writes `op_type` / `quant_type`, so int4 and int8 exports emit `QMoE` instead of the float `MoE` op.
- Restores `quant_type` and the fc1/fc2 global scale inputs for MXFP4 and NVFP4.
- Rewords the unsupported-EP error in `make_qmoe_weights` to name the EP and suggest `--precision fp16`, now that int4 MoE on `dml` reaches it instead of falling through to an unloadable float graph.

**Testing**

- Exported a small GPT-OSS model (2 layers, 4 experts) before and after: int4 goes from `MoE`/8 inputs to `QMoE`/14 inputs with `block_size=32`; mxfp4 goes from `MoE`/8 inputs to `QMoE`/17 inputs with `quant_type='fp4'` and the global scales at inputs 15 and 16.
- New unit tests pin `(op_type, quant_type, expert_weight_bits)` per scheme and the `dml` error; `test/python/builder` passes (702 passed, 3 skipped).

Refs #2575


ykhrustalev:
Closing this. The key fix stays in #2575, where it was originally reviewed.

PR #2594, opened 2026-09-18
Load model builders and Transformers architectures lazily

xieofxie:
## Summary

- Import builder classes only after selecting the model architecture, and resolve `builders` package exports lazily.
- Resolve only the selected Transformers model class when loading weights, so unrelated newer architectures do not force a Transformers upgrade. Keep shared `AutoTokenizer` and `GenerationConfig` imports at module scope.
- Preserve incoming LFM2-VL support with lazy dispatch, and update affected tests and design documentation.

## Validation

- 155 targeted lazy-import, configuration, and model-loading tests passed with Transformers 4.57.6 before merging main.
- After resolving the merge, all 35 lazy-import and LFM2-VL regression tests passed.
- Regression coverage checks every public builder export, disallows module-level architecture imports, verifies selected-class-only Transformers loading, and checks errors for missing selected classes.
- Ruff checks passed for the changed tests.


test/python/builder/test_lazy_imports.py
```diff
@@ -0,0 +1,185 @@
...
+
+
+def test_unknown_builder_attribute_raises():
+    models_dir = Path(__file__).parents[3] / "src" / "python" / "py" / "models"
+    result = subprocess.run(
+        [
+            sys.executable,
+            "-c",
+            "import builders; assert not hasattr(builders, 'UnknownModel'); "
+            "assert not hasattr(builders, '__missing__')",
```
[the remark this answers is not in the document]
xieofxie: Fixed in c8390a03. Replaced the adjacent string literals in the argument list with an explicit multiline `script` variable, preserving the single `python -c` argument and both assertions without implicit concatenation. The targeted test passes.

src/python/py/models/builders/base.py
```diff
@@ -5796,30 +5784,33 @@ def load_weights(self, input_path):
             )
 
         else:
+            import transformers
```
kunal-vaishnavi, line 5787: The model builder is intentionally designed and supported against the latest `transformers` version. Given that policy, I’m concerned that introducing local imports and lazy package exports adds complexity in order to accommodate an older dependency version for a downstream consumer. Traditionally, global imports are preferred at the top of the file. Local imports are only used when necessary.

Why can’t Foundry Toolkit pin or upgrade its `transformers` dependency instead? I don’t think that downstream constraint should determine the import architecture here.

xieofxie: because there are recipes that using other passes together with model builder pass. To use latest olive, I need latest onnxruntime-genai, then latest transformers is needed. However, other pass's dependencies are incompatible with latest transformer and the model to be optimized also doesn't need the latest transformers.

PR #2597, opened 2026-09-20
[CUDA] Support CPU embedding offload with graph capture

tianleiwu:
## Description

Offload packed Engine model embedding lookups to a separate CPU-only ONNX session while preserving CUDA graph capture for target and DFlash/DSpark decoding. The target and drafter share one embedding session, and captured execution reuses persistent device buffers whose contents are refreshed before each graph invocation.

### Runtime

- Add a validated `CpuEmbedding` session that binds CUDA-bound output to pinned host memory and asynchronously copies only selected rows to the consumer device.
- Feed external embeddings into decoder and block-drafter sessions while preserving DFlash token IDs for candidate selection.
- Reserve persistent embedding buffers in the Engine CUDA graph plan and use existing step-tensor lifetime handling for the drafter.
- Reject unsupported non-Engine and non-CPU embedding configurations explicitly.

### Conversion and Documentation

- Add `split_cpu_embedding.py` to extract a packed `Gather` or `GatherBlockQuantized` without re-exporting or copying external weight data.
- Verify target/drafter lookup compatibility and remove offloaded initializers from shared-initializer configuration.
- Document configuration, graph-lifetime behavior, supported scope, tradeoffs, and validation guidance.

### Tests

- Cover float and INT4 graph splitting, retained selector IDs, external-data offsets, shared-weight validation, and unsafe split rejection.
- Add CPU and CUDA Engine acceptance coverage for embedding session lifetime, plus optional real-model parity across eager and captured CUDA execution.

## Testing

- `PYTHONPATH="$PWD/src/python/py" python -m pytest -q test/python/builder/test_split_cpu_embedding.py` (`6 passed`)
- `cmake --build build/cuda130/Release --target onnxruntime-genai onnxruntime-genai-cuda -j "$(nproc)"` (passed)
- Exact-source CUDA build and synthetic CPU/CUDA runtime parity (`2 passed`)
- The three real-model CUDA graph parameterizations require a matching ORT/model pair; the local ORT build rejected the available model's `uint8` PagedAttention input before embedding execution.
- `lintrunner -a -r origin/main` (passed)

## Checklist

- [x] Tests added
- [x] No breaking changes; CPU embedding remains opt-in
- [x] Documentation added


src/models/cpu_embedding.cpp
```diff
@@ -0,0 +1,82 @@
...
+  auto result = OrtValue::CreateTensor(*output_memory, host.data(), host.size_bytes(), shape, type_);
+  const char* input_name = config_.inputs.input_ids.c_str();
+  const char* output_name = config_.outputs.embeddings.c_str();
+  OrtValue* input_value = input.get();
+  OrtValue* output_value = result.get();
+  session_->Run(run_options_.get(), &input_name, &input_value, 1, &output_name, &output_value, 1);
+  // The pinned mirror has the destination buffer's lifetime, so this asynchronous copy can
+  // remain ordered with the consuming CUDA session without entering graph capture.
+  if (output.p_device_->GetType() == DeviceType::CUDA) {
+    bytes.CopyCpuToDevice();
```
baijumeswani: `bytes` owns the pinned host buffer returned by `CpuSpan()`, but `bytes` is local to this function. After this asynchronous copy is queued, that host buffer is freed when the function returns. This can force a wait on every lookup, and the code does not clearly keep the source alive until the copy finishes. It also means allocating and freeing pinned memory for every token step. Please keep a reusable pinned staging buffer with the target/DFlash workspace and retain it until the copy is complete. A CUDA event can be used if the buffer may be reused before the stream finishes.

src/dflash2_drafter.cpp
```diff
@@ -827,6 +833,11 @@ bool Dflash2Drafter::Propose(Tensor& aux_hidden_states, std::span<const Feed> fe
 
   auto& input_ids = StepTensor(step_tensors_.input_ids, device, Ort::TypeToTensorType<int64_t>,
                                {static_cast<int64_t>(num_block_rows)});
+  Tensor* embeddings = nullptr;
+  if (model_->cpu_embedding_) {
+    embeddings = &StepTensor(step_tensors_.embeddings, device, model_->cpu_embedding_->type_,
```
baijumeswani, line 850: This creates a persistent DFlash GPU buffer, but its memory is not included when the Engine decides how much memory the KV cache may use. Near the configured memory limit, the first proposal or a later buffer growth can fail because the cache has already claimed that space. Please reserve the maximum expected embedding-buffer size before `CacheManager::Create()`, or allocate the full workspace before probing free memory. The estimate should also allow for the old and new buffers to exist briefly during growth.

src/dflash2_drafter.cpp
```diff
@@ -345,10 +345,16 @@ ONNXTensorElementDataType ValidateDflash2ModelCompatibility(
...
 
-Dflash2Model::Dflash2Model(std::unique_ptr<Config> config, OrtEnv& ort_env)
-    : Model{std::move(config)} {
+Dflash2Model::Dflash2Model(std::unique_ptr<Config> config, OrtEnv& ort_env,
+                           std::shared_ptr<CpuEmbedding> cpu_embedding)
+    : Model{std::move(config)}, cpu_embedding_{std::move(cpu_embedding)} {
   session_ = CreateSession(ort_env, config_->model.decoder.filename, session_options_.get());
   session_info_.Add(*session_);
+  if (cpu_embedding_) {
+    cpu_embedding_->ValidateConsumer(session_info_, config_->model.dflash2.inputs.embeddings);
```
baijumeswani, line 358: This checks the embedding input's type and shape, but not whether its name is already used by another DFlash input. For example, if it matches `aux_hidden_states`, the same ONNX input name is later bound twice and the drafter fails at runtime. Please add the embedding name to the existing input-name uniqueness check and add a test for an alias.

src/python/py/models/split_cpu_embedding.py
```diff
@@ -0,0 +1,174 @@
...
+    initializers = {i.name: i for i in model.graph.initializer}
+    weight_names = {name for index, name in enumerate(node.input) if index != 1 and name}
+    if not weight_names <= initializers.keys():
+        raise ValueError("Embedding parameters must be initializers")
+    weight = initializers[node.input[0]]
+    if len(weight.dims) != 2 or weight.dims[1] != hidden_size:
+        raise ValueError("Embedding weight does not match hidden_size (packed UINT8 weights are not supported)")
+    dtype = initializers[node.input[2]].data_type if node.op_type == "GatherBlockQuantized" else weight.data_type
+    old_output = node.output[0]
+    if output_name in {i.name for i in model.graph.input} or any(output_name in n.output for n in model.graph.node):
```
baijumeswani: This check also sees the selected lookup node itself. If that node already outputs `inputs_embeds`, a valid model is rejected even though removing the node and reusing that name is safe. Please exclude the selected node from this check, or allow `old_output == output_name`, while still rejecting collisions with other nodes and graph inputs.

src/python/py/models/split_cpu_embedding.py
```diff
@@ -0,0 +1,174 @@
...
+            i for i in settings.get("shared_initializers", []) if i["name"] not in removed
+        ]
+        graphs[filename] = graph
+    model_config["embedding"] = {
+        "filename": "embedding.onnx",
+        "session_options": {"intra_op_num_threads": 1},
+        "inputs": {"input_ids": "input_ids"},
+        "outputs": {"inputs_embeds": "inputs_embeds"},
+    }
+    graphs["embedding.onnx"] = embedding
```
baijumeswani: This fixed name can overwrite a rewritten component graph in `graphs`. For example, a valid model may already use `embedding.onnx` as its decoder filename; this assignment then replaces the decoder with the one-node embedding graph while the config still points the decoder at that file. Please either reject filename collisions before conversion or choose an unused filename and store it in `model.embedding.filename`. A converter test with a decoder named `embedding.onnx` would cover this.

baijumeswani (commented):
Thanks for working on this. The split CPU session addresses a real current limitation: ORT can offload nodes to CPU in a mixed CPU/CUDA session, but a real CPU tensor followed by a host-to-device copy prevents CUDA Graph replay. Keeping the decoder CUDA-only is therefore a reasonable short-term design. Longer term, ORT should consider capturing and replaying the CUDA parts of a mixed session while running CPU work normally, but that would be a larger ORT execution-engine feature rather than a small change here.

I found five concerns, described inline. The most important is the lifetime of the pinned host buffer. Before making performance claims, please also compare GPU embedding with and without CUDA Graphs, ORT's normal mixed CPU/CUDA execution, and this split-session path. Please verify that CUDA Graph replay actually happens, not only that it is enabled.

src/models/cpu_embedding.cpp
tianleiwu: Fixed in 9621a0971caffa437eeecfba4343c415638a46df. The target decoder (including eager steps) and DFlash now each retain a pinned staging workspace. The host allocation is reused across destinations and smaller shapes, grows only when needed, and waits for outstanding uploads before reuse or teardown. Only active rows are uploaded. A real-CUDA regression covers pointer reuse, growth, alternating destinations, back-to-back uploads, and teardown; CPU/CUDA Engine parity also passes.

This uses the existing device stream synchronization API, not a CUDA event. The synchronization cost is documented, and the expected-speedup claim was removed. The requested four-way real-model benchmark and actual target/DFlash graph-replay trace are still pending; the three real-model tests were skipped because model paths were not supplied. Synthetic parity does not establish replay or latency improvement.

src/dflash2_drafter.cpp
tianleiwu: Fixed in 9621a0971caffa437eeecfba4343c415638a46df. Before CacheManager::Create, the Engine now reserves three times max_batch_size * block_size * hidden_size * element_size for the DFlash embedding workspace. This bounds the old allocation plus a doubled replacement during StepTensor growth, and applies to windowed/full-attention drafters and explicit block budgets. Checked arithmetic rejects overflow. The FP16/FP32 reservation and overflow regressions pass, along with the DFlash/decoder test suite.

src/dflash2_drafter.cpp
tianleiwu: Fixed in 9621a0971caffa437eeecfba4343c415638a46df. With CPU embedding enabled, the embedding input now joins the existing runtime input-name uniqueness set before cache inputs are validated. Empty names and aliases with ordinary or cache inputs are rejected before execution. The regression covers aux_hidden_states, input_ids, a cache input, and a valid distinct embedding name; all selected DFlash/decoder tests pass.

src/python/py/models/split_cpu_embedding.py
tianleiwu: Fixed in 9621a0971caffa437eeecfba4343c415638a46df. The collision check excludes only the selected lookup node, so an existing inputs_embeds output can be reused. Collisions with another node, graph input, or initializer still fail. Numeric parity tests now cover both output names for float/INT4 lookups, with and without retained selector token IDs. All 18 converter tests pass.

src/python/py/models/split_cpu_embedding.py
tianleiwu: Fixed in 9621a0971caffa437eeecfba4343c415638a46df. The converter chooses an unused embedding_N.onnx filename when embedding.onnx is occupied and records that exact filename in model.embedding.filename. It checks both source files and rewritten component names. The regression converts a decoder named embedding.onnx, loads both resulting ONNX sessions, and checks numerical parity and preservation of external data. All 18 converter tests pass.

baijumeswani (approved):

PR #2601, opened 2026-09-20
feat: add LFM2-Audio / LFM2.5-Audio speech input and output support

ykhrustalev:
**Problem**
LFM2-Audio and LFM2.5-Audio cannot be used here: nothing turns a sound into something the model can read or makes room for it in the prompt, nothing lets the model answer in speech, and the model builder cannot read their checkpoints, which have no model type and nest the LFM2 decoder inside the config.

**Solution**
- Adds an `lfm2_audio` pipeline (speech encoder → embedding → LFM2 decoder) with a C++ processor that computes NeMo's log-mel front end and swaps each `<|audio|>` marker for one placeholder per encoder frame; clips are resampled with a port of `torchaudio`'s windowed sinc, as the reference does
- Adds speech output (`model.audio_output`): the depthformer turns each decoder hidden state into a frame of eight audio codes whose embedding is the decoder's next input, switching between text and speech as `generate_sequential` and `generate_interleaved` do (`search.audio_interleaved`, `audio_temperature`, `audio_top_k`); the sequence keeps one placeholder per frame and the codes come from `get_output("audio_codes")`
- Reads these checkpoints in the model builder, by folder or by name, as `lfm2_audio` or the text-only `lfm2_audio_text`; a text-only build stops at `<|audio_start|>` and `<|text_end|>`
- Requires one marker per clip and one prompt at a time; refuses beam search and `adapter_path` (an adapter trained on these checkpoints is keyed `lfm.*`, so PEFT would attach untrained weights with only a warning)
- Touches shared code in two places: `Tokenizer::Decode` returns an empty string for no tokens instead of dying with SIGFPE, and `Images.open` / `Audios.open` no longer keep pointers into a growing path list
- Adds a tutorial: build, precisions, modes, speech output and its waveform, known limitations

**Testing**
- A tiny model under `test/models/lfm2-audio` runs the whole pipeline in CI against a plain ONNX Runtime replay, speech output included: tokens and audio codes must match the replay of the reference's text/speech switching exactly
- Against `liquid-audio`, fp32: ASR and chat agree to 6e-4 in the logits with every token the same on all three published checkpoints; with the audio taken greedily on both sides, TTS and interleaved answers match in every text token and every audio frame (200 items on the reference's `question.wav`)
- LiquidAI's published depthformer graph never applies its rotary embedding and builds the tables for the wrong `theta`, so seven of its eight codebooks disagree with the PyTorch model; the tutorial has the six-line fix, after which all eight agree to 1e-5
- Speech made with the model card's sampling reads back through this runtime's own ASR as the text the model wrote


src/models/multi_modal.cpp
```diff
@@ -603,6 +625,152 @@ DeviceSpan<float> SpeechState::Run(int current_length, DeviceSpan<int32_t>& next
...
+  auto clip_mel = OrtValue::CreateTensor(Ort::Allocator::GetWithDefaultOptions(),
+                                         std::vector<int64_t>{1, num_frames, bindings.num_mels}, bindings.mel_type);
+  std::memcpy(clip_mel->GetTensorMutableRawData(), mel_data + static_cast<size_t>(index) * clip_stride,
+              static_cast<size_t>(num_frames * bindings.num_mels) * mel_element_size);
+
+  auto clip_length = OrtValue::CreateTensor<int64_t>(Ort::Allocator::GetWithDefaultOptions(), std::vector<int64_t>{1});
+  clip_length->GetTensorMutableData<int64_t>()[0] = num_frames;
+
+  auto clip_features = OrtValue::CreateTensor(
+      model_.p_device_->GetAllocator(),
```
tianleiwu, line 700: `clip_features` is allocated with the decoder's `p_device_` but bound as an output of `speech_session_`, whose own `model.speech.session_options` may select another provider. The tutorial documents that CPU speech/embedding with a GPU decoder corrupts memory, and model construction does not reject that configuration. Could this allocate on the speech session's actual output device and explicitly copy to the next session, or reject mismatched providers until that path is supported? A mixed-provider regression test would also prevent this from recurring.

src/models/lfm2_audio_output.cpp
```diff
@@ -0,0 +1,255 @@
...
+  num_layers_ = keys_shape[0];
+  num_kv_heads_ = keys_shape[2];
+  head_size_ = keys_shape[4];
+
+  // The search has to pick the placeholder whatever its sampling settings are.
+  const size_t vocab_size = static_cast<size_t>(model_.config_->model.vocab_size);
+  placeholder_logits_ = model_.p_device_->Allocate<float>(vocab_size);
+  auto logits = placeholder_logits_.CpuSpan();
+  std::fill(logits.begin(), logits.end(), std::numeric_limits<float>::lowest());
+  logits[static_cast<size_t>(placeholder_token_id_)] = 0.0f;
```
tianleiwu, line 92: `audio_token_id` comes from configuration, but the processor only checks that it is positive. A value greater than or equal to `vocab_size` writes past this span. Please validate `0 < audio_token_id < vocab_size` before this access and add boundary tests.

src/models/lfm2_audio_output.cpp
```diff
@@ -0,0 +1,255 @@
...
+    const auto previous = MakeTensor<int64_t>(std::span<const int64_t>{&previous_code, 1}, {1});
+    const auto past_length = MakeTensor<int32_t>(std::span<const int32_t>{&step, 1}, {1});
+    const auto total_length = MakeScalar<int32_t>(step + 1);
+    const OrtValue* inputs[] = {hidden_tensor.get(), slices.get(), step_index.get(), previous.get(),
+                                keys.get(), values.get(), past_length.get(), total_length.get()};
+    auto outputs = model_.depthformer_session_->Run(nullptr, kDepthformerInputs, inputs, std::size(inputs),
+                                                    kDepthformerOutputs, std::size(kDepthformerOutputs));
+
+    const float* logits = outputs[0]->GetTensorData<float>();
+    previous_code = SampleCode({logits, static_cast<size_t>(config_.codebook_size)});
```
tianleiwu, line 196: `codebook_size` is accepted from JSON and used as the span length without checking the depthformer output shape. A negative value becomes a huge `size_t`, while an oversized positive value reads past the logits tensor. Please require a positive value and validate it against the logits output's final dimension before sampling, with negative and oversized configuration tests.

src/python/py/models/builders/lfm2.py
```diff
@@ -190,6 +194,110 @@ def update_genai_config(self, genai_config):
...
+        from transformers import Lfm2Config, PretrainedConfig
+
+        architecture = "Lfm2AudioForConditionalGeneration"
+        config_dict, _ = PretrainedConfig.get_config_dict(model_name_or_path, **kwargs)
+        if config_dict.get("architectures") != [architecture]:
+            return None
+
+        config = Lfm2Config(**config_dict["lfm"])
+        config.architectures = [architecture]
+        config._name_or_path = model_name_or_path
```
tianleiwu: The model-builder review rules prohibit newly added leading-underscore variables. Transformers exposes the public setter here, and `config.name_or_path = model_name_or_path` updates the same internal value. Please use that public property instead.

tianleiwu (changes requested):
The LFM2-Audio feature is comprehensive and generally well tested, and the latest revision addresses the earlier adapter, builder-structure, and documentation concerns. I am requesting changes for two memory-safety areas: submodel outputs must not be bound to buffers from a different execution provider, and configuration-derived indices/span lengths must be validated before memory access. I also left one model-builder convention comment.

src/models/multi_modal.cpp
ykhrustalev: Went with rejecting for now. #2605 is the general fix (allocate on the writing session's device and copy at the hand-off) for every pipeline model, so this PR doesn't duplicate it.

`CheckLfm2AudioSessionDevices` refuses a speech or embedding session left on CPU by its own `session_options` when it would be handed device memory: the embedding at load when the decoder's inputs are on the device, and both when a prompt carries audio. WebGPU without graph capture keeps the inputs on the host, so text-only prompts with a CPU export's config still run. Tests: `Lfm2AudioSessionDevicesTest` (no GPU needed) plus pipeline tests that run on CUDA and WebGPU; checked on an A10 and on WebGPU.

src/models/lfm2_audio_output.cpp
ykhrustalev: Fixed: `Lfm2AudioOutput` and the processor both require `0 < audio_token_id < vocab_size` now, with boundary tests for a negative id and for `vocab_size`.

src/models/lfm2_audio_output.cpp
ykhrustalev: Fixed: `codebook_size` must be positive at load, and each depthformer run checks it against the logits it actually produced (float, last dim >= `codebook_size`), which also covers a symbolic width. It's `>=` because one test samples from the first 16 of 2049 logits. Tests cover negative, zero, and one past the logits.

src/python/py/models/builders/lfm2.py
ykhrustalev: Done, it uses `config.name_or_path` now.

src/models/multi_modal.cpp
```diff
@@ -603,6 +668,154 @@ DeviceSpan<float> SpeechState::Run(int current_length, DeviceSpan<int32_t>& next
...
+
+  auto clip_mel = OrtValue::CreateTensor(Ort::Allocator::GetWithDefaultOptions(),
+                                         std::vector<int64_t>{1, num_frames, bindings.num_mels}, bindings.mel_type);
+  std::memcpy(clip_mel->GetTensorMutableRawData(), mel_data + static_cast<size_t>(index) * clip_stride,
+              static_cast<size_t>(num_frames * bindings.num_mels) * mel_element_size);
+
+  auto clip_length = OrtValue::CreateTensor<int64_t>(Ort::Allocator::GetWithDefaultOptions(), std::vector<int64_t>{1});
+  clip_length->GetTensorMutableData<int64_t>()[0] = num_frames;
+
+  auto clip_features = OrtValue::CreateTensor(
```
tianleiwu, line 742: The temporary `clip_features` output is allocated with the decoder-side allocator and then rebound as the `speech_session_` output. In mixed-provider setups this can create a device/allocator mismatch; please reject that configuration or allocate the temporary tensor on the speech session’s device before the run.

tianleiwu (commented):
The LFM2-Audio feature is comprehensive and generally well tested. I re-checked the current head and the earlier builder/documentation concerns are no longer present. I still recommend one follow-up on the per-clip speech encoder path: the temporary output tensor is allocated with the decoder allocator while it is rebound as the speech session output, so mixed-provider configurations should be rejected or allocated on the speech session’s device before execution.

src/models/multi_modal.cpp
tianleiwu: Non blocking. It's fine to be in a follow up PR.

tianleiwu (approved):

src/models/multi_modal.cpp
ykhrustalev: Thanks. Follow-up is #2616: the encoder output is now allocated on the speech session's own device and copied into the decoder-side features buffer (for a single clip too, when the devices differ). That lets a CPU encoder run next to a CUDA or WebGPU decoder instead of being refused. In this PR, the check at the top of `Lfm2AudioSpeechState::Run` already rejects that config before `clip_features` is allocated.

ykhrustalev:
@tianleiwu thank you for collaboration

PR #2607, opened 2026-09-22
fix: size fp32 logits buffer with last-token shape during prefill

amd-genmingz:
### Problem
`Logits::Get()` allocates the fp32 conversion buffer with the full prefill shape
instead of the last-token shape

`shape_` is `[batch_beam_size, sequence_length, vocab_size]`; `shape_last` is
`[batch_beam_size, 1, vocab_size]`.

**The buffer is never used**
`Cast()` discards any existing output whose shape differs from its input:
```cpp
if (output && shape != output->GetTensorTypeAndShapeInfo()->GetShape())
  output = nullptr;
if (!output)
  output = OrtValue::CreateTensor(device.GetAllocator(), shape, output_type);
```
Its input here is `output_last_tokens_`, shape `[B, 1, V]`. This branch is only
entered when `shape_[1] != 1`, so the shapes can never match and the oversized
buffer is always freed and reallocated at `[B, 1, V]`. It is allocated, never read,
never written, and immediately released.

### Impact
The waste is `batch_beam_size * sequence_length * vocab_size * 4` bytes per prefill,
scaling with vocabulary size. For Gemma-4 (`vocab_size = 262144`) that is 1 MB per
prompt token:
| prompt tokens | wasted allocation |
|---|---|
| 8319 | 8.72 GB |
| 12511 | 13.12 GB |
| 16375 | 17.17 GB |
| 32487 | 34.07 GB |


src/models/io/logits.cpp
```diff
@@ -45,7 +45,7 @@ DeviceSpan<float> Logits::Get() {
     output_last_tokens_ = OrtValue::CreateTensor(model_.p_device_logits_->GetAllocator(), shape_last, type_);
 
     if (type_ == Ort::TypeToTensorType<Ort::Float16_t> || type_ == Ort::TypeToTensorType<Ort::BFloat16_t>)
-      logits_of_last_token_fp32_ = OrtValue::CreateTensor<float>(model_.p_device_logits_->GetAllocator(), shape_);
+      logits_of_last_token_fp32_ = OrtValue::CreateTensor<float>(model_.p_device_logits_->GetAllocator(), shape_last);
```
tianleiwu: Could you add a regression check for the fp32 conversion buffer allocation on an unpruned fp16/bf16 model with a multi-token prompt (S > 1)? It should allocate [B, 1, V], not [B, S, V]. The existing fp16 greedy test passes even with the old oversized allocation because Cast discarded and reallocated that buffer, so a correctness-only assertion will not protect this memory improvement. A pruned [B, 1, V] model would cover the other graph layout.

tianleiwu (commented):
The allocation change is correct for the standard decoder: Cast now receives an fp32 buffer shaped like the selected last-token input. prune_lm_head is disabled by default, but this path already infers a fixed last-token output from ONNX metadata via Model::IsPruned(); the merged #2614 Engine path separately sizes full/selected packed rows and draft-verification rows from the graph contract. The inline test suggestion is non-blocking.

tianleiwu (approved):

src/models/io/logits.cpp
tianleiwu: Local test passed. I'll add a PR later.

